PR #14, opened 2016-12-12
Fix references, add helpful comments to StringChecker example

nickyr:


nickyr:
Thanks for providing that example, btw!

PR #25, opened 2016-12-16
Adding PHPCodeSniffer as a dev dependency.

erunion:
This just adds PHPCodeSniffer in as a dev dependency. I must have forgotten to add it in back when I did the PSR-2 work.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         }
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


muglug:
Closed bc I did it the same in https://github.com/vimeo/psalm/commit/35e3b175456ff29976888d90fe8746934e403f76

PR #65, opened 2017-01-16
Auto-create cache directory

### LINKED ISSUE #63: Support Windows
Splitting this from #60 to keep issues separate. You wrote last:

> That directory should be created here: https://github.com/vimeo/psalm/blob/dde39ed841598b06d8fb7a752c1393590a9b4568/src/Psalm/Checker/FileChecker.php#L563
but maybe it's failing due to nesting?

I added a `var_dump` directly in front of the `mkdir` and it is never printed to screen. My setup of psalm is endlessly simple, I just executed `composer require --dev vimeo/psalm` and executed `./vendor/bin/psalm` in my project. Obviously that means that I used version `0.3.12`. I wanted to use the current `dev-master` but that seems to be broken right now and always ends in:

```
Fatal error: Uncaught UnexpectedValueException: eesh in X:\some-project\vendor\vimeo\psalm\src\Psalm\Config.php:454
Stack trace:
#0 X:\some-project\vendor\vimeo\psalm\src\Psalm\Checker\ProjectChecker.php(334): Psalm\Config->isInProjectDirs('X:\\...')
#1 X:\some-project\vendor\vimeo\psalm\src\Psalm\Checker\ProjectChecker.php(219): Psalm\Checker\ProjectChecker->checkDirWithConfig('X:\\...', Object(Psalm\Config))
#2 X:\some-project\vendor\vimeo\psalm\bin\psalm(156): Psalm\Checker\ProjectChecker->check(false)
#3 {main}
  thrown in X:\some-project\vendor\vimeo\psalm\src\Psalm\Config.php on line 454
```

Fleshgrinder:
Closes #63

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -245,25 +245,14 @@ class Config
         }
 
         if (isset($config_xml['cacheDirectory'])) {
-            $path = (string)$config_xml['cacheDirectory'];
-
-            if ($path[0] !== '/') {
-                if (realpath(dirname($path)) !== dirname($path)) {
-                    $cache_base_dir = realpath(dirname($config->base_dir . $path));
-
-                    if (!$cache_base_dir) {
-                        die('Could not locate cache parent directory ' . dirname($config->base_dir . $path) . PHP_EOL);
-                    }
-
-                    $path = $cache_base_dir . DIRECTORY_SEPARATOR . basename($path);
-                }
-            } elseif (realpath(dirname($path)) === false) {
-                die('Could not locate cache parent directory ' . dirname($path) . PHP_EOL);
-            }
-
-            $config->cache_directory = $path;
+            $config->cache_directory = (string)$config_xml['cacheDirectory'];
         } else {
             $config->cache_directory = sys_get_temp_dir() . '/psalm';
+        }
+
+        if (@mkdir($config->cache_directory, 0777, true) === false && is_dir($config->cache_directory) === false) {
+            trigger_error('Could not create cache directory: ' . $config->cache_directory, E_USER_ERROR);
+            exit(255);
         }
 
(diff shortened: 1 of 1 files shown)


src/Psalm/Config.php
```diff
@@ -245,27 +245,16 @@ public static function loadFromXML($file_path, $file_contents)
...
-
-            $config->cache_directory = $path;
+            $config->cache_directory = (string)$config_xml['cacheDirectory'];
         } else {
             $config->cache_directory = sys_get_temp_dir() . '/psalm';
         }
 
+        if (@mkdir($config->cache_directory, 0777, true) === false && is_dir($config->cache_directory) === false) {
+            trigger_error('Could not create cache directory: ' . $config->cache_directory, E_USER_ERROR);
+            exit(255);
```
Fleshgrinder: One cannot really set the exit code in PHP (long standing bug) but PHP generally exits with 255 upon fatal errors and this is to ensure that this script exits (with the same code) if some weird settings are in place.

muglug:
Thanks! Is the path relative to the config file location, or the base directory that you run Psalm from?

At the moment all other paths in the config are relative to the config's location (which allows you to run Psalm in vendor packages without having to clone them into a separate directory).

Fleshgrinder:
If configured as such, however, I think that this behavior is wrong. Almost any executable honors the current working directory and yours should too. Regardless of the position of the configuration file or any other file. Only the directory the command was executed in should count (have a look at PHP's `getcwd()`).

muglug:
Fair point. I've committed https://github.com/vimeo/psalm/commit/b3726c9de50f0a02e116b5a4b542bbedfad06ac2 which makes everything relative to `getcwd()` and removes some extraneous code along the way.

PR #67, opened 2017-01-17
Simplify installation procedure

mbrodala:
Composer "require" implies "install", also suggest installing the latest stable version instead.


muglug:
While keeping dev-master in might lessen the number of bug reports, I agree that it makes it more stable overall. Thanks!

mbrodala:
Thanks. I already imagined something like that. ;-)

PR #72, opened 2017-01-19
Fixed Dependency Bounds

Fleshgrinder:
Current definition of dependencies would result in **any** version being installed that is greater than the one specified. This is problematic since new major releases imply breaking changes and nothing ensures that psalm will be compatible with those updated dependencies. Using an operator that instructs Composer to ensure that this does not happen is therefore inalienable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,8 +10,8 @@
         }
     ],
     "require": {
-        "php": ">=5.5",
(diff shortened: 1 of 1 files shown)


muglug:
Awesome, thanks!

PR #91, opened 2017-02-09
Fix Psalm errors in StringChecker example, add it to psalm.xml

nickyr:
StringChecker example is out of date again. I fixed the errors and added it to Psalm's psalm.xml file so this won't happen again. 

The TemplateChecker had a lot more issues, so I've just exempted it for now, but it'd be cool to add it to the XML also.


muglug:
Thanks!

I'll port over Vimeo-land changes from the template checker later today.

nickyr:
Sweet, thanks!

PR #141, opened 2017-04-16
composer: bump to PHP 5.6, as in travis

TomasVotruba:


muglug:
Any particular need for this?

TomasVotruba:
To be compatible with Travis: https://github.com/vimeo/psalm/blob/bd6a623bfd9cfde18d1b0bfd47dd2c0447785423/.travis.yml#L4

It doesn't make sense to have different min versions in those files.

SignpostMarv:
getting the tests to run on php 5.5 is a pain in the ass, not least because ReflectionParameter::isVariadic() isn't available on 5.5, so the tests have to be skipped anyway.

muglug:
Makes sense. Thanks!

TomasVotruba:
Thank you

PR #147, opened 2017-04-24
Rewriting and streamlining every unit test with data providers.

erunion:
*This is a massive PR. I'm sorry.*

### Summary
I've gone through every unit test to rewrite, and streamline them with a consistent PHPUnit dataProvider framework.

### Why
The main problem that I found with the existing test suite was that every test was full of the same basic test case:

1. Take a piece of code.
2. Pass it into `$stmts = self::$parser->parse($code);`
3. Run `(new FileChecker('somefile.php', $this->project_checker, $stmts))->visitAndAnalyzeMethods();`
4. If we are looking for failures, verify that a `\Psalm\Exception\CodeException` exception was thrown.

With this, it didn't make any sense to not do this as the maintainability gains are immediately apparent. When you need to change a use case, you only need to change it once instead of messing with IDE refactoring. 

Having a common framework in place for tests also makes it easier for new people to come in and add/fix/update tests for new features and bug fixes.

Frankly, before this, it was incredibly daunting.

And less important to maintainability and overall friendliness, doing this also removed a **lot** of repeated test code that added upwards of 100KB+.

### Details
Since the majority of the unit test suite was doing a code parse and `FileChecker->visitAndAnalyzeMethods` analysis, I've pulled these into a `FileCheckerInvalidCodeParseTestTrait` and `FileCheckerValidCodeParseTestTrait` trait. These do as they're named. Given a piece of code and some options, they either check that it's valid, or invalid (throws exceptions).

#### Framework
Since these tests are now utilizing a common framework for validation and invalidation of code, there needed to be some special cases written in to support PHP 7 and skipped tests with the data providers.

* For test cases that require >= PHP7, you can prefix the data provider name with `PHP7-`. If the test is not run on PHP 7, it is marked as skipped with the following message: "Test case requires PHP 7."
* For tests that need to be skipped, you can do the same, but with `SKIPPED-`. They will then be marked as skipped with the following message: "Skipped due to a bug."

### DIFF
--- a/tests/AnnotationTest.php
+++ b/tests/AnnotationTest.php
@@ -1,218 +1,13 @@
 <?php
 namespace Psalm\Tests;
 
-use PhpParser\ParserFactory;
-use PHPUnit_Framework_TestCase;
 use Psalm\Checker\FileChecker;
-use Psalm\Config;
 use Psalm\Context;
 
-class AnnotationTest extends PHPUnit_Framework_TestCase
+class AnnotationTest extends TestCase
 {
-    /** @var \PhpParser\Parser */
-    protected static $parser;
-
-    /** @var TestConfig */
-    protected static $config;
-
-    /** @var \Psalm\Checker\ProjectChecker */
-    protected $project_checker;
-
-    /**
-     * @return void
-     */
-    public static function setUpBeforeClass()
-    {
-        self::$parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
-        self::$config = new TestConfig();
-    }
-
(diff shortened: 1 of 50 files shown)


muglug:
This is great.

You’re gonna hate me, but: the only thing I dislike is the dashes in array keys for different tests, mainly because you also have as an array key `error_message`. Instead, could you have them be camelCase, as if to say, these were once separate test methods, and can be treated as such.

erunion:
@muglug There you go. There were only ~600 instances to change, so it wasn't too bad.

PR #149, opened 2017-04-25
PSR-2 adjustments and adding PHPCS checks into the Travis config.

erunion:
This makes some minor adjustments to comply with PSR-2, as well as adding an execution of PHPCS in the Travis config so these things are caught in tests.

### DIFF
--- a/src/Psalm/Checker/FunctionChecker.php
+++ b/src/Psalm/Checker/FunctionChecker.php
@@ -98,7 +98,9 @@ class FunctionChecker extends FunctionLikeChecker
         $file_storage = FileChecker::$storage[$file_path];
 
         if (!isset($file_storage->functions[$function_id])) {
-            throw new \UnexpectedValueException('Not expecting ' . $function_id . ' to not have storage in ' . $file_path);
+            throw new \UnexpectedValueException(
+                'Not expecting ' . $function_id . ' to not have storage in ' . $file_path
+            );
         }
 
         return $file_storage->functions[$function_id];
@@ -204,7 +206,12 @@ class FunctionChecker extends FunctionLikeChecker
         $function_type_options = [];
 
         if (!self::$callmap_file_checker) {
-            self::$callmap_file_checker = new FileChecker('callmap.php', \Psalm\Checker\ProjectChecker::getInstance(), [], false);
+            self::$callmap_file_checker = new FileChecker(
+                'callmap.php',
+                \Psalm\Checker\ProjectChecker::getInstance(),
+                [],
+                false
+            );
         }
 
         foreach ($call_map_functions as $call_map_function_args) {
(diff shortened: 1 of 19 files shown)


src/Psalm/Checker/FunctionChecker.php
```diff
@@ -602,24 +614,28 @@ public static function getFQFunctionNameFromString($function_name, StatementsSou
...
-                return $imported_function_namespaces[strtolower($first_namespace)] . '\\' . implode('\\', $function_name_parts);
+            if (isset($imported_function_namespaces[$first_namespace_lcase])) {
+                return $imported_function_namespaces[$first_namespace_lcase] . '\\' .
+                    implode('\\', $function_name_parts);
             }
-        } elseif (isset($imported_namespaces[strtolower($function_name)])) {
-            return $imported_namespaces[strtolower($function_name)];
-        } elseif (isset($imported_function_namespaces[strtolower($function_name)])) {
-            return $imported_function_namespaces[strtolower($function_name)];
+        } elseif (isset($imported_namespaces[$function_name_lcase])) {
```
erunion: These weren't a PSR-2 thing, but it just seemed wasteful to be running `strtolower` every time here.

src/Psalm/Checker/Statements/Expression/CallChecker.php
```diff
@@ -318,11 +318,11 @@ public static function analyzeFunctionCall(
                     $stmt->args,
                     $method_id,
                     $function_params,
+                    $code_location,
```
erunion: These arguments were shifted around because PSR-2 requires that method arguments with default values go at the end of the list.

muglug: Instead of this, could you just put the type into the docblock? It's designed to be a nullable type, not an optional one. It was bad of me to abuse the syntax, but :man_shrugging:

erunion: Not quite sure I understand. You want `code_location` to be nullable?

muglug: No – I want the docblock to be totally unchanged – just remove the type hinting in on the parameter types:

```php
/**
     * @param   StatementsChecker                       $statements_checker
     * @param   array<int, PhpParser\Node\Arg>          $args
     * @param   string|null                             $method_id
     * @param   array<int,FunctionLikeParameter>        $function_params
     * @param   FunctionLikeStorage                     $function_storage
     * @param   ClassLikeStorage                        $class_storage
     * @param   array<string, Type\Union>|null          $generic_params
     * @param   CodeLocation                            $code_location
     * @param   bool                                    $check_variables
     * @return  false|null
     */
    protected static function checkFunctionArgumentsMatch(
        StatementsChecker $statements_checker,
        array $args,
        $method_id,
        array $function_params,
        $function_storage,
        $class_storage,
        &$generic_params,
        CodeLocation $code_location,
        $check_variables
    ) {
```

erunion:
@muglug Pushed up those adjustments.

muglug:
Thanks!

Issue #84, opened 2017-01-31
Exclude files based on pattern


Mte90:
Actually it's not possible to use patterns like:
```
<directory name="*lib*" />
```

muglug:
The second example `<directory name="/vendor/*" />` is redundant (as `<directory name="/vendor" />` has the same effect.

But I'll add support for wildcard matching.

Mte90:
yes, for the second one you are right, I update the first commit

fesor:
And how about `file` patterns? For example I want to supress specific issues for specific files, all of which has some kind of suffix.

```
<file name="*DTO.php">
```

muglug:
@fesor have added that (and directory pattern matching works as well)

kkmuffme:
Doesnt seem to work?

```
<UndefinedGlobalVariable>
	<errorLevel type="suppress">
		<directory name="*/templates" />
	</errorLevel>
</UndefinedGlobalVariable>
```

Results in:
Could not resolve config path to /var/www/vhosts/somepath//*/templates

olleharstedt:
Hm this is not working very well for me. Wanting to ignore all files in different `views/` folders:

            <directory name="*/views/*" />

What's the correct syntax?

ctf0:
@quisse @olleharstedt after search issues and try & error, 
i found that atm the pattern wont work unless all the scanned directories have that `thing` u r looking for, so for example `<directory name="*/views/*" />` wont work because any directory psalm will search will have to have the `views` folder or it wont work.

same goes for `*.phtml`, all the directories will have to have this file present or it wont work.

imo wild cards should be treated differently from absolute/fixed paths.

@muglug plz correct me if am wrong or missing something

mringler:
Could not figure it out from documentation or comments. But this works:
```xml
                        <ignoreFiles>
                            <directory name="src/**/Checker" />
                            <file name="src/Psalm/Checker/*Checker.php" />
                        </ignoreFiles>
```

PR #150, opened 2017-04-26
added quick globbing support for includes, does not satisfy vimeo/psalm#84 but is sort of related

SignpostMarv:
Because I wanted to do this:
```xml
<?xml version="1.0"?>
<psalm
    name="Example Psalm config with recommended defaults"
    stopOnFirstError="false"
    useDocblockTypes="true"
    totallyTyped="false"
>
    <projectFiles>
        <directory name="./vendor/foo/*/src" />
    </projectFiles>
```

### DIFF
--- a/src/Psalm/Config/FileFilter.php
+++ b/src/Psalm/Config/FileFilter.php
@@ -59,6 +59,23 @@ class FileFilter
         if ($e->directory) {
             /** @var \SimpleXMLElement $directory */
             foreach ($e->directory as $directory) {
+                if (strpos((string)$directory['name'], '*') !== false) {
+                    $globs = array_map(
+                        'realpath',
+                        array_filter(
+                            glob(getcwd() . DIRECTORY_SEPARATOR . (string)($directory['name'])),
+                            'is_dir'
+                        )
+                    );
+                    foreach ($globs as $globIndex => $directory_path) {
+                        if (!$directory_path) {
+                            die('Could not resolve config path to ' . getcwd() . DIRECTORY_SEPARATOR .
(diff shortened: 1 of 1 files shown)


src/Psalm/Config/FileFilter.php
```diff
@@ -59,6 +59,23 @@ public static function loadFromXMLElement(
         if ($e->directory) {
             /** @var \SimpleXMLElement $directory */
             foreach ($e->directory as $directory) {
+                if (strpos((string)$directory['name'], '*') !== false) {
```
muglug: How about
```php
$prospective_directory_path = getcwd() . DIRECTORY_SEPARATOR . $directory['name'];
if (strpos($prospective_directory_path, '*') !== false) {
```
and then you can use `$prospective_directory_path` below

src/Psalm/Config/FileFilter.php
```diff
@@ -59,6 +59,23 @@ public static function loadFromXMLElement(
...
+                    $globs = array_map(
+                        'realpath',
+                        array_filter(
+                            glob(getcwd() . DIRECTORY_SEPARATOR . (string)($directory['name'])),
+                            'is_dir'
+                        )
+                    );
+                    foreach ($globs as $globIndex => $directory_path) {
```
muglug: could you use `$snake_case` here please?

SignpostMarv: addressed in 4c181f2

src/Psalm/Config/FileFilter.php
SignpostMarv: addressed in 1d6f654

PR #154, opened 2017-05-03
Allow Psalm to be run outside of the directory it's installed in

### LINKED ISSUE #132: Psalm as a global referenced dependency
Is there a way to install Psalm as a global referenced dependency?

If I run the following code snippet on a PHP 7.1 docker image, it keeps telling me that "class or interface <xxx> does not exist":
```
composer global require vimeo/psalm
~/.composer/vendor/bin/psalm
```

Might it be that psalm is using the composer instance it is installed to to autoload classes? Or am I missing something else?

muglug:
Fixes #132

Running Psalm with `--root` will change the directory it runs on. This allows you to install Psalm globally and run it anywhere, as long as the target directory uses the composer autoloader.

### DIFF
--- a/bin/psalm
+++ b/bin/psalm
@@ -61,7 +61,7 @@ Options:
         Enable monochrome output
 
     -r, --root
-        If running as a global
+        If running Psalm globally you'll need to specify a project root. Defaults to cwd
 
     --show-info[=BOOLEAN]
         Show non-exception parser findings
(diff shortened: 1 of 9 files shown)


bin/psalm
```diff
@@ -70,6 +60,9 @@ Options:
     -m, --monochrome
         Enable monochrome output
 
+    -r, --root
+        If running Psalm globally you'll need to specify a project route. Defaults to cwd
```
TysonAndre: "root", not route?

muglug: Thanks – dealing with routing in my day job...

MetalArend:
Big thanks!

PR #157, opened 2017-05-15
Fix false UndefinedClass for resource used in namespace

phaldan:
Using `resource` as return type of a function or a class method withing a custom namespace will cause a UndefinedClass error. I have added a testcase for this issue and also tried to find a solution. Feel free to improve my solution.

My Setup: PHP 7.1.4 via docker php:7-alpine

### DIFF
--- a/tests/ReturnTypeTest.php
+++ b/tests/ReturnTypeTest.php
@@ -339,6 +339,14 @@ class ReturnTypeTest extends TestCase
                             return null;
                         }
                     }'
+            ],
+            'resourceReturnType' => [
+                '<?php
+                    namespace bar;
+
+                    function fooFoo() : ?resource {
+                        return null;
+                    }'
             ]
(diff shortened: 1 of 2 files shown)


tests/ReturnTypeTest.php
```diff
@@ -339,6 +339,14 @@ function getBarWithIsset() {
                             return null;
                         }
                     }'
+            ],
+            'resourceReturnType' => [
+                '<?php
+                    namespace bar;
+
+                    function fooFoo() : ?resource {
```
muglug: `resource` is not a valid return type in PHP 7.1. Use `/** @return ?resource **/` instead.

phaldan:
Oh I totally missed that resource cannot be used as return type. Looks like this is more a bug in php. Even the error message in #158 looks kind of buggy.

PR #163, opened 2017-05-27
adding editorconfig that seems to match the php files

SignpostMarv:
just a teensy little touch, stops supporting editors from accidentally messing with the desired indentation etc. (i.e. in my setup, the default indentation in Atom is 2, but the PHP indentation is definitely 4) :D

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*.php]
+indent_style = space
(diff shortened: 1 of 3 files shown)


SignpostMarv:
re: supporting editors: http://editorconfig.org/#download

muglug:
Thanks!

PR #206, opened 2017-08-27
Add XML as possible output format + add report generation

MacFJA:
Add a new output format: `xml`

Add an option to create a log/report file, available in 3 format: XML, JSON and emacs.  
The console options ~are~<ins>is</ins>:
- ~`--report-xml=PATH_TO_THE_FILE`~
- ~`--report-json=PATH_TO_THE_FILE`~
- ~`--report-emacs=PATH_TO_THE_FILE`~
- <ins>`--report=PATH_TO_FILE` (The file extension define the format)</ins>

~The option `--report-emacs` is not in the console help, as the format `emacs` is not documented in the wiki~

### DIFF
--- a/bin/psalm
+++ b/bin/psalm
@@ -15,7 +15,7 @@ $options = getopt(
         'help', 'debug', 'config:', 'monochrome', 'show-info:', 'diff',
         'file:', 'self-check', 'update-docblocks', 'output-format:',
         'find-dead-code', 'init', 'find-references-to:', 'root:', 'threads:',
-        'report-xml:', 'report-json:', 'report-emacs:'
+        'report:'
     ]
 );
 
@@ -89,11 +89,9 @@ Options:
     --threads=INT
         If greater than one, Psalm will run analysis on multiple threads, speeding things up.
 
-    --report-xml=PATH
-        The path where to output the XML report file
-
-    --report-json=PATH
-        The path where to output the JSON report file
+    --report=PATH
+        The path where to output report file. The output format is base on the file extension.
+        (Currently supported format: ".json", ".xml", ".txt")
 
 HELP;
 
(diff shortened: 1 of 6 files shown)


bin/psalm
```diff
@@ -14,7 +14,8 @@ $options = getopt(
     [
         'help', 'debug', 'config:', 'monochrome', 'show-info:', 'diff',
         'file:', 'self-check', 'update-docblocks', 'output-format:',
-        'find-dead-code', 'init', 'find-references-to:', 'root:', 'threads:'
+        'find-dead-code', 'init', 'find-references-to:', 'root:', 'threads:',
+        'report-xml:', 'report-json:', 'report-emacs:'
```
muglug: Could we shorten this just to `--report=filename.xml` and use the `--output-format=xml` to specify the type. That way the list of commands is more compact and composable.

MacFJA: That was how I write it first, but the `--output-format` will be use for two different things (and it force the console output to be the same as the report output).  
For example, the XML output is nice for external tool (like a **ci**) but not really readable while the `console` output is the opposite.

I guess a can add a `--report-format` option (and we lose the ability to have multiple report, but I don't really see usage of it)

muglug: Ohhh. So you have a scenario where you want to output multiple types of report at the same time?

MacFJA: I can't find any real usage of a multiple report output.  
I guess it can be useful if you send your result into different _post processing/aggregator_ tool that don't read the same type of data (a java based tool is more likely to use XML, where a Javascript one prefer to use JSON)

This options is available on PHPMD, PHPUnit, phploc, phpcs, PhpMetrics. (It's mainly because it's found on several tools that I came up with this solution)

muglug: `phpcs` doesn't seem to have separate commands for different file formats (though PHPUnit does)

MacFJA:
- Change the console options (only one option now, report format is guessed on the file extension)
- Add some tests

muglug:
Thanks!

PR #208, opened 2017-09-10
Fix some cURL function return values

fkooman:
At least the functions referenced here can also return boolean
`false` according to the PHP manual. Add this information to the
CallMap file.


muglug:
Thanks!

PR #209, opened 2017-09-12
fix json_encode return values

fkooman:
`json_encode` can also return `false`


fkooman:
```
$ bin/psalm

ERROR: InvalidScalarArgument - src/Psalm/Clause.php:90:20 - Argument 1 of md5 expects string, string|false provided

        return md5(json_encode($this->possibilities)) .

```

Heh ;-)

muglug:
Could you add a check for `false` there?

Something like

```php
$possibility_string = json_encode($this->possibilities);
if (!$possibility_string) {
  throw new \LogicException('Error converting possibilities to json');
}
return md5($possibility_string);
```

fkooman:
I'm trying to do that, but running into another issue when running Psalm itself, unrelated to my changes:

```
ERROR: PropertyNotSetInConstructor - src/Psalm/Checker/ClassLikeChecker.php:140:15 - Property Psalm\Checker\TraitChecker::$storage is not defined in constructor of Psalm\Checker\TraitChecker or in any private methods called in the constructor
    protected $storage;
```

fkooman:
It seems travis-ci is fine with running `bin/psalm` on itself. So yeah, not sure what to think of #212

muglug:
oops sorry

muglug:
Thanks for this!

PR #219, opened 2017-09-22
Adjusted signatures of DateTime and friends

weirdan:
* `__wakeup` is `void -> void`, see [docs](http://php.net/manual/en/language.oop5.magic.php#object.wakeup)
* `DateTime::__construct()` does not accept `null` as time
* `DateTimeZone::getLocation()` has no last `$object` parameter
* ditto for `DateTimeZone::getName()`
* ditto for `DateTimeZone::getTransitions()`


muglug:
Thanks!

PR #221, opened 2017-09-26
Fix case of ZipArchive properties

steffenweber:
The properties "numFiles" and "statusSys" use camelCase.

https://secure.php.net/manual/en/class.ziparchive.php


muglug:
Thanks!

PR #222, opened 2017-09-26
Fix not generated report

MacFJA:
The previous code don't generate report file if there are no errors nor warnings.

Which:
 - can lead to false positive if the previous file is not removed (because last file always contains at least one issue)
 - can break CI because the report file don't exist if the previous file is purge at each run


muglug:
Thanks!

PR #226, opened 2017-09-29
Adding a `--version` argument to the Psalm executable.

erunion:
```
$ ./bin/psalm --version
Psalm 0.2.63

```

When you tag releases, you'll need to remember to update this for 0.2.x and 0.3.x.


weirdan:
Wouldn't it be more maintainable to query current version through [ocramius/package-versions](https://packagist.org/packages/ocramius/package-versions) instead of hardcoding it in a constant value?

erunion:
Oh that's a clever package. That's a good idea.

erunion:
Pushed up the change, though ocramius/package-versions requires >=7.0, so we might need to make this available only on Psalm 0.3.x (or a 0.4.x).

muglug:
Could you instead fork that package to be PHP 5.6-compatible? I think that's the best way forward, because it keeps the minimum dep for Psalm as 5.6, which I'd prefer.

Psalm is used by a lot of existing packages that want to keep preserver support for PHP 5, so it's the only static analysis tool available to them (as the other major ones all require PHP 7).

The changes required to fork that package are minimal, I think – you can keep the generators, but replace `??` with `isset` checks.

weirdan:
There are existing backports already available, here's one I found: https://github.com/samsonasik/PackageVersions

erunion:
Backported to https://github.com/samsonasik/PackageVersions.

Everything should be passing and PHP 5.6 compatible now.

muglug:
Thanks!

weirdan:
When psalm is installed globally (`composer global require vimeo/psalm`) this seems to be breaking `composer install` for packages that don't require `samsonasik/package-versions`.
This problem was fixed in upstream (see Ocramius/PackageVersions#38 and Ocramius/PackageVersions#39), but apparently hasn't been backported to @samsonasik's fork.

muglug:
@weirdan I've created my own fork at https://github.com/muglug/PackageVersions that has the above bugfix, and Psalm now uses that fork.

erunion:
Thanks @muglug for fixing that.

wuethrich44:
@muglug I would like to install psalm in a Symfony app which already uses `ocramius/package-versions`. When I try to execute `composer require --dev vimeo/psalm` I get this error:
```

  Problem 1
    - don't install ocramius/package-versions 1.2.0|don't install muglug/package-versions 1.2.1
    - don't install muglug/package-versions 1.2.1|remove ocramius/package-versions 1.2.0
    - don't install muglug/package-versions 1.2.1|don't install ocramius/package-versions 1.2.0
    - vimeo/psalm 0.3.70 requires muglug/package-versions ^1.2.1 -> satisfiable by muglug/package-versions[1.2.1].
    - Installation request for vimeo/psalm ^0.3.70 -> satisfiable by vimeo/psalm[0.3.70].
    - Installation request for ocramius/package-versions (locked at 1.2.0) -> satisfiable by ocramius/package-versions[1.2.0].
```
Is there a way to fix this?

muglug:
Hmm aside from just using `muglug/package-versions` instead? Not as far as I know.

wuethrich44:
I had to add `"muglug/package-versions": "^1.2"` manually to my `composer.json` then I could install `vimeo/psalm`. 

It's strange that `composer require muglug/package-versions` didn't work...

weirdan:
Given that it has the same namespace and classes, what if `muglug/package-versions` provided `ocramius/package-versions` instead of replacing it, and `vimeo/psalm` depended on `ocramius/package-versions`? This way, I imagine, `muglug...` would get installed on PHP 5.6 (assuming it has only than in platform reqs) and `ocramius...` on PHP 7.0. Failing that, a couple of intermediate packages with provides could also solve this (theoretically):
```
                                        ----> package-versions-legacy (provides package-versions-virtual) ----> muglug/package-versions ----> php:5.6 
vimeo/psalm -> package-versions-virtual-|
                                        ----> package-versions-current (provides package-versions-virtual) ----> ocramius/package-versions ----> php:7.0+
```

muglug:
@fabwu @weirdan Psalm now uses `muglug/package-versions-56` which does not replace `ocramius/package-versions`, but can live along side it.

PR #227, opened 2017-09-29
Stripping preceding slashes on class names in the CallChecker.

erunion:
Since `ReflectionClass` resolves class names to the non-global namespace, we need to assert that `CallChecker` is assuming it the same. Not doing this results in the `ClassLikeStorage` system not being able to find some classes in its local store.

I uncovered this while running Psalm on a fresh Laravel application.


muglug:
I think this issue should be fixed by https://github.com/vimeo/psalm/commit/1cc13bdeebd0f1151451256d070cc2bd392be87c

erunion:
Yep that fixed it.

PR #237, opened 2017-10-12
Adjusted Closure methods signatures

weirdan:
This fixes the following false-positive:
```php
function a(): bool {
  $a = function(): bool { return true; }
  return $a->call(new stdClass);
}
```
where psalm previously complained about (wrongly) inferred type to be Closure.


src/Psalm/CallMap.php
```diff
@@ -848,9 +848,9 @@
 'closelog' => ['bool'],
 'Closure::bind' => ['Closure', 'old'=>'Closure', 'to'=>'object|null', 'scope='=>'object|string'],
 'Closure::bindTo' => ['Closure', 'new'=>'object|null', 'newscope='=>'object|string'],
-'Closure::call' => ['Closure', 'to'=>'object', 'parameter='=>'', '...='=>''],
+'Closure::call' => ['mixed', 'to'=>'object', 'parameter='=>'mixed', '...='=>'mixed'],
```
muglug: for brevity, `''` means `mixed`, so you can remove the *other* changes here

muglug:
Thanks! I need to make this a special case so it behaves like `$a()` when the function is defined locally, as above

weirdan:
Will it properly consider that `new stdClass` as `$this` inside that closure? Otherwise, IMHO, this might cause more harm than good.

PR #238, opened 2017-10-12
Corrected case of Traversable typehints

weirdan:
Affected functions/methods:
* `iterator_apply()`
* `iterator_count()`
* `IteratorIterator::__construct()`
* `iterator_to_array()`

This fixes the following false-positive:
```php
class Iter extends IteratorIterator {
    public function __construct(Traversable $a) {
	parent::__construct($a);
    }
}
```
where psalm previously wanted `parent::__construct()` argument to be
`traversable` (all-lowercase).


muglug:
Thanks!

Issue #260, opened 2017-10-25
Too few arguments for method gmp_import


steffenweber:
Thank you for this useful tool!

I've found a small bug. Psalm thinks that the function [gmp_import](http://php.net/gmp_import) requires more than one argument although its second and third parameters have default values according to the PHP documentation.

```php
<?php
gmp_import(inet_pton('127.0.0.1'));
```

```
$ php vendor/bin/psalm test.php 
ERROR: TooFewArguments - test.php:2:1 - Too few arguments for method gmp_import
gmp_import(inet_pton('127.0.0.1'));
```

I think this might not be a bug in Psalm itself but in one of its dependencies that tells Psalm the signatures of PHP functions. But I've not found such a dependency. Please let me know if I should file the bug elsewhere.

steffenweber:
Simlarly, Psalm does not know about the argument of "MaxMind\Db\Reader::__construct" and reports a "TooManyArguments" error.

steffenweber:
Hmm, I used GitHub to search for "gmp_import" in this repository but for some reason it did not find a match in [Callmap.php](https://github.com/vimeo/psalm/blob/master/src/Psalm/CallMap.php). I'll submit a pull request.

PR #261, opened 2017-10-25
Fix signature of "gmp_import"

steffenweber:
According to the PHP manual, the second and third argument of "gmp_import" have default values and are therefore optional.

https://php.net/gmp_import

Fixes https://github.com/vimeo/psalm/issues/260

### DIFF
--- a/src/Psalm/CallMap.php
+++ b/src/Psalm/CallMap.php
@@ -2703,7 +2703,7 @@ return [
 'gmp_gcdext' => ['array', 'a'=>'', 'b'=>''],
 'gmp_gcd' => ['GMP', 'a'=>'', 'b'=>''],
 'gmp_hamdist' => ['int', 'a'=>'', 'b'=>''],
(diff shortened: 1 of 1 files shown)


src/Psalm/CallMap.php
```diff
@@ -2703,7 +2703,7 @@
 'gmp_gcdext' => ['array', 'a'=>'', 'b'=>''],
 'gmp_gcd' => ['GMP', 'a'=>'', 'b'=>''],
 'gmp_hamdist' => ['int', 'a'=>'', 'b'=>''],
-'gmp_import' => ['GMP', 'data'=>'string', 'word_size'=>'int', 'options'=>'int'],
+'gmp_import' => ['GMP', 'data'=>'string', 'word_size'=>'?int', 'options'=>'?int'],
```
weirdan: Shouldn't this be `'word_size=' => 'int', 'options=' => 'int'` ? These params are optional, but documentation does not imply they are nullable (rather, they have a default int value).

steffenweber:
Oops, fixed. Thank you!

muglug:
Thanks!

PR #265, opened 2017-11-07
Fix build failure - MixedPropertyAssignment should be expected in test.

TysonAndre:
My understanding: This is assigning to a property of type mixed, and should warn?


muglug:
Thanks – this wasn't the reason for the build failure, but it's a good catch. The reason was some _incredibly_ sloppy commit of a half-finished feature that really should have gone in a branch.

PR #273, opened 2017-11-09
Fix phpdoc in psalm's codebase referring to missing params.

TysonAndre:
Also, don't use `@return self` in __construct, it returns void.


muglug:
Thanks. A while back I made a change that stopped Psalm complaining about docblock params that were missing from the signature, or docblock params that were less specific than the signature typehints, because Psalm could work around those issues. But it makes sense that they should nevertheless be accurate everywhere in Psalm's codebase.

PR #287, opened 2017-11-12
Fix incorrect strpos argument order

TysonAndre:
Detected via static analysis heuristic

https://secure.php.net/strpos documents it as `mixed strpos ( string $haystack , mixed $needle [, int $offset = 0 ] )`

### DIFF
--- a/src/Psalm/Checker/Statements/Expression/CallChecker.php
+++ b/src/Psalm/Checker/Statements/Expression/CallChecker.php
@@ -1738,7 +1738,7 @@ class CallChecker
                         $arg->value,
                         $by_ref_type,
                         $context,
-                        $method_id && (strpos('::', $method_id) !== false || !FunctionChecker::inCallMap($method_id))
+                        $method_id && (strpos($method_id, '::') !== false || !FunctionChecker::inCallMap($method_id))
                     );
                 } else {
                     if ($arg->value instanceof PhpParser\Node\Expr\Variable) {


muglug:
Amazing find. Where was the heuristic? Just the assumption that a short string should be the needle?

TysonAndre:
https://github.com/phan/phan/commit/d3cf41850a9e975998cb87fc50d66bf93fbed062#diff-701fafd21136d915e767dbfa02266f3fR104

Assumption that the "needle" should be a short string, a global/class constant, or a concatenation, and that the "haystack" shoul not be any of those.

If both of those were wrong, it warns that the "needle" might be in the wrong order in that call

muglug:
Very cool. I noticed a `/** @suppress */` annotation there – does `php-ast` now include statement docblocks?

TysonAndre:
It doesn't; That was on a closure.

Statement docblocks are planned, but would require tolerant-php-parser-to-php-ast to be used exclusively (or php-parser-to-php-ast)


- And cache the ast\Node on disk, with igbinary if available. Haven't checked the impact on performance.
- Parsing standalone heredocs or strings would be a hackish workaround.

```php
        /**
         * @suppress PhanParamSuspiciousOrder 100% deliberate use of varying regex and constant $subject for preg_match
         */
         $err = with_disabled_phan_error_handler(function() use($pattern) {  // <-- This is a closure
```

Issue #297, opened 2017-11-14
Uncaught DocBlockParseException for invalid inline var comment


TysonAndre:
This has an invalid union type in an inline doc comment.
Observed: the inline doc comment results in an uncaught DocBlockParseException.
Expected: An INFO level notice about the doc block.

A fix is in progress.

```php
<?php

/**
 * @return array(string) This is invalid PHPdoc, and is properly warned about by psalm,
 */
function returns_strings() {
    /** @var array(string) $result also invalid, but crashes DependencyFinderVisitor*/
    $result = ['example'];
    return $result;
}
```

muglug:
Fixed in https://github.com/vimeo/psalm/commit/b2ae9a2892a62ebfe56b2478f20e23b5cf7e12ed

PR #298, opened 2017-11-14
Fix uncaught DocblockParseException on invalid inline doc comment

TysonAndre:
Emit an INFO level issue instead, the same way function doc comments do.

More locations may also need to be fixed, but these files caused uncaught
exceptions.

For issue #297


muglug:
Closed after improving issue catching more comprehensively. But thanks for raising issues!

Issue #309, opened 2017-11-15
Psalm may fail to warn if passed invalid key type for array access on an array


TysonAndre:
This is a feature that I'd want psalm to have, because it has array key types as a part of union types.

When `int`, `int|string`, `string`, or `mixed` are the array key types, I'd expect more warnings for code such as the below if the key can't cast to int|string:

```php
<?php
/**
 * @property array<int,int> $prop
 */
class A{
    /** @param string $x */
    public function __get($x) {
        return [2, $x];
    }
}
// This is incorrect code, designed to test Psalm's type checker on array keys
$c = new A();
echo strlen($c->prop);  // correctly warns $prop is array<int, int>
echo $c->prop['key'];

// Also, psalm isn't warning about array keys for regular arrays:
$x = ['value'];
var_export($x['key']);
var_export($x[[]]);  // still doesn't warn
var_export(strlen($x));  // array<int, string> provided
```

TysonAndre:
What are $inferred_key_type and $key_type supposed to be in src/Psalm/Checker/Statements/Expression/FetchChecker.php (analyzeArrayAccess)? I see that $key_type gets modified to include both types from the array and type from the key's expression.

It seems like $key_type is first created as the inferred type of $stmt->dim, then reused for the types expected by the array/array-like $stmt->var, which seems wrong.

After ExpressionChecker::analyze(), could a new union type be created for the expected types for $stmt->var?

muglug:
Lots and lots of bugs found having fixed this locally. Thanks!

TysonAndre:
I'm still having issues with magic properties, e.g. echo $c->prop['key']; should expect an int for an array key.

muglug:
> I'm still having issues with magic properties, e.g. echo $c->prop['key']; should expect an int for an array key.

Still having those issues?

muglug:
This was a pandora's box of a bug – once fixed, a WHOLE bunch of bugs related to the handling of object-like arrays bubbled up, and now Psalm is able to find a ton of bugs related to the heavy (over)use of arrays in our codebase. So thanks again for reporting.

TysonAndre:
> > I'm still having issues with magic properties, e.g. echo $c->prop['key']; should expect an int for an array key.

> Still having those issues?

No, those are fixed. Other issues were caused by `@property-read` being ignored and `null`/`string` (I forget which) being one of the possible element types. I fixed that on my fork, and should be able to fix the PR tomorrow

PR #310, opened 2017-11-15
Fix bug in array key type checks

TysonAndre:
Fixes https://github.com/vimeo/psalm/issues/309.
The test case(s) mentioned in #309 still needs to be added.

Fix a test failure, should be accessing $c, not $a

**After this change, Psalm self-checks fail.**
I haven't gotten around to investigating all of the failing self-checks.
Feel free to add fixes.

### DIFF
--- a/src/Psalm/Checker/FunctionChecker.php
+++ b/src/Psalm/Checker/FunctionChecker.php
@@ -787,7 +787,7 @@ class FunctionChecker extends FunctionLikeChecker
     /**
      * Gets the method/function call map
      *
-     * @return array<string, array<string, string>>
+     * @return array<string, array<int, string>>
      * @psalm-suppress MixedInferredReturnType as the use of require buggers things up
      * @psalm-suppress MixedAssignment
      * @psalm-suppress MoreSpecificReturnType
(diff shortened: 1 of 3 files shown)


TysonAndre:
PHPunit passed, psalm self-check didn't

```
The command "vendor/bin/phpunit" exited with 0.

10.33s$ bin/psalm

ERROR: InvalidArrayAccess - src/Psalm/Checker/ProjectChecker.php:957:40 - Cannot access value on variable $pool_data using string offset - expecting int

                IssueBuffer::addIssues($pool_data['issues']);

ERROR: InvalidArrayAccess - src/Psalm/Checker/Statements/Block/ForeachChecker.php:57:30 - Cannot access value on variable $foreach_context->vars_in_scope using null offset - expecting string

            $iterator_type = $foreach_context->vars_in_scope[$var_id];
```

muglug:
Thanks for this, ported over here: https://github.com/vimeo/psalm/commit/abe7c003abaa906eeb5fd6fc38cfec95229a9a75#diff-323c8fb14b8cc164b9a14a75316befa4

PR #317, opened 2017-11-17
Alias `@property-read` and `@property-write` to `@property`

TysonAndre:
Issues can be filed to warn if a magic property is read-only and
write-only later on.

It's more helpful to have them as aliases than to know nothing about
them, and this change shouldn't introduce false positives
(Unless read and write do different things).

### DIFF
--- a/src/Psalm/Checker/CommentChecker.php
+++ b/src/Psalm/Checker/CommentChecker.php
@@ -307,49 +307,64 @@ class CommentChecker
             }
         }
 
-        if (isset($comments['specials']['property'])) {
-            /** @var string $property */
-            foreach ($comments['specials']['property'] as $line_number => $property) {
-                try {
-                    $line_parts = self::splitDocLine($property);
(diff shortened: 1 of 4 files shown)


muglug:
Thanks!

So, read and write _do_ do different things – specifically here:
https://github.com/vimeo/psalm/blob/2a96e323e77b27ab3e8e14656fd3ac891bb6529b/src/Psalm/Visitor/DependencyFinderVisitor.php#L217-L224

Then they're used in `FetchChecker` and `AssignmentChecker` to allow different behaviour for `@property-read` and `@property-write`.

(I just added it to allow a plugin to write to the `ClassLikeStorage::$pseudo_property_(set|get)_types` arrays when initially scanning the class, before I knew about `@property-read` and `@property-write`).

So if you could add `property_read` and `property_write` properties to `ClassLikeDocblockComment`, we can assign them to the correct places.

muglug:
Thanks!

PR #326, opened 2017-11-20
Allow using nullable shorthand (`?`) within an array

TysonAndre:


muglug:
Thanks! Great addition!

muglug:
And also exposed a bug, fixed [here](https://github.com/vimeo/psalm/commit/9d025fad8fd703aece130bcf45dc68bdffafab39), where `@param int? $var` would turn into `@param intnull $var`

PR #359, opened 2017-12-01
grammar nit for TypeParseTreeException

TysonAndre:
Noticed when looking at recent changes


muglug:
Thanks! English is not my first... ok, it's my first and only language. I have no excuse.

PR #391, opened 2017-12-14
Fix a possibly undefined variable seen when issue(InvalidDocblock) is suppressed

TysonAndre:
This may also be a potential bug in the analysis of try/catch? Not sure of the intended behavior.


muglug:
Thanks!

PR #396, opened 2017-12-15
Don't convert standalone false/true to bool in combineTypes

### LINKED ISSUE #392: InvalidReturnType for global function with `return true;` and phpdoc `@return true`
Expected behavior: The below example should not fail. Psalm should infer a return type of `true`, and allow casting `true`/`false` to a declared return type of `bool` or `true`/`false`?
Observed: False positive InvalidReturnType

```php
<?php

/**
 * @return true (or throw)
 */
function test() {
    if (rand() % 2 > 0) {
        throw new RuntimeException("Unlucky");
    }
    return true;  // Unexpected  InvalidReturnType - The declared return type 'true' for test is incorrect, got 'bool'
}
```

TysonAndre:
Fixes #392

### DIFF
--- a/src/Psalm/Type.php
+++ b/src/Psalm/Type.php
@@ -527,10 +527,6 @@ abstract class Type
         }
 
         if (count($types) === 1) {
-            if ($types[0] instanceof TBool) {
-                $types[0] = new TBool;
-            }
-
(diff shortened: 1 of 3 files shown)


muglug:
Wonderful, thanks!

PR #420, opened 2018-01-03
fix: add preserve_keys on array_reverse function

StephaneBour:
An omission was made on the function array_reverse (the second parameter, optional).


muglug:
Wonderful, thanks! Looks like I need to add some special handling for that argument, but this still helps.

StephaneBour:
Oh, why ?

All errors with an `array_reverse(["c" => 1, "b" => 2], true);` are deleted

muglug:
oh forgive me, I was away from a computer and didn't check the documentation. I assumed that `array_reverse` always disregarded the keys unless `$preserve_keys = true`, but it's actually only if the keys are numeric, in which case the key type in the same.

PR #421, opened 2018-01-06
Make parameters for str_replace less specific

tvdijen:
It actually allows 'mixed' types for them:
http://php.net/manual/en/function.str-replace.php

### DIFF
--- a/src/Psalm/CallMap.php
+++ b/src/Psalm/CallMap.php
@@ -9184,7 +9184,7 @@ return [
 'strptime' => ['array|false', 'datestr'=>'string', 'format'=>'string'],
 'strrchr' => ['string', 'haystack'=>'string', 'needle'=>'string'],
 'str_repeat' => ['string', 'input'=>'string', 'mult'=>'int'],
-'str_replace' => ['string|array', 'search'=>'string|array', 'replace'=>'string|array', 'subject'=>'string|array', '&w_replace_count='=>'int'],
+'str_replace' => ['string|array', 'search'=>'', 'replace'=>'', 'subject'=>'', '&w_replace_count='=>'int'],
 'strrev' => ['string', 'str'=>'string'],
 'strripos' => ['int|false', 'haystack'=>'string', 'needle'=>'string', 'offset='=>'int'],
 'str_rot13' => ['string', 'str'=>'string'],


muglug:
> It actually allows 'mixed' types for them

The documentation is wrong, because the official PHP documentation doesn’t support union types (at least, I’ve never seen union types there.

For example, this is a fatal error, and something that Psalm should *definitely* protect against.

```
class A{}
echo str_replace(new A, new A, "foo");
```

tvdijen:
Okay, but it _does_ allow a 'true`, `false` or `null` for those parameters...

muglug:
> Can we agree on adding null|bool|int to the possible types? I'll update the PR for you then.

I don't feel great about that. Given that `str_replace` is designed to work on strings/arrays of strings, it feels weird to allow the passing of arguments that may not be strings. PHP is able to cast some values to a string:

```php
echo str_replace("five", 5, "Full fathom five thy father lies");
```

But Psalm's warning tells the programmer that they may be doing something that they didn't intend e.g.

```php
if (rand(0, 1)) {
  $a = strpos("abcdef", "f");
} else {
  $a = "Five";
}
echo str_replace("five", $a, "Full fathom five thy father lies");
```

tvdijen:
> Given that str_replace is designed to work on strings/arrays of strings, it feels weird to allow the passing of arguments that may not be strings.
I agree, but that's a PHP-issue that IMHO shouldn't be solved in Psalm...

muglug:
>  agree, but that's a PHP-issue that IMHO shouldn't be solved in Psalm...

Well, you *could* say that about property types, which don't have *any* type safety imposed upon them at runtime e.g.

```php
class A {
  /** @var ?string */
  public $foo;
}

function takesFoo(A $a) : void {
  $a->foo = 5;
}
```

Nevertheless, Psalm thinks this is problematic, because it probably violates the developer's intention.

muglug:
And, while this is a bit of an argument from authority, @rlerdorf [added this type definition](https://github.com/phan/phan/blob/1ca908372974a470c521f84315317f24a601ea81/includes/arginfo.php#L2532) to the static analysis tool he created, [Phan](https://github.com/phan/phan). 99.9% of `CallMap.php` originates from a derivative of this list.

Issue #422, opened 2018-01-07
Nothing inside isset is properly analysed


muglug:
Which leads to Psalm missing things like

```php
class A {}
$a = isset(A::foo()[0]);
```

Solution: add some sort of `Context::$inside_isset` flag so we don't worry about undefined vars, but allow other checks to carry on as normal.

PR #425, opened 2018-01-08
Fix method call missing object.

nesl247:


muglug:
Thanks! Caused by https://github.com/vimeo/psalm/issues/422

PR #437, opened 2018-01-11
Be strict about matching word boundaries for expected messages in tests

TysonAndre:
This helps when some issue types are prefixes of different issue types.

Consistently require matching word boundaries both at the beginning and the end of expected fragments in unit tests (Possibly<issue>, <issue>Suffix)


muglug:
Thanks!

PR #446, opened 2018-01-17
Return an empty string on magic __toString method

paragonie-scott:
We can't throw exceptions here, due to engine constraints.


muglug:
Thanks – do you have code that led to this?

paragonie-scott:
Not specifically; it was an unfamiliar codebase. Probably a malformed docblock.

Issue #487, opened 2018-02-05
SeekableIterator::seek() has incorrect signature, causing InvalidReturnType


TysonAndre:
Observed: `InvalidReturnType - src/iterator.php:30:5 - Not all code paths of tag_collections_tree_iterator::seek end in a return statement, return type mixed expected`
Cause is ` 'SeekableIterator::seek' => ['', 'position'=>'int'],` in src/Psalm/CallMap.php.
There are many other functions that have `''` instead of `'mixed'` or `'void'` in CallMap.php, so I expect other issues to crop up where ancestors are internal classes or interfaces.

```
<?php
// Cause is 'SeekableIterator::seek' => ['', 'position'=>'int'], in src/Psalm/CallMap.php.
// Maybe go through these and explicitly change '' to 'mixed' and 'void'?
class tag_collections_tree_iterator implements SeekableIterator {

    public function __construct() {
    }

    public function rewind() {
    }

    public function current() {
        return 2;
    }

    public function key() {
        return 1;
    }

    public function next() {
    }

    public function valid() {
        return true;
    }

    /**
     * Seek recursively to a given position in the tree.
     * @param int $position
     * @throws OutOfBoundsException if $position is negative or >= count()
     *
     * It has an inherited type of void from SeekableIterator? (Actually '')
     */
    public function seek($position) {
        $count = 0;
        if ($count <= $position) {
            // $position is greater than or equal to sum($meta['count']) - ArrayIterator at the bottom would throw.
            throw new OutOfBoundsException("Seek (sub)position $position is out of range");
        }
    }
}
```

TysonAndre:
The upstream for CallMap.php was updated. It extracted `void`/`mixed` from docs.php.net's XML sources, for functions where param/return types were previously `''`. See the following changes: [first](https://github.com/phan/phan/pull/1430), [second](https://github.com/phan/phan/pull/1431)

(The import of thousands of types was done via a script, and some of the types from docs.php.net may be incomplete or inaccurate)

PR #488, opened 2018-02-05
Fix a few signatures of Iterator classes for void

TysonAndre:
This was done automatically via sed.
Other types can be added in separate PRs

Fixes https://github.com/vimeo/psalm/issues/487 (But not the general case)


muglug:
Thanks!

PR #494, opened 2018-02-06
Improve callmap with information from docs.php.net

TysonAndre:
This modifies entries that were previously `''` for their types.
This uses [a script](https://github.com/phan/phan/blob/master/internal/internalsignatures.php) to update the signatures of a function signature array automatically and generate new signatures.

docs.php.net will have some incomplete types (and vague types such as `mixed` instead of `int|string`, but should be an improvement over the large number of missing types.


muglug:
Thanks!

PR #495, opened 2018-02-07
Make call map types slightly more specific (manually)

TysonAndre:
Feel free to make any approriate patches. (e.g. not sure if you want to include false)

Hopefully, the changes reflect the documentation and example code found at php.net.


src/Psalm/CallMap.php
```diff
@@ -4539,7 +4539,7 @@
 'kadm5_get_principals' => ['array', 'handle'=>'resource'],
 'kadm5_init_with_password' => ['resource', 'admin_server'=>'string', 'realm'=>'string', 'principal'=>'string', 'password'=>'string'],
 'kadm5_modify_principal' => ['bool', 'handle'=>'resource', 'principal'=>'string', 'options'=>'array'],
-'key' => ['mixed', 'array_arg'=>'array'],
+'key' => ['int|string|false', 'array_arg'=>'array'],
```
muglug: This is something I'd rather keep as `mixed` (or maybe more specifically `scalar`) because otherwise it may get in the way more than anything (especially as people rarely type their array keys), so it can't be inferred much.

src/Psalm/CallMap.php
```diff
@@ -2506,7 +2506,7 @@
 'getmypid' => ['int'],
 'getmyuid' => ['int'],
 'get_object_vars' => ['array', 'obj'=>'object'],
-'getopt' => ['string[]|false[]|array[]', 'options'=>'string', 'longopts='=>'array', '&w_optind='=>'int'],
+'getopt' => ['array<string,string>|array<string,false>|array<string,array<int,string|false>>', 'options'=>'string', 'longopts='=>'array', '&w_optind='=>'int'],
```
muglug: Do you want the more compact `<array,string|false|array<int,string|false>>` here?

src/Psalm/CallMap.php
```diff
@@ -242,17 +242,17 @@
...
 'array_replace' => ['array', 'arr1'=>'array', 'arr2'=>'array', '...='=>'array'],
 'array_replace_recursive' => ['array', 'arr1'=>'array', 'arr2'=>'array', '...='=>'array'],
 'array_reverse' => ['array', 'input'=>'array', 'preserve='=>'bool'],
-'array_search' => ['mixed', 'needle'=>'mixed', 'haystack'=>'array', 'strict='=>'bool'],
+'array_search' => ['int|string|false', 'needle'=>'mixed', 'haystack'=>'array', 'strict='=>'bool'],
 'array_shift' => ['mixed', '&rw_stack'=>'array'],
 'array_slice' => ['array', 'input'=>'array', 'offset'=>'int', 'length='=>'?int', 'preserve_keys='=>'bool'],
 'array_splice' => ['array', '&rw_input'=>'array', 'offset'=>'int', 'length='=>'int', 'replacement='=>'array|string'],
-'array_sum' => ['numeric', 'input'=>'array'],
+'array_sum' => ['int|float', 'input'=>'array'],
```
muglug: I'm on the fence about this too, but I could also override very easily in `FunctionChecker::getArrayReturnType` for when we know the array type args.

src/Psalm/CallMap.php
TysonAndre: I could, I didn't know if psalm supported nested union types in array types (In CallMap.php) yet. (That would be the first occurence)

muglug: Everything just goes into `Type::parseString` – if it can handle it in a docblock, it can handle it here. That also goes for `get_defined_functions`, which could use `array{user:...,internal:...}` instead of just a generic array.

TysonAndre: Fixed. Also, missed that I was using nested array types already.  A simple test of get_defined_functions() works properly.

muglug:
Thanks!

PR #496, opened 2018-02-07
Make types in the call map more specific (manual, part 2)

TysonAndre:
fix a few incorrect types.

Up to date phpredis documentation is linked at https://github.com/phpredis/phpredis (verified getKeys manually). (Many redis signatures don't exist in the call map right now)
The rest should match php.net's documentation


src/Psalm/CallMap.php
```diff
@@ -7719,7 +7719,7 @@
 'Reflection::getModifierNames' => ['array', 'modifiers'=>'int'],
 'ReflectionMethod::__construct' => ['void', 'class'=>'string|object', 'name'=>'string'],
 'ReflectionMethod::__construct\'1' => ['ReflectionMethod', 'class_method'=>'string'],
-'ReflectionMethod::export' => ['?string', 'class'=>'string', 'name'=>'string', 'return='=>'bool'],
+'ReflectionMethod::export' => ['string', 'class'=>'string', 'name'=>'string', 'return='=>'bool'],
```
muglug: The docs are wrong here – this returns `null` unless a third arg of `true` is given

src/Psalm/CallMap.php
```diff
@@ -7737,7 +7737,7 @@
 'ReflectionMethod::setAccessible' => ['void', 'visible'=>'bool'],
 'ReflectionMethod::__toString' => ['string'],
 'ReflectionObject::__construct' => ['void', 'argument'=>'object'],
-'ReflectionObject::export' => ['?string', 'argument'=>'object', 'return='=>'bool'],
+'ReflectionObject::export' => ['?string', 'argument'=>'string|object', 'return='=>'bool'],
```
muglug: Docs are incomplete here:
```php
class A {}
ReflectionObject::export("A");
```
gives
```
Warning: ReflectionObject::__construct() expects parameter 1 to be object
```

src/Psalm/CallMap.php
TysonAndre: Addressed both comments, incorrectly copied the incorrect line here

PR #500, opened 2018-02-07
Throw error if magic getter or setter called for undefined property or invalid type specified with annotations

### LINKED ISSUE #480: @property should apply to internal __get and __set calls
This should generate an issue cc @nickyr 

```php
/**
 * @property string $foo
 */
class A {
  public function __get(string $name): ?string {
    if ($name === "foo") {
      return "hello";
    }
  }

  /** @param mixed $value */
  public function __set(string $name, $value): void {
  }
  
  public function badSet(): void {
    $this->__set('foo', new stdClass());
  }
}
```

nickyr:
~~If a magic getter or setter is used to access a property on a class that is not defined but a `@property` annotation for the property exists, throw an error. If no `@property` annotation exists, it's not an error because you're allowed to make magic getters and setters do crazy things.~~

Check properties accessed with magic getters and setters.
- If `@psalm-seal-properties` is set, they must be defined.
- If an `@property` annotation is specified, the setter must set something with the correct type.

Fixes #480

This seems to work pretty well but I'm pretty sure I did not put the code in the ideal place. That analyze method is really long and there's a ton of nesting so I was rather confused. Lemme know where the ideal place is!


src/Psalm/Checker/Statements/Expression/Call/MethodCallChecker.php
```diff
@@ -542,5 +544,57 @@ public static function analyze(
...
         }
+
+        // If a magic getter or setter is used to access a property on a class that is not defined
+        // but a `@property` annotation for the property exists, throw an error. If no `@property`
+        // annotation exists, it's not an error because you're allowed to make magic getters and
+        // setters do crazy things.
+        if ($stmt->var->inferredType && is_string($stmt->name)) {
+            $method_name = strtolower($stmt->name);
+            foreach ($stmt->var->inferredType->getTypes() as $type) {
+                if ($type instanceof TNamedObject) {
```
muglug: You can remove three levels of nesting here with
```php
if (!$type instance TNamedObject) {
  continue;
}

$fq_class_name = $type->value;
$class_storage = $project_checker->classlike_storage_provider->get($fq_class_name);

if (!in_array($method_name, ['__get', '__set'], true)) {
  continue;
}

$first_arg_value = $stmt->args[0]->value;

if (!$first_arg_value instanceof PhpParser\Node\Scalar\String_) {
  continue;
}
```

src/Psalm/Checker/Statements/Expression/Call/MethodCallChecker.php
```diff
@@ -542,5 +544,57 @@ public static function analyze(
 
             $context->vars_in_scope[$var_id] = $class_type;
         }
+
+        // If a magic getter or setter is used to access a property on a class that is not defined
+        // but a `@property` annotation for the property exists, throw an error. If no `@property`
+        // annotation exists, it's not an error because you're allowed to make magic getters and
+        // setters do crazy things.
+        if ($stmt->var->inferredType && is_string($stmt->name)) {
```
muglug: This can start after `checkMethodNotDeprecated`. You probably want to create a private method that has this logic too.

nickyr:
Good ideas, @muglug! Made all the changes.

src/Psalm/Checker/Statements/Expression/Call/MethodCallChecker.php
```diff
@@ -542,4 +553,75 @@ public static function analyze(
...
+     * @param PhpParser\Node\Expr\MethodCall $stmt
+     * @param string $fq_class_name
+     *
+     * @return bool
+     */
+    private static function checkMagicGetterOrSetterUndefinedProperty(
+        StatementsChecker $statements_checker,
+        \Psalm\Checker\ProjectChecker $project_checker,
+        PhpParser\Node\Expr\MethodCall $stmt,
+        string $fq_class_name
```
muglug: Can't be having a `string` typehint

src/Psalm/Checker/Statements/Expression/Call/MethodCallChecker.php
```diff
@@ -542,4 +553,75 @@ public static function analyze(
...
+        }
+
+        $prop_name = $first_arg_value->value;
+        $property_id = $fq_class_name . '::$' . $prop_name;
+        $class_storage = $project_checker->classlike_storage_provider->get($fq_class_name);
+
+        switch ($method_name) {
+            case '__set':
+                if (!ClassLikeChecker::propertyExists($project_checker, $property_id)
+                    && isset($class_storage->pseudo_property_set_types['$' . $prop_name])
```
muglug: This is not sound logic – it makes the invalid test throw an exception, but it's the wrong error message. You want to check here for `!isset`

tests/AnnotationTest.php
```diff
@@ -912,6 +958,62 @@ function takesArrayIteratorOfString($i): void {
...
+
+                        /** @param mixed $value */
+                        public function __set(string $name, $value): void {
+                        }
+
+                        public function badSet(): void {
+                            $this->__set("foo", new stdClass());
+                        }
+                    }',
+                'error_message' => 'UndefinedThisPropertyAssignment',
```
muglug: The magic property is defined - the issue should be `InvalidPropertyAssignmentValue`

src/Psalm/Checker/Statements/Expression/Call/MethodCallChecker.php
```diff
@@ -542,4 +553,75 @@ public static function analyze(
...
+            return true;
+        }
+
+        $prop_name = $first_arg_value->value;
+        $property_id = $fq_class_name . '::$' . $prop_name;
+        $class_storage = $project_checker->classlike_storage_provider->get($fq_class_name);
+
+        switch ($method_name) {
+            case '__set':
+                if (!ClassLikeChecker::propertyExists($project_checker, $property_id)
```
muglug: You can move this check above the switch since it's shared. Also I subtly changed the API so it's now `$project_checker->codebase->properties->propertyExists($property_id)`

src/Psalm/Checker/Statements/Expression/Call/MethodCallChecker.php
nickyr: drat forgot about that

tests/AnnotationTest.php
nickyr: i thought you said it should throw the same error message as if you did `$this->foo = new stdClass()` without defining foo? that still seems to make more sense to me

src/Psalm/Checker/Statements/Expression/Call/MethodCallChecker.php
nickyr: tricky tricky. will do.

src/Psalm/Checker/Statements/Expression/Call/MethodCallChecker.php
nickyr: i am v confused about this and https://github.com/vimeo/psalm/pull/500/files#r167096939, can we talk offline when you have a few minutes?

tests/AnnotationTest.php
muglug: But you're defining `@property string $foo` above!

nickyr:
Alright @muglug updated as discussed. And updated the PR description. Lemme know how it's looking!

muglug:
Awesome, thanks @nickyr!

PR #523, opened 2018-02-18
Fix unused imports in psalm, wrong param order doc comments

TysonAndre:
(Some of the imports appear as strings elsewhere in the same file, but not as phpdoc tags)


muglug:
Thanks!

PR #528, opened 2018-02-19
Add a script to run the same tests as jenkins locally

TysonAndre:
Make `composer tests` use only phpcs

Also add `composer all-tests`, which also runs the self-check.

php-cs-fixer was removed from .travis.yml earlier.
I'm guessing that composer was overlooked?

The goal is to make it easy to create PRs passing travis checks.


muglug:
Thanks, makes sense

PR #350, opened 2017-11-28
[wip] replace composer/composer with composer/xdebug-handler

weirdan:
Fixes vimeo/psalm#349

Note: there's no stable release at the moment, so it's using dev-master for now. For this reason you may want to not merge it right away and wait for stable release.

PR #549, opened 2018-03-02
[wip] Build phar

weirdan:
This PR adds necessary changes to be able to build Psalm as .phar.
Resulting binary is ~1.6MB~ ~700KB (as opposed to ~28MB~ 23MB "production" (no-dev) composer install).
Build script is provided in `bin/build-phar.sh`.

I considered adding `humbug/box` as a dev dependency, but if I did this I would need to filter out it (and it's dependencies, and psalm's other dev dependencies) when building the phar, so instead I opted for global composer install and producing phar from "production" build.

Some of the `realpath()` calls were dropped because `realpath()` returns `false` on any (even existing) path inside the phar archive.

~Notably the phar version reports issues (possibly false positives, I haven't checked) that regular version does not.~ (not anymore).


muglug:
This is very great.

I'd like to wait until https://github.com/vimeo/psalm/pull/350 is done (and it seems *so* close), because that'll cut almost 5MB of PHP from the phar.

weirdan:
Don't merge yet, it's broken (see https://github.com/humbug/box/issues/84)

weirdan:
Resolved now.

muglug:
Could you add `file_exists` checks where before `realpath` would return `false`? Then I think this is good to merge.

weirdan:
@muglug done in `src/Psalm/Config.php`. `src/command_functions.php` already had `file_exists()` checks, so I just dropped additional variables that has become excessive with no `realpath()` calls.

muglug:
@weirdan thanks – could you rebase master to clear up merge conflicts? I fear I may have slightly trampled on this unnecessarily.

weirdan:
@muglug sure, done. Interestingly enough, github said there were no conflicts with the base branch.

muglug:
Thanks a lot!

mindplay-dk:
Is there any way to install the `phar` to the `bin` folder in a project?

Global installation doesn't work for me - I want an exact, repeatable dependency installation, so that individual developers (and the CI server) get the same version, and can run it via a Composer script.

And I want the binary phar installation only - I don't want the dependencies of psalm to potentially conflict with the dependencies of the project itself.

I'd also like to be able to package the exact version dependency and all settings as a package - so that we can create and maintain our developer standards and reuse them on many projects.

Any of you have a setup like that on any project for reference?

Thanks :-)

weirdan:
Overall, I'm not aware of a solution that would fit all of your requirements.

> Is there any way to install the phar to the bin folder in a project?

`wget https://github.com/vimeo/psalm/releases/download/3.2.2/psalm.phar` + `git add` + `git commit`. You would have to update it manually from time to time though.

> I don't want the dependencies of psalm to potentially conflict with the dependencies of the project itself.

Psalm itself uses [`bamarni/composer-bin-plugin`](https://github.com/bamarni/composer-bin-plugin), which allows you to have isolated vendor environments. It uses (environment-local) composer install though, not phar.

muglug:
PHPStan has a script that exports the phar to a separate repo: https://github.com/phpstan/phpstan/blob/96c57c66ea24a75a17ec5dff8c684439ac52ead5/.travis.yml#L44-L56

I'll do something similar

muglug:
@mindplay-dk behold:

```
composer require --dev psalm/shim
```

muglug:
This is the change that _should_ add this on every build, and create a comparable tag for every release:

https://github.com/vimeo/psalm/commit/07768cf505b8dcf2d98b2f0c7f44cecfb170e5f5

Maybe `psalm/shim` should be `psalm/phar`? Not sure about that.

mindplay-dk:
Neat! Thanks :-)

Yeah, "shim" seems to imply a filler or placeholder, something that isn't the real thing?

`psalm/phar` or maybe `psalm/bin` would seem fitting :-)

muglug:
I've changed the name everywhere to [`psalm/phar`](https://github.com/psalm/phar)

mindplay-dk:
@muglug is there any way to launch the language-server from the phar?

muglug:
Yes, but only in the latest `master` Phar (from https://github.com/psalm/phar)

Use `--language-server` plus any additional args

mindplay-dk:
Damn, you're on point, man 😃

I got a warning the first time I launched it though (but not the second) - might be a bug?

```
PHP Warning:  file_put_contents(/tmp/psalm/595983f3e8518df3464c6162d2502437041ebc1d-s/config): failed to open stream: No such file or directory in phar:///mnt/c/workspace/kodus-score/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Provider/FileReferenceCacheProvider.php on line 254
```

mindplay-dk:
No luck though, because [this](https://github.com/gtache/intellij-lsp/issues/88).

muglug:
Have you tried running Psalm directly on Windows?

mindplay-dk:
> Have you tried running Psalm directly on Windows?

No. (I don't want to maintain a Windows PHP environment - not since WSL, which lets me run the supported Linux builds of every extension etc... Windows support for PHP extensions is notoriously bad.)

muglug:
Oh interesting!

PR #612, opened 2018-03-22
fix ldap_bind parameters, allow null for bind_rdn and bind_password

fkooman:
https://secure.php.net/ldap_bind

>  If bind_password is not specified or is empty, an anonymous bind is attempted. The bind_dn can also be left empty for an anonymous bind. This is defined in https://tools.ietf.org/html/rfc2251#section-4.2.2


muglug:
Thanks!

fkooman:
It seems this change was undone with an import from Phan rules?

https://github.com/vimeo/psalm/commit/b6499c33ed9c88baf1048f4328931decce98488e

muglug:
Sorry, fixed in b1e81c2

PR #620, opened 2018-03-25
Sort the call map

TysonAndre:
Psalm's function signature map was based on an old ad hoc sorting order
(Possibly the unix sorting utility with manual updates?)
Instead, sort by `strtolower(str_replace("'", "\x0", $functionKey))`
(To group classes together, and global functions together, and alternates together,
instead of mixing them)
`phan/phan/internal/internalsignatures.php sort` was used to perform this
sorting (It's hardcoded to expect src/Phan/Internal/FunctionSignatureMap.php, though)

Goals:
- Keep this in sync with Phan's call map
- Have an understandable sort order, be able to quickly see where a given method/function is when reading a file
- Have a sort order that can be enforced by checks in CI

Other changes(Should be the only changes):

- Add signature for `_()`, which is an alias of gettext()
  https://secure.php.net/manual/en/function.gettext.php#refsect1-function.gettext-notes
- etsy/phan -> phan/phan
- Mention that __construct has type void, not the class name.
- nit: Change __halt_compiler to void.


muglug:
Thanks!

weirdan:
If this is compatible (and the goal is to keep it as such) with Phan, wouldn't it make sense to split it to it's own package and then both projects could depend on that? This way, maintenance effort wouldn't have to be duplicated.

muglug:
@weirdan (there are a few incompatibilities (e.g. `callable` syntax) I need to add to Psalm

TysonAndre:
Related to https://github.com/phpstan/phpstan/ issue 846

There are already minor differences, and trying to resolve those differences would slow down development. Periodically manually syncing any new signatures or improvements to signatures is easier:

- `callable(T1):T2` introduced in phan's FunctionSignatureMap for a handful of methods, such as set_error_handler
- Differences in the way error return values are handled: If a function will return false in an extremely rare error case (e.g. false), should false be included in the return value (Phan leans towards yes, psalm to no (to avoid "attempting to cast string|false to string"))
- Deciding on who would maintain/review the package, how that package would be versioned, publishing and verifying frequent releases of that package
  (e.g. a typo in a signature type might be ignored by one project but cause an uncaught error in another project)
- Deciding which major version of extensions (e.g. Redis) to represent (The most commonly used vs latest) may have disagreement eventually

muglug:
@TysonAndre I agree completely

> Differences in the way error return values are handled: If a function will return false in an extremely rare error case (e.g. false), should false be included in the return value (Phan leans towards yes, psalm to no

FYI this should no longer an issue - every function return type (except `*strpos` ) has `@psalm-ignore-falsable-return` so that Psalm won't care. Any downstream errors that complain about `false` in the return values of those functions should be reported as bugs.

PR #621, opened 2018-03-25
dropped `aggregate_*` functions (PHP4 only)

weirdan:
Aggregate functions are PHP4 only feature. I don't think Psalm supports checking PHP4 code, so these could be dropped safely.

This poses the question, however, what versions of PHP psalm *does* support. For example, there are `session_register()` and friends, which have been removed in PHP5.4 - could those be dropped?


muglug:
Thanks, I hate the github conflicts resolver which never works. I added changes to latest master.

Issue #595, opened 2018-03-17
Phar file cannot be renamed


jakzal:
Psalm stops working after renaming the downloaded `psalm.phar` to `psalm`. Steps to reproduce below.

environment: php 7.2, os x

```
curl -Ls https://github.com/vimeo/psalm/releases/download/1.0.3/psalm.phar > psalm
chmod +x psalm
./psalm -h
```

```
PHP Warning:  require(phar:///Users/kuba/Projects/foo/psalm/psalm): failed to open stream: phar error: no directory in "phar:///Users/kuba/Projects/foo/psalm/psalm", must have at least phar:///Users/kuba/Projects/foo/psalm/psalm/ for root directory (always use full path to a new phar) in /Users/kuba/Projects/foo/psalm on line 10

Warning: require(phar:///Users/kuba/Projects/foo/psalm/psalm): failed to open stream: phar error: no directory in "phar:///Users/kuba/Projects/foo/psalm/psalm", must have at least phar:///Users/kuba/Projects/foo/psalm/psalm/ for root directory (always use full path to a new phar) in /Users/kuba/Projects/foo/psalm on line 10
PHP Fatal error:  require(): Failed opening required 'phar:///Users/kuba/Projects/foo/psalm/psalm' (include_path='.:') in /Users/kuba/Projects/foo/psalm on line 10

Fatal error: require(): Failed opening required 'phar:///Users/kuba/Projects/foo/psalm/psalm' (include_path='.:') in /Users/kuba/Projects/foo/psalm on line 10
```

weirdan:
Only affects files without extension. For example, `file.phar` or `file.q` work fine.

weirdan:
Might be related: https://bugs.php.net/bug.php?id=76061

weirdan:
Line 10 points to: 
```php
require 'phar://' . __FILE__ . '/psalm';
```
Composer's stub (which does not seem to suffer from this problem) looks a bit different:
```php
Phar::mapPhar('composer.phar');
require 'phar://composer.phar/bin/composer';
```

muglug:
That code comes from https://github.com/humbug/box/blob/master/src/StubGenerator.php I think

muglug:
I've fixed the generation step and updated the Phar attached to the release. The repro code you pasted above now works.

Thanks for ticketing!

jakzal:
Thank you for your quick reaction! 🍺

jakzal:
[...]

PR #656, opened 2018-04-08
Simplify the PHAR build script

theofidry:
I originally wanted to try if I was hitting a too many files open limit when compressing the PHAR. I cant' seem to reproduce it on my machine so I was looking for a real project where this was happening.

I did not hit the issue and simplified the build script:

- Leverage PhpScoper from inside Box instead: it's faster
- Simplify the Box config
- Instead of relying on global dependencies (for example you won't be able to install php-scoper & box on mine), I prefer to leverage [bamarni composer bin plugin](https://github.com/bamarni/composer-bin-plugin). You can find a bit more of it's usage in its doc or [the article I wrote](https://medium.com/@tfidry/managing-your-dependencies-in-php-321d584441ab).

The last step, running the `psalm.phar` PHAR does not work: I think I'm missing something obvious but can't figure out what.

Also not if you wish to see the files added to the PHAR, e.g. to debug the PHP-Scoper effect, you can:

- use an IDE like PhpStorm which is able to index PHAR files and inspect them
- use the `--debug` mode of `compile` which is going to dump the files added to the PHAR a directory `.box`


muglug:
I can't get the Phar to build at all on my work machine (4-year-old Macbook Pro). I get OOM errors.

muglug:
I'll try to get it building on my home machine too, maybe a bad config somewhere.

theofidry:
OOM errors? If possible it would be nice to get a details on them. I’m
trying to make so that box handle the config for you, for example setting
phar.readonly=0. So let’s not rush that PR I would like as much as possible
fix the real underlying issues :) as you already have a working PHAR and
build process, we can take our time for this.

bin/build-phar.sh
```diff
@@ -4,34 +4,8 @@ if ! php -r 'extension_loaded("bz2") or exit(1);' ; then
 	exit 1
 fi
 
-composer global require 'humbug/php-scoper:^1.0@dev'
-composer global require humbug/box:dev-master
-composer install --no-dev
-[ -d build ] || mkdir build
-[ -d build/psalm ] || mkdir build/psalm
-# increase FD limit, or Phar compression will fail
-ulimit -Sn 4096
```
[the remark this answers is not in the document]
weirdan: I've seen that, when building psalm.phar including dev dependencies.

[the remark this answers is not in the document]
weirdan: And  I can reproduce it:
```console
$ cd psalm
$ git checkout c13fda4034b33eb859e8810b10b0322fca17b858 
$ patch -p1 <<EOM
diff --git a/bin/build-phar.sh b/bin/build-phar.sh
index 1e05c796..7843dd05 100755
--- a/bin/build-phar.sh
+++ b/bin/build-phar.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 composer global require humbug/box:dev-master
-composer install --no-dev
+composer install # --no-dev
 [ -d build ] || mkdir build
-box build
+php -dphar.readonly=0 `which box` build
EOM
patching file bin/build-phar.sh
Hunk #1 succeeded at 1 with fuzz 2.
$ bin/build-phar.sh
[......snip......]
Box (repo)


 // Loading the configuration file "/home/weirdan/src/psalm/box.json".                                                  

Building the PHAR "/home/weirdan/src/psalm/build/psalm.phar"
? Registering compactors
  + KevinGH\Box\Compactor\Php
? Adding main file: /home/weirdan/src/psalm/psalm
? Adding binary files
    > No file found
? Adding files
    > 4496 file(s)
? Generating new stub
  - Using shebang line: #!/usr/bin/env php
  - Using banner:
    > Generated by Humbug Box.
    > 
    > @link https://github.com/humbug/box
? Compressing with the algorithm "BZ2"

In Compile.php line 415:
                                   
  unable to create temporary file  
                                   

build [-c|--config CONFIG] [--debug] [-d|--working-dir WORKING-DIR]
[......snip......]
```

The box version is 3.0.0-alpha.0 ( f5eec71df2df3a2a1f39a50ba77e8bb4312fef51 )

bin/build-phar.sh
```diff
@@ -1,37 +1,6 @@
 #!/usr/bin/env bash
-if ! php -r 'extension_loaded("bz2") or exit(1);' ; then
```
theofidry: this is now handled in Box

bin/build-phar.sh
```diff
@@ -1,37 +1,6 @@
...
-	echo "You need to install (or enable) bz2 php extension"
-	exit 1
-fi
+composer bin box install
 
-composer global require 'humbug/php-scoper:^1.0@dev'
```
theofidry: this has been replaced by `bamarni/composer-bin-plugin`, up to you if you're cool with that or you prefer the `composer global` or maybe switch to a `wget` the Box PHAR

bin/build-phar.sh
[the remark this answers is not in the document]
theofidry: Fixed in Box now

bin/build-phar.sh
```diff
@@ -1,37 +1,6 @@
...
-[ -d build ] || mkdir build
-[ -d build/psalm ] || mkdir build/psalm
-# increase FD limit, or Phar compression will fail
-ulimit -Sn 4096
+vendor/bin/box compile
 
-rm -f bin/psalm.phar
-
-# Prefixes the code to be bundled
-php -d memory_limit=-1 `which php-scoper` add-prefix --prefix='PsalmPhar' --output-dir=build/psalm --force
```
theofidry: the scoping is done via Box now, the only issue is https://github.com/humbug/php-scoper/issues/178 so right now the prefix is really random

theofidry: the memory is also bumped by Box although the number is quite arbitrary so it may need more

bin/build-phar.sh
```diff
@@ -1,37 +1,6 @@
...
-
-# Re-dump the loader to account for the prefixing
-# and optimize the loader
-composer dump-autoload --working-dir=build/psalm --classmap-authoritative --no-dev
```
theofidry: including only the non-dev files (and avoiding to scope the dev files) is done by Box

bin/build-phar.sh
```diff
@@ -1,37 +1,6 @@
...
-
-chmod 755 build/psalm/psalm
-
-cp bin/phar.psalm.xml build/psalm/psalm.xml
-
-./build/psalm/psalm --config=build/psalm/psalm.xml --root=build/psalm
-
-php -d memory_limit=-1 -d phar.readonly=0 `which box` compile
```
theofidry: the `readonly` flag is taken care of by Box

box.json.dist
```diff
@@ -1,23 +1,16 @@
 {
-    "alias": "psalm.phar",
```
theofidry: there shouldn't be a need for an alias unless you really want to refer to `phar://psalm.phar` in the codebase which I didn't see

box.json.dist
```diff
@@ -1,23 +1,16 @@
...
-        {
-            "in" : "src"
-        }
-    ],
-    "files" : [
-        "config.xsd"
+    "output" : "build/psalm.phar",
+    "files": [
+        "src/command_functions.php",
+        "src/psalm.php",
```
theofidry: is there any reason why those files are not included in the autoload?

box.json.dist
```diff
@@ -1,23 +1,16 @@
...
     ],
+    "files-bin": ["config.xsd"],
+    "directories-bin" : ["assets"],
     "intercept" : false,
-    "compactors" : [],
+    "compactors" : [
+        "KevinGH\\Box\\Compactor\\PhpScoper"
+    ],
     "chmod" : "0755",
-    "shebang" : "#!/usr/bin/env php",
```
theofidry: this is already the default shebang

box.json.dist
```diff
@@ -1,23 +1,16 @@
...
-    "stub" : true
```
theofidry: this is the default value already

box.json.dist
```diff
@@ -1,23 +1,16 @@
...
+    "compression": "GZ"
```
theofidry: there was no compression, not sure if that was voluntary because you wanted to avoid your users to have the `zlib` extension

scoper.inc.php
```diff
@@ -1,34 +1,6 @@
 <?php
 
-use Isolated\Symfony\Component\Finder\Finder;
-
 return [
-    'finders' => [
```
theofidry: this part (finding the files) is handled by Box

scoper.inc.php
```diff
@@ -136,4 +115,7 @@ function ($filePath, $prefix, $contents) {
             return $contents;
         },
     ],
+    'whitelist' => [
```
theofidry: this whitelist is useless so could be removed

src/psalm.php
```diff
@@ -127,7 +127,7 @@
     $options['r'] = $options['root'];
 }
 
-$current_dir = (string)getcwd() . DIRECTORY_SEPARATOR;
+$current_dir = (string)getcwd();
```
theofidry: I'm not quite sure why we have the trailing `/`. This was making fail the find autoloader since we ended up with `//` in the path

box.json.dist
muglug: Yeah, this was just added for renaming in https://github.com/vimeo/psalm/issues/595, but you fixed later.

box.json.dist
muglug: They're part of Psalm's command-line startup, and themselves handle autoloading

box.json.dist
muglug: Yes, definitely on purpose (because `bzip` was causing issues for users without it). If 99% of PHP installs have `zlib`, happy to use that.

bin/build-phar.sh
muglug: The deterministic prefix was on purpose to enable better cache re-use, but I don't really mind caches being busted between new phars.

theofidry:
@muglug in theory this should be ready. I think however some code changed and maybe the PHP-Scoper config needs to be updated.

theofidry:
Updated and the builds are green.

I may have missed them, but if they is no e2e tests for the PHAR I would recommend to add one (in another PR)

muglug:
Thanks a lot!

PR #668, opened 2018-04-12
Few fixes to make tests pass on windows

srsbiz:
* changed / to DIRECTORY_SEPARATOR in expected messages
* changed most occurrences of PHP_EOL to "\n" in tests/ and src/
* added output buffering in test to avoid marking test as risky


muglug:
Thanks!

Issue #669, opened 2018-04-13
mysqli methods have wrong parameter counts


jaydiablo:
Ran into a case where psalm was complaining about some methods in a class that extends PHP's `mysqli` class, something like this perhaps:

```php
class db extends mysqli 
{
    public function close() // claims too few parameters http://php.net/manual/en/mysqli.close.php
    {
    }

    public function prepare($sql) // claims too few parameters http://php.net/manual/en/mysqli.prepare.php
    {
    }

    public function commit($flags = null, $name = null) // claims first is wrong type, too many params http://php.net/manual/en/mysqli.commit.php
    {
    }

    public function real_escape_string($string) // claims too few parameters http://php.net/manual/en/mysqli.real-escape-string.php
    {
    }
}
```

There may be others, I didn't audit all of them. It seems the issue stems from psalm's definitions of these methods being copies of the procedural style functions, which have different parameter requirements (usually take the `mysqli` resource as the first parameter).

jaydiablo:
Another case where `mysqli::prepare` isn't marked that it may return `false` (or `mysqli_stmt`).

http://php.net/manual/en/mysqli.prepare.php

>mysqli_prepare() returns a statement object or FALSE if an error occurred.

muglug:
Thanks! Should be fixed now.

weirdan:
@muglug there's several more link arguments to be dropped. Also, there's quite a few **properties** of `mysqli` class in the CallMap, and I don't think they should be there. Or should they?

PR #671, opened 2018-04-13
Mysqli signatures fixes

weirdan:
There was a number of entries in the method list that were actually properties (likely imported from php mysqli docs). These are now gone.

Also it drops several $link parameters you missed in 564cebbaad679128c67cbd696a7727019d288d90.

Refs #669


muglug:
Thanks!

PR #688, opened 2018-04-21
Check that __clone, __construct, __destruct signatures omit return types

rightfold:
PHP complains when you have a return type declared for these three methods.

### DIFF
--- a/src/Psalm/Checker/MethodChecker.php
+++ b/src/Psalm/Checker/MethodChecker.php
@@ -9,6 +9,7 @@ use Psalm\Issue\ImplementedReturnTypeMismatch;
 use Psalm\Issue\InaccessibleMethod;
 use Psalm\Issue\InvalidStaticInvocation;
 use Psalm\Issue\MethodSignatureMismatch;
+use Psalm\Issue\MethodSignatureMustOmitReturnType;
 use Psalm\Issue\MoreSpecificImplementedParamType;
 use Psalm\Issue\LessSpecificImplementedReturnType;
 use Psalm\Issue\NonStaticSelfCall;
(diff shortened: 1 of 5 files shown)


tests/MethodSignatureTest.php
```diff
@@ -472,4 +472,26 @@ class B extends A {
...
+     * @expectedException        \Psalm\Exception\CodeException
+     * @expectedExceptionMessage MethodSignatureMustOmitReturnType
+     *
+     * @return                   void
+     */
+    public function testMustOmitReturnType()
+    {
+        $this->addFile(
+            'somefile.php',
+            '<?php
```
muglug: this code can be added to the above data provider as

```php
'mustOmitReturnType' => [
    '<?php
        class A
        {
            public function __construct(): void
            {
            }
        }',
    'error_message' => 'MethodSignatureMustOmitReturnType'
]
```

src/Psalm/Checker/MethodChecker.php
```diff
@@ -660,4 +661,35 @@ public static function compareMethods(
...
+        MethodStorage $method_storage,
+        CodeLocation $code_location
+    ) {
+        if ($method_storage->signature_return_type === null) {
+            return null;
+        }
+
+        $cased_method_name = $method_storage->cased_name;
+        switch ($cased_method_name) {
+        case '__clone':
```
muglug: Is this a common pattern? I'm used to indenting `case` statements

rightfold: Don't know. I can change it to a series of `===`/`||` or an `in_array` check instead.

muglug: no I just meant I'm used to seeing/writing

```php
switch ($foo) {
    case 'bar':
        // do something
}
```

rightfold: Non-indented case is pretty common, yes. But from a quick look at the code base I see that it indents them everywhere.

muglug:
Thanks!

Issue #16, opened 2016-12-14
Add support for enums as parameters or return types


nickyr:
You always have to fake enums in PHP since they're not supported. Psalm could help this!

One option would be to do a syntax like this:

```php
/**
 * @return 'A'|'B'|'C'
 */
```

The quotes would differentiate the strings from classes, and this would fit pretty naturally with the way Psalm currently allows notation with `false` like

```php
/**
 * @return string|false
 */
```

Or, you could do something like

```php
/**
 * @return enum{'A', 'B', 'C'}
 */
```

which would kind of match the current fancy array syntax.

nickyr:
https://github.com/vimeo/psalm/issues/17 could make this even awesomer

muglug:
What do you think of https://flowtype.org/docs/builtins.html#literal-types? Pretty much identical to your first proposition. And it could be used together with https://github.com/vimeo/psalm/issues/18 to avoid repetition

erunion:
Seems like this might break from the standard PHPDoc `@return` annotation.

nickyr:
Looks good!

@erunion there's a `@psalm-return` annotation for things that don't match PHPDoc

muglug:
How about

```php
/*
 * @psalm-param enum-string('some-option'|Class::CONSTANT) $foo
 */
function takesConstant(string $foo) {}
```

erunion:
Oh I like that.

nickyr:
that looks good! but it should support ints as well as strings

muglug:
@JanTvrdik did you have any plan to support a particular format in PHPStan, given you already have value types?

muglug:
`@psalm-param enum-string("hello"|"goodbye")|enum-int(1, 2, C::A)`

and also

`@psalm-param keys-of(C::A)|keys-of(C::B)|values-of(C::D)`

weirdan:
`enum-string("hello"|"goodbye")` should probably use comma as a separator, to match that of `enum-int`.

I reckon it's just a syntax proposal at this stage, and it's not implemented yet, right?

muglug:
Actually it can be simpler like
`@psalm-param ('hello' | 'goodbye' | 1 | 2 | C::A)`
or (necessarily without spaces)
`@psalm-param 'hello'|'goodbye'|1|2|C::A`

nickyr:
>  @psalm-param 'hello'|'goodbye'|1|2|C::A

That's perfect

PR #707, opened 2018-05-03
Support value types

muglug:
[PHPStan](https://github.com/phpstan/phpstan) is adding support for constant types. I was very skeptical of its benefit, but spent a bit of time adding it to Psalm.

Turns out, despite my reservations, it found two bugs in Vimeo's codebase, so I'm putting the functionality into master in case it's useful for others.

Increases time taken by about 5-10%, depending on codebase size.

### DIFF
--- a/src/Psalm/Checker/Statements/Expression/Fetch/ArrayFetchChecker.php
+++ b/src/Psalm/Checker/Statements/Expression/Fetch/ArrayFetchChecker.php
@@ -309,6 +309,16 @@ class ArrayFetchChecker
                         }
                     }
 
+                    if (!$stmt->dim && $type->count && $type->count->values) {
+                        $new_counts = [];
+
+                        foreach ($type->count->values as $count => $_) {
+                            $new_counts[(string)((int)$count + 1)] = true;
+                        }
+
+                        $type->count->values = $new_counts;
+                    }
+
                     if ($in_assignment && $replacement_type) {
                         $type->type_params[1] = Type::combineUnionTypes(
                             $type->type_params[1],
@@ -409,10 +419,17 @@ class ArrayFetchChecker
                                 $offset_type
                             );
 
+                            $property_count = $type->sealed ? count($type->properties) : null;
+
                             $type = new TArray([
                                 $new_key_type,
                                 $generic_params,
                             ]);
+
+                            if (!$stmt->dim && $property_count) {
+                                ++$property_count;
+                                $type->count = new Type\Atomic\TInt([(string)$property_count => true]);
+                            }
 
                             if (!$array_access_type) {
                                 $array_access_type = clone $generic_params;
@@ -485,7 +502,7 @@ class ArrayFetchChecker
                 continue;
             }
 
-            if ($type instanceof TMixed || $type instanceof TGenericParam ||  $type instanceof TEmpty) {
+            if ($type instanceof TMixed || $type instanceof TGenericParam || $type instanceof TEmpty) {
                 $codebase->analyzer->incrementMixedCount($statements_checker->getCheckedFilePath());
 
                 if ($in_assignment) {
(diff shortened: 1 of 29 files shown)


src/Psalm/Checker/Statements/Expression/Fetch/ArrayFetchChecker.php
```diff
@@ -409,11 +421,18 @@ public static function getArrayAccessTypeGivenOffset(
...
                             );
 
+                            $property_count = $type->sealed ? count($type->properties) : null;
+
                             $type = new TArray([
                                 $new_key_type,
                                 $generic_params,
                             ]);
 
+                            if (!$stmt->dim && $property_count) {
```
TysonAndre: How are arrays of 0 length handled?

Why is count treated as a property instead of something calculated from the array fields (and whether or not they are possibly undefined)

src/Psalm/Type/Atomic/TString.php
```diff
@@ -15,4 +26,12 @@ public function getKey()
     {
         return 'string';
     }
+
+    /**
+     * @return string
+     */
+    public function getId()
+    {
+        return $this->values ? 'string(\'' . implode('\',\'', array_keys($this->values)) . '\')' : 'string';
```
TysonAndre: Long-term, this would need to be quoted (So would array shape keys)

A whitelist might include `\0x7f-\0xff` and anything that isn't a special character. 

E.g. `$x = "O')|string('" would be rendered as having a type of `string('O')|string('')

src/Psalm/Type/Atomic/TString.php
```diff
@@ -3,6 +3,17 @@
 
 class TString extends Scalar
```
TysonAndre: I've been thinking about something similar. What's the reason for choosing this implementation over `TLiteralString extends Scalar` (something that can cast to/from a TString).

E.g. if a variable could have three possible string values, it could be a union type of 3 TLiteralString types instead of a union type with a single type (with 3 possible strings), which would seem like it would make merging branches easier

TysonAndre:
Are there plans for a standard to indicate expected values via phpdoc (Or is that already there)? E.g. `@param string('on')|string('off') $flag` or something along those lines?

---

Unrelated notes on future implications of literals being in the type system (I think TypeScript does something similar):

`switch($strVal) { case 'someLiteral': echo $obj->$strVal; }` would be able to infer this is accessing $obj->someLiteral (Same for string concatenation of a literal and $strVal, or `$obj->{$strVal}()` invoking a method with that name (e.g. in __get()/__call())

src/Psalm/Type/Atomic/TString.php
muglug: Yes, this definitely needs to be toughened up a little

muglug:
> Are there plans for a standard to indicate expected values via phpdoc

@TysonAndre yeah it's one of the earliest open issues: https://github.com/vimeo/psalm/issues/16

And to the other points, definitely would be cool to integrate value types a little more.

Some people do horrible things e.g.

```php
$a = Class::METHOD_NAME;
$b->$a();
```

and it'd be cool to be able to resolve those, too.

src/Psalm/Type/Atomic/TString.php
muglug: No good reason - I think `TLiteralString` is a great classname.

src/Psalm/Checker/Statements/Expression/Fetch/ArrayFetchChecker.php
muglug: https://github.com/vimeo/psalm/blob/1a4ddc7cdd1b332e04405a11db2b5227e4b1587e/src/Psalm/Type.php#L689

The issue suppression is redundant now, will remove

Issue #722, opened 2018-05-08
simplexml_load_string should return false|SimpleXMLElement


jaydiablo:
Given this code:

```php
<?php
  
$content = 'xml content':
  
$simpleXml = simplexml_load_string($content);

if ($simpleXml === false) {
	echo 'error';
}
```

I get a `TypeDoesNotContainType` error:

>Cannot resolve types for $simpleXml - SimpleXMLElement does not contain false

According to PHP docs this function will return `false` on error: http://php.net/simplexml_load_string

>Returns an object of class SimpleXMLElement with properties containing the data held within the xml document, or FALSE on failure.

weirdan:
Similar to #262

jaydiablo:
Working on a PR to make the stubs match php.net's docs for SimpleXML classes/functions, worthwhile to add tests for this? Seems like most tests don't really cover the stubs (and in the referenced issue no tests were added for the change there).

weirdan:
@muglug I'm getting uncaught exception with this code, btw (latest psalm master)

muglug:
> Ah, ignore me, it's `throwExceptionOnError` in action.

Yeah I added that for debugging purposes over the weekend, will remove

PR #723, opened 2018-05-08
simplexml callmap update

weirdan:
fixes vimeo/psalm#722 and includes a number of manual updates to simplexml function/method signatures (based on php.net docs)


muglug:
Thanks!

PR #725, opened 2018-05-08
Updating zlib function stubs to match php.net docs

jaydiablo:
I noticed that some of the `false` error condition returns on a couple of these functions weren't being evaluated properly by psalm, so figured I'd just go through all of the zlib function from php docs and update what didn't match the docs: https://secure.php.net/manual/en/ref.zlib.php


muglug:
Thanks!

PR #727, opened 2018-05-09
Updating stubs for several string functions

jaydiablo:
Ran into some code that made use of the `stristr` function, and was checking if the return was `false` and psalm complained:

```php
<?php
$result = stristr('s', 'string');

if ($result === false) {
}
```

https://getpsalm.org/r/40ee44d5ae

Looked through a number of other string functions on php.net and updated any stubs that weren't correct.  https://secure.php.net/manual/en/ref.strings.php (I didn't have time to look at all of them though).


muglug:
Thanks - I've also (separately) made `str_ireplace` and `substr_replace` behave like `str_replace`, inference-wise.

PR #752, opened 2018-05-16
Updating openssl callmaps to reflect php documentation

jaydiablo:
Ran into a case where an openssl function wasn't marked as returning `false` but our code was checking that condition.

Look through PHP docs on the openssl functions, and updated the callmaps as necessary.

https://secure.php.net/manual/en/ref.openssl.php


muglug:
Thanks!

PR #759, opened 2018-05-20
Updated MessageFormatter (ICU) signatures

weirdan:
Manual update of MessageFormatter methods signatures, based on PHP.net docs


muglug:
Thanks!

PR #777, opened 2018-05-27
Remove unused variables

TysonAndre:
$method_name is overwritten 3 lines below in `explode('::', $declaring_method_id);`

use_case_sensitive_file_names seems to be a WIP


muglug:
Psalm's dead code detection now catches these issues.

PR #784, opened 2018-05-31
Fixing mysqli callmap

jaydiablo:
I ran into a case today where Psalm was complaining about `mysqli::commit` not having any parameters, but 2 parameters were being passed.

According to docs that's not the case: http://php.net/manual/en/mysqli.commit.php

So I took the time to not only fix that, but also get the rest of the mysqli/mysqli_result/mysqli_stmt methods/functions in the callmap to match the php docs as well.


src/Psalm/CallMap.php
```diff
@@ -7051,108 +7052,96 @@
...
+'mysqli_send_query' => ['bool', 'link'=>'mysqli', 'query'=>'string'],
+'mysqli_set_charset' => ['bool', 'link'=>'mysqli', 'charset'=>'string'],
+'mysqli_set_local_infile_default' => ['void', 'link'=>'mysqli'],
 'mysqli_set_local_infile_handler' => ['bool', 'link'=>'mysqli', 'read_func'=>'callable'],
 'mysqli_slave_query' => ['bool', 'link'=>'mysqli', 'query'=>'string'],
 'mysqli_sqlstate' => ['string', 'link'=>'mysqli'],
 'mysqli_ssl_set' => ['bool', 'link'=>'mysqli', 'key'=>'string', 'cert'=>'string', 'ca'=>'string', 'capath'=>'string', 'cipher'=>'string'],
-'mysqli_stat' => ['', 'link'=>'mysqli'],
-'mysqli_stmt::__construct' => ['void', 'link='=>'mysqli', 'query='=>'string'],
-'mysqli_stmt::affected_rows' => ['int'],
```
jaydiablo: I wanted to point out that I did remove a few lines like this one where they're not actual methods, but rather properties on the class (http://php.net/manual/en/mysqli-stmt.affected-rows.php).  I assume this is correct (as I understand it, this file is for functions/methods, not class properties). If I'm wrong in that assumption, let me know and I can re-instate.

I did confirm that the ones I removed aren't actually functions via reflection:

```
$ref = new ReflectionMethod('mysqli_stmt', 'affected_rows');
PHP Warning:  Uncaught ReflectionException: Method mysqli_stmt::affected_rows() does not exist in php shell code:1
Stack trace:
#0 php shell code(1): ReflectionMethod->__construct('mysqli_stmt', 'affected_rows')
#1 {main}
  thrown in php shell code on line 1
```

muglug:
Thanks! Definitely right to remove properties

PR #787, opened 2018-06-01
array_column return type inference

weirdan:
It takes a safe approach and only tries to be smart when array row type is definite single shape. In other words, Psalm now will be able to infer `array<string,int>` from 
```php
$c = array_column([["k" => "a", "v" => 1], ["k" => "b", "v" => 2]], "v", "k");
```
call.


muglug:
Thanks!

PR #788, opened 2018-06-01
Attempt to fix builds on scrutinizer-ci

weirdan:
- Dropped php56 (this branch is not compatible with php5 anymore due to nikic/php-parser:^4 dependency)
- Dropped postgresql and redis from environment (they don't seem to be used)


weirdan:
Oh, it worked.

muglug:
Thanks!

Issue #815, opened 2018-06-13
Linking variables and functions to definitions


BlakeMScurr:
Hey yo, 

I was using https://github.com/nikic/PHP-Parser (which I understand you use as a base for your [DependencyFinderVisitor](https://github.com/vimeo/psalm/blob/master/src/Psalm/Visitor/DependencyFinderVisitor.php#L37)), and I hit some limitations that the maintainer [said](https://github.com/nikic/PHP-Parser/issues/508) you may have worked on here. 

In particular, I was trying to find all call sites of a function from its definition and vice versa, and find the declaration of a variable from a place where it's used. My hope is that the info required for this is in your [storage objects](https://github.com/vimeo/psalm/blob/master/docs/how_psalm_works.md#storing-data-from-scanning) after parsing - does it seem plausible? Is there any extension that needs to be done to make it happen?

weirdan:
There's [Psalm\CodeBase::findReferencesToMethod()](https://github.com/vimeo/psalm/blob/6bfde90db4f676a98b48270074a93f1fa406d49c/src/Psalm/Codebase.php#L319) (and a similar method for class-like things), but I strongly suspect it requires running full analysis pass before you can call that.

muglug:
Hey, yeah as @weirdan says, it requires a full analysis pass, because to correctly understand 

```php
$foo->bar()->bat()->bang()
```

every individual expression needs to be resolved, which can only be done in the analysis phase.

[These two methods](https://github.com/vimeo/psalm/blob/dc02ca8a7a325e0eca0db2e4c5d9a8c8ebe9297c/src/Psalm/Codebase.php#L314-L379) find uses of methods and classes when `Codebase::$collect_references` is turned on.

`Psalm\Storage\FunctionLikeStorage::$referencing_locations` contains all the referencing locations that Psalm can reason about for methods (regular functions aren't yet supported, but wouldn't be hard to add).

`Psalm\Storage\ClassLikeStorage::$referencing_locations` contains the respective ones for classes, traits, interfaces etc.

Obviously the analysis isn't necessarily complete - there may often be times when Psalm cannot infer the

BlakeMScurr:
Nice, thanks for the help, guys. This is exactly what I'm looking for with respect to method references.

[...]

PR #832, opened 2018-06-21
Add find all references command

BlakeMScurr:
Output all references to functions as a JSON dump of the call graph when  `--find-references-to="all"` is set. 

This may not be generally useful, but it solves my own [issue](https://github.com/vimeo/psalm/issues/815).

### DIFF
--- a/src/Psalm/CallGraphNode.php
+++ b/src/Psalm/CallGraphNode.php
@@ -1,0 +1,45 @@
+<?php
+namespace Psalm;
+
+class location {
+    public $start_line;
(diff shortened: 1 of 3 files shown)


weirdan:
It probably would be better to have a separate config switch for that. `all` could happen be a valid class / function name users may want to search for.

muglug:
Hey Blake - would you be able to fix the issues that Psalm found? You can run `./psalm` on the codebase, then `./psalm --find-dead-code` once the initial errors have gone away.

Issue #852, opened 2018-06-30
Integer comparision after decrementing a value in a do-while loop triggers error


mjaschen:
The following code creates a false positive error in Psalm 1.1.4 (it worked fine up to Psalm 1.1.3):

```php
<?php
$i = 5;
do {} while (--$i > 0);
echo $i === 0;
```

The condition `$i === 0` evaluates of course to *true* after the loop finishes.

Demonstration of this issue: https://getpsalm.org/r/556d8678bc

Psalm doesn't complain when using a while-loop:

```php
<?php
$i = 5;
while (--$i > 0) {}
echo $i === 0;
```

Demostration: https://getpsalm.org/r/2fed4dd9ca

mjaschen:
PR #853 provides a test case for this behaviour.

PR #853, opened 2018-06-30
add test for a false-positive number equality

mjaschen:
This ist a test case for issue #852


muglug:
Thanks, this is now in master

PR #869, opened 2018-07-06
Fix misuse of preg_quote

TysonAndre:
1. Variable ids can contain slashes or dots, e.g. for `${'my/var'} = 2`
   or psalm's local representation of properties or array indexes
   (I forget which)
2. preg_quote's escape character should **always** be the same as the
   pattern starting character.
   Using DIRECTORY_SEPARATOR will fail to escape '/' on windows.

See https://secure.php.net/manual/en/function.preg-quote.php#refsect1-function.preg-quote-parameters and https://secure.php.net/manual/en/function.preg-quote.php#refsect1-function.preg-quote-examples

### DIFF
--- a/src/Psalm/Context.php
+++ b/src/Psalm/Context.php
@@ -477,7 +477,7 @@ class Context
             $quoted_remove_var_id = preg_quote($remove_var_id);
 
             foreach ($clause->possibilities as $var_id => $_) {
-                if (preg_match('/' . $quoted_remove_var_id . '[\]\[\-]/', $var_id)) {
+                if (preg_match('/' . preg_quote($quoted_remove_var_id, '/') . '[\]\[\-]/', $var_id)) {
                     break 2;
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Psalm/Context.php
```diff
@@ -477,7 +477,7 @@ public static function filterClauses(
             $quoted_remove_var_id = preg_quote($remove_var_id);
 
             foreach ($clause->possibilities as $var_id => $_) {
-                if (preg_match('/' . $quoted_remove_var_id . '[\]\[\-]/', $var_id)) {
+                if (preg_match('/' . preg_quote($quoted_remove_var_id, '/') . '[\]\[\-]/', $var_id)) {
```
TysonAndre: I was seeing `Warning: preg_match(): Unknown modifier 'a' in /path/to/project/vendor/vimeo/psalm/src/Psalm/Context.php on line 480` here

muglug:
Thanks!

src/Psalm/Context.php
muglug: I undid this change somewhat in 2100166, as it caused bugs

TysonAndre: The new change makes more sense, I was looking at the variable name but not the definition

PR #876, opened 2018-07-09
875 addendum

SignpostMarv:
checking port args for parse_url().

### DIFF
--- a/tests/FunctionCallTest.php
+++ b/tests/FunctionCallTest.php
@@ -857,25 +857,13 @@ class FunctionCallTest extends TestCase
                     }
 
                     $port1 = portismaybeint("");
-                    $port2 = portismaybeint("localhost");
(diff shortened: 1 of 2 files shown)


tests/FunctionCallTest.php
```diff
@@ -838,8 +838,34 @@ function hereisthelastone(string $s) : string {
...
+                            return $parsed["port"];
+                        }
+
+                        return 80;
+                    }
+
+                    function portismaybeint(string $s) : ? int {
+                        $parsed = parse_url($s);
+
+                        $parsed["port"] ?? null;
```
muglug: this doesn't return anything

SignpostMarv: fixed in 2cbc55e

muglug:
Thanks!

PR #877, opened 2018-07-10
Simplify installation of a development version

sanmai:
Simplify installation of a development version from Composer.

https://getcomposer.org/doc/articles/aliases.md#branch-alias

(This probably needs backporting to 1.x to work.)


muglug:
Thanks, will backport

PR #880, opened 2018-07-10
Allow empty default of `0` for optional parameter in magic method

### LINKED ISSUE #879: Another false positive MisplacedRequireParam for magic method parameter default of `0` (Checking for empty())?
https://getpsalm.org/r/fd3b7fe84d

```
<?php

/**
 * @method int|string|array eval(string $script, array $args = [], int $num_keys = 0)
 */
class RedisWrapper {
    public function __call(string $method, array $args) {}
}
````

Expected: no error, `= 0` is a reasonable default to indicate something is optional (Is this using `empty($default)` ?)

Observed:

```
ERROR: MisplacedRequiredParam - 7:58 - Required param $num_keys should come before any optional params in @method eval
```

TysonAndre:
Fixes #879

`!= ''` will reject both null and '', but not `0`


muglug:
Thanks!

PR #888, opened 2018-07-13
Update PHPUnit

weirdan:
This updates to PHPUnit 6/7, including the minor tweaks to the tests that were needed to satisfy Psalm and PHP for various PHP/PHPUnit combinations.

One Psalm failure is still there (PHP 7.0 / low-deps, https://travis-ci.com/weirdan/psalm/jobs/134301736#L855), but I hope it'll be fixed in phpunit (reported here: sebastianbergmann/phpunit#3209)

Nightly fails for unrelated reason (phpcs has some ugly switch/continue code).

Also it adds another variable into build matrix, causing travis to run with both lowest possible and highest possible dependencies. That's useful to see if your dependency version constraints actually make sense.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,10 +29,11 @@ cache:
 
 before_install:
     # determine INI file
-    - export INI=~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+    - export INI_DIR=~/.phpenv/versions/$(phpenv version-name)/etc/conf.d
+    - export INI=$INI_DIR/travis.ini
     # disable default memory limit
     - echo memory_limit = 2G >> $INI
-    - if [[ $COVERALLS = "" ]]; then phpenv config-rm xdebug.ini; fi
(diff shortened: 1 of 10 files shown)


weirdan:
> I erroneously disabled coveralls on 7.2 when I added nightly, but it only needs to run on a single build

That's how it's already works in this PR, but I'll rebase to fix the conflicts. I'm going to mark 7.0/high as allowed_failure for the time being, until sebastianbergmann/phpunit#3209 is released in PHPUnit 6.5.10

weirdan:
@muglug I got all green on builds here and on my own repo. Coveralls build has truncated output for vimeo/psalm build though.

weirdan:
Note that travis build matrix now uses *exclude* instead of *include*

muglug:
Thanks! Sorry for my confusion!

muglug:
Also the test for low deps caught a bug with a new feature, thanks for that!

PR #924, opened 2018-07-31
Fix typos detected by `misspell`

TysonAndre:
Both of initiali[sz]e were used, z was more common.


muglug:
Thanks! I tend to use American English for code and British English for documentation, but it makes sense to use American English for code comments.

Issue #192, opened 2017-07-24
Incorrect detection of type defined by interface


mkoppmann:
Hi there, thanks for this great project.

I currently have a problem with Psalm detecting the wrong wanted type for an argument. When I try to implement the PHP [SessionHandlerInterface](https://secure.php.net/manual/en/sessionhandlerinterface.gc.php), I get the following error:

```php
class MySessionHandler implements \SessionHandlerInterface
{
  […]
    /**
     * Cleanup old sessions
     * @link http://php.net/manual/en/sessionhandlerinterface.gc.php
     * @param int $maxlifetime <p>
     * Sessions that have not updated for
     * the last maxlifetime seconds will be removed.
     * </p>
     * @return bool <p>
     * The return value (usually TRUE on success, FALSE on failure).
     * Note this value is returned internally to PHP for processing.
     * </p>
     * @since 5.4.0
     */
    public function gc($maxlifetime)
    {
        // TODO: Implement gc() method.
        return true;
    }
  […]
}
```

```shell
ERROR: MethodSignatureMismatch - path/MySessionHandler:59 - Argument 1 of path\MySessionHandler::gc has wrong type 'int', expecting 'string' as defined by SessionHandlerInterface::gc
    public function gc($maxlifetime)
```

But this test file works fine:
```php
<?php

interface TestInterface
{
    /**
     * @param int $maxlifetime
     * @return bool
     */
     public function gc($maxlifetime);
}

class TestHandler implements TestInterface
{
    /**
     * @param int $maxlifetime
     * @return bool
     */
    public function gc($maxlifetime)
    {
        return true;
    }
}
?>
```

Also, adding `@psalm-suppress MethodSignatureMismatch` doesn’t suppress the error.

I’m using the lastest version of Psalm from master. Thanks for your help.

muglug:
Thanks for the praise! Should be fixed!

mkoppmann:
Wow, that was incredibly fast. Thanks a lot.

PR #939, opened 2018-08-13
Fix for not suppressing MethodSignatureMismatch and InvalidReturnType

pieterdeschepper:
If you have an inherited class that overrides a method with a different return type, you can now suppress it with
@psalm-suppress MethodSignatureMismatch

Example code:

```
<?php
declare(strict_types=1);

class Animal
{
	public function getId(): string
	{
		return "Animal";
	}
}

class Dog extends Animal
{
	/**
	 * @psalm-suppress MethodSignatureMismatch
	 */
	public function getId(): int
	{
		return 1;
	}
}
```

Also related to the last line in issue #192

### DIFF
--- a/src/Psalm/Checker/FunctionLike/ReturnTypeChecker.php
+++ b/src/Psalm/Checker/FunctionLike/ReturnTypeChecker.php
@@ -318,7 +318,8 @@ class ReturnTypeChecker
                         'No return statements were found for method ' . $cased_method_id .
                             ' but return type \'' . $declared_return_type . '\' was expected',
                         $return_type_location
-                    )
+                    ),
+                    $suppressed_issues
                 )) {
                     return false;
                 }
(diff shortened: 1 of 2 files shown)


muglug:
Why is this needed? `MethodSignatureMismatch` should always correspond to a fatal PHP error. Otherwise it shouldn't be emitted (and that's a bug).

pieterdeschepper:
I had it in some left-over code that needed afterwards cleanup, but was not possible at that moment (because of going down a major refactor).
Since we rely on psalm for our validation on commits, it would be use-full to be able to also suppress issues like that, even if they would result in a fatal PHP error.
It's just more consistent to be able to suppress issues (any issue), if you have the option to do so.

dbalabka:
It is impossible to suppress such errors in Psalm `3.18.2`. I'm fully supporting @muglug that there should not be such a possibility to suppress fatal errors. Still, there is a scenario when it does not lead to a fatal error:
https://psalm.dev/r/187f6eec4c
Psalm treat `@fatal` annotation same as `final class`. In my case, I've solved the issue by removing the entire class, so I think it is `wont-fix` problem.

PR #1006, opened 2018-09-30
Language Server Protocol support

### LINKED ISSUE #521: Explore language server compatibility
This is a v3 feature, may never happen. But it'd be cool.

muglug:
Piggybacks off @TysonAndre's early adaptations to @felixfbecker's LSP support.

Only supports error/warning checking at the moment and just uses a single thread. It builds maps for definition lookup, but doesn't currently use them.

Time to analyse a single-line change in a 10,000-line file is down to well-under a second, but this can definitely be improved even further.

Fixes #521

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,11 +11,18 @@
     ],
     "require": {
         "php": "^7.0",
-        "nikic/PHP-Parser": "^4.0",
+        "nikic/PHP-Parser": "dev-master",
         "openlss/lib-array2xml": "^0.0.10||^0.5.1",
         "muglug/package-versions-56": "1.2.4",
         "php-cs-fixer/diff": "^1.2",
-        "composer/xdebug-handler": "^1.1"
+        "composer/xdebug-handler": "^1.1",
+        "felixfbecker/language-server-protocol": "^1.2",
+        "felixfbecker/advanced-json-rpc": "^3.0.3",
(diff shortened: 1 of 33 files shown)


src/Psalm/Fork/Pool.php
```diff
@@ -253,15 +253,29 @@ function ($data) {
...
 
+        $command = $argv[0];
+
         // Wait for all children to return
         foreach ($this->child_pid_list as $child_pid) {
-            posix_kill($child_pid, SIGALRM);
-            if (pcntl_waitpid($child_pid, $status) < 0) {
-                error_log(posix_strerror(posix_get_last_error()));
+            /** @psalm-suppress ForbiddenCode */
+            $process_lookup = @shell_exec(
```
TysonAndre: Can this use `--pid`? Also, should check if windows once you're done prototyping.

```
       --pid pidlist
              Select by process ID.  Identical to -p and p.
```

muglug: Ah thanks!

Re Windows: I [commented out](https://github.com/vimeo/psalm/pull/1006/commits/91376b8cdc665d033baa21bb76f65007692db8c9#diff-495e0ae9fc6e4853664145d5a9e1985bR915) the bit in `getCpuCount` where we get the correct count of windows processes - given Windows doesn't support `pcntl`, this area should be all *nix-only.

PR #1012, opened 2018-10-08
Bump Box

theofidry:
- Bump box to the latest stable release
- Remove unncessary settings from `box.json.dist`
- Add the `Php` compactor which should remove some unnecessary comments and whitespaces from the
  PHAR


muglug:
Thanks!

PR #1013, opened 2018-10-09
Update signature for various xml functions

fisharebest:
FYI, I made the same change to phan https://github.com/phan/phan/pull/2031


muglug:
Thanks!

PR #1018, opened 2018-10-11
ADD extra documentation about psalm-assert and how it can handle null checks

DaveLiddament:
Maybe this will be obvious to others, but it probably took me about 3 hours to realise this was possible. Hopefully this doc update will save others time too.


muglug:
Thanks! Great clarification!

PR #1037, opened 2018-10-25
DateTime*::getTimestamp can return false

fkooman:
When the represented date is after 2038-01-19 03:14:07 on 32 bit systems.

Initially I thought it would also return false on dates before 1970-01-01 00:00:00, but it returns a negative integer instead... Oh well.


muglug:
Thanks!

fkooman:
This PR was undone somehow...

muglug:
Yes, in https://github.com/vimeo/psalm/commit/2f30132d#diff-4355ecf0bcc89fe8c145b64be0e02ddfR1691 - I'll patch, sorry

PR #1041, opened 2018-10-27
Fix wording nits in documentation

TysonAndre:
Feel free to cherry-pick the changes that make sense


muglug:
Thanks - I normally err on the side of too many commas (for Americans, at least), but I guess I came up short here.

PR #1043, opened 2018-10-30
Fixed the type of 1st parameter in odbc_fetch_object()

sj-i:
odbc_fetch_object() takes resource in its 1st parameter, not integer.
http://php.net/manual/en/function.odbc-fetch-object.php


muglug:
Thanks!

PR #1049, opened 2018-10-31
Introduce issue baseline with --set-baseline and --with-baseline

ErikBooijCB:
This PR aims to introduce an issue baseline for Psalm. The issue baseline will mark existing errors as `INFO` during subsequent test runs.

The baseline will contain an overview of files with errors and the number of occurrences. That way adding another instance of an existing issue, will still cause the test run to fail.

The file will look like this:

```xml
<files>
  <file src="sample/sample-file.php">
    <MixedAssignment occurrences="3"/>
    <InvalidReturnStatement occurrences="1"/>
    <MissingParamType occurrences="1"/>
    <InvalidReturnType occurrences="1"/>
  </file>
  <file src="sample/sample-file2.php">
    <MixedAssignment occurrences="2"/>
  </file>
</files>

```

To create the baseline, use:

`--set-baseline=baseline.xml`

To run Psalm while taking the baseline into consideration, use:

`--with-baseline=baseline.xml`

To update the baseline, run:

`--update-baseline=baseline.xml`

Updating will only ever remove issues from the baseline, or decrease their occurrence count, never add any.

As you can see I chose to not automatically update the baseline while running Psalm, because I doubt that would be desired behaviour in most cases. While running in a CI environment that's not writable for example, or when actively working on code and testing in the meantime, you might not want issues to disappear from the baseline while you're still working on it.

If you have comments/suggestions, let me know 👍

### DIFF
--- a/src/psalm.php
+++ b/src/psalm.php
@@ -197,7 +197,7 @@ Options:
     --set-baseline=PATH
         Save all current error level issues to a file, to mark them as info in subsequent runs
 
-    --ignore-baseline=PATH
+    --ignore-baseline
         Ignore the error baseline
         
     --update-baseline
--- a/src/psalm.php
+++ b/src/psalm.php
@@ -1,6 +1,7 @@
 <?php
 require_once('command_functions.php');
 
+use Psalm\ErrorBaseline;
 use Psalm\Checker\ProjectChecker;
 use Psalm\Config;
 use Psalm\IssueBuffer;
@@ -25,10 +26,12 @@ $valid_long_options = [
     'debug',
     'debug-by-line',
     'diff',
+    'diff-methods',
     'disable-extension:',
     'find-dead-code',
     'find-references-to:',
     'help',
+    'ignore-baseline',
     'init',
     'monochrome',
     'no-cache',
@@ -36,13 +39,14 @@ $valid_long_options = [
     'plugin:',
     'report:',
     'root:',
+    'set-baseline:',
     'show-info:',
     'show-snippet:',
     'stats',
     'threads:',
+    'update-baseline',
     'use-ini-defaults',
     'version',
-    'diff-methods',
 ];
 
 $args = array_slice($argv, 1);
@@ -189,6 +193,15 @@ Options:
 
     --disable-extension=[extension]
         Used to disable certain extensions while Psalm is running.
+        
+    --set-baseline=PATH
+        Save all current error level issues to a file, to mark them as info in subsequent runs
+
+    --ignore-baseline=PATH
+        Ignore the error baseline
+        
+    --update-baseline
+        Update the baseline by removing fixed issues. This will not add new issues to the baseline
 
 HELP;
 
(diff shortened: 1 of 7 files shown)


weirdan:
> will contain an overview of files with errors and the number of occurrences.

Edge case: fixing one occurrence of a specific issue and introducing it somewhere else in the same file between runs will not cause the introduced issue to be reported as error. Not much of a problem, but probably worth documenting.

src/Psalm/Baseline.php
```diff
@@ -0,0 +1,159 @@
...
+     * @throws Exception\ConfigException
+     */
+    public static function update(string $baselineFile, array $issues)
+    {
+        $existingIssues = self::read($baselineFile);
+        $newIssues = self::countIssueTypesByFile($issues);
+
+        foreach ($existingIssues as $file => &$existingIssuesCount) {
+            if (!isset($newIssues[$file])) {
+                unset($existingIssues[$file]);
```
muglug: Would you be able to add a test case for this? It's the only block not covered

src/psalm.php
```diff
@@ -252,6 +256,26 @@ function ($arg) use ($valid_long_options, $valid_short_options) {
...
+if (isset($options['set-baseline'])) {
+    if (is_array($options['set-baseline'])) {
+        die('Only one baseline file can be created at a time' . PHP_EOL);
+    }
+}
+
+$issue_baseline = [];
+
+if (isset($options['with-baseline'])) {
+    if (is_array($options['with-baseline'])) {
```
muglug: Please add an error if `!is_string($options['with-baseline'])`

src/Psalm/Baseline.php
```diff
@@ -0,0 +1,159 @@
+<?php
+namespace Psalm;
+
+class Baseline
```
muglug: One last thing - I have a slight feeling the `ErrorBaseline` might be better. My I don't have any firm reason for that, and maybe `Baseline` is obvious enough.

src/Psalm/Baseline.php
ErikBooijCB: Yes, good one 👍  had that covered before but missed another case, thought I fixed that, but didn't verify because the coverage run took too long.

ErikBooijCB:
Thanks for the quick feedback!

Fixed most of the issues, but am running into a bit of trouble automatically adding the baseline file to the config, since most of the locating of a config file (if you don't explicitly add one) actually happens within the config file itself.

For now I've implemented it with a message reminding the user to add the baseline file to the config. If you have a suggestions on a proper way to automatically add it to the config, I'd be happy to take another look at that.

I've removed the `--with-baseline` and changed it to `--ignore-baseline`.

`--update-baseline` does not take a file argument anymore and simply updates the baseline defined in the config file.

Files and issues in the baseline are now sorted alphabetically and only include actual error level issues.

Used the `FileProvider` instead of native functions. Removed the virtual filesystem dependency again.

Added typehints for the `IssueBuffer::finish` function.

Renamed the `Baseline` class to `ErrorBaseline`.

weirdan:
> since most of the locating of a config file (if you don't explicitly add one) actually happens within the config file itself.

Had similar problem in my plugin-loading branch, solved that by extracting the config file search to its own method: https://github.com/vimeo/psalm/pull/855/commits/31f643b87ec2b4aebe84e7c99cd40abfb9798c61 .

muglug:
> solved that by extracting the config file search to its own method: 31f643b .

@weirdan I've added that in https://github.com/vimeo/psalm/commit/1d7ea1292ec7adcc9091b2038fbf78cea433d11b

PR #1064, opened 2018-11-06
add nullable bool result to `mysql_stmt::fetch`

vasily-kartashov:
According to documentation it can return `true`, `false` and `null`, where `null` indicates the end of result set. See here: http://php.net/manual/en/mysqli-stmt.fetch.php


muglug:
Thanks!

PR #1076, opened 2018-11-11
Plugin interface segregation

weirdan:
Replaces Plugin class with `Psalm\PluginApi\Hook\*` interfaces


weirdan:
Now that's Plugin is gone it's an option. It is API for plugins though, and
I imagine Plugin manager may want to operate on some plugins, so there are
merits for keeping it available

muglug:
Thanks!

PR #1078, opened 2018-11-12
Re-enable plugin autoloading

weirdan:
Plugins are supposed to be autoloaded when they are referenced by class names. This check prevented that.


muglug:
Oops, sorry!

PR #1080, opened 2018-11-12
Don't choke on non-existing composer.lock

weirdan:
It's useful to see enabled plugins and be able to disable them even if no composer files are found.


muglug:
Awesome, thanks!

PR #1082, opened 2018-11-14
reinstate extra.psalm "namespace"

weirdan:
Existing plugins have been updated as well (skeleton, doctrine-collections)

Also fixed path to echo-checker plugin so that it's once again directly installable from the psalm distribution.


muglug:
Thanks!

PR #1085, opened 2018-11-15
Fix typos noticed in docs

TysonAndre:


muglug:
Thanks!

muglug:
Do you understand why I created Psalm now? Also, emailed you two days ago (but from a different, non-github address) - not sure you received

TysonAndre:
> Do you understand why I created Psalm now? 

Yes - because it's useful. I'm not sure why you ask?

>  Also, emailed you two days ago 

I received that message. Thanks - I didn't know if you were expecting a response.

muglug:
> > Do you understand why I created Psalm now?

> Yes - because it's useful. I'm not sure why you ask?

It was a joke - because of the occasional random mistakes in my work, a few of which were fixed in this PR

> I didn't know if you were expecting a response.

Oh I wasn't really, just thought there was a chance it might have gone to junk bc it was from a different address. Anyway, carry on.

PR #1089, opened 2018-11-21
Fix usage of substr_count($haystack, $needle)

TysonAndre:
Detected via a check that literals and variables occur in the most
typical order for a global function.

See http://php.net/manual/en/function.substr-count.php#refsect1-function.substr-count-examples


muglug:
Thanks! This escaped notice because those values are just sent back to the host IDE

Issue #15, opened 2016-12-13
Emit issues when encountering uncaught throw


muglug:
Both when

 - calling a method that has `@throws` in its docblock and the callee doesn't also have a matching `@throws`
 - when a method throws an exception and doesn't have a `@throws` in the docblock.

cc @nickyr

muglug:
I don't think this is useful (but this could be made into a plugin)

jaydiablo:
Just a heads up, the release notes for 2.0.5 (https://github.com/vimeo/psalm/releases/tag/2.0.5) state to use the config option `checkThrowsDocblock="true"` but it's actually `checkForThrowsDocblock="true"` (https://github.com/vimeo/psalm/commit/e3ae1bf26f1501fb14f760cd33adf092448e2f47#diff-52926e4f815aeddeeac38218c09784f5R43).

muglug:
Thanks, fixed

PR #1092, opened 2018-11-24
Fix config option typo

aidantwoods:
Similar to: https://github.com/vimeo/psalm/issues/15#issuecomment-400170628


muglug:
Thanks!

PR #1106, opened 2018-11-29
Fixed collator method return types

weirdan:
This fixes signatures for `Collator::compare()` and `Collator::getSortKey()` methods that can fail with `false` return value when passed strings that cannot be converted to utf-16, as well as `Collator::create()` constructor that can possibly return `null`


muglug:
Thanks!

Issue #1163, opened 2019-01-02
[bug] filter_var - incorrect handling for default option


andrew-demb:
FunctionAnalyzer has wrong handling for filter_var options.
It expects, that $options will be like
```php
['default' => 'value']
```

But correct is
```php
['options' => ['default' => 'value']]
```

https://github.com/vimeo/psalm/blob/3.0.7/src/Psalm/Internal/Analyzer/FunctionAnalyzer.php#L624-L635
tests:
https://github.com/vimeo/psalm/blob/3.0.7/tests/FunctionCallTest.php#L1215-L1220
https://github.com/vimeo/psalm/blob/3.0.7/tests/FunctionCallTest.php#L1237-L1239

PR #1164, opened 2019-01-02
:bug: #1163 Fix handling for default option in `filter_var`

andrew-demb:


andrew-demb:
I don't add regression tests, and may break phpcs rule for line length.
Please fix, if it necessary.

PR #1204, opened 2019-01-16
Tweak error message

ro0NL:
got this error, looked at

https://github.com/vimeo/psalm/blob/90dc39c29651a7cc20bc8ff594ce1ea5d10a5f8b/src/Psalm/Config.php#L441-L445

ended up debugging a global PHAR to see if its config.xsd was missing, etc.

finally installed psalm locally, and figured out my autoloader path was wrong :joy:


muglug:
Thanks!

PR #1212, opened 2019-01-19
Use Amp for handling language server event loop

muglug:
`sabre/event` turns out to have more conflicts given its popularity and how broad the `sabre` suite is.


src/Psalm/Internal/Stubs/SabreEvent.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace Sabre\Event;
+namespace Amp;
```
TysonAndre: Nit: Could rename this file from SabreEvent.php now that it's not for Sabre

muglug:
Thanks @trowski for your input!!

PR #1224, opened 2019-01-21
Fixes #1222 - incorrect config value used for `allowStringToStandInForClass`

### LINKED ISSUE #1222: False-Positive InvalidStringClass when string is passed and used for class
Please see this code example:

https://getpsalm.org/r/abae86f87f

It reports that the string cannot be used as a class although the string is just used for instantiating the new class.

andrew-demb:
Fixes #1222

Please make attention, that `allowCoercionFromStringToClassConst` are mentioned in docs, but not used now in config

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -570,7 +570,7 @@ class Config
         }
 
         if (isset($config_xml['allowStringToStandInForClass'])) {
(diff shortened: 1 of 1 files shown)


muglug:
Thank you!

andrew-demb:
@muglug maybe we need to drop from docs mentions about `allowCoercionFromStringToClassConst`, because it's not used anywhere in the project now.

PR #1237, opened 2019-01-24
Fix typo

tvdijen:


muglug:
So, ah, this isn't a typo. I meant "provably low". Not quite sure how I though it could be proven, but anyway.

PR #1264, opened 2019-02-01
Remove redundant catching and rethrowing

TysonAndre:
I assume these used to have additional logging,
but they're no-ops right now.

Noticed while looking at recent changes, found others by grepping


muglug:
Thanks!

PR #1270, opened 2019-02-01
Set proper return type for ldap methods

tvdijen:
http://php.net/manual/en/function.ldap-count-entries.php
"Returns number of entries in the result or FALSE on error."

http://php.net/manual/en/function.ldap-get-dn.php
"Returns the DN of the result entry and FALSE on error."


weirdan:
Are you trying to change `src/Psalm/Internal/CallMap.php`? It appears you pushed it as `src/Psalm/CallMap.php`, creating a new file.

tvdijen:
Ohh, that sucks.. Dunno what went wrong, but it's fixed now!

muglug:
Thanks!

PR #1278, opened 2019-02-04
Use lowercase nikic/php-parser for composer 2.0 compatibility

TysonAndre:
Seen in composer 1.8.3:

> Deprecation warning: require.nikic/PHP-Parser is invalid, it should not
> contain uppercase characters. Please use nikic/php-parser instead. Make
> sure you fix this as Composer 2.0 will error.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": "^7.0",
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

PR #1289, opened 2019-02-06
Update Amp usage

trowski:
This supplements the refactor from Sabre Event to Amp in #1212 with a few changes to fix some errors in Promise usage and simplifies reading and writing messages in the Language Server.

I'm not sure what I'm doing with the docblock annotations and stubs for Amp. If someone could offer some guidance as to how to fix them I'd be happy to update this PR. Running `./psalm --find-dead-code` results in several errors related to closures I declared and `Success` not being compatible with a declared `Promise<…>` return type.

I think incorporating Amp into this project will be a great step forward. `amphp/process` and `amphp/parallel` could be used to simplify a lot of the parallel process code. I didn't have the time to dig into exactly how this should be done, but would be happy to assist in the future if the maintainers are interested in further refactoring.


muglug:
Thanks - I'll work on this further in the branch, and will ping you if I have questions.

Part of the challenge, return type wise, is getting Psalm to understand this templated param type: https://getpsalm.org/r/16e221f44f

It currently things `TReturn` is a union of `string` and `Generator<mixed,mixed,mixed,string>` when I want it to just match the most specific thing it found.

muglug:
https://github.com/vimeo/psalm/pull/1212/commits/c0282d6c32801927a7c3e944c9266ad92c5e4627 contains what the stubbed return types should be, but I need to change the above behaviour and use a plugin to tell Psalm how `yield $promise` actually yield's that promise's value.

muglug:
This is broken at the moment - the current implementation buffers messages, and emits a `readMessageGroup` event once all the messages have been consumed which in turns triggers analysis (so multiple sequential calls only result in a single Psalm analysis run).

I maintained that pattern (I think) in the original WIP branch, but these changes remove that emitted event (so analysis never actually happens).

Does using `byte-stream` obviate the message buffering?

trowski:
Oh, I see where I went wrong, I missed that `readMessageGroup` emit when `readMessages()` returns 0. [This](https://github.com/vimeo/psalm/pull/1289/files#diff-7ee09b3bf72d4706655e1fcc57b4a321R47) should still be checking the return and emitting that message. The while loop should also check `is_accepting_new_requests`. If you can give me push access to your branch I'd be happy to fix it.

muglug:
I don't know how to enable push access to a single branch - would you mind PRing against it?

trowski:
I think it can be done with branch restrictions, but it's probably easier if I just PR. What commands are you running so I can check my work?

muglug:
It's not easy to replicate unless you're [running Psalm as a language server](https://getpsalm.org/docs/language_server/)

PR #1291, opened 2019-02-06
Update ProtocolStreamReader

trowski:
Related to #1212 and #1289.

@muglug I installed the LSP plugin in PhpStorm and configured it to use `./psalm-language-server` when editing this project. Seems to work… I think. Let me know if more changes are needed here.


muglug:
Awesome, thanks! Tested locally and it works.

PR #1292, opened 2019-02-06
Fix/windows tests

SignpostMarv:
some of the tests fail under windows, but not exactly consistently;

* some tests that reference `src/someFile.php` don't trigger an error
* some tests fail with PHP_EOL, some tests fail if not using PHP_EOL


muglug:
Thanks!

SignpostMarv:
yw :)

i'm a little confused that not all of the instances of `src/someFile.php` needed amending o_O

Issue #1191, opened 2019-01-09
Add --php-version to psalm?


TysonAndre:
A common type of bug would be using features from newer PHP versions when the project needs to support older versions.

Obviously there's other ways to detect this, such as providing a plugin to run `php --no-php-ini --syntax-check < file_or_tmpfile.php`, or doing that with a separate tool

A reasonable default would be the php version used to run psalm (or guess from composer.json). For backwards compatibility, I'd assume 7.3 would be the most similar

Examples:

- Nullable types/void in PHP 7.1, object in php 7.2, etc 
- Psalm does not warn about extra trailing commas in function calls (e.g. if running with PHP 7.1). Those became valid in PHP 7.3. (I'd assume PHP-Parser provides this)

```
<?php
call_user_func(/** @return void */function () {
    $other = 'x';
    var_export(is_string(
        $other,  // Psalm doesn't warn about trailing commas
    ));
    var_export(is_string($missing));  // Psalm correctly warns about this
});
```

PR #1295, opened 2019-02-07
HashContext is only implemented in `php:>=7.2`

SignpostMarv:
hash_init() is detected as returning `resource|HashContext`, but `hash_update()` is currently detected as only accepting `HashContext`.

* https://getpsalm.org/r/a27717736e
* using 69d4407

```php
<?php
$hash = hash_init('sha256');

hash_update($hash, 'foo');
```

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreGenericFunctions.php
+++ b/src/Psalm/Internal/Stubs/CoreGenericFunctions.php
@@ -154,3 +154,9 @@ function array_change_key_case(array $arr, int $case = CASE_LOWER) {}
  * @return array<int, array<array-key, T>>
  */
 function array_chunk(array $arr, int $size, bool $preserve_keys = false) {}
+
+
+/**
+* @param resource|HashContext
+*/
+function hash_update($hash, string $data) : bool {}


src/Psalm/Internal/Stubs/CoreGenericFunctions.php
```diff
@@ -154,3 +154,9 @@ function array_change_key_case(array $arr, int $case = CASE_LOWER) {}
  * @return array<int, array<array-key, T>>
  */
 function array_chunk(array $arr, int $size, bool $preserve_keys = false) {}
+
+
+/**
+* @param resource|HashContext
+*/
+function hash_update($hash, string $data) : bool {}
```
weirdan: I think this should go to `src/Psalm/Internal/CallMap.php` instead. `CoreGenericFunctions` implies generics, i.e. functions with `@template` tags.

SignpostMarv: I can get that done properly later tonight- just did it direct on github because it cropped up during the day job.

Does the CallMap allow per-php-version typing btw?

weirdan: It's possible upstream (in phan/PHPSignatures), but not in Psalm as far as I'm aware. Would require at least #1191.

muglug:
Thanks, I'm gonna solve this by solving #1191

SignpostMarv:
no callmap patch needed then?

muglug:
Lots of callmap patches: https://github.com/vimeo/psalm/commit/17541b415fb164cee528f76092beb6ff47c3a45a

PR #1305, opened 2019-02-09
Show total fixed errors when updating baseline

maludecks:
This PR adds a "total errors fixed" in the output report, every time Psalm runs with the flag --update-baseline, so then it's possible to check the number of errors that were removed from the codebase, comparing the current baseline against the updated one. 

We use Psalm in one of our projects at work, and we are having sessions every two weeks in order to reduce and eventually, fix **all** the errors Psalm found in our codebase, for this reason, it would be great (and motivating) to easily see the amount of errors removed by our changes, whenever we update our baseline. 

Feedback is very welcome :)


muglug:
Thanks, this is great!

PR #1308, opened 2019-02-09
Enable PHP 7.4snapshot on travis

weirdan:
Now that nightly is 8.x, Psalm can continue to be tested with 7.4 branch.


muglug:
Thanks!

PR #1323, opened 2019-02-11
docs index tweaks

SignpostMarv:
There's an additional tweak I didn't want to make on the web interface that possibly merits a little discussion beforehand;

* Templates are a sufficiently useful topic that they'd deserve their own doc file to make them easier to find.
* Are supported annotation types a sub-topic of the topic "Typing in psalm"?
* If there isn't already one, should there be a script to auto-generate the [Builtin templated classes and interfaces](https://github.com/vimeo/psalm/blob/master/docs/supported_annotations.md#builtin-templated-classes-and-interfaces) section?

re:sub-topic:
```markdown
- [Typing in Psalm](typing_in_psalm.md)
  - [Supported Annotations](supported_annotations.md)
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -5,8 +5,9 @@
 - [Introduction](#introduction)
 - [Installation](installation.md)
 - [Configuration](configuration.md)
-- [Running Psalm via command line](running_psalm.md)
-- [Running Psalm in your IDE](language_server.md)
+- Using Psalm
+  - [Running Psalm via command line](running_psalm.md)
+  - [Running Psalm in your IDE](language_server.md)
 - [Using Psalter](fixing_code.md)
 - [Supported Annotations](supported_annotations.md)
 - [Dealing with code issues](dealing_with_code_issues.md)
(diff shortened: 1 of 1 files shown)


muglug:
> Templates are a sufficiently useful topic that they'd deserve their own doc file to make them easier to find.

Agreed - would you mind PRing that?

> Are supported annotation types a sub-topic of the topic "Typing in psalm"?

I suppose, insofar as most everything Psalm does is type-related

> should there be a script to auto-generate the [Builtin templated classes and interfaces](https://github.com/vimeo/psalm/blob/master/docs/supported_annotations.md#builtin-templated-classes-and-interfaces) section?

There isn't one, but there definitely could be.

SignpostMarv:
I was going to have a look at doing the auto-generation script myself, but I'm not sure what the best approach is- is there a public/protected method for pulling out stubbed annotations from psalm before & after plugins have ran (i.e. to flag out builtin stubs separately from plugin stubs)?

PR #1324, opened 2019-02-11
Docs/tweaks

SignpostMarv:
as discussed in #1323, with a few additional maintenance tweaks.


docs/templated_annotations.md
```diff
@@ -0,0 +1,217 @@
...
+ * @template T
+ * @psalm-param T $t
+ * @return T
+ */
+function mirror($t) {
+    return $t;
+}
+
+$a = 5;
+$b = mirror(5); // Psalm knows the result is an int
```
weirdan:
```suggestion
$b = mirror(5); // Psalm knows the result is an int

$c = "foo";
$d = mirror($c); // Psalm knows the result is string
```

SignpostMarv: can't quite tell from github's preview- is the suggestion to remove line 21?

weirdan: no, to add couple of more lines, showing the method return type being polymorphic based on argument type.

SignpostMarv: ah :D

PR #1327, opened 2019-02-12
Fix XML node name in the docs

jdreesen:


muglug:
Thanks! That was an _old_ name

PR #1341, opened 2019-02-17
Fix password_hash() return type

bugreportuser:
Revert #1189.

password_hash() returns null on failure: https://bugs.php.net/bug.php?id=77218


muglug:
Thanks!

weirdan:
@muglug @bugreportuser for 7.3 and below `password_hash` certainly returns *either* false or null, see  https://github.com/php/php-src/blob/PHP-7.3/ext/standard/password.c#L472-L492

bugreportuser:
@weirdan You are absolutely correct. I originally found that bug before the discussion of the regression went on. I'll open a new pull request that should fix that.

PR #1345, opened 2019-02-17
Fix mb return types

bugreportuser:
The mb function return types are not well documented. A lot of them return `false` for an unknown encoding.


muglug:
Thanks!

PR #1346, opened 2019-02-17
Fix invalid return type in callmap.

ADmad:
apu_delete() only returns boolean.
http://php.net/manual/en/function.apcu-delete.php


muglug:
Originally added in https://github.com/phan/phan/commit/9c51ecb199014563d11df2e8e03b42bb3dc9f905#diff-51d74031f390b02c2f3cced18a13d964 - was that a mistake @TysonAndre?

TysonAndre:
That seems to be the case

muglug:
No worries! Thanks @ADmad!

ADmad:
No problem :)

PR #1349, opened 2019-02-18
Fix array_unique stub

### LINKED ISSUE #1347: Incorrect return type for array_unique?
Hopefully I'm not overlooking anything here - it looks like the definition for `array_unique` specifies an input array with mixed keys and a return array with integer keys.

In reality keys should be preserved - see the first example from the docs: https://secure.php.net/manual/en/function.array-unique.php#refsect1-function.array-unique-examples

Basically:

```php
array_unique(['a' => 'green', 'b' => 'green']) == ['a' => 'green']
```

But given an input `array<string, string> $arr` and an output `array_unique($arr)` Psalm is complaining about an invalid return type: https://getpsalm.org/r/e426e77a0c

bugreportuser:
Fixes #1347


muglug:
Thanks!

PR #1352, opened 2019-02-18
Added option to force dead code detection

weirdan:
Previously Psalm would only check class references when there were no
paths passed on the command line. Now it's possible to force this check
even when paths are specified, with `--find-dead-code=always`.

`--find-dead-code` without a value (or with any value apart from `always`)
will continue to work the same way it did before.


muglug:
Thanks!

PR #1358, opened 2019-02-18
Fix define() with a namespace

### LINKED ISSUE #1301: define() with a namespace acts strange
Constants defined with namespaces cause strange errors.

Example (https://getpsalm.org/r/6f2a74c453):
```php
<?php
namespace {
define('ns1\\cons1', 0);

echo \ns1\cons1; // false positive
echo ns1\cons1; // false positive
}
namespace ns2 {
define(__NAMESPACE__.'\\cons2', 0);

echo \ns2\cons2;
echo \ns3\cons2; // false negative
echo cons2;
}
namespace ns2 {
echo \ns2\cons2; // false positive
echo cons2; // false positive
}
namespace {
echo \ns2\cons2; // false positive
echo ns2\cons2; // false positive
}
```

Example output:
```
Psalm output (using commit 88b4f82):

ERROR: UndefinedConstant - 5:6 - Const ns1\cons1 is not defined

ERROR: UndefinedConstant - 6:6 - Const ns1\cons1 is not defined

ERROR: UndefinedConstant - 16:6 - Const ns2\cons2 is not defined

ERROR: UndefinedConstant - 17:6 - Const cons2 is not defined

ERROR: UndefinedConstant - 20:6 - Const ns2\cons2 is not defined

ERROR: UndefinedConstant - 21:6 - Const ns2\cons2 is not defined
```

bugreportuser:
Fixes #1301

The namespace is ignored for now. Psalm can't find the constant when it has a namespace. I'll open another issue for that.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -607,10 +607,11 @@ class FunctionCallAnalyzer extends \Psalm\Internal\Analyzer\Statements\Expressio
                     return false;
                 }
             } elseif ($function->parts === ['define']) {
-                if ($first_arg && $first_arg->value instanceof PhpParser\Node\Scalar\String_) {
+                $const_name = StatementsAnalyzer::getConstName($first_arg->value);
+
+                if ($const_name !== null) {
                     $second_arg = $stmt->args[1];
                     ExpressionAnalyzer::analyze($statements_analyzer, $second_arg->value, $context);
-                    $const_name = $first_arg->value->value;
 
                     $statements_analyzer->setConstType(
                         $const_name,
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -607,17 +607,21 @@ class FunctionCallAnalyzer extends \Psalm\Internal\Analyzer\Statements\Expressio
                     return false;
                 }
             } elseif ($function->parts === ['define']) {
-                $const_name = StatementsAnalyzer::getConstName($first_arg->value);
+                if ($first_arg) {
+                    $const_name = StatementsAnalyzer::getConstName($first_arg->value);
 
-                if ($const_name !== null) {
-                    $second_arg = $stmt->args[1];
-                    ExpressionAnalyzer::analyze($statements_analyzer, $second_arg->value, $context);
+                    if ($const_name !== null) {
+                        $second_arg = $stmt->args[1];
+                        ExpressionAnalyzer::analyze($statements_analyzer, $second_arg->value, $context);
 
-                    $statements_analyzer->setConstType(
-                        $const_name,
-                        isset($second_arg->value->inferredType) ? $second_arg->value->inferredType : Type::getMixed(),
-                        $context
-                    );
+                        $statements_analyzer->setConstType(
+                            $const_name,
+                            isset($second_arg->value->inferredType) ?
+                                $second_arg->value->inferredType :
+                                Type::getMixed(),
+                            $context
+                        );
+                    }
                 } else {
                     $context->check_consts = false;
                 }
--- a/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
@@ -147,12 +147,27 @@ class ExpressionAnalyzer
                     $stmt->inferredType = Type::getLiteralClassString($context->self);
                     break;
 
+                case '__namespace__':
+                    $namespace = $statements_analyzer->getNamespace();
+                    if ($namespace === null &&
+                    IssueBuffer::accepts(
+                        new UndefinedConstant(
+                            'Cannot get __namespace__ outside a namespace',
+                            new CodeLocation($statements_analyzer->getSource(), $stmt)
+                        ),
+                        $statements_analyzer->getSuppressedIssues()
+                    )) {
+                        // fall through
+                    }
+
+                    $stmt->inferredType = Type::getString($namespace);
+                    break;
+
                 case '__file__':
                 case '__dir__':
                 case '__function__':
                 case '__trait__':
                 case '__method__':
-                case '__namespace__':
                     $stmt->inferredType = Type::getString();
                     break;
             }
--- a/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
@@ -145,14 +145,15 @@ class StatementsAnalyzer extends SourceAnalyzer implements StatementsSource
                     && isset($stmt->expr->args[1])
                     && $stmt->expr->args[0]->value instanceof PhpParser\Node\Scalar\String_
                 ) {
-                    $const_name = $stmt->expr->args[0]->value->value;
-
-                    $this->setConstType(
-                        $const_name,
-                        self::getSimpleType($codebase, $stmt->expr->args[1]->value, $this->getAliases(), $this)
-                            ?: Type::getMixed(),
-                        $context
-                    );
+                    $const_name = static::getConstName($stmt->expr->args[0]->value);
+                    if ($const_name !== null) {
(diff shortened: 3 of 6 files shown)


src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
```diff
@@ -1464,4 +1465,36 @@ public function getFunctionAnalyzers()
...
+     */
+    public static function getConstName($first_arg_value)
+    {
+        $const_name = null;
+
+        if ($first_arg_value instanceof PhpParser\Node\Scalar\String_) {
+            $const_name = $first_arg_value->value;
+        } elseif (isset($first_arg_value->inferredType)) {
+            $possible_value_types = $first_arg_value->inferredType->getTypes();
+            if (count($possible_value_types) === 1) {
```
muglug: There are convenience functions added by @TysonAndre  - `Psalm\Type\Union::hasSingleStringLiteral` && `Psalm\Type\Union::getSingleStringLiteral` that'll do this job for you

src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
```diff
@@ -145,14 +145,15 @@ public function analyze(
                     && isset($stmt->expr->args[1])
                     && $stmt->expr->args[0]->value instanceof PhpParser\Node\Scalar\String_
```
muglug: this condition can now be removed I think? If we're using the string value below (though this is rarely used, so it won't make much of a difference)

src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
```diff
@@ -147,12 +147,27 @@ public static function analyze(
                     $stmt->inferredType = Type::getLiteralClassString($context->self);
                     break;
 
+                case '__namespace__':
+                    $namespace = $statements_analyzer->getNamespace();
+                    if ($namespace === null &&
+                    IssueBuffer::accepts(
+                        new UndefinedConstant(
```
TysonAndre: You can't be outside of a namespace, you're just in the global namespace.

Silently change this from null to the empty string, instead?

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
```diff
@@ -607,16 +607,21 @@ function (Assertion $assertion) use ($generic_params) : Assertion {
...
-                    $second_arg = $stmt->args[1];
-                    ExpressionAnalyzer::analyze($statements_analyzer, $second_arg->value, $context);
-                    $const_name = $first_arg->value->value;
-
-                    $statements_analyzer->setConstType(
-                        $const_name,
-                        isset($second_arg->value->inferredType) ? $second_arg->value->inferredType : Type::getMixed(),
-                        $context
-                    );
+                if ($first_arg) {
```
TysonAndre: NYC, not a new issue: One issue reported in Phan was that if there were multiple `define()` calls for a given constant (e.g. at src/a.php:6 and src/a.php:10), then the value the psalm infers would vary based on what file was being analyzed and the phase of analysis. I'm not sure if psalm would have a similar issue.

Another issue reported with Phan was dynamic constants, e.g. `$x = CONST_FROM_OTHER_FILE . 'suffix'; define($x, 'value');`. In that case, the constant name could only be resolved once all other files were parsed.

- those issues were resolved by allowing the first constant definition to be overridden only by a value from the same node (i.e. same file + line (+ column)).

src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
```diff
@@ -1464,4 +1465,36 @@ public function getFunctionAnalyzers()
...
+                reset($possible_value_types);
+                $possible_value_type = current($possible_value_types);
+                if ($possible_value_type instanceof Type\Atomic\TLiteralString) {
+                    $const_name = $possible_value_type->value;
+                }
+            }
+        }
+
+        if ($const_name !== null) {
+            $namespace_pos = strrpos($const_name, '\\');
```
TysonAndre: Why is the namespace being removed? This looks like this would attempt to add `SomeConst` instead of `Ns\SomeConst` for `define('Ns\SomeConst', 'value')` (from any namespace)

src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
```diff
@@ -147,12 +147,27 @@ public static function analyze(
...
+                        ),
+                        $statements_analyzer->getSuppressedIssues()
+                    )) {
+                        // fall through
+                    }
+
+                    $stmt->inferredType = Type::getString($namespace);
+                    break;
+
                 case '__file__':
```
TysonAndre: Aside, out of the scope of this PR: All of this code was probably written before Psalm added support for literal string/int types.

It's useful to have the literal values for these constants:

- __FUNCTION__ for when validating that a callback is a valid callback
- __DIR__/__FILE__ for various checks, e.g. checking that calls to require_once with complex expressions are valid PHP files (absolute path is the most useful)
- __TRAIT__ for the same reason as __CLASS__, for checking for valid class-likes (make sure to remove leading the `\\` character)

src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
TysonAndre: Adding `setFullyQualifiedConstType($fully_qualified_const_name, Type\Union $const_type)` would probably be the simplest fix in src/Psalm/Internal/Analyzer/NamespaceAnalyzer.php and StatementsAnalyzer.php (or rethink $context->constants)
 
```
    /**
     * @param string     $const_name
     * @param Type\Union $const_type
     *
     * @return void
     */
    public function setConstType($const_name, Type\Union $const_type)
    {
        self::$public_namespace_constants[$this->namespace_name][$const_name] = $const_type;
    }
```

src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
bugreportuser: I thought about this. The warning could be clearer, sure, but `__CLASS__` is also an empty string outside of classes and has a warning. I figured that the same reasoning applies.

The warning is helpful with `define()`. Try defining a constant with `define(__NAMESPACE__.'\\cons')` in the global namespace. There's no way to access it.

src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
bugreportuser: Hmm. I didn't realize that two of the files call that function, meaning they do use a namespace already. Wouldn't it be the correct namespace, meaning that `setFullyQualifiedConstType()` is unnecessary? The only file that doesn't use that method is src/Psalm/Internal/Visitor/ReflectorVisitor.php, which is used the most.

I'm not sure why, but `define()` always ignores namespaces from user code, so this is nothing new. I do agree that this should be fixed, but I'm not sure if it makes sense for it to happen in this PR. This fixes a lot of issues, and I'm pretty sure that fixing `define()` to consider namespaces would be more involved.

src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
bugreportuser: I agree, and yes - best left out of scope for this PR.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
bugreportuser: > One issue reported in Phan was that if there were multiple `define()` calls for a given constant (e.g. at src/a.php:6 and src/a.php:10), then the value the psalm infers would vary based on what file was being analyzed and the phase of analysis. I'm not sure if psalm would have a similar issue.

It does, including constants defined with `const`.

src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
bugreportuser: Good to know. Now fixed.

src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
bugreportuser: You're correct. I missed this one.

muglug:
Thanks!

PR #1369, opened 2019-02-20
Expose more internal methods through Codebase

weirdan:
- `bool Codebase::canTypeBeContainedByType(Union $input, Union $container)`
- `array{Union,Union} Codebase::getKeyValueParamsForTraversableObject(Atomic $type)`

For usage, see psalm/phpunit-psalm-plugin#15

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1179,7 +1179,7 @@ class Codebase
      * If you consider `Type\Union` as a set of types, this will tell you if intersection
      * of `$input_type` with `$container_type` is not empty.
      *
-     * $input_type ∩ $container_type ≠ ∅ , e.g. they are not dijoint.
+     * $input_type ∩ $container_type ≠ ∅ , e.g. they are not disjoint.
      *
      * Useful for emitting issues like PossiblyInvalidArgument, where argument at the call
      * site should be a subtype of the function parameter type, but it's has some types that are
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1162,8 +1162,6 @@ class Codebase
      *
      * Useful for emitting issues like InvalidArgument, where argument at the call site
      * should be a subset of the function parameter type.
-     *
-     * @psalm-suppress PossiblyUnusedMethod
      */
     public function isTypeContainedByType(
         Type\Union $input_type,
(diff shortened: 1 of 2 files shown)


src/Psalm/Codebase.php
```diff
@@ -1159,4 +1160,31 @@ public function isTypeContainedByType(
...
+        Type\Union $container_type
+    ): bool {
+        return TypeAnalyzer::canBeContainedBy($this, $input_type, $container_type);
+    }
+
+    /**
+     * @return array{Type\Union,Type\Union}
+     * @psalm-suppress PossiblyUnusedMethod
+     */
+    public function getKeyValueParamsForTraversableObject(Type\Atomic $type): array
```
muglug: Would you mind add a test or two for this? That’d save needing a PossiblyUnusedMethod annotation.

src/Psalm/Codebase.php
```diff
@@ -1159,4 +1160,31 @@ public function isTypeContainedByType(
     ): bool {
         return TypeAnalyzer::isContainedBy($this, $input_type, $container_type);
     }
+
+    /**
+     * @psalm-suppress PossiblyUnusedMethod
+     */
+    public function canTypeBeContainedByType(
```
muglug: Maybe add a test for this and the method above, and add comments to explain to other potential plugin authors how they’re different (I realise I’m sort of asking you to do my work for me, but since you’re here...)

weirdan: Sure.

src/Psalm/Codebase.php
weirdan: Would you suggest a specific test file?

muglug: I'd actually encourage you to create a new file, `CodebaseTest` for these specific things

src/Psalm/Codebase.php
weirdan: Added docs.

src/Psalm/Codebase.php
weirdan: Added. I'm getting errors trying to check any classes though, so I resorted to testing with built-in types only.

weirdan: Errors like 'cannot get classstorage for exception'

weirdan:
@muglug this is ready to merge.

muglug:
Thanks!

PR #1380, opened 2019-02-23
Allow @psalm-property and @psalm-method

bugreportuser:
Allow `@psalm-property` to replace `@property` and `@psalm-method` to replace `@method`.


weirdan:
Does `@psalm-method` override `@method` when used with the same method name? Can you add a test clarifying this case?

bugreportuser:
Yes and the same for `@psalm-property`. Good idea on the tests.

muglug:
Thanks!

PR #1385, opened 2019-02-24
Added custom storage metadata

weirdan:
This will allow plugins to store their data associated with storage nodes (including persisting this data to cache, if called in appropriate phase).

### DIFF
--- a/src/Psalm/Storage/ClassLikeStorage.php
+++ b/src/Psalm/Storage/ClassLikeStorage.php
@@ -327,6 +327,9 @@ class ClassLikeStorage
      */
     public $has_docblock_issues = false;
 
+    /** @var array<string,mixed> */
+    public $custom_metadata = [];
+
     /**
      * @param string $name
      */
--- a/src/Psalm/Storage/ClassLikeStorage.php
+++ b/src/Psalm/Storage/ClassLikeStorage.php
@@ -6,6 +6,8 @@ use Psalm\Type;
 
 class ClassLikeStorage
 {
+    use CustomMetadataTrait;
+
     /**
      * A lookup table for public class constants
      *
@@ -326,9 +328,6 @@ class ClassLikeStorage
      * @var bool
      */
     public $has_docblock_issues = false;
-
-    /** @var array<string,mixed> */
-    public $custom_metadata = [];
 
     /**
      * @param string $name
--- a/src/Psalm/Storage/CustomMetadataTrait.php
+++ b/src/Psalm/Storage/CustomMetadataTrait.php
@@ -1,0 +1,11 @@
+<?php
+namespace Psalm\Storage;
+
+/**
+ * @psalm-type _MetadataEntry scalar|scalar[]|scalar[][]|scalar[][][]|scalar[][][][]|scalar[][][][][]
+ */
+trait CustomMetadataTrait
+{
+    /** @var array<string,_MetadataEntry> */
+    public $custom_metadata = [];
+}
--- a/src/Psalm/Storage/FileStorage.php
+++ b/src/Psalm/Storage/FileStorage.php
@@ -3,6 +3,8 @@ namespace Psalm\Storage;
 
 class FileStorage
 {
+    use CustomMetadataTrait;
+
     /**
      * @var array<string, string>
      */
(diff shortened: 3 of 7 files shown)


muglug:
What sort of information were you planning to store?

weirdan:
Something like this: https://github.com/psalm/phpunit-psalm-plugin/pull/17/files#diff-be87d043420df8d1976f3213fd562c0dR71

Doctrine plugin may store data from annotations and config. Pretty much anything, as long as it's small, serializable, and depends on the source code of the element.

muglug:
What if you could instead register a hook that forced any `ClassLikeStorage` object relating to some class `Foo::class` to be an instance of `FooClassLikeStorage`? That'd open the door to implementing more custom behaviour (e.g. changing the results of `$codebase->methodExists` for a given class).

Here a custom `TestCaseClassLikeStorage` class would have store the necessary data as appropriate.

weirdan:
Wouldn't this cause issues if `TestCaseClassLikeStorage` was no longer available? Like, after plugin is uninstalled?

weirdan:
Also, how would that work when two plugins have their own idea on what the class name should be? Would each of them get their own classes?

muglug:
Well that’s also a problem here - two plugins could use the same key names. But if the whole class is different the error can be bubbled up to the user.

muglug:
If the class is no longer available I think deserialisation just fails, and the data is re-serialised.

weirdan:
> Well that’s also a problem here - two plugins could use the same key names.

While they *could* use the same name, they don't have to (it's a question of namespacing, basically, well understood in the industry). But with a global classname override they *will* step on each others toes.

weirdan:
> If the class is no longer available I think deserialisation just fails

For straightforward `unserialize()` you'll get `__PHP_Incomplete_Class` objects. It won't throw or return any other kind of failure indicator. You can define callback to handle missing classes (with [ini setting](http://php.net/manual/en/var.configuration.php#ini.unserialize-callback-func)), but you probably can't throw there either (PHP's serialization mechanism is very sensitive to exceptions).

muglug:
But the UI can step in there - “Found an error when reflecting class Foo - Plugin X is not compatible with Plugin Y, as they’re both trying to manipulate the same class’s storage - please disable one of the plugins”

muglug:
Maybe my idea is bad, though

weirdan:
As a plugin author, I'd be wary to use exclusive mechanisms, as it's likely to cause conflicts.

Separate storage for plugin data might be an option, but it looks more complicated both from the Psalm side (quite a bit of code to be written) and from plugin side as well (if you want to tie entries' lifecycle to code elements changes).

muglug:
Ok, nevermind - but would it be ok to restrict values to `scalar|array<scalar>`?

weirdan:
It's a bit too restrictive. I can easily see how someone may need to store multiple annotations, each having their own attributes. Taking doctrine as example (https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/basic-mapping.html#line-number-74f96c9d606323ff59c4f57374e70618413c5af9-7) this may be conveniently stored as 
```php
$class_storage->custom_metadata['my\\plugin']['SequenceGenerator'] = 
    ['sequenceName' => "message_seq", 'initialValue' => 1, 'allocationSize' => 100]
```
Of course one can always put json or serialized string there, but it feels a little bit ... wrong.

muglug:
Happy for it to be `scalar|array<scalar>|array<array<scalar>>` etc.. I just don’t want to mess up deserialisation with objects

weirdan:
> `scalar|array<scalar>|array<array<scalar>>`

Recursive type would be handy here :)
Ok, I'll adjust docblocks a bit later. Probably will define it like 4-5 levels deep arrays, which should be plenty enough.

weirdan:
@muglug done.

muglug:
Thanks! I think the way forward on the other score is to support a bunch of smaller hooks to supplement `FunctionReturnTypeProviderInterface` and `MethodReturnTypeProviderInterface`: 

  - `MethodExistsProviderInterface`
  - `MethodVisibilityProviderInterface`
  - `MethodParamsProviderInterface`
  - `PropertyExistsProviderInterface`
  - `PropertyVisibilityProviderInterface`
  - `PropertyTypeProviderInterface`
  - `FunctionExistsProviderInterface`
  - `FunctionParamsProviderInterface`

weirdan:
Both fine and coarse grained hooks are useful. Speaking of `*ReturnTypeProvider`s, they seem to miss a way to register them from plugins. I would suggest to add another method in `Psalm\Plugin\RegistrationInterface`

PR #1388, opened 2019-02-24
Added Hook\AfterCodebasePopulatedInterface

weirdan:
It could be used to perform operations after files have been scanned, but before main analysis is started.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -306,6 +306,12 @@ class Config
      */
     public $after_visit_classlikes = [];
 
+    /**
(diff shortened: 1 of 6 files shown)


muglug:
Mind adding a test or two?

weirdan:
Oh, it's missing the registration part. Gonna add this in a moment.

muglug:
Looks like a problem with the 7.4 snapshot and PHPUnit :/

weirdan:
Yeah, I've seen bugs reported about anonymous classes causing crashes in 7.4

PR #1404, opened 2019-02-28
fixing some tests under windows again (following on from #1292)

SignpostMarv:
any chance of windows ci being set up ?


muglug:
Happy to have one, if you can finagle the details

PR #1415, opened 2019-03-01
Use PHPUnit plugin

weirdan:
- Added PHPUnit plugin
- Dropped suppressions that have become obsolete
- Dropped unused method (`ProjectAnalyzerTest::getAllIssues()`)
- Fixed inconsistencies (mostly typos and key omissions in data providers)


muglug:
Thanks! Slightly annoyed that Psalm doesn’t yet have a bulletproof way to annotate ordered array keys - will think on that.

PR #1420, opened 2019-03-02
Increment CallMap minor version

bugreportuser:
A callmap exists for version 7.4, so it should be used.

### DIFF
--- a/src/Psalm/Internal/Codebase/CallMap.php
+++ b/src/Psalm/Internal/Codebase/CallMap.php
@@ -13,7 +13,7 @@ use Psalm\Storage\FunctionLikeParameter;
 class CallMap
 {
     const PHP_MAJOR_VERSION = 7;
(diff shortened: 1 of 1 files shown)


muglug:
I'm going to hold out on this, as PHP 7.4 is not a supported version at the moment

bugreportuser:
This could use some explanation. The reason for this request is that the return type for `password_hash` is wrong. The call map has the newest version from #1341. Psalm thinks that it returns `null`, but that's the 7.4 return type. This change fixes it for lower versions. This would have made more sense to include in #1342, but I didn't know it was required. Would it be better to revert those requests?

muglug:
I've fixed the `password_hash` return type in 8d59ab7 - happy to keep the 7.4 callmap diff around, in expectation of its eventual use

bugreportuser:
Looks good.

Issue #54, opened 2017-01-14
Constants inside functions are not recognised


svivian:
Test case:

```php
/**
 * @return void
 */
function defineConstant()
{
    define('CONSTANT', 1);
}

defineConstant();

echo CONSTANT;
```
Output:

> ERROR: UndefinedConstant - test-constant.php:13 - Const CONSTANT is not defined
> echo CONSTANT;

Constants are always global, so there is no error in the above script.

svivian:
Also fails in the opposite example:


	define('TEST', 2);

	/**
	 * @return int
	 */
	function useTest()
	{
		return TEST;
	}

Output:

> ERROR: UndefinedConstant - test-constant.php:30 - Const TEST is not defined
> return TEST;

muglug:
Thanks for this – non-trivial to fix the initial version, as it means storing the constants declared by a function. Even then, there'll be a little fudging – I think it should still fail in this scenario:

```php
/**
 * @return void
 */
function defineConstant()
{
    if (rand(0, 1)) {
        define('CONSTANT', 1);
    }
}

defineConstant();

echo CONSTANT;
```

i.e. when the constant is not defined on the function in the main block.

muglug:
In both cases I think the solution is to tend towards false negatives.

muglug:
The latter is now fixed

muglug:
Thanks for detailing these

PR #1423, opened 2019-03-03
Fix more undefined constant errors

bugreportuser:
Test `constantDefinedInFunctionButNotCalled` had to be removed, but it only worked before by the constant not being defined in the namespace. Since the namespace issue was fixed, that test no longer passes.

Example: https://psalm.dev/r/9f56fbea8c

Maybe #54 should be re-opened? Or defining constants in functions that aren't ever called could be unsupported? Either is fine with me.

### DIFF
--- a/src/Psalm/Internal/Analyzer/NamespaceAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/NamespaceAnalyzer.php
@@ -130,9 +130,13 @@ class NamespaceAnalyzer extends SourceAnalyzer implements StatementsSource
      *
      * @return void
      */
-    public function setConstType($const_name, Type\Union $const_type)
+    public static function setConstType($const_name, Type\Union $const_type)
     {
-        self::$public_namespace_constants[$this->namespace_name][$const_name] = $const_type;
+        $const_name_parts = explode('\\', $const_name);
+        $const_name = array_pop($const_name_parts);
+        $namespace_name = implode('\\', $const_name_parts);
+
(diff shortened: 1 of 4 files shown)


muglug:
Fixed this in fb9e9be

bugreportuser:
Should've also added this test:
```php
<?php
namespace ns;
function func(): void {}
define(__NAMESPACE__.'\\cons', 0);

namespace ns;
cons;
```

And another good one:
```php
<?php
namespace ns1;
function func(): void {}
define('ns3\\cons', 0);

namespace ns2;
\ns3\cons;
```

muglug:
The fix for both is just to put the `define` above the function definition

bugreportuser:
That's true but weird. What about something like this?
```php
<?php
namespace ns;
function func(): void {}
define(__NAMESPACE__.'\\cons', func());

namespace ns;
cons;
```

PHP does hoist functions, so it doesn't matter for anything other than code style, but I think it could make sense to support. I would think defining constants in a function is less common, but maybe that's just me.

muglug:
Yeah, but it happens a bunch in WordPress's codebase.

bugreportuser:
I guess you're right about needing to support it then. If I create an issue for this, are you willing to keeping it open until something that works for both can be created?

muglug:
Absolutely - is it something you encountered in real-world code?

bugreportuser:
Only private projects but yes - the examples are all simplified from actual code. I'm now applying the patch to my local clone, so it can wait until there's a better solution.

I'm surprised these `define` issues haven't come up before.

PR #1425, opened 2019-03-04
fix --help end of line

ro0NL:
Given `psalm -h`

```
    --alter
        Run Psalter/app $ 
```

but should be

```
    --alter
        Run Psalter
/app $ 
```


muglug:
Thanks!

Issue #1111, opened 2018-11-30
Annotation is not recognized on a global declaration


Rarst:
This annotation works fine:

```php
global $global;

/** @var array<string, array> $global */
$filter = $global['name'];
```
But this one doesn’t:
```php
/** @var array<string, array> $global */
global $global;

$filter = $global['name'];
```
```
INFO: MixedArrayAccess - 6:11 - Cannot access array value on mixed variable $global
INFO: MixedAssignment - 6:1 - Cannot assign $filter to a mixed type
```
https://getpsalm.org/r/90ba05e725

Issue #1087, opened 2018-11-21
Type of `$_GET`


vasily-kartashov:
Is there a way to specify the type of `$_GET` superglobal (or alternatively the result's type of `parse_str`)?

```
0=                                                 array<int,string>                                 
&&&&a=example                                      array<string,string>                              
x=&y[]=2&y[xxx]=null&0=false                       array<int|string,array<int|string,string>|string> 
x=&y[]=2&y[xxx]=null&0=false&[1]=23                array<int|string,array<int|string,string>|string> 
x=&y[][]=2&y[][1]=null&y[][][]=0&false=-1          array<string,array<int,array<int,array<int,string>>|array<int,string>>|string>
5=6                                                array<int,string> 
```

With recursive types it would be something like

```
@var V=array<int|string,string|V>
```

weirdan:
That might be quite tricky since $_GET is not read-only: https://3v4l.org/qLFMP . And it's superglobal nature further complicates things, since it could be assigned anywhere.

TysonAndre:
It's implemented in src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php

```php
        if (in_array(
            $stmt->name,
            [
                'GLOBALS',
                '_SERVER',
                '_GET',
                '_POST',
                '_FILES',
                '_COOKIE',
                '_SESSION',
                '_REQUEST',
                '_ENV',
            ],
            true
        )
        ) {
            $stmt->inferredType = Type::getArray();
            $context->vars_in_scope['$' . $stmt->name] = Type::getArray();
            $context->vars_possibly_in_scope['$' . $stmt->name] = true;

            return null;
        }
```

It's easy to do, the main issue is deciding what you want to do. E.g. `0=something` is going to be pretty rare in most web apps. (`array<string,array|string>` might work)

My thoughts would be to use the following union types, or something similar (will have false positives in the very rare cases when you use integers as keys):

I don't think psalm supports recursive union types, or that there are plans to

```php
[
        '_GET' => 'array<string,string|string[]>',  // or array<string,string|array>, if you want to support arbitrary nesting - same for the others
        '_POST' => 'array<string,string|string[]>',
        '_COOKIE' => 'array<string,string|string[]>',
        '_REQUEST' => 'array<string,string|string[]>',
        '_SERVER' => 'array<string,mixed>',
        '_ENV' => 'array<string,string>',
[...]

Issue #951, opened 2018-08-20
Recognize globals from autoloader or allow defining them in the configuration


RinkAttendant6:
Psalm doesn't recognize variables defined in the global scope throughout the entire analysis. Ideally it should  recognize global variables in the file referenced by the `autoloader` property (or have an option to do so).

Alternatively, an option like ESLint's [`globals`](https://eslint.org/docs/user-guide/configuring#specifying-globals) configuration property would suffice (though I'd suggest a hashmap of var name to FQCN instead of var name to boolean, if this is the desired implementation).

muglug:
Fixed by https://github.com/vimeo/psalm/pull/1430

RinkAttendant6:
Nice, I guess the documentation hasn't been updated yet

PR #1430, opened 2019-03-05
Allow setting globals in config

bugreportuser:
Fixes these issues:
- #951
- #1087

Please look at line 1176 of ConfigTest.php, which should maybe give a warning. I think it would be difficult to support and not worth it. Should I remove that line to not give a preference?

#1111 is also related.


muglug:
Awesome, thanks!

PR #1440, opened 2019-03-07
Allow capitalized @inheritdoc

bugreportuser:
The PHPDoc standard capitalizes `@inheritdoc` as [`@inheritDoc`](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#41-making-inheritance-explicit-using-the-inheritdoc-tag), which is not supported now.

<s>It also defines [`{@inheritDoc}`](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#42-using-the-inheritdoc-inline-tag-to-augment-a-description) to only inherit the documentation, so it should be ignored.</s>

### DIFF
--- a/src/Psalm/Internal/Analyzer/CommentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/CommentAnalyzer.php
@@ -392,7 +392,7 @@ class CommentAnalyzer
         }
 
         if (strpos(strtolower($comments['description']), '@inheritdoc') !== false
-            || isset($comments['specials']['inheritdoc'])) {
+            || isset($comments['specials']['inheritdoc']) || isset($comments['specials']['inheritDoc'])) {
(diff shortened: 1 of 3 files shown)


weirdan:
> It also defines {@inheritDoc} to only inherit the documentation, so it should be ignored.

Looks like a serious BC break to me.

bugreportuser:
@weirdan Agreed. I'm fine with removing that part or putting it in its own request, but I thought it made sense to include it if I was already looking at the spec.

muglug:
> I'm fine with removing that part or putting it in its own request

Yeah, would you mind doing both? Then I can figure out if that's something that should change, or (as I suspect) if common usage treats them as identical.

bugreportuser:
@muglug Done.

PR #1441, opened 2019-03-07
Flesh out callable arg types

bugreportuser:
`callable(self)` didn't resolve `self` when checking parameters.


bugreportuser:
Thanks for reviewing all these pull requests! I know there are a lot.

PR #1443, opened 2019-03-08
Ignore inline {@inheritDoc}

bugreportuser:
The PHPDoc standard defines [`{@inheritDoc}`](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#42-using-the-inheritdoc-inline-tag-to-augment-a-description) to only inherit the documentation, so it should be ignored.


muglug:
Not going to merge this - inheritance is now automatic

bugreportuser:
I figured when I saw that change. I think that's for the best, but I do find it odd PHPDoc made that the default.

PR #1445, opened 2019-03-09
Fix return value for password_hash().

ADmad:
http://php.net/manual/en/function.password-hash.php


bugreportuser:
This was discussed in #1420 and #1342. The type is correct for now.

PR #1463, opened 2019-03-14
Allow reconciling superglobal types

bugreportuser:
Now that superglobal types can be changed, this expands on 0ea22e8.


muglug:
Thanks!

Issue #1387, opened 2019-02-24
Codebase::classExtends() with unpopulated ClassLikeStorage instance


weirdan:
It's possible to call `Codebase::classExtends()` during `afterClassLikeVisit()`, but it appears it would only work with immediate parent, probably because classlikes have not been crosslinked during this phase yet. Works fine during later stages (like in `AfterClassLikeAnalysisInterface::afterStatementAnalysis()`).

Am I supposed to check ClassLikeStorage::$populated before calling the method? It seems to be false when `classExtends()` produces false negatives. If so, I'd like to propose PR integrating this check into `classExtends()` itself, to save plugin authors a lot of confusion).

muglug:
Yeah, please PR that - it's a good and necessary check

PR #1476, opened 2019-03-16
Codebase::classExtends() now rejects unpopulated classes

weirdan:
This is done to prevent false negatives, when class storage hasn't been populated yet.

Fixes vimeo/psalm#1387.


muglug:
Thanks!

PR #1483, opened 2019-03-20
Ignore unused superglobals

bugreportuser:
Superglobals can be used between functions and by PHP functions, so I don't think it's useful to mark them as unused.


muglug:
Thanks!

PR #1493, opened 2019-03-22
replace ps with posix_kill to determine process is still running and …

### LINKED ISSUE #1485: Fork Pool model uses `ps` to check process is running
Seems like this commit https://github.com/vimeo/psalm/commit/47e987ccf8b1dd227e77a20973a380103946c78c introduced an issue with Alpine-based images of PHP from Docker Hub, as Alpine is using busybox and has no such params for `ps` invocation (https://busybox.net/downloads/BusyBox.html).

Maybe this line:
https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Fork/Pool.php#L268

Can be replaced with sending a `kill -0` signal to PID to check if it's alive?

`$process_lookup = posix_kill($child_pid, 0);`

neanton:
…prevent alpine docker hub image problems

Fixes #1485

### DIFF
--- a/src/Psalm/Internal/Fork/Pool.php
+++ b/src/Psalm/Internal/Fork/Pool.php
@@ -253,19 +253,14 @@ class Pool
      *
      * @return array
      */
-    public function wait()
+    public function wait(): array
     {
-        global $argv;
-
         // Read all the streams from child processes into an array.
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

PR #1501, opened 2019-03-24
Add an error for throws in global scope

bugreportuser:
Adds `checkForThrowsInGlobalScope`, which is similar to `checkForThrowsDocblock`, but it checks in global scope instead of function scope.

I also made the code location for `MissingThrowsDocblock` more accurate. It now shows the line that throws instead of the function that has a throw inside. I can revert that if you want to keep the old location.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -45,6 +45,7 @@
         <xs:attribute name="hoistConstants" type="xs:string" />
         <xs:attribute name="addParamDefaultToDocblockType" type="xs:string" />
         <xs:attribute name="checkForThrowsDocblock" type="xs:string" />
+        <xs:attribute name="checkForThrowsInGlobalScope" type="xs:string" />
         <xs:attribute name="forbidEcho" type="xs:string" />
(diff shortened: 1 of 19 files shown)


weirdan:
> I also made the code location for MissingThrowsDocblock more accurate.

How will this behave when a scope has the same exception thrown at several distinct points?

```php
function f(): void {
  if (rand(0,1)) throw new \Exception;
  else throw new \Exception;
}
```

bugreportuser:
> How will this behave when a scope has the same exception thrown at several distinct points?

Right now, there's not a simple answer. It chooses one, but it's not the first or last with any consistency.

It shouldn't be hard to show both, which is what I assume you're asking about. However, it would mean more memory, since the exceptions would be stored as `array<string, array<int, CodeLocation>>` instead of `array<string, CodeLocation>`.

bugreportuser:
The Travis error is unrelated to this PR.

muglug:
Looks great! Thanks!

PR #1508, opened 2019-03-26
Add support for language server in PHAR

greg0ire:


scoper.inc.php
```diff
@@ -102,18 +102,9 @@ function ($filePath, $prefix, $contents) {
...
-                $prefix . '\Psalm\Plugin\\',
-                'Psalm\Plugin\\',
-                $contents
-            );
-            return $ret;
-        },
     ],
     'whitelist' => [
         \Composer\Autoload\ClassLoader::class,
-        'Psalm\Plugin\*',
```
weirdan: Wouldn't these changes cause problems running plugins with phar-packaged Psalm?

muglug: Probably - I was just seeing errors when running on Psalm’s own codebase, and assumed the errors were universal. I’ll remove tomorrow

muglug:
I added in c1434ee2120c16090d6aaef42f244e6aaa90ca9b, but thanks for your help!

scoper.inc.php
muglug: Do plugins work with the Phar at all?

greg0ire:
No, thanks for _your_ help!

scoper.inc.php
weirdan: Partially. For example, PHPUnit plugin loads, but will probably fail on some versions where it accesses internal classes. Plugins can work with phar, given enough care both from the plugin and Psalm's sides. As long as phar build does not mangle any class names plugin may use it should be working just fine.

muglug: Yeah, but the plugin hooks demand non-namespaced things like `Psalm\StatementsSource`

weirdan: - Not every plugin may have hooks (doctrine plugin is still stubs-only, but that's going to change)
- My original idea was to provide interfaces in `Plugin` namepsace for everything plugins may need to depend on

PR #1517, opened 2019-03-29
Allow suppressing exception issues inline

bugreportuser:
#1501 made exception issues show the line where the exception was thrown, so the next step is to allow suppressing issues on that line.

This also records all code locations, so https://github.com/vimeo/psalm/pull/1501#issuecomment-476003365 gives an error for both exceptions. This PR can work without this, but it would make suppressing errors in a scope annoying, since only one error would be given at a time.

This also makes scopes less confusing. Without this PR, `UncaughtThrowInGlobalScope` is only given once for two different files that throw the same exception and include each other.


src/Psalm/Internal/Stubs/CoreGenericFunctions.php
```diff
@@ -153,3 +153,13 @@ function array_chunk(array $arr, int $size, bool $preserve_keys = false) {}
  * @return bool
  */
 function array_key_exists($key, array $search) : bool { }
+
+/**
+ * @psalm-template TKey as array-key
+ * @psalm-template TValue
+ *
+ * @param array<TKey, TValue> $arr1
```
weirdan:
```suggestion
 * @param array<TKey, TValue> $arr
```

bugreportuser: Nice one. Thanks.

muglug: Yeah, this file is not checked by Psalm currently (because method bodies are missing) but probably should, at least in some guise

muglug:
Thanks!

PR #1500, opened 2019-03-24
Appveyor config

SignpostMarv:
so a few things:
1) ~Only the March 23rd commit 6235a5e is needed (probably) at present.~
2) ~The last 3 a4e1310 7e47f8e and 508297b were only added because of my misunderstanding the composer error output on the first build; `psalm/plugin-phpunit` refuses to install because it's not currently on *dev-master*. 508267b features a workaround, I'm hoping there's a better workaround should it actually become an issue in future~
3) ~I am insufficiently familiar with appveyor to determine what'll happen with regards to pull requests (i.e. how travis etc. can run tests on pull requests).~
4) ~Running the tests on appveyor highlights the recurring windows-specific path issues described in #667, #1292, #1404, and #1467 - it does not fix them; that is to come later as described in #1467 re: "seeing what happens if everything is switched to unix paths"~ seemingly resolved in 877345a, hoping this doesn't break anything.

@muglug @weirdan if you both can take a look at your convenience at the config and the [build output](https://ci.appveyor.com/project/SignpostMarv/psalm/builds/23314606)  to see if we're missing anything useful before I start the experiment with switching to all unix paths?

Issue #1255, opened 2019-01-29
array shapes & array_shift()


SignpostMarv:
Is it within the scope of psalm's purpose to keep track of array modification with funcs like array_shift()?

https://getpsalm.org/r/632db6aee8
```php
<?php
declare(strict_types=1);

/**
* @return array{0:string, 1:int}
*/
function foo() : array {
  return ['foo', 1];
}

function bar(int $val) : void {
  echo (string) $val, "\n";
}

$foo = foo(); // $foo is array{0:string, 1:int}
$str = array_shift($foo); $foo is now array{0:int}: https://3v4l.org/slCsu
echo $str; // just here to get the psalm playground to not make noise about an unused variable
bar($foo[0]); // psalm thinks that $foo[0] is string|int, but if array_shift() were tracked, psalm would know it was an int.
```

muglug:
What I really need here, I think, is to use `array{string, int}` to mean a special `TTuple` type that extends `ObjectLike`.

It'd work like this:
```php
/**
 * @param array{0:string, 1:int} $arr
 */
function takesUnordered(array $arr) : void {}
/**
 * @param array{string, int} $arr
 */
function takesOrdered(array $arr) : void {}

takesUnordered([0 => "hello", 1 => 5]); // works
takesUnordered([1 => 5, 0 => "hello"]); // works
takesOrdered([0 => "hello", 1 => 5]); // works
takesOrdered([1 => 5, 0 => "hello"]); // fails
```

Then the `array_shift` etc. logic could be applied safely.

cc @TysonAndre who persuaded me to keep the `array{string, int}` shorthand

TysonAndre:
Another approach I've been thinking about would be to have a different notation for array types that are exclusively limited to those keys. I can't think of a good name, `tuple` only works for consecutive int keys

This is useful when APIs expect a limited set of fields (e.g. to catch typos)

- Unfortunately, the *order* of the keys might be an orthogonal check, so that might not work here. I guess consecutive keys starting from 0 could be a special case.

---

- Uppercase ARRAY would be confusing 
- Extra symbols such as `array*{0:string,1:int}` would also be confusing
- `strict-array{0:string,1:int}` would warn if extra keys were used (e.g. `['zero',1,2]`, `array{0:string,1:int,2?:int}`, etc

muglug:
I'm happy with `strict-array{...}`

muglug:
[...]

PR #1519, opened 2019-03-30
modifying callmap entry for error_get_last()

SignpostMarv:
I've seen error loggers say an error occurs on line 0 of unknown if an error was triggered in a shutdown function.
not sure what happens with type.
pretty sure that message is always there as a string.

* makes conditions checking for type redundant
* makes conditions checking for presence of `$err['message'];` redundant if `is_array($err)` has passed.

### DIFF
--- a/src/Psalm/Internal/CallMap.php
+++ b/src/Psalm/Internal/CallMap.php
@@ -2440,7 +2440,7 @@ return [
 'Error::getTrace' => ['array'],
 'Error::getTraceAsString' => ['string'],
 'error_clear_last' => ['void'],
-'error_get_last' => ['?array'],
+'error_get_last' => ['?array{message:string, type?:int, file?:string, line?:int}'],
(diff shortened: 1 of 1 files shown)


SignpostMarv:
this pr came up during #1500, filing separately in the event the review on this passes before we're fully done with the review on #1500.

bugreportuser:
> not sure what happens with type.

The [docs](https://www.php.net/manual/en/function.error-get-last.php) and [code](https://github.com/php/php-src/blob/7f262edaf3e12f81dfa204d760139149b681dce1/ext/standard/basic_functions.c#L4868-L4871) appear to confirm that `message`, `type`, `file`, and `line` always exist. I would think the type should be `?array{type:int, message:string, file:string, line:int}`.

Notice that I changed the order too. It doesn't matter for Psalm right now, but could [later](https://github.com/vimeo/psalm/issues/1255#issuecomment-458417105).

SignpostMarv:
yup, it does indeed appear to be the case I am operating under the assumption of old edge-cases, ty for pointing me towards that :)

muglug:
Thanks!

Issue #1467, opened 2019-03-15
mkdir failure throws ErrorException in some circumstances


SignpostMarv:
https://github.com/vimeo/psalm/blob/3704e75049703e861cda4585b66d49e3eb6810bd/src/Psalm/Config.php#L553

So there's an issue that occurs relatively infrequently at the dayjob, and I'm not 100% sure what the root issue is but the tl:dr is that I'm seeing `ErrorException` get thrown instead of mkdir just returning false, so the `@`  doesn't kick in and the `is_dir()` never executes.

The workaround I generally employ on the rare occasion that the problem occurs is to swap the order if `mkdir()` & `is_dir()`- is there a particular reason the `is_dir()` comes last ?

muglug:
> is there a particular reason the is_dir() comes last ?

Not AFAIK - mind PRing it?

SignpostMarv:
currently seeing local test failures on.... 3704e75049703e861cda4585b66d49e3eb6810bd - should I PR regardless ?

muglug:
The first one seems to be Windows-specific, the second is hmmm

SignpostMarv:
re: the windows-specific problem, is there ever an instance where `\` is actually needed on psalm? i.e. removing `DIRECTORY_SEPARATOR` in favour of '/' ? I'm not seeing any failures on windows if I change FileProvider & FileFilter to use `\RecursiveDirectoryIterator::UNIX_PATHS`

muglug:
Not AFAIK - if you know how to do it, it'd be great to add Windows to one of the Psalm build steps

SignpostMarv:
resolved by #1520

PR #1520, opened 2019-03-31
swapping order of mkdir & is_dir, as per #1467

SignpostMarv:
everything passes on travis except for the phar build


muglug:
Thanks!

PR #1527, opened 2019-04-03
Fix memory leak for exception code locations

bugreportuser:
After testing `checkForThrowsInGlobalScope` on a larger code base, I found that way too many code locations are stored. Psalm crashes with `Allowed memory size of 4294967296 bytes exhausted`. The same happens for `checkForThrowsDocblock`, since it also causes all code locations to be stored.

### DIFF
--- a/src/Psalm/Context.php
+++ b/src/Psalm/Context.php
@@ -213,7 +213,7 @@ class Context
     /**
      * A list of classes or interfaces that may have been thrown
      *
-     * @var array<string, array<int, CodeLocation>>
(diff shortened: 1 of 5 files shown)


muglug:
Thanks - I was planning to try out https://github.com/Roave/no-leaks (from @ocramius) which may catch this and other errors in the future

bugreportuser:
That's a great idea. I'm not sure it would catch this, since the array just stores the same objects way too many times, but it's worth trying out.

PR #1538, opened 2019-04-10
Allow ignoring exceptions only in global scope

bugreportuser:
Allows requiring that some exceptions are documented but ignoring them in global scope.

```xml
<ignoreExceptions>
    <class name="Exc2" onlyGlobalScope="true" />
</ignoreExceptions>
```

It can be a little unclear how `onlyGlobalScope` is interpreted, so I'm open to other names for it.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -99,8 +99,8 @@
 
     <xs:complexType name="ExceptionsType">
         <xs:sequence>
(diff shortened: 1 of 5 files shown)


muglug:
Would you mind documenting this somewhere?

bugreportuser:
@muglug Is this good?

muglug:
It's great, thanks

Issue #1547, opened 2019-04-12
Built-in ForbiddenCode check (var_dump, shell_exec) is case-sensitive


weirdan:
Psalm fails to report forbidden call to `var_dump()` when it's cased oddly.
https://psalm.dev/r/4e2bc2dc9c
```php
<?php
 vaR_dump("is not reported");
```
**Expected**: ForbiddenCode
**Actual**:
```
Psalm output (using commit 6e010d9): 

No issues!
```

PR #1553, opened 2019-04-13
Case-insensitive checks for var_dump and shell_exec

weirdan:
Fixes vimeo/psalm#1547

Seeing that other checks in the vicinity of the changed lines are also done in a case-sensitive fashion it might make sense to attempt more general fix.


muglug:
Thanks!

PR #1573, opened 2019-04-21
Link to human-readable package list

rmccue:
The Using Plugins page links to Packagist for Psalm plugins, but goes to the raw backend filter. I've added a link to the human-readable search instead, as this is likely what humans will want to look at. :)


muglug:
Thanks!

PR #1577, opened 2019-04-25
Fix RedundantCondition error messages for `$boolVar !== null`

TysonAndre:
Seen for this example:

    $b = rand() % 2 > 0;
    if ($b !== null) { ... }

These issue messages seem to have been copied without modification
from the code in the same file checking for impossible comparisons to the empty array


muglug:
Thanks!

PR #1602, opened 2019-05-03
Add support for absolute plugin paths.

jwage:
Add support for absolute plugin paths:

```xml
<?xml version="1.0"?>
<psalm>
    <projectFiles>
        <directory name="src" />
    </projectFiles>
    <plugins>
        <plugin filename="/absolute/path/to/StringChecker.php" />
    </plugins>
</psalm>
```

### DIFF
--- a/docs/authoring_plugins.md
+++ b/docs/authoring_plugins.md
@@ -24,9 +24,16 @@ Here are a couple of example plugins:
  - [FunctionCasingChecker](https://github.com/vimeo/psalm/blob/master/examples/plugins/FunctionCasingChecker.php) - checks that your functions and methods are correctly-cased
 
 To ensure your plugin runs when Psalm does, add it to your [config](configuration.md):
-```php
+```xml
     <plugins>
         <plugin filename="src/plugins/SomePlugin.php" />
+    </plugins>
+```
+
+You can also specify an absolute path to your plugin:
+```xml
+    <plugins>
+        <plugin filename="/path/to/SomePlugin.php" />
     </plugins>
 ```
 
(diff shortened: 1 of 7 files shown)


muglug:
Looks good!

> Checking for an absolute path can get messy to make it work between linux & windows

How so?

jwage:
To make it work consistently in windows you have to support things like `file://c:/path/to/something` and `zlib://c:/path/to/something`. In linux we can just check `if ($path[0] === '/')`. If you are ok with adding something like a `isAbsoluteMethod(string $path)` method somewhere I can do that. Let me know.

muglug:
> If you are ok with adding something like a isAbsoluteMethod(string $path) method somewhere I can do that.

I am enormously OK with that :)

muglug:
It could go in `command_functions.php`, which is available on every req.

jwage:
Ok, I will add it. Do you have any thoughts on how and what to assert here? https://github.com/vimeo/psalm/pull/1602/files#diff-afdc28abe8d80c678910cc44767a2759R799

I noticed some other tests don't have assertions so maybe we don't need it?

muglug:
Yeah I'm not worried about assertions, as long as there's a test that captures the opposite, failing case (here a plugin path that doesn't exist)

jwage:
To make `src/command_functions.php` available to the test suite I had to make a few changes to phpunit bootstrap.

jwage:
Also, this approach may technically be considered a BC break. If someone accidentally had an absolute path before this, I think it would still work as a relative path since it'd be concatenated with the base dir and would have double slashes but it would still work.

muglug:
Thanks!

PR #1605, opened 2019-05-05
fix some more return types

voku:
-> re-sync with https://github.com/JetBrains/phpstorm-stubs/pull/553


src/Psalm/Internal/Stubs/CoreGenericFunctions.php
```diff
@@ -133,7 +133,7 @@ function usort(array &$arr, callable $callback): bool {}
  * @psalm-template T
  *
  * @param array<string, T> $arr
- * @return array<string, T>
+ * @return array<string, T>|false
  */
 function array_change_key_case(array $arr, int $case = CASE_LOWER) {}
```
weirdan: I don't see where it would return `false` having passed valid arguments: https://github.com/php/php-src/blob/master/ext/standard/array.c#L4453-L4484

voku: yep, I think you are right ... maybe the documentation is not up-to-date https://www.php.net/manual/en/function.array-change-key-case.php ... I will revert it.

src/Psalm/Internal/CallMap.php
```diff
@@ -393,7 +393,7 @@
 'array_rand' => ['int|string|array<int,int>|array<int,string>', 'input'=>'array', 'num_req'=>'int'],
 'array_rand\'1' => ['int|string', 'input'=>'array'],
 'array_reduce' => ['mixed', 'input'=>'array', 'callback'=>'callable(mixed,mixed):mixed', 'initial='=>'mixed'],
-'array_replace' => ['array', 'arr1'=>'array', 'arr2'=>'array', '...args='=>'array'],
+'array_replace' => ['?array', 'arr1'=>'array', 'arr2'=>'array', '...args='=>'array'],
```
bugreportuser: `array_replace` returns null for non-array types, which Psalm checks: https://github.com/php/php-src/blob/master/ext/standard/array.c#L3746-L3752

muglug:
Thanks!

PR #1606, opened 2019-05-06
Replace php-cs-fixer/diff in favor of sebastian/diff

andrew-demb:
`php-cs-fixer/diff` marked as "internal" for php-cs-fixer tool and not recommended for extrernal usage.

I suggest it was added for compatibility with php 5.6. Now psalm requires ^7.0.

How I check for same output:
```bash
cd /path/to/psalm
./psalter --dry-run --issues=UnusedProperty
```

Output for both versions (`sebastian/diff` vs `php-cs-fixer/diff`)
```
Scanning files...
Analyzing files...
/path/to/psalm/psalm/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php:
--- Original
+++ New
@@ -59,15 +59,6 @@
     /** @var null|string */
     private $new_psalm_return_type;
 
-    /** @var array<string, int> */
-    private $param_typehint_area_starts = [];
-
-    /** @var array<string, int> */
-    private $param_typehint_starts = [];
-
-    /** @var array<string, int> */
-    private $param_typehint_ends = [];
-
     /** @var array<string, string> */
     private $new_php_param_types = [];
 

No errors found!

Checks took 20.68 seconds and used 326.509MB of memory
Psalm was able to infer types for 99.7984% of the codebase

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,6 @@
         "nikic/php-parser": "^4.0.2 || ^4.1",
         "openlss/lib-array2xml": "^0.0.10||^0.5.1",
         "muglug/package-versions-56": "1.2.4",
-        "php-cs-fixer/diff": "^1.2",
         "composer/xdebug-handler": "^1.1",
         "felixfbecker/language-server-protocol": "^1.2",
         "felixfbecker/advanced-json-rpc": "^3.0.3",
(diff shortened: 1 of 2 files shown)


andrew-demb:
I missed, that `sebastian/diff:^3.0` requires at least php 7.1.

`sebastian/diff:^2.0` do not provide compatible format from the box

andrew-demb:
Now it's compatible :)

muglug:
Thanks - there was a conflict so I applied myself: https://github.com/vimeo/psalm/commit/0f3385ce95e1d9f3a5bca0ddf5681892263c6c01

PR #1615, opened 2019-05-08
Fix isset check

ramon-b:
Check if index exists in the array with open files, instead of checking
if index exists in the array with temporary files.

I was receiving errors in Visual Studio Code: ErrorException: Undefined index.


muglug:
Oh thanks!

Issue #1614, opened 2019-05-07
Extend @internal syntax to encapsulate elements within a namespace


bdsl:
It would be useful to be able to restrict the use of classes and other elements to within the library or submodule that they live in, particularly to allow changes to be made without breaking backwards compatibility. Although PHP has no concept of either a library or a submodule, the [@internal](https://docs.phpdoc.org/references/phpdoc/tags/internal.html) annotation can sometimes be used to achieve this.

Psalm interprets`@internal` as meaning that the declaring and calling code must not be in 'unrelated namepsaces', and emits issues as shown at https://psalm.dev/r/90a95c1dd8 . 

However this only works because the two namespaces are completely unrelated, and so might not be usable if `Foo` and `Bat` were both written by members of the same organisation, or in the particular case of Drupal if `Foo` was part of Drupal Core, and `Bat` was part of a bespoke applications controller, since the Drupal docs encourage the use of related namespaces.

So my proposal is to introduce a new syntax for internal, probably initially as `@psalm-internal`, which would take a a parameter to declare exactly what module the element is internal to. Psalm would act as shown below:

```php
<?php
namespace A\FooLibrary {
    /**
     * @psalm-internal A\FooLibrary
     */
    class Foo { }
}

namespace A\FooLibrary\Bar {
    class Bat {
        public function batBat(): void {
            new \A\FooLibrary\Foo(); // no issue emitted
        }
    }
}

namespace A {
    class Bat {
        public function batBat(): void {
            new \A\FooLibrary\Foo(); // proposed variant of InternalClass emitted
        }
    }
}

namespace A\FooLibraryEtc {
    class Bat {
        public function batBat(): void {
            new \A\FooLibrary\Foo(); // proposed variant of InternalClass emitted
        }
    }
}
```

I haven't contributed to Psalm before, but if there's interest in this I'd like to try to implement it and submit a PR.

bdsl:
If I was to work on this I'd be tempted to also consolidate the 12 occurrences of the regex `'/^([^\\\]+).*/'`that are in the codebase as part of the implementation of `@internal`.

bdsl:
So it needs more work, but I've got the start of an implementation. Currently it just stops a class extending a parent which is marked as internal to a different namespace. I'll implement checks for other types of usage and tidy stuff up a bit before I make the PR.

Would have been a bit quicker if I hadn't spent about half an hour debugging before I realised I'd put the docbloc on to the namespace instead of the class in the test code.

[...]

PR #1623, opened 2019-05-11
Allow internal annotation to specify namespaces

bdsl:
Extend @internal syntax to encapsulate elements within a namespace (https://github.com/vimeo/psalm/issues/1614)

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -211,12 +211,12 @@ class ClassAnalyzer extends ClassLikeAnalyzer
                     }
                 }
 
-                if ($parent_class_storage->psalmInternal &&
-                    strpos($fq_class_name, trim($parent_class_storage->psalmInternal, '\\') . '\\') !== 0
+                if ($parent_class_storage->psalm_internal &&
+                    strpos($fq_class_name, trim($parent_class_storage->psalm_internal, '\\') . '\\') !== 0
                 ) {
                     if (IssueBuffer::accepts(
                         new InternalClass(
-                            $parent_fq_class_name . ' is internal to ' . $parent_class_storage->psalmInternal,
+                            $parent_fq_class_name . ' is internal to ' . $parent_class_storage->psalm_internal,
                             $code_location
                         ),
                         array_merge($storage->suppressed_issues, $this->getSuppressedIssues())
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -194,11 +194,7 @@ class ClassAnalyzer extends ClassLikeAnalyzer
                         $class_context ? $class_context->include_location : null,
                         true
                     );
-
-                    $self_root = preg_replace('/^([^\\\]+).*/', '$1', $fq_class_name);
-                    $declaring_root = preg_replace('/^([^\\\]+).*/', '$1', $parent_fq_class_name);
-
-                    if (strtolower($self_root) !== strtolower($declaring_root)) {
+                    if (! NamespaceAnalyzer::nameSpaceRootsMatch($fq_class_name, $parent_fq_class_name)) {
                         if (IssueBuffer::accepts(
                             new InternalClass(
                                 $parent_fq_class_name . ' is marked internal',
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -212,7 +212,7 @@ class ClassAnalyzer extends ClassLikeAnalyzer
                 }
 
                 if ($parent_class_storage->psalm_internal &&
-                    strpos($fq_class_name, trim($parent_class_storage->psalm_internal, '\\') . '\\') !== 0
+                    ! NamespaceAnalyzer::isWithin($fq_class_name, $parent_class_storage->psalm_internal)
                 ) {
                     if (IssueBuffer::accepts(
                         new InternalClass(
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -211,6 +211,22 @@ class ClassAnalyzer extends ClassLikeAnalyzer
                     }
                 }
 
+                if (
+                    $parent_class_storage->psalmInternal &&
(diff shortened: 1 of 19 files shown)


bdsl:
It might be worth also throwing a DocblockParseException if any docblock contains @psalm-internal and doesn't also contain @internal, since other tools will make use of @internal and I don't think there's any good use case for the former without the latter.

bdsl:
This PR adds three uses of `InternalClass`, two of `InternalProperty` and one of `InternalMethod` to the codebase. They are when psalm detects code that improperly:

1. extends an internal class,
2. calls a function of an internal class,
3. instantiates an internal class,
4. fetches an internal property,
5. sets an internal property, or
6. call an internal function

I think 1. may include implementing interfaces and using traits.

This directly parallels the six similar calls that are already in master for dealing with the `@internal` tag.

I think there are some more situations psalm could usefully catch that could be added to both of these lists. Some I can think of are:

1. Internal class used as type annotation for property or variable
2. Internal class used as type annotation or native type hint for function param or return
3. Internal class or interface used with instanceOf
4. Internal class constant fetched (including `::class`)
5. Internal constant fetched
6. Property of Internal Class fetched
7. Property of Internal Class set.

Checks for 6 and 7 are might not be necessary if 1, 2, and 3 are checked.

src/Psalm/Internal/Scanner/FunctionDocblockComment.php
```diff
@@ -38,6 +38,13 @@ class FunctionDocblockComment
      */
     public $deprecated = false;
 
+    /**
+     * If set, the class is internal to the given namespace.
```
bdsl:
```suggestion
     * If set, the function is internal to the given namespace.
```

src/Psalm/Internal/Scanner/VarDocblockComment.php
```diff
@@ -41,4 +41,11 @@ class VarDocblockComment
      * @var bool
      */
     public $internal = false;
+
+    /**
+     * If set, the class is internal to the given namespace.
```
bdsl:
```suggestion
     * If set, the property is internal to the given namespace.
```

muglug:
Sorry for the radio silence - I’ll review and hopefully merge tomorrow evening

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
```diff
@@ -211,6 +211,20 @@ public function analyze(
                     }
                 }
 
+                if ($parent_class_storage->psalmInternal &&
+                    strpos($fq_class_name, trim($parent_class_storage->psalmInternal, '\\') . '\\') !== 0
+                ) {
+                    if (IssueBuffer::accepts(
```
bdsl: I followed the practice of wrapping calls to `IssueBuffer::accepts` in `if`, but I'm wasn't clear on the purpose of the `if`.

src/Psalm/Internal/Analyzer/MethodAnalyzer.php
```diff
@@ -206,6 +206,26 @@ public static function checkMethodNotDeprecatedOrInternal(
             }
         }
 
+        if ($storage->psalmInternal
+            && $context->self
+            && !$context->collect_initializations
+            && !$context->collect_mutations
+        ) {
+            if (strpos($context->self, trim($storage->psalmInternal, '\\') . '\\') !== 0) {
```
bdsl: I've made six copies of essentially same function built around `strpos` to check if one namespace is inside another. I'd like to extract it to a function, but I'm not sure if there's a good place in the codebase to put it. Same applies to the pre-existing code used to check if two namespaces are unrelated.

bdsl:
Within any namespace, e.g. `Foo\Bar\Baz`, the annotation `@internal` is exactly equivalent to `@internal Foo`. So perhaps we should treat it as the latter during parsing, and delete the older code in analyser classes that handles it.

src/Psalm/Internal/Analyzer/CommentAnalyzer.php
```diff
@@ -126,13 +126,13 @@ public static function getTypeFromComment(
                 $var_comment->deprecated = isset($comments['specials']['deprecated']);
                 $var_comment->internal = isset($comments['specials']['internal']);
                 if (isset($comments['specials']['psalm-internal'])) {
-                    $psalmInternal = reset($comments['specials']['psalm-internal']);
-                    if ($psalmInternal) {
-                        $var_comment->psalmInternal = $psalmInternal;
+                    $psalm_internal = reset($comments['specials']['psalm-internal']);
```
bdsl: reset is needed because the array $comments['specials']['psalm-internal'] doesn't start at zero - it starts at a different number. I didn't track down quite why that is.

src/Psalm/Internal/Analyzer/MethodAnalyzer.php
bdsl: And would be good to also make the check case-insensitive.

bdsl:
There's also potential for preventing various types of leaky abstractions - e.g. if a class is internal then it would be an error to use it as a type hint or annotation in a non-internal method.

muglug:
> Within any namespace, e.g. Foo\Bar\Baz, the annotation @internal is exactly equivalent to @psalm-internal Foo. So perhaps we should treat it as the latter during parsing, and delete the older code in analyser classes that handle it.

That sounds like a great idea!

bdsl:
@muglug I'm not sure if you've had a good chance to review this yet. Will you tell me what you think when you have? Thanks.

muglug:
It looks good - I just thought you might want to implement the idea you mentioned above?

bdsl:
Ah right, thanks. Sorry I didn't want to start on that until I knew if you'd looked through the changes so far. I'll try to implement the combined parsing in the next few days. Can you suggest a good place to put the namespace comparison function so it isn't repeated six times?

muglug:
there's a `NamespaceAnalyzer` - may not be the worst place

bdsl:
So I tried implementing the idea above, but I found that in some of the analysis functions, e.g. `CommentAnalyzer::extractFunctionDocblockInfo` there didn't seem to be an easy to way to access the containing namespace. I thought it was better to keep things consistent rather than remove just some of the $internal properties.

I have done a bit of refactoring, and I also implemented a rule that `@psalm-internal` can never be used without `@internal`. I think since people are likely to use IDEs and other analysis tools alongside psalm, it would make sense to not just use an annotation that only psalm understands.

I've also allowed things to be marked internal to classes as well as namespaces. I'm not sure if there's a good use case for that, but I don't see why we shouldn't allow it.

So @muglug  if you're happy with this please go ahead and merge when the checks complete. Let me know if you'd like me to rebase on master or anything like that.

muglug:
Thanks!

PR #1631, opened 2019-05-13
Fix #1622 - add allow-backwards-incompatible-changes flag to psalter

### LINKED ISSUE #1622: No way to prevent psalter adding backwards-incompatible changes for packages that extend your classes
This is probably more of a question than an error report.

We have our own legacy framework that we use as a composer package in many projects. I'd like to improve type hints using `psalter` but the problem is that `psalter` adds return types even when that is not safe.

For example, `psalter` will add native `void` return types. Unfortunately, some of those methods will be overridden in the project and return something.

Basically, my question is: Can the code base be marked as a library? In that case `psalter` should not add native type hints to code that can be overridden. I'd still like to have native type hints for `final` classes or `private` methods.

I realize this is probably not a common use case, feel free to close this if you think it's not worth it.

iluuu1994:
The flag prevents psalter from adding native return types

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
@@ -721,14 +721,31 @@ class ReturnTypeAnalyzer
             $function_like_analyzer->getMethodId(),
             $function
         );
+
+        $codebase = $project_analyzer->getCodebase();
+        $is_final = true;
+        $fqcln = $source->getFQCLN();
+        if ($fqcln !== null) {
+            $class_storage = $codebase->classlike_storage_provider->get($fqcln);
+            assert($function instanceof ClassMethod);
+            $is_final = $function->isFinal() || $class_storage->final;
+        }
+
+        $allow_native_type = !$docblock_only
+            && $codebase->php_major_version >= 7
+            && (
+                $codebase->allow_backwards_incompatible_changes
+                || $is_final
+            );
+
         $manipulator->setReturnType(
-            !$docblock_only && $project_analyzer->getCodebase()->php_major_version >= 7
+            $allow_native_type
                 ? $inferred_return_type->toPhpString(
                     $source->getNamespace(),
                     $source->getAliasedClassesFlipped(),
                     $source->getFQCLN(),
-                    $project_analyzer->getCodebase()->php_major_version,
-                    $project_analyzer->getCodebase()->php_minor_version
+                    $codebase->php_major_version,
+                    $codebase->php_minor_version
                 ) : null,
             $inferred_return_type->toNamespacedString(
                 $source->getNamespace(),
--- a/src/psalter.php
+++ b/src/psalter.php
@@ -21,6 +21,7 @@ $valid_long_options = [
     'help', 'debug', 'debug-by-line', 'config:', 'file:', 'root:',
     'plugin:', 'issues:', 'php-version:', 'dry-run', 'safe-types',
     'find-unused-code', 'threads:', 'codeowner:',
+    'allow-backwards-incompatible-changes:',
 ];
 
 // get options from command line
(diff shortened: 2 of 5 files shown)


src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
```diff
@@ -721,8 +721,22 @@ private static function addOrUpdateReturnType(
             $function_like_analyzer->getMethodId(),
             $function
         );
+
+        $is_final = true;
+        $fqcln = $source->getFQCLN();
+        if ($fqcln !== null) {
+            $class_storage = $source->getCodebase()->classlike_storage_provider->get($fqcln);
```
muglug: You may want to have `$codebase = $source->getCodebase();` up top, then reference that variable in the places below.

src/psalter.php
```diff
@@ -117,6 +118,9 @@ function ($arg) use ($valid_long_options, $valid_short_options) {
 
     --codeowner=[codeowner]
         You can specify a GitHub code ownership group, and only that owner's code will be updated.
+
+    --allow-backwards-incompatible-changes=BOOL
```
muglug: Add a not to say it defaults to `true`

tests/FileManipulation/FileManipulationTest.php
```diff
@@ -30,10 +30,11 @@ public function setUp()
      * @param string $php_version
      * @param string[] $issues_to_fix
      * @param bool $safe_types
+     * @param bool $allow_backwards_incompatible_changes
      *
      * @return void
      */
-    public function testValidCode($input_code, $output_code, $php_version, array $issues_to_fix, $safe_types)
+    public function testValidCode($input_code, $output_code, $php_version, array $issues_to_fix, $safe_types, bool $allow_backwards_incompatible_changes)
```
muglug: you could add a default, and then you wouldn't need to update every test (I don't think)

src/psalter.php
bdsl: Rather than a `--allow-backwards-incompatible-changes=<BOOL>` option who'ss omission is equivalent to  `--allow-backwards-incompatible-changes=true`, I think it might be more intuitive to have a  `--no-backwards-incompatible-changes` option that doesn't take a value.

src/psalter.php
```diff
@@ -302,6 +306,20 @@ function (string $line) : bool {
...
 
+if (isset($options['allow-backwards-incompatible-changes'])) {
+    $allow_backwards_incompatible_changes = filter_var(
+        $options['allow-backwards-incompatible-changes'],
+        FILTER_VALIDATE_BOOLEAN,
+        ['flags' => FILTER_NULL_ON_FAILURE]
+    );
+
+    if ($allow_backwards_incompatible_changes === null) {
+        die('--allow-backwards-incompatible-changes expectes a boolean value [true|false|1|0]' . PHP_EOL);
```
bdsl: This gives the user a choice of five options - the four listed above, plus not passing the option at all. Only two options are really needed, and in general the fewer options we give people the easier it should be for them to make good decisions.

muglug: Yeah, this was my suggestion but I'd also be happy with `--prevent-backwards-incompatible-changes`, too

iluuu1994: I prefer not to negate boolean variables (or flags as in this case). 'incompatible' is already the negated form of 'compatible' so that would make it a double negation. But I'll let you decide @muglug.

muglug: Yeah, it's generally seen as bad design (hence why I have `--show-info=false` vs `--hide-info`). On the other hand, it's less wordy.

Your PR, you choose (and it sounds like you want to stick with the existing one)

muglug: Say the word, and I'll merge

iluuu1994: Let's keep it this way to stay consistent with the other flags. Thanks @bdsl for the suggestion though!

muglug:
Thanks all!

PR #1637, opened 2019-05-14
Remove unused properties

iluuu1994:


muglug:
Yeah at _some_ point I plan to use these, but in the meantime it just generates noise

iluuu1994:
Saw the `info`s when playing around, wasn't sure if this was left-over code or intentional.

PR #1640, opened 2019-05-15
Document intersection and no-return types in docblock syntax

bdsl:
Also removed spaces from union type example.


muglug:
Thanks!

Issue #1621, opened 2019-05-10
LSP Completion: UnanalyzedFileException


joshdifabio:
I'm running VS Code with the Psalm extension, and VS Code is not sending any didOpen, didClose, didSave notifications to Psalm's language server at all. (I've determined this by logging all messages in [`Psalm\Internal\LanguageServer\LanguageServer`](https://github.com/vimeo/psalm/blob/751253ddf030f402453bea2cad035c0db0ad3242/src/Psalm/Internal/LanguageServer/LanguageServer.php#L98).)

This means that completion requests tend to [error in `Psalm\Codebase`](https://github.com/vimeo/psalm/blob/751253ddf030f402453bea2cad035c0db0ad3242/src/Psalm/Codebase.php#L1070) as the file being edited is not 'open' [according to `Psalm\Internal\Provider\FileProvider`](https://github.com/vimeo/psalm/blob/751253ddf030f402453bea2cad035c0db0ad3242/src/Psalm/Internal/Provider/FileProvider.php#L92). I'm not sure why didOpen is not being sent, but according to the LSP docs, ["a server’s ability to fulfil requests is independent of whether a text document is open or closed,"](https://microsoft.github.io/language-server-protocol/specification#textDocument_didOpen) so Psalm should probably deal with the case gracefully.

Now, although didOpen is not being sent, didChange _is_ being sent, so `FileProvider` does contain an entry in its `temp_files` array for the file which is being edited. So, the most straightforward workaround (tested and working) would be to modify [`FileProvider::isOpen()`](https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Provider/FileProvider.php#L90) so that it returns true if either `open_files` _or_ `temp_files` contains the file in question.

This could be considered a bit of a hack, as the file isn't really 'open' as per the LSP definition if didOpen has not been sent. However, there is already precedent [[1]](https://github.com/vimeo/psalm/blob/751253ddf030f402453bea2cad035c0db0ad3242/src/Psalm/Internal/LanguageServer/Server/TextDocument.php#L172)[[2]](https://github.com/vimeo/psalm/blob/751253ddf030f402453bea2cad035c0db0ad3242/src/Psalm/Internal/LanguageServer/Server/TextDocument.php#L215) in Psalm's LS implementation which suggests that `FileProvider`'s concept of 'open' is not identical to LSP's definition of 'open'.

muglug:
Could this be related to https://github.com/vimeo/psalm/pull/1615

[...]

PR #1653, opened 2019-05-17
FileProvider: Consider file open if temp changes

joshdifabio:
The VS Code LSP client does not always send didOpen messages when a file
is being edited. This causes code completion requests to fail because the
completion request refers to files which are not 'open' according to
FileProvider.

This commit changes FileProvider so that a file is considered 'open' if it is
being modified, even if no didOpen message was received from the LSP
client.


joshdifabio:
Fixes #1621.

muglug:
Thanks!

PR #1662, opened 2019-05-21
Allow coercion from empty to array param

muglug:
This allows users to assign templates with `empty`-inferred types to values with known types.

`empty` is a special type in Psalm that represents a key/value of a yet-to-be-filled array.

Can anyone think of a reason why this might be bad?

cc @vasily-kartashov @iluuu1994 @joshdifabio @SignpostMarv @ocramius

### DIFF
--- a/tests/Template/TemplateTest.php
+++ b/tests/Template/TemplateTest.php
@@ -2230,6 +2230,29 @@ class TemplateTest extends TestCase
                     '$y' => 'array<array-key, A&B>',
                 ]
             ],
+            'templateEmptyParamCoercion' => [
+                '<?php
+                    namespace NS;
+                    use Countable;
+
+                    /** @template T */
+                    class Collection
+                    {
+                        /** @psalm-var iterable<T> */
+                        private $data;
+
+                        /** @psalm-param iterable<T> $data */
+                        public function __construct(iterable $data = []) {
+                            $this->data = $data;
+                        }
+                    }
+
+                    class Item {}
+                    /** @psalm-param Collection<Item> $c */
+                    function takesCollectionOfItems(Collection $c): void {}
+
+                    takesCollectionOfItems(new Collection());',
+            ],
         ];
     }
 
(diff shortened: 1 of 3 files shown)


tests/Template/TemplateTest.php
```diff
@@ -2230,6 +2230,29 @@ interface B {}
...
+                        public function __construct(iterable $data = []) {
+                            $this->data = $data;
+                        }
+                    }
+
+                    class Item {}
+                    /** @psalm-param Collection<Item> $c */
+                    function takesCollectionOfItems(Collection $c): void {}
+
+                    takesCollectionOfItems(new Collection());',
```
Ocramius: So the idea is that `[]` is implicitly upcast to an `iterable<T>` here?

To make this more readable, I'd use an explicit `[]` at call-time, rather than as default parameter value: that would make the example more readable.

iluuu1994:
Would psalm catch something like this?

```php
/** @template T */
class Collection
{
    /** @psalm-var iterable<T> */
    private $data;

    /** @psalm-param iterable<T> $data */
    public function __construct(iterable $data = []) {
        $this->data = $data ?: [1, 2, 3];
    }
}
```

Ocramius:
@iluuu1994 I think this should already error out (without caller):

```php
/** @template T */
class Collection
{
    /** @psalm-var iterable<T> */
    private $data;

    /** @psalm-param iterable<T> $data */
    public function __construct(iterable $data = []) {
        $this->data = $data ?: [1, 2, 3];
    }
}
```

muglug:
Yeah, template type checks _inside_ templated classes are weak, but could probably be tougher

iluuu1994:
What about this?

```php
$collection = new Collection();
takesInts($collection);
takesStrings($collection); // Error, as the collection could have been mutated
```

Does the first call change the type of the collection?

If that works I see nothing wrong with this PR. 👍

Ocramius:
```
// Error, as the collection could have been mutated
```

The type of an object cannot mutate over time: this should be prevented by inspecting the object internals

tests/Template/TemplateTest.php
muglug: I'll add both examples

muglug:
@iluuu1994 

> What about this?

Prep work for that done in https://github.com/vimeo/psalm/commit/4ecf370900e22060cc5275a3f8eb4f00bedb8f2f

muglug:
@iluuu1994 

```php
/** @param Collection<int> $c */
function takesIntCollection(Collection $c): void {}

/** @param Collection<string> $c */
function takesStringCollection(Collection $c): void {}

$collection = new Collection([]);
takesIntCollection($collection);
takesStringCollection($collection);
```

now gives the same sort of error as 

```php
function takesInt(int $i) : void {}

function bar($s) : void {
  takesInt($s);
  takesString($s);
}
```

PR #1668, opened 2019-05-23
Remove useless echo from tests

iluuu1994:


muglug:
Thanks, dumb of me to add accidentally

PR #1671, opened 2019-05-24
Write Psalm version to baseline

ErikBooijCB:
For our baseline we've found that it's quite relevant to note on which version of Psalm the baseline was created, primarily because the rapid development of Psalm means we often run into "new" issues, because they simply weren't detected before. Since the "rule" within our development department is to only ever add issues to the baseline when the Psalm version updates, this would help transparency.

Also, we've started committing a readable version of the baseline and generating that with the version number in the baseline would make it more reliable and easier to build.

Screen Shot 2019-05-24 at 14 38 19


ADmad:
> the rapid development of Psalm means we often run into "new" issues

Yeah @muglug's efficiency is annoying sometimes :stuck_out_tongue_closed_eyes:.

muglug:
Thanks!

PR #1685, opened 2019-05-25
Ignore Tnever when in union with any other type

bdsl:
Previously an InvalidReturnType would have been emitted for the code
in the new tests. I don't know if there's a practical value to
including Tnever in a union, but I believe it's correct for
TNever|T to always equal T, whatever type T is.

### DIFF
--- a/tests/ReturnTypeTest.php
+++ b/tests/ReturnTypeTest.php
@@ -563,7 +563,7 @@ class ReturnTypeTest extends TestCase
                 '<?php
                     namespace Foo;
                     /**
-                     * @return never-returns|int
+                     * @return int|never-returns
                      */
                     function foo() : int {
                         return 3;
--- a/tests/ReturnTypeTest.php
+++ b/tests/ReturnTypeTest.php
@@ -549,6 +549,26 @@ class ReturnTypeTest extends TestCase
                         exit();
                     }',
             ],
+            'neverReturnsUnion' => [
+                '<?php
+                    namespace Foo;
+                    /**
+                     * @return never-returns|int
+                     */
+                    function foo() : int {
+                        return 3;
+                    }',
+            ],
+            'neverReturnsReversed' => [
+                '<?php
+                    namespace Foo;
+                    /**
+                     * @return never-returns|int
+                     */
+                    function foo() : int {
+                        return 3;
+                    }',
+            ],
             'neverReturnsCovariance' => [
                 '<?php
                     namespace Foo;
--- a/tests/ReturnTypeTest.php
+++ b/tests/ReturnTypeTest.php
@@ -559,7 +559,7 @@ class ReturnTypeTest extends TestCase
                         return 3;
                     }',
             ],
-            'neverReturnsReversed' => [
+            'neverReturnsUnionReversed' => [
                 '<?php
                     namespace Foo;
                     /**
--- a/src/Psalm/Type/Union.php
+++ b/src/Psalm/Type/Union.php
@@ -15,6 +15,7 @@ use Psalm\Type\Atomic\TLiteralInt;
 use Psalm\Type\Atomic\TLiteralString;
 use Psalm\Type\Atomic\TGenericObject;
 use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Atomic\TNever;
 use Psalm\Type\Atomic\TString;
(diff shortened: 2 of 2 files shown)


tests/ReturnTypeTest.php
```diff
@@ -549,6 +549,26 @@ function foo() : void {
...
+                '<?php
+                    namespace Foo;
+                    /**
+                     * @return never-returns|int
+                     */
+                    function foo() : int {
+                        return 3;
+                    }',
+            ],
+            'neverReturnsReversed' => [
```
bdsl:
```suggestion
            'neverReturnsUnionReversed' => [
```

tests/ReturnTypeTest.php
```diff
@@ -549,6 +549,26 @@ function foo() : void {
...
+                     */
+                    function foo() : int {
+                        return 3;
+                    }',
+            ],
+            'neverReturnsUnionReversed' => [
+                '<?php
+                    namespace Foo;
+                    /**
+                     * @return never-returns|int
```
bdsl:
```suggestion
                     * @return int|never-returns
```

bdsl:
@muglug Will you let me know whether you think this is a change worth making? Thanks.

iluuu1994:
`never-returns` isn't useful in a union so maybe it's better to just disallow that in the first place.

muglug:
Yeah, sorry. I’m not sure this is particularly valuable as-is. I’d definitely welcome a change that made unions with `never-return` impossible

bdsl:
@iluuu1994 Maybe, but then how can Psalm really 'disallow' anything. It has to analyse whatever codebase exists.

Psalm could raise an issue if there was an attempt to use `never-returns` in a union, but presumably it would still keep going.

I don't see any good reason for a programmer to manually annotate something as returning a union with `never-return`, since it's like adding zero, but I imagined it could happen either internally in psalm or by some other automatic tool. 

I just had a a go at triggering this with an if condition but psalm already gets the right result here, i.e. infers the string return type, so I suppose it doesn't treat this as `string|no-return`:

```
<?php
  
/** @return never-returns */
function dieLoudly()
{
  echo 'Arrrgh';
  exit(1);
}

function foo(string $s) {
    if ($s > 'foo') {
      return "bar";
    }
    dieLoudly();
}
```

iluuu1994:
@bdsl Good point. Maybe your PR in combination with an error on usage of `never-return` in a union would make most sense.

bdsl:
Typescript allows never in a union without complaint: https://typescript-play.js.org/#code/FAYw9gdgzgLgBAMzGOBeRBXCIYEtIAUAlAFxywBOuEA5gD4QCmAboxXAN7BxwWMwYKEOAHIkYEQG5gAX0lA

To me it seems like something that should be allowed without any issue in the interests of consistency, in the same way that my calculator lets me do `5 + 0` if I want to.

iluuu1994:
I'd argue an error would still be helpful as the user could misunderstand what `never-returns` does exactly. It's also a direct contradiction, something can't return something and never return at the same time.

bdsl:
I suppose there must be a much may general change that would make this PR unnecessary, where instead of ignoring specifically `never-returns` in union with anything, we ignore any type `T` in a union with any other type `E` when `T` is contained within `E`.

bdsl:
@iluuu1994 I agree a message might be helpful. I don't think it's a contradiction though. A union isn't about doing two things at the same time. Something can't return a string and return an int at the same time, but `@return string|int` is meaningful.

iluuu1994:
@bdsl A return type of `int` doesn't mean the function can't not return. It just means that if it returns it will return an `int`. `never-returns` promises that the function will under no circumstances return which would contradict it returning a different type under some other circumstances.

bdsl:
Ultimately what we take things to mean is a choice, so there's no absolute right answer, but I don't agree. As you say, a return type of `int` 'means that if it returns it will return an `int`. In other words it promises that the function will under no circumstances return anything which is not an it.

That 'contradicts' `string` in exactly the same way that a return type of `never-returns` does.

There's no reason we have to follow it, but I think this from the Typescript docs is a good way to think of it: '[The never type is a subtype of, and assignable to, every type](https://www.typescriptlang.org/docs/handbook/basic-types.html)'

PR #1708, opened 2019-05-30
Fix readability for object-like types when running `psalter`

Kocal:
Hi, this PR try to _fix_ (improve?) the output of object-like types when running `psalter`.

**before:**
```
array{a:string, b?:string}
array{a:int, b:int, c:array{a:int, b:int, c:array{a:int, b:int, c:int}}}
```

**after:**
```
array{ a: string, b?: string }
array{ a: int, b: int, c: array{ a: int, b: int, c: array{ a: int, b: int, c: int } } }
```

I know that can be subjective, but I really think it improves readability. Also this PR will only affects `psalter` (right?) so it's okay I guess.
What do you think?
Thanks!


vasily-kartashov:
And scrollability :)

bdsl:
I'm not sure this improves the readability for me. I think the spaces inside the brackets are too much, and they are inconsitent with PSR-2 which disalows them (as enforced by https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.15/src/Fixer/Whitespace/NoSpacesInsideParenthesisFixer.php#L26 ) . On the other hand the [Wordpress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage) require spaces in parens.

Maybe a good compromise would be:

```
array{a: string, b?: string}
array{a: int, b: int, c: array{a: int, b: int, c: array{a: int, b: int, c: int}}}
```

But even with that I'm not sure if I prefer it to the current output.

It could be made a lot more readable when types get complex by using line breaks rather than spaces,
but I think that would be a lot more work to implement something like the following:

```
array{a: string, b?: string}


array{a: int, b: int, c: array{
    a: int,
    b: int,
    c: array{a: int, b: int, c: int},
}}

```

ADmad:
> I think the spaces inside the brackets are too much

Me too.

Kocal:
I think `array{a: string, b?: string}` is a good compromise. 
In fact the main issue for me was `varname:type` which can quickly start to be unreadable.

muglug:
I've updated the spacing, should be good

muglug:
Thanks!

Kocal:
Ah, thanks you too :)

Issue #1714, opened 2019-06-01
`tests/Config/PluginTest.php` fails when run in isolation


weirdan:
```console
$ vendor/bin/phpunit --no-coverage tests/Config/PluginTest.php
PHPUnit 8.1.6 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.4-2 with Xdebug 2.7.0
Configuration: /home/weirdan/src/psalm/phpunit.xml

FFFEFFEFEE..........EF                                            22 / 22 (100%)

Time: 4.63 seconds, Memory: 36.00 MB

There were 5 errors:

1) Psalm\Tests\Config\PluginTest::testEchoAnalyzerPluginWithJustHtml
Error: Call to undefined function Psalm\isAbsolutePath()

/home/weirdan/src/psalm/src/Psalm/Config.php:806
/home/weirdan/src/psalm/tests/Config/PluginTest.php:192

2) Psalm\Tests\Config\PluginTest::testEchoAnalyzerPluginWithEscapedString
Error: Call to undefined function Psalm\isAbsolutePath()

/home/weirdan/src/psalm/src/Psalm/Config.php:806
/home/weirdan/src/psalm/tests/Config/PluginTest.php:303

3) Psalm\Tests\Config\PluginTest::testFloatCheckerPluginIssueSuppressionByConfig
Error: Call to undefined function Psalm\isAbsolutePath()

/home/weirdan/src/psalm/src/Psalm/Config.php:806
/home/weirdan/src/psalm/tests/Config/PluginTest.php:390

4) Psalm\Tests\Config\PluginTest::testFloatCheckerPluginIssueSuppressionByDocblock
Error: Call to undefined function Psalm\isAbsolutePath()

/home/weirdan/src/psalm/src/Psalm/Config.php:806
/home/weirdan/src/psalm/tests/Config/PluginTest.php:429

5) Psalm\Tests\Config\PluginTest::testPluginFilenameCanBeAbsolute
Error: Call to undefined function Psalm\isAbsolutePath()

/home/weirdan/src/psalm/src/Psalm/Config.php:806
/home/weirdan/src/psalm/tests/Config/PluginTest.php:839

--

There were 7 failures:

1) Psalm\Tests\Config\PluginTest::testStringAnalyzerPlugin
Failed asserting that exception of type "Error" matches expected exception "Psalm\Exception\CodeException". Message was: "Call to undefined function Psalm\isAbsolutePath()" at
[...]

PR #1715, opened 2019-06-01
Fixed vimeo/psalm#1714

weirdan:
Moved `isAbsolutePath()` into a functions file (`src/functions.php`) that is now automatically loaded by composer autoload.

/cc: @jwage


weirdan:
Done.

muglug:
Thanks!

Issue #1710, opened 2019-05-30
Support for finding plugins when using composer-bin-plugin


gjuric:
I am using https://github.com/bamarni/composer-bin-plugin to install psalm to get around an unrelated issues.

When psalm is installed in a separate **bin**  and a plugin is installed in the same **bin** it looks like the psalm_plugin can not find the plugins.

To reproduce:

    composer require --dev bamarni/composer-bin-plugin
    composer bin psalm require vimeo/psalm weirdan/doctrine-psalm-plugin
    mkdir src
    vendor/bin/psalm --init src
    vendor/bin/psalm-plugin

Got output:

```Enabled
-------

 ! [NOTE] No plugins enabled

Available
---------

 ! [NOTE] No plugins available
```

Expected output:

```
Enabled
-------

 ! [NOTE] No plugins enabled

Available
---------

 ------------------------------- ------------------------------------
  Package                         Class
 ------------------------------- ------------------------------------
  weirdan/doctrine-psalm-plugin   Weirdan\DoctrinePsalmPlugin\Plugin
 ------------------------------- ------------------------------------
```

weirdan:
If you enable the plugin manually (by editing the `psalm.xml` file), does plugin work? In other words, is it limited to just `psalm-plugin`, or plugins are fully unusable in this setup?

gjuric:
It looks like it can't be found:

```
Scanning files...
PHP Fatal error:  Uncaught UnexpectedValueException: Weirdan\DoctrinePsalmPlugin\Plugin is not a known class in {removed}/vendor-bin/psalm/vendor/vimeo/psalm/src/Psalm/Config.php:985
Stack trace:
#0 {removed}/vendor-bin/psalm/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(458): Psalm\Config->initializePlugins(Object(Psalm\Internal\Analyzer\ProjectAnalyzer))
#1 {removed}/vendor-bin/psalm/vendor/vimeo/psalm/src/psalm.php(563): Psalm\Internal\Analyzer\ProjectAnalyzer->check('{removed}...', false)
#2 {removed}/vendor-bin/psalm/vendor/vimeo/psalm/psalm(2): require_once('{removed}...')
#3 {main}

[...]

PR #1717, opened 2019-06-01
Fall back to autoload when loading plugin classes

weirdan:
This allows to use plugins installed alongside Psalm (when it's not installed in the project being analyzed).

This fixes (partially) vimeo/psalm#1710


weirdan:
@gjuric @muglug Please test if this works for you use cases. This implies manual plugin configuration, `psalm-plugin` is yet to be fixed to support global plugins.

muglug:
Thanks!

PR #1718, opened 2019-06-01
Mark passphrase in openssl as nullable

jornane:
In the [openssl_pkey_export](https://github.com/php/php-src/blob/5b7acc64eb6754b3dafad8f58886359072a60356/ext/openssl/openssl.c#L4657) and [openssl_pkey_export_to_file](https://github.com/php/php-src/blob/5b7acc64eb6754b3dafad8f58886359072a60356/ext/openssl/openssl.c#L4572) functions, the passphrase is optional but also nullable.


muglug:
Thanks!

PR #1720, opened 2019-06-01
Don't attempt to deploy phar from forked repos

bdsl:
Only vimeo/psalm should be deploying. 

This makes it possible to get a successful build status from github repos other than **vimeo/psalm**, so as contributors we can more easily test our work before submitting a PR.

Build passed at https://travis-ci.org/bdsl/psalm/builds/540079192

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,7 @@ jobs:
       env: DEPS="high"
       script:
         - bin/build-phar.sh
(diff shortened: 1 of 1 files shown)


muglug:
Thanks, sorry about that

bdsl:
Don't worry about it, vimeo/psalm is of course the main use case for the travis config, it's a bonus if it also works for forks.

weirdan:
Great minds think alike: https://github.com/weirdan/psalm/commit/cc68343205fef4969a8555f92dff5c2f1bf1641a

PR #1721, opened 2019-06-02
Simplify build/sign/deploy

weirdan:
Even though @bdsl has fixed the issue with phar deploys on third-party repos (my original motivation for working on this), I still think the changes here make build's easier to understand. 

Deployments are not tested though, so this may require some further tweaking.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,22 +62,27 @@ jobs:
       script:
         - vendor/bin/phpcs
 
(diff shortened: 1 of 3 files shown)


muglug:
Thanks!

muglug:
If anything goes wrong I may revert, but it should be fairly obvious what's up

PR #1724, opened 2019-06-02
Namespace autoloaded functions

weirdan:
Since functions defined in files referenced in 'files' section in composer autoload config are automatically loaded when composer autoloader is included (for example when Psalm is installed into project's vendors), it's good idea to keep them namespaced. Otherwise it would prevent dependents to declare their own functions in the global namespace with the same names.


muglug:
Thanks!

PR #1726, opened 2019-06-03
Renamed `Fn` to `TFn` to allow for PHP 7.4 compatibility

Ocramius:
Note: this is a BC break, so I'm not sure if it requires a major release


muglug:
Thanks! Very unlikely that anyone was using Fn explicitly

zerkms:
> Very unlikely that anyone was using Fn explicitly
and if they did - their code would have been broken during 7.4 upgrade nevertheless.

PR #1727, opened 2019-06-03
Fix whitespace

BackEndTea:
There was an extra space that shouldn't be there.


muglug:
Thanks!

BackEndTea:
Thank you @muglug

PR #1745, opened 2019-06-06
Require stable version of openlss/lib-array2xml

jaikdean:
`openlss/lib-array2xml` now has a stable 1.0.0 version.


muglug:
Thanks!

PR #1748, opened 2019-06-06
Add executable PHP files to PHAR distribution

bdsl:
Previously a user of the phar distribution would have to invoke psalm as
`vendor/bin/psalm.phar`. This is different to the command given in
the psalm documentation, `vendor/bin/psalm`

I also copied all files from the psalm/phar repo into
assets/psalm-phar, so that development can be concentrated in this repo.

The travis-deploy-phar.sh should copy any changes made back into the
psalm/phar.git repo.

### DIFF
--- a/assets/psalm-phar/composer.json
+++ b/assets/psalm-phar/composer.json
@@ -1,0 +1,9 @@
+{
+	"name": "psalm/phar",
+	"description": "Composer-based Psalm Phar",
+	"license": ["MIT"],
+	"require": {
+		"php": "^7.0"
+	},
+	"bin": ["psalm.phar", "psalm", "psalter", "psalm-language-server", "psalm-plugin"]
+}
--- a/assets/psalm-phar/composer.json
+++ b/assets/psalm-phar/composer.json
@@ -3,7 +3,7 @@
 	"description": "Composer-based Psalm Phar",
 	"license": ["MIT"],
 	"require": {
-		"php": "^7.0"
+		"php": "^7.1"
 	},
 	"bin": ["psalm.phar", "psalm", "psalter", "psalm-language-server", "psalm-plugin"]
 }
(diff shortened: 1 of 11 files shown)


bdsl:
If this PR is merged I would like to also replace `vimeo/psalm` with `psalm/phar` in the readme, as I think for most users the phar is a better way to install Psalm. https://github.com/vimeo/psalm/blob/326249e13b7e32586b0f2acac319cb39aeec7038/README.md#L25

bin/travis-deploy-phar.sh
```diff
@@ -1,11 +1,15 @@
...
 git clone https://${GITHUB_TOKEN}@github.com/psalm/phar.git > /dev/null 2>&1
-cp build/psalm.phar build/psalm.phar.asc phar/
 cd phar
+rm -rf !(".git")
+cp ../build/psalm.phar ../build/psalm.phar.asc ../assets/psalm-phar/* .
+mv dot-gitignore .gitignore
 git config user.email "travis@travis-ci.org"
 git config user.name "Travis CI"
-git add psalm.phar psalm.phar.asc
+git add --all .
```
bdsl: This means that deleting a file from assets/psalm-phar will automatically delete it from the psalm/phar repo. The latter becomes purely a deployment medium.

bdsl:
It might also be a good idea in a later PR to automatically edit the readme for the phar distribution to add the Psalm version number, or branch and commit ID when not build a tag. Could also add the build date, and maybe copy the docs directory into the phar repo.

assets/psalm-phar/composer.json
```diff
@@ -0,0 +1,9 @@
+{
+	"name": "psalm/phar",
+	"description": "Composer-based Psalm Phar",
+	"license": ["MIT"],
+	"require": {
+		"php": "^7.0"
```
weirdan:
```suggestion
		"php": "^7.1"
```
because main repo requires that.

muglug:
Thanks!

PR #1749, opened 2019-06-07
Change vimeo/psalm to psalm/phar in readme

bdsl:
I think for most users the phar is better than the separate PHP files. The phar comes with a very short composer.json file, meaning it won't conflict with any other packages they may require, and the fact that everything is packaged into one file may make it easier for people to concentrate on their own code and the code of the libraries they use, by not looking at Psalm's internal code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Documentation is available on [Psalm’s website](https://psalm.dev/docs), gener
 Install via [Composer](https://getcomposer.org/):
 
 ```bash
-composer require --dev vimeo/psalm
+composer require --dev psalm/phar
 ```
 
 Add a config:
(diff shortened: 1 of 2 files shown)


bdsl:
Probably this shouldn't be merged until after eecd21a7d500a66098c45519b603653887afe652 is included in a release, since people will mostly read the master version of the readme, not the latest released version of the readme.

muglug:
I'm not sure about this.

Currently the one thing the Phar cannot do is run plugins that either inspect PHP Parser nodes or manipulate Psalm types, as those classes are scoped.

If I want to do that then I need to also require a compatible version of `nikic/php-parser`, and then it stops being nicely self-contained.

bdsl:
Ok. In that case maybe worth keeping the readme as is, and just adding a mention of `psalm/phar` as an alternate option to installation.md

bdsl:
Closing since https://github.com/psalm/phar/issues/1 need to be fixed first. I tried to fix it in #1748 but that caused #1758 , so I undid the change with #1763

PR #1750, opened 2019-06-07
Fix syntax error in travis deploy script

bdsl:
This should resolve the build failure https://travis-ci.org/vimeo/psalm/jobs/542700463

See https://unix.stackexchange.com/a/153863/6974

It might also help diagnostics to add `set -x` to the script, enabling output of all command used. I think it would be fine but I'm slightly paranoid that Travis might fail to properly mask the GITHUB_TOKEN

### DIFF
--- a/bin/travis-deploy-phar.sh
+++ b/bin/travis-deploy-phar.sh
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 
 set -e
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

muglug:
Any ideas for how to fix the current issue?

PR #1765, opened 2019-06-09
Create file to list all Psalm types

DaveLiddament:
Adding more documentation for the list of types that Psalm supports. 

Not 100% sure if this is best way of capturing it, maybe splitting the psalm-assert and template of into different lists or tables would be better? Eitherway I think it would be good to capture this information somewhere,

Based on current documentation and also work done in this PR: https://github.com/sebastianbergmann/phpunit/pull/3708/files

### DIFF
--- a/docs/annotating_code/supported_types.md
+++ b/docs/annotating_code/supported_types.md
@@ -27,24 +27,24 @@ Where `<type>` can be one of:
 
 | Type          | Template Type | Template constraint | Assertion | Notes |
 |---------------|---------------|---------------------|-----------|-------|
-| bool          | Y             | N                   | Y         |       |
-| int           | Y             | N                   | Y         |       |
-| string        | Y             | N                   | Y         |       |
-| float         | Y             | N                   | Y         |       |
-| array         | Y             | N                   | Y         |       |
+| bool          | Y             | Y                   | Y         |       |
+| int           | Y             | Y                   | Y         |       |
+| string        | Y             | Y                   | Y         |       |
+| float         | Y             | Y                   | Y         |       |
+| array         | Y             | Y                   | Y         |       |
 | iterable      | Y             | Y                   | Y         |       |
 | object        | Y             | Y                   | Y         |       |
 | callable      | Y             | Y                   | Y         |       |
 | resource      | Y             | Y                   | Y         |       |
-| null          | N             | N                   | Y         |       |
 | numeric       | Y             | Y                   | Y         | int, float, string (as string could represent a number) |
 | scalar        | Y             | Y                   | Y         | int, float, string, bool |
-| never-return  | N             | N                   | Y         | Means function never returns |
+| array-key     | Y             | Y                   | Y         | int or string |
+| (user class)  | Y             | Y                   | Y         |       |
+| never-return  | N             | N                   | N         | Means function never returns |
 | true          | N             | N                   | Y         |       |
 | false         | N             | N                   | Y         |       |
 | empty         | N             | N                   | Y         |       |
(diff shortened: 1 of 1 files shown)


muglug:
I've beefed-up this page to add much more information: https://psalm.dev/docs/annotating_code/docblock_type_syntax

Would you mind trying to figure out how this PR could fit into that page?

borekb:
Just going to note that it's great to see the effort around this. I'm new to Psalm and the structure of the docs seemed a bit unintuitive at times, for example, I had a hard time finding out how Psalm treats string literals and their unions / intersections.

What I like about this PR is that it proposes a clear place where to talk about types. Maybe it could be directly in [`typing_in_psalm.md`](https://github.com/vimeo/psalm/blob/71a184312ac4bd15b8e9da1a823363729c923dc0/docs/annotating_code/typing_in_psalm.md) but it would be great to have a single place for that.

muglug:
> What I like about this PR is that it proposes a clear place where to talk about types

Yeah but there's a lot to say about them. I agree the `typing_in_psalm.md` page is pretty scattershot, though.

docs/annotating_code/supported_types.md
```diff
@@ -0,0 +1,52 @@
...
+
+And also constraining templates: 
+```php
+  @template T of <type> 
+```
+
+Where `<type>` can be one of:
+
+| Type          | Template Type | Template constraint | Assertion | Notes |
+|---------------|---------------|---------------------|-----------|-------|
```
muglug: I'm not sure this table is terribly useful - the takeaway is that value types and bottom types cannot be used as single template constraints - that could probably said in words somewhere.

borekb:
> Yeah but there's a lot to say about them.

That is true :) I'd love to read about topics like how Psalm "ingests" types (type inference? PHP type hints? doc blocks? asserts?), how to define custom types in code vs. doc blocks, what additional type features (on top of PHP) Psalm adds, etc.

DaveLiddament:
@muglug I agree with 

> I've beefed-up this page to add much more information: https://psalm.dev/docs/annotating_code/docblock_type_syntax

Now this is in place I think we can disregard my PR. 

Only a couple of points I'd think about adding to your updated doc (I'm pretty busy today but will raise a PR later in the week)...

Explicitly state what things like `array-key` are. I did work it out eventually but it took me a few minutes before it clicked what it actually meant. 

Also it'd be nice to see somewhere that explains this notation  [@psalm-assert =ExpectedType $actual](https://github.com/sebastianbergmann/phpunit/pull/3708/files#diff-9ae7a972d07df5f73629d5d315bf405aR1578)
I'm probably being a bit slow, but couldn't work it out.

muglug:
> Explicitly state what things like `array-key` are

see https://psalm.dev/docs/annotating_code/type_syntax/scalar_types/#array-key

> Also it'd be nice to see somewhere that explains this notation `@psalm-assert =ExpectedType $actual`

You're not slow - it's weird! I'll add documentation for this.

`@psalm-assert =KnownType` and `@psalm-assert KnownType` is the difference between `if ($foo === $varOfKnownType) {}` and `if ($foo instanceof KnownType) {}`.

In both cases Psalm will warn if `$foo` can never have the type `KnownType`, but Psalm will also warn in the second case if `$foo` is always a `KnownType`. Additionally, the second case can be meaningfully negated when considering the `else` statement.

DaveLiddament:
> see https://psalm.dev/docs/annotating_code/type_syntax/scalar_types/#array-key

Yes that's great. I think there is no need for my PR anymore so I'll close. 

Thanks for updating the docs.

PR #1769, opened 2019-06-12
LSP Completion: Enable completion when whitespace follows known type

joshdifabio:
This PR enables completion when there is whitespace following a known type.

```php
$foo->
      ^ Psalm provides completion suggestions at this offset

$foo
    ->
      ^ Psalm does not currently provide completion suggestions at this offset (this PR fixes this)
```


muglug:
Thanks!

PR #1774, opened 2019-06-12
LS Completion: Add failing test

joshdifabio:
Method calls where a param is a closure containing a return statement seem to break completion. I've added a failing test but I don't know how to fix this.

Note that the first assertion in the new test case passes but the second one fails.

### DIFF
--- a/tests/LanguageServer/CompletionTest.php
+++ b/tests/LanguageServer/CompletionTest.php
@@ -508,4 +508,42 @@ class CompletionTest extends \Psalm\Tests\TestCase
         $this->assertSame(['B\A', '->'], $codebase->getCompletionDataAtPosition('somefile.php', new Position(10, 32)));
         $this->assertSame(['B\A', '->'], $codebase->getCompletionDataAtPosition('somefile.php', new Position(15, 26)));
     }
+
+    /**
+     * @return void
+     */
+    public function testCompletionOnMethodReturnValueWhereParamIsClosure()
+    {
+        $codebase = $this->project_analyzer->getCodebase();
+        $config = $codebase->config;
+        $config->throw_exception = false;
+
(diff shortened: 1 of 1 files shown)


bdsl:
@joshdifabio this might be better as an Issue instead of a PR, since it's probably not desirable to merge anything that makes the build fail, unless the issue is new since the last release and serious enough to mean a new release shouldn't be performed.

You could avoid breaking the build by adding a call to [markTestIncomplete](https://phpunit.readthedocs.io/en/8.2/incomplete-and-skipped-tests.html?highlight=marktestincomplete) before the failing assertion `$this->assertSame(['B\Collection', '->'], $codebase->getCompletionDataAtPosition('somefile.php', new Position(14, 61)));` .

If you do that you'd still probably want to add the issue, and include the issue id or link in the argument passed to markTestIncomplete.

joshdifabio:
I'm not expecting this PR to get merged, it's just the easiest way to demonstrate the issue and provides a branch from which @muglug can potentially branch.

muglug:
Yeah, I don't mind it, thanks! But I'm also gonna close.

muglug:
Added skipped test in f700502

muglug:
and fixed in 0ba535d

muglug:
the search is a miniscule amount slower, but it definitely shouldn't be noticeable

it's searching through an array that looks like
```
start => [end, ...other data]
34 => [39, ...]
18 => [20, ...]
15 => [25, ...]
4 => [30, ...]
1 => [2, ...]
```

and trying to find the element nearest the end. The fix stops it getting stuck on the `18 => [20, ...]` element when trying to locate the completion at position `30`

joshdifabio:
Fantastic, thanks. Completion is getting seriously good now!

PR #1778, opened 2019-06-13
LS Completion: Improve autocomplete suggestions

joshdifabio:
Various improvements to the autocomplete suggestions.

1. Do not suggest static members when gap is `->`
2. Fix some bad logic in `Codebase::getCompletionItemsForClassishThing()` which was repeatedly merging the same items into the list of completion items
3. Sort completion items by visibility first, as protected and private scope members are suggested even when they are out of scope (public, then protected, then private)
4. Set `filterText` to the member name for method and property items so the IDE does not search `public function`, etc., when the user begins typing the member name

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1205,7 +1205,7 @@ class Codebase
                             \LanguageServerProtocol\CompletionItemKind::METHOD,
                             null,
                             null,
-                            null,
+                            (string)$method_storage->visibility,
                             null,
                             $method_storage->cased_name . '()'
                         );
@@ -1227,7 +1227,7 @@ class Codebase
                             \LanguageServerProtocol\CompletionItemKind::PROPERTY,
                             null,
                             null,
-                            null,
+                            (string)$property_storage->visibility,
                             null,
                             ($gap === '::' ? '$' : '') . $property_name
                         );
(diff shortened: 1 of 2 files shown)


muglug:
> Do not suggest static members when gap is `->`

Mind reverting that? Pretty sure it’s valid PHP (sadly)

joshdifabio:
> Mind reverting that? Pretty sure it’s valid PHP (sadly)
It's invalid for properties: https://3v4l.org/03aRQ
And deprecated for several years on method calls: https://3v4l.org/YLG9U

It really sucks to have all of the static methods suggested on `->`; given that it's been deprecated for years, and all we are doing with this change is _not providing suggestions for those static methods_ (not actually breaking analysis), I think it's better not to suggest them. Do you still want me to revert?

muglug:
Nah you have it the wrong way around. `::instanceMethod()` Is deprecated, not `->staticMethod()`: https://3v4l.org/RO8vt. Psalm complains about the first, but not the second.

But yes, I take your point - either way it’s not _good_ code. Your change can stay.

joshdifabio:
> Nah you have it the wrong way around. ::instanceMethod() Is deprecated, not ->staticMethod(): https://3v4l.org/RO8vt. Psalm complains about the first, but not the second.

Of course! Thanks for correcting.

Issue #1788, opened 2019-06-14
Various testing improvments


bdsl:
I'm initially posting this to get some feedback from @muglug and others before I start on any of this or make individual issues.

I'm potentially interested in making various additions to the test pipeline for Psalm, e.g.

1. Restore the coverage collection, and make html coverage reports visible as an aid to writing tests.
2. Add mutation testing using [Infection](https://infection.github.io/)
3. Add a few end-to-end tests, that invoke psalm as a separate process and assert on the return code and output.
4. Test the built phar file before deploying it
5. Maybe add some tests that use big well known PHP codebases as test cases, e.g. Wordpress, Drupal, Symfony, Laravel etc.
6. Possibly some unit testing for smaller parts of the code. Currently I think psalm tests are mostly functional and based on ValidCodeAnalysisTestTrait and InvalidCodeAnalysisTestTrait, which work by passing a string of PHP Code to the FileAnalyzer and asserting on the error message Psalm should emit, and there are relatively few tests that exercise individual functions, classes etc.
7. Consider refactoring tests that use those two traits, so that instead of each string of PHP being a literal string within the test class, it is its own PHP or PHPT file. Possibly this would hurt performance too much to be worthwhile, but the advantage would be that IDEs would assist much more with writing this test code, and we could see at a glance how they analyse it, and if we used PHP it would be trivial to manually run Psalm with any given file. If I was to do this I think the change would be too big to review manually, so instead I would write a script to create all the files and ask @muglug to run it and commit the result.

However for a lot of this I think Travis is too limiting, as it has a 1 hour limit for each job, and it doesn't seem to have any easy way to carry artefacts between jobs, or offer artefacts for download. Instead I'd probably want to use CircleCI, since it supports artefacts, apparently will allow builds of up to 5 hours, and allows free unlimited use of up to three concurrent container instances for open source projects.

Some of these ideas could lead to a full test pipeline that would take multiple hours to run, but I think if it's well organised so that the most useful tests run first and the others only rarely fail that could be manageable.

muglug:
Thanks for these suggestions! I have thoughts:

> Restore the coverage collection, and make html coverage reports visible as an aid to writing tests.

Yeah, will need to migrate to pcov for this (xdebug now too slow)

> Add mutation testing using Infection

Open to this, though not as high a priority as other stuff on this list.

> Add a few end-to-end tests, that invoke psalm as a separate process and assert on the return code and output.

Sounds good

> Test the built phar file before deploying it

[...]

PR #1795, opened 2019-06-15
CircleCI test and build config, including end to end, coverage and mutation testing

bdsl:
As discussed in #1788 , this introduces a config for CircleCI that does several things. To use it you would need to use the 'add project' button at https://circleci.com

To review you should be able to view the builds from my fork, at e.g. https://circleci.com/workflow-run/0db26e28-75d1-4534-b0c2-b87a782d71a6

I've used the CircleCI workflow feature. Each commit triggers seven jobs, or less if something fails before the right hand side:

Screenshot from 2019-06-15 19-51-09

Coverage is measured using pcov, which as @muglug predicted is much quicker than the old xdebug based coverage. An advantage of CircleCI over Travis is that artifacts can be viewed and downloaded directly from the CircleCI website, so the coverage for this commit is at https://187-184879327-gh.circle-artifacts.com/0/home/docker/project/build/phpunit/coverage-html/index.html and findable as so:

Screenshot from 2019-06-15 19-54-25

The phar file is also built for every commit, and for this commit is downloadable from https://188-184879327-gh.circle-artifacts.com/0/home/docker/project/build/psalm.phar

Another advantage of CicleCI is that it can carry files in the 'workspace' between jobs within a workflow, so the coverage report is used in a mutation testing job. However that job fails and I haven't been able to work out why, so for now I've used the shell `||` operator to suppress the error.

I've also added a shell script that downloads PHPUnit and checks out the commit for version 8.2.2 and runs psalm against it, using the baseline file provided by the authors of PHPUnit. This checks that Psalm doesn't crash, and it checks it doesn't unexpectedly start detecting new issues in PHPUnit, but it doesn't check that it doesn't fail to detect new issues. And I'm not sure how to handle the case when it's desired that Psalm should start failing on PHPUnit due to a new check.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
           name: PHPUnit test
           command: php vendor/bin/phpunit --log-junit build/phpunit/phpunit.xml
       - store_test_results:
-          path: build/phpunit/phpunit.xml
+          path: build/
       - store_artifacts:
           path: build/phpunit
       - persist_to_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
           command: composer global require infection/infection
       - run:
           name: Mutation coverage testing
-          command: php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit || echo 'Temporarily ignoring unexplained infection failure'
+          command: php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit --only-covered || echo 'Temporarily ignoring unexplained infection failure'
   phar-build:
     executor: php-72
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
           command: composer global require infection/infection
       - run:
           name: Mutation coverage testing
-          command: php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit --only-covered || echo 'Temporarily ignoring unexplained infection failure'
+          command: php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit --only-covered --threads=2 || echo 'Temporarily ignoring unexplained infection failure'
   phar-build:
     executor: php-72
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
           command: composer global require infection/infection
       - run:
           name: Mutation coverage testing
-          command: php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit
+          command: php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit || echo 'Temporarily ignoring unexplained infection failure'
   phar-build:
     executor: php-72
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,75 @@ jobs:
       - run:
           name: Static analysis
           command: ./psalm
+  test:
+    executor: php-72
+    steps:
+      - attach_workspace:
+            at: /home/docker/project/
+      - run:
+          name: PHPUnit test
+          command: php vendor/bin/phpunit --log-junit build/phpunit/phpunit.xml
+      - store_test_results:
+          path: build/phpunit/phpunit.xml
+      - store_artifacts:
+          path: build/phpunit
+      - persist_to_workspace:
+          root: /home/docker/project/
+          paths:
+            - .
+  coverage:
+    executor: php-72
+    steps:
+      - attach_workspace:
+          at: /home/docker/project/
+      - run:
+          name: PHPUnit test with coverage
+          command: php -dextension=pcov.so vendor/bin/phpunit --log-junit build/phpunit/phpunit.xml --coverage-xml build/phpunit/coverage-xml --coverage-html build/phpunit/coverage-html
+      - store_artifacts:
+          path: build/phpunit
+      - persist_to_workspace:
+          root: /home/docker/project/
+          paths:
+            - .
+  mutation:
+    executor: php-72
+    steps:
+      - attach_workspace:
+            at: /home/docker/project/
+      - run:
+          name: Install infection
+          command: composer global require infection/infection
+      - run:
+          name: Mutation coverage testing
+          command: php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit
+  phar-build:
+    executor: php-72
+    steps:
+      - attach_workspace:
+          at: /home/docker/project/
+      - run:
+          name: Build Phar file
+          command: bin/build-phar.sh
+      - run:
+          name: Smoke test Phar file
+          command: build/psalm.phar --version
+      - store_artifacts:
+          path: build/psalm.phar
+      - persist_to_workspace:
+          root: /home/docker/project/
+          paths:
+            - build/psalm.phar
+  test-with-real-projects:
+    executor: php-72
+    steps:
+      - checkout # used here just for the side effect of loading the github public ssh key so we can clone other stuff
+      - attach_workspace:
+          at: /home/docker/project/
+      - run:
+          name: Analyse PHPUnit
+          command: bin/test-with-real-projects.sh
+      - store_artifacts:
+          path: build/psalm.phar
 
 # Orchestrate or schedule a set of jobs, see https://circleci.com/docs/2.0/workflows/
 workflows:
@@ -41,3 +110,18 @@ workflows:
       - static-analysis:
           requires:
             - install
+      - test:
+          requires:
+            - static-analysis
+      - coverage:
+          requires:
+            - test
+      - mutation:
+          requires:
+            - coverage
+      - phar-build:
+          requires:
+            - test
+      - test-with-real-projects:
+          requires:
+            - phar-build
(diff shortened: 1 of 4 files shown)


bdsl:
87% total line coverage! I think that's a pretty good score, although mutation coverage is probably meaningful.

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+          paths:
+            - .
+  static-analysis:
+    executor: php-72
+    steps:
+      - attach_workspace:
+          at: /home/docker/project/
+      - run:
+          name: Static analysis
+          command: ./psalm || echo 'not sure why this is failing - ignoring for now.'
```
bdsl: This is the output I'm seeing on CircleCI:

```
ERROR: PossiblyUnusedMethod - src/Psalm/Internal/Fork/ForkProcessDoneMessage.php:12:21 - Cannot find any calls to method Psalm\Internal\Fork\ForkProcessDoneMessage::__construct
    public function __construct($data)


ERROR: PossiblyUnusedMethod - src/Psalm/Internal/Fork/ForkTaskDoneMessage.php:12:21 - Cannot find any calls to method Psalm\Internal\Fork\ForkTaskDoneMessage::__construct
    public function __construct($data)


ERROR: UnusedVariable - src/Psalm/Internal/Fork/Pool.php:136:13 - Variable $task_result is never referenced
            $task_result = $task_closure($i, $task_data);


ERROR: UnusedVariable - src/Psalm/Internal/Fork/Pool.php:155:9 - Variable $results is never referenced
        $results = $shutdown_closure();


4 errors found
```

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+    executor: php-72
+    steps:
+      - attach_workspace:
+            at: /home/docker/project/
+      - run:
+          name: Install infection
+          command: composer global require infection/infection
+      - run:
+          name: Mutation coverage testing
+          command: php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit || echo 'Temporarily ignoring unexplained infection failure'
```
bdsl: Output seen here is:
```

     ____      ____          __  _
    /  _/___  / __/__  _____/ /_(_)___  ____
    / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
  _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
 /___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Running initial test suite...

PHPUnit version: 8.2.2

Generate mutants...

Processing source code files: 543
/bin/bash: line 1:   183 Killed                  php -d memory_limit=4G ~/.composer/vendor/bin/infection --coverage=build/phpunit
```

phpunit.xml.dist
```diff
@@ -1,13 +1,15 @@
...
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/8.2/phpunit.xsd"
          bootstrap="vendor/autoload.php"
          backupGlobals="false"
          beStrictAboutCoversAnnotation="true"
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTestsThatDoNotTestAnything="false"
          beStrictAboutTodoAnnotatedTests="true"
-         verbose="true">
+         verbose="true"
+         executionOrder="default"
```
bdsl: I added the `executionOrder` because unless its explicitly specified infection runs tests in random order, and one of them failed. Would be good probably to fix the failure and set it back to random.

phpunit.xml.dist
```diff
@@ -1,13 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/5.7/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/8.2/phpunit.xsd"
```
bdsl: This had to be updated because the 5.7 xml schema doesn't have executionOrder, and infection validates the config against the schema.

bdsl:
Each of the seven jobs shows up as a separate check in the Github UI:

Screenshot from 2019-06-15 20-10-26

bdsl:
Unfortunately you do have to be logged in to CircleCI to find the links to download artefacts. I don't think it matters what account you're logged in to, for publicly readable projects, so probably not a big problem, but it's still unfortunate.

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+executors:
+  php-72:
+    docker:
+      - image: thecodingmachine/php:7.2-v2-cli
+jobs:
+  install:
+    executor: php-72
+    steps:
+      - checkout
+      - run: date "+%F" > /tmp/cachekey; cat composer.json >> /tmp/cachekey
```
bdsl: For non library PHP projects where composer.lock is committed that is normally used as a cache-key. Since composer.lock isn't committed here, and circleci caches never expire, I've used a combination of the date and the contents of the composer.json as the cache key. So when an upstream releases a new version that we can use ever build within the same calendar day will show the change. After that the new version of vendor will be cached.

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+      - restore_cache:
+          keys:
+            - composer-v2-{{ checksum "/tmp/cachekey" }}
+            # fallback to using the latest cache if no exact match is found (See https://circleci.com/docs/2.0/caching/)
```
bdsl: This is from the CircleCi template config file.

bdsl:
This is a good introduction for people not familiar with mutation testing: https://medium.com/appsflyer/tests-coverage-is-dead-long-live-mutation-testing-7fd61020330e

These are the docs for the mutation testing tool I'm running: https://infection.github.io/guide/

.circleci/config.yml
muglug: What does this mean?

bdsl: I don't know. All it means to me is that I couldn't get infection to actually work. I'd be grateful if anyone else wants to take a look and can see what's wrong. The `|| echo` above is a hack to avoid breaking the build.

.circleci/config.yml
muglug: Huh, that is super mysterious. You might want to add --debug-by-line to see if Psalm is hitting any of the lines below `$task_closure`

bdsl: Thanks, let's see what happens:

```suggestion
          command: ./psalm --debug-by-line || echo 'not sure why this is failing - ignoring for now.'
```

bdsl: @muglug this was a reply to the conversation at https://github.com/vimeo/psalm/pull/1795#discussion_r294056727 but GitHub doesn't seem to have tracked that properly. Here's the job with debug by line switched on. Not sure how to use that output though. https://circleci.com/gh/bdsl/psalm/193

bdsl: And the output does include `/home/docker/project/src/Psalm/Internal/Fork/Pool.php:137`, which is where `$task_result` is used after its initialised on 136.

.circleci/config.yml
muglug: OK - I wouldn't prioritise this, as (running infection now locally) it takes bloody ages and I'm in favour of fast feedback.

.circleci/config.yml
muglug: OK how about `./psalm --no-cache`?

.circleci/config.yml
bdsl: Right. At present it's taking about 5 minutes to get this failure on Circle, but I'm sure it would take a lot longer to produce an actual result. If you think other parts of this PR are useful I can delete the Infection part.

bdsl:
Screenshot from 2019-06-15 22-19-38

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+          paths:
+            - .
+  static-analysis:
+    executor: php-72
+    steps:
+      - attach_workspace:
+          at: /home/docker/project/
+      - run:
+          name: Static analysis
+          command: ./psalm --debug-by-line || echo 'not sure why this is failing - ignoring for now.'
```
bdsl:
```suggestion
          command: ./psalm --debug-by-line --no-cache || echo 'not sure why this is failing - ignoring for now.'
```

bdsl:
Trying `--no-cache` now. 'Allow edits from maintainers' is ticked so feel free to add commits to this branch if you want.

bdsl:
Doesn't seem to make any difference: https://circleci.com/gh/bdsl/psalm/213

bdsl:
I've been staring at a screen too long. Think I need to take a break for at least 12 hours.

muglug:
> I've been staring at a screen too long. Think I need to take a break for at least 12 hours.

Sure - I've just added the static-analysis portion of your CircleCI script, to debug the Psalm failures (as they could have wider implications for users of CircleCI)

bdsl:
Right. The two commits I just added here don't seem to have made any difference - infection is still apparently crashing.

muglug:
Ha I figured it out. The `extension_loaded('pcntl')` call (which returns `false` in this image) disables class checks in Psalm. You can reproduce locally with `./psalm --disable-extension=pcntl`

muglug:
Fixed that issue in c4551bd

muglug:
OK - Psalm now passes - feel free to rebase your config, but no hurry

phpunit.xml.dist
maks-rafalko: it's better to use `xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"`

bdsl:
Thanks @muglug . I've rebased on to the latest master and stopped it allowing Psalm to fail.

phpunit.xml.dist
bdsl:
```suggestion
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/8.2/phpunit.xsd"
```

muglug:
Thanks!

PR #1807, opened 2019-06-17
Add PHPCS job to CircleCI workflow

bdsl:
Screenshot from a sample job with a deliberate code style error:

Screenshot from 2019-06-17 22-26-20

Only the first three jobs in the workflow are run in this situation.


muglug:
Thanks!

PR #1831, opened 2019-06-21
Improve autocompletion by using CompletionItem::$detail for type info

iluuu1994:
Currently, autocompletion looks like this in Visual Studio Code:

Screen Shot 2019-06-21 at 20 10 25

As you can see, if the method signature is very long you can't see the method name. To improve this we could use the  `description` property in the `CompletionItem`:

Screen Shot 2019-06-21 at 22 50 32

It's not quite as neat as I'd like (no highlighting, hard to read, signature only visible for one element at a time) but it's better than not seeing the name of the element.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1201,9 +1201,9 @@ class Codebase
                         $method_storage = $this->methods->getStorage($declaring_method_id);
 
                         $completion_item = new \LanguageServerProtocol\CompletionItem(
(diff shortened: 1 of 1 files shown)


iluuu1994:
Let me know if you have a better suggestion.

muglug:
literally just added this locally, thanks!

iluuu1994:
Lol. Thanks for merging!

PR #1833, opened 2019-06-22
Display function signature on multiple lines when it has parameters

iluuu1994:
Before:

Screen Shot 2019-06-22 at 18 35 08

After:

Screen Shot 2019-06-22 at 18 34 18


iluuu1994:
The length can still be an issues for parameters with long types or long return types.

muglug:
Thanks!

PR #1836, opened 2019-06-23
Add parameter placeholders when autocompleting method

iluuu1994:
Screen Shot 2019-06-23 at 14 58 10

Questions:

- [ ] Does this need a fallback? [The LSP spec](https://microsoft.github.io/language-server-protocol/specification) defines a `snippetSupport` flag. I have not checked if most LSP implementations implement snippets.
- [ ] Should we add an option to disable this behaviour? Not everybody might likes this.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -16,6 +16,7 @@ use Psalm\Progress\Progress;
 use Psalm\Progress\VoidProgress;
 use Psalm\Storage\ClassLikeStorage;
 use Psalm\Storage\FileStorage;
+use Psalm\Storage\FunctionLikeParameter;
 use Psalm\Storage\FunctionLikeStorage;
 
 class Codebase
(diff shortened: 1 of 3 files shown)


src/Psalm/Codebase.php
```diff
@@ -1229,8 +1230,28 @@ public function getCompletionItemsForClassishThing(string $type_string, string $
                             null,
                             (string)$method_storage->visibility,
                             $method_storage->cased_name,
-                            $method_storage->cased_name . '()'
+                            $method_storage->cased_name . '(' . implode(
+                                ', ',
+                                array_map(function (FunctionLikeParameter $parameter) {
+                                    /** @var int $i */
+                                    static $i = 0;
```
psalm-shepherd: Left operand cannot be mixed
```
                                    ++$i;
                                      ^^
```

src/Psalm/Codebase.php
```diff
@@ -1229,8 +1230,28 @@ public function getCompletionItemsForClassishThing(string $type_string, string $
...
+                                    ++$i;
+                                    return '${'
```
psalm-shepherd: Right operand cannot be mixed
```
                                        . $i
                                          ^^
```

src/Psalm/Codebase.php
```diff
@@ -1229,8 +1230,28 @@ public function getCompletionItemsForClassishThing(string $type_string, string $
...
+                                        . $i
+                                        . ':'
+                                        . (
+                                            $parameter->type !== null
```
psalm-shepherd: Argument 3 of str_replace expects array<array-key, mixed>|string, Psalm\Type\Union provided with a __toString method
```
                                                ? str_replace(['$', '}'], ['\$', '\}'], $parameter->type) . ' '
                                                                                        ^^^^^^^^^^^^^^^^
```

psalm-shepherd:
Psalm also found errors in other files

```
src/Psalm/Codebase.php:1211:99 - Param #3 is never referenced in this method
```

psalm-shepherd:
Psalm didn’t find any errors!

iluuu1994:
Maybe it would be better to implement `SignatureHelp`. The only thing this PR should do then is to jump into the parentheses of the method.

muglug:
Yeah I prefer `SignatureHelp` - feels like it’s the LSP-preferred way.

iluuu1994:
Cool, I'll look into that.

PR #1839, opened 2019-06-24
Jump between parens when autocompleting method with parameters

iluuu1994:


muglug:
Thanks!

iluuu1994:
You're welcome!

Issue #1837, opened 2019-06-24
Improve performance


muglug:
Running in a single process* (and using Phan's fallback parser), [Phan](https://github.com/phan/phan) is 20% faster than Psalm. Part of that is maybe due to Psalm doing some slightly deeper analysis, but Phan is also architected to be faster in a number of ways.

## Cloning var types when cloning context

A significant (~25%) speedup can be gained by not cloning Context::$vars_in_scope (which contains the known types of variables and properties in a given scope) when branching (e.g. in `if` statements, `foreach` loops).

Turning off this cloning introduces a numbebr of bugs, but it's likely a surmountable problem.

## PHP-Parser node types

Psalm stores them on PHP Parser nodes (which uses a "feature" of PHP allowing arbitrary properties set on nodes), but Phan stores them in its Context, which is a little faster.

Example here: https://gist.github.com/muglug/e3b8bb7dbbec4431c71ae541f61476e6

cc @TysonAndre

\*when checking for dead code, Psalm is faster because it can do the analysis in multiple threads

TysonAndre:
`phan --quick` (plus unused variable detection and redundant condition detection) might be a better comparison to Psalm, since psalm doesn't recurse when union types aren't provided (after the extra checks, it'd be slower). It wouldn't make much of a difference in timing if the codebase was already fully qualified.

Phan does set dynamic properties on `ast\Node` when there was no other choice (search the codebase for `suppress.*PhanUndeclaredProperty`). It might be faster if I reused `ast\Node->children` - I haven't checked.

```php
<?php

function example($x) {
    echo strlen($x);
}
example([]);
```

Dead code detection on multiple processes was something I thought about for Phan but haven't gotten around to trying to implement.

https://gist.github.com/muglug/e3b8bb7dbbec4431c71ae541f61476e6 - Phan stores inferred types in the Context because Phan analyzes the same function more than once when parameter types are missing. Caching it permanently in the `Node` would do the wrong thing the second time the node was analyzed and make it harder to free up that memory.

----------------

## Suggestions for Optimizations

1. Fully qualify global constants and global function calls, especially frequently used ones - e.g. is_string in ExpressionAnalyzer, strtolower in Reconciler, etc. (maybe a ~3% speedup if everything was fully qualified)

    `./phan --automatic-fix -P NotFullyQualifiedUsagePlugin` can do that automatically - there's no way to specify the issue types to fix, but the issue type whitelist in `.phan/config.php` would help.

[...]

PR #1849, opened 2019-06-26
Fully qualify constants and function calls

LeSuisse:
This should give a small performance boost.
Part of #1837.

The change is enforced via phpcs and can be autofixed with phpcbf.


muglug:
Thanks - this is awesome.

iluuu1994:
Do you think this is measurable? A benchmark would be interesting. Psalm is largely CPU-bound but I still doubt this makes a whole lot of a difference. Maybe I'm wrong.

muglug:
@TysonAndre said it’s about a 3% speed up, and I’ve seen similar figures quoted elsewhere.

LeSuisse:
I did not benchmark the change on Psalm itself but you can find some benchmarks here:
 * https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/
 * https://github.com/Roave/FunctionFQNReplacer#benchmark

Even without the perfs improvement this is IMO an interesting change (especially since it can be automated). It removes the temptation of doing ugly things for test purposes like redefining global function in the namespace scope to change the behavior.

iluuu1994:
Fair enough 🙂

PR #1850, opened 2019-06-26
Remove non-working mutation coverage step from CircleCI config

bdsl:
Also simplifying the PHPUnit with coverage step to stop collecting data
required for mutation testing


muglug:
Thanks!

PR #1854, opened 2019-06-27
Fix MethodParamsProvider to accept an empty param list

Daeroni:
If a plugin implementing MethodParamsProviderInterface returns an empty array (ie. the method does not take any parameters), psalm will not use the result and will ultimately throw an exception at https://github.com/vimeo/psalm/blob/f29826b958b3fcb4456f5ea0aa9eefe021b92d1e/src/Psalm/Internal/Codebase/Methods.php#L434


muglug:
Thanks!

PR #1862, opened 2019-06-28
Implement signature help - closes #1841

### LINKED ISSUE #1841: LSP: Signature Help
I've been investigating on how to implement the `SignatureHelp` API. Can you give me a few pointers? It's pretty similar to `Hover`. The main challenge is that `Codebase::getReferenceAtPosition` can't be used because it only records the position of the identifier (excluding the parentheses). Furthermore we need to not only know which method is being called but also at what parameter the cursor is at.

iluuu1994:
- [x] Better tests

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -236,6 +236,13 @@ class FunctionCallAnalyzer extends \Psalm\Internal\Analyzer\Statements\Expressio
                 strtolower($function_id)
             );
 
+            ArgumentMapPopulator::recordArgumentPositions(
+                $statements_analyzer,
+                $stmt,
+                $codebase,
+                $function_id
+            );
+
             if (!$namespaced_function_exists
                 && !$stmt->name instanceof PhpParser\Node\Name\FullyQualified
             ) {
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -236,12 +236,16 @@ class FunctionCallAnalyzer extends \Psalm\Internal\Analyzer\Statements\Expressio
                 strtolower($function_id)
             );
 
-            ArgumentMapPopulator::recordArgumentPositions(
-                $statements_analyzer,
-                $stmt,
-                $codebase,
-                $function_id
-            );
+            if (!$context->collect_initializations
+                && !$context->collect_mutations
+            ) {
+                ArgumentMapPopulator::recordArgumentPositions(
+                    $statements_analyzer,
+                    $stmt,
+                    $codebase,
+                    $function_id
+                );
+            }
 
             if (!$namespaced_function_exists
                 && !$stmt->name instanceof PhpParser\Node\Name\FullyQualified
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "openlss/lib-array2xml": "^1.0",
         "ocramius/package-versions": "^1.2",
         "composer/xdebug-handler": "^1.1",
-        "felixfbecker/language-server-protocol": "^1.3",
+        "felixfbecker/language-server-protocol": "^1.4",
         "felixfbecker/advanced-json-rpc": "^3.0.3",
         "netresearch/jsonmapper": "^1.0",
         "webmozart/glob": "^4.1",
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1143,6 +1143,60 @@ class Codebase
     }
 
     /**
+     * @return array{0: string, 1: int, 2: Range}|null
+     */
+    public function getFunctionArgumentAtPosition(string $file_path, Position $position)
+    {
+        $is_open = $this->file_provider->isOpen($file_path);
+
+        if (!$is_open) {
+            throw new \Psalm\Exception\UnanalyzedFileException($file_path . ' is not open');
+        }
+
+        $file_contents = $this->getFileContents($file_path);
+
+        $offset = $position->toOffset($file_contents);
+
+        list(,, $argument_map) = $this->analyzer->getMapsForFile($file_path);
+
+        $reference = null;
+        $argument_number = null;
+
+        if (!$argument_map) {
+            return null;
+        }
+
+        $start_pos = null;
+        $end_pos = null;
+
+        ksort($argument_map);
+
+        foreach ($argument_map as $start_pos => list($end_pos, $possible_reference, $possible_argument_number)) {
+            if ($offset < $start_pos) {
+                break;
+            }
+
+            if ($offset > $end_pos) {
+                continue;
+            }
+
+            $reference = $possible_reference;
+            $argument_number = $possible_argument_number;
+        }
+
+        if ($reference === null || $start_pos === null || $end_pos === null || $argument_number === null) {
+            return null;
+        }
+
+        $range = new Range(
+            self::getPositionFromOffset($start_pos, $file_contents),
+            self::getPositionFromOffset($end_pos, $file_contents)
+        );
+
+        return [$reference, $argument_number, $range];
+    }
+
+    /**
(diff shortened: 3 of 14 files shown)


iluuu1994:
@muglug I've noticed that the cache is not update to date on the first invocation of the signature help after a change of the source code. This problem goes away when I call `LanguageServer::doAnalysis` [first](https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/LanguageServer/LanguageServer.php#L125). What is the correct fix for this issue?

muglug:
> What is the correct fix for this issue?

I'll be able to spend some time tomorrow on Language Server issues - will find out then!

muglug:
You'll probably want to merge/rebase master - I removed the alias maps, as there was a better way to achieve the same thing.

muglug:
Update: I did the rebase myself

muglug:
> This problem goes away when I call LanguageServer::doAnalysis first

I think that may be the way forward.

I just added an explicit signature number offset which was required to get it working in Sublime Text.

iluuu1994:
What's the reason that the `StatementsAnalyzer` stops analyzing statements when one is incorrect?

https://github.com/vimeo/psalm/blob/220fa31404ccc46a934a5a328388b3d7f19bc383/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php#L574-L576

This causes the argument positions not to be recorded when a previous method call is incorrect.

muglug:
> What's the reason that the `StatementsAnalyzer` stops analyzing statements when one is incorrect?

If an expression is doing something _really_ naughty then it might be ok to stop analysis (because the messages on the code below can be confusing).

What's the particular circumstance here? I've been removing a lot of `return false;` downstream recently to prevent this happening wherever possible.

iluuu1994:
My test case contains multiple calls to foo:

```php
$this->foo();
$this->foo('Foo');
```

The second one won't work.

muglug:
> The second one won't work.

It presumably doesn't work because one of the analysed expressions is returning `false` - if you can track down what's doing that (and that use case indeed doesn't seem like a serious sin) then you can have the method return `null` instead.

muglug:
I see some low-hanging fruit, will push to master shortly.

muglug:
see https://github.com/vimeo/psalm/commit/0e8fa0e73e3e4530ed88d457194689eabec9b905

iluuu1994:
That solves it, thanks @muglug!

iluuu1994:
> Why not for just the signaturehelper request?

Oh of course, that works too.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
```diff
@@ -236,6 +236,13 @@ public static function analyze(
                 strtolower($function_id)
             );
 
+            ArgumentMapPopulator::recordArgumentPositions(
+                $statements_analyzer,
+                $stmt,
```
psalm-shepherd: Argument 2 of Psalm\Internal\Analyzer\Statements\Expression\Call\ArgumentMapPopulator::recordargumentpositions expects PhpParser\Node\Expr\MethodCall|PhpParser\Node\Expr\StaticCall, PhpParser\Node\Expr\FuncCall provided
```
                $stmt,
                ^^^^^
```

psalm-shepherd:
Psalm found some errors

psalm-shepherd:
Psalm didn’t find any errors!

iluuu1994:
This now works for:

* Method
* Static methods
* Functions
* Instantiations

I've encountered some super weird issue where there were `StaticCall` nodes in the AST for `class Foo extends Bar`. The `startFilePos` and `endFilePos` positions of the arguments are off completely. I have no clue what's going on there. Any ideas?

Apart from that issue this should be done.

muglug:
Can you reproduce the weird behaviour in a test?

iluuu1994:
I'll try tomorrow. 🙂 It happens about a dozen times when running Psalm over its own codebase.

iluuu1994:
[This here](https://github.com/vimeo/psalm/pull/1862/files#diff-63bf0c25596421c9e72bd4c70a413599R39) is what I'm talking about. When you comment that piece of code you'll see that there are various tokenizing errors (because as mentioned the position attributes are off).

muglug:
OHH I know what's happening - I create some AST nodes when checking for property initialisation, and those nodes are then analysed as part of that cycle. Should be a simple fix.

iluuu1994:
Ok, I suspected something like that 😃

muglug:
It should behave better now

muglug:
This is fantastic - let me know if there's anything else you need from me

iluuu1994:
Great! I've noticed that the server crashes when a signature help is invoked on native functions (`array_map` for example). I'll fix that, then it should be ready.

iluuu1994:
@muglug Can you tell me how to get the `StatementsAnalyzer` [here](https://github.com/vimeo/psalm/pull/1862/commits/3902bb75a411a1c9e6606737f6da07a671161d17#diff-e264d1221b8ed04b20093a91a2f7650cR297)? That's the last missing piece.

muglug:
> @muglug Can you tell me how to get the `StatementsAnalyzer` [here](https://github.com/vimeo/psalm/pull/1862/commits/3902bb75a411a1c9e6606737f6da07a671161d17#diff-e264d1221b8ed04b20093a91a2f7650cR297)?

I've just updated the function signature being called in the master branch - you'll want to merge/rebase that. But still wrap in a try/catch in case the function can't easily be discovered (I think this workaround won't work if a function is declared inside a class method).

iluuu1994:
Done, seems to work now! 👌

iluuu1994:
It would be nice if the popup was triggered on autocomplete, not sure if that is possible.

muglug:
> It would be nice if the popup was triggered on autocomplete, not sure if that is possible.

I think that's up to the LSP integrator on the editor side.

muglug:
Is this PR ready for review yet?

iluuu1994:
> Is this PR ready for review yet?

Oh yeah sorry.

iluuu1994:
> I think that's up to the LSP integrator on the editor side.

[Intelephense](http://intelephense.net/) can do it so it must be possible.

muglug:
Thanks, this is great!

> It would be nice if the popup was triggered on autocomplete, not sure if that is possible.

Do you mind showing me what that looks like (via a screen recording or similar)

iluuu1994:
This is what I'd like:

Jul-01-2019 22-05-59

This happens with Psalm:

Jul-01-2019 22-10-12

iluuu1994:
Thanks for your help @muglug!

muglug:
> It would be nice if the popup was triggered on autocomplete

That actually happens for me in Sublime Text, trying it out locally (one minor problem is that `<` characters are escaped properly)

file_analyzer

muglug:
But I can reproduce the bug in VS Code

iluuu1994:
Interesting. There must be a difference in how Intelephense does autocompletion and how Psalm does it. I'll see if I can find out.

muglug:
I couldn’t see anything obvious in the JS source, but shouldn’t be too difficult to see the back-and-forth

iluuu1994:
```log
[Trace - 11:06:55 PM] Received response 'textDocument/completion - (47)' in 1ms.
Result: {
    "isIncomplete": false,
    "items": [
        {
            "kind": 2,
            "label": "functionExists",
            "detail": "functionExists(StatementsAnalyzer $statements_analyzer, string $function_id)",
            "sortText": "functionExists",
            "textEdit": {
                "range": {
                    "start": {
                        "line": 51,
                        "character": 15
                    },
                    "end": {
                        "line": 51,
                        "character": 28
                    }
                },
                "newText": "functionExists($0)"
            },
            "insertTextFormat": 2,
            "command": {
                "title": "Trigger Parameter Hints",
                "command": "editor.action.triggerParameterHints"
            },
            "data": 3833510146090248
        }
    ]
}
```

Looks like it sends a `editor.action.triggerParameterHints` command. Should be super simple to implement.

muglug:
the unescaped characters in Sublime was fixed in https://github.com/tomv564/LSP/pull/608

muglug:
> Looks like it sends a editor.action.triggerParameterHints command. Should be super simple to implement.

Wonderful - want to do that or should I?

iluuu1994:
Feel free, I won't have time until tomorrow.

muglug:
Boom!

vscode_suggest

PR #1866, opened 2019-06-28
Exit with error status on baseline read errors

mpesari:
This PR changes `psalm` to exit with status 1 on baseline read errors. Current behaviour is to `die()` which exits with status 0.

The current behaviour is risky because, for example, CI jobs do not fail on corrupt baseline files, causing them to pass any code without emitting errors.


muglug:
Thanks!

PR #1867, opened 2019-06-28
Disable progressbar when GitLab CI is detected

mpesari:
This makes job output clean by default on GitLab as well.

https://docs.gitlab.com/ee/ci/variables/predefined_variables.html


muglug:
Thanks!

PR #1871, opened 2019-06-28
Rate limit the progress bar updates

TysonAndre:
Writing to the terminal adds time to the analysis, and can also cause flickering if it's fast enough.
Once per 0.1 seconds should be frequent enough for most use cases.

Previously, if there were 5000 files and it took a minute, this would write to stderr 5000 times. After this, it would write at most 600 times


muglug:
Thanks!

PR #1874, opened 2019-06-29
Fix redundant casts/conditions in psalm

TysonAndre:
This includes a check for array_map that was accidentally an assignment


muglug:
Thanks!

PR #1890, opened 2019-07-01
Fix cal_from_jd function return types

vadim2991:


psalm-shepherd:
Psalm found errors in other files

```
src/Psalm/CodeLocation/Raw.php:8:7 - Class Psalm\CodeLocation\Raw is never used
```

muglug:
Thanks!

PR #1898, opened 2019-07-04
Remove unused variables

TysonAndre:


muglug:
Thanks!

iluuu1994:
Shouldn't we adjust the config so that this is caught by CI?

muglug:
They should have been caught in CI - I need to fix three bugs: #1899, #1900, #1901

iluuu1994:
Ah I see

muglug:
and those are now fixed

Issue #1524, opened 2019-04-02
Make relative paths in config resolve from the config file


Taluu:
I have the following directory organization (Google App Engine things, IIUC) :

```
./.git
./cloudbuild.yml
./psalm.xml
./app/src/...
./app/vendor/...
./app/bin/...
...
```

So in my config, I have declared the paths to be in `app/src`, `app/bin`, .. and so on. So far so good.

But if I run from the `app` directory `psalm -c ../psalm.xml`, it doesn't analyze anything and stops with the error "cannot find the path to ../app/app/src" or something along those lines.

So it seems that the xml config resolves the path with the current working directory and not the config file location.

bdsl:
I've had a go at fixing this. I don't think it would be good to simply stop resolving from the CWD and start resolving from the file location, as people may be relying on the current behaviour. So instead I would introduce a new parameter in the config file, `resolveFromConfigFile`. 

If that's set to 'true' then all paths would be resolved from the config file, if not they'd continue to be resolved from the current working directory. I set it to true in the files psalm uses as templates for new installations.

The problem with this approach is that some things may now need to have tests run with both `resolveFromConfigFile='true' `and `resolveFromConfigFile='false' `

What I've done so far is at https://github.com/vimeo/psalm/compare/master...bdsl:resolveFromConfigFile . It seems to fix this for the `psalm` command, but more work will be needed for `psalter` and `psalm-language-server`, as well as probably some tests and code cleanup.

bdsl:
@muglug Do you think adding a new config option is a good way to fix this?

muglug:
Yeah, I think a config option is ideal. Maybe it can be a breaking change in v4 to use the new flag by default.

bdsl:
Ok, I'll keep working on this and try and get it ready to PR in the next day or two then. If we want to make a breaking change in v4 we could deprecate not using the config option, and output a warning for everyone that doesn't use it, or people who don't use and have the config in a directory other than CWD. Maybe even an option to automatically add it to config.

bdsl:
@Taluu This should be fixed in [348](release https://github.com/vimeo/psalm/releases/tag/3.4.8). Apologies for the delay. Do you want to give it a go?

Taluu:
I upgraded but didn't take the time to test that yet. I'll try in the next few days, thanks !

PR #1910, opened 2019-07-06
Allow resolving directories from config file location (continued)

bdsl:
I think this PR with the previous one should now provide a good resolution of #1524. 

I'm still not entirely in love with `resolveFromConfigFile` as an attribute name, but I'm not sure what would be better. @muglug please feel free to rename it before the release.

I also did a bit of refactoring that hopefully makes some of the relevant parts of the code easier to read.


muglug:
Thanks!

Issue #1913, opened 2019-07-07
Test for plugin with absolute filename fails on Windows / Appveyor


bdsl:
Error message from [AppVeyor console](https://ci.appveyor.com/project/muglug/psalm/builds/25801733/job/f83i0rhj1xl5d6e2):
```

1) Psalm\Tests\Config\PluginTest::testPluginFilenameCanBeAbsolute
InvalidArgumentException: Cannot find plugin file C:\projects\php-project-workspace\C:\projects\php-project-workspace\tests\Config/../../examples/plugins/StringChecker.php
C:\projects\php-project-workspace\src\Psalm\Config.php:974
C:\projects\php-project-workspace\src\Psalm\Config.php:839
C:\projects\php-project-workspace\src\Psalm\Config.php:582
C:\projects\php-project-workspace\tests\Config\PluginTest.php:835

```

I don't have a windows machine myself, so I'm not sure if this is specific to the AppVeyor set up or would affect all windows environments. Also not sure whether it's a problem with the test or if this is a bug in Psalm.

bdsl:
I've submitted PR https://github.com/vimeo/psalm/pull/1914  to stop this failing the build and link to this issue from the test output.

muglug:
These tests also fail on my mac at home, but not (for whatever mysterious reason) on my mac at work.

weirdan:
I think `isAbolutePath` just fails on paths containing the drive letter.

muglug:
It was a bug in Psalm, fixed with https://github.com/vimeo/psalm/commit/3d4adee6b332aa3408a934e4c2c7871364e20fcd

PR #1917, opened 2019-07-07
Allow backslash after drive name

weirdan:
Before, paths like `C:/path/to/something` were considered absolute, but `C:\path/to/something` were not

Refs vimeo/psalm#1913

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -16,7 +16,7 @@ function isAbsolutePath($path)
     $regex = '%^(?<wrappers>(?:[[:print:]]{2,}://)*)';
 
     // Optional root prefix.
-    $regex .= '(?<root>(?:[[:alpha:]]:/|/)?)';
+    $regex .= '(?<root>(?:[[:alpha:]]:[\\\/]|[\\\/])?)';
 
     // Actual path.
     $regex .= '(?<path>(?:[[:print:]]*))$%';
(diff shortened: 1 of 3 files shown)


bdsl:
Do you want to revert 210ac39d0004334b65675a6aca70b81f7a35a0a0 as part of this PR?

weirdan:
Perhaps, but it needs to pass on Appveyor first

bdsl:
You won't see `testPluginFilenameCanBeAbsolute` pass on AppVeyor as things are, since 210ac39 stops that particular test running in any Windows environment. 

But the latest commits seem to be failing on AppVeyor only for a different reason - Psalm is emitting two DuplicateClass errors when analysing itself there. See https://ci.appveyor.com/project/muglug/psalm/builds/25804844/job/q7c8vk3jle3r0686 . No idea why that happens in AppVeyor but not in Travis or CircleCI.

weirdan:
One reason could be that psalm actually analyzes those files twice: as `path\to\file.php` and `path/to/file.php` (note the path separators). At least it's what it reports.

muglug:
Thanks a I’ll fix those errors separately

weirdan:
Is there a reason why Psalm uses its own `isAbsolutePath()` even though there's `webmozart/path-util` direct dependency (which includes a similar method)?

muglug:
No - feel free to create a PR to fix that

PR #1921, opened 2019-07-08
Deprecated Psalm\isAbsolutPath in favor of Webmozart\PathUtil\Path::isAbsolute

weirdan:
Since this is a part of public API surface, dropping it entirely requires a major release, with a deprecation phase before.


muglug:
Thanks!

PR #1924, opened 2019-07-09
Change signature on setcookie / setrawcookie

tvdijen:
Version 	Description
7.3.0 	An alternative signature supporting an options array has been added. This signature supports also setting of the SameSite cookie attribute.


muglug:
Thanks!

tvdijen:
Very welcome!

PR #1804, opened 2019-06-17
adding php & extension version info to baseline file

SignpostMarv:
as discussed in [twitter thread](https://twitter.com/SignpostMarv/status/1140360049577091073?s=20); 

not sure how to make the info dumping conditional?

PR #1936, opened 2019-07-11
add psalm config attribute re: #1804

SignpostMarv:
* added a psalm config attribute for controlling the default inclusion of php & php extension version info in the error baseline
re: https://github.com/vimeo/psalm/pull/1804#issuecomment-510525056

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -446,6 +446,11 @@ class Config
     /** @var string|null */
     public $error_baseline = null;
 
+    /**
+     * @var bool
+     */
+    public $include_php_versions_in_error_baseline = false;
+
     /** @var string */
     public $shepherd_host = 'shepherd.dev';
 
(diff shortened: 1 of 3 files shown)


src/Psalm/Config.php
```diff
@@ -740,6 +745,11 @@ private static function fromXmlAndPaths(string $base_dir, string $file_contents,
             $config->error_baseline = $attribute_text;
         }
 
+        if (isset($config_xml['includePhpVersionsInErrorBaseline'])) {
+            $attribute_text = (string) $config_xml['includePhpVersionsInErrorBaseline'];
+            $config->include_php_versions_in_error_baseline = $attribute_text === 'true' || $attribute_text === '1';
+        }
+
```
bdsl: I think you can replace the five lines above with one in the `$booleanAttributes` array [above](https://github.com/vimeo/psalm/pull/1936/files#diff-20a62f59764c9aa89c920a98d1b6ec16R658), with no loss of clarity.

SignpostMarv: wasn't too sure about that- have implemented as suggested in 0c523ae

Issue #634, opened 2018-04-01
Qualifies optional keys in object-like arrays


weirdan:
This uses undocumented optional key syntax for object-like arrays.
```php
<?php
namespace N;

/**
 * @psalm-param array{key?:string} $p
 */
function f(array $p): void
{
    echo isset($p['key']) ? $p['key'] : "";
}
```
Fails with 
```
ERROR: InvalidArrayOffset - q.php:9:16 - Cannot access value on variable $p using offset value of 'key', expecting 'N\key'
    echo isset($p['key']) ? $p['key'] : "";
```
Expected: no issues emitted.

Same code in global namespace works fine.

weirdan:
Possibly related to #619

muglug:
Definitely related. Will fix tonight.

PR #1942, opened 2019-07-13
Add docs for optional ObjectLike

ro0NL:
ref #634

github search point me to it :)

as a side note; this doc page is :100: but IMHO too much hidden in the current tree: https://psalm.dev/docs/running_psalm/plugins/plugins_type_system/#arrays

somehow intuitively i always go to https://psalm.dev/docs/annotating_code/typing_in_psalm/ first :thinking:

### DIFF
--- a/docs/running_psalm/plugins/plugins_type_system.md
+++ b/docs/running_psalm/plugins/plugins_type_system.md
@@ -125,7 +125,7 @@ if (true === $first) {
 
 ``` php
 $a = ["a" => 1, "b" => 2]; // is ObjectLike, array{a:int(1),b:int(2)}
-$a = ["a" => null]; // is ObjectLike with optional keys/values, array{a:?int,b?:string}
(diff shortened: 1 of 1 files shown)


docs/running_psalm/plugins/plugins_type_system.md
```diff
@@ -125,6 +125,7 @@ if (true === $first) {
 
 ``` php
 $a = ["a" => 1, "b" => 2]; // is ObjectLike, array{a:int(1),b:int(2)}
+$a = ["a" => null]; // is ObjectLike with optional keys/values, array{a:?int,b?:string}
```
muglug: This is not accurate (because it’s being totally redefined). If this was in a conditional it would be true after the conditional scope was closed.

ro0NL: done

PR #1945, opened 2019-07-13
Fix throws docblock inheritance

smelesh:
This PR fixes throws docblock inheritance that was broken in 3.4.5.


muglug:
Oh wow I’m sorry, thanks for fixing

PR #1947, opened 2019-07-13
Inherit throws docblock from parent class by default

smelesh:
1. Inherit throws docblock when `{@inheritdoc}` is not set.

Today Psalm doesn't inherit `@throws` from parent class unless inheritDoc is set, unlike params and return docblock. In order to keep the same behavior, `@throws` is also inherited by default.

2. Merge inherited `@throws` with a docblock defined in the subclass.

The same behavior as in PHPStorm. When the subclass defines its own exceptions and `@inheritDoc` is set, exceptions documented in parent class are merged to the child.


muglug:
Thanks!

PR #1948, opened 2019-07-16
add php 7.4snapshot

andreybolonin:


bdsl:
So PHPUnit reports 2148 errors when running on 7.4. 

Personally I think it would be a shame to merge this as is, since it causes the red x next to travis in the build results, and you have to click in to the details to see whether it's just the 7.4 builds that fails or anything else. Personally I would probably also want to remove AppVeyor while the builds there are not passing. I think if they're left in then contributors will get use to seeing the Xs and probably not always check what exactly has failed.

But getting Psalm working on 7.4 well in advance of the 7.4 release is a very good aim. Maybe set up travis so that the PHPUnit from 7.4 is available to view for anyone working on that, but doesn't break the build? Or only breaks the build if the number or errors or failures increases?

weirdan:
> So PHPUnit reports 2148 errors when running on 7.4.

I'm working on getting tests pass on 7.4, and I'm down to 42 errors already.

PR #1957, opened 2019-07-19
Fixed PHP 7.4 related issues

weirdan:
- `s/fn/f/g` (fn is a reserved keyword in 7.4)
- `ReflectionType::__toString()` warning (this method is deprecated in 7.4)

Note that some issues are caused by `phpspec/propecy` code and will be fixed when phpspec/propecy#432 lands. Most were caused by `ReflectionType::__toString()` usage in
`Psalm\Internal\Codebase\Reflection` though.


muglug:
Thanks!

PR #1962, opened 2019-07-22
Mark private properties unused when referenced only in constructor

2e3s:
If a private class property is used only in constructor then it's a dead code since it makes no sense to have the property (unless it's unlocked with a reflection). Usually such situations happen when the property usage has been deleted and the property has been left forgotten.
But such static properties can be accessed between the constructor calls, e.g. as a cache (it would be nice to have a cardinal or at countable `FileReferenceProvider::$method_references_to_class_members` instead of a boolean to catch such cases).

I couldn't think of any case where it wouldn't be true. Is this the right place for the code? Otherwise I'd try to make a plugin for our codebase.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -315,7 +315,6 @@ class Codebase
         );
 
         $this->methods = new Internal\Codebase\Methods(
-            $config,
             $providers->classlike_storage_provider,
             $providers->file_reference_provider,
             $this->classlikes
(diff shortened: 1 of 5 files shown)


muglug:
Thanks, this is awesome - would you mind fixing the one bug in Psalm's own codebase as well? It's just removing this property: https://github.com/vimeo/psalm/blob/d13d37b75d6a60d112ac311db8040fe58bb4ab96/src/Psalm/Internal/Codebase/Methods.php#L39-L42

muglug:
Thanks!

Issue #1951, opened 2019-07-17
DOM extension types


ShiraNai7:
According to [PHP docs](https://www.php.net/manual/en/class.domnode.php#domnode.props), many of the`DOMNode` properties can be `null`, but Psalm assumes that they cannot, resulting in false positives (like `RedundantCondition`).

The `$contextNode` argument of `DOMXpath::query()` and `DOMXPath::evaluate()` does actually accept `NULL`.

https://psalm.dev/r/573bfb6b68

muglug:
would you mind creating a PR? The list of inferred property types is here: https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/PropertyMap.php#L316-L333

ShiraNai7:
I can try :)

ShiraNai7:
Well it looks like this won't be so easy after all. Changing any of the `domnode` properties to `DOMNode|null` causes a loop in `Scanner.php` when running tests. 

```
There was 1 error:

1) Psalm\Tests\AnnotationTest::testPhpStormGenericsWithValidArrayIteratorArgument
Error: Maximum function nesting level of '256' reached, aborting!

src/Psalm/Internal/Type/ParseTree.php:45
src/Psalm/Internal/Type/ParseTree.php:48
src/Psalm/Internal/Type/ParseTree.php:556
src/Psalm/Type.php:161
src/Psalm/Type.php:127
src/Psalm/Internal/Codebase/Scanner.php:324
src/Psalm/Type/Atomic.php:705
src/Psalm/Type/Union.php:1738
[SNIP]
src/Psalm/Internal/Codebase/Scanner.php:325
src/Psalm/Type/Atomic.php:705
src/Psalm/Type/Union.php:1738
src/Psalm/Internal/Codebase/Scanner.php:325
src/Psalm/Type/Atomic.php:705
src/Psalm/Type/Union.php:1738
src/Psalm/Internal/Codebase/Scanner.php:325
src/Psalm/Internal/Visitor/ReflectorVisitor.php:1016
src/Psalm/Internal/Visitor/ReflectorVisitor.php:278
vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:200
vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:91
src/Psalm/Internal/Scanner/FileScanner.php:96
src/Psalm/Internal/Codebase/Scanner.php:606
src/Psalm/Internal/Codebase/Scanner.php:386
[...]

PR #1965, opened 2019-07-23
Improve DOM extension types

ShiraNai7:
#1951

- `DOMNode::insertBefore()` accepts `null` in place of `$refnode`
- `DOMXPath::evaluate()` and `DOMXPath::query()` accept `null` in place of `$contextnode`
- `DOMDocument::$ownerDocument` is always `null`
- `DOMNode::$attributes` is `null` unless the node is instance of `DOMElement`
- `DOMNode::$parentNode` is `null` for root nodes
- `DOMNode::$firstChild, $lastChild, $previousSibling, $nextSibling` can be `null` if there is no such node
- `DOMNode::$namespaceURI, $baseURI` can be `null` if they aren't defined for the document
- `DOMNode::$prefix` seems to always be a string contrary to what the PHP docs say


ShiraNai7:
I guess it might be better to assume that `parentNode` won't be `null` (unless Psalm knows that it's a `DOMDocument` instance).

muglug:
> I guess it might be better to assume that parentNode won't be null (unless Psalm knows that it's a DOMDocument instance).

I think it's fine - if there are complaints I can allow Psalm to not care about `null` there.

ShiraNai7:
> I think it's fine - if there are complaints I can allow Psalm to not care about `null` there.

I've already reverted it via fixup before the merge 🙈 Should I send another PR or will you change it? `test-with-real-projects ` will not pass though.

PR #1983, opened 2019-08-01
Fix issue-1563: Invalid use of backslash in doc block

### LINKED ISSUE #1563: PHP Notice: Uninitialized string offset
When running Psalm on my project I get the following notice:
```
PHP Notice: Uninitialized string offset: 0 in ...\vendor\vimeo\psalm\src\Psalm\Type.php on line 498
```

As far as I can tell from looking at the source of the error, it seems that it's missing a check for empty strings:
```
if ($parse_tree->value[0] === '"' || $parse_tree->value[0] === '\'') {
```

jeffreyyoo:
Catch case of invalid use of backslash in doc block1564
Fixes #1563

### DIFF
--- a/src/Psalm/Type.php
+++ b/src/Psalm/Type.php
@@ -944,6 +944,12 @@ abstract class Type
                 continue;
             }
 
+            if ($string_type_token[0][0] === '\\'
+                && strlen($string_type_token[0]) === 1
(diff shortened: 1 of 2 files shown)


muglug:
Thanks!

PR #1993, opened 2019-08-09
Allow setLocale to have 0 or null as parameters

### LINKED ISSUE #1988: setlocale second parameter can be 0
I get the following error in my code: 
```
ERROR: InvalidScalarArgument - src/Assert.php:1101:40 - Argument 2 of setlocale expects string, int(0) provided
        $locale = \setlocale(LC_CTYPE, 0);
```

But 0, or `null` is allowed for this function.

From the documentation(https://www.php.net/manual/en/function.setlocale.php): 
* If locale is NULL or the empty string "", the locale names will be set from the values of environment variables with the same names as the above categories, or from "LANG".

* If locale is "0", the locale setting is not affected, only the current setting is returned.

* If locale is an array or followed by additional parameters then each array element or parameter is tried to be set as new locale until success. This is useful if a locale is known under different names on different systems or for providing a fallback for a possibly not available locale.

BackEndTea:
Fixes #1988

### DIFF
--- a/src/Psalm/Internal/CallMap.php
+++ b/src/Psalm/Internal/CallMap.php
@@ -11744,7 +11744,7 @@ return [
 'setcookie\'1' => ['bool', 'name'=>'string', 'value='=>'string', 'options='=>'array'],
 'setLeftFill' => ['void', 'red'=>'int', 'green'=>'int', 'blue'=>'int', 'a='=>'int'],
 'setLine' => ['void', 'width'=>'int', 'red'=>'int', 'green'=>'int', 'blue'=>'int', 'a='=>'int'],
-'setlocale' => ['string|false', 'category'=>'int', 'locale'=>'string', '...args='=>'string'],
+'setlocale' => ['string|false', 'category'=>'int', 'locale'=>'string|0|null', '...args='=>'string'],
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/CallMap.php
```diff
@@ -11744,7 +11744,7 @@
 'setcookie\'1' => ['bool', 'name'=>'string', 'value='=>'string', 'options='=>'array'],
 'setLeftFill' => ['void', 'red'=>'int', 'green'=>'int', 'blue'=>'int', 'a='=>'int'],
 'setLine' => ['void', 'width'=>'int', 'red'=>'int', 'green'=>'int', 'blue'=>'int', 'a='=>'int'],
-'setlocale' => ['string|false', 'category'=>'int', 'locale'=>'string', '...args='=>'string'],
+'setlocale' => ['string|false', 'category'=>'int', 'locale'=>'string|0|null|', '...args='=>'string'],
```
muglug: Would you mind removing the trailing `|`?

BackEndTea: The trailing `|` has been removed

muglug:
Thanks!

BackEndTea:
Thank you @muglug

PR #1904, opened 2019-07-06
Work in progress - resolve directories from config file location

bdsl:
As discussed a while back at https://github.com/vimeo/psalm/issues/1524

Build of ef7005f at https://circleci.com/workflow-run/8cadee31-973a-4e66-a724-c82bf0ae2d13

PR #2001, opened 2019-08-11
Fix setting of current_dir with 'root' command parameter

mitelg:
Hey,
I'm quite new to psalm and I am currently trying to introduce it to [Shopware](https://github.com/shopware/platform). Works good so far, but I had an issue, when I was using psalm for our pre-commit hook. Due to our development setup it is necessary to set the `--root` command parameter, but psalm fails to load the autoloader files.
After some researching I found out, that the `$current_dir` variable was not set to the given root directory, before searching for the composer files. This issue was introduce with #1904 
This PR fixes the issue and works nice in my environment.

### DIFF
--- a/src/psalm.php
+++ b/src/psalm.php
@@ -180,6 +180,20 @@ if (isset($options['root'])) {
 
 $current_dir = (string)getcwd() . DIRECTORY_SEPARATOR;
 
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

PR #2002, opened 2019-08-11
Allow class constant names to begin with lower-case

TysonAndre:
Followup to e58ade5803aacb4ff21eebc7475c75fc3fa4960e

(I assume this is what it should do, at least)


muglug:
Thanks!

PR #2022, opened 2019-08-15
Type $matches as an array if non-zero flags are passed to preg_match()

ShiraNai7:
The `preg_match()` function sets its result into the passed `$matches` reference. This reference is typed as `string[]` by Psalm, but that does not take into account that the type may vary based on the given `$flags`:

- `0` (no flags) - `string[]`
- `PREG_OFFSET_CAPTURE` - `array<array{string, int}>`
- `PREG_UNMATCHED_AS_NULL` - `array<string|null>`
- `PREG_OFFSET_CAPTURE|PREG_UNMATCHED_AS_NULL` - `array<array{string|null, int}>`

This PR changes the call map of `preg_match()` to:

- if `$flags` are `0` or not passed, `$matches` is `string[]`
- otherwise `$matches` is `array`

I know this is not 100% (`0` can be passed in dynamically, for example), but it should cover most real use cases.


src/Psalm/Internal/CallMap.php
```diff
@@ -9954,7 +9954,8 @@
 'preg_filter' => ['null|string|string[]', 'regex'=>'mixed', 'replace'=>'mixed', 'subject'=>'mixed', 'limit='=>'int', '&w_count='=>'int'],
 'preg_grep' => ['array', 'regex'=>'string', 'input'=>'array', 'flags='=>'int'],
 'preg_last_error' => ['int'],
-'preg_match' => ['int|false', 'pattern'=>'string', 'subject'=>'string', '&w_subpatterns='=>'string[]', 'flags='=>'int', 'offset='=>'int'],
+'preg_match' => ['int|false', 'pattern'=>'string', 'subject'=>'string', '&w_subpatterns='=>'string[]', 'flags='=>'0|', 'offset='=>'int'],
```
ShiraNai7: Typing flags as `0` doesn't work, but `0|` does. Is this a bug in `Type::parseTokens()`?

muglug:
Would you mind adding a test for this (maybe in `FunctionCallTest.php`)?

ShiraNai7:
Added a few tests for the various calls

src/Psalm/Internal/CallMap.php
muglug: Yes, at some point I felt that declaring a single value type was useless, but I now see its value in some situations (e.g. here)

muglug:
Thanks!

Issue #1374, opened 2019-02-20
Suppress MixedAssignment when variable is explicitly annotated mixed


rightfold:
The following code triggers MixedAssignment when f() is of type mixed:

```php
/** @var mixed */
$x = f();
```

However, since the variable is explicitly annotated mixed, it might as well be allowed. The real danger is when the variable is not given a type explicitly.

PR #2030, opened 2019-08-17
Allow explicitly-annotated mixed assignments

lhchavez:
This change makes it possible to assign mixed expressions to variables
if an annotation is explicitly added to the variable. This allows the
use of `call_user_func` without needing to suppress issues.

Fixes: #1374


muglug:
Thanks!

PR #2040, opened 2019-08-21
Test on nightly build

BackEndTea:
Nightly is PHP 8, and therefore dependencies can't be installed normally.
So we ignore the platform reqs


muglug:
Thanks!

PR #2043, opened 2019-08-21
Disable progress bar automatically when GitHub Actions is detected

LeSuisse:
https://developer.github.com/actions/creating-github-actions/accessing-the-runtime-environment/#environment-variables


muglug:
Thanks!

Issue #2054, opened 2019-08-24
psalter bc (backwards compatiblity) preserving modes


staabm:
Psalter already supports some safety features, e.g. https://psalm.dev/docs/manipulating_code/fixing/

It would be great to have a safety option `-preserve-bc` or similar which makes sure psalter only adds typehints where it does not technically break bc e.g.
- in final classes it can change private and protected properties/methods
- in non final classes only private things may change
- maybe also consider `@internal` phpdoc
- classes with private constructor may be allowed to change

There are certainly a lot more rules we could think about.

Goal would be to allow automated refactoring of mostly un-typed codebases without breaking the world.

muglug:
Ah it already has `--allow-backwards-incompatible-changes=false`, just not documented - mind creating new issues for the things it doesn't do yet?

staabm:
Tried to describe the separate steps in the linked issues

muglug:
Thanks!

PR #2056, opened 2019-08-24
document psalter --allow-backwards-incompatible-changes=false option

staabm:
Refs https://github.com/vimeo/psalm/issues/2054#issuecomment-524545486


muglug:
Thanks!

PR #2059, opened 2019-08-24
Fix Generator::throw parameter type

kelunik:


muglug:
Seems like that triggered a weird bug - will fix and merge on Sunday. Thanks!

muglug:
Thanks!

Issue #2064, opened 2019-08-25
array_key_exists to also accept `ArrayObject`


zerkms:
I think the stub should be like

```php
/**
 * @psalm-template TKey as array-key
 *
 * @param TKey $key
 * @param array<TKey, mixed>|\ArrayObject<TKey, mixed> $search
 *
 * @return bool
 */
function array_key_exists($key, array $search) : bool
{
}
```

muglug:
Thanks - would you mind PRing?

zerkms:
Yep, I'll have a quick look whether any other similar functions require it as well first.

PR #2066, opened 2019-08-26
Added `ArrayObject` to the `array_key_exists` signature

zerkms:
After a quick check of the documentation and the php sources it looks like `array_key_exists` is implemented to support the `ArrayObject` specifically and in general array functions don't.

Hence a PR just for `array_key_exists`.

Addresses: https://github.com/vimeo/psalm/issues/2064

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreGenericFunctions.php
+++ b/src/Psalm/Internal/Stubs/CoreGenericFunctions.php
@@ -195,7 +195,7 @@ function array_chunk(array $arr, int $size, bool $preserve_keys = false)
  * @psalm-template TKey as array-key
  *
  * @param TKey $key
- * @param array<TKey, mixed> $search
+ * @param array<TKey, mixed>|ArrayObject<TKey, mixed> $search
  *
  * @return bool
  */


muglug:
Thanks!

weirdan:
I'd suggest to revert this. `array_key_exists` on objects is deprecated in 7.4: https://3v4l.org/NPHmO
Also it doesn't work as expected when  STD_PROP_LIST flag is set.

muglug:
Yeah, I agree, sorry!

zerkms:
> Also it doesn't work as expected when STD_PROP_LIST flag is set.

any example?

Nevertheless, is there a way to make function signature conditional to php version? 7.4 is not released and for other 7.x versions the `ArrayObject` is still valid and used (and requires typing).

muglug:
> is there a way to make function signature conditional to php version

for most functions, yes - but some of the array functions are handled in that `CoreGenericFunctions` file, and are therefore a little more brittle.

In this case I think the deprecation makes clear that this is not something that the PHP gods intend to work, and so I'm proactively preventing Psalm from treating it as valid code (so that necessary changes can be made today).

weirdan:
> > Also it doesn't work as expected when STD_PROP_LIST flag is set.
> 
> any example?

https://3v4l.org/IhE67

Issue #1551, opened 2019-04-13
Class constants using other constants are mixed


bugreportuser:
The type of `Clazz::cons2` is `int`, but Psalm infers it as `mixed`.

Example (https://psalm.dev/r/f479e3cc0a):
```php
<?php
const cons1 = 0;

class Clazz {
    const cons2 = cons1;
    const cons3 = 0;
}

echo cons1;
echo Clazz::cons2;
echo Clazz::cons3;
```

Example output:
```
Psalm output (using commit 21fd385):

INFO: MixedArgument - 10:6 - Argument 1 of echo cannot be mixed, expecting string
```

If this is a hard problem, #942 should fix it.

TysonAndre:
I'm not familiar with how psalm implements this.

One way to deal with this is to create a class to represent lazily loaded union type (with a closure that's called to resolve the value in the codebase), and only resolve the actual value of the union type the first time analysis needs to know the union type (which, most of the time, is after all of the classes are parsed)

- The file or line with the constant's value won't be available at the time the constant is first parsed, which is why lazy loading is suggested
- This might be harder for class constants in classes that are already lazily loaded (e.g. `vendor/`)
- https://github.com/phan/phan/blob/master/src/Phan/Language/FutureUnionType.php and https://github.com/phan/phan/blob/1.3.1/src/Phan/Parse/ParseVisitor.php#L1318-L1332 are examples of how that might work - parameter defaults, properties, etc. would also benefit from this

muglug:
> One way to deal with this is to create a class to represent lazily loaded union type

Psalm _sort of_ does this for class constants from other classes - when it can’t immediately infer the type it stores the node, then [tries again in the constant population stage](https://github.com/vimeo/psalm/blob/cd74e227c173e9a258c6623b5daf95e07bc012fd/src/Psalm/Internal/Codebase/Populator.php#L120-L157).

TysonAndre:
> Psalm sort of does this for class constants from other classes - when it can’t immediately infer the type it stores the node, then tries again in the constant population stage.

This approach is suggested for issues such as the below (haven't tested these - this is just for an idea of what I mean)

- Subclasses inheriting a constant
- Multiple layers of indirection

```php
<?php
class A {
    const X = self::Y;
    const Y = 3;
}

class C extends B {
}

const Z = C::X;

class B extends A {
    const Z = self::X;
}
```

bugreportuser:
This is great! Thanks for the hard work!

PR #2069, opened 2019-08-26
Allow class constants to refer to other constants

lhchavez:
This change introduces the TMixedDeferredConstant type, which marks a
type as being a constant that needs to be resolved at a later stage.

Fixes: #1551


lhchavez:
hmm after doing a last-minute edit before uploading, i relized this actually does not solve the issue. i'll need a little bit more time to fix this, unfortunately :S

muglug:
I had my own go in 0b4981f, so I think this can be closed.

lhchavez:
thank you for fixing this, and sorry for not closing this PR before since it obviously was not ready u_u

Issue #2099, opened 2019-09-06
Allow @psalm-assert to propagate member assertions to the caller's context


lhchavez:
I don't know if this is a bug or a feature request, but it would be super helpful to allow method calls that use `@psalm-assert` to propagate their assertions to the caller context.

```php
<?php
  
class A {
  public ?int $x = 0;
  
  /**
   * @psalm-assert !null $this->x
   */
  public function assertProperty() : void {
    if (is_null($this->x)) {
      throw new RuntimeException();
    }
  }
}

$a = new A();
$a->assertProperty();
echo (2 * $a->x);
```

```text
Psalm output (using commit 071fae9):

ERROR: PossiblyNullOperand - 18:11 - Right operand cannot be nullable, got int|null
```

PR #2100, opened 2019-09-06
Allow @psalm-assert to propagate member assertions to the caller's context

lhchavez:
This change injects method call assertions that involve member variables
to the caller context by replacing `$this->` with the lhs of the member
call.

Fixes: #2099


lhchavez:
something to note is that i didn't add support for `@psalm-assert-if-{true,false}`, mostly because i didn't know where to add it ^^;;

muglug:
Thanks! Happy to keep to `@psalm-assert` now - the others are much more niche

PR #2119, opened 2019-09-10
Fix the signature of sodium_crypto_scalarmult_base()

villfa:
`sodium_crypto_scalarmult_base()` is an alias of `sodium_crypto_box_publickey_from_secretkey()`
 so it should have the same parameters.

See:
https://www.php.net/manual/en/function.sodium-crypto-scalarmult-base.php


muglug:
Thanks!

PR #2122, opened 2019-09-11
Fix signatures of \ldap_bind() and \ldap_bind_ext()

LeSuisse:
Both functions accepts a null $bind_rdn parameter to do an
anonymous bind and they also accept a fourth optional
parameter.

References:
https://www.php.net/manual/en/function.ldap-bind.php
https://github.com/php/php-src/blob/php-7.3.9/ext/ldap/ldap.c#L1152
https://github.com/php/php-src/blob/php-7.3.9/ext/ldap/ldap.c#L1209


muglug:
Thanks!

Issue #2118, opened 2019-09-10
query re: xdebug filter


SignpostMarv:
I'm thinking that the xdebug filter (e571c182993b055fef190dfa0607a4271f1059b4) explains why psalm thinks `XDEBUG_FILTER_CODE_COVERAGE` etc. are undefined?

If this is the case, is it possible for psalm to ship with a stub that's only loaded if xdebug was loaded before check is called?

muglug:
Mind creating a stub?

SignpostMarv:
sure :)

just having a tinker here- it's hypothetically possible to get the xdebug constants/functions at runtime, although I'm not sure how that'd impact startup performance of psalm (as it'd be doing runtime reflection while xdebug is active).

SignpostMarv:
explored using source code parsing last night (i.e. automating stub generation), but one tool wasn't running on windows & couldn't make head nor tails of the other; I'll give the docs a look & do it manually.

SignpostMarv:
closed as per #2133

PR #2133, opened 2019-09-12
load an xdebug stub when required, re: vimeo/psalm#2118

SignpostMarv:
couldn't quite figure out how to load the extension conditionally, variable seems to get lost after restart.


SignpostMarv:
> To see if the process was restarted you could use one of `XdebugHelper` methods, e.g. [`getSkippedVersion()`](https://github.com/composer/xdebug-handler#getskippedversion).

@weirdan I didn't catch that last night- guessing that check would go here? https://github.com/vimeo/psalm/blob/849d1e1574b18162d98a4de0a05cec1d592e776f/src/psalm.php#L317

My initial plan was to make `Config::$load_xdebug_stub` a nullable bool, allowing the end-user to prevent the stub from being loaded by specifying `loadXdebugStub="false"`.

i.e.
```php
if (is_null($config->load_xdebug_stub) && '' !== $ini_handler->getSkippedVersion()) {
    $config->load_xdebug_stub = true;
}
```

SignpostMarv:
@weirdan got that sorted in 5613688, ta :)

muglug:
Thanks!

SignpostMarv:
you're welcome :)

PR #2134, opened 2019-09-13
Fix some return values in the language server

jorams:
These requests did not conform to the specification.

Note that I didn't check any other requests, these are just the ones I ran into while trying to use it.


muglug:
Thanks! I'll try to figure out why Psalm didn't catch these.

jorams:
Thanks for merging. The return types of these methods were perfectly specified in the codebase, they just didn't correspond to the spec. :)

PR #2140, opened 2019-09-14
Ensure BC of API, compat with `roave/you-are-using-it-wrong`, completeness of `composer.json`

Ocramius:


Ocramius:
Btw, do you already have github actions enabled for this repo? They are quite a lot faster :+1:

bin/test-with-real-projects.sh
```diff
@@ -13,3 +13,11 @@ git checkout 302faed # release 8.3.2
 composer install
 ~/project/build/psalm.phar --config=.psalm/config.xml --monochrome --show-info=false
 ~/project/build/psalm.phar --config=.psalm/static-analysis.xml --monochrome
+
+cd /tmp/testing-with-real-projects
+
+git clone git@github.com:roave/you-are-using-it-wrong.git
+cd you-are-using-it-wrong
+composer install
```
Ocramius: This seems to run on a PHP 7.2+ circle-ci instance, so it fails:

```

  Problem 1
    - This package requires php ^7.3.0 but your PHP version (7.2.22) does not satisfy that requirement.
```

.travis.yml
```diff
@@ -63,6 +63,20 @@ jobs:
       script:
         - vendor/bin/phpcs
 
+    - stage: Ensure API BC
+      php: 7.3
+      env: DEPS="high"
+      script:
+        - composer require --dev roave/backward-compatibility-check:^4.0.0
```
Ocramius: Won't work due to conflicting deps: will try making a new dir and installing it there

```

Updating dependencies (including require-dev)

  Problem 1

    - Installation request for roave/backward-compatibility-check ^4.0.0 -> satisfiable by roave/backward-compatibility-check[4.0.0].

    - Can only install one of: phpdocumentor/type-resolver[1.0.1, 0.4.0].

    - Can only install one of: phpdocumentor/type-resolver[0.4.0, 1.0.1].

    - Can only install one of: phpdocumentor/type-resolver[0.4.0, 1.0.1].

    - Conclusion: install phpdocumentor/type-resolver 0.4.0

    - Installation request for phpdocumentor/type-resolver (locked at 1.0.1) -> satisfiable by phpdocumentor/type-resolver[1.0.1].
```

bin/test-with-real-projects.sh
muglug: Feel free to bump the version it runs on!

.travis.yml
muglug: Yeah, this should just be on circleci I think

muglug:
Thanks!

Ocramius:
@muglug is it OK even though it is red?

muglug:
If it's red now I'll figure out how to make it green in time for release!

Also Psalm now picks up two errors:

```
ERROR: UndefinedDocblockClass - tests/ProxyManagerBench/LazyLoadingValueHolderPropertyAccessBench.php:203:22 - Docblock-defined class or interface ProxyManagerBench\RealObjectType does not exist
     * @psalm-return RealObjectType&ValueHolderInterface<RealObjectType>&VirtualProxyInterface


ERROR: MixedMethodCall - tests/ProxyManagerTest/Functional/MultipleProxyGenerationTest.php:89:49 - Cannot call constructor on an unknown class
            self::assertInstanceOf($proxyClass, new $proxyClass(), 'Proxy can be instantiated via normal constructor');
```

Ocramius:
```
ERROR: UndefinedDocblockClass - tests/ProxyManagerBench/LazyLoadingValueHolderPropertyAccessBench.php:203:22 - Docblock-defined class or interface ProxyManagerBench\RealObjectType does not exist
     * @psalm-return RealObjectType&ValueHolderInterface<RealObjectType>&VirtualProxyInterface
```

`RealObjectType` should be the templated type

```
ERROR: MixedMethodCall - tests/ProxyManagerTest/Functional/MultipleProxyGenerationTest.php:89:49 - Cannot call constructor on an unknown class
            self::assertInstanceOf($proxyClass, new $proxyClass(), 'Proxy can be instantiated via normal constructor');
```

This is possibly a new inspection, right?

muglug:
No – it's one that should have run before, but wasn't, due to a bug

Ocramius:
I'm fine with suppressing or working around that one, since the set of known classes is known to maintainers: will do so in the downstream project. It does indeed make sense to prevent `new $unknown` :-)

Issue #2136, opened 2019-09-13
query: re: php constraints


SignpostMarv:
I noticed that doing a dev install of psalm on `php < 7.1.3` can get a little funky (due to the `symfony/process` dev-dependency).

psalm *can* be required on < 7.1.3, but there's also the remote possibility of a version constraint in the unlikely event someone is requiring psalm but also has `symfony/process < 4` required.

I've got a pair of branches for two possible resolutions to this; one that nudges up the version constraint from `php:^7.1` to `php:^7.1.3`, and another that adds adjusts the `symfony/process` constraint to enable dev install on < 7.1.3.

Both of these options may be moot, given that 7.1 goes EOL at the end of december & could be resolved then by bumping releasing a new major version that goes from `php:^7.1` to `php:^7.2`.

weirdan:
> psalm can be required on < 7.1.3, but there's also the remote possibility of a version constraint in the unlikely event someone is requiring psalm but also has symfony/process < 4 required.

Just like with any other dependency. If the tool you need conflicts with your project dependencies you either use phar bundle or an isolated composer environment (either global or something like `bamarni/composer-bin-plugin`).

> I've got a pair of branches for two possible resolutions to this; one that nudges up the version constraint from `php:^7.1` to `php:^7.1.3`, and another that adds adjusts the `symfony/process` constraint to enable dev install on < 7.1.3.

On a quick glance it appears it should be safe to allow older `symfony/process`, like 3.4.

SignpostMarv:
@weirdan so I guess the remaining question is re: EOL mootness?

muglug:
Feel free to PR a version bump to 7.1.3!

PR #2141, opened 2019-09-14
symfony/process:^4.3 requires php:^7.1.3, adjust dependency constraint to prevent install on < 7.1.3

SignpostMarv:
re: #2136

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ stages:
   # - Code coverage analysis
 
 php:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,6 +8,7 @@ stages:
   # - Code coverage analysis
 
 php:
+  - 7.1.3
   - 7.1
```
Ocramius: Probably want to remove this?

PR #2189, opened 2019-10-01
Fix the arguments order of rar_entry_get()

villfa:
Doc: https://www.php.net/manual/en/rararchive.getentry.php
Source code: https://github.com/cataphract/php-rar/blob/957fc77/rararch.c#L677


muglug:
Thanks!

PR #2194, opened 2019-10-01
Disable composer plugins for end-to-end tests

weirdan:
When Psalm is installed with `composer global require vimeo/psalm` it brings in `ocramius/package-versions`, effectively making it a global composer plugin (so it runs for all `composer install`s, even for
totally unrelated projects).

However `ocramius/package-versions` has a peculiar quirk: it fails when there's no `composer.lock` present, and there's no `composer.lock` for projects that have no dependencies (like the end-to-end test projects).

The net result is that end-to-end tests always fail wherever Psalm is installed globally.

This PR fixes it by disabling composer plugins for end-to-end tests.


muglug:
Thank you! I had always wondered why those test would fail one computer but pass on another.

PR #2210, opened 2019-10-04
allowing null as type for mysqli real_connect for host and passwd parameters

### LINKED ISSUE #2202: Mysqli real_connect signature issues
If `null` is passed to the first or third parameter of the `mysqli` `real_connect` method (https://www.php.net/manual/en/mysqli.real-connect.php), Psalm complains that `null` cannot be passed:

>Argument 1 of mysqli::real_connect cannot be null, possibly null value provided
>Argument 3 of mysqli::real_connect cannot be null, possibly null value provided

According to the docs, both parameters can accept `null`:

> host
>    Can be either a host name or an IP address. Passing the NULL value or the string "localhost" to this parameter, the local host is assumed. When possible, pipes will be used instead of the TCP/IP protocol.

> passwd
>    If provided or NULL, the MySQL server will attempt to authenticate the user against those user records which have no password only. This allows one username to be used with different permissions (depending on if a password as provided or not).

(I believe for the `passwd` argument the docs mean "If **not** provided or NULL" but that's not important here)

jaydiablo:
Fixes #2202.

Modifying the `CallMap.php` file to allow `null` type for the `host` and `passwd` parameters of `mysqli::real_connect` and `mysqli_real_connect`.

I also added tests for both. Not 100% sure that's where these sorts of tests belong, if they belong in a different test file, let me know and I'll move them.


jaydiablo:
Regarding the CircleCI failures:

>The method PHPUnit\Framework\TestCase::expectExceptionMessageRegExp has been marked as deprecated

That's not related to this diff specifically, but probably due to a new version of PHPUnit being pulled down that contains this new deprecation.  Here's the commit that deprecates that method:

https://github.com/sebastianbergmann/phpunit/commit/d1199cb2e43a934b51521656be9748f63febe18e

jaydiablo:
Ah, on AppVeyor the tests I added failed due to `mysqli` extension not being installed...

>Psalm\Tests\MethodSignatureTest::testValidCode with data set "mysqliRealConnectMethodAllowsNullParameters" ('<?php\n                    $m...null);')
Psalm\Exception\CodeException: UndefinedClass - src\somefile.php:3:21 - Class or interface mysqli does not exist

Should I look into adding that to AppVeyor or just remove these tests (or is there another solution, perhaps using stubs?)?

muglug:
Thanks! I'll address the appveyor issues separately

PR #2212, opened 2019-10-06
Fix #2201 - remove 4 params stub restriction on variadic array functions

### LINKED ISSUE #2201: Psalm thinks array_diff cannot have more than 4 parameters
This code:

```
<?php

$one = [];
$two = [];
$three = [];
$four = [];
$five = [];

$diff = array_diff($one, $two, $three, $four, $five);

print_r($diff);
```

Outputs this error:

>ERROR: TooManyArguments - 9:9 - Too many arguments for method array_diff - expecting 4 but saw 5

but shouldn't, as `array_diff` can accept a variable number of arguments (at least 2) according to php docs: https://www.php.net/array_diff

https://psalm.dev/r/d4cfc60c1f

Guuzen:


muglug:
Thanks!

PR #2227, opened 2019-10-13
Add impure functions for mcrypt and iterator_apply

2e3s:
These are among the functions whose result isn't needed.
There is also `ob_implicit_flush` [report](https://psalm.dev/r/9b36f33fc7) with `UnusedFunctionCall`, how is that possible for a void function?


muglug:
Yeah, I've made that change in https://github.com/vimeo/psalm/commit/d0a9bf24f8ee2567ebba7da04f850b03f31bd316

PR #2239, opened 2019-10-17
Normalize stringified type names

lhchavez:
This change makes stringified types more normalized. Concretely it sorts
all union types, reconciled types, and sorts the keys within object-like
types.


muglug:
thanks!

PR #2253, opened 2019-10-20
Fix checking if union type with mixed is nullable

### LINKED ISSUE #2249: TypeDoesNotContainType false positive
https://psalm.dev/r/59bab2953b

`filter_input_array(INPUT_POST)  ?? [];`
> ERROR: TypeDoesNotContainType - 3:1 - false|mixed is always defined and non-null

filter_input_arrays return type is not always defined and can be null, if the designated input array is not populated, see the docs: https://www.php.net/manual/en/function.filter-input-array.php

Do I not see something here or is this a false positive?

TysonAndre:
Fixes #2249


muglug:
Thanks!

PR #2254, opened 2019-10-21
Fix ArrayIterator::{uasort,uksort} parameter type

jorams:
The manual specifies these as having the type `callable`. It looks like they were (incorrectly?) once documented to be `string`.


src/Psalm/Internal/Stubs/CoreGenericClasses.php
```diff
@@ -597,7 +597,7 @@ public function ksort() { }
     /**
      * User defined sort
      * @link http://php.net/manual/en/arrayiterator.uasort.php
-     * @param string $cmp_function <p>
+     * @param callback $cmp_function <p>
```
muglug: Thanks - these should instead be `callable`

jorams: Ok, I'll change them. I used `callback` because that's what's used elsewhere in the same file.

PR #2255, opened 2019-10-21
Add method stubs for Ds\Set

mpesari:
Thanks for adding support for ext-ds in 89752b8574993f1f5ff2d5bd88d1b26ba8b9f315. However, the stub file was old and did not include methods for `Ds\Set` so I added them here.

Reference: https://www.php.net/manual/en/class.ds-set.php


muglug:
Thanks!

PR #2274, opened 2019-10-25
docs: change boolean flags to be xs:boolean (instead of xs:string)

dkarlovi:
The schema now better reflects the actual data model, with the more
exact type and the defaults, where known.

Also, the attributes are now separated by type (string vs bool) and sorted by name within their groups.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -21,42 +21,49 @@
             <xs:element name="globals" type="GlobalsType" minOccurs="0" maxOccurs="1" />
         </xs:choice>
 
-        <xs:attribute name="name" type="xs:string" />
-        <xs:attribute name="useDocblockTypes" type="xs:string" />
-        <xs:attribute name="useDocblockPropertyTypes" type="xs:string" />
-        <xs:attribute name="throwExceptionOnError" type="xs:string" />
-        <xs:attribute name="hideExternalErrors" type="xs:string" />
         <xs:attribute name="autoloader" type="xs:string" />
         <xs:attribute name="cacheDirectory" type="xs:string" />
-        <xs:attribute name="allowFileIncludes" type="xs:string" />
-        <xs:attribute name="totallyTyped" type="xs:string" />
-        <xs:attribute name="strictBinaryOperands" type="xs:string" />
-        <xs:attribute name="requireVoidReturnType" type="xs:string" />
-        <xs:attribute name="useAssertForType" type="xs:string" />
-        <xs:attribute name="rememberPropertyAssignmentsAfterCall" type="xs:string" />
-        <xs:attribute name="serializer" type="xs:string" />
-        <xs:attribute name="allowPhpStormGenerics" type="xs:string" />
-        <xs:attribute name="allowCoercionFromStringToClassConst" type="xs:string" />
-        <xs:attribute name="allowStringToStandInForClass" type="xs:string" />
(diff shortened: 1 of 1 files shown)


config.xsd
```diff
@@ -21,42 +21,43 @@
...
         <xs:attribute name="maxStringLength" type="xs:string" />
-        <xs:attribute name="resolveFromConfigFile" type="xs:string" />
-        <xs:attribute name="includePhpVersionsInErrorBaseline" type="xs:string" />
-        <xs:attribute name="loadXdebugStub" type="xs:string" />
-        <xs:attribute name="ensureArrayStringOffsetsExist" type="xs:string" />
-        <xs:attribute name="ensureArrayIntOffsetsExist" type="xs:string" />
+        <xs:attribute name="name" type="xs:string" />
+        <xs:attribute name="serializer" type="xs:string" />
+
+        <xs:attribute name="addParamDefaultToDocblockType" type="xs:boolean" default="false" />
```
dkarlovi: I've used this as a reference to set the type and default:
https://psalm.dev/docs/running_psalm/configuration/#optional-ltpsalm-gt-attributes

Some of these are not mentioned in the docs, guess the docs are lagging behind?

config.xsd
```diff
@@ -21,42 +21,43 @@
...
+        <xs:attribute name="ensureArrayStringOffsetsExist" type="xs:boolean" default="false" />
+        <xs:attribute name="findUnusedCode" type="xs:boolean" default="false" />
+        <xs:attribute name="findUnusedVariablesAndParams" type="xs:boolean" default="false" />
+        <xs:attribute name="forbidEcho" type="xs:boolean" default="false" />
+        <xs:attribute name="hideExternalErrors" type="xs:boolean" default="false" />
+        <xs:attribute name="hoistConstants" type="xs:boolean" default="false" />
+        <xs:attribute name="ignoreInternalFunctionFalseReturn" type="xs:boolean" default="false" />
+        <xs:attribute name="ignoreInternalFunctionNullReturn" type="xs:boolean" default="true" />
+        <xs:attribute name="includePhpVersionsInErrorBaseline" type="xs:boolean" default="false" />
+        <xs:attribute name="loadXdebugStub" type="xs:boolean" /><!-- default is runtime-specific -->
```
BackEndTea: My xsd knowledge is a bit rusty, but i think you can add some kind of 'notes' to these things. 
Generally IDEs will then dislay those on hover or something.

I think this note about it being dependant on run time could be nice there.

dkarlovi: Not sure what you mean, quickly checked and there's nothing like that AFAIK.

Maybe any searchable hints? :)

dkarlovi: This?
https://www.w3schools.com/xml/el_annotation.asp

dkarlovi: Right, it does work, nice!

BackEndTea: Yeah it looks like that was the one i was thinking about, but it looks like that is only for top level. So i guess i miss remembered

dkarlovi: No, it actually does work:
Screenshot from 2019-10-25 12-26-01

BackEndTea: Thats pretty great 😄

dkarlovi: Thanks for the review! <3

muglug:
Thanks!

PR #2280, opened 2019-10-29
Properly warn when a short option is invalid.

TysonAndre:
Previously, Psalm would not warn or exit - it would skip over the arg and
continue with analysis. (e.g. for `psalm -Z`)

### DIFF
--- a/src/psalm.php
+++ b/src/psalm.php
@@ -110,7 +110,7 @@ array_map(
                 );
                 exit(1);
             }
-        } elseif (substr($arg, 0, 2) === '-' && $arg !== '-' && $arg !== '--') {
+        } elseif (substr($arg, 0, 1) === '-' && $arg !== '-' && $arg !== '--') {
             $arg_name = preg_replace('/=.*$/', '', substr($arg, 1));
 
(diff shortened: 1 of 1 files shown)


src/psalm.php
```diff
@@ -110,7 +110,7 @@ function ($arg) use ($valid_long_options, $valid_short_options) {
                 );
                 exit(1);
             }
-        } elseif (substr($arg, 0, 2) === '-' && $arg !== '-' && $arg !== '--') {
+        } elseif (substr($arg, 0, 1) === '-' && $arg !== '-' && $arg !== '--') {
             $arg_name = preg_replace('/=.*$/', '', substr($arg, 1));
 
             if (!in_array($arg_name, $valid_short_options) && !in_array($arg_name . ':', $valid_short_options)) {
```
TysonAndre: The bug fix in this PR was noticed when looking at a check that `in_array()` is always explicitly passed a third argument for $strict.

It doesn't matter now, but might matter if `-0` becomes an option, because `'0' == '0e1'` (strings are cast to numbers to compare if both are numeric)

muglug:
Thanks!

PR #2285, opened 2019-10-30
Add support for Iterator<X> type

enumag:
Also added `IteratorAggregate` since `Traversable` is supported but I don't really care about that. I'll revert that if you want.

Should I add a test somewhere?

### DIFF
--- a/docs/running_psalm/issues.md
+++ b/docs/running_psalm/issues.md
@@ -1106,12 +1106,22 @@ extending all its template params.
 
 ```php
 /**
- * @template-implements IteratorAggregate<int>
+ * @template-implements ArrayAccess<int>
  */
-class SomeIterator implements IteratorAggregate
+class SomeIterator implements ArrayAccess
 {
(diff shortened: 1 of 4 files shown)


muglug:
Would you mind moving the failing test into the valid test set? Because your changes mean it _should_ pass now!

enumag:
Sure I'll look into the failing tests when I get home and try to fix them.

weirdan:
Keep in mind that it'll change the types inferred for codebases that had incomplete params for `Iterator` and some errors suppressed: https://psalm.dev/r/aeb390621b. Previously `Iterator<X>` was interpreted as iterator with keys of type `X` and now it'll become iterator with **values** of type `X`.

muglug:
Yeah, happy for that to happen (for these three fairly obvious cases)

enumag:
Ok, the tests are fixed now. Please review.

muglug:
Thanks!

PR #2286, opened 2019-10-30
Improve return types of explode() with limit

ShiraNai7:
Currently, `explode()` calls with 3 arguments are skipped by `FunctionAnalyzer` and fallback to the default `array<int, string>|false` return type.

https://psalm.dev/r/2fecda20b4

This PR aims to improve this by extending the existing rules by:

- `explode()` with `$limit` >= 0 should always return a non-empty list
- `explode()` with `$limit` < 0 (or dynamic) may return an empty list


muglug:
Thanks!

PR #2293, opened 2019-11-02
Fix nit on union type '0|' in callmap

TysonAndre:


weirdan:
Psalm interprets it as a single-element enum (https://psalm.dev/r/07f7cc0cc9), and I believe this form was used in callmap intentionally.

weirdan:
See https://github.com/vimeo/psalm/pull/2022#discussion_r314274404

PR #2311, opened 2019-11-07
Support comments in ErrorBaseline

### LINKED ISSUE #2310: Undefined property tagName on comment in baseline file
For some reasons (see #2282), I add to add a baseline error on my baseline file so that the jenkins doesn't report a failure on a known error, which doesn't seem to be detected on my env.

So I manually added this error in the baseline, and added a comment so that I would later know why I added this error... and it makes Psalm crash with a PHP Notice, stating that the `DOMComment` doesn't have a `tagName` type.

A quick fix would be adding on the `ErrorBaseline`, around line 108 a check if the tag has a `nodeName` ? I will remove the comment for now.

Taluu:
Fix #2310, I chose to "skip" comment nodes.

### DIFF
--- a/src/Psalm/ErrorBaseline.php
+++ b/src/Psalm/ErrorBaseline.php
@@ -104,8 +104,11 @@ class ErrorBaseline
 
             $files[$fileName] = [];
 
-            /** @var \DOMElement $issue */
             foreach ($file->childNodes as $issue) {
+                if (!$issue instanceof \DOMElement) {
+                    continue;
+                }
+
                 $issueType = $issue->tagName;
 
                 $files[$fileName][$issueType] = [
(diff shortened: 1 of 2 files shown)


src/Psalm/ErrorBaseline.php
```diff
@@ -104,8 +104,11 @@ public static function read(FileProvider $fileProvider, string $baselineFile): a
 
             $files[$fileName] = [];
 
-            /** @var \DOMElement $issue */
             foreach ($file->childNodes as $issue) {
+                if (!$issue instanceof \DOMElement) {
```
Taluu: I hesitated on using `/** @var \DOMNode|\DOMComment $issue */` instead, but I thought as both have `DOMNode` as an ancestor would be better to check that we iterate only on `DOMElement` node.

I can apply that instead if you'd rather do this.

muglug: As long as everything still works, I don't really care :)

PR #2314, opened 2019-11-08
Implement type-checking and forbidding for print

### LINKED ISSUE #2268: Differences between echo and print
I think they should behave the same in these cases...

https://psalm.dev/r/b47c3140cc

jorams:
Fixes #2268 

`print` was basically not considered at all, while echo had quite a number of special cases. This adds full type checking for `print`, and allows it to be forbidden just like all functions.

### DIFF
--- a/tests/FunctionCallTest.php
+++ b/tests/FunctionCallTest.php
@@ -2622,6 +2622,16 @@ class FunctionCallTest extends TestCase
                     array_push();',
                 'error_message' => 'TooFewArguments',
             ],
+            'printOnlyString' => [
+                '<?php
+                    print [];',
+                'error_message' => 'InvalidArgument',
+            ],
+            'printReturns1' => [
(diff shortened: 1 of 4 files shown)


tests/FunctionCallTest.php
```diff
@@ -2622,6 +2622,16 @@ function test() : void {
...
+            'printOnlyString' => [
+                '<?php
+                    print [];',
+                'error_message' => 'InvalidArgument',
+            ],
+            'printReturns1' => [
+                '<?php
+                    (print "test") === 2;',
+                'error_message' => 'TypeDoesNotContainType',
+            ],
```
jorams: Is this a good place for these test cases?

muglug: Yes!

muglug:
Thanks!

Issue #2331, opened 2019-11-13
Referencing a built-in class name in a shape causes crash


weirdan:
https://psalm.dev/r/d6dbc7bfbb
```php
<?php
  /** 
   * @return array{d:Exception}
   */
  function f() {}
```
**Expected**: no crashes
**Actual**:
```
Psalm runner output:

Internal Psalm error on line 47 - /var/www/vhosts/psalm.dev/httpdocs/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php: Could not get class storage for exception
```

PR #2359, opened 2019-11-18
Scan shape properties

weirdan:
This prevents crashes when built-in classes are referenced by shape properties.

Fixes vimeo/psalm#2331


muglug:
Thanks!

PR #2366, opened 2019-11-22
Allow usage of symfony/console v5

LeSuisse:
Also dropped support for symfony/console:^3.3, as the minimal version of PHP supported by Psalm is also supported by symfony/console:^4.0.


andrew-demb:
I think we don't need to drop support ^3.3 version.

I have multiple projects runned on php 7.3 and symfony 3.4, that I can't update yet.

LeSuisse:
Understandable, I added back the ^3.3 version.

muglug:
Thanks!

PR #2373, opened 2019-11-25
Allow resource as type for imagepng argument 2

### LINKED ISSUE #2365: InvalidArgument Argument 2 of imagepng expects null|string, false|resource provided
When I ran psalm on my project, I encountered this message.

```
InvalidArgument - src/xxxxx/xxxxx - Argument 2 of imagepng expects null|string, false|resource provided
        if (!imagepng($img, $handle)) {
```

the code is below.
```
        $handle = tmpfile();
        if (!imagepng($img, $handle)) {
            throw new Exception('Failed to create png.');
        }
        imagedestroy($img);
```

I think the argument 2 of imagepng can accept resource.
https://www.php.net/manual/en/function.imagepng.php

Is this an issue? or how do I run psalm correctly?

k2rn:
Fixed #2365


muglug:
thanks!

PR #2377, opened 2019-11-26
Improve array function list handling

ShiraNai7:
This PR improves `list` type handling when certain array functions are called:

- `array_column()` should return a list unless non-null `$index_key` is given
  - https://psalm.dev/r/aea9d4f962
- `array_pad()` should return a list when padding a list
  - https://psalm.dev/r/18f1ac0993
- `array_chunk()` should return `list<non-empty-array<TKey, TValue>>`if `$preserve_keys` is enabled, `list<non-empty-list<T>>` otherwise
  - https://psalm.dev/r/a196043831
- `array_map()` should return a list when a single list is passed
  - https://psalm.dev/r/a0d423ad92

### DIFF
--- a/src/Psalm/Internal/Type/ArrayType.php
+++ b/src/Psalm/Internal/Type/ArrayType.php
@@ -1,0 +1,59 @@
+<?php declare(strict_types=1);
+
+namespace Psalm\Internal\Type;
+
+use Psalm\Type;
+
+/**
+ * @internal
+ */
+class ArrayType
+{
+    /** @var Type\Union */
+    public $key;
+
+    /** @var Type\Union */
+    public $value;
+
+    /** @var bool */
+    public $is_list;
+
+    public function __construct(Type\Union $key, Type\Union $value, bool $is_list)
+    {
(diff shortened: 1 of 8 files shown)


src/Psalm/Internal/Type/ArrayType.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Psalm\Type;
+
+/**
+ * @internal
+ */
+class ArrayType
+{
+    /**
+     * @return array{key: Type\Union, value: Type\Union, is_list: bool}|null
```
muglug: This is my only quibble to an otherwise excellent PR - a few people have pointed out in the past that arrays use more memory than corresponding objects.

What if this instead returned an instance of `ArrayType` with the properties `$key`, `$value` and `$is_list`?

ShiraNai7: Sure, that would be better

muglug:
Thanks! This cleans up a lot of duplicate code!

PR #2385, opened 2019-11-28
fixed RarArchive signatures

alfredbez:
I am using psalm version `Psalm 3.7.0@a154191922d14dcdde8ef348520a88d735d82304`

`psalm` complains that I don't invoke [`RarArchive::getEntries()`](https://www.php.net/manual/de/rararchive.getentries.php) and [`RarArchive::close()`](https://www.php.net/manual/de/rararchive.close.php) with one parameter, but there are no required parameters.

`app.php` contains the following code:
```php
<?php
$rar = RarArchive::open("/tmp/baz.rar");
$rar->getEntries();
$rar->close();
```
```
abez@abez-XPS-13-9370:~$ vendor/bin/psalm.phar app.php
Scanning files...
Analyzing files...

E

ERROR: TooFewArguments - app.php:3:7 - Too few arguments for method RarArchive::getEntries - expecting 1 but saw 0
$rar->getEntries();


ERROR: TooFewArguments - app.php:4:7 - Too few arguments for method RarArchive::close - expecting 1 but saw 0
$rar->close();


2 errors found

Checks took 1.38 seconds and used 318.374MB of memory
Psalm was able to infer types for 91.6990% of the codebase
```
---
Maybe it would make sense to create one central repository which then could be used by many tools:
* phan has https://github.com/phan/phan/blob/master/src/Phan/Language/Internal/FunctionSignatureMap.php
* phpstan has https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/SignatureMap/functionMap.php
* psalm has https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/CallMap.php

I hit the exact same problem with PHPstan today, see https://github.com/phpstan/phpstan/issues/2625

Phan has the exact same issue:
```
abez@abez-XPS-13-9370:~$ php phan.phar --allow-polyfill-parser app.php
app.php:3 PhanParamTooFewInternal Call with 0 arg(s) to \RarArchive::getEntries(\RarArchive $rarfile) which requires 1 arg(s)
app.php:4 PhanParamTooFewInternal Call with 0 arg(s) to \RarArchive::close(\RarArchive $rarfile) which requires 1 arg(s)
```

What do you think @TysonAndre @ondrejmirtes @muglug ?


ondrejmirtes:
There were some attempts at a central repository but the static analysers aren't 100% compatible (and their creators have different opinions) to use the same signature map. At this point I'd rather wait until PHP 8 which will have most of the stuff in its reflection AFAIK (and should be possible to rely on it).

alfredbez:
Looks like the broken AppVeyor build is not related to this change, since it was broken before as well: https://ci.appveyor.com/project/muglug/psalm/builds/29185574

weirdan:
Apparently `choco search` cannot find php7.3

muglug:
Thanks!

alfredbez:
Oh noes! bc8952c79 reverted this, the problem is now back, I'll create a new PR for that.

PR #2389, opened 2019-11-28
Added WeakReference support

weirdan:
**Docs**: https://www.php.net/manual/en/class.weakreference.php
**RFC**: https://wiki.php.net/rfc/weakrefs


muglug:
Thanks!

Issue #2395, opened 2019-11-29
password_hash() signature/PASSWORD_* constants incorrect in PHP 7.4


jaikdean:
The `PASSWORD_*` constants were changed from `int` to `string` (or `null` in the case of `PASSWORD_DEFAULT`) in PHP 7.4. The [migration guide](https://www.php.net/manual/en/migration74.incompatible.php) states that “applications correctly using [these] will continue to function correctly”. However, this means the signature for `password_hash()` has now changed, with the second argument accepting `string|null` as of PHP 7.4, rather than `int`.

I'm also seeing an issue when using `PASSWORD_DEFAULT`.

```
ERROR: UndefinedConstant - redacted.php:123:57 - Const PASSWORD_DEFAULT is not defined
            $this->password = \password_hash($password, \PASSWORD_DEFAULT);


ERROR: InvalidArgument - redacted.php:152:51 - Argument 2 of password_hash expects int, string(argon2i) provided
            $password = \password_hash($password, \PASSWORD_ARGON2I);
```

PR #2396, opened 2019-11-29
Update password function types for PHP 7.4

ShiraNai7:
Possible fix for #2395


muglug:
Thanks!

ShiraNai7:
@muglug 654a792 broke part the fix though. `PASSWORD_DEFAULT` is `null` since PHP 7.4 so `isset()` won't work with it.

https://psalm.dev/r/0e4386d00b

It works in the playground, but not locally with PHP 7.4:

```
ERROR: UndefinedConstant - test.php:3:28 - Const PASSWORD_DEFAULT is not defined
echo \password_hash('123', PASSWORD_DEFAULT


ERROR: MixedArgument - test.php:3:28 - Argument 2 of password_hash cannot be mixed, expecting int|null|string
echo \password_hash('123', PASSWORD_DEFAULT
```

muglug:
OH I'm dim, sorry

muglug:
Fixed now

PR #2399, opened 2019-11-30
Added SuicidalAutoloader test

weirdan:
The idea behind this is that Psalm should not use project autoloader for its own things. So if we have a project with autoloader and no code, then any project autoloader hit means Psalm failed to load something.

Right now it highlights several issues in CoreGenericClasses stub:
- usage of `callback` instead of `callable`
- `@property-read` not resolving template parameters


weirdan:
I've stumbled upon this when I tried to use Psalm with some of our internal code - the autoloader was using `require` instead of `include`.

muglug:
Thanks! Will fix those issues

muglug:
Fixed in c65540a

PR #2402, opened 2019-12-01
Improve CoreGenericClasses

weirdan:
- general formatting
- Better definition for iterator class for `ArrayObject`
- Better definition for callable parameters to `ArrayObject`/`ArrayIterator` `uksort`/`uasort`
- Fixed signature for `ArrayIterator::current()`
- Fixed signature for `DOMNodeList::item()`

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreGenericClasses.php
+++ b/src/Psalm/Internal/Stubs/CoreGenericClasses.php
@@ -24,8 +24,9 @@ interface IteratorAggregate extends Traversable {
     /**
      * Retrieve an external iterator
      * @link http://php.net/manual/en/iteratoraggregate.getiterator.php
-     * @return Traversable<TKey, TValue> An instance of an object implementing <b>Iterator</b> or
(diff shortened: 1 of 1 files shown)


weirdan:
PHPUnit failure is legit - PHPUnit uses [templated ObjectProphecy](https://github.com/sebastianbergmann/phpunit/blob/1d937a963f5755ad8298c400447f812c65ddc687/src/Framework/TestCase.php#L1825), but Prophecy [has no `@template` tags](https://github.com/phpspec/prophecy/blob/f6811d96d97bdf400077a0cc100ae56aa32b9203/src/Prophecy/Prophecy/ObjectProphecy.php#L26).

muglug:
> PHPUnit failure is legit

Yeah, that bug was revealed by https://github.com/vimeo/psalm/commit/1e3ecb5fd077534af6da42cf51b0b9d59aef227e and I've ticketed here: https://github.com/sebastianbergmann/phpunit/pull/3966

muglug:
Thanks!

PR #2410, opened 2019-12-03
Update PhpStorm usage

leocavalcante:
For https://psalm.dev/docs/running_psalm/language_server/#phpstorm

There is a new option that defaults to Artifact. It should be Executable.
Windows can't figure out that the PHP interpreter will run a shell script that then calls  PHP (not sure if this even works on Unix)
And added a PHP bin dir example for Wins.

Basically I was only able to make it run with:
image

Maybe helps another Windows users.

### DIFF
--- a/docs/running_psalm/language_server.md
+++ b/docs/running_psalm/language_server.md
@@ -37,10 +37,11 @@ When you install the plugin, you should see a "Language Server Protocol" section
 
 In the "Server definitions" tab you should add a definition for Psalm:
 
+ - Select `Executable`
  - Extension: `php`
- - Path: `<path-to-php-binary>` e.g. `/usr/local/bin/php`
+ - Path: `<path-to-php-binary>` e.g. `/usr/local/bin/php` or `C:\php\php.exe`
     - this should be an absolute path, not just `php`
- - Args: `vendor/bin/psalm-language-server`
+ - Args: `vendor/bin/psalm-language-server` (on Windows use `vendor/vimeo/psalm/psalm-language-server`)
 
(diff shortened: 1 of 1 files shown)


docs/running_psalm/language_server.md
```diff
@@ -37,10 +37,11 @@ When you install the plugin, you should see a "Language Server Protocol" section
 
 In the "Server definitions" tab you should add a definition for Psalm:
 
+ - Select `Executable`
  - Extension: `php`
- - Path: `<path-to-php-binary>` e.g. `/usr/local/bin/php`
+ - Path: `<path-to-php-binary>` e.g. `/usr/local/bin/php` or `C:\php\php.exe`
     - this should be an absolute path, not just `php`
- - Args: `vendor/bin/psalm-language-server`
+ - Args: `vendor/bin/psalm-language-server` (on Windows use `vendor/vimeo/psalm/psalm-language-server`)
```
weirdan: Does `composer` install `.bat` file into `vendor/bin` instead of a php file on Windows?

leocavalcante: Exactly. `vendor/bin/psalm-language-server` is a shell script and there is a `.bat` version at `vendor/bin/psalm-language-server.bat`.
Then the LSP plugin will try to execute them using the given PHP interpreter, that just outputs the contents since there is no opening PHP tag `<?php`.
image

muglug:
Thanks!

weirdan:
> (not sure if this even works on Unix)

On Linux it just links files that are specified in the `bin` array in `composer.json`:
image

leocavalcante:
Oh, now I got it.

Issue #1083, opened 2018-11-15
Add a custom PHP error handler/exception handler to add more information for notices/crashes


TysonAndre:
Motivation: This will help add useful information for bug reporting. (E.g. if a user sees a PHP notice such as "undefined index" while running psalm, or a crash)

Something along the lines of https://github.com/phan/phan/blob/1.1.2/src/Phan/Bootstrap.php#L41-L123 may be helpful. You might want to leave out the `exit(1)` part, since Psalm occasionally loads php files, custom autoloaders, etc.

Features to consider

- Mention the most recently parsed/analyzed file (useful if you're trying to figure out why psalm crashes or emits a notice)
- Mention psalm version (useful for maintainers looking at a bug report). Possibly also mention php-parser version.
- Print a backtrace for the notice
- (Optional) Add `--suppress-errors` (not in that example)

TysonAndre:
The error handler I mentioned was also extended to dump variables in the backtrace (including special handling for types/union types, elements, etc). 

This is useful for reading bug reports, etc.

```
/path/to/phan/src/Phan/Language/Type.php:2304 [8] Undefined variable: some_undefined_variable
(Phan 1.3.1 crashed)
#0  phan_error_handler() called at [/path/to/phan/src/Phan/Language/Type.php:2304]
#1  Phan\Language\Type->canCastToAnyTypeInSet() called at [/path/to/phan/src/Phan/Language/UnionType.php:1673]
#2  Phan\Language\UnionType->canCastToUnionType() called at [/path/to/phan/src/Phan/Language/Element/Clazz.php:1512]
#3  Phan\Language\Element\Clazz->addMethod() called at [/path/to/phan/src/Phan/Language/Element/Clazz.php:2389]
#4  Phan\Language\Element\Clazz->importAncestorClass() called at [/path/to/phan/src/Phan/Language/Element/Clazz.php:2184]
#5  Phan\Language\Element\Clazz->importParentClass() called at [/path/to/phan/src/Phan/Language/Element/Clazz.php:2096]
#6  Phan\Language\Element\Clazz->importAncestorClasses() called at [/path/to/phan/src/Phan/Language/Element/Clazz.php:2837]
#7  Phan\Language\Element\Clazz->hydrateOnce() called at [/path/to/phan/src/Phan/Language/Element/Clazz.php:2986]
#8  Phan\Language\Element\Clazz->hydrate() called at [/path/to/phan/src/Phan/CodeBase.php:925]
[...]

PR #2418, opened 2019-12-04
Add an exception handler that prints full exception trace

TysonAndre:
By default, php will only print the first few thousand bytes of the
exception, for an uncaught exception
(I think that's the default, and not just my configuration)

Instead, print all of the parts of the exception.
Also, print the psalm version, to make handling bug reports easier.

For #1083 (this PR does not cover set_error_handler)

Example output:

```
./psalm
Uncaught Exception: test in /path/to/psalm/src/psalm.php:283
Stack trace:
#0 /path/to/psalm/psalm(2): require_once()
#1 {main}
(Psalm 3.x-dev@99db774c99b53279984eeb8026d29c0c72dcafde crashed due to an uncaught Throwable)
```

### DIFF
--- a/src/psalm-language-server.php
+++ b/src/psalm-language-server.php
@@ -9,6 +9,8 @@ gc_disable();
 // show all errors
 error_reporting(-1);
 
+require_once __DIR__ . '/Psalm/Internal/exception_handler.php';
+
 $valid_short_options = [
     'h',
     'v',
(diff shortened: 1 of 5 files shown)


muglug:
This is very good, thanks for implementing (and sorry I was delinquent in doing so)

src/psalm-language-server.php
```diff
@@ -9,6 +9,8 @@
 // show all errors
 error_reporting(-1);
 
+require_once __DIR__ . '/Psalm/Internal/exception_handler.php';
```
staabm: Would it make sense to include this file via composee files-autoloading instead?

PR #2420, opened 2019-12-04
Workaround for uncaught InvalidArgumentException fetching constant of undeclared class

### LINKED ISSUE #2419: Uncaught InvalidArgumentException: Could not get class storage for class constant of undeclared class
Expected: Report an error, but don't crash
Observed: There is an uncaught fatal error

```php
<?php

// Fatal error: Uncaught InvalidArgumentException: Could not get class storage for other in /path/to/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:47
class Request {
    const FIELD = Other::ID;
    const SORT = [self::FIELD => true];
}
```

TysonAndre:
Fixes #2419


muglug:
Thanks!

PR #2434, opened 2019-12-06
fix return type of stream_bucket_new

### LINKED ISSUE #2430: stream_bucket_new has wrong return type
`stream_bucket_new()` returns `object|bool`, but psalm thinks it's returning `resource|bool`.

This is a problem because the return value of `stream_bucket_new()` is often passed into `stream_bucket_append()` which psalm knows takes `object` as its second argument.

This way it's impossible to write a custom  stream filter that passes psalm's `InvalidArgument` check.

I would have made a quick PR, but I'm not sure whether/if this has changed between PHP versions.

pilif:
the function returns `object` since 2003 in
php/php-src@c4a491e12cfb9dbd84d88d04f49e00c4ebf9e289 and
`stream_bucket_append()` is type hinted to take `object` too.

This fixes #2430


muglug:
Thanks!

PR #2447, opened 2019-12-09
ssh2-functions can return false on error, added missing parameters

Daeroni:
All ssh2 functions (except ssh2_poll) can return false on error, also added some missing parameters to them.

Source: http://git.php.net/?p=pecl/networking/ssh2.git;a=tree


muglug:
Thanks!

PR #2492, opened 2019-12-20
Drop non-existent param from docblock

weirdan:
I wonder why psalm didn't catch it.


muglug:
It didn't catch it because Psalm doesn't care, so long as a function's params are all documented adequately.

Issue #2508, opened 2019-12-24
InvalidScalarArgument callable(T of object, T of object)


vudaltsov:
https://psalm.dev/r/eddc5cd072

vudaltsov:
@muglug , it seems the problem still occurs in the class.
https://psalm.dev/r/c7bba7e45e

PR #2524, opened 2019-12-29
Test doctrine collections

weirdan:
Not sure if the fail I'm seeing now indicate a problem with Psalm or collections.


muglug:
I think it’s with collections, at least last time I checked

muglug:
ALSO the templates tests have a lot of reproductions of code that exists in that repo (but it’s still good to have the actual thing to test on)

muglug:
Here's the PR with necessary fixes: https://github.com/doctrine/collections/pull/224/files

Fixing https://github.com/vimeo/psalm/issues/2508 also fixed a few issues in `doctrine/collections`, so I cut a new release with that inside.

PR #2528, opened 2019-12-29
Explain bool assertions

ruudk:


muglug:
Thanks!

docs/annotating_code/assertion_syntax.md
```diff
@@ -64,6 +64,18 @@ This asserts that `$bar` is not an object of type `SomeObjectType`:
...
+## Bool assertions
+
+This asserts that `$bar` is `true`:
+```php
+/** @psalm-assert true $bar  */
+```
+
+This asserts that `$bar` is not `false`:
+```php
+/** @psalm-assert !false $bar  */
```
staabm: Is this compared with == or ===

PR #2538, opened 2020-01-01
Add sample for optional keys in array shapes

mfn:
I found this lacking and only figured it by reading it somewhere else by accident.


muglug:
Thanks!

Issue #2548, opened 2020-01-03
`(int) <bool>` not detected as `int(0)|int(1)`


SignpostMarv:
Typecasting a bool to an int should allow psalm to identify the value as `int(0)|int(1)`, since that's literally all it can be.

I could have a look into PRing this if you could give me a headsup as to where to go poking around in psalm?

```php
<?php
/**
* @param 0|1 $int
*/
function foo(int $int) : void {
    echo (string) $int;
}

foo((int) ((bool) ((int) '1')));
```
> Psalm output (using commit d5d4a18):
> 
> ERROR: ArgumentTypeCoercion - 9:5 - Argument 1 of foo expects int(0)|int(1), parent type int provided

https://psalm.dev/r/10d6fd4bbb (forgot to paste the link in when I copied the php then copied the error output, re: psalm/psalm.dev#7)

SignpostMarv:
@psalm-github-bot go have a peek at psalm/psalm.dev#7 ¬_¬

muglug:
> I could have a look into PRing this if you could give me a headsup as to where to go poking around in psalm?

It's here: https://github.com/vimeo/psalm/blob/0e39759ed63d747b9b78d7f12d68c66c96ba3c8d/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php#L553-L558

muglug:
Now fixed

PR #2550, opened 2020-01-03
Typecast bool to int

SignpostMarv:
re: #2548, not sure which test case to add tests to (if needed) ?

### DIFF
--- a/tests/TypeReconciliation/ValueTest.php
+++ b/tests/TypeReconciliation/ValueTest.php
@@ -656,6 +656,39 @@ class ValueTest extends TestCase
                         private $type = "easy";
                     }'
             ],
+            'typecastTrueToInt' => [
+                '<?php
+                /**
+                * @param 0|1 $int
+                */
+                function foo(int $int) : void {
+                    echo (string) $int;
+                }
+
+                foo((int) true);',
+            ],
+            'typecastFalseToInt' => [
+                '<?php
+                /**
+                * @param 0|1 $int
+                */
+                function foo(int $int) : void {
+                    echo (string) $int;
+                }
+
+                foo((int) false);',
+            ],
+            'typecastedBoolToInt' => [
+                '<?php
+                /**
+                * @param 0|1 $int
+                */
+                function foo(int $int) : void {
+                    echo (string) $int;
+                }
+
+                foo((int) ((bool) 2));',
+            ],
         ];
     }
 
--- a/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
@@ -72,6 +72,7 @@ use function is_array;
 use function array_merge;
 use function array_values;
 use function array_map;
+use function current;
 
 /**
  * @internal
--- a/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
@@ -553,35 +553,24 @@ class ExpressionAnalyzer
         } elseif ($stmt instanceof PhpParser\Node\Expr\Cast\Int_) {
             if (self::analyze($statements_analyzer, $stmt->expr, $context) === false) {
                 return false;
-            } elseif (($stmt->expr instanceof PhpParser\Node\Expr\ConstFetch) &&
-                (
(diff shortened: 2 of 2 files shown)


SignpostMarv:
> `node_data->setType($stmt->expr);`

is this not meant to be `getType($stmt->expr)` ?

muglug:
oops yes, corrected

SignpostMarv:
Does this look about right?

```diff
diff --git a/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php b/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
index 54b9b989..1382c6ac 100644
--- a/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php
@@ -553,34 +553,15 @@ class ExpressionAnalyzer
         } elseif ($stmt instanceof PhpParser\Node\Expr\Cast\Int_) {
             if (self::analyze($statements_analyzer, $stmt->expr, $context) === false) {
                 return false;
-            } elseif (($stmt->expr instanceof PhpParser\Node\Expr\ConstFetch) &&
-                (
-                    ['true'] === $stmt->expr->name->parts ||
-                    ['false'] === $stmt->expr->name->parts
-                )
-            ) {
-                $int = 0;
-
-                if (['true'] === $stmt->expr->name->parts) {
-                    $int = 1;
-                }
-
-                $statements_analyzer->node_data->setType($stmt, Type::getInt(false, $int));
-            } elseif ($stmt->expr instanceof PhpParser\Node\Expr\Cast\Bool_) {
-                if ($stmt->expr->expr instanceof PhpParser\Node\Scalar\LNumber) {
-                    $int = 0;
+            }

-                    if (0 !== $stmt->expr->expr->value) {
-                        $int = 1;
-                    }
+            $maybe_bool = $statements_analyzer->node_data->getType($stmt->expr);

-                    $statements_analyzer->node_data->setType($stmt, Type::getInt(false, $int));
-                } else {
+            if (1 === count($maybe) && $maybe_bool[0] instanceof Type\Atomic\TBool) {
                     $statements_analyzer->node_data->setType($stmt, new Type\Union([
                         new Type\Atomic\TLiteralInt(0),
                         new Type\Atomic\TLiteralInt(1),
                     ]));
-                }
             } else {
                 $statements_analyzer->node_data->setType($stmt, Type::getInt());
             }
```

muglug:
`NodeTypeProvider::getType` returns a union type. `Union::getTypes` returns atomic types, so you need to call both

muglug:
I've just renamed `Union::getTypes` to `Union::getAtomicTypes` to make things a bit more explicit: 9fa2db1

muglug:
Also write tests! I think these could go in `ValueTest` or `AssignmentTest`

SignpostMarv:
> you need to call both

i.e. ? 

```diff
+            $maybe = $statements_analyzer->node_data->getType(
+                $stmt->expr
+            )->getAtomicTypes();
```

muglug:
yes, but also (in rare circumstances) `$maybe` can return `null`, so you need to check for that.

SignpostMarv:
> so you need to check for that.

forgetting how you feel about inline assignments:
```diff
+            if (
+                ! is_null($maybe_type) &&
+                1 === count($maybe = $maybe_type->getAtomicTypes()) &&
+                current($maybe) instanceof Type\Atomic\TBool
+            ) {
```

muglug:
I prefer explicit `!== null` checks and yes, inline assignments are generally to be avoided – especially inside function calls. Easily-scannable code is more maintainable code.

SignpostMarv:
yeah just ignore 698f94e, I had the wrong file open ¬_¬

SignpostMarv:
is a test that catches stuff like `random_int(0, 2)` necessary ?

tests/TypeReconciliation/ValueTest.php
```diff
@@ -656,6 +656,39 @@ class A {
...
+                /**
+                * @param 0|1 $int
+                */
+                function foo(int $int) : void {
+                    echo (string) $int;
+                }
+
+                foo((int) true);',
+            ],
+            'typecastFalseToInt' => [
```
muglug: I'd say this test isn't necessary

muglug:
> is a test that catches stuff like random_int(0, 2) necessary ?

No

muglug:
Thanks!

Issue #2529, opened 2019-12-29
Can't make inferences about PDOStatement->fetch?


duskwuff:
Psalm seems to have trouble dealing with the return value of `PDOStatement::fetch()`, even with a fetch mode specified.

```
$p = new \PDO("sqlite:::memory:");
$sth = $p->prepare("SELECT 1");
$sth->execute();
while (($row = $sth->fetch(\PDO::FETCH_ASSOC)) !== false) {
  //    ^-- MixedAssignment: Cannot assign $row to a mixed type
}
```

The only thing that seems to help is adding a `/** @var array|false $row */` _directly before_ `$row`, inside the condition of the while loop -- but this seems pretty awkward.

muglug:
I've fixed it for your one use-case, but would you mind opening a PR for [all the other values](https://www.php.net/manual/en/pdostatement.fetch.php) of `PDOStatement::fetch`?

PR #2553, opened 2020-01-05
Add support for more PDOStatement fetch modes

duskwuff:
The `PDO::FETCH_CLASSTYPE` and `PDO::FETCH_PROPS_LATE` flags aren't supported yet, as implementing them would require some way to read the value of an expression like `PDO::FETCH_CLASS | PDO::FETCH_CLASSTYPE` as the argument to `PDOStatement::fetch()`. This seems like something that might be useful down the road, but I'm hesitant to implement it just for this feature. :)

There are a few other really obscure fetch modes which aren't supported either, like `PDO::FETCH_FUNC` and `PDO::FETCH_SERIALIZE`. They're barely even documented and I'm not sure they even make sense to use outside of `fetchAll()`, so... whatever.

(in re. #2529)

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php
@@ -36,10 +36,56 @@ class PdoStatementReturnTypeProvider implements \Psalm\Plugin\Hook\MethodReturnT
             && ($first_arg_type = $source->getNodeTypeProvider()->getType($call_args[0]->value))
             && $first_arg_type->isSingleIntLiteral()
         ) {
-            $value = $first_arg_type->getSingleIntLiteral()->value;
+            $fetch_mode = $first_arg_type->getSingleIntLiteral()->value;
 
-            if ($value === \PDO::FETCH_ASSOC) {
-                return Type::getArray();
+            switch ($fetch_mode) {
+                case \PDO::FETCH_ASSOC: // array<string,scalar>
+                    return new Type\Union([
+                        new Type\Atomic\TArray([
+                            Type::getString(),
+                            Type::getScalar()
+                        ]),
+                    ]);
+
+                case \PDO::FETCH_BOTH: // array<array-key,scalar>
+                    return new Type\Union([
+                        new Type\Atomic\TArray([
+                            Type::getArrayKey(),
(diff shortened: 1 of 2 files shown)


muglug:
Thanks!

I think `PDO::FETCH_CLASS` can use information from the second parameter, right?

staabm:
> Thanks!
> 
> I think `PDO::FETCH_CLASS` can use information from the second parameter, right?

Nope, I dont think so:
https://www.php.net/manual/en/pdostatement.fetch.php

> PDO::FETCH_CLASS: returns a new instance of the requested class, mapping the columns of the result set to named properties in the class, and calling the constructor afterwards, unless PDO::FETCH_PROPS_LATE is also given. If fetch_style includes PDO::FETCH_CLASSTYPE (e.g. PDO::FETCH_CLASS | PDO::FETCH_CLASSTYPE) then the name of the class is determined from a value of the first column.

I guess you mixed it up with fetchObject()
https://www.php.net/manual/en/pdostatement.fetchobject.php

muglug:
Ahh right yes, sorry.

MarkusRodler:
> The `PDO::FETCH_CLASSTYPE` and `PDO::FETCH_PROPS_LATE` flags aren't supported yet, as implementing them would require some way to read the value of an expression like `PDO::FETCH_CLASS | PDO::FETCH_CLASSTYPE` as the argument to `PDOStatement::fetch()`. This seems like something that might be useful down the road, but I'm hesitant to implement it just for this feature. :)
> 
> There are a few other really obscure fetch modes which aren't supported either, like `PDO::FETCH_FUNC` and `PDO::FETCH_SERIALIZE`. They're barely even documented and I'm not sure they even make sense to use outside of `fetchAll()`, so... whatever.
> 
> (in re. #2529)

I don't think that PDO::FETCH_FUNC is very obscure. It is acutally a really handy one. For example if you want to initialize a class with a private constructor. Then you can use FETCH_FUNC to call your static factory method instead.

Example:
```php
// Repository.php
$query->fetchAll(PDO::FETCH_FUNC, [ValueObject::class, 'createFromDb']); // Will return an instance of the ValueObject class
```
```php
// ValueObject.php
class ValueObject
{
    private function __construct(DateTimeImmutable $test) {}
    public static function createFromDb(string $test): self
    {
        return new self(new DateTimeImmutable($test));
    }
}
```

It would be very nice if psalm could support this case, too.

PR #2559, opened 2020-01-06
GithubActions: update actions/checkout to v2

staabm:
See https://github.com/actions/checkout/releases/tag/v2.0.0


muglug:
Thanks!

PR #2566, opened 2020-01-07
Improve array function types

ShiraNai7:
This PR aims to improve type definitions for the following functions using generics:

- `shuffle()`
- `sort()`
- `rsort()`
- `usort()`
- `uksort()`
- `array_diff_assoc()`
- `array_intersect_assoc()`

I've skipped `ksort()` and similar key-preserving sorting functions that are hardcoded in `CallAnalyzer`, because the `ksortPreserveShape` test would fail. I'm not sure if that is something that can be dropped, but `uksort()` didn't and doesn't have this special handling.

`array_uintersect_assoc()`, `array_intersect_ukey()`, `array_udiff_assoc()`, `array_diff_ukey()` have dynamic signatures with varargs in the middle that I didn't feel like dealing with right now 😃

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php
@@ -729,10 +729,8 @@ class CallAnalyzer
         );
 
         $builtin_array_functions = [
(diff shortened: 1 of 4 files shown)


muglug:
Thanks!

Issue #2562, opened 2020-01-06
Support modulo operator


staabm:
Both cases should error https://psalm.dev/r/5912c6fcaa because they can never be true

muglug:
Was there a bug that this check would have prevented?

staabm:
I am not aware of a psalm or php-src bug you might refering to?

muglug:
As in did you run into a bug during your own development that having this check would have prevented? I'm wary of adding specific arithmetic checks as I think they're not terribly useful in real-world situations.

bdsl:
I suppose if it was worth adding support for `%`, `+`, `-`, `*` and `/` would be higher priorities. Psalm [currently](https://psalm.dev/r/8c6afa51ae) types the result of `0+0` as `int`, not `int(0)`.

staabm:
> As in did you run into a bug during your own development that having this check would have prevented? I'm wary of adding specific arithmetic checks as I think they're not terribly useful in real-world situations.

In the past I was experiencing bugs e.g. in $i % 2 == 0 expressions when trying to implement alternating colors for Html-table-rows

SignpostMarv:
@muglug if both sides of a modulo are the same, does psalm set the type to a literal int 0 ?

bdsl:
@SignpostMarv No, it returns `int`, not `int(0)`. You can see by the error message given for https://psalm.dev/r/3e8bf4f1ba - "string() cannot be identical to int"

SignpostMarv:
@bdsl: having a poke around at this particular case.

SignpostMarv:
@bdsl fixed both, tinkering for others.

muglug:
Going to close this as not worth it.

PR #2568, opened 2020-01-07
improve support for modulo operator

SignpostMarv:
~can't quite figure out how to indicate that `101 % 0` is `DivisionByZeroError`~ figured it out in 5ca2518b

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
@@ -1438,17 +1438,21 @@ class BinaryOpAnalyzer
             if ($left_type_part instanceof TInt && $right_type_part instanceof TInt) {
                 if ($parent instanceof PhpParser\Node\Expr\BinaryOp\Mod) {
                     if ($left_type_part instanceof TLiteralInt
-                        && $right_type_part instanceof TLiteralInt
-                    ) {
-                        if ($left_type_part->value === $right_type_part->value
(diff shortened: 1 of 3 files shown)


SignpostMarv:
Should `(literalfloat|literalint)  % (literalfloat|literalint)` be resolved as well as literal ints ?

muglug:
Thanks – I appreciate this, but I can't think there's any real-life code that computes the modulus of two known numbers.

SignpostMarv:
> real-life code that computes the modulus of two known numbers.

That was mostly to fix the particular cases in #2562, however, hold that thought....

SignpostMarv:
one could return a union if only one side were known. prepping a diff.

SignpostMarv:
> real-life code that computes the modulus of two known numbers.

* with `random_int()` it can be known
* with `range()` it can be known
* with `foreach ($arr as $index => $value){ echo $index % 2 ? 'odd' : 'even'; }` where `$arr = <list<T>|array<int, T>>` and the size of `$arr` is known, the possible values of `$index` will be known.

muglug:
I think the only useful situation is something like `$i % 3`, where you can say for certain that the result must lie in the range `(1-$i)` to `($i - 1)`. You could transform that into a union of literal numbers for suitably low values of `$i` like 5 or so, but I still think this is a not particularly useful analysis.

SignpostMarv:
> You could transform that into a union of literal numbers for suitably low values

It looks like BinaryOpAnalyzer already handles unions/arrays & boils them back down to literal comparsions.

Issue #2579, opened 2020-01-09
TooManyArguments is triggered if the variadic function declared at the autoloader file


incrize:

incrize:
I have an autoloader file that declares global functions. Some of the declared functions are variadic.
There is a part of code that uses these global functions. When psalm analyzes that code, I get TooManyArguments errors.

Unit test was implemented as follows: https://github.com/incrize/psalm/commit/6576d1dcefc9393b703e46bc1d89f5ad75835222#diff-f5429097d608f3340a97154a2745842aR38.

Issue can't be reproduced at https://psalm.dev, because it doesn't allow to configure the autoloder file.

incrize:
Maybe i misuse a psalm. In the near future I will prepare a more detailed example.

muglug:
Possibly, or possibly not – haven't had time to examine this in detail yet.

incrize:
Changes in PR do not completely solve my problem, global functions which are includes in autoloader, alse are part of coda base  that will be analized.

incrize:
I figured it out. Some of global functions were included through the dynamic include expression which has resulted functions were not declared as stabbed (https://github.com/vimeo/psalm/blob/7482b9e56fba9003948e0c679ff756acae4f96db/src/Psalm/Internal/Codebase/Functions.php#L33). So PR changes are completely solve my problem.

muglug:
Great! Closing.

PR #2580, opened 2020-01-09
#2579 TooManyArguments was triggered if the variadic function declared at the autoloader file. Fixed.

incrize:
Issue #2579


muglug:
Thanks!

PR #2584, opened 2020-01-10
Support suppressing all issues with @psalm-suppress all

ShiraNai7:
This PR implements `@psalm-suppress all`, which suppresses all issues on a statement or a function. Works with `--find-unused-psalm-suppress` too.

Main use case is for code that isn't supported by Psalm (e.g. closures with variables inherited by reference or other `wontfix` stuff). Such code can generate a lot of false-positive issues, especially with `totallyTyped=true`.

There are existing alternatives, but I didn't find them as "clean" to use:

- `<ignoreFiles>` in config can only ignore entire files, not single functions or lines, one would have to move code to separate files only because of Psalm and that might not be feasible
- `@psalm-suppress` for every single issue, perhaps combined with some `@var` annotations or `assert()`s - this can get ugly quick

I considered other syntaxes, but found `@psalm-suppress all` to be most readable:

- `@psalm-suppress *` looks weird in single-line doc-comments
- `@psalm-suppress` without an argument, it isn't obvious what this does and could be used unintentionally when someone forgets the argument

What do you think?

### DIFF
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -91,6 +91,8 @@ function addString(?string $s) {
 }
 ```
 
+`@psalm-suppress all` can be used to suppress all issues instead of listing them individually.
+
 ### `@psalm-assert`, `@psalm-assert-if-true` and `@psalm-assert-if-false`
 
 See [Adding assertions](adding_assertions.md).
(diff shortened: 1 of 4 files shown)


muglug:
Add a note to the docs and then I'll merge – this looks good, thanks!

PR #2597, opened 2020-01-11
removing trailing comma in function call, 

SignpostMarv:
not detectable by php-cs-fixer re: FriendsOfPHP/PHP-CS-Fixer#4135


muglug:
Thanks!

PR #2598, opened 2020-01-11
Fix: Set environment variable via env section

localheinz:
This PR

* [x] sets the `COMPOSER_ROOT_VERSION` environment variable via `env` section

💁‍♂ For reference, see https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables#about-environment-variables.


muglug:
Thanks!

localheinz:
Thank you, @muglug!

PR #2609, opened 2020-01-12
Fix required parameter after optional in Psalm

TysonAndre:
Psalm supports php 7.1 at minimum, so this is no longer necessary.

Other reason: Future php versions may end up deprecating this, even for null.


muglug:
thanks!

PR #2612, opened 2020-01-12
callmap changes

orklah:
Sorry, I made a mistake and committed two changes where I wanted to create two PR...

1)
set_error_handler returns the previous callback that was set (or null if there wasn't any). Which means the returned type is the same as the parameter

It's used a lot by library to temporary replace the handler

$oldHandler = set_error_handler($callableDoNothing);
//do stuff
set_error_handler($oldHandler);

And with the old "?callable" type, Psalm warns that "null|callable(int,string,string=,int=,array=):bool" is more specific

2)
ReflectionClassConstant::getValue will return the value of a constant. A constant can only be a scalar, an array of scalar or null. Same for constant() function

### DIFF
--- a/src/Psalm/Internal/CallMap.php
+++ b/src/Psalm/Internal/CallMap.php
@@ -11046,7 +11046,7 @@ return [
 'ReflectionClassConstant::getDocComment' => ['string|false'],
 'ReflectionClassConstant::getModifiers' => ['int'],
 'ReflectionClassConstant::getName' => ['string'],
-'ReflectionClassConstant::getValue' => ['mixed'],
+'ReflectionClassConstant::getValue' => ['scalar|array<scalar>|null'],
 'ReflectionClassConstant::isPrivate' => ['bool'],
 'ReflectionClassConstant::isProtected' => ['bool'],
 'ReflectionClassConstant::isPublic' => ['bool'],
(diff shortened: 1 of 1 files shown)


orklah:
Tests are failing when scanning PHPUnit

There are 2 errors:
- The first one (https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/Annotation/DocBlock.php#L335) is in a bloc if(strpos($message, '::') !== false) . I'd argue it's an assert string and it shouldn't be raised
- The second one (https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/Configuration.php#L387) probably never happen as it seems to be a config, but technically, constant can return a non-string-castable type: https://3v4l.org/WXRua

muglug:
@orklah I know it's a bit of a pain, but do you mind sending a PR with the changes PHPUnit should make to `muglug/phpunit`? I'll accept and then point `test-with-real-projects.sh` to that repo temporarily.

orklah:
Nice, thanks!

muglug:
I've reverted part of this in df8f208, but also improved inference of `constant` calls in https://github.com/vimeo/psalm/commit/adc658901a95e8f2593fca0f80662b1cf5e2bc17 so things like `constant("PHP_VERSION")` are have the right types.

muglug:
Reason for the change: the return type is pretty close to `mixed` from a user’s perspective, and Psalm has better handling for that return value (for users who opt into that)

Issue #1576, opened 2019-04-23
False positive TooFewArguments for version_compare when one argument is an array key


TysonAndre:
Seen on dev-master with this snippet:

```php
<?php
function test(string $a, array $x) : bool {
    foreach ($x as $b => $_) {
        // TooFewArguments: Too few arguments for method version_compare - expecting 3 but saw 2
        // This seems to be related to $b being an array key
        if (version_compare($a, $b) < 0) {
            return false;
        }
    }
    return true;
}
test('1', ['2']);
```

Possibly related to #1543

SignpostMarv:
grabbing psalm.dev link for reference: https://psalm.dev/r/b7b4f90587

ArSn:
Can not reproduce this on current master - was it fixed with #2630?

TysonAndre:
The error message for https://psalm.dev/r/b7b4f90587 is now correct. It might have been fixed by that or something else, I think the original issue was also fixed

PR #2630, opened 2020-01-15
int-return form of version_compare takes only two arguments

SignpostMarv:
re: #1576


SignpostMarv:
I'm a little wary of 4efdc6b, as I'm not sure why `\x3c` and `\x3e` are in use instead of the literal brackets?

SignpostMarv:
thinking I might need to PR phpunit, as `PHPUnit\Util\Configuration::satisfiesPhpVersion()` is typed for bool, but "===" would result in null. https://3v4l.org/VE3PX

muglug:
@SignpostMarv please PR in `muglug/phpunit`

SignpostMarv:
muglug/phpunit#3

muglug:
Thanks, I added this manually

PR #2634, opened 2020-01-15
Use floor() instead of round() for long progress percentage

TysonAndre:
PHPUnit and many other applications use floor.
Otherwise, the second-last lines would show 100% despite not being
complete.

```
\sprintf(
    ' %' . $this->numTestsWidth . 'd / %' .
    $this->numTestsWidth . 'd (%3s%%)',
    $this->numTestsRun,
    $this->numTests,
    \floor(($this->numTestsRun / $this->numTests) * 100)
)
```


muglug:
Thanks!

PR #2635, opened 2020-01-16
Modular config files

weirdan:
This change introduces an option to have the configuration split across several files using standard XInclude tags. This may be useful for more complex configs, or to include auto-generated parts into a manually written config file.


muglug:
Thanks!

PR #2643, opened 2020-01-16
check if phpstorm meta path is a file or directory

SignpostMarv:
* swaps `file_exists()` for `is_file()`
* adds a debug note in case someone wonders why their phpstorm meta path isn't being handled

leaving this in draft for now, as I'm wondering if it's worth dropping a `.gitkeep` file in an appropriately named folder for coverage purposes?

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -1560,8 +1560,10 @@ class Config
 
         $phpstorm_meta_path = $this->base_dir . DIRECTORY_SEPARATOR . '.phpstorm.meta.php';
 
-        if (file_exists($phpstorm_meta_path)) {
+        if (is_file($phpstorm_meta_path)) {
             $stub_files[] = $phpstorm_meta_path;
+        } elseif (is_dir($phpstorm_meta_path)) {
+            $progress->debug('phpstorm meta path not included in stub files list as it is a directory' . "\n");
         }
 
         if ($this->load_xdebug_stub) {
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -21,6 +21,7 @@ use function get_defined_functions;
 use function in_array;
 use function intval;
 use function is_dir;
+use function is_file;
 use function json_decode;
 use function libxml_clear_errors;
 use const LIBXML_ERR_ERROR;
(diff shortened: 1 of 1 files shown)


src/Psalm/Config.php
```diff
@@ -1560,8 +1560,10 @@ public function visitStubFiles(Codebase $codebase, Progress $progress = null)
 
         $phpstorm_meta_path = $this->base_dir . DIRECTORY_SEPARATOR . '.phpstorm.meta.php';
 
-        if (file_exists($phpstorm_meta_path)) {
+        if (is_file($phpstorm_meta_path)) {
             $stub_files[] = $phpstorm_meta_path;
+        } elseif (is_directory($phpstorm_meta_path)) {
+            $progress->debug('phpstorm meta path not included in stub files list as it is a directory' . "\n");
```
weirdan: The docs say files in this folder have the same format, so they could all be added to `$stub_files` perhaps?

SignpostMarv: 🤷‍♂️ *does not use phpstorm*

SignpostMarv: How deep does the folder go?

weirdan: I'm not a PHPStorm user either. I suppose *all* files should be read, regardless of the folder depth.

SignpostMarv: @weirdan https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata

> 2016.2 + -> you can have multiple separate files in FOLDER named `.phpstorm.meta.php`

muglug:
> Maybe it makes more sense to include all files from that directory as stub_file instead of skipping them.

Yeah 👍 on this idea, otherwise this is great.

muglug:
Oh, then why is it still a draft?

SignpostMarv:
> Oh, then why is it still a draft?

I think github might've lost the review comment; wasn't sure if doing the [`realpath()` conditional](https://github.com/vimeo/psalm/commit/7ad5f184a4942137b41018e50aa4d57b25f1d2aa#diff-20a62f59764c9aa89c920a98d1b6ec16R1571) was overly-cautious?

muglug:
> wasn't sure if doing the realpath() conditional was overly-cautious?

I'm fine with that

Issue #2646, opened 2020-01-17
Respect that \PDO:fetch() may return `false`


l-x:
This bug was introduced in psalm 3.8.2 (#2553) and is preventing us from upgrading.

```php
$pdo = new \PDO('');
$result = $pdo->query('SELECT * FROM fu')->fetch(\PDO::FETCH_ASSOC);

if ($result === false) {
    throw new \Exception();
}
```
Actual result: (>= 3.8.2)
```
ERROR: TypeDoesNotContainType - dev/psalm_fetch.php:7:5 - array<string, scalar> does not contain false
if ($result === false) {
```

Expected result: (<= 3.8.1)
```
No errors found!
```

The php documentation for `\PDO:fetch()` states:
`The return value of this function on success depends on the fetch type. In all cases, FALSE is returned on failure.`

l-x:
https://psalm.dev/r/fd41a94da3

muglug:
Thanks!

PR #2647, opened 2020-01-17
Allow `false` as \PDO::fetch return type

l-x:
Fix for issue #2646


SignpostMarv:
cc @paragonie-scott re: easydb

muglug:
Thanks!

PR #2648, opened 2020-01-17
Mention how to annotate psalm specific annotations

HenkPoley:
Might be nice to mention how to annotate psalm specific annotations, without trampling over PhpDoc annotations.


HenkPoley:
I'd even recommend to to somehow tell you need to put this annotation second, otherwise the PhpDoc annotation _seems_ to overwrite the psalm specific one.

PR #2660, opened 2020-01-17
Allow Psalm to be installed on PHP 8

scottarc:
PHP 8 won't be released until later this year, but it'd be nice to be able to install Psalm without Composer complaining loudly.


muglug:
Thanks!

PR #2663, opened 2020-01-19
Fix typo in PropertyMap

orklah:
Every other line has two \\ for separator. I figured this must be a typo


muglug:
Just a style issue – PHP treats them the same.

PR #2671, opened 2020-01-22
remove unnecessary loops

bendavies:
spotted some improvements that can be made in https://github.com/vimeo/psalm/commit/8202af957d044c4ff23842878b2a718f45d3c2c9.
This should perform better for large arrays.

### DIFF
--- a/src/Psalm/ErrorBaseline.php
+++ b/src/Psalm/ErrorBaseline.php
@@ -20,6 +20,8 @@ use RuntimeException;
 use function str_replace;
 use function strpos;
 use function usort;
+use function count;
+use function array_values;
 
 class ErrorBaseline
 {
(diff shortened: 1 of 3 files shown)


muglug:
I had it in my head that `array_merge(...$arr)` only worked in PHP 7.4, glad that it's all good. Thanks!

bendavies:
FYI the `array_values` is needed because `array_merge(...$arr)` doesn't work if `$arr` has string indexes.

src/Psalm/ErrorBaseline.php
```diff
@@ -198,16 +200,11 @@ public static function update(
      */
     private static function countIssueTypesByFile(array $issues): array
     {
-        $normalized_data = [];
-
-        foreach ($issues as $file_issues) {
-            foreach ($file_issues as $issue_data) {
-                $normalized_data[] = $issue_data;
-            }
+        if (0 === count($issues)) {
```
snapshotpl: `[] === $issues` is "much" faster than `count`

muglug: updated in 947765c

bendavies: good spot. strangely that i knew that and already do it in all my code, but didn't here! thanks!

someniatko:
@muglug prior to PHP 7.4, `array_merge(...$arr)` will fail if $arr is empty array.

someniatko:
i used to workaround it as `array_merge([], ...$arr)`

Issue #2534, opened 2019-12-30
list<foo> in stubs


orklah:
Hi, 

I noticed there isn't any list<foo> notation in the callmap file. However, I think most native functions actually return list<foo> instead of array<int,foo>

Is this by design? If not, is there a point in trying to submit a PR on this?

muglug:
I'm ok with it if @TysonAndre is. Some of the most commonly-used functions like `explode` have their own return type providers that explicitly return lists: https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Provider/ReturnTypeProvider/ExplodeReturnTypeProvider.php

TysonAndre:
Phan's already using the `list<...>` syntax (and the `associative-array` syntax suggested as an enhancement to Psalm) in its version of the call map.

```
'array_chunk' => ['list<array>', 'input'=>'array', 'size'=>'int', 'preserve_keys='=>'bool'],
'array_keys' => ['list<string>|list<int>', 'input'=>'array', 'search_value='=>'mixed', 'strict='=>'bool'],
'array_values' => ['list<mixed>', 'input'=>'array'],
'call_user_func_array' => ['mixed|false', 'function'=>'callable', 'parameters'=>'list<mixed>'],
'call_user_method_array' => ['mixed', 'method_name'=>'string', 'obj'=>'object', 'params'=>'list<mixed>'],
'debug_backtrace' => ['list<array>', 'options='=>'int|bool', 'limit='=>'int'],
'dns_get_record' => ['list<array>|false', 'hostname'=>'string', 'type='=>'int', '&w_authns='=>'array', '&w_addtl='=>'array', 'raw='=>'bool'],
'explode' => ['non-empty-list<string>', 'separator'=>'string', 'str'=>'string', 'limit='=>'int'],
'fgetcsv' => ['list<?string>|false', 'fp'=>'resource', 'length='=>'int', 'delimiter='=>'string', 'enclosure='=>'string', 'escape='=>'string'],
'file' => ['list<string>|false', 'filename'=>'string', 'flags='=>'int', 'context='=>'resource'],
'forward_static_call_array' => ['mixed|false', 'function'=>'callable', 'parameters'=>'list<mixed>'],
'fscanf' => ['list<mixed>', 'stream'=>'resource', 'format'=>'string'],
'func_get_args' => ['list<mixed>'],
'get_class_methods' => ['list<string>', 'class'=>'mixed'],
[...]

PR #2675, opened 2020-01-22
add list<> notation in stubs

orklah:
As discussed here https://github.com/vimeo/psalm/issues/2534, this PR use the list<> notation where applicable.

I used the list that @TysonAndre posted here https://github.com/vimeo/psalm/issues/2534#issuecomment-569805992. I made some adjustments when types seemed more detailed in Psalm to avoid erasing good data or when I was not sure a particular notation was interpreted by Psalm.

Judging by the number of `array<int, ...` and `type[]` left in stubs, I think I'll create a second PR later if this one goes well.


orklah:
The list I used contained a `list<>` type in the parameters of call_user_func_array. While it's coherent with [docs](https://www.php.net/manual/en/function.call-user-func-array.php)
>param_arr
The parameters to be passed to the callback, as an indexed array.

It also works with string indexed arrays:
https://3v4l.org/Xhqmo

IMHO, return types should be as precise as we can (i.e. `list<>` instead of `array<int, ...>`) but params types should be as broad as possible if it's not provoking errors or unexpected behaviour. I think we should ignore the suggested param type for call_user_func_array and call_user_method_array. Any objection?

src/Psalm/Internal/CallMap.php
```diff
@@ -357,7 +357,7 @@
 'ArithmeticError::getTrace' => ['array<int,array<string,mixed>>'],
 'ArithmeticError::getTraceAsString' => ['string'],
 'array_change_key_case' => ['array|false', 'input'=>'array', 'case='=>'int'],
-'array_chunk' => ['array[]', 'input'=>'array', 'size'=>'int', 'preserve_keys='=>'bool'],
```
TysonAndre: I missed the $preserve_keys=true case when adding this to Phan. For psalm, two separate signatures may make sense

orklah: Good catch. I'll revert for now.

TysonAndre:
> It also works with string indexed arrays:

I'm aware of this - Some of the use of associative arrays or string indexed arrays might be valid and false positives in Phan, but others might be bugs (e.g. an analyzed program where call_user_func should have been used instead of call_user_func_array). Both approaches have arguments for them.

src/Psalm/Internal/CallMap.php
TysonAndre: I forgot what array_chunk was doing, the original version shouldn't have needed changes. The **inner arrays** are lists (`list<list<...>>`) if preserve_keys is false, but they're not lists if preserve_keys is true.

```
php > var_export(array_chunk([1,2,3], 2, true));
array (
  0 => 
  array (
    0 => 1,
    1 => 2,
  ),
  1 => 
  array (
    2 => 3,  // not a list
  ),
)
```

orklah:
@muglug what's your opinion about https://github.com/vimeo/psalm/pull/2675#issuecomment-577334221 ?
I guess it'll be up to you to decide what you want :)

muglug:
@orklah I’d rather not break existing legacy code (`call_user_func_array`) but the other changes are good

muglug:
Thanks!

PR #2693, opened 2020-01-25
List notation in callmap

orklah:
Following #2675, here's another set of changes in callmap.

I only changed return types and I focused on cases annotated `array<int, ...>` or `...[]` and checked documentation to see if it could be replaced by `list<...>`


orklah:
Weird, changing the return type of libxml_get_errors will provoke a CI error: https://travis-ci.org/vimeo/psalm/jobs/641593775#L460

I reverted this line for now.

src/Psalm/Internal/CallMap.php
```diff
@@ -354,13 +354,13 @@
...
 'ArithmeticError::getPrevious' => ['?Throwable'],
-'ArithmeticError::getTrace' => ['array<int,array<string,mixed>>'],
+'ArithmeticError::getTrace' => ['list<array<string,mixed>>'],
 'ArithmeticError::getTraceAsString' => ['string'],
 'array_change_key_case' => ['array|false', 'input'=>'array', 'case='=>'int'],
 'array_chunk' => ['list<array[]>', 'input'=>'array', 'size'=>'int', 'preserve_keys='=>'bool'],
 'array_column' => ['array', 'array'=>'array', 'column_key'=>'mixed', 'index_key='=>'mixed'],
 'array_combine' => ['array|false', 'keys'=>'string[]|int[]', 'values'=>'array'],
-'array_count_values' => ['int[]', 'input'=>'array'],
+'array_count_values' => ['array<string|int, int>', 'input'=>'array'],
```
muglug: Please revert this change – various tools (including Psalm) treat `array<int>` and `array<string|int, int>` slightly differently.

orklah: Done. Pretty curious what is treated differently and why though. int[] looks like missing details to me. Even `array-key` is ambiguous and doesn't say much more about actual type.

If it has any incidence in user land annotated code, it would be nice to know what to expect :)

Thanks!

muglug: `array-key` is treated slightly like `mixed` – it's a "not sure" type. `string|int` means "it can either be a string or an int – either is valid". 

So `array-key` is to `string` and `int` what `mixed` is to every sort of type – a less-specific version thereof.

PR #2699, opened 2020-01-27
mark additional impure functions

### LINKED ISSUE #2687: vprintf and printf are treated differently
https://psalm.dev/r/89a13d8115

both `vprintf` and `printf` return `int`, but only in case of `vprintf` psalm warns about the unused return value via "the call to vprintf is not used".

I would argue that the behavior with `printf` is more correct I don't think anybody really cares or should care about how many characters were just printed to stdout.

pilif:
both `fflush` and `vprintf` have very similar cousins already declared as having side-effects.

This fixes #2687


muglug:
Thanks!

PR #2710, opened 2020-01-29
do not corrupt internal state when a baseline is present

### LINKED ISSUE #2709: psalm master reports errors which are baselined
Even though we have base-lined `MissingPropertyType` but otherwise treat it as errors, psalm `master` randomly reports some instances as `ERROR` but then still exits with zero exit code and the "No issues found" tagline.

I have tracked this down to 8202af957d044c4ff23842878b2a718f45d3c2c9 where a reference in `foreach()` is reused later on which causes corruption of internal state.

I have a fix too, but I want to provide a failing test too.

pilif:
this fixes #2709

### DIFF
--- a/src/Psalm/IssueBuffer.php
+++ b/src/Psalm/IssueBuffer.php
@@ -447,6 +447,7 @@ class IssueBuffer
                         return $d1['file_path'] > $d2['file_path'] ? 1 : -1;
                     }
                 );
+                unset($file_issues);
             }
 
             if (!empty($issue_baseline)) {
--- a/tests/IssueBufferTest.php
+++ b/tests/IssueBufferTest.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace Psalm\Tests;
+
+use Psalm\Codebase;
+use Psalm\Config;
+use Psalm\Internal\Analyzer\ProjectAnalyzer;
+use Psalm\Internal\Codebase\Analyzer;
+use Psalm\IssueBuffer;
+use Psalm\Report\ReportOptions;
+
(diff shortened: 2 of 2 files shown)


pilif:
I'm a bit unhappy with the "unit"-test because that's basically an integration test.

However, I don't see another solution as `IsssueBuffer` is all `static` methods and `IssueBuffer::finish()` is huge and accesses a huge swath of dependencies.

(also, this is currently failing the self-test - I'll fix that in a followup)

pilif:
I have force-pushed to fix the issues I was causing. However, self-analyze still reports an issue that's completely unrelated to this PR and also doesn't seem to happen on my local machine 🤷‍♀

pilif:
never mind - I f'ed up a mock. Now there's much more green.

pilif:
the issues on Travis CI / PHP 7.4 are caused by an outdated version of phpunit being used where `$this->createMock()` throws a deprecated warning. 

As I'm unconvinced of the value of this test anyways, I'm not going to dump the dependency version requirement without first having your input.

muglug:
Thanks! So I should wait to merge?

pilif:
> Thanks! So I should wait to merge?

that's up to you to decide.

I can't make the test any better without seriously reworking things but for that I don't feel familiar enough with the project.

If you accept the test as-is, the phpunit dependency would need to be bumpted to `7.5.16` for 7.5. If that's ok, I'll make the change to composer.json

If you think the test is too integration-testy, then I can also remove it.

muglug:
Oh please bump the dep!

muglug:
Actually I'll do it myself

PR #2715, opened 2020-01-30
allow setting php version from config or composer.json

### LINKED ISSUE #2628: Add `<psalm phpVersion="7.2"` support to psalm.xml
(the syntax isn't important)

Related to #1191

Some applications will only be deployed on one version of php most of the time.
It's inconvenient to add `--php-version 7.2` to all invocations (language server, manual runs, etc).

`--php-version` should override the psalm.xml setting.

pilif:
if a `composer.json` is present and a PHP version requirement is configured, we set the php version to the minimal PHP version that satisfies the composer requirement.

Additionally, this adds a `phpVersion` attribute to the `<psalm>` tag. If that's set, it takes precedence over what has been detected in `composer.json`.

And finally, the `--php-version` command line flag continues to work and takes precedence over the setting in the `<psalm>` tag

this fixes #2628


src/Psalm/Config.php
```diff
@@ -1861,8 +1873,42 @@ public function addStubFile(string $stub_file)
...
+        $composer_json_path = $this->base_dir . DIRECTORY_SEPARATOR. 'composer.json';
+
+        if (file_exists($composer_json_path)) {
+            if (!$composer_json = json_decode(file_get_contents($composer_json_path), true)) {
+                throw new \UnexpectedValueException('Invalid composer.json at ' . $composer_json_path);
+            }
+            $php_version = $composer_json['require']['php'] ?? null;
+
+            if ($php_version) {
+                foreach (['5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0'] as $candidate) {
```
pilif: I don't like having to do this loop here hard-coding versions, but IMHO there's no way to deterministically reverse a composer range to end up with a minimum version.

What still is a problem is that the knowledge of what versions of PHP psalm has actual support for isn't centralized but so far only existed in this regex here:

https://github.com/vimeo/psalm/blob/5d15ef062e0efdf7d3d820311148f4816943d176/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php#L1164

maybe for the future it would be worth to actually have proper API to get to this information so further duplication isn't needed.

PR #2730, opened 2020-02-01
small stubs changes for reflection

orklah:
Added class-string/trait-string instead of string when relevant

Thanks


orklah:
Appveyor timed out.

The error flagged by circleci seems a false-positive, it was not reported on appveyor and the variable seems correctly initialized.

muglug:
Yeah that’s an issue in the `--diff-methods` command that I need to debug.

muglug:
Thanks!

PR #2732, opened 2020-02-02
Remove non-existent $serverctrls parameter for \ldap_bind()

LeSuisse:
In a8e58b0ba0242819ce201870ac56e05f7604e6ef a `$serverctrls` parameter
was incorrectly added to the \ldap_bind() function.

See php/doc-en#20.


muglug:
Thanks!

Issue #2729, opened 2020-01-31
Support @psalm-private-mutate annotation


christian-kolb:
With PHP 7.4 using public properties in a type safe way is finally possible. In some languages you have the concept of marking a property to be able to be read from outside the class, but not modified. Basically public get, but private set. 
Psalm already supports `@psalm-readonly` which prevents the property by being modified anywhere. It would be awesome to have a similar annotation which allows to modify the property with class methods. With this the public get, private set paradigm would be available to PHP users 🙂 
Please let me know when there is already concept within psalm with which I could get the same result, I haven't found one yet.

muglug:
Have you read https://psalm.dev/articles/immutability-and-beyond?

christian-kolb:
@muglug Yes, but thanks for the reminder (was a while ago). This might be a good example to help me explain what I want.

```
<?php

class Counter {
  /** @psalm-new-annotation */
  public int $count = 0;
  
  public function increment() : void {
    $this->count++;
  }
}

function makeCounter() : Counter {
  $a = new Counter();
  $a->increment(); // this is fine
  return $a;
}

function canNotUpdateDirectly(Counter $a) : Counter {
  $a->count = 10; // This is not allowed
}
```

So basically I can use a public property as getter, not as setter, but modify it through its class methods.

As far as I'm aware this behaviour can only made available with getter methods but not with psalm annotations. Basically I want to get rid of getters, because I have many instances where I only have them to get this result. And at the moment this means I have "clean" classes without getters and ones cluttered with them. It would be amazing if I could solve this issue with psalm 🙂

muglug:
I think you're asking for `/** @psalm-external-mutation-free */` – does its use in the article line up with what you want?

christian-kolb:
@muglug Kind of, but this only applies to methods, doesn't it? I would need it to be applied for classes and properties. I tried with psalm.dev, but it doesn't raise any issues: 

https://psalm.dev/r/69691fb410

It would describe the behaviour I want 🙂

muglug:
OK, I understand now – if you contribute an implementation and/or documentation, I'll do the rest.

I suggest an annotation of `@psalm-private-set`

christian-kolb:
Awesome, I have no idea how I would start an implementation for it. But I'm sure I can write documentation for it. `@psalm-private-set` sounds great. I'll start working on it tomorrow  🙂

PR #2733, opened 2020-02-02
Add new @psalm-allow-private-mutation annotation

christian-kolb:
Pull request to solve #2729.

@muglug As discussed here is the documentation part of the annotation. I tried to make it as similar to the other annotations as possible. Let me know if I should change anything there.
I'm very curious to see how you'll implement the annotation. Hopefully I will understand it with this example, then I might be able to offer the implementation for further annotations in the future myself 🙂 
Thanks again for your offer to do the implementation for this one!

### DIFF
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -324,13 +324,13 @@ echo Arithmetic::addCumulative(3); // outputs 3
 echo Arithmetic::addCumulative(3); // outputs 6
 ```
 
-### `@psalm-private-set`
+### `@psalm-private-mutate`
 
-Used to annotate properties which can only be set in a private context. With this, public properties can be read from another class but only be modified within a method of its own class.
+Used to annotate properties which can only be mutated in a private context. With this, public properties can be read from another class but only be mutated within a method of its own class.
 
 ```php
 class Counter {
-  /** @psalm-private-set */  
+  /** @psalm-private-mutate */  
   public int $count = 0;
     
   public function increment() : void {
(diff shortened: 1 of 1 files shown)


docs/annotating_code/supported_annotations.md
```diff
@@ -324,6 +324,27 @@ echo Arithmetic::addCumulative(3); // outputs 3
 echo Arithmetic::addCumulative(3); // outputs 6
 ```
 
+### `@psalm-private-set`
```
staabm: The word „set“ has a lot of meanings: what do you think about „mutate“, „change“ or similar

christian-kolb: In the description I use "modify". 
It is technically a "set" though and as it's to get rid of "getters and setters", I think "set" does fit.
"change" would be to generic for my taste.

"set", "modify" and "mutate" would all work for me, but I guess it's most important what most people will understand. @muglug What do you think?

muglug: Oh mutate is probably the best since that’s used elsewhere. So `@psalm-private-mutate`

christian-kolb: Ok, I've updated the documentation, issue and pull request naming

muglug:
Thanks! I decided that this should just be a modifier for `@readonly` properties, as those semantics were already well-defined within Psalm, and understood externally.

Adding `@psalm-allow-private-mutation` to a property with a `@readonly` annotation prevents allows it to be edited within the class.

Another alternative to this would be `@psalm-readonly-public`, which would combine the two. In fact I think I'll document that myself.

muglug:
Added in b3f6b56

christian-kolb:
> Thanks! I decided that this should just be a modifier for `@readonly` properties, as those semantics were already well-defined within Psalm, and understood externally.
> 
> Adding `@psalm-allow-private-mutation` to a property with a `@readonly` annotation prevents allows it to be edited within the class.
> 
> Another alternative to this would be `@psalm-readonly-allow-private-mutation`, which would combine the two. In fact I think I'll document that myself.

@muglug Yes, that makes so much more sense 🙂 Thanks for the implementation

PR #2737, opened 2020-02-02
add magic properties missing from PHP-Parser

orklah:
These would be useful for type coverage in Roave/Better-Reflection


muglug:
Closing in favour of https://github.com/vimeo/psalm/commit/8d7fb2b4151512f3718771c4ab45be33cb1d7264, which will allow that project to pick up php-parser's `@property` types if the correct config is used.

orklah:
Even cooler!

orklah:
@muglug I think this may need to be refreshed with new setting: https://psalm.dev/schema/config

PR #2750, opened 2020-02-05
PDOStatement::fetchColumn() can return any scalar

### LINKED ISSUE #2743: PDOStatement::fetchColumn() can return int
Can't provide a `3v4l` link, as it does not have a MySQL server running. But when running the following code, one can see that `fetchColumn()` can return `int` as well, with the `mysql` driver and `PDO::ATTR_EMULATE_PREPARES` set to `false`.

```php
$pdo = new PDO('mysql:host=localhost', 'root', '');
$pdo->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);

var_dump($pdo->query('SELECT 1')->fetchColumn()); // int(1)
```

As for `float`s, these seem to be always returned as strings by the `mysql` driver. However, the return type may be entirely up to the driver, and could be the reason why it's documented as `mixed`. The [doc](https://www.php.net/manual/en/pdostatement.fetchcolumn.php) actually says:

> `PDOStatement::fetchColumn()` should not be used to retrieve boolean columns, as it is impossible to distinguish a value of `FALSE` from there being no more rows to retrieve. Use `PDOStatement::fetch()` instead.

So it looks like some drivers may even return `bool`.

As such, I would suggest to document this method with every possible **scalar** type (`PDO::FETCH_CLASS` does not make sense in this context, and I don't *think*  a driver could return an array or a resource):

```php
fetchColumn() : string|int|float|bool|null
```

Thoughts?

BenMorel:
Fixes #2743.

### DIFF
--- a/src/Psalm/Internal/CallMap.php
+++ b/src/Psalm/Internal/CallMap.php
@@ -9772,7 +9772,7 @@ return [
 'PDOStatement::execute' => ['bool', 'bound_input_params='=>'?array'],
 'PDOStatement::fetch' => ['mixed', 'how='=>'int', 'orientation='=>'int', 'offset='=>'int'],
 'PDOStatement::fetchAll' => ['array|false', 'how='=>'int', 'fetch_argument='=>'int|string|callable', 'ctor_args='=>'?array'],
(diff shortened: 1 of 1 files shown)


muglug:
thanks!

PR #2760, opened 2020-02-07
Allow sebastian/diff 4.0 to be installed

morozov:
This is required to allow installation of Psalm alongside PHPUnit 9 using Composer.


muglug:
Thanks!

PR #2800, opened 2020-02-12
Fix #2799 - DOMNode::$parentNode is nullable

### LINKED ISSUE #2799: is_null(DOMNode::$parentNode) causes TypeDoesNotContainNull error
The `TypeDoesNotContainNull` error in the following example surprises me:

https://psalm.dev/r/713d990318

Interestingly, changing the `while` condition to `$node->parentNode instanceof DOMNode` causes a `RedundantCondition` locally (`Found a redundant condition when evaluating $node->parentNode and trying to reconcile type 'DOMNode' to DOMNode`), but not on https://psalm.dev/.

caugner:
Fixes #2799.


muglug:
Thanks!

Issue #1732, opened 2019-06-04
Psalm V4 breaking changes


muglug:
Coming late this year, probably to coincide with 7.4 release

- [x] class constants should live in their own storage object
- [x] remove issues `MisplacedRequiredParam`, `TypeCoercion`, `MixedTypeCoercion`
- [x] remove support for `@template-typeof`
- [x] rename `TFn` to `TClosure`, `ObjectLike` to `TKeyedArray`
- [x] Remove return type manipulation from `AfterFunctionCallAnalysisInterface`
- [x] default to not disabling checks with `UnresolvableInclude` (#2818)
- [x] Added return and param types to plugin hooks
- [x] Remove `requireVoidReturnType`, `useAssertForType` config flags
- [x] Change default on `resolveFromConfigFile` to `"true"`

weirdan:
@muglug it seems v4 just never happened. Is the list still up to date?

muglug:
I'll probably do it with PHP 8 release, which seems like an even better reason

PR #2818, opened 2020-02-14
don't disable checks when includes can't be resolved

### LINKED ISSUE #2817: unresolvable includes disable many checks
I noticed by accident that psalm wasn't catching a typo in a static method called on a class.

After some digging, I boiled the issue down to an unresolvable include that we silenced.

Apparently, once psalm hits an unresolvable include, it silently turns off a lot of checks:

https://github.com/vimeo/psalm/blob/422bd25f04f529139b0ac80888bb65c680b262eb/src/Psalm/Internal/Analyzer/Statements/Expression/IncludeAnalyzer.php#L206-L208

I have a demo project here:

https://github.com/pilif/psalm/tree/unresolvable-include

`src/demo.php` in this project blows up with a `PHP Fatal error:  Uncaught Error:` but psalm thinks everything is fine.

As a user, I would not expect that a suppressed warning (do note the that included file is entirely irrelevant for the rest of the project) on an include somewhere has the side-effect of turning off many checks which would even cause fatal errors at runtime if not caught.

pilif:
Any unresolvable include (even suppressed ones) would lead to some subsequent tests being disabled as a side-effect.

this fixes #2817

As noted on that bug report, this might cause big swaths of already-present-but-to-date-hidden issues to be uncovered for people updating to a release with this patch applied.

I'm unsure as to whether we need a switch to toggle this updated behavior or whether we need some changelog warning or whether just fixing this is ok.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -68,6 +68,7 @@
         <xs:attribute name="useDocblockPropertyTypes" type="xs:boolean" default="false" />
         <xs:attribute name="usePhpDocMethodsWithoutMagicCall" type="xs:boolean" default="false" />
         <xs:attribute name="usePhpDocPropertiesWithoutMagicCall" type="xs:boolean" default="false" />
+        <xs:attribute name="skipChecksOnUnresolvableIncludes" type="xs:boolean" default="true" />
     </xs:complexType>
 
     <xs:complexType name="ProjectFilesType">
--- a/docs/running_psalm/configuration.md
+++ b/docs/running_psalm/configuration.md
@@ -275,6 +275,17 @@ Set the php version psalm should assume when checking and/or fixing the project.
 
 This can be overridden on the command-line using the `--php-version=` flag which takes the highest precedence over both the `phpVersion` setting and the version derived from `composer.json`.
 
+#### skipChecksOnUnresolvableIncludes
+```xml
+<psalm
+  skipChecksOnUnresolvableIncludes="[bool]"
+>
+```
+
+When `true`, Psalm will skip checking classes, variables and functions after it comes across an `include` or `require` it cannot resolve. This allows code to reference functions and classes unknown to Psalm.
+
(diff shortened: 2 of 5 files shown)


pilif:
@muglug is there something else I can do here? I think the test illustrates the issue nicely

https://github.com/vimeo/psalm/blob/14a843ae4b759c291cd1bc184a4279ab07c44513/tests/IncludeTest.php#L784-L791

and I think it's bad enough to warrant doing something about this. Right now any unresolvable include can hide fatal errors (for example the call to `foo()` which is never defined in the test).

muglug:
Could you make this configurable? I'm happy for this to be the default from now onwards, but there are probably some projects that rely on this behaviour. I can imagine someone doing something strange like

```php
require("file_that_only_exists_when_command_is_run.php");
function_defined_in_that_file();
```

and expecting Psalm not to worry.

After a _lot_ of refactoring over the years, at Vimeo we only depend on this behaviour on a single line, so I'm happy for this change in general.

pilif:
Yep. I'll add a setting. I'll update the PR.

In my case, I thought I finally got our 14 years old code-base in shape only to stumble over an obvious type that psalm didn't detect, so I dug into this.

The end result is 520 more issues that were completely hidden from me.

I've since fixed them too.

muglug:
> I've since fixed them too.

Awesome!

pilif:
I have rebased on top of current master and added a second commit that makes the behavior configurable - including the documentation to explain how it works and including an additional test (using the same fixture but the other setting) to make sure the setting is applied correctly.

In an abundance of caution I'm defaulting to the old behavior in order to not blow up all over people's faces when they upgrade psalm. 

Feel free to change the default if the next release is going to be 4.0 - and maybe add a note to  #1732 for doing so.

I still strongly believe that the default should be to not silently disable initial functionality that could prevent fatal errors in production.

muglug:
Thanks!

PR #2820, opened 2020-02-14
change `show-info` parameter default

### LINKED ISSUE #2794: --show-info=false should become the default
Why?

It creates a noisy output, and makes people think too much. People don't like to think.

pilif:
this fixes #2794


muglug:
I did this myself, sorry

PR #2825, opened 2020-02-17
Add failing test for suppress duplicateCoreFunction

leighman:
Maybe I'm being stupid and not sure if this is a useful way to do it but I would expect this test to pass and it doesn't.


muglug:
This should definitely be an issue, not a PR

PR #2839, opened 2020-02-18
Improve the reported error when a worker crashes

TysonAndre:
I saw this when I failed to update psalm/plugin-phpunit (and it was fixed by running `composer update`). It's more useful to have a stack trace than to just see that `explode` was passed an object instead of a string.

It may also be useful (for developers) to check the output of `ocramius/package-versions` against psalm's composer.json, and warn if the installed versions don't match composer.json.
This is possible with composer/semver, but this type of bug is probably rare enough for it not to be worked on.


muglug:
Thanks!

PR #2840, opened 2020-02-18
Support using paratest for local development

TysonAndre:
https://phpunit.de/manual/6.5/en/appendixes.configuration.html#appendixes.configuration.testsuites
documents that `<testsuites>` should typically wrap `<testsuite>` in
phpunit.xml.

> The <testsuites> element and its one or more <testsuite> children can
> be used to compose a test suite out of test suites and test cases.

This project may get a small performance boost
running tests in CI with paratest(2 processes) instead of phpunit.
Paratest works locally, and supports php 7.1+.

See https://github.com/paratestphp/paratest/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "brianium/paratest": "^4.0.0",
         "phpmyadmin/sql-parser": "5.1.0",
         "phpspec/prophecy": ">=1.9.0",
-        "phpunit/phpunit": "^7.5.16 || ^8.0",
+        "phpunit/phpunit": "^7.5.16 || ^8.5",
         "psalm/plugin-phpunit": "^0.9",
         "slevomat/coding-standard": "^5.0",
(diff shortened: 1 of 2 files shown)


muglug:
Is there an easy way to fix the travis build?

TysonAndre:
Maybe it was caused because the change to composer.json cleared the travis cache (or because paratest required a newer phpunit)

https://travis-ci.org/vimeo/psalm/jobs/652154102?utm_medium=notification&utm_source=github_status
mentions that phpunit 8.0.0 failed because ReflectionType->__toString was deprecated in phpunit 8 or its dependencies in createMock

TysonAndre:
Bumping the minimum development phpunit version may help, it's somehow installing 8.0.0 but 8.5.x is available https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-8.5.md

muglug:
Thanks!

PR #2850, opened 2020-02-19
Codelocation: Save docblock

jarstelfox:
When using a plugin that inspects the `codebase`, it is not easy to loop over classes and methods pulling out the docblocks.

Since we have code locations at that time, let's stash the docblock on the code location.

Note: I opened this since I could not find an easy way to look up a docblock given a `codelocation` object.

### DIFF
--- a/src/Psalm/CodeLocation.php
+++ b/src/Psalm/CodeLocation.php
@@ -77,6 +77,9 @@ class CodeLocation
     /** @var int|null */
     private $docblock_line_number;
 
+    /** @var PhpParser\Comment\Doc|null */
+    private $doc_comment;
+
     /** @var null|int */
     private $regex_type;
 
@@ -121,6 +124,7 @@ class CodeLocation
         $this->text = $selected_text;
 
         $doc_comment = $stmt->getDocComment();
+        $this->doc_comment = $doc_comment;
         $this->preview_start = $doc_comment ? $doc_comment->getFilePos() : $this->file_start;
         $this->docblock_start_line_number = $doc_comment ? $doc_comment->getLine() : null;
         $this->raw_line_number = $stmt->getLine();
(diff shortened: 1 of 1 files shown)


jarstelfox:
Note: Current error seems to be related to recent changes before this branch

```php
ERROR: UndefinedThisPropertyAssignment - src/Psalm/Configuration.php:30:9 - Instance property Roave\YouAreUsingItWrong\Psalm\Configuration::$totally_typed is not defined
        $this->totally_typed           = true;
```

muglug:
What exactly is the plugin doing?

jarstelfox:
> What exactly is the plugin doing?

@muglug I was toying around with making a php documentor library similar to https://github.com/phpDocumentor/phpDocumentor or https://github.com/FriendsOfPHP/Sami

However, neither of these work well with psalm's types.

My ideal plan is to make a plugin which dumps the psalm info about a class / it's methods into a [mdx template](https://mdxjs.com/) an generate the static documentation site using [gatsby](https://www.gatsbyjs.org/)

I am aware this is probably no-where near your intended use case for plugins - but thought it would be a nice addition to psalm projects.

muglug:
Can't you just save the start and end positions of the docblock, then use file loading later? I don't want to bloat the size of `CodeLocation` objects, as there can be a LOT of them

jarstelfox:
>Can't you just save the start and end positions of the docblock, then use file loading later?

Absolutely! I wasn't sure how big of a deal space was here.

I ran my change on a decent size code base (120,000 commits) and didn't notice a massive increase in memory usage. Thus, opened up the pull.

Out of curiosity, is there an easy way to parse just the start / end position of the doc block?

From what I could find, `StatementsProvider` only allows reparsing the entire file:
https://github.com/vimeo/psalm/blob/05783eb61617251ad6d381bcd74d0268cefd1bdd/src/Psalm/Internal/Provider/StatementsProvider.php#L89

muglug:
> Out of curiosity, is there an easy way to parse just the start / end position of the doc block?

I added `$docblock_start` and `$docblock_end` properties to `CodeLocation` in e5c262d.

To get the docblock contents for a given `CodeLocation` you can do

```php
$docblock_contents = substr(
    $codebase->getFileContents($location->file_path),
    $location->docblock_start,
    $location->docblock_end - $location->docblock_start
);
```

jarstelfox:
@muglug the help is much appreciated. Thank you!

PR #2851, opened 2020-02-19
GD functions that output an image accept stream resource

Daeroni:
Most GD functions that output an image accept either null (default, output directly), a filename, or a stream resource as the optional second parameter.

The only exception is imagexbm, which only accepts null (default) or a filename as the optional second parameter


muglug:
Thanks!

PR #2856, opened 2020-02-21
Prevent TypeError when generating XML reports

JaredCDN:
The XML generator currently in use (LSS\Array2XML) expects a nested array structure. However, due to the recent introduction of the IssueData type to replace what was previously shaped arrays, Array2XML was getting an array of objects instead. It then proceeds to crash from (eventually) passing null to DOMNode->appendChild(). This commit maps the IssueData array back into the array of shaped arrays that Array2XML knew and loved.

Feel free to edit as desired. This is an admittedly rudimentary fix; just wanted to make an attempt at helping in some capacity. 😅


muglug:
Thanks!

Issue #2869, opened 2020-02-24
Psalm Plugin: Can't seem to autoload Psalm/Test namespace


jarstelfox:
While writing a plugin, I wanted to allow the plugin to be tested similar to your existing psalm tests: I want to get an [analyzed codebase from a string](https://github.com/vimeo/psalm/blob/master/tests/ArrayAccessTest.php#L21-L32). This would allow my plugin to be tested without needing to checkout multiple codebase's / running psalm on them.

I was looking at pulling in your [Test namespace](https://github.com/vimeo/psalm/blob/master/composer.json#L77-L80) however, [autoload-dev is root-only](https://getcomposer.org/doc/04-schema.md#autoload-dev) which does not allow my project to easily pull in this test namespace.

My current composer.json does not work:

```php
{
  "name": "jarstelfox/psalmdocumentor",
  "description": "Outputs docblocks as markdown documentation with psalm's types",
  "type": "psalm-plugin",
  "require": {
      "phpunit/phpunit": "^8.0",
      "vimeo/psalm": "dev-master"
  },
  "autoload": {
      "psr-4": {
         "PsalmDocumentor\\": "src/"
      }
   }
  "autoload-dev": {
      "psr-4": {
         "Psalm\\Tests\\": "vendor/vimeo/psalm/tests/"
      }
   }
}
```

However, I can load your classes one by one with `require_once __DIR__ . '/../vendor/vimeo/psalm/tests/TestCase.php'`

Do you have any insight into how to easily autoload your test namespace when writing a plugin?

weirdan:
Seems to be working for me: https://github.com/weirdan/psalm-tests-autoload (clone, `composer install`, then run `php main.php`).

jarstelfox:
Whoops! I forgot `require 'vendor/autoload.php';`

However, now I am seeing:

```
phpunit ./tests 
PHPUnit 8.3.5 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 103 ms, Memory: 14.00 MB

There was 1 error:

1) Test::testCaseOne
Psalm\Exception\ConfigException: Could not resolve config path to /Users/Jarred/Documents/Coding/psalmdocumentor//src/Psalm/Internal/Stubs
```
given

```php
<?php
require 'vendor/autoload.php';

use Psalm\Tests\TestCase;

class Test extends TestCase
{
   public function testCaseOne(): void {}
}
```

PR #2870, opened 2020-02-24
TestCase/TestConfig: Allow plugins to reuse these files

jarstelfox:
See: https://github.com/vimeo/psalm/issues/2869\#issuecomment-590490908

Previously, If a plugin tried to reuse the TestConf / TestCase
an exception would likely occur since the default testconfig
is hardcoded to our internal psalm codebase

This commit allows a custom config to be passed into a testcase
thus, a plugin's codebase does not need to match our own.

CC: @weirdan

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -7,6 +7,7 @@ use const DIRECTORY_SEPARATOR;
 use function getcwd;
 use function ini_set;
 use PHPUnit\Framework\TestCase as BaseTestCase;
+use Psalm\Config;
 use Psalm\Internal\Analyzer\FileAnalyzer;
 use Psalm\Internal\Analyzer\ProjectAnalyzer;
 use Psalm\Internal\Provider\Providers;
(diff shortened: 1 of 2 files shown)


tests/TestCase.php
```diff
@@ -43,6 +44,14 @@ public static function setUpBeforeClass() : void
         self::$src_dir_path = getcwd() . DIRECTORY_SEPARATOR . 'src' . DIRECTORY_SEPARATOR;
     }
 
+    /**
+     * @return Config
+     */
+    protected function getConfig() : Config
```
weirdan:
```suggestion
    protected function makeConfig() : Config
```

muglug:
But test classes aren't available to plugins (the namespace is only mentioned in `autoload-dev`, I think?

weirdan:
Plugins may provide their own `autoload-dev` referencing `vendor/vimeo/psalm/tests` (see discussion in #2869). Of course it will only work with `--prefer-source` install.

PR #2873, opened 2020-02-24
Only set 4G memory_limit in psalter if lower

### LINKED ISSUE #2302: Hardcoded memory limit of 4096M in psalter makes it difficult to use more memory
My application is very large and I need more than 4096M.  psalter.php has a line that overrides the CLI's default memory_limit setting with ini_set('memory_limit', '4096M');, making it impossible to provide more without manually commenting out this line.

It would be nice if psalter.php only set the memory_limit to 4096M if it was currently less than 4096M.

ArSn:
Fixes #2302 in the proposed manner.

### DIFF
--- a/src/psalter.php
+++ b/src/psalter.php
@@ -6,12 +6,17 @@ use Psalm\Config;
 use Psalm\IssueBuffer;
 use Psalm\Progress\DebugProgress;
 use Psalm\Progress\DefaultProgress;
+use function getMemoryLimitInBytes;
 
 // show all errors
 error_reporting(-1);
 ini_set('display_errors', '1');
 ini_set('display_startup_errors', '1');
-ini_set('memory_limit', '4096M');
+$memLimit = getMemoryLimitInBytes();
+// Magic number is 4096M in bytes
+if ($memLimit > 0 && $memLimit < 4294967296) {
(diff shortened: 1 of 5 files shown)


src/psalter.php
```diff
@@ -6,12 +6,17 @@
...
+use function getMemoryLimitInBytes;
 
 // show all errors
 error_reporting(-1);
 ini_set('display_errors', '1');
 ini_set('display_startup_errors', '1');
-ini_set('memory_limit', '4096M');
+$memLimit = getMemoryLimitInBytes();
+// Magic number is 4096M in bytes
+if ($memLimit > 0 && $memLimit < 4294967296) {
```
staabm: Maybe just use constant MIN_MEM_LIMIT and MIN_MEM_LIMIT_BYTES

ArSn: Maybe, but where would I define those, since this is the only place I check them? 

Plus, `MIN_MEM_LIMIT_BYTES` would make sense but I would have `MAX_MEM_LIMIT_BYTES` for the upper boundary which seems kind of counter intuitive since it is not actually the maximum of anything but rather the memory psalter wants to make sure to have. 

Not sure this gets readable any better with this.

muglug:
Thanks!

PR #2883, opened 2020-02-26
Fixed stub files with absolute paths not being loaded.

### LINKED ISSUE #1331: stubs / file_name in config under Windows
Hi,

Just experienced the following issue under Windows:

```
Cannot resolve stubfile path C:\<path-where-psalm-config-is>\\file://C:\wp\wp-includes\link-template.php
```

No matter if I'll include `file://` or not, the issue is the same, Psalm looks for a directory relative to config path and ignores the full path provided in the config. Is this expected?

Thanks in advance!

ArSn:
Fixes #1331.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -936,7 +936,11 @@ class Config
         if (isset($config_xml->stubs) && isset($config_xml->stubs->file)) {
             /** @var \SimpleXMLElement $stub_file */
             foreach ($config_xml->stubs->file as $stub_file) {
-                $file_path = realpath($config->base_dir . DIRECTORY_SEPARATOR . $stub_file['name']);
+                $stub_file_name = (string)$stub_file['name'];
+                if (!Path::isAbsolute($stub_file_name)) {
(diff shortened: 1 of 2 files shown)


ArSn:
Hmm weird, are the CI processes running on a presumed merge? I mean, does travis fail because master is unstable? I specifically reset my branch to the previous commit on master so it would not fail (actually)

muglug:
It should just run on your commits. Anyway, this looks good, thanks!

weirdan:
@ArSn @muglug 
Travis PR builds are made on the merge-to-be:

> Rather than build the commits that have been pushed to the branch the pull request is from, we build the merge between the source branch and the upstream branch.
*[Travis Docs: How pull requests are built](https://docs.travis-ci.com/user/pull-requests/#how-pull-requests-are-built)*

I'm fairly sure other CI systems would do the same for PR builds (as opposed to builds performed on pushes to ordinary branches).

ArSn:
@weirdan good to know - still weird that just one of the build steps failed though.

Btw. I just noticed appveyor fails for a reason since it is running on windows - I was not so clear about that beforehand. Will hopefully open a PR to fix this for windows too tonight.

PR #2898, opened 2020-03-01
Change return type of `range()` from `array<int, ...>` to `list<...>`

jdreesen:
If I'm not mistaken, the `range()` function actually returns a `list<...>`, but currently it's not possible to annotate it as that because `list` is more specific than `array`.

See: https://psalm.dev/r/8c1b886927


muglug:
Thanks!

PR #2903, opened 2020-03-02
\sodium_memzero() sets to null the value of the given reference

LeSuisse:
While it is not mentionned in PHP manual \sodium_memzero() always sets
the given parameter to null [0].

[0] https://github.com/php/php-src/blob/cb933d63c2a6df13d58054da1ea2c48f21a2469e/ext/sodium/libsodium.c#L512


muglug:
Thanks! FYI @paragonie-scott

PR #2917, opened 2020-03-05
Support trailing commas in type aliases

leightonthomas:
https://psalm.dev/r/9358a2939f for an example of the issue

I've copied the original fix (b12f05185a76c30c0d42acd79d2d43f2ffaaea2d) over to where the aliases are parsed and added to the existing test case


muglug:
Thanks!

Issue #2748, opened 2020-02-05
Junit report can't be parsed by gitlab CI


djeux:
When the junit report is generated, it contains an xml that is not compatible with gitlabs. 
Looking closer at the generated file, it has multiple `<testsuite>` elements nested inside each other. 

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
    <testsuite>
        <testsuite>
          .....
        </testsuite>
        <testsuite>
          .....
        </testsuite>
    </testsuite>
</testsuites>
```
instead of the correct one (which is generated by other tools, for example Codesniffer, phpunit)
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
    <testsuite>
        ...
    </testsuite>
    <testsuite>
        ...
    </testsuite>
</testsuites>
```

muglug:
The source is here: https://github.com/vimeo/psalm/blob/master/src/Psalm/Report/JunitReport.php

Would you be able to create a PR to match the expected shape?

cc @m50

djeux:
Will do

m50:
I don't currently use the junit reports generated from psalm, but from my converter that psalm utilized, and those work correctly in my gitlab instance.

Also for reference, multiple nestes test suites is how the phpunit junit report is generated.

I'll test the psalm junit reports in my gitlab, and see if I can figure out what the issue is as well.

djeux:
@m50 true, just checked phpunit, format is the same. Just for some reason gitlab doesn't like the Junit produced by psalm.
What is the converter you use?

m50:
@djeux The converter is one I made: https://github.com/m50/psalm-json-to-junit

It takes in the json reports and outputs junit reports. Those reports get read from our Gitlab just fine (In fact, when I was writing it, I was testing it against gitlab).

Psalm's implementation was based on mine (which is why I was cc'd in), so it's weird if there is a discrepancy.

muglug:
@m50 mind using the Psalm one and seeing if output it at all different? You could presumably check offline.

djeux:
Already did that, the main different I see is in the count of "tests" in the parent testsuite 'tests' attribute.

Psalm one:
```
[...]

PR #2937, opened 2020-03-08
Junit report format fixes

dq5studios:
This fixes the junit report validation issues from #2748 by removing/renaming everything the xsd said is invalid.
The unit test has been updated to validate the output against the junit xsd file to ensure the report is valid.
Info level issues are changed to a skipped element from a failure element to be more accurate.
Code snippets and issue messages are XML encoded.
The junit xsd was checked in as tests/junit.xsd


muglug:
Thanks! Great addition

PR #2944, opened 2020-03-08
Ignore exceptions already caught by previous catch

smelesh:
Psalm incorrectly infers thrown exceptions in case of multiple `catch` sections.

If an exception from `try {}` context can be caught by previous `catch`, Psalm still tries to handle it by the next `catch`, as a result `MissingThrowsDocblock` issued.
Expecting that already caught exception won't appear in possible thrown exceptions of the next catch context, since it's an unreachable state.
Additionally, `$catch_context` should not inherit possibly thrown exceptions of the `$original_context`, it should either rethrow catched exception or suppress them, so only newly analyzed exceptions should be merged to current `$context`.

https://psalm.dev/r/200055ea8e


muglug:
Thanks!

PR #2947, opened 2020-03-09
WIP: Infer re-thrown exception type

smelesh:
Sometimes it's useful to write catch-all statement to handle all exceptions and re-throw them, for example in logging:
```
try {
    method();
} catch (\Throwable $e) {
    // log exception
    throw $e;
}
```

In that case we don't change the type of the re-thrown exception, so there is no need to require for handling base exception type because it will never be thrown inside `try` section.

https://psalm.dev/r/900c99cab3


smelesh:
Closing as this can lead to bugs. Added `\Throwable` to `<ignoreExceptions />` to fix the issue.

PR #2950, opened 2020-03-09
Fix #2914 - falsable functions who only expect to return true now produce errors

### LINKED ISSUE #2914: `@psalm-return true` not checked, while `@psalm-return false` is checked
In https://psalm.dev/r/58faa57709, psalm detects a mismatch between `@psalm-return false` and `return true`.

In https://psalm.dev/r/ddb1fce230, psalm detects a mismatch between `@psalm-return true` and `return false`.

leightonthomas:
```php
<?php

/** @psalm-return true */
function alwaysFalse()
{
    return false;
}
```

will now output a `FalsableReturnStatement` error


muglug:
Thanks!

PR #2955, opened 2020-03-10
Improve parse_url() return type

ShiraNai7:
This PR changes the return type of `parse_url()` with a single argument.

Current:
```
array<'fragment'|'host'|'pass'|'path'|'port'|'query'|'scheme'|'user', mixed>|false
````

New:
```
array{
    fragment?: string,
    host?: string,
    pass?: string,
    path?: string,
    port?: int,
    query?: string,
    scheme?: string,
    user?: string
}|false
```

This allows the specific key types to be inferred. Currently they're all mixed.

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/ParseUrlReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/ParseUrlReturnTypeProvider.php
@@ -112,28 +112,30 @@ class ParseUrlReturnTypeProvider implements \Psalm\Plugin\Hook\FunctionReturnTyp
             return $nullable_string_or_int;
         }
 
-        $component_key_type = new Type\Union([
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by muglug** [comment]
Could you reinstate the tests you removed? They still need to pass

**Comment by ShiraNai7** [comment]
I can but why should this pass the tests?

```php
function bar(string $s) : string {
    $parsed = parse_url($s);

    return $parsed["host"];
}

bar('index.html'); // Notice: Undefined index: host
```

muglug:
> I can but why should this pass the tests?

It doesn't look like that should – but instead it can be moved into the expected failures. That's what I've done in the past when a bugfix altered the running of an existing test.

ShiraNai7:
@muglug I've restored the original tests and moved some of them to invalid code cases.

muglug:
Thanks!

Issue #2953, opened 2020-03-09
Psalm breaks horribly if line breaks are CRLF


ArSn:
### Scenario
I am currently fiddling around a bit with [ubuntu for windows](https://ubuntu.com/tutorials/tutorial-ubuntu-on-windows#1-overview) and I thought psalm would be a great project to work on with this. I installed all composer packages and required PHP extensions and even the suggested `ext-igbinary`.

It took me a few days to realize that this was not the issue, but the fact that I checked out psalm with PHPStorm on windows and was using it in the ubuntu command line. This lead to the line breaks in the code being CRLF (aka `\r\n`) instead of only LF (aka `\n`).

To help debug this I created some examples ( ⚠️ careful, they only break if your linebreaks are CRLF, which is obviously not shown in the code below):

### How to reproduce
code:
```php
<?php

namespace Some\Weird\Space;


class A {
    /**
     * @param string|null $bla
     * @return array{
     *          0: \Some\Weird\Space\A|null,
     *          1: null
     *          }
     */
    public function doSomething($bla): array
    {
        echo '1' . (string)$bla;
        return [null, null];
    }
}

$bla = new A();
$bla->doSomething('2');
```

output:
```
Scanning files...
Analyzing files...

E

' in docblock for Some\Weird\Space\A::doSomething type 'Some\Weird\Space\null
    /**
     * @param string|null $bla
     * @return array{
     *          0: \Some\Weird\Space\A|null,
     *          1: null
     *          }
     */
    public function doSomething($bla): array


------------------------------
1 errors found
------------------------------

Checks took 1.40 seconds and used 117.811MB of memory
Psalm was able to infer types for 100% of the codebase
```

### Expected output
just leaving this here as a reference for the full bug report, and it works fine if the linebreaks are LF only:
```
Scanning files...
Analyzing files...

░

------------------------------
No errors found!
------------------------------

Checks took 1.41 seconds and used 117.818MB of memory
Psalm was able to infer types for 100% of the codebase
```

### Remarks
My guess is that this has something to do with the tokenization going wrong and I have it in the back of my mind that there was some piece of code that I stumbled upon during debugging that was just searching for `\n` and nothing else, which sounds like it could have to do with it, but I could not find it again yet.

That being said, my debugging efforts only led to this guess so far and I can not really put my finger on it as you can probably tell.

[...]

PR #2960, opened 2020-03-10
Added tests for docblock breaking CRLF 

ArSn:
I went ahead and added some tests for the fix that was provided to #2953 in 3cf5621. 

Doing this, I noticed that the description was not parsed properly as well, which is why I went ahead and refactored the "ignorance" to a more drastric approach by just removing all carriage returns in the very beginning.

@muglug This lead to me reverting what you did in 3cf5621. I reckon that this way the lower code part never knows of the carriage return in the first place and therefore the foreach does not get the "wrong `$line_offset `" since it never knows of a different length including the `\r` to begin with - would you agree?

### DIFF
--- a/src/Psalm/DocComment.php
+++ b/src/Psalm/DocComment.php
@@ -183,7 +183,7 @@ class DocComment
         $docblock = $docblock->getText();
 
         // Strip off comments.
-        $docblock = trim($docblock);
+        $docblock = str_replace("\r", '', trim($docblock));
 
         $docblock = preg_replace('@^/\*\*@', '', $docblock);
         $docblock = preg_replace('@\*\*?/$@', '', $docblock);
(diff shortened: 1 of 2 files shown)


muglug:
No, the `\r` need to be passed through – otherwise the length (and thus, importantly, the character offsets) are not preserved.

I think my fix was correct, and already tested in Appveyor. Psalm doesn't use the docblock descriptions anywhere.

ArSn:
I see. Hmm, any idea how one could build a suitable test for this? I would add this then and work on it again and reopen the PR once its smooth.

muglug:
CRLF analysis is now tested in Appveyor – I don't think additional tests are necessary

ArSn:
Understood. However, it looks to me that the test that currently fails in appveyor is a subsequent failure, meaning that the test is not actually for that specific case - I would call it "implicitly tested". Would it make sense to you to just add the tests that explicitly test the desired behaviour: [Those two](https://github.com/vimeo/psalm/pull/2960/files#diff-a9ea02292461945115d7818ad50aa9f0R10-R44)? (and to ignore the case with the description since as you said psalm does not use it anywhere)

muglug:
No – almost every test that runs on Appveyor is explicitly testing crlf code, so I think this is safe to close.

ArSn:
ok then

PR #2965, opened 2020-03-12
`ignoreInternalFunctionFalseReturn` default to false in XSD, but docu…

maks-rafalko:
See https://github.com/vimeo/psalm/blob/27598f508ee44911e89305da0c76fb90e9d9b5e4/config.xsd#L46

### DIFF
--- a/docs/running_psalm/configuration.md
+++ b/docs/running_psalm/configuration.md
@@ -220,7 +220,7 @@ When `true`, Psalm will check that the developer has caught every exception in g
   ignoreInternalFunctionFalseReturn="[bool]"
 >
 ```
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

maks-rafalko:
Thank you @muglug

jarstelfox:
Hey, I think this is possible wrong!

See here:

https://github.com/vimeo/psalm/blob/master/src/Psalm/Config.php#L324-L332

Both default to `true` in the code!

The docs say differently:

https://github.com/vimeo/psalm/blob/master/docs/running_psalm/configuration.md#L216-L232

and the XSD says so too:
https://github.com/vimeo/psalm/blob/master/config.xsd#L46-L47

Issue #997, opened 2018-09-14
Feature request: New psalter options


ADmad:
It would be nice if `psalter` had options for the following:
1. Use FQCN in docblocks  (`@return \Cake\Foo` instead of `@return Foo`).
2. Not add new lines above `@return` and `@throws` tags.

orklah:
Probably but the issue is that Psalm sometimes need to rewrite a whole docblock and will change the formatting. This lead to Psalter diff that are needlessly long and makes review a pain

What could be done otherwise is to make Psalter call whatever CS tool the user could have automatically before showing the results. This could be harder...

Note: the line feed was solved here, this issue only concerns the FQCN now

PR #2969, opened 2020-03-12
Dont add newline before return

ArSn:
This implements one of the features requested in #997

### DIFF
--- a/src/psalter.php
+++ b/src/psalter.php
@@ -1,6 +1,7 @@
 <?php
 require_once('command_functions.php');
 
+use Psalm\DocComment;
 use Psalm\Internal\Analyzer\ProjectAnalyzer;
 use Psalm\Config;
 use Psalm\IssueBuffer;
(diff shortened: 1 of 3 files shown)


src/psalter.php
```diff
@@ -31,6 +32,7 @@
     'plugin:', 'issues:', 'list-supported-issues', 'php-version:', 'dry-run', 'safe-types',
     'find-unused-code', 'threads:', 'codeowner:',
     'allow-backwards-incompatible-changes:',
+    'doc-block-add-new-line-before-return:',
```
muglug: Could you change this command to `add-newline-between-docblock-annotations` please?

And then slightly change your implementation, because now you wouldn't need the explicit check for `'return'` – just this flag.

I'm thinking that anyone who doesn't want newlines before `@return`  _also_ doesn't want newlines before `@throws`

ArSn: Do you ever sleep? ;) 

Yes, totally makes sense and I will do so tomorrow. Just had one too many beers today to complete it now.

muglug: Yes, I sleep in New York!

muglug:
Thanks!

PR #2986, opened 2020-03-18
Fix #2984 - Document Generator type syntax

### LINKED ISSUE #2984: Documentation: Generator<array-key, mixed, mixed, void>?
Psalm infers Generator return types like `Generator<array-key, mixed, mixed, void>`.

The first two parameters probably refer to the `yield` key and value, but what are the third and fourth parameter for?

(Unfortunately the Psalm documentation doesn't explain this.)

caugner:
Fixes #2984.


muglug:
Thanks!

PR #2997, opened 2020-03-19
$ignore parameter of \sodium_base642bin() is optional

LeSuisse:
See https://github.com/php/php-src/blob/php-7.3.15/ext/sodium/libsodium.c#L2775


muglug:
thanks!

PR #3002, opened 2020-03-21
Add \Error stub

kelunik:
Otherwise, errors are generated if \Error is extended: https://psalm.dev/r/8169465beb


muglug:
Thanks!

PR #3013, opened 2020-03-24
Add psalm-property-read and psalm-property-write

### LINKED ISSUE #3012: Feature request: add psalm-property-read/psalm-property-write like psalm-property
The tags psalm-property-read and psalm-property-write should be added, in the same vein as @psalm-var, @psalm-param and @psalm-return (https://psalm.dev/docs/annotating_code/supported_annotations/)

iggyvolz:
Fixes #3012

### DIFF
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -66,7 +66,7 @@ function addFoo(?string &$s) : void {
 }
 ```
 
-### `@psalm-var`, `@psalm-param` and `@psalm-return`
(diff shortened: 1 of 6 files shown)


muglug:
Thanks!

iggyvolz:
I think I missed one thing - in DocComment.php, the array on line 278 should match the one on 150 causing them to fail with "Unrecognized annotation" if parsePreservingLength is used instead of parse - do you want me to make another PR for that?

muglug:
Nah, I’ve got it

PR #3025, opened 2020-03-26
Fix grammar nit in "is can never happen" message

TysonAndre:
Should either be "is impossible" or "can never happen"


muglug:
thanks!

PR #3031, opened 2020-03-28
Allow netresearch/jsonmapper ^2.0

kubawerlos:
The bumping major in the library is because of [Make "createInstance" protected instead of public](https://github.com/cweiske/jsonmapper/blob/v2.0.0/ChangeLog#L5).


muglug:
Thanks!

PR #3046, opened 2020-04-01
Fix return type of \zip_open()

LeSuisse:
This function can return an int [0] or false [1] in case of error.

The return value of ZipArchive has also been fixed, it returns either a
boolean or an int [2][3].

Close #3209.

[0] https://github.com/php/php-src/blob/php-7.4.4/ext/zip/php_zip.c#L1176
[1] https://github.com/php/php-src/blob/php-7.4.4/ext/zip/php_zip.c#L1160
[3] https://github.com/php/php-src/blob/php-7.4.4/ext/zip/php_zip.c#L1483
[4] https://github.com/php/php-src/blob/php-7.4.4/ext/zip/php_zip.c#L1478


muglug:
Thanks!

PR #3051, opened 2020-04-02
Added json_decode to CoreGenericFunctions.phpstub

vudaltsov:


muglug:
What's the point of this exactly? If you just want to change its return type, that can be done in `CallMap.php`.

muglug:
Also the definition could be improved with a conditional type (which will be officially included in the next release):

```
/** 
 * @template TAssoc as bool
 * @param TAssoc $assoc
 * @return (TAssoc is true ? null|scalar|array : null|scalar|array|stdClass)
 */
function json_decode(
    string $json,
    bool $assoc = false, 
    int $depth = 512,
    int $options = 0
): null|scalar|array|stdClass
{
}
```

vudaltsov:
@muglug , thank you! I will update the PR soon.

muglug:
I did this myself in 4b36c4e

muglug:
But then I ran this against Vimeo's source code, and it introduced over 2,000 false-positive issues.

That's too high for my comfort, sorry.

vudaltsov:
No problem, thank you! I will do a stub in my project then.

vudaltsov:
Just as a matter of interest, what kind of issues did it introduce? This change specifies the mixed type, it seems to be pretty harmless, isn't it?

muglug:
> Just as a matter of interest, what kind of issues did it introduce?

Lots of the form

```php
function returnsArray() : array {
  $json = file_get_contents('/json_array_returning_endpoint');
  return json_decode($json);
}
```

With the more specific return type Psalm was complaining that the array return type was incorrect (as the JSON string could feasibly represent `null` or some other value.

But in fact that endpoint _never_ returns anything but an array. An `is_array($json)` check would only satisfy Psalm, and that's not great for Psalm's end-users.

muglug:
> I will do a stub in my project then.

I might also add it to a "strict" Psalm plugin at some point in the future.

vudaltsov:
I use strict mode in my project, so I have to add `@psalm-var array` anyway 😊.

PR #3056, opened 2020-04-02
Add support for resolving promises in Amp's coroutines

kelunik:
This is currently not fully working as intended, but mostly works. See comments for more information.


muglug:
I'm not too keen on baking in support for a given implementation at a particular point in time.

But I am a fan of Amp, obviously, so I've added support for a `@psalm-yield` annotation that you can add to `Promise` – I've added it already to the stub that Psalm uses in af4a7ca.

kelunik:
Thanks! I've had pretty much the same idea regarding the annotation.

muglug:
The only downside is that things would be wrong in Psalm if yielding a `Promise` while not inside an `Amp` call, but 🤷‍♂

kelunik:
@muglug Indeed, but I think anyone using Amp is fine with that risk compared to the value it provides. :+1:

kelunik:
@muglug Unfortunately, yielding an `Amp\Promise<string>[]` results in the type being inferred as `Amp\Promise<string>[]`, so it's still `TValue == TSend` there. I'm not sure how that can be supported in a generic way. Should I open an issue?

muglug:
Yes please

PR #3064, opened 2020-04-04
Support intersection types for object-like arrays

lhchavez:
This change allows expressing the intersection of multiple object-like
arrays. The resulting type will contain the (merged via intersection)
properties of all the object-like arrays involved in the intersection.

This should allow to express the resulting type of an `array_merge()`
operation, which is tricky to do in the context of a templated type.

### DIFF
--- a/tests/TypeParseTest.php
+++ b/tests/TypeParseTest.php
@@ -216,6 +216,66 @@ class TypeParseTest extends TestCase
         $this->assertSame('iterable<mixed, A>&iterable<mixed, B>', (string) Type::parseString('iterable<A>&iterable<B>'));
     }
 
+    /**
+     * @return void
+     */
+    public function testIntersectionOfObjectLike()
+    {
+        $this->assertSame('array{a: int, b: int}', (string) Type::parseString('array{a: int}&array{b: int}'));
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


tests/TypeParseTest.php
```diff
@@ -216,6 +216,23 @@ public function testIntersectionOfIterables()
         $this->assertSame('iterable<mixed, A>&iterable<mixed, B>', (string) Type::parseString('iterable<A>&iterable<B>'));
     }
 
+    /**
+     * @return void
+     */
+    public function testIntersectionOfObjectLike()
+    {
+        $this->assertSame('array{a: int, b: int}', (string) Type::parseString('array{a: int}&array{b: int}'));
```
lhchavez, line 224: hmm I just realized that there is not enough test coverage. Let me add a few more cases.

lhchavez: all right, i did find a couple of issues ^^;; it should be better now.

lhchavez:
After a bit more consideration, I realized that `Type::combineUnionTypes()` was the wrong thing to do and instead switched to using `Type::intersectUnionTypes()`.

muglug:
Thanks!

PR #3067, opened 2020-04-05
ZipArchive::open() does not return resource

mpesari:
I think https://github.com/vimeo/psalm/pull/3046 accidentally added `ZipArchive::open()` to return `resource`, which is against what the original pull request description said.

Source: https://github.com/php/php-src/blob/php-7.4.4/ext/zip/php_zip.c#L1431
Manual: https://www.php.net/manual/en/ziparchive.open.php
Playground: https://psalm.dev/r/e76cb4e762


muglug:
Thanks!

PR #3069, opened 2020-04-05
Fix maxStringLength handling

m0003r:
There was some hard limits defined in StatementsAnalyzer (30 for strings and 50 for concat), but Type::getString already respects maxStringLength in config so there is no need to additional checks.

This fix comes from developing SQL parsing plugin. I plan to infer `PDOStatement::fetch` return types via parsing SQL query. Also I found config variable `parseSql`. What is it intended for? I haven't found any usages...


weirdan:
Some time ago Psalm did parse strings and marked them as sql-string when they looked like a query, but then it was decided it's more suitable for a plugin, like yours.

m0003r:
So may there must be some deprecation message about that setting?

muglug:
Yes, it'll be removed in version 4 but I didn't think it was worth adding a deprecation message.

muglug:
But anyway, thanks!

PR #3076, opened 2020-04-05
Run every project as a separate step on CircleCI

weirdan:
It's easier to see which project is failing this way.

image


muglug:
Thanks!

PR #3080, opened 2020-04-06
Implement @psalm-trace annotation, fixes #3079

### LINKED ISSUE #3079: @psalm-trace to show what the analyzer is thinking at particular point in code
Would it be possible to add an annotation `@psalm-trace $x` that would output the inferred type of `$x` to the stdout when running static analyzer?

Something like

```
function f(int $x) {
    /** @psalm-trace $x */
    $y = $x . 'a';
}
```

```
function.php:323 $x: int
```

Thank you!

m0003r:
I have same thoughts about `@psalm-trace` as in #3079 , so I just implemented is as simple as it possible.

Have I done it correctly?

### DIFF
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -385,6 +385,18 @@ echo $counter->count; // outputs 1
 $counter->count = 5; // This will fail, as it's mutating a property directly
 ```
 
+### `@psalm-trace`
+
+You can use this annotation to trace inferred type (applied to the *next* statement).
+
+```php
+<?php
+
+/** @psalm-trace $username */ 
+$username = $_GET['username']; // prints something like "test.php:4 $username: mixed"
+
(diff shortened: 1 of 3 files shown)


muglug:
Very nice, thanks!

weirdan:
@m0003r outputting things directly will break most structured reports (xml, json, etc). It already breaks on psalm.dev when using `@psalm-trace`: https://psalm.dev/r/d9107243d3

I would suggest emitting a special kind of issue instead.

m0003r:
@weirdan oh, I haven't thought about it, I'll do it soon.
Only one question: which `ERROR_LEVEL` should have that special issue?

weirdan:
Lowest possible, I suppose. You don't want to break CI builds if you left the annotation in the source code accidently.

m0003r:
@weirdan did it in #3106

PR #3085, opened 2020-04-06
Do not reconcile outer_vars after analyzing If if there is continue inside

m0003r:
https://psalm.dev/r/2e491782ca

any other comparison works good, but in that case it fails.

I have tried to fix it by myself, but I'm not sure if I have done it right. I have added some tests. and there was no new failing tests, but I'm still insecure about that change.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Block/IfAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/IfAnalyzer.php
@@ -900,7 +900,8 @@ class IfAnalyzer
             }
         }
 
-        if ($has_leaving_statements && !$has_break_statement && !$stmt->else && !$stmt->elseifs) {
+        if ($has_leaving_statements && !$has_break_statement && !$has_continue_statement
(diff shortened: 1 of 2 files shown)


m0003r:
Why @psalm-github-bot doesn't do that?

<details>
<summary>https://psalm.dev/r/2e491782ca</summary>

```php
<?php

for ($i = 0; $i < 2; $i++) {
    if ($i === 0) {
        continue;
    }
}
```

```
Psalm output (using commit 11edb22): 

ERROR: RedundantCondition - 4:9 - Found a redundant condition when evaluating $i and trying to reconcile type 'int(0)' to int(0)

ERROR: InvalidOperand - 3:22 - Cannot perform a numeric operation with a non-numeric type empty
```
</details>

weirdan:
> Why @psalm-github-bot doesn't do that?

It watches issues only, an omission on my part.

muglug:
I went with a super-simple fix, generalising `for ($i = 0` to `for ($i = $some_int`

PR #3093, opened 2020-04-07
Added `ReflectionProperty#getType()` to psalm type stubs for PHP 7.4+

Ocramius:
Ref: https://github.com/Ocramius/ProxyManager/pull/585 (fails build there, noticed missing types)

### DIFF
--- a/src/Psalm/Internal/CallMap_74_delta.php
+++ b/src/Psalm/Internal/CallMap_74_delta.php
@@ -17,6 +17,7 @@ return [
 'new' => [
     'password_hash' => ['string|null', 'password'=>'string', 'algo'=>'int|string|null', 'options='=>'array'],
     'password_needs_rehash' => ['bool', 'hash'=>'string', 'algo'=>'int|string|null', 'options='=>'array'],
+    'ReflectionProperty::getType' => ['?ReflectionType'],
 ],
 'old' => [
     'password_hash' => ['string|false', 'password'=>'string', 'algo'=>'int', 'options='=>'array'],
--- a/src/Psalm/Internal/CallMap.php
+++ b/src/Psalm/Internal/CallMap.php
@@ -11443,6 +11443,7 @@ return [
 'ReflectionProperty::getDocComment' => ['string|false'],
 'ReflectionProperty::getModifiers' => ['int'],
 'ReflectionProperty::getName' => ['string'],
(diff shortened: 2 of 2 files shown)


src/Psalm/Internal/CallMap_74_delta.php
```diff
@@ -21,5 +21,6 @@
 'old' => [
     'password_hash' => ['string|false', 'password'=>'string', 'algo'=>'int', 'options='=>'array'],
     'password_needs_rehash' => ['bool', 'hash'=>'string', 'algo'=>'int', 'options='=>'array'],
+    'ReflectionProperty::getType' => ['?ReflectionType'],
```
weirdan, line 24: I think it should be in the `new` section, as it was added, not removed.

Ocramius:
Ooof, just noticed my mistake there - indeed, should be in "new", but will
fix tomorrow.
Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

weirdan:
> but will fix tomorrow.

That won't (and didn't) break anything, 7.4 delta is ignored anyway 😈

weirdan:
> 7.4 delta is ignored anyway 

Fixed in #3094

src/Psalm/Internal/CallMap_74_delta.php
weirdan: Also you need to add this to `CallMap.php` itself. Deltas are used to patch the callmap for older versions.

Ocramius: Adjusted both here and in `CallMap.php` :+1:

muglug:
Thanks!

Ocramius:
@muglug would it be possible to cut a patch release with this in? Not sure if `master` is in releasable state.

muglug:
I'm still working out the ergonomics of `mixed` unions inside arrays, as there are a number of false-positives in an internal [PHPUnit method](https://github.com/sebastianbergmann/phpunit/blob/30d5173bd35d51c48e9ed75b5732d567e0916ae6/src/TextUI/TestRunner.php#L122) that passes in an array that actually should be an object.

muglug:
Basically this: https://psalm.dev/r/d4e71b8666

muglug:
Remaining issues have been ironed out, a release should happen within the next 24 hours

PR #3094, opened 2020-04-07
Bump callmap version to 7.4

weirdan:
The content of the callmap corresponds to 7.4, but internally the version was set to 7.3. This caused 7.4 delta to be ignored, and 7.4 types to be used for 7.3.


muglug:
Thanks!

PR #3100, opened 2020-04-08
Suppress choco install progress

weirdan:
This should make appveyor logs much more readable. Currently they have thousands of lines like
```
Progress: Downloading php 7.4.3... 98%
Progress: Downloading php 7.4.3... 98%
Progress: Downloading php 7.4.3... 98%
```


muglug:
Thanks, I hated that

PR #3106, opened 2020-04-08
@psalm-trace as a low-level issue

m0003r:
`@psalm-trace` is now a specific low-level issue, because plain debug print [breaks structured output](https://github.com/vimeo/psalm/pull/3080#issuecomment-610264757)


muglug:
Thanks!

PR #3108, opened 2020-04-08
Allow plain assertions (@psalm-assert) about $this (fixes #3105)

### LINKED ISSUE #3105: Assertions about $this doesn't always work
https://psalm.dev/r/64e4c731f3

I persume there is in fact two issues.

1. `assertE` is working if it's not about `$this`:
https://psalm.dev/r/926eb03f6d

2. if there is `assert-if-true` AND `assert-if-false`, only the `-if-false` works

I would like to try it by myself first

m0003r:
I know it looks stupid, but I need it for my SQL plugin (if `PDOStatements::execute` haven't thrown any exceptions, then it can be fetched etc.)

***UPD:*** combining `@psalm-assert` with `@psalm-assert-if-false` still don't work for `$this`, so still WIP
***UPD2:*** combining two interface assertions [like this](https://psalm.dev/r/0316ca8d0a) doesn't work, I'll try to fix it

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/MethodCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/MethodCallAnalyzer.php
@@ -21,8 +21,9 @@ use Psalm\Issue\UndefinedMethod;
 use Psalm\IssueBuffer;
 use Psalm\Type;
 use Psalm\Type\Atomic\TNamedObject;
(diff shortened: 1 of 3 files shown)


m0003r:
Finally fixed it

P.S. Psalm is wonderful

m0003r:
Fixed combining `@psalm-assert ... $this` and `@psalm-assert-if-true ... $this` on the same method.

PR #3114, opened 2020-04-09
Update mysqli property maps

mpesari:
From https://github.com/phan/phan/blob/ca92e03c122dd89d42b7844d4d17645af44eaad1/src/Phan/Language/Internal/PropertyMap.php

### DIFF
--- a/src/Psalm/Internal/PropertyMap.php
+++ b/src/Psalm/Internal/PropertyMap.php
@@ -216,14 +216,6 @@ return [
     'normalizer' => [
         'name' => 'string',
     ],
-    'mysqli-driver' => [
-        'client-info' => 'string',
-        'client-version' => 'string',
-        'driver-version' => 'string',
-        'embedded' => 'string',
-        'reconnect' => 'bool',
-        'report-mode' => 'int',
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/PropertyMap.php
```diff
@@ -256,6 +248,62 @@
...
+        'error_list' => 'array',
+        'field_count' => 'int',
+        'host_info' => 'string',
+        'info' => 'string',
+        'insert_id' => 'mixed',
+        'protocol_version' => 'string',
+        'server_info' => 'string',
+        'server_version' => 'int',
+        'sqlstate' => 'string',
+        'stat' => 'mixed',
```
staabm, line 268: the phpdocs do not define a stat property...? Should it be dropped ?

mpesari: Yeah, looks like it's actually a method: https://www.php.net/manual/en/mysqli.stat.php

staabm: https://github.com/mpesari/psalm/pull/1

PR #3115, opened 2020-04-09
Fix $assertion_var_id calculation in method calls

m0003r:
Continuing #3105 

changed order to get [code like this](https://psalm.dev/r/f1456d620e) working.

Assertion about `$this` must be applied to `$thisName` even when `vars_in_scope['$this']` exists.


muglug:
Thanks!

PR #3120, opened 2020-04-11
Add missing "--output-format" example

Chemaclass:
There is another available format ("text") that provides a simple output, and different from all other formats.

This is an example:

```
$ vendor/bin/psalm --output-format=text 
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░E░
FileExample.php:38:18:error - TypeDoesNotContainType: array<array-key, mixed> is always defined and non-null
```

### DIFF
--- a/src/command_functions.php
+++ b/src/command_functions.php
@@ -336,7 +336,7 @@ Output:
 
     --output-format=console
         Changes the output format.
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

Issue #2749, opened 2020-02-05
Consider property null check in block condition when determining type of property getter result


caugner:
Since `Foo::hasBar()` checks `Foo::$bar` in the following snippet, I would expect Psalm to derive that `$bar` cannot be `null` in the block.

## Snippet

(https://psalm.dev/r/e583671af1)

```php
class Foo {
  /**
    * @var string|null
    */
  private $bar;
  
  public function __construct(?string $bar = null) {
    $this->bar = $bar;
  }
  
  public function hasBar(): bool
  {
    return !is_null($this->bar);
  }
  
  public function getBar(): ?string
  {
    return $this->bar;
  }
}

$foo = new Foo("nullable");

if ($foo->hasBar()) {
  $bar = $foo->getBar();
  echo strlen($bar);
}
```

## Output

```
Psalm output (using commit bbdb125):

ERROR: PossiblyNullArgument - 28:15 - Argument 1 of strlen cannot be null, possibly null value provided
```

muglug:
In Psalm, with some notable exceptions, what happens in functions stays in functions.

I think you can achieve what you want with a `@readonly` property: https://psalm.dev/r/626f190105

weirdan:
Or with some generic magic: https://psalm.dev/r/1dc90247a2

PR #3122, opened 2020-04-12
Getters automagic

m0003r:
When method is a plain getter, we can:

1. correct method return type if property type is known
2. auto assert-if-true that corresponding property is not falsy

Covers cases like:

```php
if ($a->getSomething()) {
   doesNotAcceptNull($a->getSomething())
}
```

and

```php
if ($a->hasSomething()) {
   doesNotAcceptNull($a->getSomething())
}
```

The last one inspired by #2749


muglug:
This is sadly not sound e.g. https://psalm.dev/r/502aa75a32

m0003r:
@muglug may be additional checking `!$storage->overridden_somewhere && !$storage->overridden_downstream` can fix that?

muglug:
Cool, then I'm happy.

m0003r:
Thanks, that fix a lot of false-positives with ActiveRecord-ish objects.

PR #3161, opened 2020-04-16
Added new language server options and functionality.

pristinesource:
Added new extended diagnostic code information to the language server.
 -- It must be enabled via a command line switch.
Added telemetry data for language server initialization and operation.
Added verbose log messages for language server.
 -- It must be enabled via a command line switch.

I will comment on the code changes directly in the diff.

### DIFF
--- a/src/Psalm/Internal/Analyzer/IssueData.php
+++ b/src/Psalm/Internal/Analyzer/IssueData.php
@@ -99,6 +99,12 @@ class IssueData
     public $error_level;
 
     /**
+     * @var int
+     * @readonly
+     */
+    public $shortcode;
+
+    /**
      * @var string
      * @readonly
      */
@@ -157,6 +163,7 @@ class IssueData
         $this->snippet_to = $snippet_to;
         $this->column_from = $column_from;
         $this->column_to = $column_to;
+        $this->shortcode = $shortcode;
         $this->error_level = $error_level;
         $this->link = 'https://psalm.dev/' . \str_pad((string) $shortcode, 3, "0", \STR_PAD_LEFT);
     }
--- a/src/Psalm/Internal/LanguageServer/LanguageClient.php
+++ b/src/Psalm/Internal/LanguageServer/LanguageClient.php
@@ -2,6 +2,7 @@
 declare(strict_types = 1);
 namespace Psalm\Internal\LanguageServer;
 
+use Amp\Promise;
 use JsonMapper;
 
 /**
@@ -16,11 +17,48 @@ class LanguageClient
      */
     public $textDocument;
 
+    /**
+     * The client handler
+     *
+     * @var ClientHandler
+     */
+    private $handler;
+
     public function __construct(ProtocolReader $reader, ProtocolWriter $writer)
     {
-        $handler = new ClientHandler($reader, $writer);
+        $this->handler = new ClientHandler($reader, $writer);
         $mapper = new JsonMapper;
 
-        $this->textDocument = new Client\TextDocument($handler, $mapper);
+        $this->textDocument = new Client\TextDocument($this->handler, $mapper);
+    }
+
+    /**
+     * Send a log message to the client.
+     *
+     * @param string $message The message to send to the client.
+     * @psalm-param 1|2|3|4 $type
+     * @param integer $type The log type:
+     *  - 1 = Error
+     *  - 2 = Warning
+     *  - 3 = Info
+     *  - 4 = Log
+
+     * @return Promise<void>
+     */
+    public function logMessage(string $message, int $type = 4, string $method = 'window/logMessage'): Promise
+    {
+        // https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_logMessage
+
+        if ($type < 1 || $type > 4) {
+            $type = 4;
+        }
+
+        return $this->handler->notify(
+            $method,
+            [
+                'type' => $type,
+                'message' => $message
+            ]
+        );
     }
 }
--- a/src/Psalm/Internal/LanguageServer/LanguageServer.php
+++ b/src/Psalm/Internal/LanguageServer/LanguageServer.php
@@ -383,7 +383,8 @@ class LanguageServer extends AdvancedJsonRpc\Dispatcher
                         'Psalm'
                     );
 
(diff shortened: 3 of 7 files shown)


src/Psalm/Internal/Analyzer/IssueData.php
```diff
@@ -98,6 +98,12 @@ class IssueData
      */
     public $error_level;
 
+    /**
+     * @var int
+     * @readonly
+     */
+    public $shortcode;
```
pristinesource, line 105: Added this so it can be used to send the error code to the client.

src/Psalm/Internal/LanguageServer/LanguageClient.php
```diff
@@ -16,11 +17,48 @@ class LanguageClient
...
+     * @psalm-param 1|2|3|4 $type
+     * @param integer $type The log type:
+     *  - 1 = Error
+     *  - 2 = Warning
+     *  - 3 = Info
+     *  - 4 = Log
+
+     * @return Promise<void>
+     */
+    public function logMessage(string $message, int $type = 4, string $method = 'window/logMessage'): Promise
```
pristinesource, line 48: This method allows us to send notifications with the LogMessageParams type as the params for the message.  The method is customizable, because we use this same method to send telemetry data using the `telemetry/event` method which can use anything for its params.

src/Psalm/Internal/LanguageServer/LanguageServer.php
```diff
@@ -200,16 +202,25 @@ public function initialize(
         return call(
             /** @return \Generator<int, true, mixed, InitializeResult> */
             function () use ($capabilities, $rootPath, $processId) {
+                $this->verboseLog("Initializing...");
+                $this->clientStatus('initializing');
```
pristinesource, line 206: These client statuses should be known strings, so the client can use them and provide UI updates as needed, such as status appropriate icons, or extended status messages.

src/Psalm/Internal/LanguageServer/LanguageServer.php
```diff
@@ -352,14 +373,34 @@ function (IssueData $issue_data) use ($file_path) : Diagnostic {
...
+                        // Added in VSCode 1.43.0 and will be part of the LSP 3.16.0 standard.
+                        // Since this new functionality is not backwards compatible, we use a
+                        // configuration option so the end user must opt in to it using the cli argument.
+                        // https://github.com/microsoft/vscode/blob/1.43.0/src/vs/vscode.d.ts#L4688-L4699
+
+                        /** @psalm-suppress InvalidPropertyAssignmentValue */
+                        $diagnostic->code = [
+                            "value" => $code,
+                            "target" => $issue_data->link,
+                        ];
```
pristinesource, line 396: here we send the detailed diagnostic code with the help link.

src/psalm-language-server.php
```diff
@@ -146,11 +148,22 @@ function ($arg) use ($valid_long_options, $valid_short_options) {
...
 HELP;
 
     exit;
 }
 
+if (array_key_exists('v', $options)) {
+    echo 'Psalm ' . PSALM_VERSION . PHP_EOL;
+    exit;
+}
+
```
pristinesource, line 166: moved this up to allow for an early exit since they just want the version.  There is no need to check other stuff before this.

pristinesource:
I have a PR for the vscode extension that uses these updates so you can see how they would be used.
https://github.com/psalm/psalm-vscode-plugin/pull/12

pristinesource:
Oh, it looks like I forgot to update the tests ... I will  get that fixed up.

muglug:
Thanks! Any reason you need the short code? Won’t the `uri` field provide all the data you need there?

pristinesource:
@muglug, yes vscode (and other IDEs) need the error code to associate with the error. Most of the tim and IDE (vscode does this) requires an error code to attach the link to in the UI.

Here is an example of what it looks like in VSCode (I had to redact some areas of the UI because they are for my work):

Screenshot 2020-04-16 13 04 10

pristinesource:
The failing test with `AppVeyor` is not an issue with my code, it looks like it is having trouble uploading a file. 🤔

pristinesource:
@muglug, also if you prefer, we can just make the error code sent to the client be the link, so like in that screenshot it would look more like so:
`Method [...] does not exist Psalm(https://psalm.dev/022) [32, 57]`
And it would still be clickable.

src/Psalm/Internal/LanguageServer/LanguageServer.php
pristinesource: @muglug here is where it uses the short code to sent to the client.  Right now it pads it to 3 chars and prepends 'PS' to it so a code of 22 would be 'PS022'.  We can make it be anything you like, even making the code sent to the client be the link also; we would just set the link to both "value" and "target".  We need to send the extended code structure instead of just a string in order to make it clickable in the IDE.

pristinesource:
Hmm, the CircleCI tests didn't actually show any tests that failed (some were skipped), but it says it failed 🤔

weirdan:
> Hmm, the CircleCI tests didn't actually show any tests that failed (some were skipped), but it says it failed thinking

It was killed while generating code coverage report, possibly by OOM killer:

> Received "killed" signal

muglug:
> we can just make the error code sent to the client be the link, so like in that screenshot it would look more like so

That sounds preferable – at some point the protocol will support links directly, but until then...

pristinesource:
@muglug ok, I will get that changed to be the link for both the value and the link.  The protocol will support links directly in LSP 3.16.0 which is what I am implementing here.  That is how it is going to be structured in the 3.16.0 specification 😉 .

src/Psalm/Internal/LanguageServer/LanguageServer.php
```diff
@@ -383,7 +383,8 @@ function (IssueData $issue_data) use ($file_path) : Diagnostic {
                         'Psalm'
                     );
 
-                    $code = 'PS' . \str_pad((string) $issue_data->shortcode, 3, "0", \STR_PAD_LEFT);
+                    //$code = 'PS' . \str_pad((string) $issue_data->shortcode, 3, "0", \STR_PAD_LEFT);
+                    $code = $issue_data->link;
```
pristinesource, line 387: Ok, the code now sent is the help link instead of the formatted short code.

muglug:
Thanks, these are wonderful additions!

PR #3171, opened 2020-04-17
Mark throwable methods as pure

greg0ire:


greg0ire:
Some of the failing tests look like they are cause by my changes, I have to look into that.

muglug:
Thanks! Looks like the failures were just due to my earlier bad merge

src/Psalm/Internal/Stubs/CoreImmutableClasses.phpstub
```diff
@@ -61,21 +91,33 @@ class Exception implements Throwable
 
     /**
      * @psalm-mutation-free
-     * @return int|string
      */
-    public final function getCode(){}
+    public final function getCode(): int {}
```
andrew-demb, line 95: `getCode()` can return string

greg0ire: Are you sure? https://3v4l.org/cJHtA

andrew-demb: Yes. PDO throw exceptions with string codes.

https://3v4l.org/lP7cb
https://www.php.net/manual/en/exception.getcode.php

greg0ire: Indeed. WTF…

greg0ire: See fix here: https://github.com/vimeo/psalm/pull/3175

PR #3175, opened 2020-04-18
Account for PDOException

greg0ire:
Probably for legacy reasons, PDOException::getCode() can return string (it's an
SQLSTATE code). Note that if you instantiate it yourself, you will get
an integer though.

I changed the signature of the Exception stub accordingly, and then
there were errors about incompatible signatures, so I had to change the
signature of Throwable too, which makes sense since a Throwable could be
a PDOException. The signature at
https://www.php.net/manual/en/throwable.getcode.php seems plain wrong,
and that same page of the docs warns about PDOException. Not sure what
are the plans for future versions of PHP.

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreImmutableClasses.phpstub
+++ b/src/Psalm/Internal/Stubs/CoreImmutableClasses.phpstub
@@ -25,8 +25,10 @@ interface Throwable
 
     /**
      * @psalm-mutation-free
+     *
+     * @return int|string https://www.php.net/manual/en/throwable.getcode.php
      */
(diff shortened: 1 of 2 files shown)


greg0ire:
>  and then there were errors about incompatible signatures, so I had to change the signature of Throwable too

I rebased since then, and it no longer happens, what should I do? I think it makes sense since [`PDOException` implements `throwable`](https://3v4l.org/Bfm9C)

muglug:
Yeah, this is correct – getCode doesn't have to return an `int`, sadly

Issue #2861, opened 2020-02-21
Invalid MixedReturnStatement ( unable to replicate )


azjezz:
the code runs with no issues in psalm.dev : https://psalm.dev/r/2fafbd8ebf

however, the same is triggering a `MixedReturnStatement` locally and in CI.

Repostiroy : https://github.com/azjezz/psl
Pull Request : https://github.com/azjezz/psl/pull/15
CI Build : https://travis-ci.org/azjezz/psl/jobs/653660377#L2205

weirdan:
@muglug 

I think there's a possible enhancement that could also help with this issue. Currently, Psalm only analyzes autoloaded functions in a limited way because it accesses them through reflection. Basically it ignores function docblocks (for autoloaded functions) altogether.

What it could do instead is either use the docblock information available through reflection, or figure out source files for those functions and add them to the list of files to be scanned. The latter may cause some performance penalty, especially when functions are included but not used. 

I tested the latter approach on `azjezz/psl` repo and it makes about a hundred issues to disappear (many are Mixed* false positives).

azjezz:
@weirdan thanks for looking into this, currently this issue is making it really hard for me to continue working on the project as i can't really figure where the real issues are due to this.

weirdan:
@azjezz you may try to add all those files containing functions as [stubs](https://psalm.dev/docs/running_psalm/configuration/#ltstubsgt), as a temporary workaround. As far as I can tell it should let Psalm know it should scan them.

azjezz:
fixed in #3183

Thank you @weirdan :) I'm currently removing all function stubs from psalm configurations in azjezz/psl#32 and hopefully, with the help of #3768, this issue won't be encountered again 🎉

PR #3183, opened 2020-04-19
Scan files included by autoloaders

weirdan:
Refs vimeo/psalm#2861

To provide a bit more context: this PR changes the way Psalm discovers what's been made available during autoloader registration. Previously it would simply deep-scan files listed in `autoloader.file` section of `composer.json`. Now it takes a snapshot of loaded files before autoloader registration and compares that to the included file list after autoloader is registered. This provides almost bullet-proof support for any custom logic autoloader may implement.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,6 @@
         "amphp/amp": "^2.4.2",
         "bamarni/composer-bin-plugin": "^1.2",
         "brianium/paratest": "^4.0.0",
-        "php-coveralls/php-coveralls": "^2.2",
         "phpmyadmin/sql-parser": "5.1.0",
         "phpspec/prophecy": ">=1.9.0",
         "phpunit/phpunit": "^7.5.16 || ^8.5 || ^9.0",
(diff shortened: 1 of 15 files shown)


weirdan:
I think better approach could be to compare the list of included files before and after autoloader registration, remove any that are related to Psalm itself and scan the remaining.

azjezz:
any updates here?

weirdan:
CircleCI reports one issue remaining with ProxyManager, but I can't seem to reproduce it outside the CI, even using CircleCI-built psalm.phar.

muglug:
Thanks, I'll debug that error myself

azjezz:
🎉 Thank you for you work!

muglug:
@weirdan the error was caused by the phar, which is built with `ocramius/packageversions:1.4.2`, which is missing a `@psalm-pure` annotation. I think Psalm is reflecting that class's method docblock from the bundled phar _after_ reflecting the one in the vendor directory.

weirdan:
I wonder why it wouldn't fail the same way when I tried running the phar I downloaded from CircleCI. Anyway, we'll see how it goes when new release is cut.

PR #3203, opened 2020-04-20
Make Throwable methods overridable

### LINKED ISSUE #3202: final Method Throwable::getMessage is declared final and cannot be overridden
See https://psalm.dev/r/2f95b7b3bf

I think the methods should be marked as final in `Exception` but not in `Throwable`. Will send a PR shortly.

greg0ire:
Methods from \Exception cannot be overriden, but that does not mean
methods from \Throwable cannot be.

Closes #3202


muglug:
Thanks!

PR #3205, opened 2020-04-20
Fix #3204 normalize namespaced function ids

### LINKED ISSUE #3204: Psalm with @template in namespace function fails to find error
The original problem from #3182 is still not resolved.

When no `namespace` is used, Psalm correctly finds mismatched `@template` types:

https://psalm.dev/r/6d185cca5b

When a `namespace` _is_ used, Psalm **fails to find** mismatched `@template` types:

https://psalm.dev/r/83cc97962c

Note that this seems to be only when using `string` callable types. From what I can tell all other callable scenarios work as expected:

https://psalm.dev/r/e2cd506b36

beaucollins:
Functions in storaged are indexed with lowercase names and preceding backslash removed.

Fixes #3204

### DIFF
--- a/src/Psalm/Internal/Analyzer/TypeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/TypeAnalyzer.php
@@ -51,7 +51,9 @@ use Psalm\Type\Atomic\TTraitString;
 use Psalm\Type\Atomic\TTrue;
 use function get_class;
 use function array_merge;
+use function strpos;
 use function strtolower;
+use function substr;
 use function in_array;
 use function array_values;
 use function count;
@@ -1869,7 +1871,7 @@ class TypeAnalyzer
             try {
                 $function_storage = $codebase->functions->getStorage(
                     $statements_analyzer,
-                    $input_type_part->value
+                    self::getNormalizedCallableFunctionId($input_type_part)
                 );
 
                 return new TCallable(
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/TypeAnalyzer.php
```diff
@@ -2796,4 +2798,14 @@ public static function simplifyUnionType(Codebase $codebase, Type\Union $union)
 
         return $unique_type;
     }
+
+    /**
+     * @param TLiteralString $atomic_type
+     * @return string
```
weirdan, line 2804: You can use actual type declarations. Psalm doesn't run on anything below PHP 7.1

beaucollins:
Fixes #3204

weirdan:
> Fixes #3204

If you add this text to the PR itself, merging it will automatically close the referenced issue.

beaucollins:
> If you add this text to the PR itself, merging it will automatically close the referenced issue.

I added it as a comment in case the same reference in the PR title didn't work. I was trying to make it match the others `master`.

I'm assuming we'll want to squash this into one commit if it were to be accepted.

weirdan:
I mean, put it here:
image

And it will display (and close, when merged) linked issues here:
image

beaucollins:
> I mean, put it here:

Gotcha. I could have sworn at one point GitHub did the same thing with PR comments as well.

Thanks!

beaucollins:
Looks like my test cases are fixed with 8f57d0c73

Issue #2084, opened 2019-08-29
Better support for compact


bugreportuser:
Psalm [gives an error](https://github.com/vimeo/psalm/commit/5770a6c904c33be736d952c31a6d596b8e1c3299) when `compact()` is used on an undefined variable. However, it still infers the type as `array<array-key, mixed>`.

Example (https://psalm.dev/r/bfab55228d):
```php
<?php
$x = 'str';
$y = 0;
$example = compact('x', 'y');
echo ord($example['x']);
echo chr($example['y']);
```

Example output:
```
Psalm output (using commit 1f0aca0):

INFO: MixedArgument - 5:10 - Argument 1 of ord cannot be mixed, expecting string

INFO: MixedArgument - 6:10 - Argument 1 of chr cannot be mixed, expecting int
```

It should be able to treat that example as:
```php
<?php
$x = 'str';
$y = 0;
$example = ['x' => $x, 'y' => $y];
echo ord($example['x']);
echo chr($example['y']);
```

muglug:
Do people use `compact`? I'd rather not support this IMO bad function.

bugreportuser:
It's debatable. I think `extract` is actually the problem, but they're usually grouped together.

`compact` is basically syntactic sugar. A parser can transform `compact('x', 'y')` into `['x' => $x, 'y' => $y]` as easily as `$x ?? 10` into `isset($x) ? $x : 10`. If you pass variable strings to `compact`, that's when it becomes a bad function.

muglug:
yeah, but it's not particularly sugary because it comes in the form of a function that takes a bunch of strings.

bugreportuser:
That's pretty common in PHP:

```php
constant('constant');
array_map('function', []);
get_parent_class('class');
```

muglug:
🙈

still-dreaming-1:
Psalm would have slightly better support for compact if it at least understood the returned array key type was a string instead of array-key, although that would not exactly fix the issues of the OP.

still-dreaming-1:
Actually I was interested in making that change myself (improving key type), but I couldn't find anywhere in Psalm's code where return type of compact was specified. It seemed likely to me this would be hard coded somewhere, but I couldn't find it. I did look through the contributions guide, but I still couldn't figure it out.. There is another similar contribution I would like to attempt as well.

bugreportuser:
[...]

PR #3209, opened 2020-04-21
Made compact's return array key type more specific

still-dreaming-1:
PHP's `compact` function always returns an array with string keys, so I changed it to be typed as a `string` key instead of the implicit `array-key`. While this improvement is somewhat similar to what was requested at https://github.com/vimeo/psalm/issues/2084, it does not actually address the issue reported.

This is my first attempt at contributing to Psalm, so feedback is very welcome. I didn't add any tests as I wasn't sure I could add any meaningful ones, or perhaps I am just not familiar enough with tests accompanying changes like this. I don't mind adding something if someone can suggest an idea.

Technically, only the change to CallMap.php was necessary, and the tests were still passing after that change. But to help me understand the implications of changing that return type, I tried changing it to return an `int` instead of `array`, which is completely wrong, but that was the point. When it was incorrectly set to `int`, some of the tests in FunctionCallTest.php started failing. Even though those same tests were not failing after making the correct change to CallMap.php, I still went ahead and added more specific types where `compact` was used in those tests. Please let me know if those changes make sense, as I don't completely understand those tests.


still-dreaming-1:
Looks like some kind of file upload timed out in the AppVeyor build...

muglug:
thanks!

PR #3211, opened 2020-04-21
Fix #3045 - ignore mixed assignments to variables named $_

### LINKED ISSUE #3045: MixedAssignment - Unable to determine the type that $_ is being assigned to
Hi,

Regarding the following code ( https://psalm.dev/r/641944c90f ):
```php
<?php

$gen = (function (): Generator {
    yield 1 => 'a';
    yield 2 => 'b';
})();

foreach ($gen as $k => $_) {
    echo "$k\n";
}
```
An error is detected:

> INFO: MixedAssignment - 8:24 - Unable to determine the type that $_ is being assigned to

I expected that by naming the variable with just an underscore I would not have an error.

villfa:
Fix #3045


muglug:
Awesome, thanks!

PR #3221, opened 2020-04-24
Allow non-string keys

greg0ire:
Although it would be stupid to provide an array with exclusively
non-string keys, it's possible to have array with a bit of both.

See for instance
https://github.com/doctrine/dbal/blob/155d028be084ef69110d09938c57c351a2126e5d/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php#L263-L276

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
+++ b/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
@@ -249,11 +249,12 @@ function uksort(array &$arr, callable $callback): bool
 }
 
 /**
+ * @psalm-template K of array-key
  * @psalm-template T
  *
- * @param array<string, T> $arr
+ * @param array<K,T> $arr
  *
- * @return array<string, T>
+ * @return array<K,T>
  * @psalm-pure
  */
 function array_change_key_case(array $arr, int $case = CASE_LOWER)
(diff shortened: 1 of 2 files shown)


greg0ire:
@muglug I would like to write a test, but I don't understand how they are organized. Can you recommend an existing file, or should I create a new one?

src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
```diff
@@ -249,11 +249,12 @@ function uksort(array &$arr, callable $callback): bool
 }
 
 /**
+ * @psalm-template K of array-key
```
greg0ire, line 252: :thinking: maybe I should remove the ` of array-key` part?

weirdan: You may do that, however it really shouldn't do any harm. After all, array keys could only be `array-key`. Come to think of it, it may even be useful to catch situations when Psalm wrongly infers key types, like here: https://psalm.dev/r/5c1dae6bd1

greg0ire:
> Can you recommend an existing file

Answering my own question https://github.com/vimeo/psalm/blob/master/tests/ArrayFunctionCallTest.php looks like a good candidate

greg0ire:
The same with a test, please review :)

muglug:
Thanks!

PR #3233, opened 2020-04-26
Fix #2845 - FunctionReturnTypeProvider for sscanf

### LINKED ISSUE #2845: sscanf reports error when it shouldn't
For code like this https://psalm.dev/r/0f20fd0654

psalm reports mixed type error even though documentation says returned value will be an array [if called with 2 params](https://www.php.net/manual/en/function.sscanf.php).

villfa:


muglug:
Thanks!

PR #3235, opened 2020-04-26
Update CONTRIBUTING.md

mkamranpk:
I could not find the 'dev-master' branch, assumed it was renamed to 'master'.


weirdan:
`dev-` is the prefix Composer adds to any branch, so when talking about versions it's proper to use `dev-master`.

PR #3242, opened 2020-04-27
chore: remove backticks because of escaping issue in docs

mr-feek:
👋 I'm sure this is probably an issue with the underlying tool used to generating these docs, but I honestly can't find where that is done in the codebase.

I haven't verified since I'm not  sure how to build the docs, but I'm guessing this will resolve the escaping issue.

## https://psalm.dev/docs/running_psalm/configuration/
image

### DIFF
--- a/docs/running_psalm/configuration.md
+++ b/docs/running_psalm/configuration.md
@@ -339,7 +339,7 @@ Allows you to hard-code a serializer for Psalm to use when caching data. By defa
 
 ## Project settings
 
-#### `<projectFiles>`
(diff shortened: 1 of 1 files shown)


weirdan:
Doc are generated in psalm.dev repo: https://github.com/psalm/psalm.dev

muglug:
Turns out this is necessary for the rendering the tags in the main headline text.

mr-feek:
Sorry, I'll try checking out the doc repo when I have some time and see where the double escaping is happening

PR #3244, opened 2020-04-27
feat: add simple helper for making a union type

mr-feek:
👋 any interest in this? I know it's incredibly simple but hey, it isn't too much simpler than `Type::getNumeric` 😂 

While developing plugins, I _constantly_ create instances of `PhpParser\NodeAbstract\UnionType` by accident instead of `Psalm\Type\Union`, and other times I forget to pass an array of types and instead pass a single type

### DIFF
--- a/src/Psalm/Type.php
+++ b/src/Psalm/Type.php
@@ -1980,4 +1980,9 @@ abstract class Type
     {
         self::$memoized_tokens = [];
     }
+
+    public static function union(Atomic ...$types): Union
+    {
+        return new Union($types);
+    }
 }
--- a/tests/TypeTest.php
+++ b/tests/TypeTest.php
@@ -1,0 +1,33 @@
+<?php
+namespace Psalm\Tests;
+
+use Psalm\Type\Atomic\TInt;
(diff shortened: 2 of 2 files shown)


mr-feek:
Is it necessary for me to explicitly override the type of the variadic function via docblocks? It seems to me that might be a bug..

```
ERROR: ArgumentTypeCoercion - src/Psalm/Type.php:1986:26 - Argument 1 of Psalm\Type\Union::__construct expects non-empty-array<int, Psalm\Type\Atomic>, parent type list<Psalm\Type\Atomic> provided (see https://psalm.dev/193)
        return new Union($types);
```

muglug:
> It seems to me that might be a false positive

Not a false-positive – `$types` could be empty.

But more generally, I don't see why

```
Type::union($a1, $a2)
```
is really preferable to
```
new Type\Union([$a1, $a2])
```

The latter is 5 extra characters, but it's more explicit about exactly what's happening (and I doubt that constructor will ever change).

andrew-demb:
You can add one non-variadic argument, and one variadic to function

PR #3259, opened 2020-04-28
Fix variable names from plugin hook

orklah:
Sorry, I realized too late that I forgot to rename some variables from last PR.

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
@@ -766,13 +766,13 @@ abstract class FunctionLikeAnalyzer extends SourceAnalyzer
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
```diff
@@ -766,13 +766,13 @@ function (FunctionLikeParameter $p) {
             }
         }
 
-        $plugin_classes = $codebase->config->after_functionlike_checks;
+        $plugin_functions = $codebase->config->after_functionlike_checks;
```
weirdan, line 769: The original name actually matched the variable contents - it's a list of classes.

orklah: Oh, you're right. I thought for a moment it designated the role of the variable's content. Thanks, I close this PR then

PR #3267, opened 2020-04-29
Fix #3265

### LINKED ISSUE #3265: Failed linter
https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Stubs/ext-ds.php#L567

andrew-demb:
#3265 
We should add function body


muglug:
Thanks!

PR #3269, opened 2020-04-29
Fix generator return type collect

### LINKED ISSUE #3268: Return type inference for generators
https://psalm.dev/r/f6160f23ad

Follow-up for https://github.com/vimeo/psalm/issues/3257#issuecomment-621002517.

kelunik:
Fixes #3268.

I'm not sure whether `$yield_types = [];` is correct, but without it a test fails.


kelunik:
This doesn't solve the issue I'm seeing in Amp. :-(

Issue #3279, opened 2020-05-01
[Bug] Generic mixins do not work when type is self or static


mr-feek:
👋From my testing, generic mixins don't work when the type passed is `self` or `static`. Consider the following: https://psalm.dev/r/a0e56051aa If the type is hardcoded to be `Dog`, the test passes

I think the resolving of the type needs to be done before actually mixing in the methods.

Related to #3237

PR #3280, opened 2020-05-01
fix: mixin parameter of self should be properly resolved

mr-feek:
👋 took a stab at fixing #3279 

I was able to get a test case passing for `self`. 

I can't get the following test passing for references to `static`, but I think the late static binding is outside of my grasp. Perhaps someone else can finish that.

```
'staticTemplatedMixin' => [
                '<?php

                    /**
                     * @template T
                     */
                    abstract class Foo {
                        /** @return T */
                        abstract public function hi();
                    }

                    /**
                     * @mixin Foo<static>
                     */
                    class Bar {}

                    class Three extends Bar { }

                    $three = new Three();
                    $three = $three->hi();',
                [
                    '$three' => 'Three',
                ]
            ],
```

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
+++ b/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
@@ -1264,7 +1264,9 @@ class ReflectorVisitor extends PhpParser\NodeVisitorAbstract implements PhpParse
                         Type::fixUpLocalType(
                             $docblock_info->mixin,
                             $this->aliases,
-                            $this->class_template_types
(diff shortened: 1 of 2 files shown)


mr-feek:
The issue around `static` actually seems to be a broader issue that mixins aren't inherited

muglug:
Sorry, I didn't notice this – mind rebasing off master?

This solution is better than the one I used – you could include a revert of https://github.com/vimeo/psalm/commit/5e764673789f8033c74c6c7b67d11c5552ee70eb in your changes

PR #3304, opened 2020-05-05
Fix link to PHPDoc types documentation

XedinUnknown:
This fixes link to PHPDoc's type syntax documentation. It appears that they have added an extra URL segment to represent the doc version.


muglug:
Thanks!

PR #3333, opened 2020-05-10
update proc_open() types for PHP 7.4

sj-i:
as of PHP 7.4, proc_open() accepts an array for its first argument.
https://www.php.net/manual/en/function.proc-open.php#refsect1-function.proc-open-parameters


muglug:
Thanks!

PR #3334, opened 2020-05-10
fix: code location is nullable

mr-feek:
Added `MethodVisibilityProviderInterface` to a class and immediately got
```
Uncaught TypeError: Argument 5 passed to Psalm\LaravelPlugin\ReturnTypeProvider\EloquentBuilderReturnTypeProvider::isMethodVisible() must be an instance of Psalm\CodeLocation, null given
```

https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Provider/MethodVisibilityProvider.php#L81 has this marked as nullable


muglug:
Thanks!

Issue #3340, opened 2020-05-11
Internal Psalm error: Intersection types


M1ke:
*Version:* Psalm 3.11.2@d470903722cfcbc1cd04744c5491d3e6d13ec3d9

Please see the [demo project here](https://github.com/m1ke/psalm-ftp-bug) for a working demo of this issue.

Alternatively the contents of [this file](https://github.com/dg/ftp-php/blob/master/src/Ftp.php) can be pasted into the psalm.dev home page to show the error.

In short: trying to analyse a project containing the dependency `dg/ftp` throws the following internal error, even in the case where `vendor/` is ignored in `psalm.xml`.

```
Uncaught Psalm\Exception\TypeParseTreeException: Intersection types must be all objects or all object-like arrays, Psalm\Type\Atomic\TString provided in ~/psalm-ftp-bug/vendor/vimeo/psalm/src/Psalm/Type.php:586
Stack trace:
#0 ~/psalm-ftp-bug/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/CommentAnalyzer.php(1041): Psalm\Type::getTypeFromTree(Object(Psalm\Internal\Type\ParseTree\IntersectionTree), Object(Psalm\Codebase))
#1 ~/psalm-ftp-bug/vendor/vimeo/psalm/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php(1105): Psalm\Internal\Analyzer\CommentAnalyzer::extractClassLikeDocblockInfo(Object(PhpParser\Node\Stmt\Class_), Object(PhpParser\Comment\Doc), Object(Psalm\Aliases))
#2 ~/psalm-ftp-bug/vendor/vimeo/psalm/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php(272): Psalm\Internal\PhpVisitor\ReflectorVisitor->registerClassLike(Object(PhpParser\Node\Stmt\Class_))
#3 ~/psalm-ftp-bug/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(200): Psalm\Internal\PhpVisitor\ReflectorVisitor->enterNode(Object(PhpParser\Node\Stmt\Class_))
#4 ~/psalm-ftp-bug/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(91): PhpParser\NodeTraverser->traverseArray(Array)
[...]

PR #3342, opened 2020-05-11
Try/catch block when getting docblock method parameter types

M1ke:
Resolves https://github.com/vimeo/psalm/issues/3340

Add try/catch block to provide a better message when types fail to parse, specifically added because of an issue where `& $var` was treated as a union type but was actually meant to be a pass-by-reference variable in the docblock

### DIFF
--- a/src/Psalm/Internal/Analyzer/CommentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/CommentAnalyzer.php
@@ -1041,7 +1041,10 @@ class CommentAnalyzer
                         try {
                             $param_type = Type::getTypeFromTree($method_tree_child->children[0], $codebase);
                         } catch (TypeParseTreeException $e) {
-                            throw new DocblockParseException('Badly-formatted @method parameters for ' . $method_entry.'; have you checked if a variable passed by reference has a space between the ampersand and the start of the variable?');
+                            $msg = 'Badly-formatted @method parameters for ' . $method_entry;
+                            $msg .= '; have you checked if a variable passed by reference has a space between';
+                            $msg .= ' the ampersand and the start of the variable?';
+                            throw new DocblockParseException($msg);
                         }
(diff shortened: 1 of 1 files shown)


M1ke:
Have resolved the phpcs check

M1ke:
Seems travis failed just the PHP 7.3 build, because of the same phpcs issue that Circle found. Looks to be the only fail so will push up.

M1ke:
The AppVeyor build is failing but for a file different to what I have changed; it is a failure of `php ./psalm --shepherd` and is saying some values can be null. Running locally this passes; also the file highlighted isn't one I've changed, and as far as I can see my changes have no type implications as they throw an exception that's already thrown in the same function.

So I am unsure what I need to do to resolve.

src/Psalm/Internal/Analyzer/CommentAnalyzer.php
```diff
@@ -1038,7 +1038,14 @@ public static function extractClassLikeDocblockInfo(
 
 
                     if ($method_tree_child->children) {
-                        $param_type = Type::getTypeFromTree($method_tree_child->children[0], $codebase);
+                        try {
+                            $param_type = Type::getTypeFromTree($method_tree_child->children[0], $codebase);
+                        } catch (TypeParseTreeException $e) {
+                            $msg = 'Badly-formatted @method parameters for ' . $method_entry;
+                            $msg .= '; have you checked if a variable passed by reference has a space between';
+                            $msg .= ' the ampersand and the start of the variable?';
```
weirdan, line 1046: I would prefer the message to be more actionable. Like 'There should be no space between & and the variable name' or something similar.

M1ke:
Happy to do that, though I might hold off until I have some insight on
those tests rather than just spend all that time running them for them to
fail again.

muglug:
I fixed in d8cd30b

M1ke:
Oh, ok then

PR #3355, opened 2020-05-12
send php native warnings/errors/notices on stderr instead of stdout

staabm:
- sending errors on stderr is best practice..  on stdout only "regular data" should be sent
- don't break unix pipe-operations which error on unexpected output like errors/warnings


muglug:
Thanks!

Issue #3347, opened 2020-05-11
Introduce a repo just for stubs?


muglug:
I'm wondering whether it would be worthwhile to have a repository just for machine-readable class and function stubs for common frameworks. A user would do something like `composer require --dev psalm/framework-stubs` and Psalm would be able use those in its analysis.

Any framework function/method whose return type can be expressed using templates or a conditional return type would be moved to that repo. All other custom functionality would remain with the plugins.

Pro:
- It would give users limited framework support instantly, without having to navigate the (still pretty easy) plugin installation process
- it would reduce the overhead for adding Psalm stubs for a framework not yet covered

Con:
- It would make maintenance for Psalm plugins trickier – if you wanted to add a stub you'd have to PR it in psalm/framework-stubs, then require the stub version bump
- it would require some sort of versioning system in `psalm/framework-stubs` so that Psalm didn't assume a given framework function exists unless some version constraint was met

cc @mr-feek @seferov @weirdan

muglug:
The obvious parallel in the Typescript world is https://github.com/DefinitelyTyped/DefinitelyTyped

joehoyle:
This sounds quite similar to the TypeScript approach to the `@types` namespace, which requires a PR to https://github.com/DefinitelyTyped/DefinitelyTyped, I think that has worked very well in TS, and I think it would be a great addition to Psalm to encourage more community contribution for stubs. Authoring plugins is a much larger undertaking!

mr-feek:
I'm not necessarily on the side of maintaining another repo just for stubs as I'd need to give it more thought, but I'd like to mention another pro!

For the [psalm laravel plugin](https://github.com/psalm/psalm-plugin-laravel), we will have to make many of the same stubs as the [larastan](https://github.com/nunomaduro/larastan) phpstan plugin. It doesn't make much sense to be maintaining them in parallel, it would be great to have a single source of truth for stubs that all static analyzers could utilize.

erunion:
I really like the idea of a single repo for stubs!

muglug:
>  it would be great to have a single source of truth for stubs that all static analyzers could utilize.

Yeah, as Psalm and PHPStan support each others type annotations (or at least a decent subset thereof) they could share in this goodness.

muglug:
for versioning I think there'd be some sort of structure like

```
- laravel/laravel/latest/model.phpstub
- laravel/laravel/6.x/model.phpstub
- laravel/laravel/shared/relations.phpstub
```

[...]

PR #3356, opened 2020-05-12
Streamline plugins docs

weirdan:
Following the discussion in #3347 here's my attempt to streamline the documentation on using and authoring plugins.


muglug:
Any reason this is still a draft?

weirdan:
I wanted to check it with psalm.dev, but couldn't get mkdocs working.

PR #3362, opened 2020-05-14
Simplify PSR4 namespace config.

dereuromark:
This seems to be some relict from refactoring back in https://github.com/vimeo/psalm/commit/a338e76ef68bce75a0cc9c5616b6413e9d92bfcd
Those additional lines do not seem needed or useful anymore.
Also consolidated paths to the best practice for psr-4 config.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,8 +68,7 @@
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

dereuromark:
The fails are normal/unrelated?

Issue #3360, opened 2020-05-14
Use JSON_PRETTY_PRINT by default


dereuromark:
When generating a report.json using `--report=PATH` the file generated is not using JSON_PRETTY_PRINT, being one huge blob that is not readable by default without IDE level extensions or alike.

Since this is a dev only tool, I propose that the default here should be with pretty print, to allow this also be read/searched by humans more easily.
After all, the few more bytes here on the disk probably don't matter. And tooling doesn't care either way.
It is also harder to read diff when committing those files or alike.

muglug:
Mind creating a PR?

PR #3364, opened 2020-05-14
Use pretty print for JSON reporting.

dereuromark:
Resolves https://github.com/vimeo/psalm/issues/3360

Any preference on the implementation? Or is it OK as such?
I only modified the reporting (not internal JSON encoding for hashing etc), and I put the constants into a dedicated class to keep it DRY and easy to maintain.
While doing that I also captured the "theoretical" `false` return to make the return result more "true" towards documented and expected type `string`.

The constants are used as such for a lot of libraries and frameworks as such, e.g.
psalm/vendor/phpunit/phpunit/src/Util/Json.php

### DIFF
--- a/src/Psalm/Internal/Json/Json.php
+++ b/src/Psalm/Internal/Json/Json.php
@@ -9,14 +9,19 @@ use const JSON_UNESCAPED_SLASHES;
 use const JSON_UNESCAPED_UNICODE;
 
 /**
- * Provides pretty printed JSON output.
+ * Provides ability of pretty printed JSON output.
  */
 class Json
 {
     /**
      * @var int
      */
-    protected const OPTIONS = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE;
+    public const PRETTY = JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE;
+
+    /**
+     * @var int
+     */
+    public const DEFAULT = 0;
 
     /**
      * @param mixed $data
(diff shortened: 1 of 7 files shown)


muglug:
Actually is there a reason you can't have a separate `json-pretty` output format? It's possible that some won't expect the json-encoded output to have newlines (as `json_encode` by default converts any in strings to the `"\n"` character), so seeing a pretty-printed output might break stuff.

dereuromark:
Yes, there could also be some flag/option for this on top, of course.

dereuromark:
The psalm errors seem to be a deficiency of psalm itself, right?

muglug:
> The psalm errors seem to be a deficiency of psalm itself, right?

Sort of. There's no _great_ way to resolve constants referenced with `static::`, as they could be overridden downstream.

One way is to take the docblock type (and also verify a constant's docblock type against the calculated one) but that tends to be a pretty edge case. I think there's an issue somewhere.

dereuromark:
Code wise static:: is always more correct than self:: as it allows the extension (which in library code should always be the case). No need to protect developers from themselves beyond basic due diligence. Yes, it can and should be able to be overwritten on project level, as long as one sets the same type of course.
But that's a topic for a different discussion :) So fair enough then.

muglug:
> Code wise static:: is always more correct than self::

I very much disagree there – PHP doesn't allow you to set constants `final`, but that seems to be a design flaw more than anything else. For example, I never want `self::DAYS_IN_WEEK` to change.

muglug:
Thanks!

dereuromark:
Yes, those are "final" constants, here you could use self indeed.
But other cases could be very well extendable, thus psalm shouldn't error in those IMO.

Issue #3345, opened 2020-05-11
InvalidCast - object cannot be cast to string


bkdotcom:
despite having a check before doing so
ie either `methoc_exists($obj, '__toString')` or `isCallable(array($obj, '__toString'))`

https://psalm.dev/r/a1cafa9d3e

```php
if (is_object($val)) {
    if (!\is_callable(array($val, '__toString'))) {
        throw new \RuntimeException(__METHOD__ . ': ' . \get_class($val) . 'is not stringable');
}
    echo (string) $val;   // yes, this can be cast to string
}
```

weirdan:
`method_exists()` seems to be working: https://psalm.dev/r/c0bca5d469

muglug:
OK, what's required here is a simple check in `AssertionFinder` that generates the same assertion for `is_callable([$a, 'someMethod'])` as it would for `method_exists($a, 'someMethod')`

PR #3372, opened 2020-05-16
Fix InvalidCast after is_callable([$val, '__toString']) check

nobuf:
Attempt to fix https://github.com/vimeo/psalm/issues/3345. First pull request here :)


muglug:
Thanks!

PR #3374, opened 2020-05-16
Avoid regenerating docblock when no modification is made

orklah:
This should prevent Psalter to regenerate docblock when no change were made. This is purely cosmetic, but it reduce drastically the review on our project (added lines, reformatted comments...)


orklah:
There are some side effects, I'll look into that

orklah:
This is better. I think the remaining errors are unrelated

muglug:
Thanks!

Issue #3329, opened 2020-05-09
Psalm doesn't notice possibly incorrect cloning on mixed/generic type


EvgeniiR:
Here's code sample - https://psalm.dev/r/069e705033

Variable $a can be not cloneable(scalar)

UPD: same with mixed - https://psalm.dev/r/9191701ffd

EvgeniiR:
@muglug Hello, I have one about how it should work. 
"clone" operator is valid in PHP, if subject __clone method is not private, so when we got an instance of concrete class, we can check for existence and visibility of this method, but what shoult we do when we got an `TObject` or an interface type, where __clone method is not specified(mean we can't make sure it's not private)?

Here is an example situation - https://psalm.dev/r/29aa969830
Should we raise PossiblyInvalidClone issue in that example?

it would be more strict to raise issue here, as in the case of InvalidCast - https://psalm.dev/r/1f7556ac06 , and force to add __clone method to interface, but will this be acceptable given broken backward compatibility, and small chance of encountering the redefined _ _ clone method on the object?

PR #3382, opened 2020-05-17
Correct analyze clone expression

EvgeniiR:
Fixes https://github.com/vimeo/psalm/issues/3329

There are still some edge-cases, primarily about the return type of expression with possible invalid clone, e.g. https://psalm.dev/r/1a6bdbac1b . My solution also outputs MixedAssignment in that code sample, but ideally it should infer type `A` from `clone $a` expression, because only `A` in union type `A|string $a` is cloneable(strings are not).

This is not so difficult to do for simple types, but it is a little difficult to do this for cases like generic types - https://psalm.dev/r/27f283d113 (note that current psalm version does not show warnings here because it doesn't treat cloning string as error, and not because it  could infer type `T:fn-foo as A` for $a)

Also 1 more actual question - https://github.com/vimeo/psalm/issues/3329#issuecomment-629769684 . Strict check here cause issues on testing with real projects on circleci.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -301,6 +301,7 @@
             <xs:element name="PossiblyInvalidArrayAssignment" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="PossiblyInvalidArrayOffset" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="PossiblyInvalidCast" type="IssueHandlerType" minOccurs="0" />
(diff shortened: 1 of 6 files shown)


muglug:
Oh, there'll also need to be `MixedClone` like `MixedPropertyFetch`

mr-feek:
Nice feature!!  👏

muglug:
Thanks – I'll add `MixedClone` myself (I rejiggered `ExpressionAnalyzer` a little to make it easier to add updates in the future.

Issue #3376, opened 2020-05-16
Inconsistent type errors on arrow functions


jwdunne:
New to using Psalm and it's awesome!

Noticed that return type checking doesn't seem to handle arrow functions as well as the non-arrow counterparts e.g:

https://psalm.dev/r/2b701d86b9

Because I'm new, I'm not sure if I'm doing something wrong but it doesn't look like the return type is considered when returning an arrow function.

muglug:
Oh @ragboyjr I think your earlier solution just needs a slight adaptation to arrow functions - want to make it, or should I?

ragboyjr:
@muglug ah, ok, will give this a look!

ragboyjr:
OK, i was able to fix the patch for arrow functions, but it looks like there's another bug that tries to infer his example as `Closure(empty):bool`. Will try to address that in a separate PR.

muglug:
Thanks!

PR #3385, opened 2020-05-17
Infer Return Types for Arrow Functions #3376

ragboyjr:
- Made a small patch to check for closure or arrow
  function when attempting to infer the functions
  params
- Added new isExprClosureLike to start to consolidate
  all checks on closure/arrow fns

Signed-off-by: RJ Garcia <ragboyjr@icloud.com>


ragboyjr:
@muglug I'm guessing some reference to psalm-assert on isExprClosureLike would fix psalm's issues in the check?

muglug:
Thanks! I simplified the checks

PR #3387, opened 2020-05-18
Solve more cases when docblock rendering is not needed

orklah:
https://github.com/vimeo/psalm/pull/3374 Didn't solve all the cases because when psalm override an attribute, it was still detected as a change.

This change check those cases as well by only flagging the docblock as modified when there is an actual modification of the content of an attribute


muglug:
Thanks!

PR #3395, opened 2020-05-18
stubs for array_sum

orklah:
This PR adds a stub for array_sum with a conditional return.
The logic is: 
- an `array<int>` will return an int
- an `array<float>` will return a float
- anything else will return int|float

During the testing, I discovered that `[4, 5.6]` will actually be considered a valid `array<float>` (instead of an `array<int|float>`) and so, will make array_sum return a float. While this is true for array_sum (and so I have a test just for this case), this is probably not strictly correct.


muglug:
Thanks, this is fantastic and I'm happy to see conditional types being used!

orklah:
> Thanks, this is fantastic and I'm happy to see conditional types being used!

I used it a lot in our legacy project, it really helped reducing issues where Psalm wasn't aware that it was working on a more narrow type than declared!

Issue #3397, opened 2020-05-19
Issue with reference argument and subclasses


afbarbaro:
This seems to be an erroneous check.   Am I missing something?  It seems the return of `bar` is correct as `B`.

https://psalm.dev/r/949e386523

```
<?php

class A { /** @var int */ public $x = 0;}

class B extends A{}

function foo(A &$obj): void { $obj->x = 1;}

function bar(): B {
	$b = new B();
	foo($b);
	return $b;
}
```

```
INFO: LessSpecificReturnStatement - 12:12 - The type 'A' is more general than the declared return type 'B' for bar

INFO: MoreSpecificReturnType - 9:17 - The declared return type 'B' for bar is more specific than the inferred return type 'A'
```

Note:  If the reference in the argument to `foo` is removed no warnings are issued.

weirdan:
This is to prevent situations like this: https://3v4l.org/fHSIl
Psalm generally performs checks at function boundaries, so at the call site (`foo($b);`) it cannot guarantee the type of `$b` won't change.

You should be able to get around this by declaring `foo()` a generic function:
https://psalm.dev/r/13b9aaa1e9
but it doesn't seem to work, which looks like a bug to me.

afbarbaro:
@weirdan yes, I can see the use of `generics` to resolve this but I bumped into that issue that now `Psalm` doesn't understand that `T` is of type that `extends A`.  So we still need a way to get pass this issue.

weirdan:
I think I fixed it in the linked PR. There are issues with master branch right now, so it can't be merged yet.

afbarbaro:
> I think I fixed it in the linked PR. There are issues with master branch right now, so it can't be merged yet.

Awesome! Thanks so much 🥇

muglug:
Above example now works

PR #3398, opened 2020-05-19
Handle templated object function parameters

weirdan:
Fixes vimeo/psalm#3397

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/PropertyAssignmentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/PropertyAssignmentAnalyzer.php
@@ -47,6 +47,7 @@ use function strtolower;
 use function explode;
 use Psalm\Internal\Taint\Sink;
 use Psalm\Internal\Taint\Source;
+use Psalm\Type\Atomic\TTemplateParam;
 
 /**
  * @internal
@@ -235,6 +236,20 @@ class PropertyAssignmentAnalyzer
                     && count($lhs_type->getAtomicTypes()) > 1
                 ) {
                     continue;
+                }
+
+                if ($lhs_type_part instanceof Type\Atomic\TTemplateParam) {
+                    $extra_types = $lhs_type_part->extra_types;
+
+                    $lhs_type_part = array_values(
+                        $lhs_type_part->as->getAtomicTypes()
+                    )[0];
+
+                    $lhs_type_part->from_docblock = true;
+
+                    if ($lhs_type_part instanceof TNamedObject) {
+                        $lhs_type_part->extra_types = $extra_types;
+                    }
                 }
 
                 if (!$lhs_type_part instanceof TObject && !$lhs_type_part instanceof TNamedObject) {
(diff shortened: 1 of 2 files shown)


weirdan:
There are tons of psalm issues currently reported in master, I'd prefer to rebase this after they are resolved.

src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/PropertyAssignmentAnalyzer.php
```diff
@@ -237,6 +238,20 @@ public static function analyzeInstance(
                     continue;
                 }
 
+                if ($lhs_type_part instanceof Type\Atomic\TTemplateParam) {
+                    $extra_types = $lhs_type_part->extra_types;
+
+                    $lhs_type_part = array_values(
+                        $lhs_type_part->as->getAtomicTypes()
+                    )[0];
```
afbarbaro, line 246: Is this guaranteed to always have at lest one element?

src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/PropertyAssignmentAnalyzer.php
```diff
@@ -789,7 +804,7 @@ public static function analyzeInstance(
                     if (IssueBuffer::accepts(
                         new InvalidPropertyAssignment(
                             $lhs_var_id . ' with non-object type \'' . $invalid_assignment_type .
-                            '\' cannot treated as an object',
+                            '\' cannot be treated as an object',
```
afbarbaro, line 807: 👍

src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/PropertyAssignmentAnalyzer.php
```diff
@@ -237,6 +238,20 @@ public static function analyzeInstance(
                     continue;
                 }
 
+                if ($lhs_type_part instanceof Type\Atomic\TTemplateParam) {
+                    $extra_types = $lhs_type_part->extra_types;
```
afbarbaro, line 242: Do we need to assign this to a variable here? It seems it's only needed if we enter into the if on lines 250-251 and it just gets re-assigned.  Maybe able to do without it here.

muglug:
Should be safe to rebase now, sorry.

I'm mid-cleanup, but I promise the next round won't be so messy.

src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/PropertyAssignmentAnalyzer.php
weirdan: Yes. `getAtomicTypes()` returns `non-empty-array`, `array_values()` turn that into `non-empty-list`, so there's always an element at offset `0`.

src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/PropertyAssignmentAnalyzer.php
weirdan: `$lhs_type_part` is overwritten on the line below. So `$extra_types` is reassigned, but to the property of a different object.

muglug:
I fixed this _slightly_ differently in f335560

PR #3400, opened 2020-05-19
Fix getSymbolInformation() for symbols containing ":"

joehoyle:
Currently a Symbol like `10-15:array{foo: string}` will return ` string}`, as the regex does not specify a string-start caret.


muglug:
OH thanks for this! I had noticed it a while back

PR #3401, opened 2020-05-19
Display variable assignment type in LSP hovers

joehoyle:
First PR, so this might be totally the wrong approach to fix this!

I wanted to support the hover action via the LSP for variable assignment statements, so I can hover the variable to see it's assigned type (e.g. `$foo = 'hello'`). When hovering `$foo` it should show me `string(hello)` as the variable type.

To do this, I had to add the node reference in the Assignment analyzer. This is because the LSP hover calls `Codebase::getReferenceAtPosition`, hence the variable assignment needs to be part of the reference map fetched via `getMapsForFile`. I'm not actually sure if the reference map is where this should live, as I don't really know what is meant by "reference" in this context. Does it refer to a reference of file positions to tokens/nodes, or is it to track statements that reference a value declared elsewhere in the codebase?

So, I'm not sure if this is the right place to add this code.

Screenshot:

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php
@@ -450,6 +450,18 @@ class AssignmentAnalyzer
                         );
                     }
 
+                    if ($codebase->store_node_types
+                        && !$context->collect_initializations
+                        && !$context->collect_mutations
+                    ) {
+                        $location = new CodeLocation($statements_analyzer, $assign_var);
+                        $codebase->analyzer->addNodeReference(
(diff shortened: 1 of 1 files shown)


muglug:
Looks good! Mind fixing the style issues (a too-long line) and also adding a test?

klesun:
Sorry for the offtopic, but @joehoyle is it vscode on the screenshot? Does this work with [Psalm (PHP Static Analysis Linting Machine)](https://marketplace.visualstudio.com/items?itemName=getpsalm.psalm-vscode-plugin) extension or is there supposed to be some other way to connect PSALM's lsp to the IDE?

joehoyle:
@klesun yes it's via the VS Code extension you mentioned.

PR #3445, opened 2020-05-24
Add false to possible parse_url return types (#3359)

### LINKED ISSUE #3359: Wrong return type for \parse_url with two arguments
[\parse_url](https://www.php.net/manual/en/function.parse-url.php) returns `false` for 'seriously malformed urls'. [This is true](https://3v4l.org/9Mepj) whether called with one param or two.

However in the latter case, psalm thinks the return type is `null|string`. It should be `null|string|false`.

https://psalm.dev/r/3234b76fb6

orklah:
Fix #3359


muglug:
Thanks!

PR #3462, opened 2020-05-26
Fix #3461 by removing file from issues after removing last issue

### LINKED ISSUE #3461: Plugin AfterAnalysisInterface receiving non-empty $issues when there are no issues
I'm using the master code. I have a project where Psalm currently is finding no issues. I have a plugin that hooks into the AfterAnalysisInterface, but for some reason the $issues are passed as not empty, even though there are no issues. I did a print_r on the $issues, and this is what I see:

```
Array
(
    [C:\hacking\hunti\src\Hunti\Button.php] => Array
        (
        )
)
```

That Button class does not have any issues, and it is not the only class in the project, so from my perspective it seems very random for this to be in the array by itself.

still-dreaming-1:


still-dreaming-1:
You can ignore the branch name being a little strange. This ended up going in a different direction than I imagined. I was originally thinking I would just change the $issues type as used in AfterAnalysisInterface. I tried changing it to `@param array<string, non-empty-list<IssueData>> $issues`, hoping that would cause Psalm to reveal where that is not honored in the code. But there is a Psalm suppression preventing it from surfacing, and it says it is because of a bug. I tried deleting that comment, but then there was other technical debt that got revealed that I didn't know how to deal with. So I abandoned that approach and just fixed the bug directly in IssueBuffer::remove() as suggested by @muglug.

muglug:
Awesome, thanks!

PR #3468, opened 2020-05-27
Fixes #3464 by typing __DIR__, __TRAIT__, and __FILE__ as non-empty-s…

### LINKED ISSUE #3464: const not considered non-empty-string if value set concatenating __DIR__
The `__DIR__` throws it off. I tried concatenating a custom const set to either an empty string or a string with characters in it, and those were both fine, but for some reason this example with concatenating __DIR__ throws it off.
https://psalm.dev/r/678b71f78c

still-dreaming-1:
…tring


weirdan:
`__TRAIT__` can be empty though, when used outside of a trait: https://3v4l.org/udW3I

muglug:
You can detect whether or not you're in a trait with

```
$statements_analyzer->getSource() instanceof TraitAnalyzer
```

muglug:
but I'm sure your solution is fine for pretty much everyone

muglug:
Thanks!

still-dreaming-1:
@muglug The change your suggesting seems easy enough to implement. Whichever way you think makes the Psalm code better I guess.

PR #3469, opened 2020-05-27
__TRAIT__ can be string or non-empty-string

still-dreaming-1:
depending on whether or not it is used from inside a trait


muglug:
Thanks! My hunch was that this would never actually break anyone's code, but you never know: https://xkcd.com/1172/

still-dreaming-1:
Your welcome. Please see the issue prompting these PRs. I guess If forgot to test these against that, as it is still not working.

PR #3471, opened 2020-05-28
fix: pass along final

### LINKED ISSUE #3470: [False Positive] Unnecessary &static required in typehint
I believe this was just a regression introduced in 3.11.5, as one of my tests over @ the laravel plugin just started failing.

https://psalm.dev/r/64c4862e7e

I don't believe the additional `&static` is necessary, since `User` is `final`.

Marking the template as covariant is a workaround

mr-feek:
closes #3470 

I can make a separate test case if you like. I figured it was fine to merge them together since the title of the test case was `inheritTemplatedMixinWithStatic`

### DIFF
--- a/tests/MixinAnnotationTest.php
+++ b/tests/MixinAnnotationTest.php
@@ -277,6 +277,62 @@ class MixinAnnotationTest extends TestCase
                         return (new FooGrandChild)->type();
                     }'
             ],
+            'inheritTemplatedMixinWithStaticAndFinalClass' => [
+                '<?php
+                    /**
+                     * @template T
+                     */
+                    class Mixin {
+                        /**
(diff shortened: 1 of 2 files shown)


tests/MixinAnnotationTest.php
```diff
@@ -243,10 +243,10 @@ public function __construct ($var) {
                         }
 
                         /**
-                         * @psalm-return T
+                         * @psalm-return self<T>
```
muglug, line 246: Any reason why you can't just create a new test?

mr-feek: figured it'd be a lot of boilerplate, happy to do so :) I'll push it up in a sec

muglug:
Thanks!

PR #3477, opened 2020-05-28
Support global functions in getSymbolInformation and getSymbolLocation

joehoyle:
Currently codebase-wide defined function are not found in `Codebase::getSymbolLocation` or `Codebase::getSymbolInformation`. This means hovers via the LSP on functions not in the current file, or "go to definition" do not work for non-locally defined functions.

It looks to me that this might have been an oversight, as methods do support this.

For stubbed functions, "go to definition" will open the stub file, which is also quite hadny.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1078,14 +1078,20 @@ class Codebase
             }
 
             if (strpos($symbol, '()')) {
-                $file_storage = $this->file_storage_provider->get($file_path);
-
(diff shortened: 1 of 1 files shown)


muglug:
I've fixed the Psalm errors, mind adding a test or two?

joehoyle:
Yup will give it a go!

muglug:
Any updates here?

joehoyle:
Sorry I didn’t add the test to this yet @muglug , I’ll add a follow up PR when I get some availability

PR #3487, opened 2020-05-29
Add checks for duplicated params and returns

### LINKED ISSUE #1400: psalm doesn't flag up duplicate param hints
* https://psalm.dev/r/dd494a658e
* using 601f303
* as per phpstan/phpstan#1931

# snippet
```php
<?php declare(strict_types = 1);

/**
* @param mixed $val
*
* @param scalar|array|object|null $val
*/
function foo($val) : void {
	echo var_export($val, true);
}
```

### LINKED ISSUE #2245: Psalm silently ignores duplicate `@return` comments
Observed: It uses the first comment and ignores the second
Expected: Combine the union types of all `@return` statements (continue to have psalm-return take precedence), or warn about multiple return comments being unsupported

```php
<?php

class X {
    /**
     * @return string in some conditions
     * @return array in some other conditions
     */
    function test() {
        // Observed: InvalidReturnType
        // Expected: No warning
        return [];
    }
}
```

orklah:
This PR should fix #1400 and fix #2245.

It checks for duplicated @param, @return (and prefixed tags too) for duplicated content.

Please tell me if there's anything to change, My goal is also to improve :)

I didn't expect that `MissingDocblockType` Issue would appear in the tests. Not sure if I did something wrong of if it's expected.

I checked PhpUnit and PHPStorm checked CS rules but I didn't launch the full CI on local. I'll fix CI if needed.

EDIT: I also fixed a case of duplicated return on a test. I think it was unrelated to the test itself

### DIFF
--- a/src/Psalm/Internal/Analyzer/CommentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/CommentAnalyzer.php
@@ -19,6 +19,8 @@ use Psalm\Internal\Type\TypeAlias;
 use Psalm\Internal\Type\TypeParser;
 use Psalm\Internal\Type\TypeTokenizer;
 use Psalm\Type;
+use function array_column;
+use function array_unique;
 use function trim;
 use function substr_count;
 use function strlen;
@@ -368,6 +370,8 @@ class CommentAnalyzer
         $comment_text = $comment->getText();
 
         $info = new FunctionDocblockComment();
+
+        self::checkDuplicatedTags($parsed_docblock);
 
         if (isset($parsed_docblock->combined_tags['return'])) {
             self::extractReturnType(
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Analyzer/CommentAnalyzer.php
```diff
@@ -1254,4 +1258,59 @@ public static function splitDocLine($return_block)
...
+     *
+     * @throws DocblockParseException  if an invalid string is found
+     *
+     * @return array<int, array<string>>
+     */
+    public static function splitAllDocLine(array $lines)
+    {
+        $splitted_lines = [];
+        foreach ($lines as $offset => $line) {
+            $splitted_lines[$offset] = self::splitDocLine($line);
```
muglug, line 1311: This the only quibble I have – `splitDocLine` does a lot we don't need – a simple `explode(' ', $line)` should be sufficient (and much faster).

```suggestion
            $splitted_lines[$offset] = \explode(' ', $line);
```

orklah:
Thank for the review!

Tests are failing for an obscure reason, If it's related to my change, I don't get how...

orklah:
Oh, test fail right now because when identation is not strictly a space, explode is not enough.

I wanted to try a regex to allow multiple spaces or even tabs, but I can find cases where it's going to fail. For example if the param is `callable(): string` there is a space in there.

Are you sure we can avoid splitDocLine?

orklah:
I still have my test failing, I'm not sure I follow the logic. I'll look closer tomorrow.

There's also other tests failing `Psalm\Exception\CodeException: MixedAssignment - somefile.php:19:17 - Unable to determine the type that $d is being assigned to`

But I have no idea why...

muglug:
Use `splitDocLine`, then, and I’ll figure out how to memoize the calculation.

orklah:
Tests pass now.

I understood why I had an error. This kind of notation: https://github.com/vimeo/psalm/pull/3487/files#diff-57230bbfd2ea55ce6c8f6ce2ee456197L28 is the same as two return types.

This means, this PR will now consider this as an error.

muglug:
Thanks!

SignpostMarv:
🥳

PR #3493, opened 2020-05-30
Fix #3437

### LINKED ISSUE #3437: Psalm doesn't know that (string) DateInterval#days is numeric-string
https://psalm.dev/r/620b62d7c3

orklah:
This PR improve the typing for DateInterval::days and fix #3437. This doesn't totally solve the issue but it's the best we can do for now


muglug:
Thanks!

PR #3494, opened 2020-05-30
Fix #3438

### LINKED ISSUE #3438: psalm should be smarter about recognizing if bcdiv can return null (if divisor is non-0)
https://psalm.dev/r/41cf16dc2d

orklah:
This PR should fix #3438.

EDIT: I hesitated to create a test but I couldn't find any Tester for bc* functions, or any Tester where this could fit. Please tell me if needed, I'll add this kind of test: https://psalm.dev/r/41cf16dc2d


muglug:
Thanks!

PR #3495, opened 2020-05-30
Improve stubs for str_replace and preg_replace

### LINKED ISSUE #3475: Allow int[] for str_replace
After upgrade from 3.11.4 to 3.11.5 I get errors for str_replace using int[] array.

Checking the docs https://www.php.net/str_replace I'm not sure we should force arrays to be string?

I'm proposing to allow using int[] in str_replace.

example: https://psalm.dev/r/c9c73429c8

orklah:
This PR expands accepted values for str_replace and preg_replace and make return conditional based on type of $subject

It fix #3475

After some tests, I found out, ints and floats are accepted and works well. No matter what, the returned type will always be a string or an array of strings. The parameter is always casted to string.

I could have added bool as an accepted value, but given that a casted boolean will be either "" or "1", it didn't make a lot of sense and this is most likely a bug in the application

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
+++ b/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
@@ -560,9 +560,9 @@ function htmlspecialchars_decode(string $string, ?int $quote_style = null) : str
 /**
  * @psalm-pure
  *
- * @param string|int|float|array<string|int|float> $search
- * @param string|int|float|array<string|int|float> $replace
- * @param string|int|float|array<string|int|float> $subject
+ * @param string|array<string|int|float> $search
+ * @param string|array<string|int|float> $replace
+ * @param string|array<string|int|float> $subject
  * @param int $count
  * @return ($subject is array ? array<string> : string)
  *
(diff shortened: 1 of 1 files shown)


orklah:
> -1 on allowing numeric types

I think flagging numeric types as issues would be a pretty opiniated rule from psalm itself. This should be the kind of issues that should be checked by a plugin IMO.

I have some of those in our legacy project, and those issues went straight to the baseline. It's not an easy or elegant issue to solve, you have to cast an entire array (possibly duplicating it) for the sake of passing the "right" type even when the function is fully compliant. If we keep this as an issue, we litteraly have to add some code to be executed, not for being sure types are right, but just on principle.

weirdan:
IMO Psalm is strict about the types even when PHP itself is not, and that's the point to use it.

orklah:
The point of Psalm is not being strict but rather being type error proof. There is no risk, no advantage and no performance to be gained here. Only strictness.

muglug:
I was surprised that PHP allows it, and I consider it a bug in PHP. _But_ I will allow arrays of ints, because it's hard to coerce those values.

muglug:
Btw the return type is currently provided in `StrReplaceReturnTypeProvider`, but we can probably get rid of that plugin.

orklah:
Seems like a good compromise. Thanks.

I'll check tomorrow if I can propose a PR for deleting StrReplaceReturnTypeProvider.

PR #3496, opened 2020-05-30
Fixed mysql::$insert_id type

staabm:
Per phpdocs:

> Return Values 
The value of the AUTO_INCREMENT field that was updated by the previous query. Returns zero if there was no previous query on the connection or if the query did not update an AUTO_INCREMENT value.
Note:
If the number is greater than maximal int value, mysqli_insert_id() will return a string.

https://www.php.net/manual/en/mysqli.insert-id.php

### DIFF
--- a/src/Psalm/Internal/PropertyMap.php
+++ b/src/Psalm/Internal/PropertyMap.php
@@ -260,7 +260,7 @@ return [
         'field_count' => 'int',
         'host_info' => 'string',
         'info' => 'string',
-        'insert_id' => 'int',
+        'insert_id' => 'int|numeric-string',
         'protocol_version' => 'string',
         'server_info' => 'string',
         'server_version' => 'int',
--- a/src/Psalm/Internal/PropertyMap.php
+++ b/src/Psalm/Internal/PropertyMap.php
@@ -260,7 +260,7 @@ return [
         'field_count' => 'int',
         'host_info' => 'string',
         'info' => 'string',
-        'insert_id' => 'int|numeric-string',
+        'insert_id' => 'int',
         'protocol_version' => 'string',
         'server_info' => 'string',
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/PropertyMap.php
```diff
@@ -260,7 +260,7 @@
         'field_count' => 'int',
         'host_info' => 'string',
         'info' => 'string',
-        'insert_id' => 'int',
+        'insert_id' => 'int|numeric-string',
```
weirdan, line 263: I wonder if this will cause unnecessary noise on common setups. Maximum autoincrement value for MySQL is BIGINT UNSIGNED (2^64-1), while PHP_MAX_INT is signed int64 (2^63-1) (on 64bit platforms).

staabm: I guess we need the numeric-string case because it depends on a flag wheter the mysql driver passes all data as strings or allows int/float etc

staabm: See MYSQLI_OPT_INT_AND_FLOAT_NATIVE on https://www.php.net/manual/en/mysqli.options.php

weirdan: Docs say it affects column types. Does it also affect `inser_id`?

staabm: Didnt tested it but it feels logical to me if it would also affect insert_id

weirdan: I just did, and it doesn't affect the type of `insert_id` property.

staabm: Thx for testing. I will change to int then

muglug:
Thanks!

PR #3497, opened 2020-05-30
Made mysqli_warning::$sqlstate a string

staabm:
Like in other mysql objects the sqlstate is a string
https://github.com/vimeo/psalm/pull/3497/files#diff-3a9cb9e92cf88595ca52bb250ebc0bd7L288
https://github.com/vimeo/psalm/pull/3497/files#diff-3a9cb9e92cf88595ca52bb250ebc0bd7L267


muglug:
Thanks!

PR #3499, opened 2020-05-30
Use scandir and glob flags to reduce their performance impact

orklah:
This PR applies suggestions from https://github.com/kalessil/phpinspectionsea about performance optimizations with scandir and glob.

The main optimization is to tell scandir and glob not to sort the results. I tried to look but the order of the files didn't seem critical.

There is also an optimization with glob which return only directories. This allows dropping the is_dir filter which need to fetch info from filesystem again.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -28,6 +28,7 @@ use function is_dir;
 use function is_file;
 use function json_decode;
 use function libxml_clear_errors;
+use const GLOB_NOSORT;
 use const LIBXML_ERR_ERROR;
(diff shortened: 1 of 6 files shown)


weirdan:
Is there any perceptible performance improvements?

orklah:
I didn't tried to measure it. It will be highly dependant on context (number of entries to sort), hardware, filesystem's cache and probably OS.

But it's not like it's a tradeoff. It's litteraly removing unnecessary operations.

muglug:
Thanks!

PR #3503, opened 2020-05-31
Improve stubs/callmap for str and preg replace functions

orklah:
This PR remove StrReplaceReturnTypeProvider following @muglug suggestion in #3495.

It also fix preg_replace return type because it can also return null (it was handled in callmap and plugin, but not in the stubs)

It appears that stubs do not respect ignore_internal_nullable_issues. Tests fails because of this.

@muglug do you have a opinion on how to fix this? I guess this should look like https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/FunctionAnalyzer.php#L254 but where stubs are parsed


muglug:
> @muglug do you have a opinion on how to fix this?

That'll have to be a _bit_ of a hack, unfortunately – I think it should be handled in `FunctionCallAnalyzer`, adding the `ignore_falsable_return` property to the returned type there instead.

orklah:
I'm making some progress.

I'll have to investigate on remaining psalm error. I don't understand why they're popping up just now. Plus, the preg_replace one is weird, it ignored my conditional return (or it read directly from the callmap instead of stubs, and that's weird)

muglug:
The problem is here: https://github.com/vimeo/psalm/blob/96175988ccfd4176f15775c6d860728711a79c44/src/Psalm/Config/Creator.php#L230

The first time the loop is entered Psalm doesn't know the type of `$path`, so it chooses the union of both branches of the conditional return type of `preg_replace`.

Why is this only a problem now? Psalm's existing hook does something a _little_ dirty. In the event it cannot determine a type for the string/array we're manipulating, it assumes it's a string, so types the output as string. Instead of "garbage in, garbage out" it's "garbage in, string out". 

The thing is, 99.9% of the time when the type passed to `preg_replace` is not known it's safe to assume it's a string. This assumption is pretty handy in code that doesn't have many types.

So I'd suggest removing the stubs for `str_replace` and `preg_replace` (and keeping the hook for those two). I know it's a pain, and I'm sorry I didn't consider this before.

orklah:
> The first time the loop is entered Psalm doesn't know the type of $path, so it chooses the union of both branches of the conditional return type of preg_replace.

Oh I wasn't aware conditional returns worked like that. So they're applied only when the type is known? (so I guess, when it's different than mixed?). 
I always thought mixed cases would fall into the "else" of the conditional...

> So I'd suggest removing the stubs for str_replace and preg_replace (and keeping the hook for those two). I know it's a pain, and I'm sorry I didn't consider this before.

It's okay :) 
I'll change that later.

Thanks for the review!

orklah:
@muglug I'm reverting my changes, but I realize that str_replace and preg_replace were added in https://github.com/vimeo/psalm/commit/187b944 as a part of your taint analysis. Should we still remove them from the stubs?

orklah:
Oh, Reading https://github.com/vimeo/psalm/pull/3495#issuecomment-636384893 again, I realize stubs and hooks work together so there is no need to remove functions in stubs as long as we keep the hook.

I'll make the changes

orklah:
Ok, so I kept improvements in stubs and callmap (even if callmap is not used anymore) and I also kept the flag for ignore internal null and false issues, even if it's not very relevant to this PR anymore (I can create a separate PR if needed)

I was thinking, it could be helpful for future contributors if there was a small comment in stubs and callmap to warn when a return type is handled by a hook. I could create a PR for that if you want

orklah:
Two tests are failing. I think the ignore_falsable_internal applied to rand(0, 1) makes it returning only truthy values.

I don't think that's how it should behave and rand() is not in stubs so it shouldn't even change that. I'll check that when I have time

orklah:
I'm closing this. It lost some of it's initial goal and I don't have time right now

PR #3514, opened 2020-06-03
Fix signatures of stream_filter_append and stream_filter_prepend

LeSuisse:
filterparams is not restricted to an array, it is perfectly to pass a
class or any other value.

See https://github.com/php/php-src/blob/php-7.4.6/ext/standard/streamsfuncs.c#L1171


muglug:
Thanks!

PR #3516, opened 2020-06-03
Proposal: Use different psalm cache for different versions

TysonAndre:
Older psalm versions may be missing cache entries that newer psalm
versions may cache.
Newer versions may remove/use a different form of cache entries
than that expected by older psalm versions.

Motivation: I've seen build failures when updating psalm on one branch but not others when the directory is on a shared CI host

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -827,7 +827,7 @@ class Config
 
         $config->global_cache_directory = $config->cache_directory;
 
-        $config->cache_directory .= DIRECTORY_SEPARATOR . sha1($base_dir);
+        $config->cache_directory .= DIRECTORY_SEPARATOR . sha1($base_dir . '::' . PSALM_VERSION);
 
         if (is_dir($config->cache_directory) === false && @mkdir($config->cache_directory, 0777, true) === false) {
             trigger_error('Could not create cache directory: ' . $config->cache_directory, E_USER_ERROR);


muglug:
The cache should be cleaned out whenever the version changes - if that’s not happening, that should be fixed instead. I don’t want a situation where each version bump increases the cache size.

muglug:
> when the directory is on a shared CI host

Also I highly recommend not running Psalm on a shared CI at the same time while sharing the same cache directory, because bad things will happen.

TysonAndre:
> Also I highly recommend not running Psalm on a shared CI at the same time while sharing the same cache directory, because bad things will happen.

It isn't - psalm is running in a folder named web-psalm, and only one psalm instance is running on that folder. If multiple psalm jobs are scheduled at the same time, they'd run on web-psalm@2, web-psalm@3, separately.
Psalm uses different folders within `/tmp/psalm/<hash>` already.

> The cache should be cleaned out whenever the version changes

However, they are run on different git branches (not simultaneously), some of which use newer psalm versions, some which use older psalm versions, and switching between new and old is typical. The psalm version is fixed for the git branch to make builds reproducible, so clearing cache would be impossible without just not using cache at all.

Would you consider making an opt-in option like `usePsalmVersionForCacheKey="true"`? Not particularly urgent, since I can easily work around this for myself

> Also I highly recommend not running Psalm on a shared CI at the same time while sharing the same cache directory, because bad things will happen.

This is the approach I planned to take to not "share the same cache (sub)directory"

PR #3518, opened 2020-06-04
Fix #3517 - Changed ignoreInternalFunctionFalseReturn default to true

### LINKED ISSUE #3517: `ignoreInternalFunctionFalseReturn` default value
The docs https://psalm.dev/docs/running_psalm/configuration/#ignoreinternalfunctionfalsereturn and the xsd https://github.com/vimeo/psalm/blob/master/config.xsd#L47 claim its default value is `false`.

But it looks like it's the opposite.

zerkms:
In config.xsd and the documentation


muglug:
Thanks!

PR #3547, opened 2020-06-08
Fix #3521 - Add --set-baseline filename to --config specified file

### LINKED ISSUE #3521: `--set-baseline` always adds to default config
Running `psalm --config=blah.xml --set-baseline=blah-baseline.xml` I would expect the baseline reference to be added to `blah.xml` but it's added to `psalm.xml`

andrei-petre:
Fixes vimeo/psalm#3521

I found that it's about the call to update_config_file() from psalm.php which uses the Config::locateConfigFile that only searches for Config::DEFAULT_FILE_NAME filenames

The call to Config::locateConfigFile('...path\blah.xml') returns actually '...path\psalm.xml' (or '\path'psalm.xml.dist)

I choosed to skip locateConfigFile based on is_dir (see the call from psalm.php , second argument like : $path_to_config ?? $current_dir)

### DIFF
--- a/src/command_functions.php
+++ b/src/command_functions.php
@@ -442,7 +442,11 @@ function update_config_file(Config $config, string $config_file_path, string $ba
         return;
     }
 
-    $configFile = Config::locateConfigFile($config_file_path);
+    $configFile = $config_file_path;
+
+    if (is_dir($config_file_path)) {
+        $configFile = Config::locateConfigFile($config_file_path);
+    }
 
     if (!$configFile) {
         fwrite(STDERR, "Don't forget to set errorBaseline=\"{$baseline_path}\" to your config.");


andrei-petre:
First time trying to contribute :) CircleCI reports 

There was 1 error:
1) Psalm\Tests\EndToEnd\PsalmEndToEndTest::testHelpReturnsMessage
Symfony\Component\Process\Exception\ProcessTimedOutException: The process "'composer' 'install' '--no-plugins'" exceeded the timeout of 60 seconds. in /home/travis/build/vimeo/psalm/vendor/symfony/process/Process.php:1253

I welcome any suggestions @muglug

muglug:
I'd ignore that – there has been some sort of issue with composer install, but not your fault. Thanks for this!

PR #3549, opened 2020-06-09
Fix RecursiveIteratorIterator stub to support Traversable iterators

simPod:


weirdan:
Docs say it accepts RecursiveIterator or IteratorAggregate (https://www.php.net/manual/en/recursiveiteratoriterator.construct.php), testing confirms it: https://3v4l.org/evDe2. Anyway it only makes sense for a recursive iterator of sorts. So what problem does this fix?

simPod:
@weirdan thanks! My bad. the issue originated from type inconsistence in Symfony interfaces and their implementations. Next time I must question their code and not php documentation.

weirdan:
> Next time I must question their code and not php documentation.

It's better to use actual PHP source code for type information. In my experience PHP docs often omit some type details (like `null` or `false` returns for invalid calls).

simPod:
Yea, I'm aware of it, thanks!

PR #3558, opened 2020-06-10
Add tests for getSymbolInformation across multiple files

joehoyle:
In #3477 support for getting symbol finformation on functions across files was added. This is a follow-up to add tests for that enhancement.

These are currently failing, because I haven't worked out how to get Psalm to analyze both files. This is currently throwing an error: `UndefinedFunction - somefile.php:2:24 - Function foo does not exist`. It seems either the codebase analyzer is not assuming `functions.php` is included, or they are somehow not getting anaylzed together.

I tried adding `require( "functions.php" );` to `somefile.php`, but this results in a "file not found" error, as it seems it's looking on the filesystem rather than looking at the in-memory file map.

@muglug could you perhaps provide me some guidance on how I should be writing a test that is across more than one file. I had a look through other tests, but I didn't find any that are adding multiple files to a project.

Thanks!

### DIFF
--- a/tests/LanguageServer/SymbolLookupTest.php
+++ b/tests/LanguageServer/SymbolLookupTest.php
@@ -94,6 +94,35 @@ class SymbolLookupTest extends \Psalm\Tests\TestCase
     /**
      * @return void
      */
+    public function testCrossFileSymbolLookup()
+    {
+        $this->addFile(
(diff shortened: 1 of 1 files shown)


orklah:
Please reopen if you want to start on that!

PR #3565, opened 2020-06-11
Fix return type for preg_grep

tvdijen:
Will return false when pattern cannot be compiled.

https://github.com/php/php-src/blob/a5e9950c7a491bb945888401196c8d685aaba61b/ext/pcre/php_pcre.c#L2861

https://psalm.dev/r/369fe6a46b


tvdijen:
Thanks for your efforts @muglug !

Issue #3037, opened 2020-03-29
Extending final classes should be prohibited


vudaltsov:
https://psalm.dev/r/433b6930be

I would like to suggest supporting the `@final` annotation. It's commonly used for Doctrine entities that cannot be final because of the proxies.

PR #3576, opened 2020-06-12
Extending final class is prohibited #3037

andrei-petre:
I tried a simple implementation for prohibiting extending final classes, both by the final keyword and also by the `@final` annotation.
I set the level -1 because in php is a fatal error.

The SHORTCODE seems to link to a doc webpage so should be updated.

### DIFF
--- a/docs/running_psalm/issues/InvalidExtendClass.md
+++ b/docs/running_psalm/issues/InvalidExtendClass.md
@@ -1,0 +1,18 @@
+# InvalidExtendClass
+
+Emitted when attempting to extends a final class or annotated as final.
+
+```php
+<?php
+
+final class A {}
+
+class B extends A {}
+
+/**
+ * @final
+ */
+class DoctrineA {}
+
+class DoctrineB extends DoctrineA {}
+```
--- a/tests/ExtendsFinalClassTest.php
+++ b/tests/ExtendsFinalClassTest.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Psalm\Tests;
+
+class ExtendsFinalClassTest extends TestCase
+{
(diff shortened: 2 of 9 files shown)


docs/running_psalm/issues/InvalidExtendClass.md
```diff
@@ -0,0 +1,18 @@
+# InvalidExtendClass
+
+Emitted when attempting to extends a final class or annotated as final.
```
weirdan, line 3:
```suggestion
Emitted when attempting to extends a final class or a class annotated with `@final`.
```

tests/ExtendsFinalClassTest.php
```diff
@@ -0,0 +1,52 @@
...
+namespace Psalm\Tests;
+
+class ExtendsFinalClassTest extends TestCase
+{
+    use Traits\InvalidCodeAnalysisTestTrait;
+
+    /**
+     * @return iterable<string,array{string,error_message:string,2?:string[],3?:bool,4?:string}>
+     */
+    public function providerInvalidCodeParse()
```
weirdan, line 12: I'd suggest adding a couple of tests showing how this new issue could be suppressed (with a docblock on the extending class, I suppose?).

andrei-petre: Thanks for the suggestions. Added two examples for suppression with docblock.

muglug:
Thanks! Awesome contribution!

andrei-petre:
One more little thing, mentioned also in the first message, the SHORTCODE of InvalidExtendClass should be updated as appropriate because now is pointing at another issue page.
Unfortunately i couldn't figure out how to set it.

Thanks a lot!

Issue #3574, opened 2020-06-12
UnusedClosureParam should not be reported for params defined before used params


ostrolucky:
https://psalm.dev/r/036ec7d178

Removing first argument would break the caller that calls the closure with both arguments

muglug:
Prepend the name with `_`: https://psalm.dev/r/8260d3972e

weirdan:
I would consider it a bug if it *wasn't* reported. You can suppress the error or mark the parameter as ignored with underscore: https://psalm.dev/r/8260d3972e

ostrolucky:
Where can I find this underscore convention documented? Ideally message itself should explain this.

weirdan:
> Where can I find this underscore convention documented? Ideally message itself should explain this.

In the linked PR (#3580) for now :smile:

PR #3580, opened 2020-06-12
Documented using underscore to suppress unused(param/variable) issues

weirdan:
Previously it was undocumented, as pointed out in vimeo/psalm#3574


muglug:
Thanks!

PR #3599, opened 2020-06-17
tidyNode->child will be null if the node does not have any children

Daeroni:
https://github.com/php/php-src/blob/master/ext/tidy/tidy.c#L696


muglug:
Thanks!

PR #3612, opened 2020-06-18
Fix #3607 - constant string class reference with leading backslash

### LINKED ISSUE #3607: Unused class/method misses const strings with leading backslash
This one is fine:
https://psalm.dev/r/17dc5a2e45

Psalm misses this one:
https://psalm.dev/r/c5d8cd84d8

But adding a leading backslash is legal:
https://3v4l.org/R1B3f

Thanks!

iluuu1994:
Let me know if this fix is appropriate 🙂

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -641,7 +641,9 @@ class FunctionCallAnalyzer extends CallAnalyzer
 
                         if (strpos($var_type_part->value, '::')) {
                             $parts = explode('::', strtolower($var_type_part->value));
-                            $potential_method_id = new \Psalm\Internal\MethodIdentifier($parts[0], $parts[1]);
+                            $fq_class_name = $parts[0];
+                            $fq_class_name = \preg_replace('/^\\\\/', '', $fq_class_name);
+                            $potential_method_id = new \Psalm\Internal\MethodIdentifier($fq_class_name, $parts[1]);
                         } else {
                             $explicit_function_name = new PhpParser\Node\Name\FullyQualified(
                                 $var_type_part->value,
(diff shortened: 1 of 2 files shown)


iluuu1994:
Hm, this breaks something... I'll take a look.

iluuu1994:
@muglug Very peculiar. As soon as I touch `$fq_class_name` in `MethodIdentifier::__construct` Psalm breaks... Even just fetching it and then doing nothing.

```php
$fq_class_name;
```

Any idea what's going on here?

src/Psalm/Internal/MethodIdentifier.php
```diff
@@ -13,6 +13,9 @@ class MethodIdentifier
      */
     public function __construct(string $fq_class_name, string $method_name)
     {
+        // remove trailing backslash if it exists
```
staabm, line 16: The comment is wrong... it removes a leading backslash, no?

staabm: Could this just be a `ltrim(...)` instead?

iluuu1994: > The comment is wrong... it removes a leading backslash, no?

Oh yeah, I just moved that code block, the comment was already wrong :) 

> Could this just be a ltrim(...) instead to ease reading...?

That would remove multiple backslashes which would make Psalm thing that `\\Foo` is valid but it's not.

muglug:
> @muglug Very peculiar. As soon as I touch $fq_class_name in MethodIdentifier::__construct Psalm breaks... Even just fetching it and then doing nothing... Any idea what's going on here?

Yeah, I didn't bother to define property types because Psalm was able to infer them in the scanning stage. Adding the input sanitisation breaks that (though you can always add the property types yourself).

I'd encourage you instead to do the sanitisation before the call to `new MethodIdentifier` – that way you're not penalising every call (which almost never comes from a specific string).

iluuu1994:
> I'd encourage you instead to do the sanitisation before the call to new MethodIdentifier – that way you're not penalising every call (which almost never comes from a specific string).

Ok, I'll do that 👍

iluuu1994:
@muglug Done 🙂

muglug:
Thanks!

PR #3615, opened 2020-06-18
use original case in error messages when reporting undefined methods

### LINKED ISSUE #3473: Invalid case in reported method names
Method `XY` is reported as `xy`. Is there any reason? Thanks!  
https://psalm.dev/r/7a4846bdca

andrei-petre:
fix #3473


weirdan:
@andrei-petre please rebase, it should fix (at least some) build failures.

muglug:
Thanks!

Issue #3655, opened 2020-06-23
Taint detection does not work for encapsulated strings (echo "$unsafe";)


TysonAndre:
Observed: No TaintedInput is emitted, but it would be emitted if the quotes are removed
Expected: TaintedInput is emitted 

`$pdo->exec("select * from users where name='" . $name . "'")` in https://psalm.dev/docs/security_analysis/ suggests taint detection already works for concatenation but not encapsulation

```php
<?php

$unsafe = $_GET['unsafe'];
echo "$unsafe";
```

TysonAndre:
src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php `analyze()` seems relevant for the concat implementation. 

src/Psalm/Internal/Analyzer/Statements/Expression/EncapsulatedStringAnalyzer.php is a candidate of where to add this taint tracking for encapsulated strings?

muglug:
just FYI you can reproduce in psalm.dev by having `<?php // --taint-analysis` at the top

weirdan:
> just FYI you can reproduce in psalm.dev by having `<?php // --taint-analysis` at the top

Does it work for other switches / settings?

muglug:
That’s a negative (but feel free to add support for individual ones over at psalm.dev’s repo)

PR #3656, opened 2020-06-23
Fix taint checking for encapsulated strings

TysonAndre:
For #3655


muglug:
Sorry! Though this was pretty much identical to what I added, so “great minds” and all that

PR #3657, opened 2020-06-23
Drop missing issues from XSD schema

weirdan:
Two unknown issues (that were only present in schema) are dropped and a test to validate that all issues are covered by XSD schema is added.


muglug:
Thanks!

PR #3683, opened 2020-06-25
Add psalm-taint-specialize for preg_replace_callback

TysonAndre:
Fixes https://psalm.dev/r/517c4a169e

```php
<?php // --taint-analysis
$x = preg_replace_callback('/a/', function($match) {
    return 'b';
}, $_GET['a']);

$other = preg_replace_callback('/a/', function($match) {
    return '';
}, 'a literal string');
echo $other;  // should not warn, but $a should
```


muglug:
Thanks!

Issue #3457, opened 2020-05-25
Psalm-internal and direct usage of method from internal class


skolman:
https://psalm.dev/r/17f080d422
```
<?php
namespace App\A {
    /**
     * @internal
     * @psalm-internal App\A
     */
    class Foo {
        public static function abc(): int { return 1; }
    }
}

namespace App\B {
    class Bat {
        
        /**
        * @var \App\A\Foo
        */
        public $a;
        
        public function __construct(\App\A\Foo $a)
        {
            $this->a = $a;
        }
        
        public function batBat() :void {
            $a = $this->a->abc();
            $b = \App\A\Foo::abc();
            $c = new \App\A\Foo();
        }
    }
}
```

Psalm returns

```
ERROR: InternalClass - 27:18 - App\A\Foo is marked internal to App\A
ERROR: InternalClass - 28:18 - App\A\Foo is marked internal to App\A
```
and this is fine, but what about the line `$a = $this->a->abc();`? Psalm did not find any issue. Why?

It works fine when I use `@internal` (but namespaces are not completely different)
https://psalm.dev/r/6add224dce

```
<?php
  namespace A {
    /**
     * @internal
     */
    class Foo {
        public static function abc(): int { return 1; }
    }
}

namespace B {
    class Bat {
        
        /**
        * @var \A\Foo
        */
        public $a;
        
        public function __construct(\A\Foo $a)
        {
            $this->a = $a;
        }
        
        public function batBat(): void
        {
            $a = $this->a->abc();
            $b = \A\Foo::abc();
            $c = new \A\Foo();
        }
    }
}
```

Output:

```
ERROR: InternalMethod - 25:28 - The method A\Foo::abc has been marked as internal
ERROR: InternalClass - 26:18 - A\Foo is marked internal to A
ERROR: InternalMethod - 26:18 - The method A\Foo::abc has been marked as internal
ERROR: InternalClass - 27:18 - A\Foo is marked internal to A
```

weirdan:
This: https://psalm.dev/r/a4f18d346f should be marked as error. I think it doesn't matter where the error is generated, could as well be reported at `@var \A\Foo` line and/or constructor.

bdsl:
[...]

PR #3698, opened 2020-06-27
Treat methods of internal or psalm internal classes as internal

bdsl:
When both the method and the class are annotated as psalm-internal,
but to different namespaces, we consider the method internal to
whichever namespace is longer, i.e. the smaller code module.

Issue reported at https://github.com/vimeo/psalm/issues/3457 by @skolman

I'd also like to gauge interest in `@internal` and `@psalm-internal` more generally. If people think it would be useful I could do some more work on catching the issues in the second list at https://github.com/vimeo/psalm/pull/1623#issuecomment-491529583

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
+++ b/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
@@ -68,6 +68,7 @@ use function substr;
 use function trim;
 use function preg_split;
 use php_user_filter;
+use function strlen;
 
 /**
  * @internal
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
```diff
@@ -2133,6 +2134,12 @@ private function registerFunctionLike(PhpParser\Node\FunctionLike $stmt, $fake_m
 
         $doc_comment = $stmt->getDocComment();
 
+
+        if ($class_storage && ! $class_storage->is_trait) {
```
bdsl, line 2138: We don't take the internal nature of the trait into account, because that would fail these existing test cases for valid code:

https://github.com/vimeo/psalm/blob/b27560617c88bed9df0ee6dc80ab9487313a4b95/tests/PsalmInternalAnnotationTest.php#L221 https://github.com/vimeo/psalm/blob/b27560617c88bed9df0ee6dc80ab9487313a4b95/tests/InternalAnnotationTest.php#L175

Issue #3697, opened 2020-06-27
`echo $taintedObject;` does not infer taintedness from __toString()


TysonAndre:
Same for string encapsulation (`"Trace: $myClass"`)

Expected: The examples in https://psalm.dev/r/43fab1bc2a should emit TaintedOutput

Related to #3696

PR #3705, opened 2020-06-28
Track taint in some implicit `__toString()` casts

TysonAndre:
Partially support #3697

There are other ways that objects can be cast to string that this PR
does not analyze (concat, `echo $unsafeStringableObject;`).
I'm not familiar enough with psalm to know the best way to implement it or if it'd make sense for psalm to support that.


muglug:
I did this myself (mostly via modifications to the `castStringAttempt` method)

PR #3724, opened 2020-07-01
TemplateChecker: Ensure it's working again

jarstelfox:
The template checker is erroring on my first template file due to storage missing on my pseudo method. I traced back to your original + noticed the optional param had been removed.


muglug:
Thanks!

PR #3725, opened 2020-07-01
docs(MoreSpecificImplementedParamType): less to more

binarious:


muglug:
Thanks! Less is more!

PR #3729, opened 2020-07-01
Mark create_function() as a taint sink

### LINKED ISSUE #3723: Mark create_function as an eval sink for taint detection?
https://www.php.net/create_function

> This function internally performs an eval() and as such has the same security issues as eval(). Additionally it has bad performance and memory usage characteristics.

https://psalm.dev/r/3824177fbb

Motivation: It's more harmful with user input than with hardcoded literal strings.

**NOTE: This is deprecated and will be removed in php 8.0 - psalm should ensure that targeting php 8.0 will infer this function does not exist**

TysonAndre:
create_function() is a thin wrapper around eval().
Fixes #3723

### DIFF
--- a/src/Psalm/Internal/InternalTaintSinkMap.php
+++ b/src/Psalm/Internal/InternalTaintSinkMap.php
@@ -4,6 +4,7 @@
 
 return [
 'exec' => [['shell']],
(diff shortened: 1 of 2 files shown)


muglug:
Thanks!

PR #3731, opened 2020-07-01
Support taint detection on Throwable::getTraceAsString()

### LINKED ISSUE #3696: Idea: Treat Throwable->getTraceAsString() and __toString() as taint source for secrets
And potentially html and text, since params in the trace can be anything and are unescaped.

```php
<?php

function login($username, $password) {
    throw new RuntimeException('login failure');
}
try {
    login('user', 'pass');
} catch (Exception $e) {
    // This output includes 'pass'
    echo $e, "\n";
    echo $e->getTraceAsString();
}
```

TysonAndre:
And `__toString()`, which uses getTraceAsString().

Fixes #3696

```php
function login($username, $password, $secret) {
    throw new RuntimeException('login failure');
}
try {
    login('user', $_GET['pass'], SECRET);
} catch (Exception $e) {
    // This output includes unescaped 'pass' and SECRET
    echo $e, "\n";
    echo $e->getTraceAsString();
}
```

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreImmutableClasses.phpstub
+++ b/src/Psalm/Internal/Stubs/CoreImmutableClasses.phpstub
@@ -55,8 +55,14 @@ interface Throwable
 
     /**
      * @psalm-mutation-free
+     * @psalm-taint-source input
      */
     public function getTraceAsString() : string;
+
+    /**
+     * @psalm-taint-source input
+     */
+    public function __toString() : string;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Stubs/CoreImmutableClasses.phpstub
```diff
@@ -195,6 +209,13 @@ class Error implements Throwable
 
     /**
      * @psalm-mutation-free
+     * @psalm-taint-source input
      */
     public final function getTraceAsString() : string {}
+
+    /**
+     * @psalm-mutation-free
```
muglug, line 217: Mind removing this? It breaks some downstream packages

muglug:
Thanks!

Issue #3735, opened 2020-07-02
\Psalm\Internal\Scanner\DocblockParser::parse() mangles the parsed docblock


lhchavez:
Since `\Psalm\Internal\Scanner\DocblockParser::parse()` calls `str_replace()` to remove the matched tags, if there are any tags that are substring of previous tags OR for whatever reason one tag's text appears in the description, they will get removed, therefore mangling the parsed description.

Example:

```php
/**
 * some self-referential desc with " * @return bool
 * " as part of it.
 *
 * @param string $bli
 * @param string $bli_this_suffix_is_kept
 * @param int $bla
 *
 * @throws \Exception
 *
 * @return bool
 */
```

will roundtrip as:

```php
/**
 * some self-referential desc with "
 * " as part of it.
 *

 _this_suffix_is_kept
 * @param string $bli
 * @param string $bli_this_suffix_is_kept
 * @param int $bla
 *
 * @throws \Exception
 *
 * @return bool
 */
```

PR #3736, opened 2020-07-02
Improve \Psalm\Internal\Scanner\DocblockParser::parse()

lhchavez:
This change avoids calling `str_replace()` on the original docblock and
instead only operates on the parsed (and modified) lines. This now makes
it so that if there are substrings of the docblock that match a tag
match, it won't get prematurely removed, therefore avoiding mangling of
the parsed docblock's description.

Fixes: #3735


muglug:
Thanks!

Issue #3744, opened 2020-07-04
Add print_r() as taint-sink


thomasbley:
print_r() can also produce html output if second parameter is not true.

example: https://3v4l.org/TUS98

https://psalm.dev/r/a7c15718ab

expected:
ERROR: TaintedInput - 11:7 - Detected tainted html
ERROR: TaintedInput - 13:6 - Detected tainted html

current:
ERROR: TaintedInput - 13:6 - Detected tainted html

TysonAndre:
Also mentioned in #3665 - As you said, the inconvenient part is that it depends on the second argument, like var_export

ohader:
This seems to be fixed with PR #5827

thomasbley:
Currently print_r($_GET['id'], true); and print_r($_GET['id']); raise "Detected tainted HTML", but that's ok for me.

emmanuelGuiton:
`Detected tainted HTML` is not that OK when the `$return` parameter of `print_r` is `true`. See this code snippet that does trigger an error while it should not : https://psalm.dev/r/6908a17232

The result of `print_r` is sanitized so there is no issue.
And unfortunately, using `@psalm-taint-escape html` has no effect to ignore this false-positive. I am using Psalm 4.8.1

Here is the analysis result :
```
Scanning files...
Analyzing files...

░

ERROR: TaintedHtml - sources/test.php:3:27 - Detected tainted HTML (see https://psalm.dev/245)
  $_GET
    <no known location>

  call to print_r - sources/test.php:3:27
echo htmlentities(print_r($_GET, true), ENT_QUOTES);



------------------------------
1 errors found
------------------------------

Checks took 0.48 seconds and used 40.769MB of memory
Psalm was able to infer types for 100% of the codebase
```

ohader:
Correct, current implementation is (too) strict and produces a false-positive here:
https://github.com/ohader/psalm/blob/689315b05fe671be80b7890aaa74dcd77604d02a/stubs/CoreGenericFunctions.phpstub#L369-L378

orklah:
I'm going to consider this as closed. Please create an issue for the conditional taint based on param if needed

PR #3754, opened 2020-07-05
Fix typo of `printr` - the function name is `print_r()`

TysonAndre:
Related to #3744

`print_r` is only a taint sink when `$return` is false or absent.


staabm:
Typo in the PR title: print_r

PR #3768, opened 2020-07-07
Test with PHP-Standard-Library

azjezz:


bin/test-with-real-projects.sh
```diff
@@ -30,13 +30,20 @@ proxymanager)
 	~/project/psalm --monochrome
 	;;
 
+psl)
+	git clone git@github.com:azjezz/psl.git
+	cd psl
+	composer install --ignore-platform-reqs
```
azjezz, line 36: ext-bcmath is required, but it is missing from CI.

azjezz: @muglug build seems to pass as psalm uses the stubs for bcmath, do you think this is fine or should i add ext-bcmath to the CI?

muglug: Tests passed, so I'm happy

muglug:
Thanks!

PR #3769, opened 2020-07-07
Fix "Could not get class storage" from cache

### LINKED ISSUE #3671: Uncaught InvalidArgumentException: Could not get class storage for other in ClassLikeStorageProvider.php:45 after removing subclass
This can occur when three things occur simultaneously before the next psalm run

1. A constructor A's statement list is modified
2. A class which invokes `A::__construct` is deleted
3. There is an issue in the way that class's constructor invokes `A::__construct`

```
» ../psalm/psalm  --threads=1
Scanning files...
Analyzing files...

Uncaught InvalidArgumentException: Could not get class storage for other in /path/to/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:45
Stack trace:
#0 /path/to/psalm/src/Psalm/Internal/Codebase/Analyzer.php(673): Psalm\Internal\Provider\ClassLikeStorageProvider->get('other')
#1 /path/to/psalm/src/Psalm/Internal/Codebase/Analyzer.php(251): Psalm\Internal\Codebase\Analyzer->loadCachedResults(Object(Psalm\Internal\Analyzer\ProjectAnalyzer))
#2 /path/to/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(621): Psalm\Internal\Codebase\Analyzer->analyzeFiles(Object(Psalm\Internal\Analyzer\ProjectAnalyzer), 1, false, true)
#3 /path/to/psalm/src/psalm.php(601): Psalm\Internal\Analyzer\ProjectAnalyzer->check('/path/to...', false)
#4 /path/to/psalm/psalm(2): require_once('/path/to...')
#5 {main}
(Psalm 3.x-dev@... crashed due to an uncaught Throwable)
```

Seen in de85e7c539033253ccece17d2b38ae6e9165cd55 and probably an issue for weeks or longer.

### Steps to reproduce

1. Create psalm.xml with cacheDirectory configured. Remove other files from src/ of the project
2. `rm -rf /path/to/cacheDir`
3. Replace src/base.php with the first file contents. Run `/path/to/psalm --threads=1`
4. Replace src/base.php with the second file contents. Run `/path/to/psalm --threads=1` again.
    Observed: Psalm crashes with the above stack trace on this run and subsequent runs

```php
<?php
// ==> src/base.php (version 1) <==
class base {
    public function __construct(string $required) {
        echo "\n";
    }
}
class other extends base {
    public function __construct() {
        parent::__construct();  // too few arguments is deliberate
    }
}
```

In the second version, the contents of `base::__construct` were changed, and class with the call to `base::__construct` which caused an issue to be emitted was removed

```php
<?php
// ==> src/base.php (version 2) <==
class base {
    public function __construct(string $required) {
    }
}
```

Psalm is also affected by manual calls to `new base()` from within a constructor.

This can be seen in CI when switching between branches. If one change adds a subclass and triggers an issue calling the parent constructor (and modifies the parent constructor), then triggering a build with an older branch will reproduce this issue.

TysonAndre:
Fixes #3671

This is better than an uncaught exception, at least, and I can detect
new issues if the constructor body changes.

### DIFF
--- a/src/Psalm/Internal/Codebase/Analyzer.php
+++ b/src/Psalm/Internal/Codebase/Analyzer.php
@@ -6,6 +6,7 @@ use function array_intersect_key;
 use function array_merge;
 use function count;
 use function explode;
+use InvalidArgumentException;
 use function number_format;
 use function pathinfo;
 use PhpParser;
@@ -669,9 +670,15 @@ class Analyzer
                             if ($referencing_base_classlike === $unchanged_signature_classlike) {
                                 $newly_invalidated_methods[$referencing_method_id] = true;
                             } else {
-                                $referencing_storage = $codebase->classlike_storage_provider->get(
-                                    $referencing_base_classlike
-                                );
+                                try {
+                                    $referencing_storage = $codebase->classlike_storage_provider->get(
+                                        $referencing_base_classlike
+                                    );
+                                } catch (InvalidArgumentException $_) {
+                                    // Workaround for #3671
+                                    $newly_invalidated_methods[$referencing_method_id] = true;
+                                    $referencing_storage = null;
+                                }
 
                                 if (isset($referencing_storage->used_traits[$unchanged_signature_classlike])
                                     || isset($referencing_storage->parent_classes[$unchanged_signature_classlike])


muglug:
Thanks!

PR #3772, opened 2020-07-08
Fix #3757 - allow multiple mixins

### LINKED ISSUE #3757: Only first @mixin recognized
Hey there,

I have the following problem using psalm in my project (using CakePHP 4):
ORM-Models can be extended with behaviors, some of them define additional methods that are magically available on the Model (through `__call`). These are annotated with `@mixin` in the class Docblock.
It is pretty common to have three or more behaviors and thus multiple `@mixin`-annotations.
However, i've noticed that only the first `@mixin` is actually respected. For the others, magic methods are not added to the class and an error is thrown upon usage.

Small Repro: https://psalm.dev/r/695ba45341

If this is indeed a bug, i might be able to find some time to try and fix it - though i have no experience with psalms code.
If this is intended, a hint would be appreciated :-) (Wasn't able to find much info on this topic)

xyng:
This pull requests resolves my issue that only the first `@mixin`-docblock is recognized.

Changes of this pr in short:

`ClassLikeDocBlockComment` has a new property `mixins`, which is a string array. The previous property `mixin` was marked deprecated since there may be some plugins that rely on it.

`ClassLikeStorage` has two new properties, one for Mixins of type `TTemplateParam` and one for Mixins of type `TNamedObject`. As in `ClassLikeDocBlockComment` the `mixin`-property is still there but marked as deprecated.

All uses of the now deprecated `mixin` attribute on these classes (that i could find) were changed so that they use the new array equivalents.
Cases where the `mixin` attribute was set have been changed so that they populate the arrays. The deprecated `mixin`-attribute is then set to the first encountered mixin (using `reset()`, as it was before).

This is my first time working on psalm, so any hints for improvements or spots that i may have missed are very much appreciated.

Thank you!


mr-feek:
Just wanted to thank you for the contribution @xyng! Classes certainly should allow multiple mixins 😄

xyng:
Just a short status report: i'm working on the failing tests (and creating a few that will test stuff for classes with multiple-mixins) but these run incredibly slow on my machine.
This might take me a little (esp. because of some other projects i have to get done 😉)

muglug:
>  i'm working on the failing tests (and creating a few that will test stuff for classes with multiple-mixins) but these run incredibly slow on my machine.

You can just run the `@mixin` tests with `vendor/bin/phpunit tests/MixinAnnotationTest.php`

xyng:
Sorry for the long delay!
Problem is fixed, tests seem to pass now.

I've added two test-cases covering multiple mixins.
Do you have further suggestions for more cases or are they fair enough for now?

xyng:
Quick note.
The failing case "test-with-real-projects" fails on the same step on master and is most likely unrelated to this pr:
https://app.circleci.com/pipelines/github/vimeo/psalm/3232/workflows/7b39fdfc-b8e0-4bcf-a930-09bcea9b502b/jobs/14953

muglug:
Thanks!

PR #3776, opened 2020-07-08
Support generating multiple report files

TysonAndre:
E.g. `psalm --report=report.txt --report=report.pylint`.
It may be useful to have both a machine-readable and human readable
report, e.g. during taint analysis.

Previously, psalm would generate neither report if the report option was
repeated.


TysonAndre:
Aside: Maybe the 'getopt' signature should be `list<string|false>` instead of `list<mixed>`, in CallMap.php

muglug:
Thanks!

PR #3780, opened 2020-07-08
Scan stub files before project files

### LINKED ISSUE #3328: Class extenting stub class is not analyzed
When I have a class that extends a class in my stub files, I don't get any analysis or failures on missing types in my subclass. I see `Populator could not find dependency (521)` when running `./vendor/bin/psalm --debug`. I can confirm the stub file is being ingested, because if I remove it, I got an `UndefinedClass` error.

I uploaded a very simple sample project to https://gist.github.com/joehoyle/203a52d80f21ce56636d45bac81f2b8f, you can clone and test locally with `git clone https://gist.github.com/203a52d80f21ce56636d45bac81f2b8f.git`

I see https://github.com/vimeo/psalm/issues/1528 which looks quite related, but then the tests in https://github.com/vimeo/psalm/commit/39f729667089bf704baba53075ce8399e772151e don't look like they are testing if invalid code shows as failing, when it should, so maybe this isn't be tested for right now.

joehoyle:
Currently project files are scanned before stub files, this means if project files have dependencies on things declared in stubs (such as child classes), analysis essentially fails, as the Populator will bail in `populateDataFromParentClass -> 'Populator could not find dependency ...'`.

I imagine there is some reason behind scanning the project files before the stubs, but I'm not sure what it is. I decided to go ahead and make this change and also add a test demonstrating the problem. without switching the order, the attached test case will not throw any errors (when it should!).

Fixes #3328.

### DIFF
--- a/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
@@ -1000,9 +1000,9 @@ class ProjectAnalyzer
 
         $this->config->initializePlugins($this);
 
-        $this->codebase->scanFiles($this->threads);
-
         $this->config->visitStubFiles($this->codebase, $this->progress);
+
+        $this->codebase->scanFiles($this->threads);
 
         $this->progress->startAnalyzingFiles();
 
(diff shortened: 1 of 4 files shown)


joehoyle:
@muglug any thoughts on this approach?

orklah:
I may be mistaken, but there is a preloadClasses attribute in config. Isn't that what you need?

orklah:
IIRC, there was work done since on the order of stubs loading. Could you check if the issue still happen on the latest release? Can you also check my last comment, maybe that could fix things

PR #3781, opened 2020-07-08
Always deep scan stubs

### LINKED ISSUE #3568: Stub files are deep scanned
I've been investigating why I'm seeing my (quite large) stub files getting a shallow scan, and then a deep scan. It seems to me that stub files should never get shallow _and_ deep scanned, as they only contain definitions, so function / method bodies etc are not present.

From the best I can tell, the reason for deep scanning is due to a project having a class that extends the stubbed class as triggered by https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Codebase/Scanner.php#L615. `$will_analyze` is `true` (analyzing the project file), and it (for some reason) forwards the `$will_analyze` to `queueClassLikeForScanning`. The subsequent lines for `required_interfaces` and `referenced_classlikes` hard code `$analyze_too` to `false`, so I'm wondering if this is an oversight?

This leads me to two questions I think:

1. Why does requiring a class in file A cause a deep scan on File B that has the class? Isn't a shallow scan enough?

1. Can stub-files always be short-circuited from deep scanning?

joehoyle:
In many sitations, stub files will receive a shallow _and_ deep scan when project files require extra analysys on things like parent classes. This makes stub file scanning inconsistent (orders become much less predictable for example), and adds extra process time to scan the files twice. In the case of stubs providing classes and functions for large projects, this is a non-trivial amount of time.

As deep scanning stubs should take just about as long as a shallow scan, it makes sense to just always deep scan them.

Fixes #3568.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -1792,7 +1792,7 @@ class Config
 
         foreach ($stub_files as $file_path) {
             $file_path = \str_replace(['/', '\\'], DIRECTORY_SEPARATOR, $file_path);
-            $codebase->scanner->addFileToShallowScan($file_path);
+            $codebase->scanner->addFileToDeepScan($file_path);
         }
 
         $progress->debug('Registering stub files' . "\n");


muglug:
Thanks!

Issue #3740, opened 2020-07-03
New annotation: psalm-self, as a pre-condition for the type of self when calling a method


olleharstedt:
Use-case: A type-safe query builder:

```php
<?php
    
class Query
{   
    /**
     * @var string|null 
     */
    public $from;
        
    /**
     * @param string|null $from
     */
    public function __construct($from = null)
    {
        $this->from = $from;
    }
  
    /**
     * @param string $from
     * @return void
     * @psalm-self-out Query<string>
     */
    public function set($from)
    {
        $this->from = $from;
    }
  
    /**
     * @psalm-self Query<string>
     * @return string
     */
    public function execute()
    {
        return 'result';
    }
}

$q = new Query();
$q->execute();  // Should show error
$q->set('tablename');
$q->execute();  // OK
```

Another alternative would be using $this together with param: `@param Query<string> $this`.

Thoughts? Using `@psalm-self-out` to change interface of $this will not work, because there will be multiple fields in the query object (both "select" and "from" need to be set at least to be a valid query, and this cannot easily be controlled by switching interface of "self" back and forth).

Related discussion: https://github.com/vimeo/psalm/issues/3207#issuecomment-650061673

Can make a PR if accepted.

muglug:
I think I prefer something more explicit – using a `@psalm-typestate` annotation, where a typestate is _similar_ to a template, with the key difference that it's mutable.

```php
<?php
   
/**
 * @psalm-typestate TTable as ?string
 */
class Query
{   
    /**
     * @var TTable
     */
    public $from;
        
    /**
     * @param TTable $from
     */
    public function __construct($from = null)
    {
        $this->from = $from;
    }
  
    /**
     * @template TNewTable as string
     * @param TNewTable $from
     * @return void
     * @psalm-self-out Query<TNewTable>
     */
    public function from($from)
    {
        $this->from = $from;
    }
  
    /**
     * @psalm-if-typestate TTable string
     * @return string
     */
    public function execute()
    {
        return 'result';
    }
}

$q = new Query();
$q->execute(); // fails
$q->from('tablename');
$q->execute(); // succeeds
```

olleharstedt:
For reference, here's the same concept with templates: https://psalm.dev/r/5281d54b75 Pretty awkward. :)

> I think I prefer something more explicit – using a @psalm-typestate annotation, where a typestate is similar to a template, with the key difference that it's mutable.

Interesting. Just have to think if we block out any use-case with this restriction. :thinking:

[...]

Issue #3766, opened 2020-07-07
New annotation: psalm-no-aliasing


olleharstedt:
Another even more dumbed down version of ownership semantics. :)

This is the other side of the "immutable structures are easier to reason about". Forcing mutable structures to be non-aliasing prevents spooky action from a distance. In essence, all classes should be either immutable or non-aliasing (or some other ownership category, like mutable only for the owner, but that's been discussed in another ticket).

For a class annotated with `psalm-no-annotation`, the following would be invalid:

```php
$f = new Foo();
$g = $f;  // Invalid
callme($f);  // Invalid
$this->bar = $f;  // Invalid
$this->baz = new Foo();  // OK? But only private? Can still be leaked, though
```

https://github.com/vimeo/psalm/issues/3614

olleharstedt:
This can actually kind of be enforced during runtime by checking the refcount of an object. if it's > 1, throw an OwnershipException.

olleharstedt:
> from an optim point of view, uniqueness/ownership/linearity are the same as an interprocedural pessimistic aliasing analysis

olleharstedt:
Gentrification  Gone  too  Far?Affordable  2nd-Class  Values  for  Fun  and  (Co-)Effect

https://www.cs.purdue.edu/homes/rompf/papers/osvald-oopsla16.pdf

PR #3782, opened 2020-07-08
New annotation: @psalm-if-this-is on methods

olleharstedt:
As discussed: https://github.com/vimeo/psalm/issues/3740

Todo in another PR: `psalm-self-out` should change to `psalm-this-out`.

Ignore the commit history, it's because I'm using the same fork for all my PRs. Or I'm stupid. :smile:


olleharstedt:
Not sure if the types should be compared with equality or with something else?

muglug:
> Not sure if the types should be compared with equality or with something else?

You should instead use `TypeAnalyzer::isContainedBy`, which would in turn mean that the condition in the tests `@psalm-if-this-is I` is always met (so you'll have to create a new "invalid" testcase).

olleharstedt:
> > Not sure if the types should be compared with equality or with something else?
> 
> You should instead use `TypeAnalyzer::isContainedBy`, which would in turn mean that the condition in the tests `@psalm-if-this-is I` is always met (so you'll have to create a new "invalid" testcase).

:+1:

olleharstedt:
@muglug Even when using `isContainedBy`, it does not check the interface; the failing test-case still fails. Missing functionality? The test `ReconcilerTest::providerTestTypeIsContainedBy` also does not check interfaces, only subclassing.

olleharstedt:
Also not sure about this error. Missing config for new test file?

> 1) Psalm\Tests\Config\ConfigTest::testAllPossibleIssues
> Psalm\Exception\ConfigException: Error on line 118:
>    Element '{https://getpsalm.org/schema/config}IfThisIsMismatch': This element is not expected.

danog:
This is __seriously__ useful, can't wait to see it merged! :)

olleharstedt:
> This is _seriously_ useful, can't wait to see it merged! :)

Didn't work on it much lately. Maybe should give it another go...

olleharstedt:
> Hi @olleharstedt! Do you intend to work on that again?

Maybe, if there's interest. Alternatively help someone else take over. I was waiting for muglug to respond to my comment, but now I've learned he's not involved in Vimeo and Psalm anymore. I don't know even know if anyone is merging or reviewing pull requests at the moment.

I'd probably have to start a new PR, I assume lots have changed the last year...

weirdan:
> I don't know even know if anyone is merging

That would be me.

> or reviewing pull requests at the moment.

Me and @orklah (and anyone interested in doing that, really).

olleharstedt:
> > I don't know even know if anyone is merging
> 
> That would be me.
> 
> > or reviewing pull requests at the moment.
> 
> Me and @orklah (and anyone interested in doing that, really).

Oh, ok! Well, this feature kind of need alias control to be tight. You want to close old PRs?

orklah:
>  You want to close old PRs?

No rush, I just wanted to clean up PR is they were abandoned, but we can keep this open if needed

olleharstedt:
> > You want to close old PRs?
> 
> No rush, I just wanted to clean up PR is they were abandoned, but we can keep this open if needed

Right. Here are some brainfarts I wrote down related to a possible no-alias feature, needed to make type-state safe: https://github.com/vimeo/psalm/issues/3766

One use-case for this feature is a type-safe query builder, as discussed here: https://github.com/vimeo/psalm/issues/3740

orklah:
I think it could even be used in Psalm's tests itself instead of using an external array of assertions

olleharstedt:
> I think it could even be used in Psalm's tests itself instead of using an external array of assertions

Not sure I get it. What could be used? What's an external array of assertions?

orklah:
This is what a unit test looks like in Psalm: https://github.com/vimeo/psalm/blob/master/tests/ArgTest.php#L23 it's a little code snippet with variable and the goal is to make Psalm succeed (by not emitting any issue). The other goal is to make variables in the snippet match an assertion array provided just below where each variable must match with a type.

This works great but it's limited to variable in the global scope (I think). With @psalm-if-this-on, I think we should be able to insert those assertions anywhere in the snippet

danog:
Resumed work on this PR in https://github.com/zoonru/psalm/tree/if-this-is, IDK if I should make a separate PR :)
Btw, I've still got a small issue with the branch I'd like to fix: I noticed that the UnionTypeComparator has issues when checking certain templated types, for example:
```php
<?php
/**
 * @template TKey as array-key
 * @template TValue
 *
 * @extends ArrayObject<TKey, TValue>
 *
 * Basic generic iterator, with additional helper functions.
 */
class MyIterator extends ArrayObject {
    /**
     * @psalm-if-this-is MyIterator<mixed, int>
     */
    public function sum(): int {
        $res = 0;
    	foreach ($this as $n) { $res += $n; }
        return $res;
    }
}

/**
 * @var MyIterator<string, int>
 */
$res = new MyIterator(['a' => 'b']);
$res->sum();
```

The following code will throw an assertion, even if `string` is effectively contained by `mixed`.

PR #3783, opened 2020-07-08
Fix offset calculation in getReferenceAtPosition

joehoyle:
`getReferenceAtPosition()` powers the LSP hover functionality, and it seems the logic to return the range of the hovered symbol is incorrect. Reading the `while` loop shows there is no sufficient breaking conditions so `$start_pos` will always be the last item in the `$reference_map`.

Before (hovering the variable on the left):

Screenshot 2020-07-08 at 9 58 40 PM

After: (again hovering on the variable on the left, show by the cursor as expected)

Screenshot 2020-07-08 at 9 57 35 PM

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1217,8 +1217,8 @@ class Codebase
             return null;
         }
 
-        $start_pos = null;
-        $end_pos = null;
+        $reference_start_pos = null;
+        $reference_end_pos = null;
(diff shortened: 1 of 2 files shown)


muglug:
Mind rebasing and fixing the code style issues?

joehoyle:
@muglug ok updated!

muglug:
Thanks!

PR #3800, opened 2020-07-10
Adding a runTaintAnalysis option into the config

erunion:
This adds a new `runTaintAnalysis` option into the root of the config for being able to run Psalm with the `--taint-analysis` option at all times.


muglug:
Thanks!

PR #3804, opened 2020-07-11
Fix jumping to definition on nullable parameters

joehoyle:
Currently it's not possible to "Go to definition" (LSP) on nullable args like `function( ?MyClass $class )` as the reference is stored as `MyClass|null` in the reference map, which will now resolve to a class name.

This PR removed any nullable type from the union before adding it to the reference map (as the reference map is only use to indicate a symbol was used in a given location, I think this makes sense).


muglug:
Thanks!

PR #3806, opened 2020-07-11
Fix going to definition on return type

joehoyle:
If a return type of a method or function is set incorrectly (with the PHP doc), then the references are not added for the `function() : MyClass` symbol, so the "Go to definition" feature of the LSP won't work. I don't believe an invalid return type or not should stop the symbol location being tracked (and not allowing code navigation).

In moved the symbol location tracking to be before the return early short circuit.


muglug:
Thanks!

PR #3830, opened 2020-07-16
Allow netresearch/jsonmapper ^3.0 to be installed

LeSuisse:
The major version seems to be the addition of the support of the PHP 7.4 class type properties.

https://github.com/cweiske/jsonmapper/blob/v3.0.0/ChangeLog


muglug:
Thanks!

Issue #3636, opened 2020-06-22
Support taint analysis for remaining common pure functions such as json_encode()


TysonAndre:
It seems like psalm only knows about functions in src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub . It may help to expand that list of functions to other pure functions such as json_encode(), base64_decode(), etc.

```php
<?php
$globalVariable = $_GET['evil'];
eval('echo ' . json_encode($globalVariable));
```

This example still allows evaluating arbitrary code, such as `echo "$(ls)"`

https://github.com/phan/phan/blob/master/src/Phan/Plugin/Internal/UseReturnValuePlugin.php may be of help, because it lists many common "pure" functions that return a value based on their inputs.

Off-topic notes:
---------------------

Aside: $_REQUEST combines $_GET and $_POST, so should that also be included as a source?
$_COOKIE can be set by browsers, so should that be considered for `eval` (but possibly not html)

Aside: json_encode() technically escapes `html`, because the only **reasonable** place to echo it is inside `<script>`. However, it might be a useful sanity check to assert that the file contains `</script>` after the echo line but before the next `<script>` substring, if any occur. Probably not worth the effort.

- e.g. `<p><?= json_encode($unsafe) ?></p>` is potentially unsafe for emitting (malformed) html tags

muglug:
> $_REQUEST combines $_GET and $_POST

Good point, I've fixed that

TysonAndre:
Miscellaneous notes I might find useful if working on this in the future:

In src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php, `$function_storage->return_source_params` and attributes such as `$function_storage->added_taints` and `$function_storage->removed_taints` seem to be how these propogate. (this may change)

- Adding core stubs may not work in all cases, such as conflicting with php 7.x or 8.x adding a brand new param? Not familiar with psalm's handling of core stubs or best practices.

  src/Psalm/PluginRegistrationSocket.php seems to provide one way for end users to register stubs with rare or internal PECL modules (config->addStubFile()), and those are handled the same way as core stubs
- A plugin could theoretically provide just the additional properties to inherit taintedness from and the removed_taints/added_taints for various functions.

muglug:
I improved things a little in e8be2c5, adding support for `(l|r)?trim` and `explode`, which lead to the discovery of 12 new XSS bugs in Vimeo's code.

muglug:
[...]

PR #3836, opened 2020-07-17
Allow taint through strval sprintf

adrienlucas:
This is partially related to issue #3636

Actually `sprintf` was already taint-able, but without any test to prove it.

:warning: the second commit is a strange and unexpected behaviour I encountered when using `implode` indirectly. I'm not sure how to fix that.


muglug:
Thanks!

Issue #3838, opened 2020-07-17
Wrong type for key() in stub


ygottschalk:
Introduced by b22e6e150066fe1ca803424c9dd0899e36d1ff7e

PR #3839, opened 2020-07-17
fixes wrong return type of key() #3838

ygottschalk:
Issue: #3838 
Don't know what to do with test `Psalm\Tests\EmptyTest::testValidCode()#unsetChangesComplicatedArrayEmptiness`. My solution was 54a3a26d6e3195f7d88611bfc540d1ee8c17d126 but you seem to not like it (c0909f318703fa15b2ae873cbc6d8bada623fbdc)

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
+++ b/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
@@ -136,7 +136,7 @@ function array_flip(array $arr)
  *
  * @param TArray $arr
  *
- * @return (TArray is array<empty, empty> ? null : (TArray is non-empty-array ? TKey : TKey|null))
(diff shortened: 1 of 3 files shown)


muglug:
Oh sorry I misunderstood what `key` does (I assumed it only returned `null` if the array is empty, but it can return `null` when the pointer is beyond the array length)

muglug:
Please add back that check I removed!

ygottschalk:
To me it looks like `Psalm\Tests\ArrayFunctionCallTest::testValidCode()#key ` ([Link](https://github.com/vimeo/psalm/blob/master/tests/ArrayFunctionCallTest.php#L765)) and `Psalm\Tests\ArrayFunctionCallTest::testValidCode()#keyNonEmptyArray` ([Link](https://github.com/vimeo/psalm/blob/master/tests/ArrayFunctionCallTest.php#L773)) are the same test (now).

PR #3840, opened 2020-07-17
Remove an if statement that repeats the previous if statement

TysonAndre:
They're the exact same variables.

This otherwise seems to be reasonable based on classHasCorrectCasing


muglug:
Thanks!

PR #3841, opened 2020-07-18
Fix multiple issues with @internal and @psalm-internal

### LINKED ISSUE #3706: Internal and psalm-internal fail to catch multiple issues
Using the `@internal` or `@psalm-internal` annotation on any symbol should prevent that symbol being referenced outside of the code module, but at present Psalm only partially achieves this. 

This issue covers multiple ways that Psalm doesn't always catch there in one issue here. Depending how much interest there is in this I might start working through these to fix, and refactoring to make the same code handle both of these annotations.

This issue is based on my comment https://github.com/vimeo/psalm/pull/1623#issuecomment-491529583. I fixed a similar issue in #3698. The issues I know about now are:

1.   Internal class used as type annotation for property or variable
2.   Internal class used as type annotation or native type hint for function param or return
3.   Internal class or interface used with instanceOf
4.   Internal class constant fetched (including ::class)
5. 
6.   Property of Internal Class fetched
7.   Property of Internal Class set.

I've numbered these in the snippet linked below. All except 6 and 7 apply equally to related namespaces, which `@psalm-internal` should control, and unrelated namespaces, which `@internal` should control. 6 and 7 are currently detected only for the latter.

https://psalm.dev/r/06ce5f098f

bdsl:
This makes the checking of `@internal` and `@psalm-internal` elements a lot more strict.

I also unified the handling of the two annotations, by treating `@internal` as an `@psalm-internal` to the first part of the relevant namespace.

However there are tests failing at present, and I can't easily work out what's gone wrong. I'd be happy if anyone else can work that out and/or fix it — I'm going to bed now.

Fixes #3706

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
+++ b/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
@@ -1309,8 +1309,12 @@ class ReflectorVisitor extends PhpParser\NodeVisitorAbstract implements PhpParse
                 }
 
                 $storage->deprecated = $docblock_info->deprecated;
-                if ($docblock_info->internal && ! $docblock_info->psalm_internal && null !== $this->namespace_name) {
-                    $storage->psalm_internal = $this->namespace_name->getFirst();
+
+                if ($docblock_info->internal
+                    && !$docblock_info->psalm_internal
+                    && $this->aliases->namespace
+                ) {
+                    $storage->psalm_internal = explode('\\', $this->aliases->namespace)[0];
                 } else {
                     $storage->psalm_internal = $docblock_info->psalm_internal;
                 }
@@ -2128,11 +2132,10 @@ class ReflectorVisitor extends PhpParser\NodeVisitorAbstract implements PhpParse
 
 
         if ($class_storage
-                && !$class_storage->is_trait
-                && $class_storage->psalm_internal
-            && (
-                null === $storage->psalm_internal ||
-                strlen($class_storage->psalm_internal) > strlen($storage->psalm_internal)
+            && !$class_storage->is_trait
+            && $class_storage->psalm_internal
+            && (!$storage->psalm_internal
+                || strlen($class_storage->psalm_internal) > strlen($storage->psalm_internal)
             )
         ) {
             $storage->psalm_internal = $class_storage->psalm_internal;
@@ -2191,13 +2194,14 @@ class ReflectorVisitor extends PhpParser\NodeVisitorAbstract implements PhpParse
             $storage->deprecated = true;
         }
 
-        if ($docblock_info->internal && ! $docblock_info->psalm_internal && null !== $this->namespace_name) {
-            $storage->psalm_internal = $this->namespace_name->getFirst();
-        }
-
-        if (null === $class_storage ||
-            null === $class_storage->psalm_internal ||
-            (null !== $docblock_info->psalm_internal
+        if ($docblock_info->internal
+            && !$docblock_info->psalm_internal
+            && $this->aliases->namespace
+        ) {
+            $storage->psalm_internal = explode('\\', $this->aliases->namespace)[0];
+        } elseif (!$class_storage
+            || !$class_storage->psalm_internal
+            || ($docblock_info->psalm_internal
                 && strlen($docblock_info->psalm_internal) > strlen($class_storage->psalm_internal)
             )
         ) {
--- a/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
+++ b/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
@@ -2127,9 +2127,9 @@ class ReflectorVisitor extends PhpParser\NodeVisitorAbstract implements PhpParse
         $doc_comment = $stmt->getDocComment();
 
 
-        if ($class_storage &&
-                !$class_storage->is_trait &&
-                $class_storage->psalm_internal &&
+        if ($class_storage
+                && !$class_storage->is_trait
+                && $class_storage->psalm_internal
             (
                 null === $storage->psalm_internal ||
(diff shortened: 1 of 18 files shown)


bdsl:
CircleCI Builds: https://app.circleci.com/pipelines/github/bdsl/psalm?branch=psalm-internal-work

bdsl:
I made a throwawy branch to see what this will do with real projects, even though the self-tests aren't 100% working yet. Testing versions of and PHPUnit, Doctrine Collections, Ocramius / Proxy Manager and azjezz/psl, there are three new issues found in total, all InternalClass errors in the tests of ProxyManager: [1](https://github.com/muglug/ProxyManager/blob/70a9faaf23b523640de6c60f41d29b05854da1c4/tests/ProxyManagerTest/Functional/AccessInterceptorScopeLocalizerFunctionalTest.php#L502), [2](https://github.com/muglug/ProxyManager/blob/master/tests/ProxyManagerTest/Functional/AccessInterceptorValueHolderFunctionalTest.php#L418), [3](https://github.com/muglug/ProxyManager/blob/70a9faaf23b523640de6c60f41d29b05854da1c4/tests/ProxyManagerTest/Functional/LazyLoadingValueHolderFunctionalTest.php#L380).

All three are based on this `@internal` annotation: https://github.com/sebastianbergmann/phpunit/blob/1a9dd1a392c6670a241e9e737cc385e50bbc6f7b/src/Framework/Exception/ExpectationFailedException.php#L22

muglug:
I think your changes to `src/*.php` can be removed (I namespaced those files), making this PR much smaller.

bdsl:
Right, your commit https://github.com/vimeo/psalm/commit/1e01e90fba5bfe4a7eacecc643f22ba6e0acfd11 makes portion of this unnecessary. I'll rebase on to the new master when the build passes.

bdsl:
> Right, your commit [1e01e90](https://github.com/vimeo/psalm/commit/1e01e90fba5bfe4a7eacecc643f22ba6e0acfd11) makes a portion of this unnecessary. I'll rebase on to the new master when the build passes.

Done - interactive rebased omitting f796554 and there were no conflicts!

src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
```diff
@@ -2122,8 +2127,14 @@ private function registerFunctionLike(PhpParser\Node\FunctionLike $stmt, $fake_m
         $doc_comment = $stmt->getDocComment();
 
 
-        if ($class_storage && ! $class_storage->is_trait) {
-            $storage->internal = $class_storage->internal;
+        if ($class_storage &&
+                !$class_storage->is_trait &&
```
muglug, line 2131: Small style thing: I prefer `&&` at the start of the line, not the end, as it makes it _slightly_ easier to scan

bdsl:
Was there a reason you didn't want `psalm-language-server.php` in the `\Psalm` namespace? I'm getting errors in that: https://app.circleci.com/pipelines/github/bdsl/psalm/249/workflows/7ac9afc2-ef48-485b-8eb3-6a0495f965d0/jobs/1060

src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
bdsl: 2df047136

muglug:
> I also unified the handling of the two annotations, by treating @internal as an @psalm-internal to the first part of the relevant namespace.

This will break existing stuff (as it has here) and I'm not terribly comfortable with that.

bdsl:
> This will break existing stuff (as it has here) and I'm not terribly comfortable with that.

I certainly didn't want to break anything. It just seemed like the same logic is applied to both and it will be easier for users to understand what's going on if they always behave consistently, and future development will be less work without the duplication. I know that I have a failing test but I've assumed that's because I've made a mistake somewhere - I'm not sure why in principle `namespace Foo; @internal` isn't always exactly equivalent to `namespace Foo; @psalm-internal Foo`

But I can try and find time in the next couple of days to un-unify the handling if you think it's needed and the other stuff in this PR is useful.

muglug:
So this will still be allowed?

```php
<?php

namespace Foo {
    new \Foo\Internal\SomeClass();
}

namespace Foo\Internal {
    /**
     * @internal
     */
    class SomeClass {}
}
```

bdsl:
> So this will still be allowed?

Yes, it should be. `@internal` in that context is interpreted as `@psalm-internal Foo`. Relevant line is https://github.com/bdsl/psalm/blob/7ce6928283087abd8b43c3bad07b8dee2188cf50/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php#L1313

muglug:
Sorry, you'll want to rebase again (I didn't properly namespace `src/psalm-language-server.php`

bdsl:
Thanks, `psalm` now passes.

muglug:
Tests should pass now – the issue was the way you were getting the namespace in `ReflectorVisitor`

bdsl:
Thanks Matt. In that case the build should pass as a whole if you merge https://github.com/muglug/ProxyManager/pull/2 .

bdsl:
This is [now failing](https://app.circleci.com/jobs/github/bdsl/psalm/1073) on the test-with-real-project PHPUnit analysis in a way that's similar to [how master is failing](https://app.circleci.com/pipelines/github/vimeo/psalm/3174/workflows/2ff22812-7aac-45c9-8acf-0c928c8487ac/jobs/14727), but not identical. Would be good if we can gen get master passing first before trying to fix that.

bdsl:
@muglug What's your thinking now about this PR? Is there anything you'd like me to change? The build is now failing only in the same way as the master build.

muglug:
Thanks! I’ll investigate that separately

Issue #3653, opened 2020-06-23
Invalid UndefinedClass thrown


ADmad:
https://psalm.dev/r/0f55461f32

PR #3842, opened 2020-07-18
Fix invalid UndefinedClass using array|callable

EvgeniiR:
Fix https://github.com/vimeo/psalm/issues/3653

Now, an array of strings will not be considered as a callable when passing it to a function that also accepts just an array of strings(`callable|array<string>`, `callable|array`, `callable|list` etc. )


muglug:
Thanks!

Issue #3844, opened 2020-07-18
psalm dependencies incompatible with composer2 and php < 7.4


jderusse:
by requiring `ocramius/package-versions`, psalm can not be installed with `composer 2` and php < 7.4.

`ocramius/package-versions` < 1.8 are not compatible with composer 2
`ocramius/package-versions` >= 1.8 requires > php 7.4

weirdan:
As a workaround, you should be able to require `composer/package-versions-deprecated` which replaces `ocramius/package-versions`.

PR #3846, opened 2020-07-19
Use composer/package-versions-deprecated instead of ocramius/package-versions

weirdan:
It allows wider range of PHP versions in addition to supporting both Composer v1 and v2

Fixes vimeo/psalm#3844

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "felixfbecker/language-server-protocol": "^1.4",
         "netresearch/jsonmapper": "^1.0 || ^2.0 || ^3.0",
         "nikic/php-parser": "^4.3",
-        "ocramius/package-versions": "^1.2",
+        "composer/package-versions-deprecated": "^1.8.0",
         "openlss/lib-array2xml": "^1.0",
         "sebastian/diff": "^3.0 || ^4.0",
         "symfony/console": "^3.4.17 || ^4.1.6 || ^5.0",
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -55,6 +55,11 @@
                 <referencedMethod name="PhpParser\Comment::getLine" />
             </errorLevel>
         </DeprecatedMethod>
+        <DeprecatedClass>
+            <errorLevel type="suppress">
+                <referencedClass name="PackageVersions\Versions"/>
+            </errorLevel>
+        </DeprecatedClass>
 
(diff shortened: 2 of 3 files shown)


weirdan:
`test-with-real-projects` can be fixed by suppressing the error in https://github.com/muglug/ProxyManager/blob/master/psalm.xml, similar to changes in this PR.

muglug:
Thanks!

Majkl578:
FWIW this change has caused dependency hell in real-world projects:
```

  Problem 1
    - Installation request for vimeo/psalm 3.13.1 -> satisfiable by vimeo/psalm[3.13.1].
    - Conclusion: remove ocramius/package-versions 1.9.0
    - Conclusion: don't install ocramius/package-versions 1.9.0
    - vimeo/psalm 3.13.1 requires composer/package-versions-deprecated ^1.8.0 -> satisfiable by composer/package-versions-deprecated[1.10.99, 1.10.99.1, 1.8.0, 1.8.1, 1.8.2].
    - don't install composer/package-versions-deprecated 1.10.99|don't install ocramius/package-versions 1.9.0
    - don't install composer/package-versions-deprecated 1.10.99.1|don't install ocramius/package-versions 1.9.0
    - don't install composer/package-versions-deprecated 1.8.0|don't install ocramius/package-versions 1.9.0
    - don't install composer/package-versions-deprecated 1.8.1|don't install ocramius/package-versions 1.9.0
    - don't install composer/package-versions-deprecated 1.8.2|don't install ocramius/package-versions 1.9.0
    - Installation request for ocramius/package-versions 1.9.0 -> satisfiable by ocramius/package-versions[1.9.0].
```

weirdan:
I believe the break is caused by composer/package-versions-deprecated#3. Can you track down what is requiring `ocramius/package-versions` and what's the specific require string is used?

Majkl578:
`ocramius/package-versions` is actually our root requirement for version `1.9.0`.
But `composer/package-versions-deprecated` has no `1.9.0` release, I don't know why but that's probably the reason.

weirdan:
Can you provide the link the project, if it's public? 

I suppose the motivation behind that change in `package-versions-deprecated` was that most people would have the requirement specified as `^1.9.0`, and that should have allowed Composer to consider `1.9.99` as acceptable and install `composer/package-version-deprecated`

PR #3847, opened 2020-07-19
Prevent tests from producing output

weirdan:
- These are considered risky by PHPUnit
- And they mess up PHPUnit output


muglug:
Thanks!

Issue #3843, opened 2020-07-18
Allow null for datefmt_create parameters


othercorey:
https://psalm.dev/r/1e9a214d71

The first 3 parameters are allowed to be null although they cannot be default.

https://3v4l.org/TSYJO

https://www.php.net/manual/en/intldateformatter.create.php

othercorey:
If I could find where this is mapped out, I could try to submit a PR.

weirdan:
See https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/CallMap.php

PR #3851, opened 2020-07-20
Allow null for locale, datetime and typetime of datefmt_create

othercorey:
Fixes https://github.com/vimeo/psalm/issues/3843


muglug:
Thanks!

Issue #3826, opened 2020-07-16
Question: how to avoid UndefinedGlobalVariable?


aivchen:
https://psalm.dev/r/4e3ca5bc9d

How to avoid UndefinedGlobalVariable error in this case (without suppression)? PhpStorm works ok with this annotation but not psalm...

Thank you!

ygottschalk:
Depending on where `$sports` is defined, you could use `$sports = include('script_defining_sports');`. Inside a function you could use `global $sports;` and in any other case you could do `$sports = $GLOBALS['sports'];`

TysonAndre:
src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php `analyze()` would be the relevant part to patch.

It might be useful to add an XML setting to psalm.xml such as a mapping from global name to a union type (e.g. `<globalsTypeMap>...</globalsTypeMap>`), to improve analysis results of projects that have globals used everywhere with a known type. Psalm's maintainers would have a better idea of whether this would be generally useful or cause issues, e.g. in combination with `--taint-analysis`

I worked on something related in a fork of psalm (TysonAndre/psalm), but that was for runkit/runkit7 superglobals, not regular globals. Something along the following lines might help in your use case, but again, XML settings for globals (not superglobals) would be better



```patch
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
@@ -153,6 +153,12 @@ class VariableFetchAnalyzer
             return ExpressionAnalyzer::analyze($statements_analyzer, $stmt->name, $context);
         }
 
+        // FIXME Hack: TODO replace with config
+        if ($stmt->name === 'sports' && $from_global) {
+            $stmt->inferredType = Type::parseString('object[]');
+            return true;
+        }
+
         if ($passed_by_reference && $by_ref_type) {
             AssignmentAnalyzer::assignByRefParam(
                 $statements_analyzer,
```

aivchen:
> Depending on where `$sports` is defined, you could use `$sports = include('script_defining_sports');`. Inside a function you could use `global $sports;` and in any other case you could do `$sports = $GLOBALS['sports'];`
> 
> **EDIT:** Or maybe add the file defining `$sports` to the files analysed by psalm

Thank you!

ygottschalk:
[...]

PR #3861, opened 2020-07-21
[DOC] Add option to supress a referenced but undefined global variable.

kesselb:
Add documentation as suggested by @ygottschalk at https://github.com/vimeo/psalm/issues/3826#issuecomment-661243709.


muglug:
Thanks!

PR #3863, opened 2020-07-21
[docs] property/method text on Configuration page

dshoreman:
When I was reading through the docs setting up my Psalm config recently, I was a little confused by the `usePhpDocPropertiesWithoutMagicCall` rule referencing `@method` and vice-versa. After some digging it seems #3748 added the Properties text (highlighted in red below) in the wrong order, at least as far as I can tell.

This PR resolves that and a couple other minor text fixes.

psalm


weirdan:
Oops, my bad.

muglug:
Thanks!

PR #3893, opened 2020-07-25
Detect redundant identity with true

greg0ire:
Using `=== true` on a known boolean results in the same boolean.


greg0ire:
The builds seem to be badly broken… looks like there is PHP 8-specific syntax somewhere…

weirdan:
> The builds seem to be badly broken… looks like there is PHP 8-specific syntax in phpunit:

Actually it's an old syntax (`class Match ...`) that's no longer valid for PHP 8 (due to the `match` keyword introduced).
I suspect `nikic/php-parser` needs to be pinned to the version that does not support PHP 8 syntax.

weirdan:
https://github.com/nikic/PHP-Parser/releases/tag/v4.7.0 - I think this is the culprit.

weirdan:
Already reported upstream: https://github.com/nikic/PHP-Parser/issues/690

muglug:
In the meantime, thanks @greg0ire!

PR #3903, opened 2020-07-27
Add negated identity with false case

greg0ire:
This is the opposite of `=== true` and works the same.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -2076,41 +2076,58 @@ class AssertionFinder
         if ($codebase
             && $source instanceof StatementsAnalyzer
             && ($var_type = $source->node_data->getType($base_conditional))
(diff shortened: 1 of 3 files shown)


greg0ire:
Note that I am not contributing something for `!== true` or `=== false` (yet) because some people are arguing that it would force them to use `!`, which is not visible enough for them.

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -626,7 +626,7 @@ private static function scrapeEqualityAssertions(
                     if (count($types) === 1 && isset($types['bool'])) {
                         if (IssueBuffer::accepts(
                             new RedundantIdentityWithTrue(
-                                'Comparing a boolean with true is redundant',
+                                'Identifying a boolean with true is redundant',
```
greg0ire, line 629: Small change in vocabulary here… it's not really a comparison.

weirdan:
Thanks! Note that you will have to enable strict binary operands option for this to have any effect.

PR #3911, opened 2020-07-30
Update docs on @psalm-internal as @internal no-longer has to be used …

bdsl:
…at same time

See #42ad366dc8ccaf34


muglug:
Thanks!

PR #3920, opened 2020-08-01
Fix hash_file signature

NicolasCARPi:
hash_file can return false if the file is not readable.

See https://3v4l.org/E0BeU


muglug:
Thanks!

PR #3939, opened 2020-08-04
Casing checker: Ignore call / call static

jarstelfox:
Line 52 ignores `__call`, but not `__callStatic`. This will help us since mainly this comes up when using https://github.com/myclabs/php-enum


muglug:
Thanks!

PR #3942, opened 2020-08-05
Update MissingReturnType.md

cedw032:
It would be useful if these pages included the fix.

If you are interested in this I could do the rest.  It would be a good learning exercise.


muglug:
> If you are interested in this I could do the rest. It would be a good learning exercise.

Sure, that'd be great! 

Some of these can also be automatically fixed e.g. https://psalm.dev/docs/manipulating_code/fixing/#missingreturntype

Issue #3945, opened 2020-08-05
Memory limit flag?


mr-feek:
Hiya 👋 

Would you be willing to accept a PR specifying a memory limit flag for invoking psalm? I have a large, legacy, monolithic codebase in [which the default memory limit](https://github.com/vimeo/psalm/blob/6919e8842332366ce9243aae12f2eb7349214578/src/psalm.php#L192) is not enough.

I also don't want to change my ini value for the whole project -- at runtime I want a different value (smaller) than what I allocate to psalm. ``

[PHPstan has a memory limit flag](https://phpstan.org/user-guide/command-line-usage#--memory-limit), would you be willing to accept a PR for the same thing here?

I tried working around it with the following:

`php -d memory_limit=5G ./vendor/bin/psalm.phar --use-ini-defaults` does not work
`ini_set("memory_limit", "5G");` in a custom autoloader does work

muglug:
sure, go crazy

PR #3947, opened 2020-08-05
feature: allow specifying memory limit

mr-feek:
should close https://github.com/vimeo/psalm/issues/3945

### DIFF
--- a/src/psalm.php
+++ b/src/psalm.php
@@ -12,6 +12,7 @@ require_once('command_functions.php');
 require_once __DIR__ . '/Psalm/Internal/exception_handler.php';
 
 use Psalm\ErrorBaseline;
+use Psalm\Exception\ConfigException;
 use Psalm\Internal\Analyzer\ProjectAnalyzer;
 use Psalm\Internal\Provider;
 use Psalm\Config;
@@ -24,6 +25,7 @@ use Psalm\Progress\VoidProgress;
 use function array_slice;
 use function getopt;
 use function implode;
+use function is_scalar;
 use function array_map;
 use function substr;
 use function preg_replace;
@@ -92,6 +94,7 @@ $valid_long_options = [
     'help',
     'ignore-baseline',
     'init',
+    'memory-limit:',
     'monochrome',
     'no-cache',
     'no-reflection-cache',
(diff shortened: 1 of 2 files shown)


src/psalm.php
```diff
@@ -189,7 +192,18 @@ function ($arg) use ($valid_long_options, $valid_short_options) {
...
     ini_set('display_startup_errors', '1');
-    ini_set('memory_limit', (string) (8 * 1024 * 1024 * 1024));
+
+    $memoryLimit = (8 * 1024 * 1024 * 1024);
+
+    if (array_key_exists('memory-limit', $options)) {
+        $memoryLimit = $options['memory-limit'];
+
+        if (!is_scalar($memoryLimit)) {
+            throw new ConfigException('Invalid memory limit specified.');
```
mr-feek, line 202: I'm not sure if this case is even possible -- however without the scalar check, psalm complained about the following: 

```
ERROR: PossiblyInvalidCast - src/psalm.php:202:38 - list<mixed> cannot be cast to string (see https://psalm.dev/190)
    ini_set('memory_limit', (string) $memoryLimit);
```

weirdan: `getopt()` will return a list if you specify the option multiple times. 
E.g `psalm --memory-limit=1 --memory-limit=2` will result in
```
array(1) {
  'memory-limit' =>
  array(2) {
    [0] =>
    string(1) "1"
    [1] =>
    string(1) "2"
  }
}
```

mr-feek: today I learned!

muglug:
Thanks!

mshamaseen:
@muglug how to use this? I tried `--memory-limit=1024M` , `--memory-limit=1024` but both of them failed.

AndrolGenhald:
@mshamaseen It [appears](https://github.com/vimeo/psalm/pull/3947/files#diff-3d788d0759fd9c0298e110bd208a42114d6b8c20a5f8f7f0221357fe40248efaR196-R199) that it must be specified in bytes (`--memory-limit=1073741824`).

orklah:
ini_set format should be accepted, a string is a valid scalar. So '1024M' should work.

Most probably you have a ini_set('memory_limit') in your own code that gets executed by your autoloader. Could you check if it could be the case? Try with --debug-by-line to try to see where Psalm crash

mshamaseen:
@orklah @AndrolGenhald it does not work, it says:
`vendor/bin/psalm --memory-limit=1073741824 ${CHANGED_FILES}
Target PHP version: 8.1 (inferred from composer.json)
Scanning files...
PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 491520 bytes) in /**/vendor/barryvdh/laravel-ide-helper/resources/views/helper.php on line 34
PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 65536 bytes) in /**/vendor/nunomaduro/collision/src/Highlighter.php on line 235
`

AndrolGenhald:
> ... `--memory-limit=1073741824` ... `Allowed memory size of 1073741824 bytes exhausted`

Looks fine to me, have you tried a higher limit?

PR #3948, opened 2020-08-05
feature: universal object crates

mr-feek:
This is a draft for implementing phpstan's concept of "[universal object crates](https://phpstan.org/config-reference#universal-object-crates)". 

What do we think? I have yet to test with an actual config file, but the unit test I added is passing.


weirdan:
I'd like to also have an annotation for this.

muglug:
Looks great! Rebase to fix the build, sorry

mr-feek:
@weirdan annotation would be cool! A bit out of my grasp though, feel free to follow up with it

muglug:
Sorry this took so long to merge, and thanks again!

PR #3950, opened 2020-08-06
narrowed `reset` and `end` return type

### LINKED ISSUE #3931: False positive `InvalidFalsableReturnType` on `reset(non-empty-array)`
https://psalm.dev/r/876aa2daf4
The [PHP docs](https://www.php.net/manual/en/function.reset.php) say that `reset` will only return `false` if the array is empty, but Psalm doesn't seem to understand that.

ygottschalk:
I was trying to fix #3931 but psalm seems to ignore the stub return type (at least on my machine atm).
Anybody thoughts on this?

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayPointerAdjustmentReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayPointerAdjustmentReturnTypeProvider.php
@@ -44,8 +44,6 @@ class ArrayPointerAdjustmentReturnTypeProvider implements \Psalm\Plugin\Hook\Fun
             return Type::getMixed();
         }
 
-        $isNotEmpty = false;
-
         if ($first_arg_array instanceof Type\Atomic\TArray) {
             $value_type = clone $first_arg_array->type_params[1];
             $isNotEmpty = $first_arg_array instanceof Type\Atomic\TNonEmptyArray;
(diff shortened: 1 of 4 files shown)


weirdan:
I think it's overridden with https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayPointerAdjustmentReturnTypeProvider.php

src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayPointerAdjustmentReturnTypeProvider.php
```diff
@@ -44,15 +44,27 @@ public static function getFunctionReturnType(
...
             $value_type = clone $first_arg_array->type_param;
+            $isNotEmpty = $first_arg_array instanceof Type\Atomic\TNonEmptyList;
         } else {
             $value_type = $first_arg_array->getGenericValueType();
+            $isNotEmpty = $first_arg_array->getGenericArrayType() instanceof Type\Atomic\TNonEmptyArray;
+        }
+
+        if ($value_type->isEmpty()) {
+            $value_type->removeType('empty');
+        }
```
ygottschalk, line 62: Not sure if this can lead to problems

src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayPointerAdjustmentReturnTypeProvider.php
```diff
@@ -44,15 +44,27 @@ public static function getFunctionReturnType(
...
+
+        if (($function_id !== 'reset' && $function_id !== 'end') || !$isNotEmpty) {
+            $value_type->addType(new Type\Atomic\TFalse);
         }
```
ygottschalk, line 66:
```suggestion
        if (($function_id !== 'reset' && $function_id !== 'end') || !$isNotEmpty) {
            $value_type->removeType('empty');
            $value_type->addType(new Type\Atomic\TFalse);
        }
```
This may be a better solution...

ygottschalk:
This should now be ready to merge.
Please review the code because I am not very familiar with the internals of psalm...
The errors reported by psalm get also reported on master so I think I didn't introduce them.

muglug:
Thanks!

ygottschalk:
You are welcome =D

PR #3953, opened 2020-08-07
Improve return type of DateTime::getLastErrors

irozgar:
Define the structure of the array returned by _date_get_last_errors()_, _DateTime::getLastErrors()_ and _DateTimeImmutable::getLastErrors()_.

The structure of the array is the one in the examples from [php docs](https://www.php.net/manual/en/datetime.getlasterrors.php#refsect1-datetime.getlasterrors-examples).


muglug:
Thanks!

PR #3956, opened 2020-08-07
Patch 1 typos

ktomk:
While I was looking in the code stumbled over some typos.

Only in docs and should be only typos (and one link correction).

### DIFF
--- a/docs/running_psalm/plugins/plugins_type_system.md
+++ b/docs/running_psalm/plugins/plugins_type_system.md
@@ -4,7 +4,7 @@ Psalm's type system represents the types of variables within a program using dif
 
 ## Union types
 
-All type Psalm's type information you are likely to use will be wrapped in a [Union Type](../../annotating_code/docblock_type_syntax.md#union-types).
+All type Psalm's type information you are likely to use will be wrapped in a [Union Type](../../annotating_code/type_syntax/union_types.md).
 
 The `Union` class constructor takes an array of `Atomic` types, and can represent one or more of these types at a time. They correspond to a vertical bar in a doc comment.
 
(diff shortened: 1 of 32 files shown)


docs/running_psalm/plugins/plugins_type_system.md
```diff
@@ -174,15 +174,15 @@ Note that these classes do sometimes change, so `Type::parseString` is always go
...
+The following example constructs types representing a string, a floating-point number, and a class called 'Foo\Bar\SomeClass'.
 
 ``` php
 new TLiteralString('A text string')
 new TLiteralFloat(3.142)
 new TNamedObject('Foo\Bar\SomeClass')
 ```
 
-Types within Psalm are always wrapped in a union as a convenience feature. Almost anywhere you may expect a type, you can get a union as well (property types, return types, argument types, etc). So wrapping a single atomic type (like TInt) in a union container allows to uniformly handle that type elsewhere, without repetitive checks like this:
+Psalm always wraps types in a union as a convenience feature. Almost anywhere you may expect a type, you can get a union as well (property types, return types, argument types, etc). So wrapping a single atomic type (like TInt) in a union container allows to uniformly handle that type elsewhere, without repetitive checks like this:
```
weirdan, line 185: This sounds like it's done automatically, but it's actually not.

ktomk: @weirdan yes thas has been turned in active language. If it's not psalm what is the subject that does the wrapping?

Or is it that this is a concept and while psalm evolves it has turned out this is a good model?

weirdan: In the context of plugins it could be the plugin itself. E.g when you need to provide a return type to Psalm in a method return type provider you would have to construct the atomic type and wrap it into a union, then return it from the hook.

ktomk: Sounds to me as if this is completely optional. I've reverted therefore to keep it passive so any plugin author can figure out their own whether or not the composite pattern is to theirs benefit or they restrain from applying it. Squashed all these changes to reset the PR.

Also removed the change of the error messages as it's too invasive for a simple PR.

muglug:
Thanks – doesn’t isn't a typo of course, but I don't mind the unabbreviated version

ktomk:
@muglug: Yes, sure. If you don't mind for some error messages I would do the same their or (could be a way for the docs as well) to use the apostrophe from us-ascii (single byte) instead of the (not consistently the case btw) the UTF-8 three-byte sequence. Esp. for error messages I think that is more portable (within the docs it's mixed, could benefit from streamlining as well). Let me know if you've got some preference, I could take care of that and file another PR.

PR #3959, opened 2020-08-08
Fix PSR-4 deprecation notices on composer install

ktomk:
Previously when checking out the repository and installing the source
package via `$ composer install` composer did spill deprecation notices for
four files not complying with the PSR-4 auto-loading standard.

This is a minor issue, composer install works and can be addressed by
fixing name-spaces (all of those three are in tests).

For the one function test, fixing the directory structure / dirname of the
(single) affected function test (fourth file).


muglug:
Thanks!

PR #3961, opened 2020-08-08
Resolve require_once command_functions.php

ktomk:
Considering the test-suite tests the code from src paired with tests,
not the working directory (or even worse the overall include_path runtime
setting) should lead to relative path resolution, but just the relative
location to the path of the test itself.

Fix by making use of the `__DIR__` constant to anchor the relative location
turning it into an absolute path.


muglug:
thanks!

Issue #3478, opened 2020-05-28
Cannot suppress DeprecatedClass on property


scheb:
Not sure if I'm doing it wrong, but I'm unable to suppress a `DeprecatedClass` error on a property.

```php
<?php

/**
 * @deprecated
 */
class TheDeprecatedClass {
    
}

class A
{
    /**
     * @psalm-suppress DeprecatedClass
     * @var TheDeprecatedClass
     */
    public $property;
}
```

```
ERROR: DeprecatedClass - 14:13 - Class TheDeprecatedClass is marked as deprecated
```

https://psalm.dev/r/7550e806a2

weirdan:
You can now: https://psalm.dev/r/9fd2bb74bd

PR #3962, opened 2020-08-08
fix for DeprecatedClass suppress on property

the-toster:
Related to #3478  
I have added test case for this issue, but can't find out, how it should be fixed.

For some reason, suppressed issues are empty here:
https://github.com/vimeo/psalm/blob/ea52b9d23add23f98cf4ae36812bd203945e7dcc/src/Psalm/Internal/Analyzer/ClassAnalyzer.php#L1084-L1092

I will be glad for any clue, in which point `supressed issues` should be filled.

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -3,6 +3,7 @@ namespace Psalm\Internal\Analyzer;
 
 use PhpParser;
 use Psalm\Aliases;
+use Psalm\DocComment;
 use Psalm\Internal\Analyzer\Statements\Expression\Call\ClassTemplateParamCollector;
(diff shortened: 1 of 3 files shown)


the-toster:
As I understand, only class-level suppressions are presened here:
https://github.com/vimeo/psalm/blob/ea52b9d23add23f98cf4ae36812bd203945e7dcc/src/Psalm/Internal/Analyzer/ClassAnalyzer.php#L1088
so, statement level supressions should be added here

the-toster:
it also leads another issues like this https://psalm.dev/r/dcafbd4219

the-toster:
One more doc-block parsing operation for properties added. Looks not very good, but it pass given test case.

Also, TypeChacker modification to respect given suppressed array (looks like this is not breaks anything, full testsuite still executing on my machine).

muglug:
Thanks!

PR #3966, opened 2020-08-09
quit using combineUnionTypes, when specific TSend provided.

### LINKED ISSUE #3944: Generator type with return tag, couldn't receive correct send type , internally.
Define Generator with override TSend by specific type , but seems couldn't analyze correct.

```php
<?php
/** @return Generator<int, string, DateTimeInterface, void> */
function gen () : Generator {
    $date = yield 1 => 'string';
    
    $date->format('m');
    
    return;
}
```

would raise
```
ERROR: PossiblyInvalidMethodCall - 6:12 - Cannot call method on possible string variable $date
```
https://psalm.dev/r/70e67c1c07

----
phpstan's behavior is expected result.
https://phpstan.org/r/20001c32-d195-48f5-9f18-30fd5348b0b5

sasezaki:
fix #3944


muglug:
Thanks!

Issue #3969, opened 2020-08-10
BeforeAnalyzeFileInterface is not being invoked


lhchavez:
`\Psalm\Plugin\Hook\BeforeAnalyzeFileInterface` is not being invoked anywhere, since `\Psalm\PluginRegistrationSocket` does not detect instances of this class in order to add it to the config.

The name of that interface is also inconsistent with the other hooks, which are of the form `AfterKINDAnalysisInterface`.

PR #3970, opened 2020-08-10
Fix File-related analysis plugin hooks

lhchavez:
This change does a few things:

* Renames `BeforeAnalyzeFileInterface` to `BeforeFileAnalysisInterface`.
* Adds a few more parameters to `BeforeFileAnalysisInterface`.
* Since the analysis was never called due to a bug, the above two
  changes are safe to do.
* Fix the above-mentioned bug, and now adds the hooks to the config.
* Update the documentation about the File-related analysis plugin hooks.
* Add a test to ensure that this does not break again in the future.

Fixes: #3969


docs/running_psalm/plugins/authoring_plugins.md
```diff
@@ -68,11 +68,12 @@ class SomePlugin implements \Psalm\Plugin\Hook\AfterStatementAnalysisInterface
 - `AfterCodebasePopulatedInterface` - called after Psalm has scanned necessary files and populated codebase data.
 - `AfterEveryFunctionCallAnalysisInterface` - called after Psalm evaluates any function call. Cannot influence the call further.
 - `AfterExpressionAnalysisInterface` - called after Psalm evaluates an expression.
+- `AfterFileAnalyisisInterface` - called after Psalm analyzes a file.
```
andrew-demb, line 71: Typo

lhchavez: Fixed!

muglug:
Thanks!

TysonAndre:
> Renames `BeforeAnalyzeFileInterface` to `BeforeFileAnalysisInterface`.

I'd consider this to be a backwards incompatible API change requiring a new major version (e.g. a third party plugin using that would break).
Instead, it might make more sense to have `BeforeAnalyzeFileInterface` extend from `BeforeFileAnalysisInterface` and be `@deprecated use BeforeFileAnalysisInterface instead in 3.x.y`

weirdan:
@TysonAndre, according to @lhchavez it never worked in the first place.

TysonAndre:
In that case, never mind, no objections.

I guess I'd fixed this at some point but forgot to make a PR to contribute that upstream after validating that bug fixes/merge conflict fixes worked.

PR #3971, opened 2020-08-10
Localize types on magic properties

### LINKED ISSUE #3949: @template + @property = ???
https://psalm.dev/r/50f4749252
here in `Bad` class we have template param `T` which is for sure `instanceof A`. Shouldn't it act like `Straightforward` class?

If `@property` is replaced with regular property, there are no errors: https://psalm.dev/r/dfb5f81c20

Templated properties also seems to have trouble with templates in general:
https://psalm.dev/r/2ec1a86af7

m0003r:
Fix #3949

I've just copied type localization logic from plain properties to magic


muglug:
Thanks!

Issue #3972, opened 2020-08-10
request for documentation


SignpostMarv:
It would be nice if the return types for `null|false` would have descriptions added, so the intent can be determined without digging through the psalm plugin infrastructure code.

mr-feek:
Some documentation around what each parameter in a hook is for would be super helpful. It took me a while to grok what each was. 

I can probably take a stab at a PR

SignpostMarv:
it looks like `return false` says "no further issues to be seen here, skip for the rest of the execution" ?

mr-feek:
I took a stab at providing some context around the hooks I use the most  --  although I think we should leave this ticket open because more effort can be put in

PR #3978, opened 2020-08-10
docs: provide some background as to what certain hooks are for

mr-feek:
Related to  https://github.com/vimeo/psalm/issues/3972

When I first started on plugin development, I was very confused as to why these return types are nullable. I hope my statements here are accurate and might help someone else out in the future


muglug:
Thanks!

Issue #3976, opened 2020-08-10
Question about applicability of allowCoercionFromStringToClassConst and allowStringToStandInForClass


ktomk:
Given I'm using class-names as verbatim strings instead of class::class constants.

When using [`allowCoercionFromStringToClassConst="true"`](https://psalm.dev/docs/running_psalm/configuration/#allowcoercionfromstringtoclassconst)  and [`allowStringToStandInForClass="true"`](https://psalm.dev/docs/running_psalm/configuration/#allowstringtostandinforclass)  I still get errors like:

> ERROR: ArgumentTypeCoercion - 22:18 - Argument 1 of assertInstanceOf expects class-string, parent type string(stdClass) provided

Example code (the options are not available there): https://psalm.dev/r/3e90f55330

Me was under the impression that I should not see the error any longer, when I run this scenario locally I have error-level 3 currently).

Is this error intended?

weirdan:
`allowCoercionFromStringToClassConst` was removed in 3.0, but it seems nobody got around to remove it's documentation. It's been [mentioned](https://github.com/vimeo/psalm/pull/1224), but failed to attract attention.

ktomk:
@weirdan Thanks for the feedback, much appreceated. 

I'm using 3.x so this explains it made no difference.

For part of my code in one project I could benefit from this (if I understand it correctly in the first place). what I need is that whenever @class-string is in use to upgrade/promote (take) any string to (as) a class-string. I don't need to call methods on these, I just need it for making these strings special in terms of class-string parameters (should be parameters only).

This would allow me to use more concrete type (here: class-string) - which comes with a lot of benefits from Psalm templates - with code backwards compatible to PHP before 5.6.

Would this be any of the (previous existing) configurations? Or would this be something new?

muglug:
> with code backwards compatible to PHP before 5.6

PHP 5.5 went EOL 4 years ago, and I don't want to encourage its usage via Psalm.

ktomk:
@muglug Yes, I can perfectly understand that, however the code is also perfectly PHP 7.4 / PHP master compatible. So I don't want to encourage Psalm use with PHP 5.x either, my main question is whether an equivalent benefit would be possible with Psalm for passing strings verbatim only for class-names where that is annotated for Psalm.

[...]

Issue #1040, opened 2018-10-27
Breaking changes for v3


muglug:
- [x] Rename `*Checker` to `*Analyzer`
- [x] Config: change `allowStringToStandInForClass` default to `false`
- [x] Config: remove `allowCoercionFromStringToClassConst`
- [x] Plugin refactor cc @weirdan

weirdan:
@muglug #855 as it stands does not have any breaking changes. `PluginApi` namespace which it introduces is an extension, old plugins would still continue to work with `LegacyPlugin` wrapper.

Remaining todo items are not strictly required, the PR can be merged as is . `PluginApi` interfaces can (and should, as currently you can only add stub files through them) be extended during v3 lifecycle.

There's couple of finishing touches I want to do before it's merged though: namespacing the `extra` key in `package.json` and renaming the `PluginFacade`.

muglug:
Removed `CamelCase vars and properties?` bc actually I think `snake_case` is more legible, and it's backed up by some science.

weirdan:
> bc actually I think snake_case is more legible, and it's backed up by some science.

What about snake_casing the `inferredType` then? Also, can coding standards be documented somewhere, as it deviates from commonly used style in PHP ecosystem (camelCasing everything)?

Sadly, neither php-cs-fixer nor codesniffer seem to have an option for snake_case vars/props.

TysonAndre:
> Sadly, neither php-cs-fixer nor codesniffer seem to have an option for snake_case vars/props.

I have a custom sniffer based on another sniffer which will warn about variables/properties that seem to be camelCase, to enforce this consistency. See https://github.com/phan/phan/blob/master/internal/PHP_CodeSniffer/Sniffs/ValidUnderscoreVariableNameSniff.php and https://github.com/phan/phan/blob/9a1e89cd615ca44ffa4ccb5db69674bd9213217d/phpcs.xml#L56-L59

ktomk:
- 7fd4cab1f7be484ce7d51c7f81993cac20543091 - Changelog (mention: allowCoercionFromStringToClassConst)
- dc834fdb01490d5774a1a1f41889fe59414c7765 - allow_coercion_from_string_to_class_const removal / upgrade away from it

PR #3982, opened 2020-08-10
Removed obsolete documenation

weirdan:
`allowCoercionFromStringToClassConst` was removed in 3.0 and never worked since.

Refs vimeo/psalm#3976

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -32,7 +32,6 @@
         <xs:attribute name="serializer" type="xs:string" />
 
         <xs:attribute name="addParamDefaultToDocblockType" type="xs:boolean" default="false" />
-        <xs:attribute name="allowCoercionFromStringToClassConst" type="xs:boolean" default="false" />
         <xs:attribute name="allowFileIncludes" type="xs:boolean" default="true" />
         <xs:attribute name="allowPhpStormGenerics" type="xs:boolean" default="false" />
         <xs:attribute name="allowStringToStandInForClass" type="xs:boolean" default="false" />
--- a/docs/annotating_code/type_syntax/scalar_types.md
+++ b/docs/annotating_code/type_syntax/scalar_types.md
@@ -30,7 +30,7 @@ class A {}
 function takesClassName(string $s) : void {}
 ```
 
-`takesClassName("A");` would trigger a `TypeCoercion` issue (or a `PossiblyInvalidArgument` issue if [`allowCoercionFromStringToClassConst`](../running_psalm/configuration.md#coding-style) was set to `false` in your config), whereas `takesClassName(A::class)` is fine.
+`takesClassName("A");` would trigger a `TypeCoercion` issue, whereas `takesClassName(A::class)` is fine.
 
 You can also parameterize `class-string` with an object name e.g. [`class-string<Foo>`](value_types.md#regular-class-constants). This tells Psalm that any matching type must either be a class string of `Foo` or one of its descendants.
 
--- a/docs/running_psalm/configuration.md
+++ b/docs/running_psalm/configuration.md
@@ -161,15 +161,6 @@ Setting this to `false` means that any function calls will cause Psalm to forget
 ```
 Allows you to specify whether or not to use the typed iterator docblock format supported by PHP Storm e.g. `ArrayIterator|string[]`, which Psalm transforms to `ArrayIterator<string>`. Defaults to `false`.
 
-#### allowCoercionFromStringToClassConst
-
-```xml
-<psalm
-  allowCoercionFromStringToClassConst="[bool]"
->
-```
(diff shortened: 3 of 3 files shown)


ktomk:
IMHO commit should reference the change-set (and/or issue no) in which this was removed.

Rest looks good from my side, XSD is part of the change-set.

ktomk:
X-Ref #1222, #1224

weirdan:
Couldn't find the specific comming with Github search, but removal was mentioned in #1040.

> maybe leave a message for older versions that this config option existed, but gone since 3.0)

It was almost two years ago, when Psalm was much less popular. I bet most of the users never used Psalm before 3.0.

ktomk:
marking it as removed in the docs would keep the (in-doc) ref with bookmarks / searches, it's more from the a webmistress' perspective than a technical requirement. esp. given that the message has at least one ref already, this is an improvement to the changes back in nov 2018 so no blocker from my perspective (I've left some refs I could dig up in #1040).

Majkl578:
I believe so, the XML configuration is checked against XML schema. While that is good for validation and error prevention it also means changes like this can't happen in minor/patch versions.

ktomk:
Yes, this can break automated builds unexpectedly if there is no version pinning for the tooling (which might not be wanted to integrate faster w/ updated tooling, but then the breakage might indeed come unexpected - depending on how this is wanted).

From what I could check there is no deprecation feature within XSD itself. One suggestion I've found is to mark those fields deprecated in it's own XML namespace and then handle these with the processor / validation.

Ideally the build should not break but give a deprecation notice (e.g. to standard error) to continue for such minor version changes. It is then the users duty to check for warnings in builds which normally is part of the job. Most users do at least when they consider to upgrade to a major version. So this might also help with migrations.

However in this case, the flaw is that this property never made it into the current major version and from that perspective the XSD was flawed and it was actually an error that the file was validated.

For the moment w/o code changes, if you need reliable automated builds, please pin the version for psalm. I can understand that this might not be what you want to do in the long run, but this prevents seeing the build breaking.

So far I like the feature that psalm validates the config against it's schema when opening the document, so would argue to keep it. Then however as there is no concept of deprecation for XML schema, I'm not entirely sure how far it's good to go.

If the validation is not mandatory for psalm to work with the configuration data later on, all validation errors could turned into deprecation warnings. This however would also constitute a b/c break, so could only be an option for the next major release and might be considered a degradation of the utility as the validation is by far not that strict as before and failing early is useful, especially IMHO when it's about configuration.

One idea I have is to take XSD changes into account as a process. There is the XSD file as before which must pass validation as we know it (strict).

Additionally there is a second XSD file that is the same as the original one but constains the minor version changes.

First the strict validation is run and it must pass. Then the updated validation is run and it could give deprecation notices.

Files could be in the same directory to be easily diffed next to each other.

This suggestion has a bit of overhead on the file level, but given the files aren't that large, it's not totally bulky. It would need some documentation though why two files are and for what they are there.

This would keep changes to the overall working low, add an additional XML validation step (which I think is normally fast), could be feature-flagged (e.g. a command-line switch to not give config deprecation warnings) and most of all would allow to improve the validation while keeping backwards compatibility in a practical sense.

This would be my first idea while chewing on the problem. Please let me know your thoughts.

XedinUnknown:
Yeah... We have to fix dozens of packages now...

XedinUnknown:
I would say, the XSD is the authoritative source of what can and what cannot be in the XML, not the documentation, and regardless of whether it is actually used in checks or not. Therefore, it should be considered as such, making this indeed a BC-breaking change, and not a bugfix. Yes, it should never have been added there, but it was, and now removing it is breaking stuff.

ktomk:
@XedinUnknown: The b/c breaking XSD has been fixed in #4046, so skipping the affected version should help.

One suggestion has been made already for having this non b/c breaking is to mark such changes deprecated so that the XSD is non-breaking but the documentation through the/a schema is already prepared for the next version and users can look for schema warnings instead of errors. Would this help in your case?

XedinUnknown:
@ktomk, deprecation instead of outright removal would surely help. But from what I read in this thread, I got the impression that it isn't possible to use XSD to deprecate something.

ktomk:
@XedinUnknown Yes, XSD itself has no mechanism for deprecation built in. The idea is to keep the current XSD and add a second one which is more precise and have those validation errors non-failing. It then can function to give deprecation warnings. That deprecating XSD is technically more strict than the first one, just not failing on schema invalidation. When the next major version is prepared, the more strict "deprecating" XSD is then copied over to the standard one. What was deprecated previously then becomes mandatory with the next major release.

With this suggestion, there is the caveat to author two XSD files and to not forget to apply the copy on next major release.

An alternative approach to mark XSD elements within their own namespace in the same file I've not yet looked into and I can't say if it is feasible, e.g. filter deprecations from errors, esp. without preventing errors.

PR #3996, opened 2020-08-16
Fixed duplicate shortcodes

weirdan:
Also added an utility to help assign new shortcodes and a test to prevent duplicate shortcodes.


muglug:
Thanks!

PR #3997, opened 2020-08-17
Improve import errors

### LINKED ISSUE #3885: @psalm-import-type should complain about malformed imports
https://psalm.dev/r/6da34cd451

Both of those imports are invalid:
1. `Types` is an unknown symbol
2. there's a typo, `form` should be `from`

weirdan:
Fixes #3885

### DIFF
--- a/src/Psalm/Issue/InvalidTypeImport.php
+++ b/src/Psalm/Issue/InvalidTypeImport.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Psalm\Issue;
+
+class InvalidTypeImport extends CodeIssue
+{
+    const ERROR_LEVEL = 7;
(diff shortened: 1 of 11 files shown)


src/Psalm/Issue/InvalidTypeImport.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Psalm\Issue;
+
+class InvalidTypeImport extends CodeIssue
+{
+    const ERROR_LEVEL = 7;
```
weirdan, line 7: Not sure about the level, feel free to adjust.

muglug:
Thanks! Not that I really care, but I think it's good to avoid aliases of functions (e.g. `join` instead of `implode`)

Issue #3975, opened 2020-08-10
query re: conditional function impurity


SignpostMarv:
As near as I can tell, `class_exists()` is only impure if argument 2 is `true`.

Would it be worth resolving this in psalm, or just flagging the function as entirely impure?

muglug:
Nah, it's still impure because you subsequently define a class and then it'll return a different value.

PR #4004, opened 2020-08-17
flag class_exists as impure per vimeo/psalm#3975

SignpostMarv:
`class_exists()` interacts with PHP's autoloader feature, which allows
user-defined behaviour to take place when PHP tries to load a given
class or interface.


muglug:
Thanks!

PR #4014, opened 2020-08-18
Fix hash_*() signatures

baukevdw:
hash_*() functions can return `false` when an invalid algorithm is supplied.

See [hash](https://3v4l.org/Fct6Y) [hash_hmac](https://3v4l.org/Unrs6) [hash_hmac_file](https://3v4l.org/NsHle) [hash_init](https://3v4l.org/uccXc) [hash_pbkdf2](https://3v4l.org/sSOa4)

As of PHP 8 theses functions will throw an exception in case an invalid algorithm is supplied.


muglug:
Thanks!

PR #4021, opened 2020-08-18
Set up taint analysis for pgsql functions

TysonAndre:
I noticed that entries in InternalTaintSinkMap don't warn
unless the functions are added to CallMap.php
e.g. `pg_exec($conn, $_GET['query']);`
(pg_exec is an alias of pg_query)


muglug:
Thanks!

PR #4032, opened 2020-08-20
Add test case for false positive RedundantCondition with boolean operator

MorrisJobke:
Found while introducing Psalm to the Nextcloud code base.

Unfortunately this is only the test case that shows the problem and is missing the actual fix. Feel free to just push to this branch to fix the issue.

cc @rullzer @kesselb


MorrisJobke:
@weirdan @muglug Anything I can help here with?

muglug:
This PR isn't, itself, particularly useful, so I'm closing – it needs to exist as an issue (which I believe it already does).

PR #4046, opened 2020-08-23
Revert config schema, fix b/c break

ktomk:
Previous in d5055ea the allowCoercionFromStringToClassConst attribute has
been removed from the XML configuration file per its schema.

While technically correct (was removed in 3.0), this breaks b/c between
minor versions, breaks with the release of 3.14.0.

Fix is revert.

Ref: d5055ea1d43cb724de2e7e27ae11c6b743a6f727
Caused-by: #3982


muglug:
Thanks!

PR #4051, opened 2020-08-24
Fix prophecy warnings

weirdan:
This fixes annoying warnings about `prophesize()` being deprecated. The dependency added (`weirdan/prophecy-shim`) provides forward-compatible (empty) implementation of ProphecyTrait for older PHPUnit versions that are not supported by `phpspec/prophecy-phpunit`.


muglug:
Thanks!

Issue #4044, opened 2020-08-23
Constant initialized from other constant is not handled


greg0ire:
I'm trying to use [`ParameterType::*`](https://github.com/doctrine/dbal/blob/2.10.x/lib/Doctrine/DBAL/ParameterType.php) as a phpdoc for a method argument, but get error messages showing that only 16 or null is accepted:

```
ERROR: InvalidArgument - tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php:908:42 - Argument 3 of Doctrine\DBAL\Query\QueryBuilder::setParameter expects int(16)|null, int(2) provided (see https://psalm.dev/004)
        $qb->setParameter('name', 'foo', ParameterType::STRING);
```
I dumbed the issue down to the following:

https://psalm.dev/r/6dbe1ad2dd

muglug:
Seems to be fixed

PR #4065, opened 2020-08-26
Resolve unresolved constants in TypeExpander

weirdan:
This fixes vimeo/psalm#4044 by passing down `$statements_analyzer` all the way down to `getConstantForClass()`

### DIFF
--- a/src/Psalm/Internal/Type/TypeExpander.php
+++ b/src/Psalm/Internal/Type/TypeExpander.php
@@ -2,6 +2,7 @@
 namespace Psalm\Internal\Type;
 
 use Psalm\Codebase;
+use Psalm\Internal\Analyzer\StatementsAnalyzer;
 use Psalm\Type;
 use Psalm\Type\Atomic\TTemplateParam;
 use Psalm\Type\Atomic\TNamedObject;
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Type/TypeExpander.php
```diff
@@ -33,7 +34,8 @@ public static function expandUnion(
         ?string $parent_class,
         bool $evaluate_class_constants = true,
         bool $evaluate_conditional_types = false,
-        bool $final = false
+        bool $final = false,
+        ?StatementsAnalyzer $statements_analyzer = null
```
weirdan, line 38: Not sure if it's the best way to do it, but `expandUnion()` is just too handy.

muglug:
I'm not sure about this particular one, mainly because I think only globally-defined constants should be referenced in a class constant.

This means constants defined in `autoload_files`, as well as the `autoload` config flag files are acceptable, but no others.

weirdan:
Any reason why other constants won't do?

muglug:
Any constants that are assumed to be available to _every_ class are fine IMO. It's the ones that are only available to certain classes (and thus need the `$statements_analyzer`) that trouble me.

weirdan:
@muglug are you referring to visibility (as in `private const WHATEVER=213`) or the load order (i.e. absence of const/function autoloading in PHP)?

muglug:
the load order – some constants are always assumed to be present if they're in `autoload_files` or defined in the `autoloader="foo.php"` file in the config

orklah:
@weirdan seems like the original ticket was fixed. Is this still needed?

weirdan:
@orklah I don't think the original issue is fixed, snippets there still fail in various ways. This may not be the best way to fix that though.

PR #4066, opened 2020-08-26
add pure-callable type

### LINKED ISSUE #4062: add `pure-callable` type and ban using callable arguments in pure context.
similar to #4052, the `callable` type doesn't guarantee that the function will be pure. in this case, I think it would be better to add a `pure-callable` type so that pure functions can still take a callable as an argument, and psalm can verify that it is pure.

pure-callable is just a normal callable but declared using the `@psalm-pure` annotation.

see: https://psalm.dev/r/50f8a9d0c2

impure behavior: https://3v4l.org/jQ4IT

azjezz:
closes #4062 

TODO:
- [x] add pure-callable type
- [x] type check pure-callable
- [x] handle [$obj, 'method'], ['foo', 'staticmethod'], and invokable objects.
- [x] ban `callable` type usage in pure context.
- [x] tests

### DIFF
--- a/src/Psalm/Internal/Diff/AstDiffer.php
+++ b/src/Psalm/Internal/Diff/AstDiffer.php
@@ -23,8 +23,6 @@ class AstDiffer
      * @param array<int, PhpParser\Node\Stmt> $b
      *
      * @return array{0:array<int, array<int, int>>, 1: int, 2: int, 3: array<int, bool>}
-     *
-     * @psalm-pure
      */
     protected static function calculateTrace(
         \Closure $is_equal,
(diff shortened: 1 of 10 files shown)


src/Psalm/Internal/Diff/AstDiffer.php
```diff
@@ -23,8 +23,6 @@ class AstDiffer
      * @param array<int, PhpParser\Node\Stmt> $b
      *
      * @return array{0:array<int, array<int, int>>, 1: int, 2: int, 3: array<int, bool>}
-     *
-     * @psalm-pure
```
azjezz, line 27: `$is_equal` is not pure 😄

azjezz: ( caught by this PR 😃 https://app.circleci.com/pipelines/github/vimeo/psalm/3408/workflows/155fa96e-c151-4f53-9f7d-6f89886902bd/jobs/15663 )

azjezz:
ping @muglug :)

muglug:
Thanks!

PR #4084, opened 2020-08-29
Fix expected php 8.0 syntax error

TysonAndre:
This will very likely be parsed as an attribute token in php 8.0
See
https://wiki.php.net/rfc/shorter_attribute_syntax_change#secondary_vote


muglug:
haha thanks

Issue #3271, opened 2020-04-30
Links to source code


enumag:
When solving PHPStan issues it's very useful for me to have the ability to open the source code directly from the console output with a click. I achieved that using this [Formatter](https://github.com/grifart/phpstan-oneline/blob/master/src/CompactErrorFormatter.php) with the following configuration:

```
	errorFormatter.compact:
		class: Grifart\PhpstanOneLine\CompactErrorFormatter
		arguments:
			format: "]8;;phpstorm://open/?file={absolutePath}&line={line}\\{path}:{line}]8;;\\\n ↳ {error}\n"
```

Is there any way I could configure psalm similarly?

enumag:
ping @muglug

andrew-demb:
I think we can put logic to report formatters (they inherits from https://github.com/vimeo/psalm/blob/master/src/Psalm/Report.php).

For example - console formatter - https://github.com/vimeo/psalm/blob/d8cd30b34a438e5e980533922fc18bea49bc7402/src/Psalm/Report/ConsoleReport.php#L37-L39

enumag:
Can I configure Psalm to use my custom Report implementation?

andrew-demb:
AFAIK, no you can't.

But we can introduce new option to current report formatter to show file links in IDE-compatible way.

Where formatters instantiated:
https://github.com/vimeo/psalm/blob/4bfdda597efc724e44d89a75dc71adc849140c55/src/Psalm/IssueBuffer.php#L631-L704

enumag:
@andrew-demb One issue is that the link needs to contain an absolute path to the file. `IssueData::$file_name` is relative. How can I get the absolute path in `ConsoleReport`?

andrew-demb:
I don't know codebase so much, but I think we can add argument in `getOutput()` - `string $base_dir` and put to it value from config - `$project_analyzer->getConfig()->base_dir`

ostrolucky:
Here is how to setup iterm2 to do this, although limitation here is that because of hardcoded path, this is viable for single project only:

Preferences -> Profiles - {profileName} -> Advanced -> Triggers (Edit) -> Add:
Regular Expression: `ERROR: \w+ - ([a-zA-Z\/.\d]+):(\d+)`
Action: Make Hyprelink
Parameters: `phpstorm://open?file=/Users/gabriel.ostrolucky/Documents/enrola/project/\1&line=\2`

Done

[...]

PR #4085, opened 2020-08-29
Add option to show link to source in console output

the-toster:
It adds `--php-storm` option to `psalm` command, which make console output fit to phpStorm pattern of source links (`\n at {file}:{line}:{col}\n`)

I see two problems in this PR: 1) option name is not very good, and 2) this pattern is not documented (but stacktrace printed like this).

For me this option is very handy (also, it was requested here #3271 ), I'll be glad of any suggestions, how this can be done.


andyg0808:
I don't really know anything about implementation details, but there are a bunch of existing report formats that are switchable via the `--output-format` option:
```sh
    --output-format=console
        Changes the output format.
        Available formats: compact, console, emacs, json, pylint, xml, checkstyle, junit, sonarqube, github
```
I would suspect that it might make sense to make this available via that list as well.
It looks to me like the implementations may be in https://github.com/vimeo/psalm/tree/master/src/Psalm/Report.

the-toster:
I have renamed flag to `--php-storm-format`

the-toster:
@andyg0808 At first I did as you say, but looking in current codebase I have found four `if` statements, that check type of report, this:
https://github.com/vimeo/psalm/blob/64f08fa9dca23e81862e6f2f5b73ea391ecd10e5/src/Psalm/IssueBuffer.php#L456-L461
(and they should be modified to support one more console-like format)
So, I decide do not add extra complextity to it.

the-toster:
If new report format is prefered way, I'm ready to implement it.

muglug:
Yeah, I'd favour a new report output format for this, as it'll provide a little more versatility to customise things the way PHPStorm likes

the-toster:
@muglug It's ready, please take a look.
New format name is `phpstorm`, and help message changed to this (according to code style limitation of line length):
```shell
...

    --output-format=console
        Changes the output format.
        Available formats: compact, console, text, emacs, json, pylint, xml, checkstyle, junit, sonarqube, github,
                           phpstorm

...
```

Also I used `in_array` in `if` statements:
```php
        if ($this->stdout_report_options
            && in_array(
                $this->stdout_report_options->format,
                [\Psalm\Report::TYPE_CONSOLE, \Psalm\Report::TYPE_PHP_STORM]
            )
            && $this->codebase->collect_references
        ) {
```
 I dont know is it acceptable.

the-toster:
Maybe better to add `\Psalm\Report::isConsoleLike($format)` method, instead of this repeatable `in_array`..

muglug:
Thanks!

Issue #4049, opened 2020-08-24
Allow changing default severity


orklah:
Hi,

Today, at work, I wanted to create a new xml file to show me the highest errors (which are ignored in my baseline in my default xml config).

I created a new standard file with errorLevel="8" and I was surprised to see the analysis take 13 times more than my usual analysis (it went from 354 seconds to 4820 seconds)

I narrowed down the issue to the structure of the xml file. My usual file contains every issue in issueHandler with "suppress" for each issue I don't want to see whereas the new file was only `errorLevel="8"`. I tried to add every issue from level < 8 with "suppress" and it was suddenly faster again.

That means, by default, Psalm will take it's time just to display the "XXXX issues found" (because --show-info is now false by default).

My suggestion would be to add a configuration `defaultSeverity` which could be changed to "suppress" to replace the hard coded "info" value here: https://github.com/vimeo/psalm/blob/46013a35c84a0d547e0167b87e3a7da4f045cd8d/src/Psalm/Config.php#L1600 . This config should be ignored wherever --show-info was set to true.

Additionally, I would try to define a threshold which would automatically set this config to "error" in case the number of errors in lower levels is too high when running --init.

Of course, this would mean we won't be able to count issues for the report at the end. A placeholder "Remove defaultSeverity config to see the number of errors" may be added to replace the missing report line.

Would you accept such PR? Do you see improvements or obstacles?

weirdan:
This is likely fixed in #4095

orklah:
Yep, it was, thanks @weirdan

PR #4087, opened 2020-08-30
Allow changing config for severities

orklah:
I posted on #4049 that using errorLevel will automatically report lower errors as "info", which can be a massive slowdown for big projects.

With this PR, I added a "reportLowerLevelsAsInfo" config, which, if false, will suppress those errors

I tried to create a psalm.xml with errorLevel="8" for my work codebase. With reportLowerLevelsAsInfo to true, it took 8156 seconds. With reportLowerLevelsAsInfo to false, it took 570 seconds

I also added a defaultSeverity config which allow setting the severity by default. For example, if I create a xml file to check only one issue, I can put suppress by default and then use issueHandler more easily.

If the PR is validated, I'll create documentation for those two configs.


muglug:
Mind adding documentation with this PR? I don't object to anything that speeds up Psalm.

I think though a flag like `reportInfo="false"` might be simpler?

orklah:
I messed up something. I'll open a new PR

PR #4091, opened 2020-08-31
Include variadic and pass-by-ref params in generated stubs

joehoyle:
When generating stubs, params that are passed by ref or variadic don't get added to the generated code stub output.


muglug:
Thanks!

PR #4100, opened 2020-09-01
Add `null` as a possible type parameter to `strval`

lhchavez:
https://www.php.net/manual/en/language.types.string.php#language.types.string.casting mentions

> `NULL` is always converted to an empty string.

Which seems to indicate that it is a valid argument for it.

As opposed to stringifying other types (like arrays), `strval(null)` does not cause a warning in PHP.


muglug:
Sure, thanks!

Issue #4070, opened 2020-08-26
`strpos` with a constant first arg


staabm:
similar to https://github.com/vimeo/psalm/issues/3877 I guess psalm should emit a error, when someone uses `strpos` with a string literal as a first argument.

I can't think of a good use-case where I want the $haystack to be a string literal.

in contrast when the $haystack and $needle params are mixed up, it can lead to bugs, see e.g.
https://github.com/joomla/joomla-cms/pull/30489
https://github.com/composer/composer/pull/9154
https://github.com/justinrainbow/json-schema/pull/641

https://psalm.dev/r/22d89c087e -> the `strpos()` call should emit an error

staabm:
additionally to my above request, I would expect psalm to error, because strpos() can never return `-1`

weirdan:
@psalm-github-bot, please fetch this: https://psalm.dev/r/22d89c087e

muglug:
Fixed in 422271b

staabm:
Thanks!

staabm:
Hmm wondering why I dont get a error on https://psalm.dev/r/e0c5b3e258 after you fixed this issue?

staabm:
@muglug could you re-open the issue, as it seems not to be fixed atm

weirdan:
> I can't think of a good use-case where I want the $haystack to be a string literal.

This does cause false positives in the code that uses strings as lookup tables (we have quite a few of those for generating barcodes). Can we make the issue more specific, so that it could be selectively suppressed?

muglug:
> Can we make the issue more specific, so that it could be selectively suppressed?

Sure, I'll do that

staabm:
> This does cause false positives in the code that uses strings as lookup tables (we have quite a few of those for generating barcodes)

@weirdan Im curios.. could younshow a example for what you are talking about?

staabm:
@muglug its working now, thx for fixing

One question!

https://psalm.dev/r/22d89c087e

Outputs

> ERROR: DocblockTypeContradiction - 8:7 - int can never contain false|int|int

Shouldn‘t it be

> ERROR: DocblockTypeContradiction - 8:7 - int can never contain positive-int|0|false

?

weirdan:
> Im curios.. could younshow a example for what you are talking about?

Sure. E.g. [Codabar](https://en.wikipedia.org/wiki/Codabar) encoding may look like this: https://psalm.dev/r/ffe1e64c0f

[...]

PR #4101, opened 2020-09-01
Added more fine grained strpos stub

staabm:
A shot in the dark.. Might fix the comment https://github.com/vimeo/psalm/issues/4070#issuecomment-680942551

> I would expect psalm to error, because strpos() can never return -1 (this can be a common mistake, because people might be used to javascript in which the equivalent function returns -1)

https://psalm.dev/r/22d89c087e

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
+++ b/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
@@ -486,7 +486,7 @@ function strtoupper(string $str) : string {}
  *
  * @psalm-return positive-int|0|false
  */
-function strpos(string $haystack, string $needle, int $offset = 0) : int {}
+function strpos(string $haystack, $needle, int $offset = 0) : int {}
 
 /**
  * @psalm-pure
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
```diff
@@ -481,6 +481,13 @@ function strtolower(string $str) : string {}
  */
 function strtoupper(string $str) : string {}
 
+/**
+ * @psalm-pure
+ *
+ * @psalm-return positive-int|0|false
```
staabm, line 487: actual point of the pr is to catch callers which falsly assume the return could be negative

muglug:
Looks great! I think it might also be worth creating a `non-literal-string` type that disallows explicit strings.

staabm:
Thx for merging.

If you tell me where to add it, I could try covering the change by a test

staabm:
> Looks great! I think it might also be worth creating a `non-literal-string` type that disallows explicit strings.

Sounds great. Contributing such a change is outside my psalm fu, though :)

staabm:
I am wondering whether this change also requires a change in https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/CallMap.php ?

I am wondering, because the error message we face now in https://psalm.dev/r/22d89c087e looks strange

PR #4102, opened 2020-09-01
Short list syntax

orklah:
This PR propose to replace the use of list() with the short list syntax []. This is mainly motivated by the fact that the use of PhpStorm refactoring automatically changes those and this pollutes every PR I want to push.

### DIFF
--- a/src/Psalm/Internal/Diff/ClassStatementsDiffer.php
+++ b/src/Psalm/Internal/Diff/ClassStatementsDiffer.php
@@ -34,7 +34,7 @@ class ClassStatementsDiffer extends AstDiffer
     {
         $diff_map = [];
 
-        list($trace, $x, $y, $bc) = self::calculateTrace(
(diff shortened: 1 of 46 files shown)


src/Psalm/Internal/Diff/ClassStatementsDiffer.php
```diff
@@ -109,13 +109,13 @@ function (
                 }
 
                 if ($a instanceof PhpParser\Node\Stmt\ClassMethod && $b instanceof PhpParser\Node\Stmt\ClassMethod) {
-                    if ((string) $a->name !== (string) $b->name) {
+                    if ((string)$a->name !== (string)$b->name) {
```
weirdan, line 112: These whitespace changes seem unrelated.

orklah: You're right, they must have slipped through. I reverted them

muglug:
Thanks!

PR #4143, opened 2020-09-06
Fixes to LSP protocol

matthijskooijman:
This PR makes two changes:
 1. Add a missing parameter documentation in `textDocument.didChange`, that caused the JSON-RPC dispatcher to generate an error.
 2. Prevent `textDocument.completion` from operating on out-of-project files, to be consistent with other methods.

Wrt out-of-project files, this commit follows the lead of other methods to simply return an empty response for such out-of-project files and otherwise ignore them. I do wonder if it would not be better to return an error response (e.g. raise an exception) in all of these methods instead, or maybe only in `didOpen`, to let the LSP client and possibly the user know that no analysis is actually happening?


muglug:
Thanks!

Issue #4141, opened 2020-09-06
Iterator::current() assumed not nullable


gmazzap:
It seems that Psalm assumes `Iterator::current()` to always return the type assigned to the iterator via template annotations, but the reality is that `Iterator::current()` is always nullable because it returns null when the iterator is empty or consumed.

Consider the following code:

```php
/**
 * @return Iterator<string>
 */
function buildIterator(int $size): Iterator {
 
    $strings = [];
    for ($i = 0;  $i < $size; $i++) {
       $strings[] = random_bytes(1); 
    }
 
    return new ArrayIterator($strings);
}

$current = buildIterator(0)->current();
if (is_string($current)) {
    echo "First item is $current";
}
```

(https://psalm.dev/r/7db06f8ddd)

The condition `is_string($current)` is valid, because when the iterator is empty (and that's the case here) `$current` will be null (see https://3v4l.org/2job5).

Another example: https://psalm.dev/r/324b008117 this time false negative. The code will throw an exception if the iterator is empty (see https://3v4l.org/kDMSE) but for Psalm everything is fine.

---

In short, **for an iterator of type `Iterator<T>`, the method `Iterator::current()` has a return value of `T|null` and not `T`**.

Unless Psalm is able to infer that the iterator is not empty, e.g. in the following code:

```php
$iterator= buildIterator(2);
if ($iterator->valid()) {
   // In this case the condition is redundant because iterator is surely not empty at this point
   // so $current is surely a string according to the doc bloc
   if (is_string($current)) {
   }
}
```

Please note that this is not only valid for empty iterators, but also for non-empty iterators that have been consumed. For example, see here: https://3v4l.org/tmvX3 how for the same iterator instance `Iterator::current()` is either `string` or `null` depending on the iterator being valid or not.

muglug:
Mind creating a PR on `CoreGenericClasses.phpstub`?

gmazzap:
As you can see I've created #4146 that fixes stubs.

But I think this will disrupt a lot of codebases.

IMO it would be better to make use of `@psalm-assert-if-true` / `@psalm-assert-if-false` on the `Iterator::valid()` to assert that the `Iterator::current()` will return `TValue` when `Iterator::valid()` returns `true` and will return `null`  when when `Iterator::valid()` returns `false`.

To introduce this assertion I think it is required to add a new template type `TCurrent` of type `TValue|null` so that `current()` could return `TCurrent`. But I'm not sure this is doable...

[...]

PR #4146, opened 2020-09-07
Fix Iterator::current in stubs (#4141)

gmazzap:
See #4141 

`Iterator::current()` will return `null` if the iterator is empty.

This PR fixes Psalm internal stubs for:
* Iterator
* IteratorIterator
* Generator
* SplDoublyLinkedList
* SplObjectStorage


muglug:
Mind also fixing the one test?

muglug:
Actually I'm going to close this in favour of a slightly different fix – Psalm uses the implicit return type of `current()` when typing iterators, so this should only take effect when called explicitly.

PR #4149, opened 2020-09-07
Document cacheDirectory behaviour

alfredbez:
See https://github.com/vimeo/psalm/blob/8c7423505a4282d1cdd38f3ff9552b911cbd5fa9/src/Psalm/Config.php#L869-L875

And https://github.com/dnoegel/php-xdg-base-dir/blob/12f5b94710c8f5b504432d57ce353075fc434339/src/Xdg.php#L78-L84


muglug:
Thanks!

PR #4164, opened 2020-09-10
Document Xdebug usage

adrienlucas:
Just a little hint for anyone wondering why their `xdebug.remote_enable=1` is not working when debugging their plugin.

### DIFF
--- a/docs/running_psalm/plugins/authoring_plugins.md
+++ b/docs/running_psalm/plugins/authoring_plugins.md
@@ -105,10 +105,10 @@ You can also specify an absolute path to your plugin:
 
 ### Using Xdebug
 
-As Psalm disables _Xdebug_ at runtime, if you need to debug your code step-by-step when authoring a plugin, you can allow the extension by doing :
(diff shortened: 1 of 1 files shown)


docs/running_psalm/plugins/authoring_plugins.md
```diff
@@ -103,6 +103,14 @@ You can also specify an absolute path to your plugin:
     </plugins>
 ```
 
+### Using Xdebug
+
+As Psalm disables _Xdebug_ at runtime, if you need to debug your code step-by-step when authoring a plugin, you can allow the extension by doing :
+
+```bash
+$ export PSALM_ALLOW_XDEBUG=1
```
weirdan, line 111: I think it's better to suggest a non-persistent override for a single command only. This way you wouldn't accidentally forget to reset the env variable when you no longer need the override. 
```suggestion
As Psalm disables _Xdebug_ at runtime, if you need to debug your code step-by-step when authoring a plugin, you can allow the extension by running Psalm as following:

```console
$ PSALM_ALLOW_XDEBUG=1 path/to/psalm
```

weirdan (changes requested):

muglug:
Thanks!

PR #4171, opened 2020-09-11
Docs: Add missing issues to Error levels

caugner:
I noticed that the [Error levels](https://psalm.dev/docs/running_psalm/error_levels/) documentation is missing some issues, so I wrote [an ugly script](https://gist.github.com/caugner/7ffa504f61991256f1f5357529a8a2a2) that lists those, and added them manually.


muglug:
Thanks!

PR #4137, opened 2020-09-05
add native param types

orklah:
Following #4116, here's a PR proposing adding native types in parameters.

As before, I used phpstorm to add them and psalm to refine.

PR #4173, opened 2020-09-12
More return types

orklah:
Hi,

This PR propose adding yet more native return types in Psalm. This time, I mainly added them through Psalter and refine with Psalm and PHPStorm.

There are still many cases where return types could theoretically be added but there are some code constructs that gets in the way for that (for example `return;` is not a valid return for a method that should return null), so I'll push a future PR that addresses that.

Before beginning a review to propose cases I missed, be aware that I had to rollback a lot of return types because there was incompatibility between hierarchy or with some issue Psalm detected. I'm totally open to try to put some back, but this may be wasted time

### DIFF
--- a/examples/TemplateChecker.php
+++ b/examples/TemplateChecker.php
@@ -72,9 +72,6 @@ class TemplateAnalyzer extends Psalm\Internal\Analyzer\FileAnalyzer
     }
 
     /**
-     * @param  \Psalm\Internal\MethodIdentifier         $method_id
-     * @param  PhpParser\Node $stmt
-     *
      * @return Context|false
      */
     private function checkMethod(\Psalm\Internal\MethodIdentifier $method_id, PhpParser\Node $stmt, Codebase $codebase)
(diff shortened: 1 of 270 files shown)


lexidor:
This PR is a little bit of a large to review all at once. I believe there is a lot of value here and I'd love to give you my approval. Could you maybe please split this PR into multiple PRs? For example, upgrading all `@return void` to `) : void` is already a big PR's worth of changes.

You're free to ignore this advice. I am not strongly related to the psalm open source project. Just your friendly neighborhood type system enthusiast.

orklah:
Hi @lexidor, thanks for your input!

I don't quite have the control I'd need for that with the tools that I'm using however. It would take a massive amount of time just to make that division manually.

I already did my best to separate return types and param types in previous PRs(#4116 , #4137) and in them, I tried to separate the commits with code modifications and those with comments modifications to help review.

But I'll try to divide even more if needed :)

muglug:
Thanks!

Issue #4172, opened 2020-09-11
Language server not available on psalm.phar v3.15


ngyuki:
hi!

Running Language server with psalm.phar v3.15 results in an error.

```sh
curl -LO https://github.com/vimeo/psalm/releases/download/3.15/psalm.phar
php psalm.phar --init
php psalm.phar --language-server --tcp-server --verbose
```

```
Uncaught Error: Class 'Psalm\Internal\Analyzer\LanguageServer' not found in phar:///c/Users/ngyuki/example-psalm/psalm.phar/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php:384
Stack trace:
#0 phar:///c/Users/ngyuki/example-psalm/psalm.phar/src/psalm-language-server.php(205): Psalm\Internal\Analyzer\ProjectAnalyzer->server()
#1 phar:///c/Users/ngyuki/example-psalm/psalm.phar/src/psalm.php(131): require_once('phar:///c/Users...')
#2 phar:///c/Users/ngyuki/example-psalm/psalm.phar/psalm(5): require_once('phar:///c/Users...')
#3 /c/Users/ngyuki/example-psalm/psalm.phar(14): require('phar:///c/Users...')
#4 {main}
(Psalm 3.15@de6e7f324f44dde540ebe7ebd4eb481b97c86f30 crashed due to an uncaught Throwable)
```

ngyuki:
I extracted from psalm.phar.
Class name is not resolved correctly.

```sh
phar extract -f psalm.phar -i ProjectAnalyzer psalm/
cat psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php | sed -n 384p
```

```php
new \Psalm\Internal\Analyzer\LanguageServer(new \Psalm\Internal\Analyzer\ProtocolStreamReader(\STDIN), new \Psalm\Internal\Analyzer\ProtocolStreamW
riter(\STDOUT), $this);
```

This may be a bug in the PhpScoper?

PR #4174, opened 2020-09-12
Workaround to PhpScoper group use bug in Box

ngyuki:
See #4172 https://github.com/box-project/box/issues/476


muglug:
Thank you!

PR #4181, opened 2020-09-14
[help wanted] [merge ready?] fix #4180 adding default case impossible check in SwitchCaseAnalyzer.…

### LINKED ISSUE #4180: ParadoxicalCondition not detected in default with return
https://psalm.dev/r/0fa3a37e03
https://psalm.dev/r/6aa20eea86

ygottschalk:
…php to 'return_throw'-ing case statements

Edit: solves issue #4180
Edit 2: I'm not sure what the correct behaviour should be, discussion is in issue #4180 (currently implemented behaviour is described in ~~https://github.com/vimeo/psalm/issues/4180#issuecomment-691981037~~)
Edit3: ... described in https://github.com/vimeo/psalm/pull/4181/commits/990b1046dc9a22b2bc68961debdbfbf5a54ee7bc#diff-0e8f19d6674acee7a312b68efbf6bb2893e4189cd9a91df913da6f2bc9d0cfc1R361-R368

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -31,6 +31,7 @@
         <xs:attribute name="name" type="xs:string" />
         <xs:attribute name="phpVersion" type="xs:string" />
         <xs:attribute name="serializer" type="xs:string" />
+        <xs:attribute name="allowParadoxicalDefaultForExitTypes" type="xs:string" default="throw_or_exit" />
 
         <xs:attribute name="addParamDefaultToDocblockType" type="xs:boolean" default="false" />
         <xs:attribute name="allowCoercionFromStringToClassConst" type="xs:boolean" default="false" />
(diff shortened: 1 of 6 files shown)


ygottschalk:
I added a config option to control psalms behaviour.
I noticed that `Psalm\Internal\Analyzer\ScopeAnalyzer::getFinalControlActions` does not classify a function defined like
```
/** @return no-return */
 function foo() { exit(); }
```
as an `exit statement`, which results in a test failiure (run phpunit with `./tests/SwitchTypeTest.php --filter "testValidCode@impossibleCaseDefaultWithCustomDefinedExitFunction"`) which should pass.

ygottschalk:
@muglug any feedback? 
Shall I try to fix the issue described in my last comment?
Shall I give a reasoning why this feature is useful?

orklah:
Sorry for not coming back to you earlier on this. I was not sure how to answer.

I don't think adding a config on such an obscure thing is the best solution. For all I know of Psalm, I really struggle to see what exactly its behaviour is currently and what it should be, so I'm not convinced other users will either.

However, the behaviour difference in your two snippets on the original issue trouble me and I'd like for those two to behave the same.
This is even more important now that we have the match structure that should behave like that also (emitting an error when all cases were already matched).

would you be willing to start a new PR without introducing a config for now that brings consistency in all those?

I'll close this one in the meantime, it's too old to be of use for now

PR #4189, opened 2020-09-15
Fix errors in return types

orklah:
While adding return types in Psalm, I found some cases where declared return types didn't match with the code.

This PR contains 3 commits, each one fix a specific issue and it's consequences if any.


muglug:
Thanks!

Issue #4196, opened 2020-09-16
random_int and positive-int


DanielBadura:
https://psalm.dev/r/e074fe9e3a

I dont know if psalm should check if random_int could only return a positive or negative integer. What do you think?

muglug:
Hey!

Would you mind creating a PR against `CoreGenericFunctions.phpstub`?

You'll want to add a return type with a [conditional return type](https://psalm.dev/articles/conditional-love), something like

```php
/**
 * @return ($min is positive-int ? ($max is positive-int ? positive-int : int) : int)
 */
function random_int(int $min int $max) : int {}
```

DanielBadura:
Sure thing  👍

PR #4199, opened 2020-09-17
add random_int as stub, returns positive-int  if min is positive-int

DanielBadura:
fix https://github.com/vimeo/psalm/issues/4196


muglug (approved):

muglug:
Thanks!

PR #4204, opened 2020-09-17
Improve PhpStormReport with source code links

enumag:
#3271

### DIFF
--- a/src/Psalm/Report/PhpStormReport.php
+++ b/src/Psalm/Report/PhpStormReport.php
@@ -36,7 +36,8 @@ class PhpStormReport extends Report
         $issue_reference = ' (see ' . $issue_data->link . ')';
 
         $issue_string .= ': ' . $issue_data->type
-            . "\nat " . $issue_data->file_path . ':' . $issue_data->line_from . ':' . $issue_data->column_from
+            . "\nat \e]8;;phpstorm://open/?file=" . $issue_data->file_path . '&line=' . $issue_data->line_from
+            . "\e\\" . $issue_data->file_path . ':' . $issue_data->line_from . ':' . $issue_data->column_from . "\e]8;;\e\\"
             . "\n" . $issue_data->message . $issue_reference . "\n";
 
 


Rarst:
I am curious what does the escape codes do here? Google doesn't tell...

I applied the patch and it still just outputs the absolute path for me (on Windows). I am not sure if link is supposed to be somehow masked or?..

enumag:
> I am curious what does the escape codes do here? Google doesn't tell...

I don't really know the details. I just found it somewhere before and have been using it to get terminal links for PHPStan ever since. Now I just copied it from my PHPStan solution.

> I applied the patch and it still just outputs the absolute path for me (on Windows). I am not sure if link is supposed to be somehow masked or?..

It's basically similar to `<a href="path:line">text</a>` in HTML.

```
// this is `<a href="path:line">`
\e]8;;phpstorm://open/?file=" . $issue_data->file_path . '&line=' . $issue_data->line_from . "\e\\"
// this is `text`
. $issue_data->file_path . ':' . $issue_data->line_from . ':' . $issue_data->column_from
// this is `</a>`
. "\e]8;;\e\\"
```

If you hover the text in PhpStorm terminal while holding Ctrl it should underline. Then by clicking it should open.

Then you ofc need something to handle the `phpstorm://` protocol and call PHPStorm. When I last set that up it wasn't built-in PHPStorm but it might be now - I'm not sure.

Rarst:
So for me (on Windows) the result of this patch is same as unpatched behavior - full path is output and linked up by PhpStorm terminal (no ctrl necessary). The URL seems to be ignored completely.

I mean, it still works, but this seems to be thoroughly confusing issue, so interested what's the most clear and compatible way to handle this for everyone. :)

weirdan:
> I am curious what does the escape codes do here? Google doesn't tell...

@Rarst See https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

enumag:
@Rarst Currently I'm using this: https://github.com/sanduhrs/phpstorm-url-handler

This repository provides instructions for mac and win users: https://github.com/aik099/PhpStormProtocol

I believe you also need to have `phpstorm` command in PATH (at least in Ubuntu). Jetbrains Toolbox is able to generate that:

Screenshot from 2020-09-18 09-07-29

Rarst:
> See https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

Ah, so I am observing correct fallback behavior on Windows, link omitted. If this works well on non-Win platforms then this might be the way to go. Weird that path-based linkifying doesn't work consistently.

> Currently I'm using this: https://github.com/sanduhrs/phpstorm-url-handler
> This repository provides instructions for mac and win users: https://github.com/aik099/PhpStormProtocol

I think I remember PhpStorm got `phpstorm://` _natively_ on OSX (other nixes?) at some point, but not on Windows (which is probably why it doesn't recognize it on Windows either).

I am using a different solution for Win, the [Remote Call](https://plugins.jetbrains.com/plugin/6027-remote-call) plugin, which works through `http://localhost` w/o need for custom protocol or any system-level configuration.

enumag:
Since it doesn't work out of the box it might be better to not put this to psalm directly yet. Is there a way for me to register my own `Psalm\Report` implementation and have Psalm use it?

Rarst:
Given erratic support for related features in PhpStorm, why not output both?
```
ERROR
at /absolute/path/to/
link phpstorm://....
```
Just plain link, no escape codes. Whatever the platform, one or the other would work? Just click the one that is clickable.

enumag:
yeah I think it would be okay to output both... even when the OS can't handle the link it shouldn't cause any issues afaik

Rarst:
Ok, let's enumerate the cases:

syntax | Windows | Linux | OSX
-|-|-|-
at [absolute path] | ✔ | ❓ | ❓
at [relative path] | ❌ | ❓ | ❓
bare `phpstorm://...` | ❌(no native protocol) | ❓ | ❓
escape codes `phpstorm://...` | ❌ (no terminal support) \| ✔(fallback to anchor) | ❓ | ❓

Can someone please test and report these for other platforms?

Then we'll have a better understanding what would be preferable for all around compat.

enumag:
Linux: no (I think there was some plugin?), no, no, yes (with sanduhrs/phpstorm-url-handler)

Rarst:
Cheers, added!

Strange that `at ...` doesn't work on Linux, even though it's native PhpStorm feature. 🤔

PR #4211, opened 2020-09-19
Remove empty() and use strict comparison when safe

orklah:
Hi,

This PR propose changing some syntax into more strict versions when possible.


orklah:
I added some replacement of is_null() to === null too.

orklah:
I think CI failures are unrelated. I still reproduce them when I revert the last commit locally and they weren't there before the last commit

muglug:
Thanks!

The errors were due to a change in latest nikic/php-parser version, since patched

PR #4212, opened 2020-09-19
respect static keyword in method access

orklah:
Hi,

Some static methods were called through `->` and some dynamic method through `::`. This PR propose to replace those with the normal syntax for each.


weirdan (commented):
Psalm ought to have caught static calls on non-static methods. Were those issues suppressed?

orklah:
> Psalm ought to have caught static calls on non-static methods. Were those issues suppressed?

I don't think so, I don't see any reason they'd not been caught

muglug:
```php
<?php

class A {
    public function foo() : void {
        echo "called\n";
    }
    
    public function bar() {
        self::foo();
        A::foo();
    }
}

(new A)->bar();
```

This is perfectly permissible PHP, so I don't treat it as a bug.

muglug:
I've fixed some of this in the opposite direction in 1ac527b

PR #4219, opened 2020-09-20
add visibilities to constants

orklah:
this PR propose adding visibilities to constants.

I tried to make them private when possible, the rest are public.


muglug:
Thanks!

PR #4221, opened 2020-09-20
Use array destructuring when possible

orklah:
This PR propose replacing multiple assignments of array elements into the array destructuring syntax

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
@@ -537,8 +537,7 @@ class ForeachAnalyzer
                         continue;
                     }
 
-                    $value_type_part = $iat->type_params[1];
-                    $key_type_part = $iat->type_params[0];
+                    [$value_type_part, $key_type_part] = $iat->type_params;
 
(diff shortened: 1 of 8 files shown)


src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
```diff
@@ -537,8 +537,7 @@ public static function checkIteratorType(
                         continue;
                     }
 
-                    $value_type_part = $iat->type_params[1];
-                    $key_type_part = $iat->type_params[0];
+                    [$value_type_part, $key_type_part] = $iat->type_params;
```
weirdan, line 540:
```suggestion
                    [$key_type_part, $value_type_part] = $iat->type_params;
```
Vars were in the reverse order in the original code.

weirdan (changes requested):

src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
orklah: well spotted! I was still trying to see what could have caused the issue in CI! Thanks!

weirdan (approved):

PR #4222, opened 2020-09-20
remove unused code

orklah:
There was a whole block of seemingly unused code. Not sure if that's a code that was broken since or if the functionnality never worked.

The block fills a $new_file_paths variable but never use it anywhere after that.

### DIFF
--- a/src/Psalm/Internal/Codebase/Analyzer.php
+++ b/src/Psalm/Internal/Codebase/Analyzer.php
@@ -365,28 +365,7 @@ class Analyzer
             };
 
         if ($pool_size > 1 && count($this->files_to_analyze) > $pool_size) {
-            $shuffle_count = $pool_size + 1;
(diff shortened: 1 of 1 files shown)


weirdan:
Apparently it was intended to populate `$process_file_paths` below, but wasn't actually used after e1e2ff3e57fa192492af46cc9defbbae6563ddd1. But according to https://github.com/vimeo/psalm/commit/e1e2ff3e57fa192492af46cc9defbbae6563ddd1#commitcomment-35482010 it did improve user experience somewhat :rofl:

orklah:
Interesting!

Maybe placebo effect? 😄

muglug:
Fixed in cf8dcc1

muglug:
But thanks, Psalm should have caught this. Mind reducing and creating an issue?

PR #4223, opened 2020-09-20
useless comparison

orklah:
This comparison seems useless. This is probably not the right fix, feel free to close this and fix this an other way

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php
@@ -706,7 +706,7 @@ class ArrayAssignmentAnalyzer
                     }
                 } elseif ($type instanceof TNonEmptyList
                     && $key_value instanceof Type\Atomic\TLiteralInt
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php
```diff
@@ -706,7 +706,6 @@ private static function updateTypeWithKeyValues(
                     }
                 } elseif ($type instanceof TNonEmptyList
                     && $key_value instanceof Type\Atomic\TLiteralInt
-                    && ($key_value->value === $key_value->value || $key_value->value < $type->count)
```
weirdan, line 709: while the lhs of that or looks really redundant, the rhs does not immediately strike as useless

orklah: yep, the first commit deleted only the first part, but this is a "or". So `true || A` is true. Deleting only the first part changed the behaviour, and indeed, CI failed as a result.

muglug:
Thanks!

PR #4227, opened 2020-09-20
misc changes

orklah:
another batch of small improvements. I'll explain them in review

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
@@ -519,7 +519,7 @@ abstract class ClassLikeAnalyzer extends SourceAnalyzer implements StatementsSou
         array $suppressed_issues,
         bool $emit_issues = true
     ): ?bool {
-        [$fq_class_name, $property_name] = explode('::$', (string)$property_id);
+        [$fq_class_name, $property_name] = explode('::$', $property_id);
 
         $codebase = $source->getCodebase();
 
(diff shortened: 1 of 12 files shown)


src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
```diff
@@ -519,7 +519,7 @@ public static function checkPropertyVisibility(
         array $suppressed_issues,
         bool $emit_issues = true
     ): ?bool {
-        [$fq_class_name, $property_name] = explode('::$', (string)$property_id);
+        [$fq_class_name, $property_name] = explode('::$', $property_id);
```
orklah, line 522: The variable is already a string

src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
```diff
@@ -716,9 +716,7 @@ public function interpretRefactors() : void
 
                 $source_class_storage = $this->codebase->classlike_storage_provider->get($source_parts[0]);
 
-                $destination_parts = explode('\\', $destination);
-
-                array_pop($destination_parts);
+                $destination_parts = explode('\\', $destination, -1);
```
orklah, line 719: explode has a param which can be negative. If positive, it can reduce the resulting array, if negative, it will remove last elements. This allow dropping the array_pop

src/Psalm/Type/Atomic/TTypeAlias.php
```diff
@@ -36,9 +36,7 @@ public function __toString(): string
             return $this->getKey() . '&' . implode(
                 '&',
                 array_map(
-                    function ($type): string {
-                        return (string) $type;
-                    },
+                    'strval',
```
orklah, line 39: more clear

muglug:
Thanks!

PR #4229, opened 2020-09-20
add strict_types declarations

orklah:
This PR propose adding strict_types everywhere it's possible.

I added them automatically in files in src/ . I just had to remove it from two files.

One of them does an implicit call to __toString that require not being in strict_types and the other one triggers a "redundant" error.
EDIT: I had to remove some more due to failing tests. I'll try to fix them one by one on a future PR.

### DIFF
--- a/src/Psalm/Aliases.php
+++ b/src/Psalm/Aliases.php
@@ -1,4 +1,7 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Psalm;
 
 class Aliases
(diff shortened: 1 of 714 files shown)


muglug:
I really don't think this is necessary – the sort of point of Psalm is to add ensure that types are all accurate, and I don't believe there's any added benefit of adding the strict_types directive from a speed standpoint

orklah:
I was wondering if this would be your response :)

Well, please let this open for now, I'll try to find the reasons why I had to revert some of them. I have the feeling there was implicit casts and errors not detected by psalm, It would be interesting to find the patterns

orklah:
Ok, I reported all I could, there were 3 types of issues that popped while adding strict_types: #4235, #4236, #4237

I'll close this now

PR #4236, opened 2020-09-22
preg_split can't take null in limit

orklah:
This removes the null type in preg_split third param according to:
https://3v4l.org/HEdl6

This should fail on CI because of:
https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php#L130


orklah:
The CI fails correctly:
https://github.com/vimeo/psalm/runs/1150843433#step:4:23

It should now be OK

muglug:
Thanks!

PR #4237, opened 2020-09-22
implicit int cast in array keys

orklah:
This code outputs a Type Error on strict_types=1

This is due to the infamous cast in array keys:
```php
$a = '12';
$b = [];
$b[$a] = 'a';
foreach($b as $key => $value){
    var_dump($key); // int(12)
}
```

This PR is more a discussion than an actual proposed fix.

Maybe we should have an issue in that sort of case at level 1 (alongside Mixed Issues) that only activates when the file is in strict_types and we use a string array-key where a string is expected?

### DIFF
--- a/src/Psalm/Type/Reconciler.php
+++ b/src/Psalm/Type/Reconciler.php
@@ -242,7 +242,7 @@ class Reconciler
                 ? clone $existing_types[$key]
                 : self::getValueForKey(
                     $codebase,
-                    $key,
+                    (string)$key,
                     $existing_types,
                     $new_types,
                     $code_location,


weirdan:
It's even worse with floats that are silently truncated to int, and *no* `TypeError` is thrown, even with `declare(strict_types=1)` (see https://3v4l.org/VqCqS). Psalm also happily accepts float as a key, and infers impossible type:  https://psalm.dev/r/6c4c5871dc

orklah:
> It's even worse with floats that are silently truncated to int, and _no_ `TypeError` is thrown, even with `declare(strict_types=1)` (see https://3v4l.org/VqCqS). Psalm also happily accepts float as a key, and infers impossible type: https://psalm.dev/r/6c4c5871dc

Welp, that's not great! Psalm could indeed handle this better, but at least, it was a mistake from the start by the developper. This could probably be an issue at a high level. The string case is more tricky, because there is no developper error, plus this behaviour is probably not known by many developpers and probably not expected by the rest of them! It can be fixed really easily but this is the kind of error that could mess your production (bonus points: it's only a problem if you tried to do things right and declared strict typing 😄)

orklah:
I converted to draft because, as I push my tests, I realize that this variable is used a lot more in contexts where we need a string. This fix is not adequat at all.

PR #4247, opened 2020-09-26
Fixed crash due to class/filename mismatch

weirdan:
Introduced in 56cddd16bf7, where `TaintNodeData` was renamed to `ControlFlowNodeData`


muglug:
Thanks! Sorry, have been heads-down working on a big refactor of unused variable detection.

PR #4259, opened 2020-09-29
Fix: allowMissingFiles not respected for ignored files

### LINKED ISSUE #4238: ignoreFiles must exists to run psalm
Our psalm configuration contains a directory that should be ignored:
```xml
   <projectFiles>
       <ignoreFiles>
           <directory name="tests/_support/_generated" />
        </ignoreFiles>
    </projectFiles>
```

In our CI environment we get this error message and exit with status code 1 (breaks the build):

> Problem parsing /builds/\<my-project\>/psalm.xml:
>   Could not resolve config path to /builds/\<my-project\>/src/\*/\*/tests/_support/_generated/

The reason is, that the tests are run in another job, so the generated support classes are not visible to psalm.
We can not remove the ingoreFiles-directive, as we need it in our dev environment.

Is there a way that psalm does not die when a folder, which it should ignore anyway, does not exist?

Might be related to #4025

ddeboer:
Fix #4238.

### DIFF
--- a/src/Psalm/Config/FileFilter.php
+++ b/src/Psalm/Config/FileFilter.php
@@ -259,7 +259,7 @@ class FileFilter
 
                 $file_path = realpath($prospective_file_path);
 
-                if (!$file_path) {
+                if (!$file_path && !$allow_missing_files) {
                     throw new ConfigException(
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

PR #4271, opened 2020-10-02
add some more stubs for array functions

marcosh:
This adds some more stubs for array functions

### DIFF
--- a/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
+++ b/src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
@@ -364,34 +364,6 @@ function array_merge_recursive(array $arr, array ...$arr2)
  * @psalm-template TKey as array-key
  * @psalm-template TValue
  *
- * @param array<TKey, TValue> $arr
- * @param array<TKey, TValue> ...$arr2
- *
- * @return array<TKey, TValue>
- */
-function array_merge(array $arr, array ...$arr2)
-{
-}
-
-/**
- * @template TValue
- *
- * @param int $startIndex
- * @param int $num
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Stubs/CoreGenericFunctions.phpstub
```diff
@@ -49,6 +49,25 @@ function array_intersect_key(array $arr, array $arr2, array ...$arr3)
...
  * @param array<TKey, TValue> $arr
+ * @param array $arr2
+ * @param array ...$arr3
+ * @param callable(TKey): int $keyCompareFunc
+ *
+ * @return array<TKey, TValue>
+ *
+ * @psalm-pure
+ */
+function array_intersect_ukey(array $arr, array $arr2, array ...$arr3, callable $keyCompareFunc): array {}
```
andrew-demb, line 60: Brackets mentioned twice

marcosh: thanks @andrew-demb

muglug:
Please remove the six stubs that are already covered by return type providers in Psalm (currently `array_fill`, `array_merge`, `array_pad`, `array_pop`, `array_rand`, `array_reduce`).

marcosh:
@muglug done. Just for the sake of understanding, are return type providers used when annotations are not enough to type correctly the function?

muglug:
> are return type providers used when annotations are not enough to type correctly the function?

Yes, exactly. When I introduced conditional return types I removed (I think) all the return type providers that were simple enough to be implemented instead as conditional return types.

muglug:
Thanks!

PR #4291, opened 2020-10-06
Fix #3543 - add documentation for @psalm-type and @psalm-import-type

### LINKED ISSUE #3543: Include the type tags in the psalm documentation
It seems that the documentation does not list at least these two tags: @psalm-type and @psalm-import-type. Please, include them.

leightonthomas:
Just added the examples from https://github.com/vimeo/psalm/releases/tag/3.11.5 as proposed in #3543


muglug:
Thanks!

PR #4325, opened 2020-10-14
Fix preg_match_all stub

Synchro:
[`preg_match_all`](https://www.php.net/preg_match_all) can return false given a bad pattern.

This PR fixes the stub, and adds a test for it. The test triggers a warning, so I suppress that since we're only interested in the `false` return value.


muglug:
Add `@psalm-ignore-falsable-return` to the signature as well

muglug:
Thanks!

PR #4331, opened 2020-10-14
ambiguous php version checks

orklah:
This PR aims to fix some case where the PHP version checks are ambiguous and could lead to regressions when the php_minor_version start from 0 again in PHP 8.0


weirdan (approved):

muglug:
Thanks!

PR #4332, opened 2020-10-14
changing the templates order to match existing stub in PHPStan

orklah:
This PR change the order of the templates in SplPriorityQueue to match those in PHPStan.

For reference, the original issue and discussion can be seen/continued here: https://github.com/phpDocumentor/phpDocumentor/pull/2625


muglug:
Thanks!

PR #4352, opened 2020-10-16
improve phpdoc

orklah:
In this PR, I used Psalter to improve a little the types in PHPDoc, mainly focusing on array -> list


muglug:
Thanks!

PR #4353, opened 2020-10-16
Report fails to write in not yet existing folders

pascalheidmann:
Run `psalm --report=in/a/subfolder/test.junit.xml somefile.php`
This will result in an error like `Warning: file_put_contents(in/a/subfolder/test.junit.xml): failed to open stream: No such file or directory in /app/src/Psalm/IssueBuffer.php on line 573` and the report will not be written.

This PR checks for the target dir and tries to create it otherwise.


muglug:
Thanks!

Issue #4356, opened 2020-10-17
[Idea] Allow traits to impose requirements on using class


NiclasvanEyk:
It would be really helpful for traits to be able to define a set of requirements, that the classes that use these traits need to fulfill. 

## Motivation

Often times, traits are created to be used by a single class or interface, like the [`InvalidCodeAnalysisTestTrait`](https://github.com/vimeo/psalm/blob/master/tests/Traits/InvalidCodeAnalysisTestTrait.php) from psalms own test suite. Among others, this trait requires an `addFile` method to be present on the using class, but this requirement is not being statically enforced or explicitly mentioned anywhere. 

One could make the effort of defining _all_ of such requirements as abstract methods, but this could be a lot of work. On top of that it introduces duplication, as one has to change these abstract methods, when the corresponding methods from the class for which the trait was created changes. If the signature of [`TestCase::addFile`](https://github.com/vimeo/psalm/blob/37ab91aa25d6312d058878cc76f697593eb3364c/tests/TestCase.php#L84) changes, and we would have defined this requirement in the `InvalidCodeAnalysisTestTrait`as an abstract method, this abstract method would also have to be updated.

As an added benefit, these requirements would enable IDEs to provide auto-completion for the methods that the trait depends on.

With this issue I just wanted to start the discussion and maybe get an idea on how likely it is to get implemented some day.

## Examples from other languages

- [Hacks `require extends` and `require implements`](https://docs.hhvm.com/hack/classes/trait-and-interface-requirements#traits). The documentation provides some good examples. The terminology used there is the same as in the following potential implementation example.
- [Dart Mixins and the `on`-keyword](https://dart.dev/guides/language/language-tour#adding-features-to-a-class-mixins). Darts `mixin`s are very similar to PHP, but support an `on` keyword, which allows for restricting the types of classes that the mixing can be used with.

## Potential implementation

This scenario is inspired by the [`SoftDeletes`-trait](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/SoftDeletes.php) from [Laravel](https://laravel.com/docs/8.x/eloquent#soft-deleting). The `@psalm-require-extends`-annotation is an example syntax for how to define the requirements that a trait imposes on the using class.

```php
abstract class DatabaseModel {
  // methods, properties, etc.
}

/**
 * @psalm-require-extends DatabaseModel
 */
trait SoftDeletingTrait {
[...]

PR #4361, opened 2020-10-17
Initial proposal for psalm-require-{extends, implements}

NiclasvanEyk:
This is an attempt to implement the `@psalm-require-extends` and `@psalm-require-implements` annotations, as discussed in #4356. It only concerns the static analysis regarding traits for now, wo the targeted behaviour is as follows:

For `@psalm-require-extends` on traits
```php
<?php
class A {}

/** @psalm-require-extends A */
trait RequiresA {}

class Invalid {
  use RequiresA; // This should throw an error
}

class Valid extends A {
  use RequiresA; // This is valid, as A is extended
}
```

For `@psalm-require-implements` on traits
```php
<?php
interface A {}
interface B {}

/** 
 * @psalm-require-implements A 
 * @psalm-require-implements B
 */
trait RequiresAB {}

class Invalid {
  use RequiresAB; // This should throw an error
}

class AlsoInvalid implements A {
  use RequiresAB; // This should throw an error
}

class Valid implements A, B {
  use RequiresAB; // This is valid, as A *and* B are implemented
}
```


muglug:
Great start!

You're missing an entry in `config.xsd` for `ExtensionRequirementViolation` and also documentation for the issue – the markdown file you create for documentation will need a snippet of PHP that demonstrates the issue (you can copy-paste the code from a test).

NiclasvanEyk:
Thank you for the quick feedback, I will implement it!

What is the recommended approach for filling the `SHORTCODE` and `ERROR_LEVEL` constants on a subclass of `CodeIssue`?

weirdan:
For `SHORTCODE` I'd suggest running `php bin/max_used_shortcode.php` and taking the next available value.

muglug:
`ERROR_LEVEL` should be `-1` — that makes it always trigger, which is what we want for any user-opted-in situations (where Psalm-specific annotations the user has added trigger issues)

NiclasvanEyk:
Ok, I added docs, entries in `config.xsd` and shortcodes taken from `bin/max_used_shortcode.php`. If there is still something missing, just let me know.

As this is now enough to impose both types of restriction and have them be checked by psalm I'll clear the draft status. The CI fails, but the issues seem to be unrelated to the added code, so I am not sure how to proceed here.

I would also want to add some code, so the language server would provide methods from the requirements like this:
```php
class A {
  public function foo() {}
}

/** @psalm-require-extends A */
trait B {
  public function bar()
  {
    $this-> // LS/IDE would suggest foo, because of the require-extends annotation
  }
}
```

but I feel this is better implemented in a following pr.

muglug:
Thanks, I'll deal with the remaining issues

muglug:
You accidentally removed a `use` statement here: https://github.com/vimeo/psalm/commit/0261024aa6b2921afb02b3f0c1efb57dd9bd7020#diff-4e427d91e2dcbfd421d5b849eb4e67efbe4885079b87ac618001820dc44fa339L20

PR #4385, opened 2020-10-21
WIP: start working on array_intersect_ukey return type provider

marcosh:
I'm trying to create a return type provider for `array_intersect_ukey` since the stub does not work as expected.

I have some questions:
- is the idea of return type providers to check the requirements on the parameters and if they are satisfied, then return a precise type, and otherwise return a broad type?
- at line 42 of `ArrayIntersectUKeyReturnTypeProvider` I's like to assert that the first type of the callable is broaded than the type of the keys of the first array argument. How should I do this?


muglug:
> is the idea of return type providers to check the requirements on the parameters and if they are satisfied, then return a precise type, and otherwise return a broad type?

Yes, there are _also_ param type providers that allow you to tell Psalm, given some arguments, what params it should expect (which is what's needed for a function with this weird signature)

> How should I do this?

Have a look at the `UnionTypeComparator`!

orklah:
@marcosh do you still intend to work on that?

PR #4401, opened 2020-10-22
Document ArrayAccess in PHP-DS

BenMorel:
The following classes/interfaces in PHP-DS implement `ArrayAccess`:

- `Map`
- `Stack`
- `Sequence` (`Deque`, `Vector`)

But this is not documented, and currently fails if you try to access such an object as an array:
https://psalm.dev/r/25e869aeb4

This PR fixes the stubs.

---

Script to check which classes/interfaces implement `ArrayAccess`:

```php
$classes = array_merge(get_declared_interfaces(), get_declared_classes());

foreach ($classes as $class) {
    if (substr($class, 0, 3) === 'Ds\\') {
        if ((new ReflectionClass($class))->implementsInterface(ArrayAccess::class)) {
            echo "$class\n";
        }
    }
}
```

```
Ds\Sequence
Ds\Vector
Ds\Deque
Ds\Stack
Ds\Map
```

### DIFF
--- a/stubs/ext-ds.php
+++ b/stubs/ext-ds.php
@@ -2,6 +2,7 @@
 
 namespace Ds;
 
+use ArrayAccess;
 use Countable;
 use JsonSerializable;
 use OutOfBoundsException;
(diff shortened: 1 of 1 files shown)


BenMorel:
Also reported on php-doc: https://github.com/php/doc-en/pull/161

muglug:
Thanks!

PR #4415, opened 2020-10-26
Ds\Queue & Ds\Set implement ArrayAccess as well

BenMorel:
Forgot these in #4401.

See [Queue](https://github.com/php-ds/polyfill/blob/master/src/Queue.php) & [Set](https://github.com/php-ds/polyfill/blob/master/src/Set.php) in the polyfill.

See also https://github.com/php-ds/ext-ds/pull/163.


muglug:
Thanks!

PR #4425, opened 2020-10-27
Add psalm-mutation-free to ext-ds

enumag:
I noticed that adding `@psalm-mutation-free` to `Collection::isEmpty()` doesn't affect `Set::isEmpty()` even though `Set implements Collection`. Do I have to copy the methods in such cases or is there a way to make the inheritance work?

### DIFF
--- a/stubs/ext-ds.php
+++ b/stubs/ext-ds.php
@@ -599,13 +599,13 @@ final class Pair implements JsonSerializable
 interface Sequence extends Collection, ArrayAccess
 {
     /**
-     * @return Sequence<TKey, TValue>
+     * @return Sequence<int, TValue>
      * @psalm-mutation-free
      */
     public function copy(): Sequence;
 
     /**
-     * @return array<TKey, TValue>
+     * @return array<int, TValue>
      * @psalm-mutation-free
      */
     public function toArray(): array;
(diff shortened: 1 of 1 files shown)


muglug:
You have to copy the methods – `@psalm-mutation-free` and similar are not inherited

enumag:
@muglug Okay, all done.

stubs/ext-ds.php
```diff
@@ -462,37 +598,80 @@ public function copy(): Pair
...
+     * @return Sequence<TKey, TValue>
+     * @psalm-mutation-free
+     */
+    public function copy(): Sequence;
+
+    /**
+     * @return array<TKey, TValue>
+     * @psalm-mutation-free
+     */
+    public function toArray(): array;
```
enumag, line 611: @muglug The failing tests are not related in my opinion. However I found a different error - TKey type doesn't exists in Sequence. This tells me that there is no test that the stubs are valid which I think there should be.

```suggestion
    /**
     * @return Sequence<int, TValue>
     * @psalm-mutation-free
     */
    public function copy(): Sequence;

    /**
     * @return array<int, TValue>
     * @psalm-mutation-free
     */
    public function toArray(): array;
```

muglug:
@enumag the stubs in general are tested, but I don't think the `ext-ds` ones are because they're only loaded conditionally.

enumag:
Could you somehow manually test them? Just to check if I didn't mess up something...

PR #4440, opened 2020-10-28
add stub with conditionnal return for is_a

### LINKED ISSUE #4439: psalm should complain about string input if $allow_class = false in is_a function
This is invalid
https://psalm.dev/r/61b64202dd
3rd argument needs to be true for this to work

orklah:
This PR propose adding a stub for is_a with a conditionnal return to try to fix #4439 

I made some tests here and it seems to work well:
https://psalm.dev/r/d5f0c18633

If we try with the example of the issue, it outputs:
https://psalm.dev/r/5eb4292c0b

Which may not be the result expected in the issue, but it makes the function always return false when called with a class-string when not allowed by $allow_string

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -779,3 +779,14 @@ function random_int(int $min, int $max): int {}
  * @psalm-pure
  */
 function array_count_values(array $array): array {}
+
+/**
+ * @template T of object
+ * @template U of object
+ * @param class-string<T>|T $object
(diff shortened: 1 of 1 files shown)


orklah:
This breaks some tests.

Some of them because psalm can now declare checks as redundant (which is cool!) but some of them because a string is pushed where we expect a class-string. This is not bad per se but Psalm has a config to allow string to class-string coercion and stubs don't respect it it seems...

I'm not sure I can fix this though. What do you think @muglug? Is this a good idea?

PR #4444, opened 2020-10-29
Pass CodeLocation to FunctionParamsProviderInterface interface

joehoyle:
Currently the `getFunctionParams()` method of the `FunctionParamsProviderInterface` is never passed the CodeLocation of the analyzed function call. As this is in-scope in the only call site, we can pass the CodeLocation. This means the `getFunctionParams()` is able to issue it's own Issues (which required the code location to attached the Issue to)


muglug:
Thanks!

PR #4469, opened 2020-11-02
improve mb_strtolower return type

### LINKED ISSUE #4455: mb_strtolower() should return lowercase-string rather than string
https://psalm.dev/r/b141bec26e

I found this while working on getting a psalm plugin updated that's creating a `\Psalm\Internal\MethodIdentifier` and I wanted to be correct and used `mb_strtolower()` on the method name - not that we're supposed to have method names where this matters, but still.

pilif:
this fixes #4455


muglug:
Thanks!

Issue #3113, opened 2020-04-09
Incorrect type inference with spread operator on arrays


tux-rampage:
Psalm incorrectly infers `non-empty-list<mixed>` on arrays constructed with the spread operator from typed iterable:

https://psalm.dev/r/f4951ecf90

tux-rampage:
Fixed by #4487

PR #4487, opened 2020-11-06
Fix type inference when unpacking typed iterables

tux-rampage:
This attempts to fix issue  #3113

When a typed iterable is unpacked into an array, it's template types are discarted and `list<mixed>` is assumed.


tux-rampage:
Failing Test: E2E asserting Psalm Version 4 on cli output. This was already failing in `master`.

muglug:
Thanks!

PR #4495, opened 2020-11-06
Add a `proxy` capability to the flow annotation

adrienlucas:
The idea is to be able to add control flow nodes using the `@psalm-flow` annotation.
You could want to : 
 - indicate that a call to a certain stubbed method is "actually" a call to another stubbed method. In the case of Doctrine stubs, we could indicate that a call to `Doctrine\DBAL\Connexion::insert` is in fact a call to `Doctrine\DBAL\Connexion::executeStatement` which is really a call to `Doctrine\DBAL\Connexion::prepare` which is a taint sink.
 - indicate that a call to a certain stubbed method is a call to another totally different method and that you want it analysed and have its return type taints be wired to the taint of the actual call. In the case of Symfony, we could indicate that a call to  `Symfony\Bundle\FrameworkBundle\Controller\AbstractController::renderView` is a call to `Twig\Environment::render` which could have a taint sink (or not depending on the template analysis).

What do you think ?

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -1062,7 +1062,32 @@ class FunctionCallAnalyzer extends CallAnalyzer
         }
 
         if ($function_storage) {
-            self::taintReturnType($statements_analyzer, $stmt, $function_id, $function_storage, $stmt_type);
+            $return_node = self::taintReturnType(
+                $statements_analyzer,
+                $stmt,
+                $function_id,
+                $function_storage,
+                $stmt_type
+            );
+
+            foreach ($function_storage->passthru_calls as $passthru_call) {
+                $pass_arguments = [];
+                foreach ($passthru_call['params'] as $i) {
+                    $pass_arguments[] = $stmt->args[$i];
+                }
+
+                $fake_call = new PhpParser\Node\Expr\FuncCall(
+                    new PhpParser\Node\Name\FullyQualified($passthru_call['fqcn']),
+                    $pass_arguments
+                );
+                ExpressionAnalyzer::analyze($statements_analyzer, $fake_call, $context);
+
+                if (null !== $return_node && $passthru_call['return'] === true) {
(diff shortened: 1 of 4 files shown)


muglug:
I'm not sure I see the value here.

If you can annotate that some function passes through to a taint sink, can't you just label _that_ as a taint sink?

adrienlucas:
> If you can annotate that some function passes through to a taint sink, can't you just label that as a taint sink?

Yes I guess, but what about the second use case, when you want another function to be analysed ?

Ultimately, I would like to be able to write something like this in the symfony plugin :

```php
/* Stubs/5/AbstractController.stubphp */

namespace Symfony\Bundle\FrameworkBundle\Controller;

use Symfony\Contracts\Service\ServiceSubscriberInterface;
use Psr\Container\ContainerInterface;

class AbstractController implements ServiceSubscriberInterface
{
    /**
     * @psalm-flow passthru Twig\Environment::render($view, $parameters) -> return
     */
    public function renderView($view, array $parameters = []) {}

    /**
     * @psalm-flow passthru self::renderView($template, $parameters) -> return
     */
    protected function render($view, array $parameters = [], Response $response = null) {}
}
```

But maybe I'm over-thinking it and I should only code this directly in the plugin.

muglug:
OK, would `proxy` be a better keyword perhaps?

adrienlucas:
Ok, I renamed to `proxy` and allowed to proxy to a method.

muglug:
Thanks!

zmitic:
@adrienlucas Will this solution allow to make proxy for forms?

Example:

```php

// UserType

/** @extends AbstractType<User> */
class UserType extends AbstractType
{
    // all methods know that $data will be ?User
}

class MyController extends AbstractController
{
    public function __invoke()
    {
        $form = $this->createForm(UserType::class);  // psalm should know this is FormInterface<User>
    }
}

abstract class AbstractController
{
    /**
     * PROBLEM here: for $type = UserType::class, psalm should return FormInterface<User>
     */
    protected function createForm(string $type, /** other */ ): FormInterface {}
}
```

Stub used: https://github.com/strictify/psalm-plugin-symfony/blob/symfony-forms/src/Stubs/common/forms/FormBuilderInterface.stubphp#L16

So could a proxy to ``FormBuilderInterface:getForm()`` be used, as long as T from AbstractType can be resolved?

PR #4552, opened 2020-11-14
Detect trying to access to a list with a negative offset

### LINKED ISSUE #4551: array-key of array returned by `explode()` cannot be negative
this should error, because the array key cannot exist

https://psalm.dev/r/c01f81a66f

would this work in case `explode` would return a `list` instead of a `array`?

orklah:
The goal of this PR is to fix #4551

It detects when the offset key is a negative int and flag the offset as invalid.

The error message is sub-optimal though. I wanted to display something like 
`Cannot access value on variable $a using offset value of -1, expecting positive-int|int(0)`

I saw two way of doing that:
- Adding an Union in $expected_offset_type => This fails and only displays `int`,
- Adding two types in $expected_offset_type => This fails because we take only the first in https://github.com/orklah/psalm/blob/1497cfb98c84f8487b7150593816e5f7baef907b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php#L1471

However, I think the current message is explicit enough given it will look like 
`Cannot access value on variable $a using offset value of -12, expecting positive-int`
It should be obvious that 0 is still an option for a list.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
@@ -797,9 +797,16 @@ class ArrayFetchAnalyzer
                                     $statements_analyzer
                                 );
                             }
+                            $has_valid_offset = true;
+                        } elseif (count($key_values) === 1
+                            && is_int($key_values[0])
+                            && $key_values[0] < 0
+                        ) {
+                            $expected_offset_types[] = Type::getPositiveInt();
+                            $has_valid_offset = false;
+                        } else {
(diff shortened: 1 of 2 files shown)


muglug:
Thanks, this is good enough –– though in the future you can construct the desired type with

`new Type\Union([new Type\Atomic\TLiteralInt(0), new Type\Atomic\TPositiveInt])`

orklah:
> 
> 
> Thanks, this is good enough –– though in the future you can construct the desired type with
> 
> `new Type\Union([new Type\Atomic\TLiteralInt(0), new Type\Atomic\TPositiveInt])`

Yeah, I did that, but Psalm display just ìnt`in the error message when doing so

PR #4554, opened 2020-11-14
$expr->name is not prefixed with $

LukasReschke:
The example as-is would currently not flag the following code:

```
	public function foo() {
		$foo = $bad_data;
		\shell_exec($foo);
        } 
```

Switching it to `bad_data` made it work.


muglug:
Thanks!

PR #4560, opened 2020-11-15
Fix #4549 Better intersection between parent types

### LINKED ISSUE #4549: dev-master: still issues with stubs and inheritance
Using Psalm dev-master@f65868c0236640dfd560854005decac38a4b7ef4

Same as #4539 but using this format to return the same value of the return value of the callable param:

```
    /**
     * @template U of B
     * @param callable(T): U $function
     * @return U
     */
    public function bind(callable $function);
```

If I remove the docblock from the original file, it works as expected.

Source code: [branch `stub-issue-3`](https://github.com/thomasvargiu/psalm-issues-tests/tree/stub-issue-3)
CI test: [branch `stub-issue-3`](https://github.com/thomasvargiu/psalm-issues-tests/runs/1398850170?check_suite_focus=true#step:8:1)

CI Result:

```
Running psalm on the following code
\n------------------ TEST CODE -----------------\n
<?php

use Original\Foo;

require __DIR__ . '/vendor/autoload.php';

/** @psalm-trace $foo */
$foo = new Foo('foo');

/** @psalm-trace $value */
$value = $foo->bind(function (string $a): Foo {
    return new Foo(strlen($a));
});

// Value should be Foo<int>

\n---------------- END TEST CODE ---------------\n
Using Psalm dev-master@f65868c0236640dfd560854005decac38a4b7ef4
---------------- STARTING TEST ---------------\n
Scanning files...
Analyzing files...

E

ERROR: Trace - test.php:8:1 - $foo: Original\Foo<string(foo)> (see https://psalm.dev/224)
/** @psalm-trace $foo */
$foo = new Foo('foo');


ERROR: Trace - test.php:11:1 - $value: empty (see https://psalm.dev/224)
/** @psalm-trace $value */
$value = $foo->bind(function (string $a): Foo {
    return new Foo(strlen($a));
});
```

Value should be `Foo<int>`

thomasvargiu:
This fix #4549 doing a better intersection, getting the most specific type.

The problem occurs with stubs, when we have something like this:

```php
interface Either
{
    /**
     * @param callable $left
     * @param callable $right
     * @return mixed
     */
    public function either(callable $left, callable $right);
}
```

and we have a tow stubs, one for the interface and another for an implementation that override the interface method types:

```php
/**
 * @template E
 * @template A
 */
interface Either
{
    /**
     * @template TE
     * @template TA
     * @param callable(E): TE $left
     * @param callable(A): TA $right
     * @return TE|TA
     */
    public function either(callable $left, callable $right);
}

/**
 * @template E
 * @template A
 * @template-implements Either<E, A>
 */
class Right
{
    /**
     * @template TE
     * @template TA
     * @param callable(E): TE $left
     * @param callable(A): TA $right
     * @return TA
     */
    public function either(callable $left, callable $right);
}
```

Without this change, the intersection will take type references from the stubbed interface and not from the implementation, breaking the template_type inference.

Sorry, I wasn't able to reproduce it in a test.

### DIFF
--- a/src/Psalm/Internal/Codebase/Methods.php
+++ b/src/Psalm/Internal/Codebase/Methods.php
@@ -779,11 +779,19 @@ class Methods
                 if ((!$old_contained_by_new && !$new_contained_by_old)
                     || ($old_contained_by_new && $new_contained_by_old)
                 ) {
-                    $attempted_intersection = Type::intersectUnionTypes(
-                        $overridden_storage->return_type,
-                        $candidate_type,
-                        $source_analyzer->getCodebase()
-                    );
+                    if ($old_contained_by_new && $new_contained_by_old) {
+                        $attempted_intersection = Type::intersectUnionTypes(
+                            $candidate_type,
+                            $overridden_storage->return_type,
+                            $source_analyzer->getCodebase()
+                        );
+                    } else {
+                        $attempted_intersection = Type::intersectUnionTypes(
+                            $overridden_storage->return_type,
+                            $candidate_type,
+                            $source_analyzer->getCodebase()
+                        );
+                    }
 
                     if ($attempted_intersection) {
                         $self_class = $overridden_method_id->fq_class_name;


thomasvargiu:
I see that this broke another test for the nested interface definitions :(

muglug:
Thanks!

PR #4572, opened 2020-11-16
Add sinks for popen and proc_open

LukasReschke:
User input in those two functions could lead to a RCE.

popen: https://www.php.net/manual/en/function.popen.php
proc_open: https://www.php.net/manual/en/function.proc-open.php


muglug:
Thanks!

PR #4575, opened 2020-11-16
Allow named arguments to variadic functions

### LINKED ISSUE #4563: False positive InvalidNamedArgument for named arguments to variadic functions
Psalm doesn't recognize that variadic functions can be passed arbitrary named arguments:

https://psalm.dev/r/ad5fcd9926

I _think_ this fixes it, but I'm not familiar enough with the codebase to know if it might have unwanted side effects. It might also be worth considering an annotation to indicate whether a variadic parameter should be allowed to handle named parameters.

``` diff
diff --git a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
index 1ce336501..b0c6811ff 100644
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
@@ -665,6 +665,10 @@ class ArgumentsAnalyzer
                         $arg_function_params = [$candidate_param];
                         break;
                     }
+                    if ($candidate_param->is_variadic) {
+                        $arg_function_params = [$candidate_param];
+                        break;
+                    }
                 }
 
                 if (!$arg_function_params) {

```

duskwuff:
Closes #4563

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
@@ -661,7 +661,7 @@ class ArgumentsAnalyzer
                 }
             } elseif ($arg->name && $function_storage && $function_storage->allow_named_arg_calls) {
                 foreach ($function_params as $candidate_param) {
-                    if ($candidate_param->name === $arg->name->name) {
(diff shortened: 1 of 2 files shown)


muglug:
Thanks!

PR #4576, opened 2020-11-16
Add more Psalm flows for string functions

LukasReschke:
This adds string functions from https://www.php.net/manual/en/ref.strings.php

This commit adds the flows for functions from "addcslashes" to "sprintf".

More are to follow in later commits, if that is deemed useful :)

Ref #3636

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -443,6 +443,7 @@ function date(string $format, int $timestamp = 0) {}
  *
  * @param mixed $vars
  * @param-out string|int|float $vars
+ * @psalm-flow ($str, $format) -> return
  * @return (func_num_args() is 2 ? list<float|int|string> : int)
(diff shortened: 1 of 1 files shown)


muglug:
This is great, thanks!

thomasbley:
Using the new release 4.2, I get "Detected tainted header" using https://psalm.dev/r/21c22b8fb6

What is the best way to handle this?

LukasReschke:
@thomasbley You could create a new function that is marked as sanitizer.

https://psalm.dev/r/5fb5ba744f

```
<?php // --taint-analysis

$test = (string) ($_POST['test'] ?? '');


/**
 * @psalm-taint-escape header
 */
function addFileNameToContentDisposition(string $fileName) : string {
    return addslashes($fileName);
}

header('Content-Disposition: attachment; filename="' . addFileNameToContentDisposition($test) . '"');
```

PR #4577, opened 2020-11-16
mysqli::$insert_id can be a string

BenMorel:
Source: https://www.php.net/manual/en/mysqli.insert-id.php

> Note:
> If the number is greater than maximal int value, mysqli_insert_id() will return a string.

### DIFF
--- a/dictionaries/PropertyMap.php
+++ b/dictionaries/PropertyMap.php
@@ -260,7 +260,7 @@ return [
         'field_count' => 'int',
         'host_info' => 'string',
         'info' => 'string',
-        'insert_id' => 'int',
+        'insert_id' => 'int|string',
         'protocol_version' => 'string',
         'server_info' => 'string',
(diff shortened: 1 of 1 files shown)


dictionaries/PropertyMap.php
```diff
@@ -260,7 +260,7 @@
         'field_count' => 'int',
         'host_info' => 'string',
         'info' => 'string',
-        'insert_id' => 'int',
+        'insert_id' => 'int|string',
```
staabm, line 263: Wouldn’t  it make sense to have `numeric-string` here?

BenMorel: Probably. @muglug?

BenMorel: Actually, this is one of the many places where I would like to see an `int-string` type. `numeric-string` is just way too broad to be relevant in real life, IMHO.

PR #4579, opened 2020-11-16
DateTimeInterface::getTimeZone() can return false

### LINKED ISSUE #4515: DateTimeInterface::getTimeZone() can return false
[DateTimeInterface::getTimeZone()](https://www.php.net/manual/en/datetime.gettimezone.php) can return `false`:

> public DateTimeInterface::getTimezone ( ) : DateTimeZone|false

However, Psalm thinks it can return `DateTimeZone` only:

https://psalm.dev/r/cdb412cd2a

BenMorel:
Fixes #4515


muglug:
Thanks!

PR #4580, opened 2020-11-16
Fix curl_multi_getcontent signature

TysonAndre:
See https://www.php.net/curl_multi_getcontent and php-src ext/curl/curl.stub.php


muglug:
Thanks, and sorry!

Issue #4561, opened 2020-11-15
PossibilyNullIterator and PossiblyInvalidIterator on preg_filter($array) result


mstilkerich:
Hello,

  I am getting spurious errors reported by psalm when attempting to iterate an array returned by the `preg_filter` PHP function.

According to the [preg_filter doc](https://www.php.net/manual/en/function.preg-filter.php):

> Returns an array if the subject parameter is an array, or a string otherwise.
> If no matches are found or an error occurred, an empty array is returned when subject is an array or NULL otherwise. 

In my case, psalm knows that `$subject` is an array, so according to the above it can be assumed an array will be returned by `preg_filter`, even in case of error. Consequently, psalm should throw neither of these error messages which are based on either a `null` or a `string` return value.

Snippet:
https://psalm.dev/r/fa931259f1

(I know this is a very specific case and preg_filter may return null or a string given different parameters, but it would nevertheless be great if psalm could handle this without throwing errors)

orklah:
Hey!

This should be pretty easy to add, you just have to copy the stub made for preg_replace here:
https://github.com/vimeo/psalm/blob/master/stubs/CoreGenericFunctions.phpstub#L654

And adapt it for preg_filter.

Do you want to try to propose a PR?

mstilkerich:
Thanks for the hint, that really was easy to fix.

PR #4587, opened 2020-11-17
Stub for preg_filter

mstilkerich:
As suggested here is a pull request for issue #4561. It seems to work for me.

However, a few notes from my side:
* According to the documentation, preg_filter returns an array if $subject is an array. If $subject is a string, it returns a string _or null_ in case of error. The null case is missing from the `@return` annotation. However, `preg_replace` can also return null in case of error, so I assumed the author of `preg_replace` stub knows better than I and left it this way.
* I have no idea what `@psalm-flow` does and could not find documentation.

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -664,6 +664,19 @@ function str_replace($search, $replace, $subject, &$count = null) {}
 /**
  * @psalm-pure
  *
+ * @param string|string[] $pattern
+ * @param string|array<string|int|float> $replacement
+ * @param string|array<string|int|float> $subject
+ * @param int $count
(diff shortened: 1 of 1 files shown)


LukasReschke:
> I have no idea what @psalm-flow does and could not find documentation.

This is for the taint flow analysis (e.g. used by https://psalm.dev/docs/security_analysis/.)

```
 * @psalm-flow ($replacement, $subject) -> return
```

Defines that whatever input flows in `$replacement` and `$subject` is also in the response. (e.g. if an attacker can control the input to `$replacement` the output would be tainted as well)

Looking at the PHP documentation for `preg_filter`, that seems accurate.

muglug:
Thanks!

mstilkerich:
Thanks for the explanation, Lukas.

Issue #915, opened 2018-07-24
strtr - Too few arguments for method strtr - expecting 3 but saw 2


iluuu1994:
> Too few arguments for method strtr - expecting 3 but saw 2

This issue was fixed before:
https://github.com/vimeo/psalm/issues/831

Unfortunately, I cannot reproduce the problem online...
https://getpsalm.org/r/3fcf383db0

I'm on Psalm 2.0.9.

Any clue what might be going on here?

muglug:
What are the types of the args you're passing? Does explicitly casting them help?

iluuu1994:
For whatever reason, suddenly the error message is different:

> Argument 2 of strtr expects string, array{%from_date%:mixed, %to_date%:mixed} provided

I have not opened the project since 10 days ago. I can see in `git` that the code has not changed. No `composer update`. Someone please explain that 😆

Explicitly casting the values in the array results in a very similar error message:

> Argument 2 of strtr expects string, array{%from_date%:string, %to_date%:string} provided

muglug:
And what if you cast the first arg?

iluuu1994:
That actually fixes the issue! I cannot for the life of me reproduce this issue on getpsalm.org:
https://getpsalm.org/r/a07c5ba9a1

Maybe it has to do something with my config. There's currently no way to add a config to the online version of psalm? That could make it easier to reproduce issues.

muglug:
Thanks!

iluuu1994:
Cool, thanks man!

Issue #698, opened 2018-04-25
Function/method overloading


weirdan:
Basically, I want the following to typecheck:
```php
<?php
  interface Query {}
  class LimitedQuery implements Query { public function getPageSize(): int { return 10; }}
  class UnlimitedQuery implements Query {}
  
  interface Result {}
  class PagedResult implements Result { public function getPageCount(): int { return 1; }}
  class PlainResult implements Result { public function getAllRecords(): array { return []; }}

  /**
   * @method PagedResult search(LimitedQuery $q)
   * @method PlainResult search(UnlimitedQuery $q)
   * @psalm-seal-methods
   */
  class Searcher {
    public function __call(string $methodName, array $args = []): Result {
      if ('search' !== $methodName) throw new \BadMethodCallException;
      if ($args[0] instanceof LimitedQuery) return new PagedResult;
      else return new PlainResult;
    }
  }
  (new Searcher)->search(new LimitedQuery)->getPageCount(); // ok
  (new Searcher)->search(new LimitedQuery)->getAllRecords(); // not ok, PagedResult does not have that method
  (new Searcher)->search(new UnlimitedQuery)->getAllRecords(); // ok
  (new Searcher)->search(new UnlimitedQuery)->getPageCount(); // not ok, PlainResult doesn't have that method
```
I understand this must be a lot of work (going from a single signature for a symbol to multiple is not an easy thing to implement). Name mangling might be a viable approach (a-la C++).

Ideally this would be supported by other major players in the field (static analyzers, completion providers, language servers, doc generators, etc), so it might be beneficial to bring this up in the context of PSR-5

muglug:
Psalm would have to generate a new type of return type here

```
class TConditional extends Psalm\Type\Atomic {
  public $base_type;
  /** @var array<int, array{0: ArgRequirement, 1: Psalm\Type\Union}> */
  public $cases;
}
```

weirdan:
Some syntax for functions (where `@method` is not applicable) would be useful as well.
This might allow more precise contracts to be expressed, when, say, you can only call the function with 2 or 4 parameters, but never 3.

rzvc:
> Some syntax for functions (where `@method` is not applicable) would be useful as well.

Would be nice if that same syntax could be used for methods too, or at the very least, have ```@method``` available in the method's docblock too. Having the overloads declared in the class' docblock creates a disconnection between declaration and implementation. The only reason it currently needs to be there is because it's used to declare magic methods.

[...]

PR #4591, opened 2020-11-17
Add psalm-flow for string functions from sscanf to wordwrap

LukasReschke:
This should conclude all string functions from https://www.php.net/manual/en/book.strings.php

Continuation of https://github.com/vimeo/psalm/pull/4576

Ref https://github.com/vimeo/psalm/issues/3636

**TODO**

- [x] strtr return provider
- [x] Fix tests 
- [x] Cleanup commits + remove accidentially commited files

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -890,6 +890,7 @@ function sprintf(string $format, ...$args) : string {}
 /**
  * @psalm-pure
  * @return string|false
+ * @psalm-ignore-falsable-return
  * 
  * @psalm-flow ($format, $args) -> return
  */
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -487,6 +487,20 @@ function strtoupper(string $str) : string {}
 /**
  * @psalm-pure
  *
+ * @param string|array<string> $string
+ * @param string|array<string> $replacement
+ * @param int|array<int> $start
+ * @param null|int|array<int> $length
+ *
+ * @return ($string is array<string> ? array<string> : string) 
+ *
+ * @psalm-flow ($string, $replacement) -> return
+ */
+function substr_replace($string, $replacement, $start, $length) {}
+
+/**
+ * @psalm-pure
+ *
  * @param string $haystack
  *
  * @psalm-return positive-int|0|false
@@ -613,6 +627,20 @@ function strip_tags(string $str, ?string $allowable_tags = null) : string {}
 /**
  * @psalm-pure
  *
+ * @psalm-flow ($str) -> return
+ */
+function stripcslashes(string $str) : string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($str) -> return
+ */
+function stripslashes(string $str) : string {}
+
+/**
+ * @psalm-pure
+ *
  * @psalm-taint-escape html
  *
  * @psalm-flow ($string) -> return
@@ -664,6 +692,104 @@ function str_replace($search, $replace, $subject, &$count = null) {}
 /**
  * @psalm-pure
  *
+ * @param string|array<string|int|float> $search
+ * @param string|array<string|int|float> $replace
+ * @param string|array<string|int|float> $subject
+ * @param int $count
+ * @return ($subject is array ? array<string> : string)
+ *
+ * @psalm-flow ($replace, $subject) -> return
+ */
+function str_ireplace($search, $replace, $subject, &$count = null) {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($input, $pad_string) -> return
+ */
+function str_pad(string $input, int $pad_length, $pad_string = '', int $pad_type = STR_PAD_RIGHT): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($input) -> return
+ */
+function str_repeat(string $input, int $multiplier): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($str) -> return
+ */
+function str_rot13(string $str): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($str) -> return
+ */
+function str_shuffle(string $str): string {}
+
+/**
+ * @psalm-pure
+ * @return array<string>
+ *
+ * @psalm-flow ($string) -> return
+ */
+function str_split(string $string, int $split_length = 1) {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function strstr(string $haystack, string $needle, bool $before_needle = false): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function stristr(string $haystack, string $needle, bool $before_needle = false): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function strchr(string $haystack, string $needle, bool $before_needle = false): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function strpbrk(string $haystack, string $char_list): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function strrchr(string $haystack, string $needle): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($string) -> return
+ */
+function strrev(string $string): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($string) -> return
+ */
+function strtok(string $str, string $token): string {}
+
+/**
+ * @psalm-pure
+ *
  * @param string|string[] $pattern
  * @param string|array<string|int|float> $replacement
  * @param string|array<string|int|float> $subject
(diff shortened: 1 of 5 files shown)


stubs/CoreGenericFunctions.phpstub
```diff
@@ -661,6 +696,103 @@ function htmlspecialchars_decode(string $string, ?int $quote_style = null) : str
...
+ * @psalm-flow ($str) -> return
+ */
+function str_shuffle(string $str): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($string) -> return
+ */
+function str_split(string $string, int $split_length = 1): array {}
```
LukasReschke, line 745: I guess this should be more specific considering the test failure seems related to https://github.com/vimeo/psalm/blob/ebf33723cb2128cb2f34bb9cd87321c661e18fac/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php#L160-L161

LukasReschke:
Any pointers why the following test is failing? I defined the second argument as `mixed` as per php.net (https://www.php.net/manual/en/function.strstr.php). But the test seems to intentionally test wrong behaviour?

https://github.com/vimeo/psalm/blob/494ec407775c30c2ff8c60dddfac34efb83bd17c/tests/FunctionCallTest.php#L504-L515

weirdan:
> But the test seems to intentionally test wrong behaviour?

The test was added along with the fix for #915, and was probably related to how Psalm selects the overloaded signature (`strtr()` has two) from the callmap.

>  defined the second argument as mixed as per php.net

That's too wide. It should be either string or array, depending on the number of parameters (string for 3 arg signature and array for 2).

LukasReschke:
> That's too wide. It should be either string or array, depending on the number of parameters (string for 3 arg signature and array for 2).

Oof. Seems the PHP documentation omits that behaviour for arrays straight out. Even though it works fine as per https://3v4l.org/7PJG7 :-/ 

I'm going to fix this, thanks! Also the CallMap was new to me, good to know. :)

stubs/CoreGenericFunctions.phpstub
```diff
@@ -484,6 +484,27 @@ function strtolower(string $str) : string {}
...
+ * @psalm-flow ($str, $replace_pairs) -> return
+ */
+function strtr(string $str, array $replace_pairs) : string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($string, $replacement) -> return
+ */
+function substr_replace(mixed $string, mixed $replacement, mixed $start, ?mixed $length) : mixed {}
```
LukasReschke, line 506: TODO: Fix the mixed.

stubs/CoreGenericFunctions.phpstub
```diff
@@ -661,6 +696,104 @@ function htmlspecialchars_decode(string $string, ?int $quote_style = null) : str
...
+ * @psalm-flow ($string) -> return
+ */
+function str_split(string $string, int $split_length = 1) {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function strstr(string $haystack, mixed $needle, bool $before_needle = false): string {}
```
LukasReschke, line 753: Fix mixed

stubs/CoreGenericFunctions.phpstub
```diff
@@ -661,6 +696,104 @@ function htmlspecialchars_decode(string $string, ?int $quote_style = null) : str
...
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function stristr(string $haystack, mixed $needle, bool $before_needle = false): string {}
```
LukasReschke, line 760: Fix mixed

stubs/CoreGenericFunctions.phpstub
```diff
@@ -661,6 +696,104 @@ function htmlspecialchars_decode(string $string, ?int $quote_style = null) : str
...
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function strchr(string $haystack, mixed $needle, bool $before_needle = false): string {}
```
LukasReschke, line 767: Fix mixed

stubs/CoreGenericFunctions.phpstub
```diff
@@ -661,6 +696,104 @@ function htmlspecialchars_decode(string $string, ?int $quote_style = null) : str
...
+ * @psalm-flow ($haystack) -> return
+ */
+function strpbrk(string $haystack, string $char_list): string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($haystack) -> return
+ */
+function strrchr(string $haystack, mixed $needle): string {}
```
LukasReschke, line 781: Fix mixed

stubs/CoreGenericFunctions.phpstub
LukasReschke: It would be really nice if conditional types for parameters would be supported. They don't seem to be at the moment.

Example:

```
 * @param ($string is array<string> ? array<string> : string) $replacement
 * @param ($string is array<string> ? array<int> : int) $start
 * @param ($string is array<string> ? array<int> : int) $length
```

LukasReschke: I also need to read up on `FunctionReturnTypeProviderInterface` / `StrReplaceReturnTypeProvider` to understand how one propagates taint there.

weirdan: > It would be really nice if conditional types for parameters would be supported. They don't seem to be at the moment.

Yep. Also related: #698

stubs/CoreGenericFunctions.phpstub
```diff
@@ -484,6 +484,27 @@ function strtolower(string $str) : string {}
...
+ * @psalm-flow ($str, $to) -> return
+ */
+function strtr(string $str, string $from, string $to) : string {}
+
+/**
+ * @psalm-pure
+ *
+ * @psalm-flow ($str, $replace_pairs) -> return
+ */
+function strtr(string $str, array $replace_pairs) : string {}
```
muglug: Psalm doesn't handle duplicate stubbed signatures for the same function, unfortunately. This will need to be a plugin instead (similar plugins are here: https://github.com/vimeo/psalm/tree/master/src/Psalm/Internal/Provider/ReturnTypeProvider)

stubs/CoreGenericFunctions.phpstub
muglug: @LukasReschke taints are propagated by copying the `parent_nodes` property on a type.

As an aside it's slightly weird to have so much information on a type object – I may migrate some of it to a `Data` object or similar, but that'd be a major version breaking change.

src/Psalm/Internal/Provider/ReturnTypeProvider/StrReplaceReturnTypeProvider.php
```diff
@@ -1,62 +0,0 @@
...
-class StrReplaceReturnTypeProvider implements \Psalm\Plugin\Hook\FunctionReturnTypeProviderInterface
-{
-    public static function getFunctionIds() : array
-    {
-        return [
-            'str_replace',
-            'str_ireplace',
-            'substr_replace',
-            'preg_replace',
-            'preg_replace_callback',
```
LukasReschke, line 20: Those are all defined in `stubs/CoreGenericFunctions.phpstub` now and have proper return types. I believe this file can be removed.

orklah: See https://github.com/vimeo/psalm/pull/3503#issuecomment-643867424, there are still some tweaks in the Provider that can't be deleted

LukasReschke: Joy :-) - Guess this PR will be tougher then I anticipated :-)

staabm:
Side-note: could someone elaborate why/when psalm-flow should be used?

Its not documented.. is it a psalm internal-only phpdoc?

orklah:
> Side-note: could someone elaborate why/when psalm-flow should be used?
> 
> Its not documented.. is it a psalm internal-only phpdoc?

See https://github.com/vimeo/psalm/pull/4587#issuecomment-729114908

LukasReschke:
This should be ready to review now. For `strtr` I added a provider (`StrTrReturnTypeProvider`) that will propagate the taint flow for all parameters.

LukasReschke:
The error being thrown is from https://github.com/sebastianbergmann/phpunit/blob/58f69d18fd1e61b54c18cf9d06b762507be55562/src/TextUI/DefaultResultPrinter.php#L149-L154:


```
        if (!in_array($colors, self::AVAILABLE_COLORS, true)) {
            throw InvalidArgumentException::create(
                3,
                vsprintf('value from "%s", "%s" or "%s"', self::AVAILABLE_COLORS)
            );
        }
```

As per php.net:

> https://www.php.net/manual/en/function.vsprintf.php

```
vsprintf ( string $format , array $args ) : string|false

Return array values as a formatted string according to format, or FALSE on failure.
```

And the code doesn't check for null. - Should this be accepted behaviour? For `sprintf` it seems we just ignore the fact that it could return `false`.

muglug:
Thanks!

Possibly-false returns in standard library functions are annoying. In > 99.9% of their common usage those functions don’t return `false`, so Psalm has an annotation `@psalm-ignore-falsable-return` that triggers special behaviour in a bunch of places.

orklah:
I have a few differents false positives that popped in my work codebase that seems related to this PR.

I'll need to verify them and post issues but if someone is faster than me:
- stristr return type seems to not contain false anymore
- substr_replace 4th param should be optional and it became mandatory
- I have new errors related to the offset of the array returned by str_split. Not sure if it's a false positive or if the type was narrowed

Issue #4596, opened 2020-11-17
[Security] Improve file sinks


LukasReschke:
Thoughts:

- Create custom category for file sinks
- Differentiate between read/write/delete
- Investigate missing sinks (e.g. `readfile`, `unlink`, etc) - requires going through https://www.php.net/manual/en/ref.filesystem.php

LukasReschke:
> Differentiate between read/write/delete

Should probably be the same category (e.g. `file`) at the time being still so the same sanitizers can be used. Maybe something to improve in the future.

PR #4630, opened 2020-11-19
Add dedicated types for 'file', 'header' and 'cookie'

LukasReschke:
This creates the three new dedicated rules and the responding documentation:

- file
- header
- cookie

For the `file` type, this change also adds more sinks.

Fixes https://github.com/vimeo/psalm/issues/4596


dictionaries/InternalTaintSinkMap.php
```diff
@@ -5,10 +5,24 @@
 return [
 'exec' => [['shell']],
 'create_function' => [['text'], ['eval']],
-'file_get_contents' => [['text']],
-'file_put_contents' => [['shell']],
-'fopen' => [['shell']],
-'header' => [['text']],
+'file_get_contents' => [['file']],
+'file_put_contents' => [['file']],
```
LukasReschke, line 9: Technically that could also be used for RCE (by writing a file somewhere in the web root).

I'll make sure to document the different risks in the documentation for the `file` type. In the future we could consider having separate types that all potentially share similar sanitizers though.

muglug:
Thanks!

PR #4632, opened 2020-11-19
return static instead of self when static context detected

orklah:
This PR propose improving the handling of functions that return $this.

The goal is to transform this code:
```php
class HelloWorld
{
    public function sayHello()
    {
        return $this;
    }
}
```
into:

```php
class HelloWorld
{
    /**
     * @return $this
     */
    public function sayHello(): self
    {
        return $this;
    }
}
```

Bonus points: I want to propose another PR so that this example return `static` instead of `self` in PHP8 but I'm not sure how I'm supposed to access to the php version inside an Atomic. Can you give me a hint?

### DIFF
--- a/src/Psalm/Type.php
+++ b/src/Psalm/Type.php
@@ -115,7 +115,7 @@ abstract class Type
     ) : string {
         if ($allow_self && $value === $this_class) {
             if ($was_static) {
-                return '$this';
+                return 'static';
             }
             return 'self';
         }
--- a/src/Psalm/Type.php
+++ b/src/Psalm/Type.php
@@ -110,9 +110,13 @@ abstract class Type
         ?string $namespace,
         array $aliased_classes,
         ?string $this_class,
-        bool $allow_self = false
+        bool $allow_self = false,
+        bool $was_static = false
     ) : string {
         if ($allow_self && $value === $this_class) {
+            if ($was_static) {
+                return '$this';
+            }
             return 'self';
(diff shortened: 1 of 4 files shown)


src/Psalm/Type.php
```diff
@@ -110,9 +110,13 @@ public static function getStringFromFQCLN(
         ?string $namespace,
         array $aliased_classes,
         ?string $this_class,
-        bool $allow_self = false
+        bool $allow_self = false,
+        bool $was_static = false
     ) : string {
         if ($allow_self && $value === $this_class) {
+            if ($was_static) {
```
staabm, line 117: The condition should be inverse?

muglug:
I'd prefer `@return static`, especially as it's PHP-8-compatible

orklah:
> I'd prefer `@return static`, especially as it's PHP-8-compatible

Oh? I thought `$this` was more narrow than `static` (probably just semantic, `$this`means that the object itself was returned when `static` just meant the type was the same as the object being called on, but not necessary the same instance) but `static` is fine by me too.

src/Psalm/Type.php
orklah: Why do you think so? This boolean is true when in static context so we add `$this`/`static` on the phpdoc when it's true

muglug:
Thanks! Psalm treats `$this` and `static` the same in docblocks, FWIW.

PR #4639, opened 2020-11-20
feature: allow plugin manager to work without config file

### LINKED ISSUE #3640: psalm-plugin cannot run without a config file
It would be nice if you could run `psalm-plugin` without a config file, in which case it just tells you which plugins it can find currently, without being able to enable/disable anything.

Use case: packaging Psalm with several plugins into `jakzal/toolbox`: https://github.com/jakzal/toolbox/pull/252

dkarlovi:
Closes #3640.

### DIFF
--- a/src/Psalm/Internal/PluginManager/PluginList.php
+++ b/src/Psalm/Internal/PluginManager/PluginList.php
@@ -1,6 +1,7 @@
 <?php
 namespace Psalm\Internal\PluginManager;
 
+use RuntimeException;
 use function array_diff_key;
 use function array_flip;
 use function array_key_exists;
@@ -9,7 +10,7 @@ use function strpos;
 
 class PluginList
 {
-    /** @var ConfigFile */
+    /** @var null|ConfigFile */
     private $config_file;
 
     /** @var ComposerLock */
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/PluginManager/PluginList.php
```diff
@@ -34,9 +35,11 @@ public function getEnabled(): array
     {
         if (!$this->enabled_plugins) {
```
dkarlovi, line 36: This is likely an error: it's claiming to be memoized here, but if no plugins are enabled, it will retry since an empty array is also falsy.

src/Psalm/Internal/PluginManager/PluginListFactory.php
```diff
@@ -24,7 +24,11 @@ public function __construct(string $project_root, string $psalm_root)
 
     public function __invoke(string $current_dir, ?string $config_file_path = null): PluginList
     {
-        $config_file = new ConfigFile($current_dir, $config_file_path);
+        try {
+            $config_file = new ConfigFile($current_dir, $config_file_path);
+        } catch (\RuntimeException $exception) {
```
dkarlovi, line 29: Would be better to throw a more specific exception, but for now it will do.

src/Psalm/Internal/PluginManager/PluginList.php
weirdan: Should probably be `if (null === $this->enabled_plugins)`

dkarlovi: Yes, but it's unrelated to this PR.

dkarlovi (commented):
Ready for review.

muglug:
Thanks!

PR #4641, opened 2020-11-20
allow psalter to use static return type in PHP8

orklah:
This PR propose allowing `static` return type when using Psalter starting from PHP8.

I had to change the signature of `canBeFullyExpressedInPhp` to add major and minor php version in order to remove the return phpdoc when it was identical to the return type


muglug:
Oh you don't need to do that – instead you can use the `toPhpString` method, which takes version numbers

orklah:
> 
> 
> Oh you don't need to do that – instead you can use the `toPhpString` method, which takes version numbers

Yeah but that only controls the signature return type. It means that I won't ever be able to remove the phpdoc return type in php8 if I do that, right?

muglug:
Yes, sorry you're totally right

PR #4642, opened 2020-11-21
Fix whitespace in help output

dq5studios:
The help output needs an extra newline at the end so the shell prompt clears properly.


muglug:
Thanks!

Issue #4023, opened 2020-08-18
Psalter adds "@return array-key[]" instead of "@return int[]|string[]" for MissingReturnType


TysonAndre:
For `return array_keys($array);`. array-key is specific to psalm as far as I know.

Issue #4643, opened 2020-11-21
Psalter and Union types


orklah:
Hi,

Union Types in PHP8 allow us to change Psalter so it can add more types in signature than before.

Currently, the only Union Type than can be added in signature is a Nullable Type:
https://github.com/vimeo/psalm/blob/master/src/Psalm/Type/Union.php#L480

I can try to propose a PR but I think we should define some arbitrary(or not) rules to avoid messes.

Here are the rules I think of, could you tell me if you agree with those or not and if I forgot some?

Don't add the Union in signature if:
- There's a void type in the union => forbidden by RFC
- There's two or more objects, one is named and the other one is just "object" => forbidden by RFC
- There's one type in the union who returns no php type => This would break code
- There's a mixed type in the union => This adds no meaning
- There's more than 3 different types in the result signature (null included). This means string|numeric-string|class-string|callable-string will pass => to avoid mess in signature
- There's a mix of scalar and other types in the Union => to avoid mess in signature

Thanks

orklah:
Note: Types like ArrayKey will be an issue as they represent two types by themselves. This means if it's nullable, we can't just do `?<<result>>` because we will get `?int|string` and this syntax is forbidden by RFC.
Ideally, we would transform this into `null|int|string` but currently, the role of adding the "?" is left to Union and it doesn't know what each part will return.

orklah:
Note: TFalse could also be a special case given that `string|false` is allowed. However, this will have to be handled by Union because `false` and `false|null` are not allowed by themselves

muglug:
I've done a basic thing (that doesn’t allow `string|false`) which we can refine

PR #4645, opened 2020-11-21
use int|string in phpdoc format for array-key

orklah:
This PR aims to fix https://github.com/vimeo/psalm/issues/4023

This is kind of a special case as array-key is in itself an union of two types. This means I had to add parenthesis to avoid errors in type construction. (For example, without them, we could get something like `int|string[]` to represent `array<array-key>`)

I also considered looking at $php_major_version to return int|string directly in return type but was confronted to an issue with nullable types: https://github.com/vimeo/psalm/issues/4643#issuecomment-731580141


muglug:
Thanks!

PR #4651, opened 2020-11-21
avoid formating useless type

orklah:
This is a small PR to change a sub-optimal condition I encountered when reading the code.

The `$type_string` is first assigned with the return of a formatting function and possibly overrided by a litteral just below.


muglug:
Great, thanks!

PR #4652, opened 2020-11-21
return string in phpdoc for a literal class-string

orklah:
I don't think classname::class is valid phpdoc. It's not recognized by phpstorm either.

This PR format it as `string` in the `@return`. It will have no impact on `@psalm-return`


muglug:
👍

Issue #3298, opened 2020-05-04
Incoherent types between two similar operations


orklah:
Hi,

I found a weird thing today after Psalter changed a return type in one of my function.

Consider the following example:
https://psalm.dev/r/dcd7f53538

Here, `$a` and `$b` should behave the same but they don't. One is inferred to null and the other one to mixed

I kinda understand why it can't be inferred to float (the correct behaviour in PHP) , but the fact that one is inferred to null is a big issue when using psalter (even in safe mode): 

If the function returns $a, the type will be `int|mixed` and added in phpdoc (which is almost correct)
If the function returns $b, the type will be `?int` and added in signature

If strict_type is not enabled, Psalter will slightly change the behaviour of the function by returning a rounded int instead of a float

PR #4655, opened 2020-11-22
null operations should return mixed results

orklah:
This PR should fix https://github.com/vimeo/psalm/issues/3298


muglug:
Thanks!

Issue #4235, opened 2020-09-22
Returning an object on a string typed method


orklah:
https://psalm.dev/r/b1b361febc

This should be considered as an ImplicitToStringCast.

The risk here is that if we want to put the file in strict_types=1, this will fail with a TypeError.

Example: 
https://github.com/vimeo/psalm/blob/master/src/Psalm/Codebase.php#L841

orklah:
While we're on that type of issue:
https://psalm.dev/r/bae3e0422b

This should be one too. For example, It happens here:
https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/Statements/Expression/Call/MethodCallAnalyzer.php#L232
`$result->non_existent_magic_method_ids` is documented as array<string> but it contains class with __ToString

Maybe this should be reviewed in relation with the Stringable interface that comes in PHP 8?

weirdan:
Original issue is now flagged. @orklah do you want to keep this open?

orklah:
@weirdan Yes please, I'd like to revisit the second issue later. It seemed complicated earlier

muglug:
Fixed

PR #4657, opened 2020-11-22
Emit an issue when returning a Stringable object when a string is expected

orklah:
This PR fix partially https://github.com/vimeo/psalm/issues/4235

It emits an ImplicitToStringCast when returning a Stringable object instead of a string to a method.

The other part of https://github.com/vimeo/psalm/issues/4235 was to do the same on property assignment but it seems more complicated.


muglug:
Thanks!

Issue #4273, opened 2020-10-02
Error in Psalter when manipulating an ignored file


orklah:
Once I understood what caused the error of #4272, I fixed my code of every occurence of this pattern so I could run psalter correctly.

Unfortunately, it still failed with this error but on a file I had in my ignore files in my config.

I think this means Psalm was trying to manipulate the properties in a file it's supposed to ignore. (I have never noted any actual change in an ignored file so Psalm must discard them at some point)

It made me think of #4265. Maybe there's an issue with the scopes of analysis?

muglug:
Was the ignored file inherited by a file that is in scope?

orklah:
Oh, good catch. Indeed it was.

I get why Psalm had to analyse it now. However, this was in a directory similar to vendor. The fact that it contains code that can't be manipulated shouldn't lead to a crash in Psalm

PR #4658, opened 2020-11-22
Multiple inline properties: only crash Psalm when the faulty file is in project

orklah:
This PR should fix https://github.com/vimeo/psalm/issues/4273. It checks that the faulty file is in the project. If not, it just return and do nothing


muglug:
Thanks!

PR #4659, opened 2020-11-22
Add some more flows

LukasReschke:
- Adds all relevant JSON flows https://www.php.net/manual/en/ref.json.php
- Adds all relevant misc flows https://www.php.net/manual/en/ref.misc.php
- Adds a small subset of URL flows https://www.php.net/manual/en/ref.url.php
- Added some file and SSRF sinks


muglug:
Thanks!

PR #4660, opened 2020-11-22
Prevent illegal array keys

orklah:
This PR propose emitting issue when trying to create an illegal array offset. It address this comment: https://github.com/vimeo/psalm/pull/4237#issuecomment-696890651

This is not finished yet. I think I left aside the "Mixed" case and the "Stringable object" case at least, but I wanted to run the full CI and get @muglug approval on this before working too long on it


muglug:
Thanks!

Issue #4674, opened 2020-11-23
file_exists is considered impure


tux-rampage:
Is a file existence check considered to be impure? This might be true, but it seems to be inconsistent, since other functions based on FS-State (like: `is_readable`) are not considered impure:

https://psalm.dev/r/4986ddd590

weirdan:
I believe fs-related functions should be impure.

muglug:
Oops, committed the wrong change

PR #4676, opened 2020-11-23
`is_file` and `is_dir` should be impure

staabm:
similar to https://github.com/vimeo/psalm/issues/4674

https://psalm.dev/r/2ec0c4323a


staabm:
separate issue: what about method calls on [FileInfo objects](https://www.php.net/manual/en/class.finfo.php)

muglug:
> what about method calls

All method calls are considered impure

PR #4682, opened 2020-11-23
Fix PDOStatement::fetchObject() return type

BenMorel:
According to [the manual](https://www.php.net/manual/en/pdostatement.fetchobject.php):

> Returns an instance of the required class with property names that correspond to the column names or FALSE on failure.

So: `object|false`


muglug:
Thanks!

PR #4683, opened 2020-11-23
Fix PdoStatementReturnTypeProvider

### LINKED ISSUE #4382: Scalar does not contain null with PDOStatement::fetch()
I have the following code:

```php
$statement = $this->pdo->prepare(
    'SELECT activity_id, parent_id, replyCount FROM Comment WHERE id = ? FOR UPDATE'
);

$statement->execute([$commentId]);
$result = $statement->fetch(\PDO::FETCH_ASSOC);

if ($result === false) {
    throw new \DomainException('The parent comment does not exist.');
}

$commentId = $result['parent_id'];

if ($commentId === null) {
    break;
}
```

Psalm complains on the last `if()`:

> ERROR: TypeDoesNotContainNull - src/Domain/Service/CommentService.php:95:21 - scalar does not contain null (see https://psalm.dev/090)
>                 if ($commentId === null) {

Looks like it assumes that `PDOStatement::fetch()` returns `scalar[]` (*pseudo-type, I don't know how Psalm represents types internally*), when it should be `(scalar|null)[]`.

BenMorel:
Methods returning scalars may return `null` as well.

May fix #4382, although at first glance it looks like this provider only applies to `PDOStatement::fetch()`, unless I missed something?

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php
@@ -37,20 +37,26 @@ class PdoStatementReturnTypeProvider implements \Psalm\Plugin\Hook\MethodReturnT
             $fetch_mode = $first_arg_type->getSingleIntLiteral()->value;
 
             switch ($fetch_mode) {
-                case \PDO::FETCH_ASSOC: // array<string,scalar>|false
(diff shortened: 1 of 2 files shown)


muglug:
Mind fixing the tests?

BenMorel:
I'm on it. I was on laptop battery and couldn't afford 20 mins of burning CPU, so I delegated tests to CI!

muglug:
Thanks!

muglug:
Draft PR mode is also good for that (though finicky in the UI)

Issue #4684, opened 2020-11-23
How to: dynamically suppress specific issue type based on inheritance


ludofleury:
Hi there, 

I'm struggling since hours to achieve this.
I tried almost every hooks on the plugin interface.

Here's my use-case:

If an interface "MyInterface" is implemented by a class, I want to disable "PropertyNotSetInConstructor" issue.

I know that I can suppress them with xml config or annotation (note that class annotation "@psalm-suppress" is not working in 4.X-dev). I do not want to go this way, is there a "plugin" way ?

Right now I'm so desperate that the only solution after hours is: modifying the code of `Psalm\Internal\Analyzer\ClassAnalyzer` L1492

orklah:
Did you try suppressing the issue through XML config?
```xml
    <issueHandlers>
        <PropertyNotSetInConstructor>
            <errorLevel type="suppress">
                <file name="path_to_file1"/>
            </errorLevel>
        </PropertyNotSetInConstructor>
    </issueHandlers>
```

ludofleury:
It does work with the XML. This i not what i'm trying to achieve. 
In order to ease maintainability & still enforce a standard, I'd like to dynamically add suppression rule based in inheritance :)

orklah:
> It does work with the XML. This is not what i'm trying to achieve.
> In order to ease maintainability & still enforce a standard, I'd like to dynamically add suppression rule based in inheritance :)

If I understand correctly, you don't want to list the files in the xml because this would imply changing the config on each new class?

Did you consider changing the code to allow ignoring a specific implemented class in XML config instead of files?

Something that would look like:
```xml
    <issueHandlers>
        <PropertyNotSetInConstructor>
            <errorLevel type="suppress">
                <inheritedClass name="\Name\Space\Of\Interface"/>
            </errorLevel>
        </PropertyNotSetInConstructor>
    </issueHandlers>
```

ludofleury:
I tried:

`AfterClassLikeVisitInterface`: sadly here the class info are partially populated (inheritance/interface are not available)

`afterStatementAnalysis`: with `StatementsSource->->addSuppressedIssues()`, the doc on `addSuppressedIssues()` is not clear itself, I tried an array of short class name or an array of issue FQCN. yet it seems "too late" to suppress as the CLI progress is reporting an "E" and the errors is still there on the final output

`afterAnalysis`: with parsing the issue and `unset`, seems brutal but didn't work.

ludofleury:
[...]

PR #4686, opened 2020-11-23
Plugin example: dynamic class suppression based on interface implementation

ludofleury:
@see https://github.com/vimeo/psalm/issues/4684

### DIFF
--- a/examples/plugins/DynamicallySuppressClassIssueBasedOnInterface.php
+++ b/examples/plugins/DynamicallySuppressClassIssueBasedOnInterface.php
@@ -1,0 +1,35 @@
+<?php
+namespace EventSourcing\Psalm;
+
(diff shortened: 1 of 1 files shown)


ludofleury:
Do we fix example ? :)

muglug:
Thanks, I've added this example to documentation here: https://github.com/vimeo/psalm/blob/master/docs/running_psalm/dealing_with_code_issues.md

muglug:
I don't think it's necessary to have the actual code in the `examples` directory

weirdan:
Using reflection is an antipattern in SA context though. There are `Codebase::classExists()` and `Codebase::classImplements()` methods that do not execute actual application code, unlike `class_exists()` and `ReflectionClass`.

Issue #4685, opened 2020-11-23
PdoStatement fetchObject with class


staabm:
Would be great psalm would be able to infer the class of the object returned by PdoStatement->fetchObject in https://psalm.dev/r/6431cc5c64

Dont expect the error

> INFO: LessSpecificReturnStatement - 13:14 - The type 'false|object' is more general than the declared return type 'User' for User::load_by_id

muglug:
Create a stub!

PR #4693, opened 2020-11-24
Added PDOStatement->fetchObject() stub

clxmstaab:
adding a PDO Stub as mentioned in the referenced issue.

closes https://github.com/vimeo/psalm/issues/4685

This turns unexpected error message of https://psalm.dev/r/6431cc5c64

from

> LessSpecificReturnStatement - 13:14 - The type 'false|object' is more general than the declared return type 'User' for User::load_by_id

into

> InvalidReturnStatement - test.php:16:16 - The inferred type 'User::class|false' does not match the declared return type 'User|false' for User::load_by_id (see https://psalm.dev/128)

so its obvious the new stubs are picked up, but it seems psalm is still not able to match the types as expected.
it doesn't realize that `User::class` is the same as `self` in this context?

my testfile:

```
<?php

class User {
    // Private properties
    private $id, $name;

    private function __construct () {}

    /**
     * @return false|self
     */
    public static function load_by_id ($id) {
        $pdo = new pdo("my-dsn");
        $stmt = $pdo->prepare('SELECT id, name FROM users WHERE id=?');
        $stmt->execute([$id]);
        return $stmt->fetchObject(__CLASS__);
    }

}

$user = User::load_by_id(1);
/** @psalm-trace $user */
```

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -1848,7 +1848,7 @@ class Config
             $xdebug_stub_path = dirname(__DIR__, 2) . '/stubs/Xdebug.php';
 
             if (!file_exists($xdebug_stub_path)) {
-                throw new \UnexpectedValueException('Cannot locate XDebug stub');
+                throw new \UnexpectedValueException('Cannot locate Xdebug stub');
             }
 
             $stub_files[] = $xdebug_stub_path;
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -1822,6 +1822,16 @@ class Config
             $core_generic_files[] = $stringable_path;
         }
 
+        if (class_exists('PDO', false)) {
+            $ext_pdo_path = dirname(__DIR__, 2) . '/stubs/pdo.php';
+
+            if (!file_exists($ext_pdo_path)) {
+                throw new \UnexpectedValueException('Cannot locate pdo classes');
+            }
+
+            $core_generic_files[] = $ext_pdo_path;
+        }
+
         if (\extension_loaded('ds')) {
             $ext_ds_path = dirname(__DIR__, 2) . '/stubs/ext-ds.php';
 
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -1822,7 +1822,7 @@ class Config
             $core_generic_files[] = $stringable_path;
         }
 
-        if (class_exists('PDO', false)) {
+        if (\extension_loaded('PDO')) {
             $ext_pdo_path = dirname(__DIR__, 2) . '/stubs/pdo.php';
 
             if (!file_exists($ext_pdo_path)) {
--- a/stubs/pdo.php
+++ b/stubs/pdo.php
@@ -1,0 +1,10 @@
+<?php
+
+class PdoStatement {
+    /**
+     * @template T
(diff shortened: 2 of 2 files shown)


orklah:
I think this is an issue with the difference between `User` as the type that represent an instance of User Class and `User::class`, aka class-string<User> which is a string with the name of the class

Your stub should take that into account, you receive a class-string and return the actuel object

staabm:
Good catch. I will try this tomorrow. 

Thx

clxmstaab:
the suggestion from orklah works like a charm.

everything looks like expected now:

```
 cat test.php
<?php

class User {
    // Private properties
    private $id, $name;

    private function __construct () {}

    /**
     * @return false|self
     */
    public static function load_by_id ($id) {
        $pdo = new pdo("my-dsn");
        $stmt = $pdo->prepare('SELECT id, name FROM users WHERE id=?');
        $stmt->execute([$id]);
        return $stmt->fetchObject(__CLASS__);
    }

}

$user = User::load_by_id(1);
/** @psalm-trace $user */


php psalm test.php
Scanning files...
Analyzing files...

E

ERROR: MissingPropertyType - test.php:5:13 - Property User::$id does not have a declared type (see https://psalm.dev/045)
    private $id, $name;


ERROR: MissingParamType - test.php:12:40 - Parameter $id has no provided type (see https://psalm.dev/154)
    public static function load_by_id ($id) {


ERROR: UnusedVariable - test.php:21:1 - $user is never referenced or the value is not used (see https://psalm.dev/077)
$user = User::load_by_id(1);


ERROR: Trace - test.php:22:0 - $user: User|false (see https://psalm.dev/224)
/** @psalm-trace $user */


4 errors found
Psalm can automatically fix 1 of these issues.
Run Psalm again with
--alter --issues=UnusedVariable --dry-run
to see what it can fix.

Checks took 1.74 seconds and used 160.969MB of memory

```

the return type of `fetchObject` is correctly infered.

src/Psalm/Config.php
```diff
@@ -1838,7 +1848,7 @@ public function visitStubFiles(Codebase $codebase, ?Progress $progress = null):
             $xdebug_stub_path = dirname(__DIR__, 2) . '/stubs/Xdebug.php';
 
             if (!file_exists($xdebug_stub_path)) {
-                throw new \UnexpectedValueException('Cannot locate XDebug stub');
+                throw new \UnexpectedValueException('Cannot locate Xdebug stub');
```
clxmstaab, line 1851: drive-by cleanup: the official name is Xdebug, see http://xdebug.org/

muglug:
Thanks, this is great!

stubs/pdo.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+class PdoStatement {
+    /**
+     * @template T
+     * @param class-string<T> $class
+     * @return false|T
+     */
+    public function fetchObject($class = "stdclass") {}
```
staabm, line 9: @muglug should parameters like this one, which make the php engine create a object based on a userland class-name, be considered a taint-sink?

staabm: Just went ahead and proposed a change: https://github.com/vimeo/psalm/pull/4701

PR #4694, opened 2020-11-24
restore str_split as a list

orklah:
As I reported here: https://github.com/vimeo/psalm/pull/4591#issuecomment-732769280 there were at least 3 false positives introduced in the PR.

I saw you fixed two of them earlier, this PR should fix the third one


muglug:
Thanks!

Issue #1767, opened 2019-06-10
Emit RedundantCondition for `(int)$int` when type is not from phpdoc?


TysonAndre:
This may be useful for checking for code that's no longer necessary after adding types elsewhere.

```
<?php
function testcast (int $i) : int {
    return (int)$i;  // does not emit RedundantCondition
}
```

PR #4695, opened 2020-11-24
Detect redundant cast

orklah:
This PR intends to fix https://github.com/vimeo/psalm/issues/1767

However, when I look at issues emitted in Psalm itself, I'm surprised to see this for example:
```
ERROR: RedundantConditionGivenDocblockType - src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php:1505:38 - Redundant cast to string (see https://psalm.dev/156)
                    $resolved_name = (string) $context->self;

```
This is a cast to string from a property defined as:
```
   /**
     * @var string|null
     */
    public $self;
```

This cast seems valid. That means, the detection of Union::isString() seems broken and don't see that the string is nullable. But I don't get it, it's used many times in the code and doesn't seem to cause issues. Am I missing something?

### DIFF
--- a/src/Psalm/DocComment.php
+++ b/src/Psalm/DocComment.php
@@ -104,7 +104,7 @@ class DocComment
                         $special[$type] = [];
                     }
 
-                    $line_number = $line_map && isset($line_map[$full_match]) ? $line_map[$full_match] : (int)$m;
+                    $line_number = $line_map && isset($line_map[$full_match]) ? $line_map[$full_match] : $m;
 
                     $special[$type][$line_number] = rtrim($data);
                 }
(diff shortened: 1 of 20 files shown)


muglug:
Mind fixing the other valid redundant casts in this same PR in the meantime?

orklah:
> Mind fixing the other valid redundant casts in this same PR in the meantime?

Done. Seems I have a case with nullable bool too.

orklah:
I'm a moron. Psalm saw perfectly well that null couldn't happen via inference. I was focused on the declared type I didn't look at the code itself. The three left cast are indeed redundant

orklah:
This now fails on test-with-real-projects. Should I try to PR muglug/phpunit and fix those? Or maybe just ignore them?

muglug:
This is great – I'll fix the PHPUnit baseline myself. Thanks!

mr-feek:
This is awesome @orklah 👏

orklah:
Thanks!

PR #4701, opened 2020-11-25
Declare PdoStatement->fetchObject($class_name) a taint sink

staabm:
IMO a string used by the engine to construct a object should never be end-user-controlled

### DIFF
--- a/stubs/pdo.php
+++ b/stubs/pdo.php
@@ -2,6 +2,8 @@
 
 class PdoStatement {
     /**
+     * @psalm-taint-sink text $class
+     *
      * @template T
      * @param class-string<T> $class
      * @return false|T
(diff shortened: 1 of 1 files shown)


stubs/pdo.php
```diff
@@ -2,6 +2,8 @@
 
 class PdoStatement {
     /**
+     * @psalm-taint-sink text $class
```
staabm, line 5: This change is not tested.. I guess this is how these taint-annotations work..?

muglug:
Thanks! `callable` is more appropriate

staabm:
> Thanks! `callable` is more appropriate

Thanks! Would it make sense to apply this sink automatically  to all php-src function which take a class-string typed parameter?

Issue #4698, opened 2020-11-25
DuplicateArrayKey with very large numeric string keys


gdejong:
https://psalm.dev/r/dc22a5ec25

I would not expect a `DuplicateArrayKey` error since the keys are different. It seems to have something to do with `PHP_INT_MAX`. Using numbers less than `PHP_INT_MAX` gives no error.

```
php > echo PHP_INT_MAX;
9223372036854775807
```

orklah:
It may be related to the fact that PHP cast string to int when possible, but it can't happen for literals that are higher than PHP_INT_MAX.

Psalm seem stuck to this value anyway:
https://psalm.dev/r/232eda4ed0
Here we don't use `9223372036854775807` at all, but Psalm still emit that this key is present twice...

gdejong:
@orklah indeed you are right:

From https://www.php.net/manual/en/language.types.array.php

```
Strings containing valid decimal ints, unless the number is preceded by a + sign, will be cast to the int type. E.g. the key "8" will actually be stored under 8. On the other hand "08" will not be cast, as it isn't a valid decimal integer.
```

A `var_dump` reveals this behavior:
```
array(2) {
  [9223372036854775807] =>
  int(1)
  '9223372036854775808' =>
  int(2)
}
```

orklah:
The root cause was that castring a string that contains an int > PHP_INT_MAX will "reduce" its value to PHP_INT_MAX:
https://3v4l.org/sOaGd.

I proposed a PR to avoid changing the value to int (PHP does too: https://3v4l.org/M0Ssb)

gdejong:
Hi thanks for having a look!

I am not sure if this fix actually solved the issue though:
https://psalm.dev/r/232eda4ed0

On commit 01ceaf7 the DuplicateArrayKey error still shows up.

gdejong:
I think the fix does not work since casting to an integer has an upper boundary: https://3v4l.org/XAuN2

A possible solution could be to use [bccomp](https://www.php.net/manual/en/function.bccomp.php) like: https://3v4l.org/eUJrA

orklah:
> 
> 
> I think the fix does not work since casting to an integer has an upper boundary: https://3v4l.org/XAuN2
> 
> A possible solution could be to use [bccomp](https://www.php.net/manual/en/function.bccomp.php) like: https://3v4l.org/eUJrA

[...]

PR #4702, opened 2020-11-25
fix wrong cast to int when string offset is a number > MAX INT

orklah:
This PR is an attempt to fix https://github.com/vimeo/psalm/issues/4698


muglug:
Thanks!

PR #4707, opened 2020-11-26
Fix PHPMAXINT offset

orklah:
I messed up in https://github.com/vimeo/psalm/pull/4702. I must have left cache active while running my test file locally but the fix was not correct. Plus, I didn't add a unit test for it. This one should be better.

It fixes https://github.com/vimeo/psalm/issues/4698 once again :)


muglug:
Thanks!

Issue #4359, opened 2020-10-17
class-string using templates incorrectly reports an error.


kapitancho:
https://psalm.dev/r/246da1ff09

There is a template defined for a class and a class-string property that uses the template (e.g. class-string<A>). The constructor receives a parameter of the same type (class-string<A>) but a property type coercion error is being reported when trying to assign to the class property:

ERROR: PropertyTypeCoercion - 18:17 - $this->test expects 'class-string<A:Test as mixed>',  parent type 'class-string<A:Test as object>' provided.

I don't think this makes any sense. There should be no error in this scenario.

orklah:
This should help:
https://psalm.dev/r/2266304964

kapitancho:
The addition of "... of object" helps there but it introduces another problem:
https://psalm.dev/r/1f0808491f

ERROR: RedundantConditionGivenDocblockType - 23:10 - Docblock-defined type object for $result is never falsy
even though A|null COULD be falsy.

orklah:
Yep, this seems like a bug. Psalm sees object|null but considers it can't be falsy:
https://psalm.dev/r/8011c03fd0

orklah:
reduced a little, it even works with false or string '0' :
https://psalm.dev/r/f321b26f76

Seems like any type in a Union with a template of object will be considered truthy

PR #4711, opened 2020-11-26
Allow <false|T as object> to be falsy

orklah:
This should fix https://github.com/vimeo/psalm/issues/4359

The issue was that `reconcileFalsyOrEmpty` is recursive in case of templates. Unfortunately, reconciling the result of the template with the assertion can emit an issue by itself. I made it so recursive calls can't emit issues anymore.

Note: I wanted to add a test for that, but I'm not sure what notation should be used for representing templates in the tests


muglug:
Thanks!

PR #4712, opened 2020-11-26
better fix for reconciling iterable and object

orklah:
As @weirdan noted, an iterable without array becomes a Traversable, not ArrayAccess


muglug:
Thanks!

Issue #4285, opened 2020-10-06
False positive array_multisort for list<int> in 3rd argument


ostrolucky:
https://psalm.dev/r/8242e51491

orklah:
This seems to be a legitimate issue but Psalm is very bad at displaying it.

The fact that your type $foos has it's offset y optional makes Psalm infer the resulting type as a `list<int>` with the flag `possibly_undefined`. I think this would be better to directly makes an optional key translate to null in the result of array_column but I will need to perform some tests before that

PR #4718, opened 2020-11-27
When the result of array_column is single, the union is redundant

orklah:
This should solve https://github.com/vimeo/psalm/issues/4285

I'm actually not sure that the provided code was totally correct given that the 'y' offset is optional. I'm not totally sure what the result of array_column is in this case.

But this fixes the reported issue. The first test is redundant given that even with the extra Union, the type was already displayed as `list<int>` but it's just in case.


orklah:
ohh, my analysis was completely wrong.

PR #4719, opened 2020-11-27
fix array_column with possibly_undefined keys

orklah:
This PR should really fix https://github.com/vimeo/psalm/issues/4285 this time!

The issue was that the provider for array_column didn't take into account the fact that the key could be possibly_undefined.

This allowed two bugs:
- The resulting type of array_column on a possibly_undefined key was a list<int> but with the possibly_undefined flag so it was not accepted by functions expecting a correct list
- When the initial array was a non-empty array, there was a special case to return a non-empty result. This is not always the case on possibly_undefined columns because if they're all undefined, it returns an empty result


muglug:
Thanks!

PR #4724, opened 2020-11-28
fix #3282

### LINKED ISSUE #3282: MissingImmutableAnnotation wording choice could be `@psalm-immutable`, not `immutable`
Originally, the annotation was `@immutable`, but was changed to `@psalm-immutable` due to other tools likely ending up defining `@immutable` in a different way until an official phpdoc standard was chosen.

(People may forget that the annotation is `@psalm-immutable`, not `@immutable`)

```
Current: MissingImmutableAnnotation: x is marked immutable, but MySubclass is not marked immutable
Proposed: MissingImmutableAnnotation: x is marked @psalm-immutable, but MySubclass is not marked @psalm-immutable
```

orklah:
Simple change of wording to fix https://github.com/vimeo/psalm/issues/3282

### DIFF
--- a/docs/running_psalm/issues/MutableDependency.md
+++ b/docs/running_psalm/issues/MutableDependency.md
@@ -1,6 +1,6 @@
 # MutableDependency
 
-Emitted when an immutable class inherits from a class or trait not marked immutable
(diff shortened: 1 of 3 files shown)


muglug:
Thanks!

Issue #3349, opened 2020-05-11
False positive MixedReturnTypeCoercion?


gharlan:
https://psalm.dev/r/5d6f32139d

The error is new since 3.11.3 or 3.11.4.

weirdan:
The error reported looks valid. The easiest way to address this given currently available Psalm features would be to add `/** @var Traversable<string, SplFileInfo> */` to the return statement: https://psalm.dev/r/18e3445436

To properly understand the shape of the iterator Psalm needs to be able to infer key and value types from the second argument of `RecursiveDirectoryIterator` constructor.

orklah:
For me, it looks like a missing stub for `RecursiveCallbackFilterIterator`. I guess if this stub was added, it would contain the `@template-implements` tags necessary to makes Psalm understand that `RecursiveCallbackFilterIterator` is indeed a Traversable with those templates.

orklah:
The original issue was kinda fixed by #4725 except that this falls back into the trap of standard iterators that return different types depending on constructor flags.

PR #4725, opened 2020-11-28
add stubs for standard iterators

orklah:
This is a PR that aims to complete the list of standard iterators listed here:
https://www.php.net/manual/fr/spl.iterators.php

I separated the iterators from the classes file for clarity. I probably made mistakes and I'm not done yet so I create this as a draft to run the full CI


orklah:
Well, I think this is finished. At least, that's all I'm willing to do, this is really tedious work.

Let's be honest, this is far from perfect. I'm sure there's some obvious errors in there but it's probably better than nothing. (And I'm not sure this can ever be perfect. For example https://www.php.net/manual/fr/filesystemiterator.current.php#refsect1-filesystemiterator.current-returnvalues FilesystemIterator will change it's template type depending on flags in the constructor :( )

For example, when I run the code from https://github.com/vimeo/psalm/issues/3349. It solves the error. The inferred type is still wrong because it can't infer types in templates but it's now a Traversable so it's better.

The risk here will be broken build because the old suppressed error due to missing stub will be replaced with an other error due to wrong/incomplete stub.

Maybe this will encourage people to submit their issues or fix it themselves?

weirdan:
@orklah I'd like to go over this to make sure we don't accidentally cause too much noise. It might be tedious, but dealing with a flurry of issues is even more so.

orklah:
> 
> 
> @orklah I'd like to go over this to make sure we don't accidentally cause too much noise. It might be tedious, but dealing with a flurry of issues is even more so.

Be my guest 👍 

I guess a good way to progress with this would be to have example of code using those to see how psalm behaves. But finding those would be difficult.

staabm:
> I guess a good way to progress with this would be to have example of code using those to see how psalm behaves. But finding those would be difficult. 

Symfony uses a lot of iterators.. https://github.com/symfony/symfony/search?q=iterator maybe this is useful

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+
+    public function beginChildren(): void {}
+    public function beginIteration(): RecursiveIterator {}
+    public function callGetChildren(): RecursiveIterator {}
+    public function callHasChildren(): bool {}
+
+    /**
+     * @param RecursiveIterator|IteratorAggregate $it
+     */
+    public function __construct($it, int $flags = RecursiveTreeIterator::BYPASS_KEY, int $cit_flags = CachingIterator::CATCH_GET_CHILD, int $mode = RecursiveIteratorIterator::SELF_FIRST) {}
```
weirdan, line 823:
```suggestion
    /**
     * @param RecursiveIterator<TKey, TValue>|IteratorAggregate<TKey, TValue> $it
     * @param int-mask<self::BYPASS_CURRENT, self::BYPASS_KEY> $flags
     * @param int-mask<self::CATCH_GET_CHILD> $cit_flags
     * @param self::LEAVES_ONLY|self::SELF_FIRST|self::CHILD_FIRST $mode
     */
    public function __construct($it, int $flags = self::BYPASS_KEY, int $cit_flags = self::CATCH_GET_CHILD, int $mode = self::SELF_FIRST) {}
```

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+    /**
+     * @return TValue Can return any type.
+     */
+    public function current(): string {}
```
weirdan, line 827: It actually returns string
```suggestion
    public function current(): string {}
```

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+    public function endChildren(): void {}
+    public function endIteration(): void {}
+    public function getEntry(): string {}
+    public function getPostfix(): string {}
+    public function getPrefix(): string {}
+    /**
+     * @return TKey scalar on success, or null on failure.
+     */
+    public function key(): string {}
```
weirdan, line 836: It actually returns string
```suggestion
    public function key(): string {}
```

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+    public function next(): void {}
+    public function nextElement(): void {}
+    public function rewind(): void {}
+    public function setPostfix(string $postfix ): void {}
+    public function setPrefixPart(int $part , string $value ): void {}
```
weirdan, line 841:
```suggestion
    /**
     * @param self::PREFIX_* $part
     * @param string $value
     */
    public function setPrefixPart(int $part , string $value ): void {}
```

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+ */
+class RegexIterator extends FilterIterator {
+    const MATCH = 0 ;
+    const GET_MATCH = 1 ;
+    const ALL_MATCHES = 2 ;
+    const SPLIT = 3 ;
+    const REPLACE = 4 ;
+    const USE_KEY = 1 ;
+
+    public function __construct(RecursiveIterator $iterator, string $regex, int $mode = self::MATCH, int $flags = 0, int $preg_flags = 0) {}
```
weirdan, line 859:
```suggestion
    /**
     * @param Iterator<TKey, TValue> $iterator
     * @param string $regex
     * @param self::MATCH|self::GET_MATCH|self::ALL_MATCHES|self::SPLIT|self::REPLACE $mode
     * @param int-mask<self::USE_KEY> $flags
     * @param int $preg_flags
     */
    public function __construct(Iterator $iterator, string $regex, int $mode = self::MATCH, int $flags = 0, int $preg_flags = 0) {}
```

orklah:
Oh, I totally missed your review earlier. Will look into it! Thanks!

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+     * @return TKey scalar on success, or null on failure.
+     */
+    public function key() {}
+}
+
+/**
+ * @template T as DirectoryIterator
+ * @psalm-implements SeekableIterator<T>
+ */
+class DirectoryIterator extends SplFileInfo implements SeekableIterator {
```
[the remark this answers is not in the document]
orklah: I added a template here because childrens will redefine this template to be more specific. Was it wrong?

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+ */
+class EmptyIterator implements Iterator {
+    /**
+     * @return empty
+     */
+    public function current() {}
+    /**
+     * @return empty
+     */
+    public function key() {}
```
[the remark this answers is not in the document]
orklah: oh, good to know. Not sure what the effect returning empty will have but this was the closest I could find

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+    public function next(): void {}
+    public function rewind(): void {}
+    public function valid(): bool {}
+}
+
+/**
+ * @template T as FilesystemIterator
+ * @psalm-extends SeekableIterator<T>
+ */
+class FilesystemIterator extends DirectoryIterator
```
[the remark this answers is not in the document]
orklah: Oh, I guess this answer my previous question

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+    const FOLLOW_SYMLINKS = 512;
+    const KEY_MODE_MASK = 3840;
+    const NEW_CURRENT_AND_KEY = 256;
+    const SKIP_DOTS = 4096;
+    const UNIX_PATHS = 8192;
+
+    /**
+     * @param self::KEY_AS_PATHNAME|self::CURRENT_AS_FILEINFO|self::SKIP_DOTS $flags
+     */
+    public function __construct(string $path, ?int $flags = null) {}
```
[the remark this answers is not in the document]
orklah: Oh, I see! Thanks!

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+    const CURRENT_MODE_MASK = 240 ;
+    const KEY_AS_PATHNAME = 0 ;
+    const KEY_AS_FILENAME = 256 ;
+    const FOLLOW_SYMLINKS = 512 ;
+    const KEY_MODE_MASK = 3840 ;
+    const NEW_CURRENT_AND_KEY = 256 ;
+    const SKIP_DOTS = 4096 ;
+    const UNIX_PATHS = 8192 ;
+
+    public function __construct(string $path, int $flags = 0) {}
```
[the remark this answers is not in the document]
orklah: Wow, I'm impressed by your attention to details and patience with this!

stubs/CoreGenericIterators.phpstub
```diff
@@ -0,0 +1,869 @@
...
+     * @return TKey scalar on success, or null on failure.
+     */
+    public function key(): string {}
+}
+
+/**
+ * @template TIterator as RecursiveIterator|IteratorAggregate
+ * @mixin TIterator
+ */
+class RecursiveIteratorIterator implements OuterIterator {
```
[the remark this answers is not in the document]
orklah: Yep, that was one of existing stubs, I only moved them from one file to the other without editing them. I just completed it

stubs/CoreGenericIterators.phpstub
```diff
@@ -875,11 +901,11 @@ class RecursiveRegexIterator extends RegexIterator implements RecursiveIterator
      * @param RecursiveIterator<TKey, TValue> $iterator
      * @param string $regex
      * @param self::MATH|self::GET_MATCH|self::ALL_MATCHES|self::SPLIT|self::REPLACE $mode
-     * @param int-mask-of<self::USE_KEY> $flags
+     * @param int-mask-of<self::*> $flags
```
weirdan, line 904: The only accepted flag is `USE_KEY`: https://www.php.net/manual/en/regexiterator.setflags.php
If Psalm chokes on single-valued bitmask I'd change it to `0|self::USE_KEY`.

orklah: It's doesn't work either. If only 1 value is accepted, I'll just use the literal. It should work the same.

stubs/CoreGenericIterators.phpstub
```diff
@@ -902,7 +902,7 @@ class RecursiveRegexIterator extends RegexIterator implements RecursiveIterator
      * @param RecursiveIterator<TKey, TValue> $iterator
      * @param string $regex
      * @param self::MATH|self::GET_MATCH|self::ALL_MATCHES|self::SPLIT|self::REPLACE $mode
-     * @param int-mask-of<self::0|self::USE_KEY> $flags
+     * @param 1 $flags
```
weirdan, line 905: What I meant was either
```suggestion
     * @param self::USE_KEY|0
```
or
```suggestion
     * @param 1|0
```
as `0` (no flag) is also a valid value.

orklah:
@weirdan Seems we broke RecursiveIteratorIterator. It was a mixin and the test fails. I think I'll revert this one, it was one of the existing one so I have no reason to think it didn't work already.

Other tests are failing because we return null now. Do you think we should change the tests?

weirdan:
> Other tests are failing because we return null now. Do you think we should change the tests?

Depends on whether actual code returns null. I.e. if tests are testing for a wrong behaviour we should fix them.

orklah:
> Depends on whether actual code returns null. I.e. if tests are testing for a wrong behaviour we should fix them.

That's the thing, in the happy path, devs should check is_valid before calling current() so the template should only return TKey, BUT, if they don't, null is returned.

The question is, should we document current as possibly always returning null? If so, then devs who check is_valid and have a template that don't contains null will probably get errors anyway

weirdan:
> If so, then devs who check is_valid and have a template that don't contains null will probably get errors anyway

That's what `@psalm-ignore-nullable-return` is designed to solve. `file_get_contents()` may return `false`, but Psalm doesn't force you to check for that (while *allowing* you to do so), because it treats it as `@psalm-ignore-falsable-return`: https://psalm.dev/r/f10a53c7bb

orklah:
Oh, I thought this was to allow the config ignoreNull in psalm.xml. My bad then, I'll fix the tests

muglug:
Thanks!

xPaw:
I think this PR made looping over `FilesystemIterator` think it's iterating strings:

https://psalm.dev/r/8930178275

orklah:
Well not exactly: https://psalm.dev/r/de250db5e3

string is one of the options possible for current() depending of flags given on creation. For example: https://3v4l.org/cgYb1

I'm not sure what it would take for Psalm to understand that the output of a function is conditioned by a parameter in another method (__construct in this example)

Can you open a new issue to discuss this please?

Issue #4726, opened 2020-11-28
TypeDoesNotContainType: False positive on uninitialized property


m-ober:
Consider this code with error level set to 3 (so  MissingConstructor is not treated as error).
https://psalm.dev/r/b0096d11d3

This will output "$this->i is not set.", because `isset()` returns false on checking an uninitialized property. Marking the property as nullable (`?int`) removes the error, but changes the intention of the code (i.e. `$i` must be assigned at some point, but not necessarily during construction, and `null` is not an acceptable value.)

Expected: TypeDoesNotContainType should not be triggered when calling isset on uninitialized, non-nullable properties.

orklah:
I agree strongly on that.

Our work codebase does not work by injecting dependancy so everything is done with getters and setters. Using isset on non nullable properties is actually our good practice but it's considered an issue in Psalm.

I'm not sure how to add this myself, I could try if it's not too difficult.

muglug:
I don’t see a way that this can happen given Psalm’s current assumptions.

Psalm assumes that all properties are set inside constructors. That assumption is pretty useful — it means that in any class that passes those checks, Psalm can guarantee that property access will work.

You’re asking for Psalm to lose that assumption for some set of classes, and I don’t think that’s good.

What you could maybe do, and @orklah could work on this, is add a Psalm plugin that set the `Union::$initalized` property to false for the candidate class properties.

That would mean the properties could _only_ be used after an `isset` check, though, which might not be desirable.

An extreme alternative would be a `@psalm-ignore-initialization` docblock annotation corresponding to a new `Union::$ignore_initialization` property set on each property’s type. Calling `isset` on a property with that flag wouldn’t raise an error, but you’d want to strip that flag from any newly-assigned type.

orklah:
mmh tricky.

Would it be possible to change the type of the issue from TypeDoesNotContainType to its own issue? Something like RedundantInitialisationCheck or something so a project can just ignore those?

muglug:
Yeah, you could set a `Union::$from_property` flag to `true` for all non-nullable properties here: https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/ClassAnalyzer.php#L1041

Then you’d treat that like `from_docblock` when handling the `isset` issue.

But the same caveat about stripping that flag when assigning to another type applies.

PR #4727, opened 2020-11-28
Handling isset on non-nullable properties

orklah:
Following discussion on https://github.com/vimeo/psalm/issues/4726#issuecomment-735253394 about isset on non-nullable properties to check initialisation, I made this small PR just to show what I had in mind instead of trying to explain it badly

This successfully remove the issue and it doesn't seem like a big deal. Maybe we could couple this with a config "allowIssetChecksOnNonNullableProperties" for old projects? Or we could emit another Issue type in this case, just for easy suppression for old projects

What do you think about this?

### DIFF
--- a/src/Psalm/Internal/Type/NegatedAssertionReconciler.php
+++ b/src/Psalm/Internal/Type/NegatedAssertionReconciler.php
@@ -89,29 +89,31 @@ class NegatedAssertionReconciler extends Reconciler
                             $failed_reconciliation = 2;
 
                             if ($code_location) {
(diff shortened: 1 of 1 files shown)


muglug:
I prefer the flag come from the type TBH — that way it can also be used outside the class that defines the property

orklah:
Oh, I see. I'll try that

orklah:
I'll submit a new PR once I progress on this

muglug:
It needs some work, but here's an attempt: https://github.com/vimeo/psalm/compare/property-flags

PR #4734, opened 2020-11-29
add documentation for RedundantPropertyInitializationCheck

orklah:
This PR adds documentation for RedundantPropertyInitializationCheck

Thanks for adding this :)

I started looking at what you did, it's interesting to look at, I thought reinitializing the from_property would have to be dispersed everywhere, but it's quite localized in assignments, that's cool.

I'm not sure why you added the third param when creating a new instance of the issue. Isn't that used for custom suppression in issueHandler


muglug:
Thanks!

PR #4738, opened 2020-11-29
Enforce class-string in is_subclass_of()

BenMorel:


orklah:
While I agree for the second param $class_name, I'm more dubious about the first. That would imply checking both class_exists and is_subclass_of when starting from a simple string.

orklah:
@BenMorel care to change the first param and rebasing?

BenMorel:
@orklah Done!

weirdan:
You need to add that to `CallMap_historical` as well, I suppose.

BenMorel:
@weirdan Done!

BenMorel:
Still failing because of [this test](https://github.com/vimeo/psalm/blob/e5b797c764c8a30a3a8f9ef61a5f25dc96a950d5/tests/ClassLikeStringTest.php#L390-L408).

Not sure if we should just remove it now?

weirdan:
Yeah, it seems so. Please do.

BenMorel:
@weirdan Done!

weirdan (approved):

weirdan:
Thanks!

Issue #4739, opened 2020-11-29
Bit shifting produces mixed


BenMorel:
```php
class Options
{
    public const LOCK_READ = 1 << 0;
    public const LOCK_WRITE = 1 << 1;
}

function isRefresh(int $options): bool
{
    return ($options & Options::LOCK_READ || $options & Options::LOCK_WRITE);
}
```

> INFO: MixedOperand - 11:24 - Right operand cannot be mixed

> INFO: MixedOperand - 11:57 - Right operand cannot be mixed

https://psalm.dev/r/6d57e26364

muglug:
Fixed by @orklah

BenMorel:
Thanks, @orklah and @muglug!

PR #4740, opened 2020-11-29
Const support more ops

orklah:
This PR should solve https://github.com/vimeo/psalm/issues/4739

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
@@ -13,6 +13,9 @@ use function count;
 use function array_shift;
 use function reset;
 
+/**
+ * This class takes a statement and return its type by analyzing each part of the statement if necessary
+ */
 class SimpleTypeInferer
 {
     /**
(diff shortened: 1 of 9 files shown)


orklah:
I'm quite confused. 

First thing, seems like bitwise or is already working, but I don't see why yet, it doesn't seem to be handled where I am. Plus, I don't get why the other two bitwise operations are not supported the same way.

Second thing is that an union between TLiteralInt(0) and TPositiveInt seems to be displayed as int|positive-int. I'm pretty sure this issue was fixed not long ago (multiple int not merging), maybe this is more localized than I thought

Third thing is that I broke some attribute test. I'm not sure how

I'll continue checking those

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -109,6 +112,17 @@ public static function infer(
                 );
             }
 
+            if ($stmt instanceof PhpParser\Node\Expr\BinaryOp\ShiftLeft
+                || $stmt instanceof PhpParser\Node\Expr\BinaryOp\ShiftRight
+            ) {
+                return new Type\Union(
+                    [
+                        new Type\Atomic\TPositiveInt(),
```
weirdan, line 120: Shifts are perfectly capable to produce negative integers: https://3v4l.org/HXdvp

orklah: Thanks!

orklah:
I think this should be ok now.

I changed a lot more than I thought I would:
- Fixed arithmetic operations and tests for pow/exponent. The symbol is **, not ^
- Added shift operations and some bitwise operations to arithmetics
- Changed positive-int to int in tests. I'm pretty sure it's because those are LiteralInt being calculated in arithmetics.

muglug:
Thanks, this is awesome!

Issue #4500, opened 2020-11-06
Psalter doesn't add return type with returning catch blocks


orklah:
Hi,

For the following example:
https://psalm.dev/r/0a7dc1a7c7

If we click on fix code, Psalter only add documentation where it could add a return type in signature.

Strangely, when we add an unreachable statement after the catch, it suddenly works:
https://psalm.dev/r/8124b89d61

weirdan:
Looks quite similar to #4499 , where that could also be fixed by adding unreachable return.

orklah:
oh nice, I didn't make the connection between the two

PR #4746, opened 2020-11-30
handle return flag for a try/catch/finally

orklah:
This PR aims to resolve https://github.com/vimeo/psalm/issues/4500

However, it seems pretty crude. I probably did something wrong (or maybe even very wrong!) and I'm about to see that in CI, but it seems to resolve my issue so I take a shot!

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Block/TryAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/TryAnalyzer.php
@@ -444,6 +444,7 @@ class TryAnalyzer
             $context->loop_scope->final_actions[] = ScopeAnalyzer::ACTION_NONE;
         }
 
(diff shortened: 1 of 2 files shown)


muglug:
Mind adding a test?

muglug:
Mind adding tests that fail on current Psalm?

muglug:
You should add a test in the `tests/FileManipulations` folder, I think

orklah:
> Mind adding tests that fail on current Psalm?

https://psalm.dev/r/baa68b1aa0 didn't do the trick? Anyway, I added my original case in FileManipulations

orklah:
oh, nvm, the error was indeed already reported correctly before. Should I keep the redundant tests I added then?

muglug:
Nah, get rid of those I think

muglug:
Thanks!

PR #4757, opened 2020-12-03
Improve Imagick call map

BenMorel:
- Fixed case of `Imagick`, `ImagickDraw` and `ImagickPixel`
- Added proper types of arrays

I've been able to infer the type of most arrays from doc and/or experiment.
I've left out arrays I wasn't sure of, for now.


muglug:
Thanks!

PR #4758, opened 2020-12-03
Fix ImagickPixel::getColor()

BenMorel:
`ImagickPixel::getColor()` takes an int, not a bool:

https://www.php.net/manual/en/imagickpixel.getcolor.php
https://github.com/Imagick/imagick/blob/06116aa24b76edaf6b1693198f79e6c295eda8a9/imagickpixel_class.c#L605


muglug:
Thanks!

PR #4761, opened 2020-12-03
Invalid syntax for PHP v7.2, psalm v4.3.0

SmetDenis:
Hi,

I'm trying to use psalm v4.3.0 + PHP 7.2.33 

According `composer.json`, it should work with v7.1+
 https://github.com/vimeo/psalm/blob/master/composer.json#L17 

I found invalid syntax here
https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/ExistingAtomicStaticCallAnalyzer.php#L251

`Type::getInt(false, $codebase->php_major_version),`

I guess it should be 
`Type::getInt(false, $codebase->php_major_version)`


muglug:
Thanks!

Issue #4765, opened 2020-12-03
False positive RedundantCast


enumag:
https://psalm.dev/r/dc8cad0f30

`hexdec` can return float so `hexdec($x) + 1` can also be float.

muglug:
When can `hexdec` return a float?

enumag:
I don't know but the documentation claims so. https://www.php.net/manual/en/function.hexdec.php

orklah:
>  The function can convert numbers that are too large to fit into the platforms int type, larger values are returned as float in that case.

muglug:
OH @orklah the fix is to not complain about int casts if the type has the `$from_calculation` flag set

orklah:
Ok, I'll check that

PR #4768, opened 2020-12-03
Allow int casts if the type comes from calculation

orklah:
This should fix https://github.com/vimeo/psalm/issues/4765


muglug:
Thanks!

Issue #4776, opened 2020-12-04
is_a accepts `mixed` as first param


gmazzap:
[`is_a()`](https://www.php.net/manual/en/function.is-a.php) signature says first param is `mixed`.

However, Psalm enforces it to be a object or a string.

Note that for the `instanceof` statement, which works pretty much same way, Psalm correctly accepts anything.

See https://psalm.dev/r/c456cebe2e

And see how in PHP that is fine code: https://3v4l.org/Rb2MP

websupporter:
The related XML definition of `is_a`: https://github.com/php/doc-en/blob/master/reference/classobj/functions/is-a.xml#L12

PR #4780, opened 2020-12-04
allow passing mixed in is_a

orklah:
This should fix https://github.com/vimeo/psalm/issues/4776

rationale: you may want to pass anything in is_a for example to split an union object|array ;
```php
$test = rand(1, 4) > 2 ? [] : new stdClass;

if(is_a($test, stdClass::class)) {
    echo 'object';
}
else{
    echo 'array';
}
```


muglug:
Thanks!

PR #4786, opened 2020-12-05
Don't add link if the issue doesn't have a shortcode

orklah:
This avoids the link `(see https://psalm.dev/000)` at the end of plugin issues


weirdan:
Shouldn't all issues have shortcode though?

orklah:
Not sure what the point would be for plugin issues as they won't be documented in psalm.dev. They even risk using a future psalm shortcode, which would mean that the documentation in psalm.dev would be about a different issue

weirdan:
You're right, forgot about third-party-supplied issues.

muglug:
Thanks!

PR #4789, opened 2020-12-06
Make sure uncached files are analyzed

weirdan:
Previously Psalm would ignore any files that were created before or at the same time (second) as it had recorded as the last run.

This is problematic in these situations:

 * File is created the same second Psalm starts, but after it created the list of files to check
 * File is copied/moved from somewhere into the project folder with original timestamps preserved

Now Psalm will additionally check if it actually has the file cached before ignoring it.

### DIFF
--- a/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
@@ -1076,7 +1076,7 @@ class ProjectAnalyzer
 
         foreach ($file_paths as $file_path) {
             if ($config->isInProjectDirs($file_path)) {
-                if ($this->file_provider->getModifiedTime($file_path) > $last_run
+                if ($this->file_provider->getModifiedTime($file_path) >= $last_run
                     && $this->parser_cache_provider->loadExistingFileContentsFromCache($file_path)
                         !== $this->file_provider->getContents($file_path)
                 ) {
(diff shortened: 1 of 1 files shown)


muglug:
I'm not thrilled about this – I'd prefer something like 

```
$this->file_provider->getModifiedTime($file_path) > $last_run - 1
```

to avoid the off-by-one-second issue, and to make the override of _that_ something you'd put in a config.

The reason being that looking up files in the cache is quite a bit slower, especially with large projects.

weirdan:
That check is probably not very effective for big projects anyway, at least in CI context, due to the fact that `git clone` does not preserve timestamps. For those cases something like https://github.com/MestreLion/git-tools#git-restore-mtime should probably be recommended. Composer generally sets the timestamp for the entire package to that of its last commit (by virtue of using `git archive`).

> `$this->file_provider->getModifiedTime($file_path) > $last_run - 1`

This also solves the issue I'm facing.

weirdan:
@muglug re-done as suggested.

muglug:
> That check is probably not very effective for big projects anyway at least in CI context, due to the fact that git clone does not preserve timestamps

You can probably guess what we do at Vimeo :)

muglug:
Thanks!

weirdan:
> You can probably guess what we do at Vimeo :)

Using CVS? :laughing:

PR #4790, opened 2020-12-06
Added a workflow to test on PHP 8 / Linux

weirdan:
Currently we already test with PHP 8 on AppVeyor, but the failures there may be windows-specific, so it makes sense to test on Linux as well


muglug:
Thanks!

I fixed at least one PHP 8 issue locally today.

Issue #4795, opened 2020-12-07
Attempting to register two stubs for the same global function results in hard to trace errors


mr-feek:
Psalm currently allows registering two different stubs for the same global method. The resultant behavior can be very hard to debug if you do not know that a second stub is being added.

I will take a stab at implementing. I can't think of any scenarios off the top of my head where two different stubs being registered for the same global function would be ideal?

muglug:
>  I can't think of any scenarios off the top of my head where two different stubs being registered for the same global function would be ideal?

It isn’t. What do you think Psalm should do in that scenario?

orklah:
Closing as stale

PR #4796, opened 2020-12-07
feature: prevent multiple stubs being added for the same function

mr-feek:
working on a potential fix for #4795


mr-feek:
After digging into this more, it appears that psalm doesn't differentiate between "global functions" and "stubbed global functions". We would need to keep track of both is my understanding -- will revisit when i have more time

PR #4805, opened 2020-12-08
Overwrite memoized return type after sum-type candidate is calculated

2e3s:
### Given

When analyzing an immutable call on a sum-typed variable, it memoizes only the first union part.
`MethodCallAnalyzer` analyzes a method call's result on `C<A>` and `C<B>` by calling `Method\AtomicMethodCallAnalyzer`->`ExistingAtomicMethodCallAnalyzer`. The first call finds `A` and remembers it. The second call calculates the proper type candidate `B` but it is overwritten by `A`.

I've created a new test case but `doNotCombineTypes` fails when adding `@psalm-mutation-free`. Probably the tests may be combined to reduce the excessive test.

### Expected

`A|B` is memoized fully.
I think the only way to do that is by memoizing the call after the full type is calculated.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/AtomicMethodCallAnalysisResult.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/AtomicMethodCallAnalysisResult.php
@@ -79,4 +79,9 @@ class AtomicMethodCallAnalysisResult
      * @var list<\Psalm\Internal\MethodIdentifier>
      */
     public $too_few_arguments_method_ids = [];
+
+    /**
+     * @var bool
+     */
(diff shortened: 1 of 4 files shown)


muglug:
uhh I'm confused. This works fine: https://psalm.dev/r/4a9e033f37

2e3s:
Yes, this is the old test, it passes indeed because there is no memoization for `C::get()`.
But this doesn't work: https://psalm.dev/r/554cdcc468
> ERROR: DocblockTypeContradiction - 33:9 - Cannot resolve types for $a_or_b - docblock-defined type A does not contain B

muglug:
Ah ok, sorry, now I understand – thanks!

PR #4808, opened 2020-12-08
Infer mutation-free on all non-overridden methods

2e3s:
I doubted at first to do this implicitly but then I found that mutation-free property is already inferred for final methods.
Finality is only interesting to verify against class inheritors. What the mutation-free heuristics apparently wants to know is that there are no inheritors on the method.

This would allow wider acceptance of, for instance, `PossiblyNull*` issues while avoiding bending the code for Psalm with unnecessary `final` keywords or `@psalm-mutation-free` in the contexts where it's obvious to have no side effects.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallPurityAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallPurityAnalyzer.php
@@ -80,7 +80,7 @@ class MethodCallPurityAnalyzer
         ) {
             if ($method_storage->mutation_free
                 && (!$method_storage->mutation_free_inferred
-                    || $method_storage->final)
+                    || !$method_storage->overridden_somewhere)
             ) {
                 if ($context->inside_conditional
                     && !$method_storage->assertions
--- a/tests/MethodCallTest.php
+++ b/tests/MethodCallTest.php
@@ -833,7 +833,7 @@ class MethodCallTest extends TestCase
                         printInt($obj->getInt());
                     }',
             ],
-            'unchainedInferredInferredFinalMutationFreeMethodCallMemoize' => [
+            'unchainedInferredNonOverriddenMutationFreeMethodCallMemoize' => [
                 '<?php
                     class SomeClass {
                         private ?int $int;
@@ -842,7 +842,7 @@ class MethodCallTest extends TestCase
                             $this->int = 1;
                         }
 
-                        final public function getInt(): ?int {
+                        public function getInt(): ?int {
                             return $this->int;
                         }
                     }
@@ -1381,6 +1381,11 @@ class MethodCallTest extends TestCase
                         }
 
                         public function getInt(): ?int {
+                            return $this->int;
+                        }
+                    }
+                    class Child extends SomeClass {
+                        public function getInt(): ?int { // overridden
                             return $this->int;
                         }
                     }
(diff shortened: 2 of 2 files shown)


2e3s:
All Psalm checks are failing because https://github.com/vimeo/psalm/blob/6f916553a8de3c5c2483162b6cc01b21b24e4d1d/src/Psalm/Type/Union.php#L234-L238

Yet there are a lot of checks like https://github.com/vimeo/psalm/blob/60ac109c01433073d83dddaeb9ffd3abb202b92f/src/Psalm/Internal/Type/SimpleAssertionReconciler.php#L449
It looks like a wrong signature in the docblock on `Union::getAtomicTypes()` which is discovered by this change (since now `@return non-empty-array<string, Atomic>` gets remembered for the method).

2e3s:
Fixing the signature of `Union::getAtomicTypes()` to `@return array<string, Atomic>` leads to https://github.com/vimeo/psalm/blob/a2b20f094d6df402194787b54a786ba39fd9384a/src/Psalm/Type/Reconciler.php#L152
:face_with_head_bandage:

muglug:
I don't like this change.

The allowance with `final` methods was to ensure that this memoisation presented the same basic trade-off as memoising property assignments (which Psalm does by default).

If users want full method memoisation that's a config option; if users want safer method memoisation they need to use `final`, and if they want _really_ safe memoisation they need to disable property memoisation and only use immutable classes.

muglug:
I appreciate the effort, though!

With Psalm there's always a balance between "what developers think is good code" and "what programming language purists think is good code". With the property memoisation I err on the side of the former, but with method memoisation I err on the side of the latter (but as above, both are configurable).

2e3s:
The config option for total memoisation looks quite useless because it's unconditional and a lot of guarantees are invalid. I could try to make this heuristic for non-overridden methods a plugin but `AfterMethodCallAnalysisInterface` has no argument `AtomicMethodCallAnalysisResult` where I've added the memoization result in the previous PR. As I understand, BC is sacred for the plugin interface.

> The allowance with final methods was to ensure that this memoisation presented the same basic trade-off as memoising property assignments (which Psalm does by default).

`final` and non-overridden methods present exactly the same fact for the static analysis for treating a method as a "class property." This only fact is "method is not overridden". Using only a subset of non-overridden methods marked by `final` keyword is confusing.

I think, if someone needs a safest and explicit memoisation, it's better to allow only `@psalm-mutation-free` which would be more expected and explicit. I initially thought that Psalm only uses that tag.

weirdan:
> final and non-overridden methods present exactly the same fact for the static analysis

There's a difference between 'cannot be overridden' (final) and 'is not currently known to be overridden' (your heuristic). The latter easily breaks when your class is inherited in a separate composer package that has yours as a dependency.

2e3s:
Okay, yes, I see how it can be theoretically different, although I can't imagine a real scenario where it happens without a dependency cycle (unlike unsound type of a memoized class property).

I'm now concerned about [`AfterMethodCallAnalysisInterface`](https://github.com/vimeo/psalm/blob/e46c68b1e58320c456ae2878925721be8f6db6df/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/ExistingAtomicMethodCallAnalyzer.php#L379) for a plugin :disappointed: I guess, the only choice here is to modify the incoming return type candidate of `Union` but it would disallow sum types resolution which #4805 fixes.

muglug:
I also made an update yesterday that immutable methods and properties will _always_ be memoised even when the config options are disabled, as that memoisation is always safe.

PR #4821, opened 2020-12-09
Type-hint implode() array parameter

BenMorel:


dictionaries/CallMap.php
```diff
@@ -5989,8 +5989,8 @@
 'imap_utf7_encode' => ['string', 'string'=>'string'],
 'imap_utf8' => ['string', 'mime_encoded_text'=>'string'],
 'imap_utf8_to_mutf7' => ['string|false', 'string'=>'string'],
-'implode' => ['string', 'separator'=>'string', 'array'=>'array'],
-'implode\'1' => ['string', 'separator'=>'array'],
+'implode' => ['string', 'separator'=>'string', 'array'=>'string[]'],
```
voku, line 5992: ```int[]``` and ```float[]``` are also accepted: https://3v4l.org/E7ZOU

BenMorel: You're right. Actually, pretty much anything is accepted: https://3v4l.org/iq8aj

voku: ;D I don't know if it makes sense but it's apparently possible.

PR #4832, opened 2020-12-12
Memoize private inferred mutation-free methods

2e3s:
More relaxed modification from the rejected [approach](https://github.com/vimeo/psalm/pull/4808): given an inferred mutation-free behaviour, if final methods are memoized because they cannot be overridden, private methods cannot be overridden as well, hence it's logical to memoize them too.

https://psalm.dev/r/91cfaaafb0


muglug:
Looks good, thanks!

Issue #4820, opened 2020-12-09
[Plugin question] retrieve a the type of a particular expression starting from File hook


orklah:
TL;DR: I used the AfterFileAnalysisInterface and I can't find a way to retrieve the type of a particular Return_ statement I get while traversing Php-Parser nodes.

I'm trying to create a plugin that will crawl every node for a particular file to find any violation to strict_types rules(the goal is to suggest adding the declare expression when safe)

I made the recursive calls to each statement and each expression in order to find any function call, any property assignment and any return statement that could be faulty and analyse each one.

I'm at the point where I found a specific Return_ statement for a particular FunctionLike and I would like to retrieve its type but I can't get it

So far, I've tried:
- `$file_analyzer->getNodeTypeProvider()->getType($stmt->expr)`
- `$file_analyzer->getNodeTypeProvider()->getType($stmt)`

At that point, I guessed there must be several NodeTypeProvider and went looking for the right one so I tried:
- `$file_analyzer->getFunctionLikeAnalyzer(new MethodIdentifier('Test', 'test'))->getNodeTypeProvider()->getType($stmt->expr);`
- `$file_analyzer->getFunctionLikeAnalyzer(new MethodIdentifier('Test', 'test'))->getNodeTypeProvider()->getType($stmt);`

I tried to debug it some more and I realized that the one and only Return_ in the code I analyze had an spl_object_id that I never receive in any node in the hook, so I guess the storage mechanism doesn't work in this case 

I'm kinda lost right now. This is not a covered topic on authoring plugins page on psalm.dev and when I tried to ask for help to plugin authors about this kind of subject, the only response I had told me this was also an obscure topic for them and I couldn't get an answer.

Is there some kind of central hub I'm missing or is there any logic I don't get to help navigate analyzed codebases?

Sorry for the long post, I wanted to show I've done actual research before reaching out :)

Thanks

orklah:
Edited the question based on accepted PR

PR #4834, opened 2020-12-12
preserve analyzers until the end of file

orklah:
Related to https://github.com/vimeo/psalm/issues/4820

This drops the $preserve_analyzers that was always false and move the removal of analyzers below the call to the File plugin hook


muglug:
Thanks!

PR #4835, opened 2020-12-13
push original statements list to file plugin

orklah:
This PR propose adding the list of statements directly to the hook for AfterFileAnalysis.

This should fix https://github.com/vimeo/psalm/issues/4820. After hours of tweaking, I came to the conclusion that I couldn't get types in the plugin because I didn't retrieved the original list of statement when I used this syntax:
`$stmts = $codebase->getStatementsForFile($file_storage->file_path);`
I think the cache is causing them to be recreated between the analysis and the plugin hook. This makes it impossible to fetch the types through this syntax:
`$type = $statement_source->getNodeTypeProvider()->getType($stmt->expr);`
It returns null because the new nodes created by cache fetching weren't analyzed and stored here.

Unfortunately, this PR is a BC break. It adds a new parameter to the hook so it will probably break every AfterFileAnalysis plugin.

I searched on GitHub and I could only find one usage of this hook (except mine) here:
https://github.com/omegaup/omegaup/blob/df92883630bd6c1efa42905985337ab535b392aa/frontend/server/src/Psalm/TranslationStringChecker.php#L109 (and it's still in Psalm V3)
But I can't be sure there isn't private usage not published somewhere...

### DIFF
--- a/src/Psalm/Internal/Analyzer/FileAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FileAnalyzer.php
@@ -271,7 +271,7 @@ class FileAnalyzer extends SourceAnalyzer
             }
         }
 
-        $event = new AfterFileAnalysisEvent($this, $this->context, $file_storage, $codebase);
+        $event = new AfterFileAnalysisEvent($this, $this->context, $file_storage, $codebase, $stmts);
         $codebase->config->eventDispatcher->dispatchAfterFileAnalysis($event);
 
         $this->class_analyzers_to_analyze = [];
--- a/src/Psalm/Plugin/EventHandler/Event/AfterFileAnalysisEvent.php
+++ b/src/Psalm/Plugin/EventHandler/Event/AfterFileAnalysisEvent.php
@@ -3,6 +3,7 @@
 
 namespace Psalm\Plugin\EventHandler\Event;
 
+use PhpParser\Node\Stmt;
 use Psalm\Codebase;
 use Psalm\Context;
 use Psalm\StatementsSource;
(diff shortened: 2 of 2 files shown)


weirdan:
Since we in the land of BC breaks I have an idea on how to make it easier down the line. The problem we have is because to supply additional data to plugins plugins themselves need to provide a way for Psalm to push that data (e.g. in a form of additional parameter as in this PR).
If, instead, they could pull it out of a DTO passed to them the DTO could be freely extended without BC breaks.

Essentially it's going from

```php
// Psalm
$plugin_class::afterAnalyzeFile($this, $this->context, $file_storage, $codebase);

// Plugin
public function afterAnalyzeFile(Analyzer $analyzer, Context $context, Storage $storage, Codebase $codebase, Statement[] $statements) {
   // work with parameters
}
```

to 

```php
// Psalm 
$plugin_class::afterAnalyzeFile(new AfterAnalyzeFileEvent($this, $this->context, $file_storage, $codebase, $statements));
// Plugin
public function afterAnalyzeFile(AfterAnalyzeFileEvent $event) {
   $storage = $event->getStorage();
   // work with fetched objects
}
```

In a way, this is similar to how DOM passes events around, which proved fairly flexible so far.

Of course the switch would need to be done in a major release.

PS: @muglug is it possible to enable GH discussions on this repo?

muglug:
> PS: @muglug is it possible to enable GH discussions on this repo?

I'm waiting a week or so to see how github.com/phpstan/phpstan/discussions looks

muglug:
I am very on board with migrating plugin hooks to use a DTO

orklah:
@weirdan is it something you want to work on?

weirdan:
@orklah I don't have much free time at the moment, unfortunately

orklah:
Now that #4881 is merged, I'll propose changing this with DTO, that way this won't be a BC break.

muglug:
Thanks!

Issue #4854, opened 2020-12-16
Possibly incorrect handling of session_name()


christeredvartsen:
According to the PHP manual, this is the signature of [session_name](https://php.net/session_name):

    session_name ([ string|null $name = null ] ) : string|false

Psalm does not seem to think that `session_name` returns anything other than a string:

https://psalm.dev/r/7fc1845367

weirdan:
Should be fixable in `dictionaries/CallMap.php`

christeredvartsen:
Guess this can be closed now, or do you want to wait until the fix has been released?

weirdan:
Thanks for the fix!

PR #4855, opened 2020-12-16
Allow session_name to also return false

christeredvartsen:
According to the PHP docs the [session_name](https://php.net/session_name) signature is:

    session_name ([ string|null $name = null ] ) : string|false

Resolves Possibly incorrect handling of session_name() #4854.


weirdan:
> According to the PHP docs

And source: https://github.com/php/php-src/blob/8067cf475205b4af1f703584beb4b7db4897223d/ext/session/session.c#L1869-L1877

muglug:
Thanks!

PR #4860, opened 2020-12-17
Keep method nodeTypeProviders at class level

orklah:
Closely related to https://github.com/vimeo/psalm/issues/4820

This PR propose saving method NodeDataProviders on class Analyzer.

This is needed to allow access to method Node types when using the `afterAnalyzeFile` hook.

This should not affect memory much, this only keeps the Nodes until destruction of ClassAnalyzer, which happens after each file analysis.


muglug:
I'm not sure about this – couldn't you instead use an `afterFunctionLikeAnalysis` hook that would give you the type provider?

orklah:
My goal is to make a plugin that detects any strict_type violation and if there isn't, I add the declare().

This means I necessarily work at file level because I need to check every statement for the whole file to be sure I don't miss a wrong statement.

There are case in where the afterFunctionLikeAnalysis is not called: there are 3 returns in the method before the call to the plugin hook.

muglug:
since all functionlikes within a class use the same instance of the plugin, you could just store a map for cross-referencing

orklah:
Interesting, I'll keep this open while I try something like that. Thanks!

Issue #4857, opened 2020-12-17
Array types


mj4444ru:
https://psalm.dev/r/4fb467dba3

https://psalm.dev/r/e58247c551

An array can actually only be of 3 types:
array<array-key, mixed>
array<int, mixed>
array<string, mixed>

Array cannot be of type: array<float|int, mixed> and array<float, mixed>

Example:
```php
$arr = [];
$arr[1.1] = 1.1;
```
The result will be: [1 => 1.1]
This is an array type: array<int, float>

_Originally posted by @mj4444ru in https://github.com/vimeo/psalm/issues/4848#issuecomment-747215056_

Issue #4858, opened 2020-12-17
Array and InvalidArrayOffset


mj4444ru:
https://psalm.dev/r/e58247c551

Issue https://github.com/vimeo/psalm/issues/4846 has been resolved. But the errors described here are still present.
Closing this issue was a mistake.

_Originally posted by @mj4444ru in https://github.com/vimeo/psalm/issues/4847#issuecomment-747218311_

muglug:
This is now improved with @orklah's recent changes

PR #4861, opened 2020-12-17
Illegal offsets access

orklah:
This should fix https://github.com/vimeo/psalm/issues/4857 and https://github.com/vimeo/psalm/issues/4858

Basically, this checks that every ArrayDimFetch is used with a valid offset.

This was not my first goal. At first, I just wanted to forbid the syntax `$a[1.1] = ''` but this went to forbid `$a[1.1]` altogether because this was simpler and still a correct issue. We could be less strict by checking that at least one type in the union is array-key compatible instead of checking that all the types should.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Block/SwitchCaseAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/SwitchCaseAnalyzer.php
@@ -110,7 +110,7 @@ class SwitchCaseAnalyzer
             if ($fake_switch_condition) {
                 $statements_analyzer->node_data->setType(
                     $switch_condition,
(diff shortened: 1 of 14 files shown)


orklah:
I allowed null in array access because this should already be covered by PossiblyNullArrayOffset.

I added casts in Psalm's code for PossiblyNullArrayOffset that were suppressed by ignore-nullable-issue. They were unnecessary. I can try to remove them if necessary

orklah:
I converted to draft. Seems like I overlooked PossiblyInvalidArrayOffset, some of the case launch two issues

orklah:
I'll propose something else. I clearly missed a lot of already handled cases.

PR #4864, opened 2020-12-18
store node type provider on functionlike level

orklah:
This PR propose storing the node type provider on a property of FunctionLikeAnalyzer.

This is needed to retrieve the node type provider from AfterFunctionLikeAnalysis Hook.


orklah:
I also included a property to store the context at functionlike level.

It seems that this is the only way to retrieve the type of $this. Another solution would be to add the expr $this to the node data provider, I'm not sure why it's not and what that would impact.

orklah:
@muglug Could you review this please? I'm sorry to rush you, I'd like to progress on this on my end 😄

orklah:
Now that #4881 is merged, I'll propose changing this with DTO, that way this won't be a BC break.

muglug:
Thanks!

PR #4865, opened 2020-12-18
Illegal offset

### LINKED ISSUE #4857: Array types

orklah:
This should fix #4857 and #4858

There was already a check for offset type to verify this was coherent with expected type. I added a new variable to check the offset was coherent by itself

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
@@ -393,7 +393,8 @@ class ArrayFetchAnalyzer
         $has_array_access = false;
         $non_array_types = [];
 
-        $has_valid_offset = false;
+        $has_valid_expected_offset = false;
+        $has_valid_absolute_offset = false;
         $expected_offset_types = [];
 
         $key_values = [];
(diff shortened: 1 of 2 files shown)


tests/ArrayAccessTest.php
```diff
@@ -1118,7 +1118,7 @@ public function __toString() {
                     function foo(array $a, $key) : void {
                         echo $a[$key];
                     }',
-                'error_message' => 'PossiblyInvalidArrayOffset',
+                'error_message' => 'InvalidArrayOffset',
```
muglug, line 1121: This is incorrect, though – since one of the types (`string`) is valid as a key

orklah: Thanks, it should be better now

muglug:
Thanks!

PR #4869, opened 2020-12-19
Add taint analysis to language server

Danysan1:
When the taint analysis feature was added in b2c0993cdc023608c1ef7c4b010ae5a0354c6b8f and 23ab0f1ddb552347fffde0ff6da5a48fb70356f7 it was added only for the `psalm` executable, not for `psalm-language-server`.
Hence, IDE plugins were not able to display security alerts (see for example https://github.com/psalm/psalm-vscode-plugin/issues/25 ).

I this commit I have ported from [`psalm.php`](https://github.com/vimeo/psalm/blob/master/src/psalm.php) to [`psalm-language-server.php`](https://github.com/vimeo/psalm/blob/master/src/psalm-language-server.php) the lines which load the taint analysis configuration.

With these changes and [psalm-vscode-plugin](https://github.com/psalm/psalm-vscode-plugin) I have been able to display taint alerts on Visual Studio Code:
immagine

### DIFF
--- a/src/psalm-language-server.php
+++ b/src/psalm-language-server.php
@@ -60,6 +60,9 @@ $valid_long_options = [
     'tcp-server',
     'disable-on-change::',
     'enable-autocomplete',
+    'track-tainted-input',
+    'taint-analysis',
+    'security-analysis',
     'use-extended-diagnostic-codes',
     'verbose'
 ];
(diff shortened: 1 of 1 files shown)


muglug:
Thanks! I won't be merging this, though – to be useful taint analysis requires a scan of the entire codebase, so it's not well-suites to operate in language server mode.

PR #4873, opened 2020-12-21
Allow by reference parameters in method docblock

### LINKED ISSUE #4625: Problem with parameters by reference using phpdoc method
Looks like if a method defined in phpdoc is using a parameter by reference makes psalm fail or do not read the rest of the annotations.

https://psalm.dev/r/a82f1d6340

franmomu:
This tries to fix #4625, I've tried to imitate the variadic example.

Debugging this, the issue was that when interpreting the `&` it was creating a `IntersectionTree` instead of treating it as a part of the parameter. Tests pass, but not sure if this is the right fix 😬  

Fixes #4625


muglug:
Looks great, thanks!

PR #4881, opened 2020-12-22
implement DTO for plugins

orklah:
This PR propose introducing Data Tansfer Objects for plugins in Psalm

Right now the CI will fail because it depends on a plugin to run and the plugin itself is not compatible yet.

This will allow to discuss about the implementation and how to proceed now

EDIT: this idea was proposed by @weirdan here: https://github.com/vimeo/psalm/pull/4835#issuecomment-744055926


docs/running_psalm/dealing_with_code_issues.md
```diff
@@ -114,26 +114,18 @@ If you want something more custom, like suppressing a certain type of error on c
 <?php
 namespace Foo\Bar;
 
-use PhpParser\Node;
-use Psalm\Codebase;
-use Psalm\FileSource;
 use Psalm\Plugin\Hook\AfterClassLikeVisitInterface;
-use Psalm\Storage\ClassLikeStorage;
+use Psalm\Plugin\Hook\TransfertObject\AfterClassLikeVisitEvent;
```
weirdan, line 118: It's 'transfer' in English

weirdan:
The main question, I guess, is whether we're doing this as a hard BC break (thus in a major version) or should we provide a transition period (and what are our options for that).

docs/running_psalm/dealing_with_code_issues.md
orklah: Damn! Excuse my French 😄

src/Psalm/Plugin/Hook/TransferObject/AfterClassLikeAnalysisEvent.php
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public function getFileReplacements(): array
+    {
+        return $this->file_replacements;
+    }
+
+    /**
+     * @param FileManipulation[] $file_replacements
+     */
+    public function setFileReplacements(array $file_replacements): void
```
weirdan, line 87: Perhaps we could also add `addFileReplacement()`. But it doesn't have to happen right now, as we would be able to extend the event class later.

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
```diff
@@ -973,15 +974,17 @@ public function analyze(
             $file_manipulations = [];
 
             foreach ($plugin_classes as $plugin_fq_class_name) {
-                if ($plugin_fq_class_name::afterStatementAnalysis(
+                $event = new AfterClassLikeAnalysisEvent(
```
weirdan, line 977: Do we need to reinitialize the event for every hook call?

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ConstFetchAnalyzer.php
```diff
@@ -13,6 +13,7 @@
 use Psalm\Context;
 use Psalm\Issue\UndefinedConstant;
 use Psalm\IssueBuffer;
+use Psalm\Plugin\Hook\TransferObject\StringInterpreterEvent;
```
weirdan, line 16: Is it actually used?

src/Psalm/Internal/Codebase/Reflection.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace Psalm\Internal\Codebase;
 
+use Psalm\Plugin\Hook\TransferObject\StringInterpreterEvent;
```
weirdan, line 4: Is it used?

src/Psalm/IssueBuffer.php
```diff
@@ -554,12 +555,13 @@ function (IssueData $d1, IssueData $d2) : int {
 
             foreach ($after_analysis_hooks as $after_analysis_hook) {
                 /** @psalm-suppress ArgumentTypeCoercion due to Psalm bug */
```
weirdan, line 557: Is this suppression still required?

weirdan (approved):

docs/running_psalm/dealing_with_code_issues.md
weirdan: Having given this a bit more thought I think it would be even better if the namespace was `Event` or `Events`.

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
orklah: We don't have control of what the plugin will do with the event. They could unset properties for example. Also file_manipulation is risky to keep.

src/Psalm/IssueBuffer.php
orklah: yeah. FYI, the error is `Argument 2 of Psalm\Plugin\Hook\TransferObject\AfterAnalysisEvent::__construct expects array<string, list<Psalm\Internal\Analyzer\IssueData>>, parent type array<string
, array<list-key<$file_issues>, Psalm\Internal\Analyzer\IssueData>> provided`

src/Psalm/Internal/Codebase/Reflection.php
orklah: thanks. Good catch!

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ConstFetchAnalyzer.php
orklah: thanks. Good catch!

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
orklah: which makes me think I should merge file_manipulations and push a clean array each time I think. I'll do that this afternoon

docs/running_psalm/dealing_with_code_issues.md
orklah: Done

muglug:
For compatibility I think the options are:
- create a new namespace for these plugins and allow both old and new to work in Psalm 4
- Use the regular namespace, but create a 5.x branch and merge into that instead of master

I slightly prefer the former (which allows for a transition), but it does sort of mess up the nice naming scheme @weirdan arranged

orklah:
So, that would mean duplicating every plugin interface, duplicating the properties where classes are stored, duplicating tests, possibly examples too.

That's a lot of work and lots of code to maintain until V5. But if that's what you want, I can begin working on that

weirdan:
Introducing event dispatcher, either single with methods like `dispatchAfterClassAnalysis` or generic `EventDispatcher<AfterClassAnalysisEvent>->dispatch()` may help to centralize if not reduce the duplication.

orklah:
@weirdan mind detailing more? I'm not sure how to proceed from what you said

weirdan:
For option 1 I mean something like this: https://gist.github.com/weirdan/5ecd80ec64bd000909359efc540020dc

Option 2 is not possible with non-uniform hook method names, so ignore that.

orklah:
Oh, I see, thanks!

@muglug, do you agree with that? I'd like to avoid working on that if it has to be refactored after

muglug:
@orklah I like option 1 as long as it doesn't have an obvious performance impact (which it doesn't look like it will)

orklah:
This should be ok now

weirdan (approved):
Looks good to me

muglug:
Thank you!

orklah:
Thank you too :) I'll be able to progress on my project now !

PR #4884, opened 2020-12-23
Add renaming of traits and interfaces

nicoorfi:
This PR makes possible to use the command

```
vendor/bin/psalm-refactor --rename "Ns1\Foo" --to "Ns2\Bar\Baz"
```

event if the `Ns1\Foo` is a `Trait` or and `Interface`.


orklah:
I'm changing it to draft to prevent cluttering the PR view. Do you intend to work on that again?

PR #4904, opened 2020-12-28
Document lowercase-string when possible

orklah:
I made a blackfire profiling on psalm today and I was surprised to see that the most called function and the one that takes the longer is `strtolower`. I wanted to see if there wasn't some redundant functions so I documented lowercase-string where I could.

Psalm didn't report any redundant strtolower :( Worse, I found two places where there was a missing strtolower(cf commit 2)

### DIFF
--- a/src/Psalm/Aliases.php
+++ b/src/Psalm/Aliases.php
@@ -4,22 +4,22 @@ namespace Psalm;
 class Aliases
 {
     /**
-     * @var array<string, string>
+     * @var array<lowercase-string, string>
      */
     public $uses;
 
     /**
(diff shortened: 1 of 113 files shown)


staabm:
IIRC strtolower() is one of the functions which php can optimize when fully qualified (e.g. using a function import or noted with a backslash)

orklah:
I believe Psalm has a CS check for unqualified functions anyway so this would have been done already

muglug:
Thanks! I noticed the same a while back, which was the main reason for introducing `lowercase-string` in the first place (and also wanting to keep it internal).

Issue #4876, opened 2020-12-21
Psalm runs out of memory analyzing long chain of && and || conditions in Mobile_Detect.php


TysonAndre:
Noticed when testing psalm on various projects, e.g. https://raw.githubusercontent.com/serbanghita/Mobile-Detect/master/Mobile_Detect.php

`psalm --debug-by-line` indicates that it's stuck on line 1360 of https://github.com/serbanghita/Mobile-Detect/blob/6f8113f57a508494ca36acbcfa2dc2d923c7ed5b/Mobile_Detect.php

I'm easily able to exclude this file from analysis, but wasn't sure if this was a performance edge case you'd be interested in.

**After around a minute, PHP exits after exhausing the memory limit of 8 gigabytes** in src/Psalm/Internal/Clause.php on line 271.

This was seen in `3390097b3` and probably earlier commits from weeks ago (from Psalm on december 21)

phpspy indicates that a lot of time is spent in Psalm\Internal\Algebra\FormulaGenerator::getFormula and Psalm\Internal\Algebra::combineOredClauses

```php
        if (
            // Apple iOS 4-7.0 – Tested on the original iPad (4.3 / 5.0), iPad 2 (4.3 / 5.1 / 6.1), iPad 3 (5.1 / 6.0), iPad Mini (6.1), iPad Retina (7.0), iPhone 3GS (4.3), iPhone 4 (4.3 / 5.1), iPhone 4S (5.1 / 6.0), iPhone 5 (6.0), and iPhone 5S (7.0)
            $this->is('iOS') && $this->version('iPad', self::VERSION_TYPE_FLOAT) >= 4.3 ||
            $this->is('iOS') && $this->version('iPhone', self::VERSION_TYPE_FLOAT) >= 4.3 ||
            $this->is('iOS') && $this->version('iPod', self::VERSION_TYPE_FLOAT) >= 4.3 ||

            // Android 2.1-2.3 - Tested on the HTC Incredible (2.2), original Droid (2.2), HTC Aria (2.1), Google Nexus S (2.3). Functional on 1.5 & 1.6 but performance may be sluggish, tested on Google G1 (1.5)
            // Android 3.1 (Honeycomb)  - Tested on the Samsung Galaxy Tab 10.1 and Motorola XOOM
            // Android 4.0 (ICS)  - Tested on a Galaxy Nexus. Note: transition performance can be poor on upgraded devices
[...]

PR #4911, opened 2020-12-29
Check if included files are listed in mustBeIgnored

TysonAndre:
Previously, when a file was included in `include()`, it was not subject
to ignoreFiles

Related to #4876


muglug:
Thanks!

PR #4915, opened 2020-12-30
Add missing param for constructor arguments in PDOStatement#fetchObject

hollodotme:
`PDOStatement#fetchObject` allows a second, optional parameter for constructor arguments, which - if given - will be passed to the given class' constructor.
See: https://www.php.net/manual/de/pdostatement.fetchobject.php

Also see the PhpStorm stubs: https://github.com/JetBrains/phpstorm-stubs/blob/master/PDO/PDO.php#L1441

Currently psalm reports a `TooManyArguments` error for code like this:

```php
<?php

class MyClass {}

/** @var PDO $pdo */
$pdo->query('some-query')->fetchObject(MyClass::class, ['ctor', 'args']);
``` 

See: https://psalm.dev/r/9426558ef1

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -9884,7 +9884,7 @@ return [
 'PDOStatement::fetch' => ['mixed', 'how='=>'int', 'orientation='=>'int', 'offset='=>'int'],
 'PDOStatement::fetchAll' => ['array|false', 'how='=>'int', 'fetch_argument='=>'int|string|callable', 'ctor_args='=>'?array'],
 'PDOStatement::fetchColumn' => ['string|int|float|bool|null', 'column_number='=>'int'],
-'PDOStatement::fetchObject' => ['object|false', 'class_name='=>'string', 'ctor_args='=>'?array'],
+'PDOStatement::fetchObject' => ['object|false', 'class_name='=>'string', 'ctor_args='=>'array'],
 'PDOStatement::getAttribute' => ['mixed', 'attribute'=>'int'],
 'PDOStatement::getColumnMeta' => ['array|false', 'column'=>'int'],
 'PDOStatement::nextRowset' => ['bool'],
--- a/stubs/pdo.php
+++ b/stubs/pdo.php
@@ -6,7 +6,8 @@ class PdoStatement {
      *
      * @template T
      * @param class-string<T> $class
+     * @param array $ctorArgs
      * @return false|T
      */
-    public function fetchObject($class = "stdclass") {}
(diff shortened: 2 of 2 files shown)


orklah:
in Callmap.php, the parameter is a `?array` instead, which *seems* more correct to me but I don't have a way to test it on hand.

Could you try something like:
```php
class MyClass {
    
    public function __construct($a){
        var_dump($a);
    }
}

/** @var PDO $pdo */
$pdo->query('some-query')->fetchObject(MyClass::class, []);
/** @var PDO $pdo */
$pdo->query('some-query')->fetchObject(MyClass::class);
/** @var PDO $pdo */
$pdo->query('some-query')->fetchObject(MyClass::class, null);
```

and report what the output is?

hollodotme:
@orklah I used the following portable code to test it:

```php
<?php

class MyClass {}

$pdo = new PDO('sqlite::memory:');

$pdo->query('SELECT 1')->fetchObject(MyClass::class, []);

$pdo->query('SELECT 1')->fetchObject(MyClass::class);

$pdo->query('SELECT 1')->fetchObject(MyClass::class, null);
```

And it throws the following exception in the last line:

```
PHP Fatal error:  Uncaught TypeError: PDOStatement::fetchObject(): Argument #2 ($ctorArgs) must be of type array, null given in /Users/hollodotme/test.php:13
Stack trace:
#0 /Users/hollodotme/test.php(13): PDOStatement->fetchObject('MyClass', NULL)
#1 {main}
  thrown in /Users/hollodotme/test.php on line 13
```

So `null` is not allowed, thus `?array` cannot be used as parameter type declaration.

I tested it on PHP 8 locally, but you can verify it here for older versions: https://3v4l.org/qBFT0

orklah:
You're right. Can you also fix the line in callmap.php and remove the `?`

hollodotme:
@orklah I'm happy to do so, but I actually don't know what you are referring to with `callmap.php`. Can you give me a link/pointer please?

orklah:
It's the file here: https://github.com/vimeo/psalm/blob/master/dictionaries/CallMap.php

It's the main source of types for php functions/methods. There's one line per function/method and it describe the types of params and return.

At line 9887, you have `PDOStatement::fetchObject` and the arg `ctor_args` is declared as `?array`. You can remove the `?` as we showed that it can't be null.

(This was the source that was used until recently when the stub you modified was added to be more detailed than the callmap file, unfortunately, there was an error added here too.)

hollodotme:
@orklah Thanks and done :)

muglug:
Thanks!

PR #4922, opened 2020-12-31
Full support for `$this` in `@var` annotations

### LINKED ISSUE #4916: `$this` outside a class is accepted with `@psalm-scope-this`, but not with `@var ... $this`
## Use-case

My codebase uses bound `$this` variables in closures in quite a few places (using `Closure::call(object $newthis)`), mostly outside a class (meaning: a closure is directly returned from a configuration file and then called from a class it is injected to).

Simplified demonstration snippet: https://psalm.dev/r/fe823c6796

If `@psalm-scope-this` is used instead of `@var`, it works. This is because [only `@psalm-scope-this` fakes the "this class"](https://github.com/vimeo/psalm/blob/4.3.2/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php#L680): https://psalm.dev/r/1c6322da42

## Output (for the first demonstration snippet with `@var`)

```
Scanning files...
Analyzing files...

E

ERROR: InvalidScope - test.php:5:3 - Use of $this in non-class context (see https://psalm.dev/013)
  /** @var \Exception $this */
  $this->getMessage();


1 errors found

Checks took 0.98 seconds and used 231.681MB of memory
```

## Expected behavior

As `@var` can also be understood by IDEs and is much more universal as `@psalm-scope-this`, it would be great if I could use `@var` annotations everywhere, even outside classes.

Required change: `@var` annotations should set the "faked this class" if the annotated var is `$this`.

## Psalm version I used for testing

```
Psalm dev-master@3b6550f892b97039bcdca0bd44f89c791da0c883
```

lukasbestle:
Fixes #4916.

### DIFF
--- a/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
@@ -460,6 +460,13 @@ class StatementsAnalyzer extends SourceAnalyzer
                         $var_comment,
                         $context
                     );
+
+                    if ($var_comment->var_id === '$this'
+                        && $var_comment->type
+                        && $codebase->classExists((string)$var_comment->type)
+                    ) {
+                        $statements_analyzer->setFQCLN((string)$var_comment->type);
+                    }
                 }
             }
         } else {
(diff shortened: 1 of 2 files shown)


muglug:
Thanks!

PR #4923, opened 2021-01-02
remove test namespace from autoload-dev to allow usage outside of Psalm (e.g. for plugins)

orklah:
I started a discussion on this subject here: https://github.com/vimeo/psalm/discussions/4887

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,17 +72,13 @@
     },
     "autoload": {
         "psr-4": {
-            "Psalm\\": "src/Psalm/"
+            "Psalm\\": "src/Psalm/",
+            "Psalm\\Tests\\": "tests/"
         },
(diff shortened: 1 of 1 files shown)


weirdan:
Note that when installed with `--prefer-dist` the `tests` folder is not present, due to it being marked as `export-ignore` in `.gitattributes`.

orklah:
As a note: I've been able to successfully run a [CI](https://github.com/orklah/psalm-strict-types/runs/1744587969) with phpunit based on [4 files](https://github.com/orklah/psalm-strict-types/tree/master/tests/PsalmInternal) copy and pasted from psalm.

It'd be great to be able to directly reuse those classes from psalm instead

PR #4924, opened 2021-01-02
SARIF report: Include help links

lukasbestle:
This will make it quicker for users to access the help as tools can directly print the link in the issue details.


muglug:
Thanks!

PR #4925, opened 2021-01-03
PHAR: Include issue docs for SARIF output

lukasbestle:
The SARIF report uses the Markdown files inside `docs/running_psalm/issues` for its output, so the files should be included in the PHAR.


muglug:
Thanks!

Issue #4933, opened 2021-01-04
Invalid return in SimpleXMLElement


snapshotpl:
https://psalm.dev/r/f5ae1a573a

https://www.php.net/manual/en/simplexmlelement.children.php

PR #4934, opened 2021-01-04
add ArrayAccess to SimpleXmlElement

orklah:
This should fix https://github.com/vimeo/psalm/issues/4933

However, I'm not sure this is the correct fix as https://github.com/vimeo/psalm/compare/master...orklah:ArrayAccess?expand=1#diff-ae46a3ded74600c7b71b7e92661c1427c514741f10191e1b9aa49c7a3e83c5afR835 already works somehow


muglug:
It's good enough for me, I think

PR #4938, opened 2021-01-05
handle concatenation with int parts

orklah:
This PR handles some concatenation cases with int parts.
When two int are concatened, it results in numeric-string. When two literal ints are concatened, we can infer the resulting string.

This is useful for example when using this plugin: https://github.com/orklah/psalm-strict-numeric-cast who require having a numeric-string before casting to int. There are cases in Psalm itself where we concat two ints (php major/minor version) where this can help

I also fixed a comment and an error message I saw


muglug:
Thanks!

PR #4940, opened 2021-01-06
Taint specialized calls even when not using a variable

adrienlucas:
Hi there ! Happy new year !

I found a strange behaviour (possibly introduced by https://github.com/vimeo/psalm/commit/95de6cf177845976f10be22340eedfb593e0007c) : when having a specialized call, if the object instance is not stored in a variable, the taint does not _go out_ (not sure about the terminology here) of the call.

I added two test cases to prove the point, but the fix I introduced is really naive (only merge two `if`s) and I'm pretty sure it will not be sufficient. Let me know how, and I'll be happy to improve it :)


muglug:
Looks fine to me! Thanks!

PR #4948, opened 2021-01-07
Avoid false positives for taint specialized calls when not using a variable

adrienlucas:
Following #4940 that was too naïve, here is the improvement :)


muglug:
Thanks!

PR #4962, opened 2021-01-08
Trigger dispatch even when only legacy hooks

adrienlucas:
Hey ! With the new `EventDispatcher` mechanism there is an issue when there are only legacy hooks registered for the `AfterMethodCallAnalysis` and `AfterClassLikeVisit` handlers : the dispatch does not occur.


orklah:
Ouch, good catch, thanks!

adrienlucas:
Should we also make `private` all the other properties of `EventDispatcher` ? It would require a little rework on the `PluginTest` though.

muglug:
Thanks!

Issue #4965, opened 2021-01-08
ParseError - vendor/symfony/polyfill-mbstring/bootstrap80.php:125:86 


klebann:
While scanning ../../mod/checklist weird Error from vendor/symfony folder is popping out.
Why this is happening, although I ignore vendor files in psalm-plugin.xml?

```xml
<projectFiles>
    <ignoreFiles>
        <directory name="vendor"/>
        <file name="psalm-autoload.php"/>
    </ignoreFiles>
</projectFiles>
``` 

```console
./vendor/bin/psalm --config=psalm-plugin.xml --show-info=true --no-diff ../../mod/checklist
Scanning files...
Analyzing files...

░░░░░░I░░░░░░░░░░░░░░░I░░░░░░░░░░░░░░I░░░░░░

ERROR: ParseError - vendor/symfony/polyfill-mbstring/bootstrap80.php:125:86 - Syntax error, unexpected '=' on line 125 (see https://psalm.dev/173)
    function mb_scrub(string $string, string $encoding = null): string { $encoding ??= mb_internal_encoding(); return mb_convert_encoding($string, $encoding, $encoding); }


INFO: PossibleSqlInjection - ../../mod/checklist/autoupdate.php:107:24 - Calling unsafe sql method $DB->get_records_sql
SQL:
    SELECT i.id AS itemid, i.checklist, cl.teacheredit, ck.*, cl.course
              FROM {checklist_item} i
              JOIN {checklist} cl ON i.checklist = cl.id
              LEFT JOIN {checklist_check} ck ON (ck.item = i.id AND ck.userid = :userid)
             WHERE cl.autoupdate > 0 AND i.moduleid = :cmid AND i.itemoptional < :heading
Documentation -
    $itemchecks = $DB->get_records_sql($sql, ['userid' => $userid, 'cmid' => $cmid, 'heading' => CHECKLIST_OPTIONAL_HEADING]);
```

gmessier:
Same as #4961

PR #4967, opened 2021-01-09
Skip conditional includes when registering autoload files

### LINKED ISSUE #4961: ParseError in (ignored) vendor symfony/polyfill-mbstring/bootstrap80.php
In two symfony 4.4 projects we use PHP 7.2.
After updating to the [newest release 1.22.0](https://github.com/symfony/polyfill-mbstring/compare/v1.20.0...v1.22.0#diff-a97bb9776f637d300bee07c901a6ca59a2378845f507441b74939327606fc0f6) of symfony/polyfill-mbstring wo got the following new error after running psalm

```
ERROR: ParseError - vendor/symfony/polyfill-mbstring/bootstrap80.php:125:86 - Syntax error, unexpected '=' on line 125 (see https://psalm.dev/173)
    function mb_scrub(string $string, string $encoding = null): string { $encoding ??= mb_internal_encoding(); return mb_convert_encoding($string, $encoding, $encoding); }
```

The vendor folder itself is ignored.

While I understand this cannot be parsed by PHP 7.2, I do not understand why the PHP 8 bootstrap file is analysed at all. The only reference is in symfony packages with a PHP version check.

```
if (\PHP_VERSION_ID >= 80000) {
    return require __DIR__.'/bootstrap80.php';
}
```

Do I miss something or is this a problem in psalm?

weirdan:
This should allow analysis of forward-compatible polyfills used in codebases targeting older PHP versions.

The following previously problematic polyfills should now emit no parse errors:

* `symfony/polyfill-ctype`
* `symfony/polyfill-intl-grapheme`
* `symfony/polyfill-intl-normalizer`
* `symfony/polyfill-mbstring`

The pattern that is now works as intended looks like this:
```php
if (\PHP_VERSION_ID >= 80000) {
    require __DIR__ . '/bootstrap80.php';
    return;
}
```

Previously Psalm would scan the required file even when the codebase targeted older PHP versions, and would emit parse errors when that file contained PHP 8 syntax.

Fixes #4961 and #4965

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ExpressionResolver.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ExpressionResolver.php
@@ -6,6 +6,10 @@ use function function_exists;
 use function implode;
 use function interface_exists;
 use PhpParser;
+use PhpParser\ConstExprEvaluationException;
+use PhpParser\ConstExprEvaluator;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\ConstFetch;
 use Psalm\Aliases;
 use Psalm\Codebase;
 use Psalm\Internal\Analyzer\ClassLikeAnalyzer;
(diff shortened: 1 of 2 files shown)


muglug:
Thanks!

klebann:
Thanks!

PR #4968, opened 2021-01-09
Removed misc files from dist archives

weirdan:
This adds more files / folders to be ignored when building dist composer packages, so that only necessary files appear in `vendor` folder:

```
/dictionaries          
/src                   
/stubs                 
 LICENSE               
 composer.json         
 config.xsd            
*psalm                 
*psalm-language-server 
*psalm-plugin          
*psalm-refactor        
*psalter               
```


muglug:
Thanks!

williamdes:
Thanks a lot!

PR #4974, opened 2021-01-11
Added GHA workflow to build phars

weirdan:
Required setup:
On `Settings / Secrets` add three secrets:
* `GPG_ENCRYPTION` - passphrase used to encrypt `keys.asc.gpg`
* `SIGNING_KEY` - passphrase for actual signing key
* `SIGNING_KEY_ID` - key ID used to sign phar with (use `8A03EA3B385DBAA1` to use the same key Travis currently uses).

### DIFF
--- a/.github/workflows/build-phar.yml
+++ b/.github/workflows/build-phar.yml
@@ -1,0 +1,53 @@
+name: Build phar
+
+on: 
+  release:
+    types:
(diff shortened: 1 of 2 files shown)


muglug:
Thanks! I think I set those up correctly, will debug a bit to check

PR #4985, opened 2021-01-11
Split unit test over multiple jobs in GitHub Actions, lowering the total run time

bendavies:
Hi Matt,

I noticed that the unit test took quite a while to run in GitHub Actions, with the last few runs taking over 20 minutes to complete.

This is quite a long feedback loop, so I've tried to speed up it up.

What I've done is:

1. List all the tests and randomize them with gnu [shuf](https://www.gnu.org/software/coreutils/manual/html_node/shuf-invocation.html)
1. consistently split these tests into a N configurable number of groups with gnu [split](https://www.gnu.org/software/coreutils/manual/html_node/split-invocation.html)
1. run these groups in N separate jobs, in M parallel processes, using gnu [parallel](https://www.gnu.org/software/parallel/) instead of `paratest`.
1. the output uses [groups](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#grouping-log-lines) so renders nicely with collapsible sections for each test. 
1. similarly, we use `::error::` to make failing tests easy to spot.

Started with 5 jobs with 5 parallel `phpunit` processes each, but can tweak this easily over time.

Runtime should be down to around 5 minutes, which is a much better feedback loop.

Thanks.

### DIFF
--- a/bin/tests-github-actions.sh
+++ b/bin/tests-github-actions.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+set -eu
+
+function get_seeded_random() {
+  openssl enc -aes-256-ctr -pass pass:"vimeo/psalm" -nosalt </dev/zero 2>/dev/null
+}
+
+function run {
+  local -r chunk_count="$1"
+  local -r chunk_number="$2"
+  local -r parallel_processes="$3"
+
+  local -r phpunit_cmd='
+echo "::group::{}";
+vendor/bin/phpunit --log-junit build/phpunit/logs/{_}.xml --colors=always {};
+exit_code=$?;
+echo ::endgroup::;
+if [[ "$exit_code" -ne 0 ]]; then
+    echo "::error::{}";
+fi;
+exit "$exit_code"'
+
+  mkdir -p build/parallel/ build/phpunit/logs/
+
+  find tests -name '*Test.php' | shuf --random-source=<(get_seeded_random) > build/tests_all
+  split --number="l/$chunk_number/$chunk_count" build/tests_all > build/tests_split
+  parallel --group -j"$parallel_processes" --rpl {_}\ s/\\//_/g --joblog build/parallel/jobs.log "$phpunit_cmd" < build/tests_split
+}
+
+if [ -z "${CHUNK_COUNT:-}" ]; then echo "Did not find env var CHUNK_COUNT."; exit 1; fi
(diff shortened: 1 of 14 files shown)


muglug:
Thanks! CircleCI is faster (and I was thinking of ditching phpunit in Github Actions) but this is a great solution

bendavies:
Yeah, the current GA runners have low resources, while circle's have lots.
It's on GAs roadmap to provide more powerful runners, although no idea if OS projects will be able to use them: https://github.com/github/roadmap/issues/95

bin/tests-github-actions.sh
```diff
@@ -0,0 +1,35 @@
...
+exit_code=$?;
+echo ::endgroup::;
+if [[ "$exit_code" -ne 0 ]]; then
+    echo "::error::{}";
+fi;
+exit "$exit_code"'
+
+  mkdir -p build/parallel/ build/phpunit/logs/
+
+  find tests -name '*Test.php' | shuf --random-source=<(get_seeded_random) > build/tests_all
```
Slamdunk, line 26: This doesn't sound right to me: this line shuffles the test based on a seed that is unique for each spawned `jobs.tests`, because `get_seeded_random` gets called indeed inside each of them.

So every `job` shuffles them differently and this lead to some test being run twice or worse, not being run at all.

The `random-source` should be generated only once in the `chuck-matrix` job.

/cc @weirdan @danog

bendavies: the random source is consistent.
you can verfy this youself by running this multiple times
`find tests -name '*Test.php' | shuf --random-source=<( openssl enc -aes-256-ctr -pass pass:"vimeo/psalm" -nosalt </dev/zero 2>/dev/null)`

bendavies: source
https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html#Random-sources

orklah: Yeah, I confirm I always saw error in jobs to be consistent from one run to another

Slamdunk: > the random source is consistent.

This is good news indeed.

I wonder why `shuf` is needed in the first place then :thinking: 
Maybe you know you have some long-running tests and other quick tests, and you want to spread them evenly across the `jobs`? It would be good to write a comment about the why, just to clear the things out

PR #4990, opened 2021-01-12
Optionally allow trait static calls and property access

danog:
This PR optionally allows static method calls and property access on traits.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -37,6 +37,8 @@
         <xs:attribute name="allowFileIncludes" type="xs:boolean" default="true" />
         <xs:attribute name="allowPhpStormGenerics" type="xs:boolean" default="false" />
         <xs:attribute name="allowStringToStandInForClass" type="xs:boolean" default="false" />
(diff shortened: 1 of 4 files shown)


danog:
Hmmm, this introduces a weird issue with typing during property access:
```
<?php
trait a
{
    public static int $a = 0;
}

echo(a::$a);
```

Here, NodeDataProvider::getType(a::$a) returns null, causing a MixedArgument error, but even just adding a `$a = a::$a` seems to correctly populate type information, removing the error:
```
<?php
trait a
{
    public static int $a = 0;
}

$a = a::$a;
echo(a::$a);
```

muglug:
What's the value of allowing this?

danog:
Hi!)
Personally, I don't see any problem with adding this: static trait methods and properties traits behave just as they would in an abstract or normal class, no weird quirks at all.

If traits are used to re-use shared code, I don't see why shouldn't I use static methods when grouping together related logic in a trait.

muglug:
But direct access is just not what traits were designed for, and I don't think Psalm should encourage this (even behind a flag)

Issue #4145, opened 2020-09-07
[feature-request] @psalm-pure-if {type} $param


azjezz:
I would like to see a new annotation added to psalm that allows defining functions as "pure" depending on their argument type.

example:

### before

```php
/**
 * @psalm-template Tk of array-key
 * @psalm-template Tv
 *
 * @psalm-param list<Tv>                    $values
 * @psalm-param (pure-callable(Tv): ?Tk)    $key_func
 *
 * @psalm-return array<Tk, list<Tv>>
 *
 * @psalm-pure
 */
function group_by(array $values, callable $key_func): array
{
    $result = [];
    foreach ($values as $value) {
        $key = $key_func($value);
        if (null === $key) {
            continue;
        }

        $result[$key]   = $result[$key] ?? [];
        $result[$key][] = $value;
    }

    return $result;
}
```

### after

```php
/**
 * @psalm-template Tk of array-key
 * @psalm-template Tv
 *
 * @psalm-param list<Tv>            $values
 * @psalm-param (callable(Tv): ?Tk)	$key_func
 *
 * @psalm-return array<Tk, list<Tv>>
 *
 * @psalm-pure-if (pure-callable(Tv): ?Tk) $key_fun
 */
function group_by(array $values, callable $key_func): array
{
    $result = [];
    foreach ($values as $value) {
        $key = $key_func($value);
        if (null === $key) {
            continue;
        }

        $result[$key]   = $result[$key] ?? [];
        $result[$key][] = $value;
    }

    return $result;
}
```

## Why?

In my case, in PSL, there's a lot of functions that work on arrays, almost all of them are defined as "pure", some take `callable` as an argument.

previously these functions were not considered pure as there was no way to guarantee that the callable itself is pure, however, after #4066, they switched to using `pure-callable`.

but this raises another problem, where the functions cannot be used in a non-pure context, but sometimes their functionality might be needed, and the developer is okay with the call resulting in mutation as their application is not mutation-free.

## Problem

I have a list of users I just fetched from the database, I wanna group these users by their status, but currently, I'm unable to use `Psl\Arr\group_by` as it only works in a mutation-free context where the callable I'm passing is not pure ( e.g fetches the user status from the database? ).

## Solution

`@psalm-pure-if (pure-callable(Tv): ?Tk) $key_fun` helps avoid this issues as it would be okay to pass a non-pure callable to `group_by`, but psalm would consider that call non-pure.

[...]

PR #4999, opened 2021-01-13
#4997 added more precise type inference for `count()` returning `0` or `positive-int` on known arrays

### LINKED ISSUE #4997: `count(array)` should be `positive-int|0`, while `count(Countable)` should be `int`
https://psalm.dev/r/c5d1ba8607

```php
<?php

/** @psalm-return positive-int|0 */
function gimmePositiveCount(array $input): int
{
    return \count($input);
}


/** @psalm-return positive-int|0 */
function gimmeCount(\Countable $input): int
{
    return \count($input); // expected violation here - `Countable#count()` can return negative!
}
```

```
Psalm output (using commit 1afce4d): 

INFO: LessSpecificReturnStatement - 6:12 - The type 'int' is more general than the declared return type 'int(0)|positive-int' for gimmePositiveCount

INFO: MoreSpecificReturnType - 3:19 - The declared return type 'int(0)|positive-int' for gimmePositiveCount is more specific than the inferred return type 'int'

INFO: LessSpecificReturnStatement - 13:12 - The type 'int' is more general than the declared return type 'int(0)|positive-int' for gimmeCount

INFO: MoreSpecificReturnType - 10:19 - The declared return type 'int(0)|positive-int' for gimmeCount is more specific than the inferred return type 'int'
```

In practice, `count()` needs better stubs:

 * `count(array): positive-int|0`
 * `count(Countable): int`

I'll try and see if I can send a patch :-)

Ocramius:
Fixes #4997

~~Note: I only provided the stub, but am unsure about where tests for stubs (if there are any) are applicable.~~

Questions open:

 * [x] `@psalm-flow` - unclear whether this is needed, seems related to taint analysis, so probably not necessary
 * [x] `@psalm-pure` - unclear whether this is to be declared. Purity depends on the given argument: perhaps @azjezz knows more here? See https://github.com/vimeo/psalm/pull/4999#issuecomment-759444684

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
@@ -300,7 +300,7 @@ class FunctionCallReturnTypeFetcher
                                     return new Type\Union([
                                         $atomic_types['array']->count !== null
                                             ? new Type\Atomic\TLiteralInt($atomic_types['array']->count)
-                                            : new Type\Atomic\TInt
+                                            : new Type\Atomic\TPositiveInt
                                     ]);
                                 }
 
@@ -308,7 +308,7 @@ class FunctionCallReturnTypeFetcher
                                     return new Type\Union([
                                         $atomic_types['array']->count !== null
                                             ? new Type\Atomic\TLiteralInt($atomic_types['array']->count)
-                                            : new Type\Atomic\TInt
+                                            : new Type\Atomic\TPositiveInt
                                     ]);
                                 }
 
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
@@ -319,6 +319,11 @@ class FunctionCallReturnTypeFetcher
                                         new Type\Atomic\TLiteralInt(count($atomic_types['array']->properties))
                                     ]);
                                 }
+
+                                return new Type\Union([
+                                    new Type\Atomic\TLiteralInt(0),
+                                    new Type\Atomic\TPositiveInt
+                                ]);
                             }
                         }
                     }
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1087,6 +1087,28 @@ function chunk_split(string $str, int $chunklen = 76, string $end= ''): string
 }
 
 /**
+ * @template TInput as \Countable|array
+ *
+ * @param TInput $value
+ * @param \COUNT_NORMAL|\COUNT_RECURSIVE $mode
+ *
+ * @return (
+ *          TInput is array<empty, empty>
+ *          ? 0
+ *          : (TInput is non-empty-array
+ *              ? positive-int
+ *              : (TInput is array
+ *                  ? positive-int|0
+ *                  : int
+ *              )
+ *          )
+ * )
+ */
+function count(\Countable|array $value, int $mode = \COUNT_NORMAL): int
+{
+}
+
+/**
  * @psalm-pure
  *
  * @psalm-flow ($data) -> return
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1087,24 +1087,27 @@ function chunk_split(string $str, int $chunklen = 76, string $end= ''): string
 }
 
 /**
- * @template TInput as \Countable|array
+ * @template TInput as \Countable|\SimpleXMLElement|\ResourceBundle|array
  *
  * @param TInput $value
- * @param \COUNT_NORMAL|\COUNT_RECURSIVE $mode
+ * @param 0|1 $mode
  *
  * @return (
  *          TInput is array<empty, empty>
  *          ? 0
- *          : (TInput is non-empty-array
- *              ? positive-int
- *              : (TInput is array
- *                  ? positive-int|0
- *                  : int
+ *          : (TInput is array{0: mixed, 1: mixed}
+ *              ? 2
+ *              : (TInput is non-empty-array
+ *                  ? positive-int
+ *                  : (TInput is array
+ *                      ? positive-int|0
+ *                      : int
+ *                  )
  *              )
  *          )
  * )
  */
-function count(\Countable|array $value, int $mode = \COUNT_NORMAL): int
+function count(\Countable|\SimpleXMLElement|\ResourceBundle|array $value, int $mode = 0): int
 {
 }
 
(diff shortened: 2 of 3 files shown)


orklah:
Seems like current CallMap expect `Countable|array|SimpleXMLElement|ResourceBundle` as parameters. SimpleXMLElement is not countable nor array but can be counted through voodoo. Not sure about ResourceBundle though...

orklah:
@psalm-flow is related to the continuity of taint though the function. It shouldn't be needed for count.

stubs/CoreGenericFunctions.phpstub
```diff
@@ -1086,6 +1086,31 @@ function chunk_split(string $str, int $chunklen = 76, string $end= ''): string
...
+ *                  ? positive-int
+ *                  : (TInput is array
+ *                      ? positive-int|0
+ *                      : int
+ *                  )
+ *              )
+ *          )
+ * )
+ */
+function count(\Countable|\SimpleXMLElement|\ResourceBundle|array $value, int $mode = 0): int
```
Ocramius, line 1110: Tempted to mark this `@psalm-pure`, but `@psalm-pure` only applies to `array` as input :-\

tests/FunctionCallTest.php
```diff
@@ -1818,7 +1818,7 @@ function example($x) : void {
                             if ($c !== 2) {}
                         }
                     }',
-                'error_message' => 'TypeDoesNotContainType',
+                'error_message' => 'DocblockTypeContradiction',
```
Ocramius, line 1821: This change is because `TypeDoesNotContainType` was probably generated in psalm internal hardcoding (which can probably be removed by now, given the new stubs), but now `if ($c !== 2) {}` clashes with the return type of `count()` instead.

azjezz:
>  @psalm-pure - unclear whether this is to be declared. Purity depends on the given argument: perhaps @azjezz knows more here?

we can't really declare `count` pure as it depends on the argument type ( #4145 ) ( unless we can do function overrides in stubs )

If `$value` is `array`, `SimpleXMLElement`, or `ResourceBundle`, then `count` is pure.
If `$value` is `Countable` and `$value` is immutable, or `count(): int` is declared with `@psalm-mutation-free`, then `count` is pure
otherwise, `count` is not pure.

stubs/CoreGenericFunctions.phpstub
```diff
@@ -1086,6 +1086,31 @@ function chunk_split(string $str, int $chunklen = 76, string $end= ''): string
 {
 }
 
+/**
+ * @template TInput as \Countable|\SimpleXMLElement|\ResourceBundle|array
+ *
+ * @param TInput $value
+ * @param 0|1 $mode
+ *
+ * @return (
```
Ocramius, line 1095: Also to be handled: `non-empty-countable` type, which I discovered somewhere in tests, and didn't know of before :D

Ocramius:
This block can probably also be refactored away somehow:

https://github.com/vimeo/psalm/blob/1afce4d613b6d7f0856a54eb3441947759e96bae/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php#L286-L326

The new stubs don't seem to contradict it, but I'm not sure how well tested it is right now.

Ocramius:
@orklah `ResourceBundle` was indeed added to the signature, but downstream it fails in real-world projects:

```
ERROR: UndefinedClass - src/Framework/MockObject/Rule/ConsecutiveParameters.php:117:18 - Class or interface ResourceBundle does not exist (see https://psalm.dev/019)
        foreach ($parameters as $i => $parameter) {
```
https://app.circleci.com/pipelines/github/vimeo/psalm/4824/workflows/6f0caadd-ba64-414d-ad4e-d453e9cc6dab/jobs/21060, `git@github.com:muglug/phpunit.git`

azjezz:
> we should really just stop allowing SimpleXMLElement and ResourceBundle :X

👍 for that

Ocramius:
Mutation modifier seems to be computed in psalm's internals:

https://github.com/vimeo/psalm/blob/1afce4d613b6d7f0856a54eb3441947759e96bae/src/Psalm/Internal/Codebase/Functions.php#L401-L421

I think we're OK as-is, if that is still taken into account whilst looking at the stubs.

I'd gladly write test scenarios for these cases, just need someone to point me at an appropriate location where to put them :-)

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
```diff
@@ -300,15 +300,15 @@ private static function getReturnTypeFromCallMapWithArgs(
                                     return new Type\Union([
                                         $atomic_types['array']->count !== null
                                             ? new Type\Atomic\TLiteralInt($atomic_types['array']->count)
-                                            : new Type\Atomic\TInt
+                                            : new Type\Atomic\TPositiveInt
```
Ocramius, line 303: This change is un-tested and requires testing

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
```diff
@@ -300,15 +300,15 @@ private static function getReturnTypeFromCallMapWithArgs(
...
                                     ]);
                                 }
 
                                 if ($atomic_types['array'] instanceof Type\Atomic\TNonEmptyList) {
                                     return new Type\Union([
                                         $atomic_types['array']->count !== null
                                             ? new Type\Atomic\TLiteralInt($atomic_types['array']->count)
-                                            : new Type\Atomic\TInt
+                                            : new Type\Atomic\TPositiveInt
```
Ocramius, line 311: This change is un-tested and requires testing

stubs/CoreGenericFunctions.phpstub
Ocramius: Need feedback based on findings at https://github.com/vimeo/psalm/pull/4999#issuecomment-759444684

orklah:
I believe FunctionCallTest.php is the appropriate location for those. I added tests here in the past (for example for parse_url)

stubs/CoreGenericFunctions.phpstub
```diff
@@ -1086,6 +1086,31 @@ function chunk_split(string $str, int $chunklen = 76, string $end= ''): string
...
+ *          TInput is array<empty, empty>
+ *          ? 0
+ *          : (TInput is array{0: mixed, 1: mixed}
```
Ocramius, line 1098: This match is wrong, since it will also match `array{0: mixed, 1: mixed, 2: mixed}`, which would then result in `2` being inferred.

Ocramius:
@orklah thanks for pushing me in the right direction: I've added detailed tests around `positive-int` and `0` return types of `count()`, as well as its purity properties.

This ended up being much simpler than expected, given that the pre-existing code needed very minor adjustments to operate as expected, so we should be good to review and hopefully merge if the solution is acceptable :+1:

muglug (approved):

muglug:
Thanks!

Issue #5006, opened 2021-01-14
RecursiveArrayIterator MethodSignatureMismatch


franmomu:
https://psalm.dev/r/ea76d92558

Looks like it's new from `4.4.0`.

EDIT: https://psalm.dev/r/9097e8d5fa this with phpdoc

Looks like it works with type declarations: https://psalm.dev/r/0604fa9bea

PR #5014, opened 2021-01-14
remove signature return types in stubs

orklah:
This PR will fix https://github.com/vimeo/psalm/issues/5006

Return types in signatures in stubs are causing an issue because Psalm will check that the return type is present on every child method. 

This should not be an issue for params as params can be wider in childs.


muglug:
Thanks!

Issue #5016, opened 2021-01-14
Narrowing template with array-key (unexpected InvalidArrayOffset)


klimick:
Array cannot be created when key has templated type:
https://psalm.dev/r/a46106b7c2

weirdan:
Simplified: https://psalm.dev/r/0c0660459c

klimick:
very fast) @orklah thakns!

PR #5019, opened 2021-01-14
don't throw InvalidArrayOffset when dealing with templates

orklah:
This PR aims to fix https://github.com/vimeo/psalm/issues/5016

This is probably not fixed the best possible way (I can imagine we could infer template type and check if it's a subtype of array-key) but I'm not quite ready to enter this module in psalm!


muglug:
Thanks!

PR #5030, opened 2021-01-15
--report=codeclimate.json

Danysan1:
Currently it's possible to generate a [codeclimate JSON](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types) output with the `--output-format=codeclimate` flag, but it's impossible to generate directly a report (you must redirect the output because if you use `--report=codeclimate.json` psalm creates a standard JSON).
With these changes it is now possible to generate directly a `codeclimate.json` report with a flag like `--report=codeclimate.json` or `--report=foo.codeclimate.json`.


muglug:
Thanks!

Issue #5021, opened 2021-01-14
RecursiveDirectoryIterator errors on FilesystemIterator constants


staabm:
https://psalm.dev/r/3d66686169

expected: no error

staabm:
Hmm should the constants re-defined here, just be inherited from the parent class?

https://github.com/vimeo/psalm/blob/0e17a3354fda992d7113b97d15c75ff4a92ad800/stubs/CoreGenericIterators.phpstub#L804-L810

weirdan:
It's not about specific constants. You can drop any of the three and it works: https://psalm.dev/r/79233b7288

weirdan:
Actual cause is Psalm's inability to infer literal value for bitwise OR when one of the sides is an expression, even if it qualifies as constexpr: https://psalm.dev/r/84621f04aa

PR #5032, opened 2021-01-16
Calculate more literal values

orklah:
This is an attempt to fix https://github.com/vimeo/psalm/issues/5021

I tweaked a condition slightly and it immediately gave me the value I expected. It seem way to easy to be right but I guess CI will tell me!

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
@@ -295,13 +295,14 @@ class NonDivArithmeticOpAnalyzer
             && $right_type_part instanceof TLiteralInt
             && ($left instanceof PhpParser\Node\Scalar
                 || $left instanceof PhpParser\Node\Expr\ConstFetch
-                || $left instanceof PhpParser\Node\Expr\ClassConstFetch)
+                || $left instanceof PhpParser\Node\Expr\ClassConstFetch
+                || $left instanceof PhpParser\Node\Expr\BinaryOp)
             && ($right instanceof PhpParser\Node\Scalar
                 || $right instanceof PhpParser\Node\Expr\ConstFetch
-                || $right instanceof PhpParser\Node\Expr\ClassConstFetch)
+                || $right instanceof PhpParser\Node\Expr\ClassConstFetch
+                || $right instanceof PhpParser\Node\Expr\BinaryOp)
         ) {
             // time for some arithmetic!
-
             $calculated_type = null;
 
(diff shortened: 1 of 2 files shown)


orklah:
Looks like everything pass, this is cool !

tests/BinaryOperationTest.php
```diff
@@ -196,6 +196,15 @@ function foo(string $s) : int {
                     '$f' => 'string',
                 ],
             ],
+            'ComplexLiteralBitwise' => [
+                '<?php
+                    /**
+                     * @return 7
+                     */
+                    function scope(){
+                        return 1 | 2 | 4;
```
weirdan, line 205: I would suggest to also add a test with `0`, bitwise and (`&`), and brackets.

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
```diff
@@ -295,13 +295,14 @@ private static function analyzeNonDivOperands(
             && $right_type_part instanceof TLiteralInt
             && ($left instanceof PhpParser\Node\Scalar
                 || $left instanceof PhpParser\Node\Expr\ConstFetch
-                || $left instanceof PhpParser\Node\Expr\ClassConstFetch)
+                || $left instanceof PhpParser\Node\Expr\ClassConstFetch
+                || $left instanceof PhpParser\Node\Expr\BinaryOp)
```
weirdan, line 299: Shouldn't this be any expression really? E.g. `(true ? 1 : 2) | 22` would not have `$left` as binary op, but it still can be calculated, I suppose.

weirdan (changes requested):

tests/BinaryOperationTest.php
orklah: Done

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
orklah: Well, we saw a good example of 1 | 2 | 4 in the form of flags and we need Psalm to understand it to be coherent. I'm not sure if it's worth adding complexity and risking edge cases for  `(true ? 1 : 2) | 22`. I'd be open to consider this if we find a real case where it change the outcome of psalm's analysis

weirdan: I can imagine things like ` & ~$this->flagMask `, but you're right, we can fix it when we see it.

weirdan (approved):

muglug:
Thanks!

Issue #5033, opened 2021-01-16
hrtime return type issue on PHP 8.0 if phpVersion is set to < 7.3


kelunik:
https://github.com/amphp/amp/runs/1714053018?check_suite_focus=true#step:9:14

When run on PHP 8.0, an error is reported, because Psalm thinks the return type of `hrtime(false)` is `int` instead of `int[]`. But this only happens if the `phpVersion` in `psalm.xml` is set below `7.3`. Works fine on PHP 7.4.

Unrelated to the hrtime issue, I had to somehow add the function files to `<stubs />` when upgrading to Psalm 4.x for some reason.

```xml
<stubs>
    <file name="lib/functions.php" />
    <file name="lib/Internal/functions.php" />
</stubs>
```

orklah:
@weirdan Can you take at look at this one?

https://psalm.dev/r/fb5c5555dc
https://psalm.dev/r/fb5c5555dc?php=7.2

I think this could be fixed by conditional execution again. The error reported by psalm is not really relevant because the code only use the function when it's available

weirdan:
> I think this could be fixed by conditional execution again.

Should be easy, just make the body of this `if` unconditional:

https://github.com/vimeo/psalm/blob/ef0d19ed9fc66152e48229d4a0d305f19d1a6aa2/src/Psalm/Internal/PhpVisitor/Reflector/ExpressionResolver.php#L243

PR #5034, opened 2021-01-16
cleanup around hrtime

orklah:
This PR propose cleaning up hrtime various data source to fix https://github.com/vimeo/psalm/issues/5033

- The FunctionCallReturnTypeFetcher.php file seem to return the same structure as detailed in Callmap so I removed it
- The hrtime function was introduced in PHP 7.3 so I removed it from Callmap: https://www.php.net/manual/fr/function.hrtime
- The parameter is called `as_number` internally so I updated Callmap_73

I'm not sure why there is this stack of sources where one should be enough. If CI breaks, I'll investigate more


weirdan:
> so I removed it from Callmap

You shouldn't. Callmap contains the data for latest and greatest supported version, and deltas are used to calculate the callmap for previous versions. Since `hrtime()` wasn't removed, it should still be in Callmap.

weirdan:
> The FunctionCallReturnTypeFetcher.php file seem to return the same structure as detailed in Callmap so I removed it

The difference seems to be the `from_calculation` flag. I'm not sure what's its intended meaning and how that affects inference, but it's used in several places in the codebase.

orklah:
> 
> 
> > so I removed it from Callmap
> 
> You shouldn't. Callmap contains the data for latest and greatest supported version, and deltas are used to calculate the callmap for previous versions. Since `hrtime()` wasn't removed, it should still be in Callmap.

oooh, I thought it was the base version and deltas were added on top. Thanks, I'll fix this

orklah:
> The difference seems to be the `from_calculation` flag. I'm not sure what's its intended meaning and how that affects inference, but it's used in several places in the codebase.

IIRC, from_calculation is used to allow some operations on the result type. For example:
`$a = (int) PHP_INT_MAX;` is a redundantCast but `$a = (int) PHP_INT_MAX + 1;` is allowed because the calculation may have changed the type
It's the trick we used for allowing string cast from string offset keys that could have been implicitly casted to int by php

I'm pretty sure it's used here to be able to do a `is_float` on the result while still declaring the type is `int` for some reason

orklah:
I'll close this PR because the issue is not on data sources. If someone stumble here on the future, there is still some small weirdness here and there:
- callmap declare the function can return false but there is no trace of this in FunctionCallReturnTypeFetcher. This should probably be added with a ignoreFalsableIssues flag
- The param name is wrong in callmap_73 (should be `as_number`)
- The `$int = Type::getInt(); $int->from_calculation = true` is weird, it's the only place in the code where we use it given that this is the first parameter of `getInt()`. This could (should?) be replaced by `$int = Type::getInt(true);`

kelunik:
It can only return `false` if it's not available on the given platform and it's a compile time decision.

https://github.com/php/php-src/blob/a06c20a17c97a76956f3454291ba0a46ee39eda1/ext/standard/hrtime.c#L183

Issue #2809, opened 2020-02-13
Detect which method is available in php version?


roukmoute:
With this example:

```php
class DatetimeImmutable extends \DateTimeImmutable
{
    public static function createFromInterface(\DatetimeInterface $datetime): \DatetimeImmutable
    {
        if (\PHP_VERSION_ID >= 80000) {
            return parent::createFromInterface($datetime);
        }
        //…
    }
}
```

psalm returns to me:
```
ERROR: UndefinedMethod - src/Datetime.php - Method DateTime::createfrominterface does not exist
            return parent::createFromInterface($datetime);
```

Is it a bug? Or I must ignore it because psalm can't detect this kind ot things.

roukmoute:
Here it is a link: https://psalm.dev/r/c6f2bbb578

pilif:
somebody (maybe you?) will have to add declarations for methods added to php 8 in a file similar to [CallMap_74_delta.php](https://github.com/vimeo/psalm/blob/dc78e9c57ba58e21e7970451ca52314b398fc811/src/Psalm/Internal/CallMap_74_delta.php), but named `CallMap_80_delta.php`

orklah:
@weirdan I wonder if this was fixed by your recent PR about conditional execution?

weirdan:
@orklah I decided to play it safe and only enabled version-conditional logic during autoload registration. This could probably be allowed to work during other execution phases too, but needs testing.

dkarlovi:
@orklah @weirdan we discussed this yesterday in Symfony Slack, I guess no need to open another ticket. :)

weirdan:
@dkarlovi as a smoke test, would you mind making that code piece I linked to on Slack unconditional and running analysis on your project?

dkarlovi:
@weirdan can you point me to a place where you do it only for files which are in Composer autoload list or something? I've tried dumping `$file_path` on top of the method and the only of my files it reaches is `src/bootstrap.php` (which is only referenced from PHPUnit's config). The other files are similar, 

```
/project/vendor/phpunit/phpunit/src/Framework/Assert/Functions.php
/project/vendor/symfony/deprecation-contracts/function.php
/project/vendor/symfony/polyfill-php80/bootstrap.php
```

etc.

weirdan:
I mean drop the `if()` and make it unconditional here:

[...]

PR #5036, opened 2021-01-16
add analysis execution depending on PHP version in userland

orklah:
This PR aims to implement conditional analysis in userland code.

For example, it would allow using a syntax from PHP 8 even when phpVersion=PHP 7.4 as long as there is a PHP_VERSION_ID check.

EDIT: it also adds DateTime::createFromInterface in PHP8 callmap

EDIT: forgot to say, it should fix https://github.com/vimeo/psalm/issues/2809 and fix https://github.com/vimeo/psalm/issues/5033

### DIFF
--- a/tests/ConditionalExecution.php
+++ b/tests/ConditionalExecution.php
@@ -1,0 +1,97 @@
+<?php
+namespace Psalm\Tests;
+
+use UnexpectedValueException;
+use function preg_quote;
+use Psalm\Config;
+use Psalm\Context;
+use Psalm\Internal\RuntimeCaches;
+use Psalm\Tests\Internal\Provider;
+use function strpos;
+use function substr;
+
+class ConditionalExecution extends TestCase
+{
+    /** @var \Psalm\Internal\Analyzer\ProjectAnalyzer */
+    protected $project_analyzer;
+
+    public function setUp() : void
+    {
+        RuntimeCaches::clearAll();
+
+        $this->file_provider = new Provider\FakeFileProvider();
+
+        $this->project_analyzer = new \Psalm\Internal\Analyzer\ProjectAnalyzer(
+            new TestConfig(),
+            new \Psalm\Internal\Provider\Providers(
+                $this->file_provider,
+                new Provider\FakeParserCacheProvider()
+            )
+        );
+    }
+
+    /**
+     * @dataProvider providerValidCodeParse
+     *
+     * @param string $code
+     * @param array<string> $error_levels
+     */
+    public function testValidCode($code, array $error_levels = []): void
+    {
+        $test_name = $this->getTestName();
+        if (strpos($test_name, 'SKIPPED-') !== false) {
+            $this->markTestSkipped('Skipped due to a bug.');
+        }
+
+        $version_position = strpos($test_name, 'PHP-');
+        if ($version_position === false) {
+            throw new UnexpectedValueException('No PHPVersion in test name');
+        }
+
+        $version = substr($test_name, $version_position + 4, 3);
+        $this->project_analyzer->setPhpVersion($version);
+
+        $file_path = self::$src_dir_path . 'somefile.php';
+
+        $this->addFile(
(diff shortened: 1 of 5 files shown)


orklah:
cc @weirdan This seems to work! However when running the tests locally (on windows) I got different errors for the two tests. I think this is broken on windows... It will have to be fixed if someone care enough.

weirdan:
8.0 *syntax* won't necessarily be supported in conditional blocks - it will (and should) fail at parser level.

weirdan:
It works in AppVeyor, so the issue could be with your local setup

orklah:
Weird, but it's better that way!

And yeah, I used the word syntax abusively. I meant symbols

tests/ConditionalExecution.php
```diff
@@ -0,0 +1,94 @@
...
+
+                    class Datetime extends \DateTime
+                    {
+                        public static function createFromInterface(\DatetimeInterface $datetime): \DateTime
+                        {
+                            if (\PHP_VERSION_ID >= 80000) {
+                                return parent::createFromInterface($datetime);
+                            }
+                        }
+                    }'
```
weirdan, line 90: I'm surprised this doesn't fail on 7.4 because it should: https://3v4l.org/8mpLv#v730

orklah: Mmh you make me second guess myself! It's the whole point of this PR, it doesn't fail in PHP 7.4 because the body of the if is behind a PHP_VERSION check, no?

orklah: oh, I see your point, there is no return anymore in this function

orklah: I removed the return type that wasn't here in the original code (I added to to remove an unrelated Psalm issue). However, you're right, in normal conditions, Psalm throw a InvalidReturnType : `Not all code paths of FooBar\Datetime::createFromInterface end in a return statement, return type DateTime expected`

weirdan: Yeah, and I wonder why it did not do that here.

orklah: I'm not sure I understood exactly what your modification does. Can you explain quickly?

weirdan: My understanding of that code is that returning `false` from `ExpressionResolver::enterConditional()` makes Psalm consider the `if` body as 'uninteresting', and then it doesn't do some of the things it does while walking the AST during scanning. The second part of the changes was to make Psalm omit scan of included files when in 'uninteresting' code.

Looking at the code, though, it doesn't seem like that flag affects method calls in any way. Does the test you added fail on master, i.e. does PR actually changes anything for method calls?

orklah:
all the tests pass? I'm not sure what's going on. I'll work on debugging that tomorrow

weirdan:
> all the tests pass? I'm not sure what's going on. I'll work on debugging that tomorrow

I guess you need to rename the test file and class to end with `Test.php` and `Test` respectively:
image

orklah:
Oof, I'm a moron. That explains why the tests failed only locally as I was running only the file

orklah:
Ok, so now it fails with
"Cannot access array value on non-array variable  of type int"
"Could not infer a return type"
As expected when running a code on a precedent version. I'll put back the modification now

orklah:
Aaand it fails exactly the same way with it

orklah:
I'll close this. It's not as easy as I thought (and I think this is will get even harder when I try to debug this).

I'm not sure this can just be solved by saying Psalm not to read some code (because it will just miss everything that code does)

For example if a method return string in PHP 8 and int in PHP 7 based on a PHP_VERSION_ID check, analysis in both version will have a partial understanding on what the code could do and this is likely to create some issue even outside of the PHP_VERSION_ID check.

This means that the best course of action would probably be to change the phpVersion on the fly to be able to understand the code as it will be run. And I suspect this will be incredibly hard and mess with caches and such.

PR #5042, opened 2021-01-18
Check iterator types on yield from

### LINKED ISSUE #4289: Psalm does not error when yielding from a non-generator or Traversable
https://psalm.dev/r/f07be189f6

danog:
Fixes #4289

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
@@ -341,7 +341,7 @@ class ForeachAnalyzer
      */
     public static function checkIteratorType(
         StatementsAnalyzer $statements_analyzer,
-        $stmt,
+        PhpParser\NodeAbstract $stmt,
         PhpParser\Node\Expr $expr,
         Type\Union $iterator_type,
         Codebase $codebase,
--- a/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
@@ -215,6 +215,7 @@ class ForeachAnalyzer
             if (self::checkIteratorType(
                 $statements_analyzer,
                 $stmt,
+                $stmt->expr,
                 $iterator_type,
                 $codebase,
                 $context,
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
```diff
@@ -335,11 +336,13 @@ public static function analyze(
     }
 
     /**
+     * @param PhpParser\Node\Stmt\Foreach_|PhpParser\Node\Expr\YieldFrom $stmt
      * @return false|null
      */
     public static function checkIteratorType(
         StatementsAnalyzer $statements_analyzer,
-        PhpParser\Node\Stmt\Foreach_ $stmt,
+        $stmt,
```
orklah, line 344: This could be declared NodeAbstract

danog: Right, of course!

orklah:
I'm not sure I see the logic in sending a Yield to ForeachAnalyzer. Can you elaborate why please?

danog:
Yield from is an expression that accepts an iterable as only parameter, forwarding all iterable values to the consumer of the generator.
It is often used to implement stackful coroutines, allowing suspension of the call stack of multiple nested generators.

These two snippets are functionally equivalent:

```
function a(iterable $it): \Generator
{
    foreach ($it as $key => $value) {
        yield $key => $value;
    }
    if ($it instanceof \Generator) {
        $res = $it->getReturn();
    }
}
```
---
```
function a(iterable $it): \Generator
{
    $res = yield from $it;
}
```

Since a yield from is functionally equivalent to a foreach, it makes sense to use the exact same code; if you think that the shared logic should be extracted somewhere else, no problem)

tests/Template/ClassTemplateTest.php
```diff
@@ -2038,11 +2038,6 @@ class A {
                         public function __construct(Foo $v) {
                             $this->vector = $v;
                         }
-
```
muglug, line 2041: Why are you removing this? The test isn't testing anything proper without this yield

danog: Foo is not an iterable, and cannot be yielded from.
In fact, if we try to iterate over the Generator returned by this method, "Can use "yield from" only with arrays and Traversables" will be returned.

muglug: Oh ok, then would you mind making `Foo` something that can be yielded from? Or skip the test entirely my renaming it to `SKIPPED-<testname>`

danog: Done!

muglug:
Thanks!

PR #5050, opened 2021-01-19
Fix for #5049 removing `true` or `false` from a `Union` should turn `bool` into respectively `false` or `true`

### LINKED ISSUE #5049: Regression: `@psalm-assert !empty` with `bool` does not reduce it to `true`
While writing type refinements for `webmozart/assert`, I found a regression between `vimeo/psalm@61f5a06` (61f5a06) and `vimeo/psalm:4.4.1` (9fd7a7).

https://psalm.dev/r/6d6857dcf8

Specifically:

```php
/** @psalm-assert !empty $input */
function assertNotEmpty(mixed $input): void { /* ... */ }

/** @psalm-return true */
function gimmeTrue(bool $input): bool {
    assertNotEmpty($input);
    
    return $input;
}
```

Produces:

```
Psalm output (using commit 2f58c6a): 

ERROR: InvalidReturnStatement - 10:12 - The inferred type 'bool' does not match the declared return type 'true' for gimmeTrue

ERROR: InvalidReturnType - 6:19 - The declared return type 'true' for gimmeTrue is incorrect, got 'bool'
 
```

When given a `bool` `$input`, I would expect this to infer `true` as result of the assertion: this used to work in `3.x` (61f5a06).

The opposite works:

```php
/** @psalm-assert empty $input */
function assertEmpty(mixed $input): void { /* ... */ }

/** @psalm-return false */
function gimmeTrue(bool $input): bool {
    assertEmpty($input);
    
    return $input;
}
```

https://psalm.dev/r/ff0756f05d

### LINKED ISSUE #5094: Bug: string(stdClass) can never contain class-string on compare get_class with string
```php
function isStdClass(object $o): bool {
    return get_class($o) !== "stdClass";
}

isStdClass(new \stdClass());
```

Result: `string(stdClass) can never contain class-string`. But should be pass.

Example: https://psalm.dev/r/6769afb49b

Ocramius:
Fixes #5049

### DIFF
--- a/src/Psalm/Type/Union.php
+++ b/src/Psalm/Type/Union.php
@@ -560,6 +560,12 @@ class Union implements TypeNode
                 unset($this->types[$literal_key]);
             }
             $this->literal_float_types = [];
+        } elseif ($type_string === 'false' && isset($this->types['bool'])) {
+            unset($this->types['bool']);
+            $this->addType(new Type\Atomic\TTrue());
+        } elseif ($type_string === 'true' && isset($this->types['bool'])) {
+            unset($this->types['bool']);
+            $this->addType(new Type\Atomic\TFalse());
         }
 
         return false;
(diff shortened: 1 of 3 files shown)


muglug:
Fixed separately in https://github.com/vimeo/psalm/commit/f1107ab4d925c095ca4f6388a1edc5046468b4d7, but I appreciate the contribution regardless

Ocramius:
@muglug careful: did you also consider `TFalse` in that? I had added unit tests for the `Union` type on purpose there :D 

Thanks anyway for the prompt action!

Issue #5048, opened 2021-01-19
Fatal error during initial scan in Lumen


mabasic:
Because `laravel/lumen-framework` is requiring `illuminate/queue` which has a class `CallQueuedClosure` which uses a trait `Illuminate\Foundation\Bus\Dispatchable` which is only available in `illuminate/foundation` which is not required in `laravel/lumen-framework` we get the error bellow when running `./vendor/bin/psalm --init`.

```
Calculating best config level based on project files
Calculating best config level based on project files
Scanning files...
PHP Fatal error:  Trait 'Illuminate\Foundation\Bus\Dispatchable' not found in /.../blog/vendor/illuminate/queue/CallQueuedClosure.php on line 14
Fatal error: Trait 'Illuminate\Foundation\Bus\Dispatchable' not found in /.../blog/vendor/illuminate/queue/CallQueuedClosure.php on line 14
```

There is a PR/Issue on `laravel/framework` with more info on this: https://github.com/laravel/framework/pull/35399

To get the initial scan to complete I had to manually remove that missing trait from `CallQueuedClosed`, run the command `psalm --init` and then restore the file `CallQueuedClosed.php` as it was.

Any ideas on how can this be solved so that it does not throw a fatal error on initial run?

weirdan:
@mabasic Please include the 'steps to reproduce' section from the linked lumen PR.

mabasic:
Steps to reproduce:

```
composer create-project --prefer-dist laravel/lumen blog
cd blog
composer require --dev vimeo/psalm
./vendor/bin/psalm --init
```

mabasic:
@weirdan Thank you. It works now.

PR #5054, opened 2021-01-19
Use composer autoloader during init

weirdan:
Fixes vimeo/psalm#5048

The issue was caused by missing dependency of a class that Psalm tried to inspect through reflection. Fixed by using composer autoloader which avoids the need to use reflection.


muglug:
Thanks!

Issue #5038, opened 2021-01-17
PHP8 false positive MixedAssignment error for "createFromInterface" methods


shandyDev:
Hello,
Latest 4.4.1 psalm version & PHP8 environment.
This code - https://psalm.dev/r/a3d4cdf5a0 show
> MixedAssignment: Cannot assign $this->date to a mixed type 

This is psalm bug? 

P.S. It seems it's not related to https://github.com/vimeo/psalm/issues/2809 & https://github.com/vimeo/psalm/pull/5036

orklah:
> P.S. It seems it's not related to #2809 & #5036

Not directly, but #5036 contains the fix anyway.

If you want you can take the 2 lines in CallMap and CallMap_80_delta and propose a separate PR to fix this issue.

We have difficulties with #5036 so I'm not sure it will be ready before next release.

shandyDev:
Then #5036 not fully.
First: method createFromInterface have in both classes - DateTimeImmutable & DateTime
Secondly: Return type detected wrong way (it is should not just DateTimeInterface and should be hard "self" reference)

orklah:
Thanks! Care to push the PR?

PR #5055, opened 2021-01-19
Add createFromInterface in PHP 8 

orklah:
This should fix https://github.com/vimeo/psalm/issues/5038


muglug:
Thanks!

orklah:
Not sure what went wrong here: https://psalm.dev/r/a3d4cdf5a0

I'll investigate tomorrow

dictionaries/CallMap.php
```diff
@@ -1774,6 +1774,7 @@
 'DateTime::add' => ['static', 'interval'=>'DateInterval'],
 'DateTime::createFromFormat' => ['static|false', 'format'=>'string', 'time'=>'string', 'timezone='=>'?DateTimeZone'],
 'DateTime::createFromImmutable' => ['static', 'datetTimeImmutable'=>'DateTimeImmutable'],
+'DateTime::createFromInterface' => ['DateTimeInterface', 'self'],
```
weirdan, line 1777: Shouldn't it be in the reverse order? `DateTime::createFromInterface()` returns `self` and accepts `DateTimeInterface`.

orklah: ouch, thanks! I got confused because first I thought the method was defined in DateTimeInterface and returned static. I fixed the wrong way

pilif: it's also missing the parameter name, causing 

```
PHP Warning:  Trying to access array offset on value of type int in src/Psalm/Internal/Codebase/InternalCallMapHandler.php on line 266
```

I have submitted #5138 to correct this.

Issue #5047, opened 2021-01-19
Psalm throwing an InvalidArgumentException


VincentLanglet:
I have a git hook which runs lint/phpstan/psalm on staged file before every commit.
When running it I'm getting
```
Uncaught InvalidArgumentException: Could not get class storage for phpunit\framework\testcase in /var/www/vagrant.assurancevie.com/current/symfony/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:42
Stack trace:
#0 /var/www/vagrant.assurancevie.com/current/symfony/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Populator.php(122): Psalm\Internal\Provider\ClassLikeStorageProvider->get('phpunit\\framewo...')
#1 /var/www/vagrant.assurancevie.com/current/symfony/vendor/vimeo/psalm/src/Psalm/Codebase.php(466): Psalm\Internal\Codebase\Populator->populateCodebase()
#2 /var/www/vagrant.assurancevie.com/current/symfony/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(1169): Psalm\Codebase->scanFiles(1)
#3 /var/www/vagrant.assurancevie.com/current/symfony/vendor/vimeo/psalm/src/psalm.php(678): Psalm\Internal\Analyzer\ProjectAnalyzer->checkPaths(Array)
#4 /var/www/vagrant.assurancevie.com/current/symfony/vendor/vimeo/psalm/psalm(2): require_once('/var/www/vagran...')
#5 {main}
(Psalm 4.3.2@57b53ff26237074fdf5cbcb034f7da5172be4524 crashed due to an uncaught Throwable)
```

If I run a full psalm run to populate the cache, the git hook is working fine after.

It might be related to the fact I don't have `phpunit` in my vendor but use a custom loader (with the symfony phpunit bridge)
```
autoloader="bin/.phpunit/phpunit/vendor/autoload.php"
```

Does a try/catch should be added in Populator l122 ; I see in other methods that it's often the case @muglug

weirdan:
Can you provide Psalm command you're using in your git hook?

VincentLanglet:
Sure, I'm using https://github.com/okonet/lint-staged

My config is 
```
[...]

PR #5061, opened 2021-01-20
Catch InvalidArgumentException

VincentLanglet:
Fix https://github.com/vimeo/psalm/issues/5047

I tried in my project and it solves the issue.
I don't know why this invalidArgumentException is thrown but it works well if psalm ignores it.


muglug:
That works – thanks!

Issue #5056, opened 2021-01-19
how to debug: "Child terminated with return code 0 and signal 9" and "Could not get class storage"


michabbb:
hi there,

while trying to scan a whole project, I got the error:

`Uncaught Exception: Could not get class storage for kint`
 [here's the output](https://gist.github.com/michabbb/74e97bd2358521e73d1139bb301f8127).
(Kint can be found [here](https://github.com/kint-php/kint))

because I don´t know how to fix this, I just tried to ignore "Kint" at this point: 
`vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Properties.php` and added:

```php
/** in method propertyExists */
if ($fq_class_name==='Kint') {
  return false;
}
```

just to see, what happens. with that change I got the error:

```
Scanning files...
Analyzing files...

E░░EE░░░░░E░░░░E░EE░EEEE░░E░░░E░E░EEE░E░EEEEEEEEEEEEEEEEE░EE  60 / 791 (7%)
░EEEEEEEEEEEEEEEEEEE░EEEEEEEEEEEEEEE░EEEE░░E░EE░E░░EEEEE░EEE 120 / 791 (15%)
EEEEEEEEEEEEEEEEEEEEEEEE░E░EEE░EEEEEEEEEEE░E░E░░░E░E░░░░E░EE 180 / 791 (22%)
EE░EEE░░EEEEEEEEEEEEEEEEEEEE░E░░░░EEEEEE░░░░░░░░EEEEE░░E░░░E 240 / 791 (30%)
EEE░EEEE░EEEEEEEEEEEEEEEEEEEEEEEEEEE░░E░E░EEEE░░EEEEEEEEEEE░ 300 / 791 (37%)
E░░░░░░EEEEEEEEEEE░░EEEEEE░░░E░EEE░EE░░░░░░░EEEEEE░░EE░░░EEE 360 / 791 (45%)
░EEE░░EEEEEEEEEEEEEEEEEEE░EEEEEE░░E░░░EEEE░░░░░░░░EEEEE░░EE░ 420 / 791 (53%)
░░░EEE░E░E░EEEEEEEEEEEEEEEEEE░EEEEEEEEE░EE░EE░E░E░░░░░░░EEEE 480 / 791 (60%)
E░░EE░░░░░░E░EEE░EEEEEE░EEEEEEEEEEE░EEEEEEEEE░EEEEEEEEE░░░░░ 540 / 791 (68%)
[...]

PR #5068, opened 2021-01-20
Resolve class aliases when accessing properties

weirdan:
Fixes vimeo/psalm#5056

### DIFF
--- a/src/Psalm/Internal/Codebase/Properties.php
+++ b/src/Psalm/Internal/Codebase/Properties.php
@@ -11,6 +11,7 @@ use Psalm\Internal\Provider\PropertyExistenceProvider;
 use Psalm\Internal\Provider\PropertyTypeProvider;
 use Psalm\Internal\Provider\PropertyVisibilityProvider;
 use Psalm\StatementsSource;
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Codebase/Properties.php
```diff
@@ -305,4 +316,18 @@ public function getPropertyType(
 
         return null;
     }
+
+    /**
+     * Returns ClassLikeStorage for the given class name, resolving aliases
+     */
+    private function getClasslikeStorage(string $fq_class_name): ?ClassLikeStorage
```
muglug, line 323: Might it make sense to move this into ClassLikeStorageProvider like `getSafe` or such?

weirdan: The method uses `Classlikes::getUnAliasedName()` so if we move it the provider would have to depend on `Classlikes` which itself depends on `ClassLikeStorageProvider`. Such circular dependency would complicate initialization of both.

weirdan: Done

muglug:
Thanks!

PR #5080, opened 2021-01-22
Add option to dump taint graph

adrienlucas:
This feature is complementary to the `--report` option, as it allows to debug whenever taints are not detected (thus not reported in SARIF).


muglug:
Thanks!

PR #5084, opened 2021-01-22
Remove file map cache when re-analyzing files

joehoyle:
When the project is re-analyzed under the single thread condition (not using a process pool), the filemaps are not cleared before re-analyzing files. This means that file maps only get appended to. If you delete the contents of a file via the LSP, the file map will still be populated with all the old values for example.


muglug:
Thanks!

muglug:
Ok actually this needs to be slightly more nuanced – there's existing logic that updates the reference map when code moves around. That code also needs to account for removing methods.

joehoyle:
@muglug ahh totally missed that behavior. Thanks for handling.

PR #5085, opened 2021-01-22
Support completions on class references

joehoyle:
This provides completions on class references (as opposed to initiated objects via the type map), so you can do `MyClass::` and get completitions for static methods and constants etc.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1286,6 +1286,12 @@ class Codebase
             if ($offset - $end_pos === 2 && substr($file_contents, $end_pos, 2) === '::') {
                 return [$possible_reference, '::', $offset];
             }
+
+            // Only continue for references that are partial / don't exist.
+            if ($possible_reference[0] !== '*') {
+                continue;
+            }
+
             if ($offset - $end_pos === 0) {
                 $recent_type = $possible_reference;
 
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1279,10 +1279,13 @@ class Codebase
         }
 
         foreach ($reference_map as $start_pos => [$end_pos, $possible_reference]) {
-            if ($offset < $start_pos || $possible_reference[0] !== '*') {
+            if ($offset < $start_pos) {
                 continue;
             }
-
+            // If the reference precedes a "::" then treat it as a class reference.
+            if ($offset - $end_pos === 2 && substr($file_contents, $end_pos, 2) === '::') {
+                return [$possible_reference, '::', $offset];
+            }
             if ($offset - $end_pos === 0) {
                 $recent_type = $possible_reference;
 
--- a/src/Psalm/Internal/Codebase/Analyzer.php
+++ b/src/Psalm/Internal/Codebase/Analyzer.php
@@ -1168,7 +1168,7 @@ class Analyzer
     }
 
     /**
-     * @param string $reference The symbol name for the reference. Prepend with an asterisk (*) SOME REASON!.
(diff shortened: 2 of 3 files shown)


src/Psalm/Internal/Codebase/Analyzer.php
```diff
@@ -1167,6 +1167,9 @@ public function addNodeArgument(
         ];
     }
 
+    /**
+     * @param string $reference The symbol name for the reference. Prepend with an asterisk (*) SOME REASON!.
```
joehoyle, line 1171: I've tried to document this, but I didn't work out what the `*` is meant to _mean_!

src/Psalm/Codebase.php
```diff
@@ -1279,10 +1279,13 @@ public function getCompletionDataAtPosition(string $file_path, Position $positio
         }
 
         foreach ($reference_map as $start_pos => [$end_pos, $possible_reference]) {
-            if ($offset < $start_pos || $possible_reference[0] !== '*') {
```
joehoyle, line 1282: As mentioned below, I'm not sure why this is only allowing refs that are tagged with `*` at the start!

joehoyle:
Looks like I broke [`testCompletionOnThisPropertyWithAnotherCharacter`](https://github.com/vimeo/psalm/blob/211553c53f09fa278946fda37b99c2f8ad1c1cbd/tests/LanguageServer/CompletionTest.php#L221) though I'm not sure I understand this test. Why should it expect no completions when adding a second valid character that is part of the method name?

src/Psalm/Internal/Codebase/Analyzer.php
muglug: It's a wildcard I think?

muglug:
> Looks like I broke testCompletionOnThisPropertyWithAnotherCharacter though I'm not sure I understand this test. Why should it expect no completions when adding a second valid character that is part of the method name?

I'm not sure why that test exists now. It could be that whatever client I was using at the time got confused by a more-refined list once another character was entered? Feel free to delete that test, or amend it with expected values.

src/Psalm/Internal/Codebase/Analyzer.php
joehoyle: The only places I see it used are https://github.com/vimeo/psalm/blob/f9fccb2b2dafacc83bf6c43c24d09aa8a589446c/src/Psalm/Internal/Analyzer/ClassAnalyzer.php#L1949 and https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/ClassAnalyzer.php#L2314 which appears to signify that the reference does not exist in storage. The only place I found that makes use / acknowledges the asterisk is https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Codebase/ClassLikes.php#L237 which ignores it.

joehoyle: Looking at https://github.com/vimeo/psalm/blob/master/tests/LanguageServer/CompletionTest.php#L656, I think it is meant to mean "doesn't exist", in this unit test, `Ex` of course doesn't exist, so the symbols is `*Ex`

joehoyle:
> I'm not sure why that test exists now. It could be that whatever client I was using at the time got confused by a more-refined list once another character was entered? Feel free to delete that test, or amend it with expected values.

Ahh this was a genuine bug, fixed up in c6b02d6b1dea9ce963a3d200d8890bf9736922f1

muglug:
Thanks!

yaegassy:
@joehoyle Great! I look forward to further "completion" feature to psalm-language-server!

The [pyright](https://github.com/microsoft/pyright) (for Python) was started as a type-linter project and now also has a language server that includes a "completion" feature.

I think `psalm` has the same potential.

Issue #5082, opened 2021-01-22
is_callable is not taken into account


transistive:
Hello,

Psalm is giving me a hard time in a case resembling this snippet: https://psalm.dev/r/583d409fde.

I am trying to accept a lazy-loaded variable which is possibly wrapped around a closure. If I test it with is_callable, it still complains the type might not be a callable. I don't know what causes this.

Kind regards,
Ghlen

transistive:
Hello Weirddan,

Thank you for the swift response! I understand why that is not callable but isn't that what `is_callable` is for? How can Psalm think it not to be a valid callable, right after I test it to be callable?

Kind regards,

Ghlen

orklah:
technically, `string(test)` could be the supertype of `class-string<test>` so it may very well be callable. A little unorthodox though...

transistive:
I still don't fully understand. If I follow this logic correctly, the key within the array `["test" => "foobar"]` is possibly callable, but how does that make the array callable? Furthermore, how does psalm essentially interprets the result of `is_callable($x)` to be a false positive?

orklah:
nope sorry, possible callable would be  `array{string(test), string}`, not  `array<string(test), string>`. I misread your code and I'm not sure I actually understand what's going on

transistive:
Thank you. To me, it just looks like a false positive. The example works in simpler cases such as in the second function in this snippet: https://psalm.dev/r/5eece5d7da.

orklah:
Can you explain why the return type of the callable is `(X|null)|X|null` ?

Psalm is doing weird things around that:
with X instead, it's a redundant is_callable:
https://psalm.dev/r/0ae0ab2322
with X|null, it becomes OK:
https://psalm.dev/r/54ec32ea26
with X|null|(X|null) your issue is back:
https://psalm.dev/r/efea950ba4
But it's not due to parenthesis because X|null|(string) is ok:
https://psalm.dev/r/fefad7a0c4

@weirdan I believe we can reopen this, there's something fishy here

transistive:
Hey Orklah,

Thank you for helping. I am using this callable structure for lazy evaluation. On the other hand, the value is also optional, hence the callable():(X|null)|X|null. My library is customisable through injections, but these injections can be quite heavy, so I am using the lazy evaluation pattern. Sometimes it seems a bit like overkill, but for some injections, it is not. So for the sake of consistency, all injections can be lazily injected.

[...]

PR #5086, opened 2021-01-22
only transform a Keyed array into callable-array if there's two elements

orklah:
This will prevent Psalm from transforming a random array into a callable-array. It does that by checking an array has exactly two elements

This will solve https://github.com/vimeo/psalm/issues/5082


muglug:
Thanks!

PR #5087, opened 2021-01-22
Use separate exit code to indicate Psalm finding issues

weirdan:
This will allow to distinguish successful run that found some issues from crash (that results in exit code=1).


ohader:
Looks good. What do you think about documenting the different exit codes e.g. at https://github.com/vimeo/psalm/blob/master/docs/running_psalm/command_line_usage.md?

weirdan:
@ohader good point, done.

muglug:
Thanks!

Issue #5089, opened 2021-01-23
ReflectionType does not have the getName method


vudaltsov:
But Psalm does not complain: https://psalm.dev/r/08daec2e9d
See https://www.php.net/manual/ru/class.reflectiontype.php

JesusValera:
The issue is now working again :)

PR #5090, opened 2021-01-23
Remove ReflectionType::getName(), move ReflectionType::isBuiltin()

weirdan:
* `ReflectionType::getName()` was never there: https://3v4l.org/1iZt4
* `ReflectionType::isBuiltin()` was moved to `ReflectionNamedType::isBuiltin()` in PHP 8

Fixes vimeo/psalm#5089


weirdan:
CircleCI failure seems unrelated

orklah:
> CircleCI failure seems unrelated

Are you sure? It seems pretty close:
> Method ReflectionType::getName does not exist

weirdan:
@orklah There was a different failure before I restarted the build job.

weirdan:
Current failure is legit as the code in question fails with methods returning union types: https://3v4l.org/8gt34E
This is already fixed in original package: https://github.com/Ocramius/ProxyManager/commit/13c118e77b927a56c06fe7824e921954c8261a93

muglug:
Thanks, I'll update my fork of that

PR #5098, opened 2021-01-24
Update documentation for taints and global configuration

ohader:
* [DOCS] Add documentation for @psalm-taint-unescape
* [DOCS] Add documentation for conditional @psalm-taint-escape
* [DOCS] Synchronize meaning of @psalm-taint-source input with source code
* [DOCS] Extend documentation on global variables configuration


muglug:
Thanks!

PR #5100, opened 2021-01-24
Fix CallMap of PDOStatement for PHP8

sj-i:
- Fixes https://psalm.dev/r/57f8bc0be4
- The definitions are based on the php-src stub. https://github.com/php/php-src/blob/4bbb98c24fa691c176e403ab00a2fe3dd4224b3e/ext/pdo/pdo_stmt.stub.php
	- The example above is valid on the runtime type checking. https://3v4l.org/ZXfT2
- The internal definitions are changed as of PHP 8.0, though the manual doesn't reflect it. 
	- e.g. https://3v4l.org/GRY5F


dictionaries/CallMap_80_delta.php
```diff
@@ -286,6 +286,17 @@
 'mktime' => ['int|false', 'hour='=>'int', 'minute='=>'int', 'second='=>'int', 'month='=>'int', 'day='=>'int', 'year='=>'int'],
 'parse_str' => ['void', 'encoded_string'=>'string', '&w_result='=>'array'],
 'password_hash' => ['string|false|null', 'password'=>'string', 'algo'=>'?string|?int', 'options='=>'array'],
+'PDOStatement::bindColumn' => ['bool', 'column'=>'mixed', '&rw_var'=>'mixed', 'type='=>'int', 'maxLength='=>'int', 'driverOptions='=>'mixed'],
```
weirdan, line 289: This would need corresponding changes added to `Callmap.php` as well.

sj-i: ... and the fix was wrongly added to 'old' only. Fixed!

muglug:
Thanks!

PR #5105, opened 2021-01-25
Completions for array keys and type literals

joehoyle:
This adds two new additions to completions support: known array-key string literals, and valid values as completions for argument types. For example, any parameter that has `bool`, ints, literal strings or constants will be provided as completions.

This requires broadening the completions triggers to `(`, `[` and ` `. From what I can see this is the same for TypeScript, which is what I've been using for completion comparisons.

For known-literals for types in arguments, it's show like so:

For known-array keys it's shown like so:

This is still marked as draft as I have an issue with the array-key completion. Currently the reference map for the file will hold `array{bar: int(2), foo: int(1)}` for the above screenshot if the type annotation is not added. `array{bar: int(2), foo: int(1)}` is not a valid value for `Type::parseString()`, so I don't see a way to be able to reverse the type "annotation" in the reference map to a type object (and then be able to pull out the array-keys).

Perhaps `array{bar: int(2), foo: int(1)}` could become a valid type string, or perhaps it could make sense to change the format of `array{bar: int(2), foo: int(1)}` to something that's reversible to a type string too? Failing that, I think I'd need to track additional information in the reference map to be able to work back to a valid type object.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1273,7 +1273,7 @@ class Codebase
                         return null;
                     }
 
-                    return [$recent_type, '[', $offset];
+                    return [$recent_type, $gap, $offset];
                 }
             }
 
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1230,7 +1230,7 @@ class Codebase
     }
 
     /**
-     * @return array{0: string, 1: '->'|'::'|'symbol', 2: int}|null
+     * @return array{0: string, 1: '->'|'::'|'['|'symbol', 2: int}|null
      */
     public function getCompletionDataAtPosition(string $file_path, Position $position): ?array
     {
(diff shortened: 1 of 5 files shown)


muglug:
Yes, I'll make `int(1)` a drop-in replacement for `1`

muglug:
Actually, thinking about it, there's no good reason why `getId()` shouldn't just return `1`. I'll move in that direction instead

joehoyle:
@muglug sounds good, I think this applies to all scalar literals too

muglug:
Ok, feel free to merge/rebase master

joehoyle:
@muglug to ready for review!

src/Psalm/Codebase.php
```diff
@@ -1499,6 +1574,90 @@ public function getCompletionItemsForPartialSymbol(
...
+                $completion_items[] = new \LanguageServerProtocol\CompletionItem(
+                    $atomic_type->value,
+                    \LanguageServerProtocol\CompletionItemKind::VALUE,
+                    $atomic_type->getId(),
+                    null,
+                    null,
+                    null,
+                    "'$atomic_type->value'",
+                );
+            } elseif ($atomic_type instanceof Type\Atomic\TLiteralFloat
```
muglug, line 1607: Mind removing the float option? I don't think specifying a `float` is a common occurrence in an enum scenario, and I'd rather Psalm not support it (also, more practically, I only fixed the `getId` types for string and int).

muglug: I think this broke a test

joehoyle:
My bad, didn't update the test for floats.

muglug:
Thanks!

PR #5115, opened 2021-01-27
Add constant fetch to reference map

joehoyle:
To support showing constant types on hover of constant references, we need to add them to the ref map.

I'm not sure on the best format for the Symbol Information, I went with `const APPLE string` but perhaps there's a better formula to use here?

For global constants (though not in a namespace), I wasn't yet able to see where they were stored (see the failing test). It looks like `Codebase::getStubbedConstantType()` doesn't have it.

I'm also adding this to support both constant compleitions, but also it will mean effectively any word in typing will become a `*unknown_const` reference.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -18,6 +18,7 @@ use const PHP_MINOR_VERSION;
 use PhpParser;
 use function preg_match;
 use Psalm\Internal\Analyzer\ProjectAnalyzer;
+use Psalm\Internal\Analyzer\NamespaceAnalyzer;
 use Psalm\Internal\Analyzer\Statements\Block\ForeachAnalyzer;
 use Psalm\Internal\Type\Comparator\UnionTypeComparator;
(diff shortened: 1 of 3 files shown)


muglug:
Mind fixing the Psalm issue and test?

joehoyle:
@muglug sorry, was looking for some guidance on:

> For global constants (though not in a namespace), I wasn't yet able to see where they were stored (see the failing test). It looks like Codebase::getStubbedConstantType() doesn't have it.

Do you know what I'm doing wrong here? Where should I be looking for storage of constants in the root NS?

muglug:
`ConstantFetchAnalyzer::getGlobalConstType` seems to be the one

joehoyle:
Ok fixed up the tests, I was wrongly assuming that defining a constant in any file will cause it to be added to global constants, but that doesn't seem to be the case. I guess just because a constant is defined in _any_ file doesn't mean it is considered globally defined, as that file may not be loaded.

muglug:
Thanks!

PR #5117, opened 2021-01-28
docs: Update config of vim-lsp in language_server.md

yaegassy:
## Description

This is about the "vim-lsp" configuration.

It has now been changed from the `whitelist` to the `allowlist`.

The old settings may not be available in the future, so we have changed the documentation.

## REF

https://github.com/prabirshrestha/vim-lsp/commit/32fae1f0e9c0c2c21361d0bb94f119f9ae65a118


muglug:
Thanks!

PR #5122, opened 2021-01-28
Track references on global variables

joehoyle:
Add global type references to the type map, and fix up unused detection on global variables. To do this I had to also register the globals in the var maps when you do `global $x`. I think this makes sense -- in doing so I also had to add the Node references to satisfy the unused variable detection.


muglug:
Thanks, looks good!

PR #5128, opened 2021-01-29
Add completions for functions

joehoyle:
Provide autocompletions in the LSP for all global functions and functions from namespaces used in the current context.

I think I caught all the places functions live, but let me know if it looks like I missed any!

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1627,7 +1627,7 @@ class Codebase
         ksort($namespace_map);
         $namespace_map = array_reverse($namespace_map);
 
-        foreach ($functions as $function) {
+        foreach ($functions as $function_lowercase => $function) {
             // Transform FQFN relative to all uses namespaces
             $function_name = $function->cased_name;
             if (!$function_name) {
@@ -1635,7 +1635,7 @@ class Codebase
             }
             $in_namespace_map = false;
             foreach ($namespace_map as $namespace_name => $namespace_alias) {
-                if (strpos(strtolower($function_name), $namespace_name . '\\') === 0) {
+                if (strpos($function_lowercase, $namespace_name . '\\') === 0) {
                     $function_name = $namespace_alias . '\\' . substr($function_name, strlen($namespace_name) + 1);
                     $in_namespace_map = true;
                 }
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -44,6 +44,7 @@ use function substr;
 use function substr_count;
 use function array_pop;
 use function implode;
+use function array_reverse;
 
 class Codebase
 {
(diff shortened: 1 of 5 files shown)


src/Psalm/Codebase.php
```diff
@@ -1611,6 +1612,51 @@ public function getCompletionItemsForPartialSymbol(
...
+
+        foreach ($functions as $function) {
+            // Transform FQFN relative to all uses namespaces
+            $function_name = $function->cased_name;
+            if (!$function_name) {
+                continue;
+            }
+            $in_namespace_map = false;
+            foreach ($namespace_map as $namespace_name => $namespace_alias) {
+                if (strpos(strtolower($function_name), $namespace_name . '\\') === 0) {
```
staabm, line 1638: Could move strtolower(..) before the loop

muglug: You can go one step further: though `$functions` is a `list<string>`, it can instead be an `array<lowercase-string, string>` if `getMatchingFunctionNames` is updated.

src/Psalm/Internal/Codebase/Functions.php
```diff
@@ -252,6 +256,100 @@ public function getFullyQualifiedFunctionNameFromString(string $function_name, S
...
+                            continue;
+                        }
+                        $function = $this->reflection->getFunctionStorage($function_name);
+                    }
+                    $matching_functions[$function_name] = $function;
+                }
+            }
+        }
+
+        return array_values($matching_functions);
```
muglug, line 350: This can be simplified to

```suggestion
        $function_map = $file_storage->functions
            + $this->getAllStubbedFunctions()
            + $this->reflection->getFunctions()
            + $codebase->config->getPredefinedFunctions();

        foreach ($function_map as $function_name => $function) {
           foreach ($match_function_patterns as $pattern) {
                if (substr($pattern, -1, 1) === '*') {
                    if (strpos($function_name, rtrim($pattern, '*')) !== 0) {
                        continue;
                    }
                } elseif ($function_name !== $pattern) {
                    continue;
                }
                if (is_bool($function)) {
                    /** @var callable-string $function_name */
                    if ($this->reflection->registerFunction($function_name) === false) {
                        continue;
                    }
                    $function = $this->reflection->getFunctionStorage($function_name);
                }
                $matching_functions[$function_name] = $function;
            }
        }

        return $matching_functions;
```

src/Psalm/Codebase.php
joehoyle: I had kept it at `list<string>` to reflect the return typeof `ClassLIkes::getMatchingClassLikeNames`.

joehoyle: I'll update this though, ATM `getAllStubbedFunctions` `reflection::getFunctions()` and `config::getPredefinedFunctions()` all return `array<string, FunctionLike>` but I think in reality though should all be also `lowercase-string`?

muglug:
Thanks!

Issue #4184, opened 2020-09-14
Enforce property type invariance


bdsl:
Non-private properties in child classes need to be covariant for reading, like function returns and simultaneously contravariant for writing, like function params. That adds up to them needing to be invariant, or have exactly the same type as the property in the parent class.

I would therefore expect all four properties in the child class at https://psalm.dev/r/d0f7266a80 to be reported as errors for failing to match the types in the parent. Currently Psalm reports no errors.

I think this means that there is no good reason to override a property, except to set a value.

34vl [shows](https://3v4l.org/1PjtI) a fatal error on the first mismatched natevily typed property:

> Fatal error: Type of ChildClass::$mightExistNative must be ?string (as in class ParentClass) in /in/1PjtI on line 27

bdsl:
Matching issue reported at https://github.com/phpstan/phpstan/issues/3868

bdsl:
Fixed in release [4.5.0](https://github.com/vimeo/psalm/releases/tag/4.5.0)

PR #5131, opened 2021-01-30
Enforce property type invariance

bdsl:
This should be a fix for https://github.com/vimeo/psalm/issues/4184

Creates a new issue type **NonInvariantPropertyType** which is emitted when any property which exists in a parent class is redeclared in a child class with a different type.

### DIFF
--- a/docs/running_psalm/issues/NonInvariantPropertyType.md
+++ b/docs/running_psalm/issues/NonInvariantPropertyType.md
@@ -1,0 +1,26 @@
+# NonInvariantPropertyType
+
+Emitted when a public or protected class property has a different type to the matching property in the parent class.
+
+```php
+<?php
+
+class A {
+    /** @var string */
+    public $foo = 'hello';
+}
+
+class B extends A {
+    /** @var null|string */
+    public $foo;
+}
+
+```
+
+## Why this is bad
+
+For typed properties, this can cause a compile error. For non-typed
+properties, it can cause a type system violation when code written against the parent class reads or writes a value on an object of the child class.
+
+If the child class widens the type then reading the value may return unexpected value that client code cannot deal with. If the child class narrows the type then code setting the value may set
+it to an invalid value.
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -30,6 +30,7 @@ use Psalm\Issue\MissingImmutableAnnotation;
 use Psalm\Issue\MissingPropertyType;
 use Psalm\Issue\MissingTemplateParam;
 use Psalm\Issue\MutableDependency;
+use Psalm\Issue\NonInvariantPropertyType;
 use Psalm\Issue\OverriddenPropertyAccess;
 use Psalm\Issue\PropertyNotSetInConstructor;
 use Psalm\Issue\ReservedWord;
@@ -651,7 +652,7 @@ class ClassAnalyzer extends ClassLikeAnalyzer
                     ) {
                         if (IssueBuffer::accepts(
                             new OverriddenPropertyAccess(
-                                'Property ' . $guide_class_storage->name . '::$' . $property_name
+                                'Property ' . $fq_class_name . '::$' . $property_name
                                     . ' has different access level than '
                                     . $storage->name . '::$' . $property_name,
                                 $property_storage->location
@@ -659,8 +660,22 @@ class ClassAnalyzer extends ClassLikeAnalyzer
                         )) {
                             // fall through
                         }
+                    }
 
-                        continue;
+                    if ($property_storage->type && $guide_property_storage->type && $property_storage->location &&
+                        !$property_storage->type->equals($guide_property_storage->type)) {
+                        if (IssueBuffer::accepts(
+                            new NonInvariantPropertyType(
+                                'Property ' . $fq_class_name . '::$' . $property_name
+                                . ' has type ' . $property_storage->type->getId()
+                                . ", not invariant with " . $guide_class_name . '::$' . $property_name . ' of type ' .
+                                $guide_property_storage->type->getId(),
+                                $property_storage->location
+                            ),
+                            $property_storage->suppressed_issues
+                        )) {
+                            // fall through
+                        }
                     }
                 }
             }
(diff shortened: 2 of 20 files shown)


docs/running_psalm/issues/NonInvariantChildProperty.md
```diff
@@ -0,0 +1,26 @@
+# NonInvariantChildProperty
```
muglug, line 1: I'd rename this `NonInvariantPropertyType` as the explanation makes clear it applies to parent-child relationships

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
```diff
@@ -651,16 +652,30 @@ public static function addContextProperties(
...
+                        !$property_storage->type->equals($guide_property_storage->type)) {
+                        if (IssueBuffer::accepts(
+                            new NonInvariantChildProperty(
+                                'Property ' . $fq_class_name . '::$' . $property_name
+                                . ' has type ' . $property_storage->type->getId()
+                                . ", not invariant with " . $guide_class_name . '::$' . $property_name . ' of type ' .
+                                $guide_property_storage->type->getId(),
+                                $property_storage->location
+                            ),
+                            $statements_source->getSuppressedIssues()
```
muglug, line 675: You need to add $property_storage->suppressed_issues to allow it to be suppressed

docs/running_psalm/issues/NonInvariantChildProperty.md
bdsl: Thanks. I'll read through your comments now but I need to sleep, so I'll make changes tomorrow.

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
bdsl: Thanks. `\Psalm\Storage\PropertyStorage` didn't have a `suppressed_issues` property so I've added it, not sure if I went about populating it in the right way.

docs/running_psalm/issues/NonInvariantChildProperty.md
bdsl: I've renamed it.

bdsl:
This is f[ailing in the test-with-real-projects step](https://app.circleci.com/pipelines/github/bdsl/psalm/286/workflows/16c1f9ac-df69-4327-84dd-d240921c9f0e/jobs/1173) at CircleCI because it detects a few instances of this issue in https://github.com/muglug/phpunit . @muglug Let me know if you'd like me to make a separate PR to that repo to update the baseline.

bdsl:
Since with this change we're not allowed to change the type of a redeclared property, which of course I think we shouldn't be, I'm not really sure what use case there is for redeclaring or why PHP allows it at all. [Visibility](https://www.php.net/manual/en/language.oop5.visibility.php) in the PHP manual shows an example of redeclaration to change the default value of a property, but you could generally achieve the same thing through setting the value in a constructor.

src/Psalm/Issue/NonInvariantPropertyType.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Issue;
+
+final class NonInvariantPropertyType extends CodeIssue
+{
+    public const ERROR_LEVEL = -1;
+    public const SHORTCODE = 235159;
```
bdsl, line 10: This is not very short. I'm not sure sure if there's a system for assigning shortcodes to issues.

src/Psalm/Type/Atomic/TArray.php
```diff
@@ -14,6 +14,7 @@ class TArray extends \Psalm\Type\Atomic
 
     /**
      * @var array{\Psalm\Type\Union, \Psalm\Type\Union}
+     * @psalm-suppress NonInvariantPropertyType
```
bdsl, line 17: Maybe this and similar properties can be deleted from the code and we can just use the properties declared in the parent classes?

tests/ClassTest.php
```diff
@@ -570,7 +570,6 @@ class ReflectionGenerator {
             'extendException' => [
                 '<?php
                     class ME extends Exception {
-                        /** @var string */
```
bdsl, line 573: Deleted this because it registers as unequal since the parent property is not `fromDocBlock`.

src/Psalm/Issue/NonInvariantPropertyType.php
weirdan: There's a script that outputs max used shortcode, so you could use that (on master) and `+1` the value it returns.

bdsl, line 10: Thanks Bruce. Taking your suggestion:

```suggestion
    public const SHORTCODE = 265;
```

bdsl: I made https://github.com/vimeo/psalm/pull/5133 while I was there.

muglug:
Thanks! I'll fix the baseline in PHPUnit separately

docs/running_psalm/issues/NonInvariantPropertyType.md
```diff
@@ -0,0 +1,26 @@
...
+
+```
+
+## Why this is bad
+
+For typed properties, this can cause a compile error. For non-typed
+properties, it can cause a type system violation when code written against the parent class reads or writes a value on an object of the child class.
+
+If the child class widens the type then reading the value may return unexpected value that client code cannot deal with. If the child class narrows the type then code setting the value may set
+it to an invalid value.
```
iluuu1994, line 26: I think narrowing the type of a `readonly` property should semantically be ok though.

https://psalm.dev/r/7af10fdcb3

```php
class A {
    /**
     * @psalm-readonly
     * @var string|null
     */
    public $foo = 'hello';
}

class B extends A {
    /**
     * @psalm-readonly
     * @var string
     */
    public $foo = 'hello 2';
}
```

This is essentially the same as covariance return types.

iluuu1994: Actually, maybe not if `__construct` is not overridden. Probably not worth pursuing.

bdsl: Right. It might be OK with a private constructor.

PR #5146, opened 2021-02-02
improve psalter capacities for anonymous class extending real classes

### LINKED ISSUE #5145: Psalter changing return type of function returning anonymous class
**Before Psalter**
```
<?php

class A {}

function f(): A
{
    return new class extends A {};
}
```
**After Psalter**
```
<?php

class A {}

function f(): object
{
    return new class extends A {};
}
```
The return type of the function `f` was changed to `object`. Is this intended?

Would it be doable to use the class the anonymous class extends from if it is the case?

orklah:
This should fix #5145


muglug:
Thanks!

PR #5152, opened 2021-02-04
[WIP][FEATURE] Introduce conditional inversion

ohader:
→ **Please ignore this PR for the time being, wanted to verify PhpUnit issues I'm experiencing locally**

```
@psalm-return (T1 is string ? string : int)
@psalm-return (T1 not int ? string : int)
```

### DIFF
--- a/src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php
@@ -352,11 +352,14 @@ class AtomicTypeComparator
                     $allow_float_int_equality,
                     $atomic_comparison_result
                 )) {
-                    return true;
+                    // with inverse=false -> true
+                    // with inverse=true -> false
+                    return true xor $container_type_part->inverse;
                 }
             }
-
-            return false;
+            // with inverse=false -> false
+            // with inverse=true -> true
+            return false xor $container_type_part->inverse;
         }
 
         if ($input_type_part instanceof TTemplateParam) {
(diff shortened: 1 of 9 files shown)


orklah:
I'm not sure if you know this already but there is actually 3 outputs for a conditional in psalm:
https://psalm.dev/r/1de73fc314
1) One when the type is known and match the condition
2) One when the type is known and doesn't match the condition
3) One when the type is unknown (in this case it returns mixed)

Which means inversing the condition will not be an exact inversion of the outputs

ohader:
> I'm not sure if you know this already but there is actually 3 outputs for a conditional in psalm:
> https://psalm.dev/r/1de73fc314
> 
> 1. One when the type is known and match the condition
> 2. One when the type is known and doesn't match the condition
> 3. One when the type is unknown (in this case it returns mixed)
> 
> Which means inversing the condition will not be an exact inversion of the outputs

Thanks for the pointer - that's actually what I'd like to fine out - my origin is  `@psalm-taint-*`...
Inversions get tricky when boolean algebra is involved. However, your example don't seem to be related to that, are they?

muglug:
> inversions get tricky when boolean algebra is involved

how so?

I'm skeptical that this is necessary – I've tried to keep the syntax [comparable to TypeScript's](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html)

ohader:
I did not manage to implement it in a generic way - the use-case I started with was `@psalm-taint-*` declarations. Currently only `@psalm-taint-escape` seems to allow conditions - I probably would be better to extend conditional support to more annotations, like e.g.

```
@psalm-taint-source <taint-type #conditional> 🛠️ 
@psalm-taint-source ($fromRequest is true ? 'input' : null)

@psalm-taint-sink <taint-type> <param-name> 🛠️
@psalm-taint-sink ($dryRun is false ? 'sql' : null) $query

@psalm-taint-escape <taint-type #conditional> ✅ 
@psalm-taint-escape ($encode is true ? 'html' : null)

@psalm-taint-unescape <taint-type> 🛠️
@psalm-taint-unescape ($decode is true ? 'html' : null)
```

PR #5165, opened 2021-02-06
allow libraries to provide psalm plugins

azjezz:
In `azjezz/psl`, i want to provide a plugin with the library, however, i don't want to change the package type, as it is still a library, even tho it defines an optional psalm-plugin, this restriction ( of requiring the package to be of `psalm-plugin` type ), is currently stopping `vendor/bin/psalm-plugin` from discovering the plugin within PSL, even tho the plugin class is declared.

see https://github.com/azjezz/psl/pull/119/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R65

### DIFF
--- a/src/Psalm/Internal/PluginManager/ComposerLock.php
+++ b/src/Psalm/Internal/PluginManager/ComposerLock.php
@@ -31,11 +31,8 @@ class ComposerLock
     public function isPlugin($package): bool
     {
         return is_array($package)
-            && isset($package['name'])
+            && isset($package['name'], $package['extra']['psalm']['pluginClass'])
             && is_string($package['name'])
-            && isset($package['type'])
-            && $package['type'] === 'psalm-plugin'
-            && isset($package['extra']['psalm']['pluginClass'])
             && is_array($package['extra'])
             && is_array($package['extra']['psalm'])
             && is_string($package['extra']['psalm']['pluginClass']);
@@ -71,7 +68,7 @@ class ComposerLock
     }
 
     /**
-     * @return list<array{type:string,name:string,extra:array{psalm:array{pluginClass:string}}}>
+     * @return list<array{name:string,extra:array{psalm:array{pluginClass:string}}}>
      */
     private function getAllPluginPackages(): array
     {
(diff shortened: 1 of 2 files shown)


tests/ComposerLockTest.php
```diff
@@ -51,10 +52,10 @@ public function seesNonDevPlugins(): void
      */
     public function seesDevPlugins(): void
     {
-        $lock = new ComposerLock([$this->jsonFile((object)[
+        $lock = new ComposerLock([$this->jsonFile([
             'packages' => [],
             'packages-dev' => [
-                (object) $this->pluginEntry('vendor/package', 'Vendor\Package\PluginClass'),
+                $this->pluginEntry('vendor/package', 'Vendor\Package\PluginClass'),
```
azjezz, line 58: I'm not sure why `object` was used here, but both objects and keyed-arrays are encoded as objects in JSON, so i assume it's redundant.

tests/ComposerLockTest.php
```diff
@@ -12,7 +12,7 @@ class ComposerLockTest extends TestCase
      */
     public function pluginIsPackageOfTypePsalmPlugin(): void
```
weirdan, line 13: The test case name becomes misleading with these changes.

tests/ComposerLockTest.php
weirdan: I vaguely recall something breaking (perhaps it was empty object vs empty array) so I just used `(object)[...]` as a cheap replacement for object literal PHP lacks.

weirdan (changes requested):
Note that plugins without `type=psalm-plugin` won't show up on https://psalm.dev/plugin

tests/ComposerLockTest.php
azjezz: renamed 👍

tests/ComposerLockTest.php
azjezz: Okay, i reverted this changes just in case

azjezz:
> Note that plugins without type=psalm-plugin won't show up on https://psalm.dev/plugin

this is not an issue, i would like to think if someone is installing `azjezz/psl`, they have read the README, which mentions the plugin right after the installation method 😛

src/Psalm/Internal/PluginManager/ComposerLock.php
```diff
@@ -30,12 +30,9 @@ public function __construct(array $file_names)
      */
     public function isPlugin($package): bool
     {
-        return is_array($package)
-            && isset($package['name'])
+        return isset($package['name'], $package['extra']['psalm']['pluginClass'])
```
weirdan, line 33: I would probably put that after `is_array($package)` check.

azjezz: 👍

weirdan (approved):

muglug:
Thanks! Could you also add documentation?

PR #5169, opened 2021-02-06
Improved PDO::pgsqlGetNotify signature

vudaltsov:
See https://www.php.net/manual/ru/pdo.pgsqlgetnotify.php

`payload` offset is present in the result only when notification has a payload.


muglug:
Thanks!

Issue #5177, opened 2021-02-07
Ignoring ContainerExceptionInterface and descendants does not work for NotFoundExceptionInterface


vudaltsov:
Given config

```xml
    <ignoreExceptions>
        <classAndDescendants name="Psr\Container\ContainerExceptionInterface"/>
    </ignoreExceptions>
```

This code

```php
function test(Psr\Container\ContainerInterface $container): void
{
    $container->get('a');
}
```

still produces

```
ERROR: MissingThrowsDocblock - .stuff/test.php:9:17 - Psr\Container\NotFoundExceptionInterface is thrown but not caught - please either catch or add a @throws annotation (see https://psalm.dev/169)
    $container->get('a');
```

although `NotFoundExceptionInterface extends ContainerExceptionInterface`, see https://github.com/php-fig/container/blob/master/src/NotFoundExceptionInterface.php

weirdan:
This is possibly caused by `ContainerException` not extending `Throwable`: https://psalm.dev/r/5e71411915. @vudaltsov can you check if that's true by modifying `psr/container` source locally?

vudaltsov:
@weirdan , I have a stub to fix that. But I checked with modified source — same result, the issue still occurs.

PR #5178, opened 2021-02-07
Allow suppressing thrown interfaces and their descendants

weirdan:
Fixes vimeo/psalm#5177

Test was redone with docblock throws instead of actual `throw` statements because Psalm was ignoring everything after the first throw.


vudaltsov (approved):

muglug:
Thanks!

PR #5197, opened 2021-02-11
Fix array_key_exists() with all int literal keys

brainlock:
(https://psalm.dev/r/a0e5cf2ff8)

When checking code like the following:

```php
<?php

function checkNegated(string $key): void {
    $arr = [
        0 => "foo",
        1 => "bar",
    ];

    if (!array_key_exists($key, $arr)) {
        printf("not found\n");
    }
}

function check(string $key): void {
    $arr = [
        0 => "foo",
        1 => "bar",
    ];

    if (array_key_exists($key, $arr)) {
        printf("found\n");
    }
}
```

the `if` in `checkNegated` would cause:

```
ERROR: RedundantCondition - 9:10 - Type string for $key is never =int(0)
```

This happens when the array keys are all int literals, but the "needle"
is a string.

`array_key_exists()` uses a loose equality comparison, but the generated
assertions for this specific case
(`AssertionFinder::getArrayKeyExistsAssertions`) was generating strict
equality clauses. This commit fixes it by changing the generated clause
from `=` to `~`.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -3770,7 +3770,7 @@ class AssertionFinder
                         }
                     } elseif ($key_type->allIntLiterals() && !$key_type->possibly_undefined) {
                         foreach ($key_type->getLiteralInts() as $array_literal_type) {
-                            $literal_assertions[] = '=' . $array_literal_type->getAssertionString();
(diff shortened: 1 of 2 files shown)


brainlock:
The psalm.dev snippet I first link could be confusing, so here's a new one to highlight the actual bug (it cannot know that the key is never in the array):

https://psalm.dev/r/d770e4c1c5

(I'll update the commit message as well)

muglug:
Thanks!

PR #5201, opened 2021-02-12
Remove the trailing comma for function call as minimum support version is 7.1

drealecs:
The new version, 4.5.*, fails on PHP 7.1 and 7.2 with:
```
Uncaught ParseError: syntax error, unexpected ')' in ...../vendor/vimeo/psalm/src/Psalm/Codebase.php:1635
```

I'm guessing this will require a new bug patch release after merge.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -1631,7 +1631,7 @@ class Codebase
                         null,
                         null,
                         null,
-                        $property_name,
+                        $property_name
                     );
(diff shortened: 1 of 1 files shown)


drealecs:
Also, dropping support for PHP 7.1 nd 7.2 might be a solution :)

muglug:
Thanks!

joehoyle:
Oops, think that was my bad!

PR #5204, opened 2021-02-12
Use ReflectionNamedType instead of ReflectionType (PHP 7.1+)

lptn:
`ReflectionNamedType` has been [introduced in PHP7.1](https://www.php.net/manual/en/class.reflectionnamedtype.php): `class ReflectionNamedType extends ReflectionType`

Technically speaking, `ReflectionFunction::getReturnType(): ReflectionType` is a valid construction, but it returns `ReflectionNamedFunction` in version PHP 7.1+. The same for other reflection API classes that have getType() or getReturnType() methods.

The problem raises when we use a chain `ReflectionFunction->getReturnType()->getName()`, it's not correct anymore: `getName()` available in `ReflectionNamedType` but doesn't in `ReflectionType`.

see https://3v4l.org/0CMfa (tests ReflectionFunction and child)
see https://3v4l.org/Y2Jqo (tests ReflectionProperty::getType())

___________
Related PR on phpstorm-stubs: https://github.com/JetBrains/phpstorm-stubs/pull/1039


caugner:
@lptn Thank you for your PR. Would you mind rebasing your branch onto the current master?

I noticed one important change in your PR, which is to move `PropertyType::getType` to `dictionaries/CallMap_74_delta.php`. 

Meanwhile, #5584 added `PropertyType::hasType` directly to `dictionaries/CallMap.php`, but your PR makes me think that both `getType` and `hasType` belong in `CallMap_74_delta.php`, because they aren't available in earlier versions.

lptn:
hey @caugner 

thanks for the feedback! Yes, your concern is valid, `PropertyType::hasType` is available from PHP7.4 only: https://3v4l.org/L6lPY . Unfortunately, I've removed my fork and local branch (it was created on 12 Feb), but  can create a new PR with the same changes + move `PropertyType::hasType` to callmap 7.4

caugner:
@lptn Doesn't GitHub offer you to restore the branch here? I wonder if you also don't have commit 1aff7e25b6d2162f1da090c3612aa9336b4e8df1 in your local or remote repository anymore?

lptn:
@caugner 
I didn't have both local and remote forks anymore (ironically I've removed them few days ago). I've created a new fork and used the same branch name - no success. So, I've created a new PR: https://github.com/vimeo/psalm/pull/6077

caugner:
@lptn Thanks! Let's close this one in favor of that one then. (I can't, but you should be able to.)

PR #5205, opened 2021-02-12
Allow @psalm-type and @psalm-import-type to be used in extends/implements

### LINKED ISSUE #4240: Type declared with @psalm-type cannot be used in @template-extends annotation
Hello,
Currently there is a bug with the using of `@psalm-type` in `@template-extends` annotation.

The following snippet (https://psalm.dev/r/f1fffe9d00), has some errors:
`ERROR: UndefinedDocblockClass - 20:13 - Docblock-defined class or interface IArgs does not exist`

```php
<?php

/**
 * @template TArgs
 * @template TReturn
 */
abstract class AbstractField
{
    /** 
     * @psalm-param TArgs $args
     * @psalm-return TReturn
     */
    abstract public function resolve($args);
}

/**
 * @psalm-type IArgs = array{id: string}
 * @template-extends AbstractField<IArgs, string>
 */
final class ClientField extends AbstractField
{
    public function resolve($args)
    {
        return $args['id'];
    }
}
```

Whereas the if I use `array{id: string}` instead of `IArgs`, that works (snippet on: https://psalm.dev/r/5173751356).

In the previous versions it worked (at least at version 3.12.1)

leightonthomas:
Fix #4240

This fixes issues with not being able to use `@psalm-type`s in the same doc-block they're defined in, and allows them to be used for `@implements`/`@extends` type parameters as expected

It's mostly just moving some things around so that it gets the types from the node's comments before carrying on with the rest of the doc-block stuff, and pulls the resolving of imported types into a separate method

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
@@ -104,15 +104,11 @@ class ClassLikeNodeScanner
      */
     public $type_aliases = [];
 
-    /**
-     * @param array<string, TypeAlias> $type_aliases
-     */
     public function __construct(
         Codebase $codebase,
         FileStorage $file_storage,
         FileScanner $file_scanner,
         Aliases $aliases,
-        array $type_aliases,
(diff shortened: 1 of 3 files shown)


muglug:
Thanks!

PR #5213, opened 2021-02-13
Rename stub to avoid phpstorm conflict

VincentLanglet:
Hi @muglug

Phpstorm is higlighting my `Exception` because I'm providing a string to the constructor.
I discovered it's because there is an Exception class in the `MissingMethod` stub.

To avoid such issue, all the stubs should used the `.phpstub` extension IMHO.


muglug:
Thanks – please update the references in PHP too!

VincentLanglet:
Fixed @muglug :)

muglug:
Thanks!

PR #5218, opened 2021-02-14
Update CoreImmutableClasses.phpstub

tarlepp:
Fixes wrong parameter names - https://psalm.dev/r/8cfdb309d9


tarlepp:
Also I noticed that `DateTime` class is behaving the same way - https://psalm.dev/r/129b9fce2b - but that class doesn't have stub like that `DateTimeImmutable` has.

muglug:
Thanks!

Issue #5215, opened 2021-02-13
Virtual nodes created by Psalm sent to plugins


orklah:
I'm currently implementing AfterExpressionAnalysisInterface in order to detect and change == into === when safe.

It seemed pretty easy but I'm stuck with an issue I don't know how to solve.

In SwitchCaseAnalyzer (and probably elsewhere), Psalm is creating a new Expr\BinaryOp\Equal in order to analyse operations. This Equal Expression is then sent to the plugin as if it was in the original code as a real "==" node.

Unfortunately, When I try to replace with a === it just breaks the code. I couldn't find a way to differentiate between those virtual nodes and real one.

Is there a way to do that I'm missing or could we maybe flag those node somehow to exclude them in plugin analysis?

orklah:
Ok, I gave it some thought, the solutions I came up were:
- Add an attribute `virtual => true` to every node created in Psalm. (probably a few hundreds)
- Create classes extending PHPParser's Nodes (for example VirtualString_) and use those in Psalm instead of native ones
- Register every virtual node created in some sort of storage (kinda like nodeTypeProvider) that the plugin could call to know if the node is virtual or not

I'm out of ideas :)

Creating classes seems the most clean to me, @muglug can you tell me if it's worth it before I start a PR?

Thanks

muglug:
> Creating classes seems the most clean to me, @muglug can you tell me if it's worth it before I start a PR?

It's definitely the most clean, and I also don't see any drawbacks to that approach. I guess each would implement a `VirtualNode` interface or somesuch to differentiate them

orklah:
Fixed by linked PR

PR #5222, opened 2021-02-14
Introduce Virtual Nodes in order to differentiate real nodes in plugins

orklah:
This should solve https://github.com/vimeo/psalm/issues/5215

A few notes:
1) I had errors in PhpParser pop up now that Psalm's code inherits PhpParser code and Psalm's PropertyMap being more precise than PhpParser's code
2) I had errors in Psalm for Unused classes because I migrated every Node from PhpParser, not only used Nodes. Not sure if I should keep it or delete those.

### DIFF
--- a/examples/TemplateChecker.php
+++ b/examples/TemplateChecker.php
@@ -11,6 +11,8 @@ use Psalm\Internal\Analyzer\StatementsAnalyzer;
 use Psalm\CodeLocation;
 use Psalm\Context;
 use Psalm\DocComment;
+use Psalm\Node\Stmt\VirtualClass;
+use Psalm\Node\Stmt\VirtualClassMethod;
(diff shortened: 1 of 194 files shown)


muglug:
Sorry, mind merging again? I did a dumb thing

orklah:
Done.

I notice there are errors in Code Style because it doesn't like the `_` at the end for classes that are also keywords. Not sure what should be done about that

muglug:
> Not sure what should be done about that

Just remove the underscore. The _only_ reason the underscore exists in PhpParser is because `Class` is a reserved word. When extending and renaming there's no need to keep that convention.

orklah:
Done :)

muglug:
Thanks!

Issue #5227, opened 2021-02-15
"RedundantCondition - positive-int can never contain 0" on a result of a bit & operation that can be 0


sad-spirit:
I suspect psalm incorrectly assumes `positive-int` type for the result of a bit operation with two `positive-int` arguments. But while `1` and `2` are positive, `1 & 2` is not.

https://psalm.dev/r/d174e99595

weirdan:
Simplified: https://psalm.dev/r/931ebf1a0c

orklah:
The bug must be around here:
https://github.com/vimeo/psalm/blob/88fe805c2741ab10ecefb289a42f6076ed5e57ee/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php#L695

Is there other binary operator when two positive int can result in a non positive int?

sad-spirit:
Bitwise xor, left and right bitwise shifts (at least)
https://psalm.dev/r/2ba369ca91

weirdan:
Substraction can also result in zero or negative result with both positive operands.

orklah:
> Substraction can also result in zero or negative result with both positive operands.

Yeah, this is handled here: 
https://github.com/vimeo/psalm/blob/2f2ee0008c7430263ea3e211919267f886480859/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php#L664

Standard operators are pretty well debugged by now I think. Edge cases are more in bitwise operators where (at least for me) the result is not immediately obvious

PR #5233, opened 2021-02-15
Fix not always positive bit operations

orklah:
This should fix https://github.com/vimeo/psalm/issues/5227

I'll wait to see failing tests to see if I just fix them or if I add this example: https://psalm.dev/r/2ba369ca91


sad-spirit:
I think you missed `\PhpParser\Node\Expr\BinaryOp\BitwiseAnd` from the original report.

orklah:
Thanks! Added!

muglug:
Thanks!

PR #5240, opened 2021-02-17
Allow installation of netresearch/jsonmapper ^4.0

jeromegamez:
Greetings 👋🏼!

Similar to https://github.com/vimeo/psalm/pull/3830, this has no immediate effect until https://github.com/felixfbecker/php-advanced-json-rpc/pull/53 is merged and released, but when it is, it promises support for PHP 8 and PHPUnit 9, while keeping support for PHP ^7.1 (https://github.com/cweiske/jsonmapper/blob/v4.0.0/ChangeLog)

:octocat:


muglug:
Thanks!

Issue #5259, opened 2021-02-21
Add error handler to convert PHP errors / warnings / notices into exceptions


weirdan:
> how do you feel about introducing an error handler to convert all notices/warnings/etc into exceptions?

I think that's a great idea! Normally a notice pre-empts a fatal error anyway.

_Originally posted by @muglug in https://github.com/vimeo/psalm/issues/5255#issuecomment-782776250_

PR #5260, opened 2021-02-21
Convert PHP errors / warnings / notices to exceptions

weirdan:
Fixed vimeo/psalm#5259


orklah:
Note: this may have an impact for plugins as well and *could* be considered a BC break.

muglug:
Thanks @weirdan – a way around the BC issue would be to make its use configurable, but I think it's fine for now

PR #5262, opened 2021-02-21
Change code-climate line reference

spyric:
It seems that CodeClimateReport generator take wrong data for lines description


muglug:
Thanks!

PR #5267, opened 2021-02-22
Add documentation to LSP

joehoyle:
This adds documentation to several features of the LSP. Specifically:

Functions, methods, classes, class constants and class properties will now display their documentation/description in the IDE's Hover information, Autocomplete lists and Signature Information.

Here's some examples:

Autocomplete documentation:

Hover documentation:

Documentation in function param signature 

I mostly modeled these off the default TypeScript documentation in VS Code.

--

This is blocked on https://github.com/felixfbecker/php-language-server-protocol/pull/17

Personally I am bullish on including documentation on all the PHP inbuilt functions, this reflects what the Typescript / JS LSP by default also does.

I didn't yet add guards around all documentation collection for the LSP only -- I wasn't quite sure how to do that (I imagine that is stored in the `config` object?), though I wasn't sure how difficult it would be to get the scope/context of that value at each point.

As a comparison, I ran `./vendor/bin/psalm --no-cache --no-reflection-cache --no-file-cache .` on a pretty large WordPress project (which makes a lot of use of phpdocumentation)

`master`
Checks took 5.19 seconds and used 552.723MB of memory

`lsp-documentation`
Checks took 5.19 seconds and used 554.804MB of memory

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "composer/xdebug-handler": "^1.1",
         "dnoegel/php-xdg-base-dir": "^0.1.1",
         "felixfbecker/advanced-json-rpc": "^3.0.3",
-        "felixfbecker/language-server-protocol": "^1.4",
+        "felixfbecker/language-server-protocol": "^1.5",
         "netresearch/jsonmapper": "^1.0 || ^2.0 || ^3.0 || ^4.0",
         "nikic/php-parser": "^4.10.1",
         "openlss/lib-array2xml": "^1.0",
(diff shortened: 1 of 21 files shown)


muglug:
> This is blocked on felixfbecker/php-language-server-protocol#17

I've just bumped that package version, so you can bump that here and all should be good

joehoyle:
@muglug I've updated that dev, thanks! Currently the listing is failing on a long line with a `@param array{...` doc block. I don't _think_ Psalm supports breaking this on to multiple lines? Not sure how to get that to pass.

muglug:
Thanks! Psalm does indeed support docblocks on multiple lines, thankfully.

Issue #5265, opened 2021-02-22
Using an unsupported PHP feature causes an unclear error message


ciaranmcnulty:
I used a nullsafe method call:

```
return $order->getId()?->asString();
```

and got the error:
```
UnrecognizedExpression - src/Application/OrdersService.php:19:16 - Psalm does not understand PhpParser\Node\Expr\NullsafeMethodCalll (see https://psalm.dev/048)
```

This was because although I was using PHP 8 my project's `composer.json` defined a lower version. The root cause isn't really clear from the error, and the linked page says "This should never happen." 

Possibly these cases could be handled with a more clear error message + linked explanation

ciaranmcnulty:
(was not reproducible on psalm.dev, which added to my befuddlement)

orklah:
I'm a little confused as to how the php version in the config (or composer) interact with the php version from the actual executable. Also the error message when running on 7.4 on psalm.dev is not even the same: https://psalm.dev/r/10b6fa221d?php=7.4

But I guess displaying the php version here:
https://github.com/vimeo/psalm/blob/f9fccb2b2dafacc83bf6c43c24d09aa8a589446c/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php#L400
should make the error clearer

muglug:
@orklah's fix should help here – closing

PR #5268, opened 2021-02-22
display PHP version when not understanding an expression

orklah:
This PR will clarify the error message from https://github.com/vimeo/psalm/issues/5265 by displaying the PHP version when an element has not be recognized. This will probably only be useful when launching an analysis in a different php version than composer.


muglug:
Thanks!

PR #5270, opened 2021-02-23
Fixed #3416 - templates with class phpdocs

### LINKED ISSUE #3416: Broken parsing of '@template of I' followed by a simple comment
When `@template of` is followed by a simple comment on a next line, psalm fails to stop parsing the template tag, and continues to the next line, making the comment part of `of I`

https://psalm.dev/r/f5aa676997

Without 'of I' everything is parsed properly

https://psalm.dev/r/74b00ac705

Same applies for at least one more tag -  `@extends`

elnoro:
This is a partial fix for https://github.com/vimeo/psalm/issues/3416

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
@@ -53,7 +53,8 @@ class ClassLikeDocblockParser
         $info = new ClassLikeDocblockComment();
 
         if (isset($parsed_docblock->combined_tags['template'])) {
-            foreach ($parsed_docblock->combined_tags['template'] as $offset => $template_line) {
+            foreach ($parsed_docblock->combined_tags['template'] as $offset => $template_tag) {
+                $template_line = preg_replace('/\n.*/m', '', $template_tag);
                 $template_type = preg_split('/[\s]+/', preg_replace('@^[ \t]*\*@m', '', $template_line));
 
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
```diff
@@ -53,7 +53,8 @@ public static function parse(
         $info = new ClassLikeDocblockComment();
 
         if (isset($parsed_docblock->combined_tags['template'])) {
-            foreach ($parsed_docblock->combined_tags['template'] as $offset => $template_line) {
+            foreach ($parsed_docblock->combined_tags['template'] as $offset => $template_tag) {
+                $template_line = preg_replace('/\n.*/m', '', $template_tag);
```
weirdan, line 57: Wouldn't it break more complex structures like:
```php
/**
 * @template T of array{
 *    a: string,
 *    b: int
 * }
 */
```

elnoro:
You are right, it will probably break it. I’ll see if I can find a better way to do this

elnoro:
Closed this since I am not sure what is the best way to resolve this without breaking BC. 

@weirdan do you think it's worth to add a test with a complex structure like the one in your example to ClassTemplateTest.php? This should save you some time next time someone attempts a similar solution.

weirdan:
> do you think it's worth it to add a test with a complex structure 

Feel free to PR that.

PR #5272, opened 2021-02-23
session_id can return false

orklah:
according to doc: https://www.php.net/manual/en/function.session-id.php#refsect1-function.session-id-returnvalues


muglug:
Thanks!

PR #5280, opened 2021-02-25
bugfix: do not throw `ConfigException` when missing files are allowed

boesing:
Hey there,

I've came along an issue with my configuration and searched for similar issues/fixes here.
Ive stumbled upon #4259 and #4238 where #4259 fixed some exception already in another case.

Since there are two more cases where an exception is thrown due to missing files, I'd like to contribute.

Thanks.

### DIFF
--- a/src/Psalm/Config/FileFilter.php
+++ b/src/Psalm/Config/FileFilter.php
@@ -239,6 +239,10 @@ class FileFilter
                     );
 
                     if (empty($globs)) {
+                        if ($allow_missing_files) {
+                            continue;
+                        }
+
                         throw new ConfigException(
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

XedinUnknown:
Still [presents](https://github.com/Dhii/collections-interface/actions/runs/10975227896/job/30474637044?pr=31#step:8:9) in [4.30.0](https://github.com/Dhii/collections-interface/actions/runs/10975227896/job/30474637044?pr=31#step:6:181).

donaldinou:
@XedinUnknown 
You need to add `allowMissingFiles="true"` attribute to `ignoreFiles` tag.

https://psalm.dev/docs/running_psalm/configuration/#projectfiles

PR #5281, opened 2021-02-25
fixed typo in contributing/index.md

elnoro:
Hello 

There is no `phpcs` composer script and this part of the doc seems to be a typo (at least it fails on my machine)

There is a `cs` command however that runs `phpcs -p` internally.


muglug:
Thanks!

PR #5300, opened 2021-03-01
Fix JSON reports with long UTF8 strings

danog:
When printing keyed arrays or literal strings containing more than 80 characters, the old logic may cut off a UTF8 codepoint in half, causing an error when encoding JSON reports.
Also added a call to json_last_error_msg to obtain a detailed error description.

### DIFF
--- a/tests/TypeParseTest.php
+++ b/tests/TypeParseTest.php
@@ -860,6 +860,15 @@ class TypeParseTest extends TestCase
         $this->assertSame($resolved_type->getId(), $docblock_type->getId());
     }
 
+    public function testLongUtf8LiteralString(): void
+    {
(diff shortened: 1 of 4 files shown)


tests/TypeParseTest.php
```diff
@@ -860,6 +861,15 @@ public function testEnumWithoutSpaces(): void
         $this->assertSame($resolved_type->getId(), $docblock_type->getId());
     }
 
+    public function testLongUtf8LiteralString(): void
+    {
+        $string = "АаБбВвГгДдЕеЁёЖжЗзИиЙйКкЛлМмНнОоПпРрСсТтУуФфХхЦцЧчШшЩщЪъЫыЬьЭэЮюЯя";
```
weirdan, line 866: I would suggest to double the string, so that it's longer than 80 (multibyte) characters. It'll make sure the string is actually truncated.

danog: Well that's exactly what I do below)

weirdan: Oops, missed that.

muglug:
Thanks!

PR #5306, opened 2021-03-02
Fix mysqli_init() return type

### LINKED ISSUE #5305: Incorrect return type for the mysqli_init() stub
https://psalm.dev/r/a23ca5ef27

`mysqli_init()`can return false according to the [PHP source code](https://github.com/php/php-src/blob/da011a312a6c6cd7ff12fe1aa0de1e33fba2f167/ext/mysqli/mysqli_api.c#L1455) and the [MySQL API documentation](https://dev.mysql.com/doc/c-api/8.0/en/mysql-init.html) (the same should be true for `mysqlnd`).

morozov:
Closes #5305.


muglug:
Thanks!

PR #5323, opened 2021-03-04
Update mysql_connect_error() return type definition

ErikBooijCB:
Tiny update to the call map.

While refactoring a very old code base, Psalm is proving to be an extremely helpful tool once again. I did however run into a little issue where the return type of `mysqli_connect_error()` is incorrectly defined. Reading the documentation on PHP.net, that's no wonder, because it *does* claim the return type is string.

However, in the `Return values` section it explains `null` is returned when no error occurred:

https://www.php.net/manual/en/mysqli.connect-error.php#refsect1-mysqli.connect-error-returnvalues


muglug:
Thanks!

PR #5329, opened 2021-03-05
Always return type inference summary, even if no files were analyzed due to caching

### LINKED ISSUE #5319: "No files analyzed" instead of "Psalm was able to infer types for X% of the codebase" when cache is up-to-date
Running `vendor/bin/psalm` twice with cache enabled seems to cause the second run to just print `No files analyzed`, while the first  run prints `Psalm was able to infer types for X% of the codebase`.

Unfortunately, neither `--stats` nor `--no-reflection-cache` helps, only `--no-cache` does.

Is there another way to force the type coverage output **with** cache?

(Background: I'm using GitLab CI to capture the output and display the diff in merge requests and I noticed since about January 2021 that in some cases the difference couldn't be calculated, because the psalm job on `main` had just printed `No files analyzed`, especially after merging merge requests whose branches where up-to-date with `main`.)

caugner:
Fixes #5319.

Thanks to @orklah for pointing the corresponding code location out.

### DIFF
--- a/src/Psalm/Internal/Codebase/Analyzer.php
+++ b/src/Psalm/Internal/Codebase/Analyzer.php
@@ -29,6 +29,7 @@ use function substr;
 use function usort;
 use function array_values;
 use const PATHINFO_EXTENSION;
+use function implode;
 
 /**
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

Issue #5248, opened 2021-02-18
False positive with assert-if-true/if-false


klimick:
If assert function has `@psalm-assert-if-false` and `@psalm-assert-if-true` at the same time then it works incorrectly:
https://psalm.dev/r/91745556ee

Quick workaround is drop `@psalm-assert-if-false` or `@psalm-assert-if-true` from assert function:
https://psalm.dev/r/a5500b4ed2

But above workaround doesn't work in more difficult cases:
https://psalm.dev/r/408248f46f

weirdan:
Also this doesn't seem right: https://psalm.dev/r/65cbc43cff

weirdan:
@muglug it looks like Psalm generates inverted assertion (`!string`) for false return here, but should it really do that?

whsv26:
Encountered the same problem for Either monad. I'd like to do something like this, but unfortunately it's not working for both psalm-assert-if-true and psalm-assert-if-false at the same time.

```php
/**
 * @psalm-assert-if-true Left<L, R> $this
 * @psalm-assert-if-false Right<L, R> $this
 */
public function isLeft(): bool {}

/**
 * @psalm-assert-if-true Right<L, R> $this
 * @psalm-assert-if-false Left<L, R> $this
 */
public function isRight(): bool {}
```

klimick:
Issue is solved: https://psalm.dev/r/b0b6a03bd6
It seems in the #9209

klimick:
I hastened to close the issue. Sorry. Issue still exists.
https://psalm.dev/r/6125f6ef36

PR #5333, opened 2021-03-06
Provide return type of ReflectionParameter::getType()

### LINKED ISSUE #5258: ReflectionParameter::hasType() should assert return type of ReflectionParameter::getType()
If `ReflectionParameter::hasType()` returns `true`, then `ReflectionParameter::getType()` will always return `ReflectionType`.

If `ReflectionParameter::hasType()` returns `false`, then `ReflectionParameter::getType()` will always return `null`.

This is not currently understood by psalm.

https://psalm.dev/r/257f4e884d

dbrekelmans:
Fixes #5258

Thanks to @orklah for giving me some hints 👍

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/ReflectionParameterGetType.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/ReflectionParameterGetType.php
@@ -7,7 +7,7 @@ namespace Psalm\Internal\Provider\ReturnTypeProvider;
 use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
 use Psalm\Type;
 use ReflectionType;
-use function array_key_exists;
+use function is_string;
 
 class ReflectionParameterGetType implements \Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface
 {
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/ReflectionParameterGetType.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/ReflectionParameterGetType.php
@@ -43,7 +43,9 @@ class ReflectionParameterGetType implements \Psalm\Plugin\EventHandler\MethodRet
         $type = $event->getContext()->vars_in_scope[$scopedVarName];
 
         if ($type->isTrue()) {
-            return Type::parseString(ReflectionType::class);
+            return new Type\Union([
+                new Type\Atomic\TNamedObject(ReflectionType::class)
+            ]);
         }
 
         return Type::getNull();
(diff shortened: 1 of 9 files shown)


nusje2000 (approved):

weirdan:
@dbrekelmans mind fixing CS (https://circleci.com/gh/vimeo/psalm/22383)?

Tests failures are unrelated, fixed in #5335

src/Psalm/Internal/Provider/ReturnTypeProvider/ReflectionParameterGetType.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $scopedVarName = '$' . $stmt->var->name . '->hastype()';
+        if (!isset($event->getContext()->vars_in_scope[$scopedVarName])) {
+            return null;
+        }
+
+        $type = $event->getContext()->vars_in_scope[$scopedVarName];
+
+        if ($type->isTrue()) {
+            return Type::parseString(ReflectionType::class);
```
weirdan, line 46:
```suggestion
            return new Type\Union([
                new Type\Atomic\TNamedObject(ReflectionType::class)
            ]);
```
There's no need to go through parser as we know exactly what we need to return.

weirdan (changes requested):

src/Psalm/Internal/Provider/ReturnTypeProvider/ReflectionParameterGetType.php
dbrekelmans: Ah yes, I tried initially, but couldn't find out how to do it. Thanks!

weirdan (approved):

muglug:
First, thanks for this!

I think I prefer the option you presented in the original issue, following [this recent dictum](https://psalm.dev/docs/contributing/philosophy/#docblock-annotations-are-better-than-type-providing-plugins).

Given that `ReflectionParameter` is an immutable class the assertion would be entirely safe (though it requires a bugfix for Psalm to allow this specific syntax, which I'm working on now.

```php
<?php

/**
 * @immutable
 */
class ReflectionParameter {
  /**
   * @psalm-assert-if-true $this->getType() ReflectionType
   * @psalm-assert-if-false $this->getType() null
   */
  public function hasType() : bool {
      return true;
  }
    
  public function getType() : ?ReflectionType {
   	return null;   
  }
}
```

The PR is not for nought, as I'll re-use your tests.

orklah:
@muglug I'm kinda lost
The else part seems to correctly infer that the type is null: https://psalm.dev/r/82732b730e

However, there is no assert-if-false in the stub like in your example above:
https://github.com/vimeo/psalm/blob/master/stubs/Reflection.phpstub#L108

How does that works then?

weirdan:
> How does that works then?

I mentioned it here: https://github.com/vimeo/psalm/issues/5248#issuecomment-781339212
Looks like a bug to me.

PR #5335, opened 2021-03-06
Suppress test issues with Symfony functions used with `::class` constant

weirdan:
These appeared in symfony/symfony#40203


muglug:
Thanks!

PR #5339, opened 2021-03-07
config.xsd: fix default errorLevel

gharlan:
Default errorLevel seems to be `2` instead of `1`: https://github.com/vimeo/psalm/blob/c05a3ea0739c11770eecec1c8cd8126ca9e3ec7c/src/Psalm/Config.php#L848


gharlan:
Otherwise in phpstorm `errorLevel="1"` is greyed out:

Bildschirmfoto 2021-03-07 um 10 55 55

muglug:
Thanks!

PR #5340, opened 2021-03-07
docs: mention default level

staabm:
As the source indicates at

https://github.com/vimeo/psalm/blob/c05a3ea0739c11770eecec1c8cd8126ca9e3ec7c/src/Psalm/Config.php#L848

### DIFF
--- a/docs/running_psalm/error_levels.md
+++ b/docs/running_psalm/error_levels.md
@@ -4,6 +4,8 @@ You can run Psalm in at different levels of strictness from 1 to 8.
 
 Level 1 is the most strict, level 8 is the most lenient.
 
+When no level is explicitly defined, psalm defaults to level 2.
+
 Some issues are [always treated as errors](#always-treated-as-errors). These are issues with a very low probability of false-positives.
 
 At [level 1](#errors-that-only-appear-at-level-1) all issues (except those emitted for opt-in features) that Psalm can find are treated as errors. Those issues include any situation where Psalm cannot infer the type of a given expression.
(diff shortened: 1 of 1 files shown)


docs/running_psalm/error_levels.md
```diff
@@ -4,6 +4,9 @@ You can run Psalm in at different levels of strictness from 1 to 8.
 
 Level 1 is the most strict, level 8 is the most lenient.
 
+When no level is explicitly defined, psalm defaults to level 2.
+In case `totallyTyped` is enabled, psalm defaults to level 1.
```
staabm, line 8: Since totallyTyped is deprecated we might just drop this line for simplicity sake

docs/running_psalm/error_levels.md
```diff
@@ -4,6 +4,9 @@ You can run Psalm in at different levels of strictness from 1 to 8.
 
 Level 1 is the most strict, level 8 is the most lenient.
 
+When no level is explicitly defined, psalm defaults to level 2.
```
dbrekelmans, line 7: Are you sure about this? I thought psalm scans the codebase and determines a appropriate level.

staabm: I guess its correct, see the cited code in the PR description.

staabm: > Are you sure about this? I thought psalm scans the codebase and determines the appropriate level from that.

I guess there is some kind of `init` command which guesses a proper default level when you use it.
I am talking about what psalm does when no level is defined in the config file at all.

muglug:
Thanks!

Issue #5331, opened 2021-03-05
Undefined variables in string interpolation are not being caught


Ocramius:
Taken from the convo @ https://twitter.com/mattbrowndev/status/1367965245084491783 (sorry, I just happen to review too much code :| )

```
<?php

(function () : void {
    echo (fn(string $b) => "$b - $undefined")("hello");
})();
```

https://psalm.dev/r/d5662fa596

It should report `$undefined` as undefined, but it doesn't.

See https://3v4l.org/D8tJR for runtime verification.

weirdan:
Simplified and added one more variation: https://psalm.dev/r/9520a78b52

Ocramius:
Thanks, @muglug!

orklah:
And thanks @weirdan for the PR! :)

PR #5343, opened 2021-03-08
Allow undefined variable detection in arrow functions

weirdan:
Previously Psalm would assume that any variable it sees in the arrow function body is defined (and mixed, if it's not available in the outer scope). This prevented undefined variable detection. Dropping that assumption allows it to work.

Fixes vimeo/psalm#5331


muglug:
Thanks!

Issue #5320, opened 2021-03-04
ImplicitToStringCast not emitted for arrays


weph:
Given this code:

```php
/**
 * @psalm-param array<array-key, string> $bars
 */
function foo(array $bars): void
{
}

class Bar {
    public function __toString()
    {
        return 'bar';
    }
}

foo([new Bar()]);
```

I'd expect an ImplicitToStringCast error to be emitted. But it doesn't.
After changing the array to a list, however, the error will be emitted

```php
/**
 * @psalm-param list<string> $bars
 */
function foo(array $bars): void
{
}

class Bar {
    public function __toString()
    {
        return 'bar';
    }
}

foo([new Bar()]);
```

https://psalm.dev/r/ee207d52b3

PR #5344, opened 2021-03-08
Emit ImplicitToStringCast in more places

weirdan:
Fixes vimeo/psalm#5320

`to_string_cast` is set on successful comparison, thus it needs to always bubble up (it will be ignored in UnionTypeComparator if some part does not match).


orklah:
Nice!

Does that solve https://github.com/vimeo/psalm/issues/4235#issuecomment-696862910 too?

weirdan:
It does, and I fixed property assignments in Psalm codebase, including the one you pointed out.

orklah:
Great! Thanks! You're on fire today :D

muglug:
Thanks!

Issue #5347, opened 2021-03-08
Psalm thinks explode() can return false


simPod:
I have enabled `ignoreInternalFunctionFalseReturn="false"`

(I cannot do it in playground) https://psalm.dev/r/96ef362b30

`explode()` returns false when value is empty string.

I asserted the value is not empty string but psalm still reports it can return false

> Cannot iterate over falsable var false|non-empty-list<string>

weirdan:
> `explode()` returns false when value is empty string.

Actually it returns false when *separator* is empty.
Reproduced in https://psalm.dev/r/6b3a3e88a6

simPod:
Ah, u're right. I had delimiter in constant as `const D = ',';`. My brain somehow ruled it out and rather focused on exploded value.

PR #5350, opened 2021-03-08
Provide more specific explode() return type

weirdan:
Fixes vimeo/psalm#5347

`explode()` now omits `false` from the return type when separator is a definitely non-empty string.


muglug:
Thanks!

Issue #5342, opened 2021-03-07
False-positive RedundantConditionGivenDocblockType


gharlan:
https://psalm.dev/r/47d24a5e84

After the `is_string` condition, `array` is missing from type for `$param`.

PR #5351, opened 2021-03-08
Added test to keep Union commutative

weirdan:
Union should be a commutative operation, such that `A|B = B|A`. This seems to be the case with types parsed from strings, but not for unions created directly in the code.

E.g.
```php
new Union([Atomic::create('int'), Atomic::create('positive-int')])
```
currently results in `positive-int`, even though `int` should subsume it.

Refs vimeo/psalm#5342

This PR just provides a failing test, as I'm not sure how best to address the issue.

### DIFF
--- a/tests/UnionTest.php
+++ b/tests/UnionTest.php
@@ -2,6 +2,7 @@
 
 namespace Psalm\Tests;
 
+use Psalm\Internal\Type\TypeCombiner;
 use Psalm\Type\Atomic;
 use Psalm\Type\Union;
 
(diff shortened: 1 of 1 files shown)


weirdan:
Cases highlighted by `TypeCombiner` directly affect users, as the resulting type depends on the order in which types are specified in docblocks: https://psalm.dev/r/ed874649d5

muglug:
I'm not sure I really see this as an issue. When union types are constructed manually their constituent types should be exclusive of one another.

weirdan:
By 'manually' you mean in Psalm's code? Or docblocks?

Either way I'm not sure requiring constituents to be exclusive is practical. I've seen types such as `S of object | Envelope<S of object>` used, where obviously Envelope is an object.

klimick:
Assertions in `foo` and `bar` are not equivalent because union is not truly commutative?
https://psalm.dev/r/24a5cc3d98

orklah:
@weirdan could you rebase? I'm curious to see what was the result of the test

weirdan:
@orklah done

tests/UnionTest.php
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Psalm\Tests;
+
+use Psalm\Internal\Type\TypeCombiner;
+use Psalm\Type\Atomic;
+use Psalm\Type\Union;
+
+class UnionTest extends TestCase
+{
+    private const KNOWN_ATOMICS = [
```
orklah, line 11: never-return => never
positive-int => `int<1, max>`
html-escaped-string => delete
false-y => delete
empty => delete

I think it should be better after that

Issue #3909, opened 2020-07-30
amphp/amp is required both in require and require-dev


fox91:
Using `composer validate` command gives a warning:
```
amphp/amp is required both in require and require-dev, this can lead to unexpected behavior
```
This leads to validation problems in any project that is using `psalm` and uses this command: `composer validate --with-dependencies`.

Which of the two versions is needed? The one under `require` or the one under `require-dev`?
Is it possible to remove the superfluous one?

fox91:
> Hey @fox91, can you reproduce the issue on https://psalm.dev ?

No, I can't as the problem is in the `composer.json` file.

muglug:
From memory the higher version is used to verify some Psalm-specific amphp behaviour, but Psalm can happily run on amphp 2.1

fox91:
> From memory the higher version is used to verify some Psalm-specific amphp behaviour, but Psalm can happily run on amphp 2.1

But in `require` or in `require-dev`?

remorhaz:
That may be okay to `require` _and_ `require-dev` the same package, if dev environment requires higher version.

fox91:
It may be correct, but `can lead to unexpected behavior` (cit).
As a developer of this package I could use only the most recent version (`require-dev`) and validate against it but a user can install it using an older version (`require`) that could not be compatible...

orklah:
This was fixes

PR #5352, opened 2021-03-08
Remove duplicate entry for 'amphp/amp' from composer.json file

junaidbinfarooq:
Removes the duplicate entry for 'amphp/amp'  from the require-dev section of _composer.json_ file.
Updates the entry in the require section to the latest version used earlier.


weirdan:
Related discussion: #3909

muglug:
Thanks!

Issue #3546, opened 2020-06-08
Possible improvement to min/max return type inference


jaikdean:
This case throws up a false-positive mixed argument info notice. It might be impractical to resolve, but would be useful if it could be.

https://psalm.dev/r/58124719cd

PR #5353, opened 2021-03-08
add return provider for min/max

orklah:
This should fix https://github.com/vimeo/psalm/issues/3546


orklah:
Note: at first I tried to add a stub but I struggled to match these 3 criterias:
- forbidding only one parameter (which is not an array) there is a test for that
- When more than one parameter is given, we must return an union of every type in parameters
- When one array is given as parameters, we must return an union of the types in the array

So I created a ReturnTypeProvider for that

muglug:
Thanks!

PR #5355, opened 2021-03-09
Mark ceil() return type as float

### LINKED ISSUE #5354: ceil() returns float, not int
https://psalm.dev/r/5bc7fb82f2

Having set `strictBinaryOperands="true"` psalm says `Cannot add ints to floats`. `ceil()` returns integerish float value

simPod:
https://www.php.net/manual/en/function.ceil.php

The return value of ceil() is still of type float as the value range of float is usually bigger than that of int.

Resolves #5354


muglug:
Thanks!

Issue #5359, opened 2021-03-09
Uncaught RuntimeException after changing the $config global in an autoloader


mauriciofauth:
When registering a custom autoloader with the following code, Psalm will crash due to an uncaught `RuntimeException` when writing error baseline to file.
 
```php
<?php
$GLOBALS['config'] = null;
```

```
vendor/bin/psalm --set-baseline=baseline.xml
```

```
Uncaught RuntimeException: PHP Error: Trying to get property 'include_php_versions_in_error_baseline' of non-object in vendor/vimeo/psalm/src/psalm.php:711 in vendor/vimeo/psalm/src/Psalm/Internal/ErrorHandler.php:66
Stack trace:
#0 vendor/vimeo/psalm/src/psalm.php(711): Psalm\Internal\ErrorHandler::Psalm\Internal\{closure}()
#1 vendor/vimeo/psalm/psalm(2): require_once('/home/...')
#2 {main}
(Psalm dev-master@3c66b755e2468b0976979768972d298f4ba25651 crashed due to an uncaught Throwable)
```

PHP 7.4.15.

weirdan:
There are more variables Psalm uses that can be destroyed this way. We should avoid using global scope as much as possible.

PR #5366, opened 2021-03-10
Wrap entrypoints into IIFE to protect their variables

weirdan:
Fixes vimeo/psalm#5359

### DIFF
--- a/src/Psalm/IssueBuffer.php
+++ b/src/Psalm/IssueBuffer.php
@@ -23,6 +23,7 @@ use function ob_start;
 use function sprintf;
 use Psalm\Internal\Analyzer\IssueData;
 use Psalm\Internal\Analyzer\ProjectAnalyzer;
+use Psalm\Internal\ExecutionEnvironment\BuildInfoCollector;
 use Psalm\Issue\CodeIssue;
 use Psalm\Issue\UnusedPsalmSuppress;
 use Psalm\Report\CheckstyleReport;
@@ -440,6 +441,7 @@ class IssueBuffer
      */
     public static function finish(
         ProjectAnalyzer $project_analyzer,
+        BuildInfoCollector $build_info_collector,
         bool $is_full,
         float $start_time,
         bool $add_stats = false,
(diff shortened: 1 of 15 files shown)


src/Psalm/IssueBuffer.php
```diff
@@ -542,7 +544,7 @@ function (IssueData $d1, IssueData $d2) : int {
         }
 
         $source_control_info = null;
-        $build_info = (new \Psalm\Internal\ExecutionEnvironment\BuildInfoCollector($_SERVER))->collect();
```
weirdan, line 545: This caused issues when `$_SERVER` was destroyed by autoloader

tests/EndToEnd/PsalmRunnerTrait.php
```diff
@@ -42,7 +42,7 @@ private function runPsalm(
             $process->mustRun();
         } else {
             $process->run();
-            $this->assertGreaterThan(0, $process->getExitCode());
+            $this->assertEquals(2, $process->getExitCode());
```
weirdan, line 45: As Psalm now always return exit code 2 when it detects issues, we expect that when `$shouldFail === true`, and let unexpected exceptions be reported as test failures.

src/psalm.php
```diff
@@ -56,810 +57,830 @@
...
     /**
-     * @param string $arg
+     * @param array<int,string> $argv
+     *
+     * Note: the $argv parameter should not be renamed as it's used in the files this function includes via require_once
      */
-    function ($arg) use ($valid_long_options, $valid_short_options): void {
-        if (substr($arg, 0, 2) === '--' && $arg !== '--') {
-            $arg_name = preg_replace('/=.*$/', '', substr($arg, 2));
+    function (array $argv): void {
```
weirdan, line 72: We should probably move those closures into methods proper, e.g. `Psalm\Internal\Cli\Psalter::run()`

muglug:
Thanks, this never occurred to me as a potential problem, stupidly

PR #5373, opened 2021-03-11
fix usage of callable with array_map

### LINKED ISSUE #5314: array_map Ignores callable return type
### Problem
Psalm usually understands how `array_map` works but we've found a case where it doesn't and it's pretty weird.
Psalm understands the closure returns a `stdClass` (as seen in the trace) but array mapping over that closure results in `array<mixed>`.

https://psalm.dev/r/4c414bde70

It's particularly weird that if I include a `@var` annotation of exactly what the trace is telling me, psalm both tells me the annotation is redundant AND that the array map now returns the expected `array<stdClass>`.

orklah:
This should fix #5314 
Only closures were used to infer return type. I added callables.


muglug:
Thanks!

Issue #5371, opened 2021-03-11
False positive NonInvariantDocblockPropertyType with template.


AndrolGenhald:
I have an abstract `ItemCollection` that can hold abstract `Item`s, and a concrete `FooCollection` that holds concrete `Foo` items. When I redeclare the `$items` property in `FooCollection` it gives a `NonInvariantDocblockPropertyType` false positive even though it's using `Foo` for the template.

https://psalm.dev/r/d84346cfe5

AndrolGenhald:
Fixed a template syntax error, issue persists. https://psalm.dev/r/5063097d3b

weirdan:
Why would you need to redefine property at all? Psalm knows it's a `list<Foo>` already: https://psalm.dev/r/c13bb4fa11

AndrolGenhald:
Well I suppose that's fine. I had it for consistency with the rest of the class's properties having `@var` annotations, but if that's the official solution, is there or should there be something like `@psalm-var-inferred` to tell psalm to infer the type while leaving the `@var` annotation for non-psalm users? I haven't found the time to dig into psalm's code yet, so I don't know the complexity of fixing it, but as a user it seems like it'd be preferable for it to "just work".

weirdan:
From Psalm's type system perspective supporting this would actually be sound, and it already seems to be supported for other types: https://psalm.dev/r/8fa0853348. PHP does not allow it for native types (e.g. https://3v4l.org/9cWYX), but PHP doesn't have generics yet. This is similar to what we had before PHP supported covariant returns - people would put parent return type as native return type and specify descendant type via `@return` doblock tag.

AndrolGenhald:
Should probably also fix this false negative with templates: https://psalm.dev/r/40bd875b72

Issue #3407, opened 2020-05-19
Incorrect types when using Amp\Promise\all()


enumag:
It's very nice that Psalm has support for Amphp. However I found out that whenever I use `[$x, $y] = yield \Amp\Promise\all([ $promiseForX, $promiseForY ])` then the `$x` variable has incorrect type detected, causing various false-positives further in the code.

muglug:
Would you mind reproducing?

enumag:
psalm.dev doesn't work with libraries so I can only provide the code here...:

```
use Amp\Promise;
use function Amp\call;
use function Amp\Promise\all;

abstract class Z
{
	public function test(): Promise
	{
		return call(
			function () {
				[$x, $y] = yield all([$this->promiseX(), $this->promiseY()]);
				assert(is_int($x));
				assert(is_string($y));
			}
		);
	}

	/**
	 * @return Promise<int>
	 */
	abstract protected function promiseX(): Promise;

	/**
	 * @return Promise<string>
	 */
	abstract protected function promiseY(): Promise;
}
```

Psalm incorrectly thinks that `$x` is a `string` and reports this:

```
ERROR: DocblockTypeContradiction - src/Z.php:16:12 - Found a contradiction with a docblock-defined type when evaluating $x and trying to reconcile type 'string' to int (see https://psalm.dev/155)
                                assert(is_int($x));
```

enumag:
Thanks @muglug!

muglug:
Addendum: the templating for `Amp\Promise\all` converts an array shape into a regular array (losing offset information). I don't think there's a good way to template this (so your `assert`  calls are still necessary).

PR #5380, opened 2021-03-13
Disable property invariance checks for properties with templates

AndrolGenhald:
Very much a work in progress, comments welcome.
Fix for #5371

I would like to have property invariance checks work correctly, but for now I'll at least settle for disabling them for properties containing a template.

So far I've changed `Union::hasTemplate` to check for arrays, lists, class-string-maps, and iterables containing templates, and the tests all pass. If this is the desired behavior of `hasTemplate` there's also `TGenericObject`, and possibly `TCallable` and others I haven't considered yet.

Any advice is much appreciated. I'm also trying to figure out how to update ClassAnalyzer.php:708 to check if `$guide_property_storage` has a template that's used by `$property_storage`, and then maybe clone `$guide_property_storage` and set the template to what `$property_storage` has before doing the `->equals` check, but I bet there's a better way to do that, cloning doesn't sound great for performance.

### DIFF
--- a/src/Psalm/Type/Atomic.php
+++ b/src/Psalm/Type/Atomic.php
@@ -597,6 +597,21 @@ abstract class Atomic implements TypeNode
         // do nothing
     }
 
+    public function hasTemplate(): bool
+    {
+        if ($this instanceof TTemplateParam) {
+            return true;
+        }
+        
+        foreach ($this->getChildNodes() as $childType) {
+            if ($childType->hasTemplate()) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     public function equals(Atomic $other_type): bool
     {
         return get_class($other_type) === get_class($this);
--- a/src/Psalm/Type/Atomic.php
+++ b/src/Psalm/Type/Atomic.php
@@ -597,21 +597,6 @@ abstract class Atomic implements TypeNode
         // do nothing
     }
 
-    public function hasTemplate(): bool
-    {
-        if ($this instanceof TTemplateParam) {
-            return true;
-        }
-        
-        foreach ($this->getChildNodes() as $childType) {
-            if ($childType->hasTemplate()) {
-                return true;
-            }
-        }
-
-        return false;
-    }
-
     public function equals(Atomic $other_type): bool
     {
         return get_class($other_type) === get_class($this);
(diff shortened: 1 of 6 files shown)


AndrolGenhald:
Simplified now with `getChildNodes()` which I'd overlooked before, but it breaks a test for #3407. I think some checks are disabled for types with templates and now it's skipping checks that used to work properly?

src/Psalm/Type/Atomic.php
```diff
@@ -597,6 +597,21 @@ public function replaceTemplateTypesWithArgTypes(
...
     }
 
+    public function hasTemplate(): bool
+    {
+        if ($this instanceof TTemplateParam) {
+            return true;
+        }
+        
+        foreach ($this->getChildNodes() as $childType) {
+            if ($childType->hasTemplate()) {
```
muglug, line 607: This subtly changes the nature of this call, and probably breaks some existing assumptions.

I think it would be better to either have a function `hasNestedTemplateParam` for this specific check, or use an instance of `TemplateTypeCollector` which does basically the same thing.

AndrolGenhald: Yeah, I was thinking earlier that this is confusing since it makes `hasTemplate` work differently from `hasArray`, `hasObject`, etc. And then I noticed NodeVisitor and wondered if that wasn't the way to go. Of course what I'd _really_ like is to figure out how to support it properly, but I'm still working my way through understanding everything.

Side note, I noticed `Union::$types` uses `'array'` for both `TArray` and `TList`, which makes [this](https://psalm.dev/r/48b13f8b7e) a bit screwy. I was just thinking about separating them but that's no good because you'd still have [this issue](https://psalm.dev/r/c23d141cfa). Is it worth opening an issue or is that something that's not worth trying to support?

muglug: That's by design – `list` is a subtype of `array`, so `array<int, int>|list<string>` is always simplified to `array<int, int|string>`

AndrolGenhald:
I think this is working properly now to disable property invariance checks for templates, I'll open a new pull request if I manage to get the checks to work correctly with templates.

muglug:
Thanks!

Issue #5390, opened 2021-03-14
array_combine doesn't preserve non-emptiness


vudaltsov:
https://psalm.dev/r/cc20d5cd31

PR #5393, opened 2021-03-14
makes array_combine return non-empty-array for most cases

orklah:
This shoud fix https://github.com/vimeo/psalm/issues/5390

Note that the stubs are not dependant on the PHP version so array_combine is still specified as returning false although PHP 8 replaced the false return by a fatal error.


muglug:
> Note that the stubs are not dependant on the PHP version so array_combine is still specified as returning false although PHP 8 replaced the false return by a fatal error.

What's there is fine for now, but FYI `(PHP_MAJOR_VERSION is 8 ? Foo : Bar)` is a valid conditional type.

muglug:
Thanks!

stubs/CoreGenericFunctions.phpstub
```diff
@@ -66,7 +66,7 @@ function array_intersect_assoc(array $arr, array $arr2, array ...$arr3)
  * @param array<mixed, TKey> $arr
  * @param array<mixed, TValue> $arr2
  *
- * @return array<TKey, TValue>|false
+ * @return ($arr is non-empty-array ? non-empty-array<TKey, TValue>|false : array<empty, empty>|false)
```
vudaltsov, line 69: @muglug , @orklah , why `array<empty, empty>`? Shouldn't it also be `array<TKey, TValue>`?

orklah: Yeah, you're right, I assumed the opposite of non-empty-array would be an empty array but it's not true. I'm fixing that

PR #5395, opened 2021-03-15
fix stub for array_combine + remove false in PHP8 + improve tests

orklah:
Fix a mistake I made in https://github.com/vimeo/psalm/pull/5393
Also improve test to cover my mistake and remove the false return in PHP 8


muglug:
Thanks!

Issue #5382, opened 2021-03-13
Taint sink dependent on structure of an array


mortenson:
Hi there - I'm writing Psalm integration for Drupal.

In Drupal, "render arrays" are used to represent elements that end up being HTML. Some documentation for this is here https://www.drupal.org/docs/drupal-apis/render-api/render-arrays

The structure of the array determines if something is safe or not. For instance, the value for the key "#markup" is sanitized for XSS and outputs filtered HTML. "#children" outputs completely raw text. In other cases, keys like "#lazy_builder" can be used to provide a callback function, which could lead to RCE. Render arrays are usually deeply nested, so these keys could appear at any nesting level in the array.

How can we check if a taint source is being assigned to a "dangerous" array key with Psalm? I'd be fine with even considering any array assignment with certain keys/structures to be a taint sink, even if it never hits a function to turn the render array into HTML. I have tried writing a plugin for this but haven't been able to add a taint sink to anything (see https://github.com/vimeo/psalm/issues/3828#issuecomment-792865166)

Here's a code snippet if it's helpful: https://psalm.dev/r/b211d1e726

muglug:
OK, that's rather tricky.

Thinking out loud here:

In my mind the array value only becomes a taint source if the key has a particular value. It would mean [making this conditional](https://github.com/vimeo/psalm/blob/90fd1c5d1ffcf055c4a8ca8073905cf22530e6d2/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php#L351-L370) on a particular array key.

There would need to be a new plugin event to override this behaviour. I think it could be generalised to be useful in any conditional that carries [this check](https://github.com/vimeo/psalm/blob/90fd1c5d1ffcf055c4a8ca8073905cf22530e6d2/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php#L343).

It would take a `PhpParser\Node` and some other common information, and return `bool` – `true` when it's ok to taint, `false` when it's not. Let me know if you'd be interested in creating a PR for such a plugin, given this would be an ideal use-case.

Here it would only allow the taint to happen when the type of the `ArrayItem` node key is a literal with the value `#children` or `#plain_text`

mortenson:
@muglug Yes! I'm happy to jump in with a bit of guidance. I didn't realize there are already conditional checks (annotations?) about arrays, would this support deeply nested keys too?

muglug:
[...]

PR #5398, opened 2021-03-15
Added event to prevent tainting.

mortenson:
Per https://github.com/vimeo/psalm/issues/5382 this PR adds a new event that allows plugins to prevent tainting in specific scenarios. I've also added an example plugin/test that shows how you can inspect the structure of an array to determine if it should be considered a taint source.

This is a WIP, so let me know if you want the event name or test changed - once that's decided I can dispatch the event from more analyzers if you'd like.


src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
```diff
@@ -220,7 +224,8 @@ private static function analyzeArrayItem(
         Context $context,
         ArrayCreationInfo $array_creation_info,
         int $int_offset,
-        PhpParser\Node\Expr\ArrayItem $item
+        PhpParser\Node\Expr\ArrayItem $item,
+        Codebase $codebase = null
```
muglug, line 228: this is never not null

mortenson: I was thinking of backwards compatibility for changing the method signature but I see it's private now.

mortenson: Fixed

muglug:
Ok, extra complication: I think this plugin should instead inform these two arguments: https://github.com/vimeo/psalm/blob/dd4d970a37a1a4dde56f7903333e2ef3c5481394/src/Psalm/Internal/Codebase/DataFlowGraph.php#L30-L31 in the call here: https://github.com/vimeo/psalm/blob/df43126626715cfe4e641c2017a6b41a8a43c433/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php#L371

It may be the case that you want to remove the `html` taint but preserve the `sql` taint, for example. In this case you'd want to remove `html` taints.

So instead of being a boolean I think it wants to provide two separate arrays of added and removed taints at every node it's used on. If both those arrays are empty there's no net effect

muglug:
I've updated the PR along those lines! The main benefit is it allows people to institute their own custom taints in a smarter fashion.

mortenson:
@muglug That looks good to me - the test I added still works as expected so I think this would give me a path forward on the render array problem. Would you like me to rename the new test method and add this event dispatch to other analyzers? (If you have time to do this don't wait on me though)

muglug:
> Would you like me to rename the new test method and add this event dispatch to other analyzers?

That would be great!

mortenson:
@muglug StaticCallAnalyzer, AtomicPropertyFetchAnalyzer, and ArrayFetchAnalyzer didn't have access to Context which is needed for this event, should I add $context as an optional parameter to the relevant public static methods on those classes?

All other analyzers that used that `TaintedInput` check you mentioned in the issue have been updated.

mortenson:
I did the above, feel free to clean up the new event invocations/calls as needed.

muglug:
Thanks, this is a great effort

mortenson:
Thank you @muglug !

PR #5400, opened 2021-03-15
iterator_to_array must always return array-key

### LINKED ISSUE #5396: iterator_to_array never returns mixed key, it throws TypeError instead
https://psalm.dev/r/ce986a03b8

I think the issue raised by Psalm is not correct. While `iterator_to_array` does throw a `TypeError` in case of a non-scalar key (https://3v4l.org/BBOJF), the returned type of `iterator_to_array(T, true)` is never `array<mixed, T>`, it is `array<array-key, T>`.

orklah:
This PR should fix #5396. When the return of iterator contains a mixed templated key, we can narrow it to array-key because PHP will throw a TypeError if it's not the case.


muglug:
Thanks!

PR #5401, opened 2021-03-15
adapt bcdiv for PHP8 in stubs

orklah:
Now that I learned we can use PHP_MAJOR_VERSION in stubs, I checked changed functions in callmap_80 that appear in stubs and this was the only one


muglug:
Thanks!

PR #5409, opened 2021-03-16
Add source attribute to checkstyle report

dsamburskyi:
CheckStyle report supports "source" attribute (example [here](https://github.com/checkstyle/checkstyle/blob/e09397dab3c24c294e7559aeb28fcd461b9cd9e1/src/test/resources/com/puppycrawl/tools/checkstyle/ant/checkstyleanttask/ExpectedCheckstyleAntTaskXmlOutput.xml#L4)).
CheckStyle parsing tools are getting category and type from the "source" attribute (Jenkins reporting [plugin](https://github.com/jenkinsci/analysis-model/blob/18a6d7cdb4fa1cb485d36a2f328667f171354102/src/main/java/edu/hm/hafner/analysis/parser/checkstyle/CheckStyleParser.java#L88-L99)).
This change will allow the reporting tools to define the meta data correctly. See the screenshot from Jenkins with applied change.
image

The source is based on getParentIssueType() and falls back to Issue's parent class if parent issue type is null.

### DIFF
--- a/src/Psalm/Report/CheckstyleReport.php
+++ b/src/Psalm/Report/CheckstyleReport.php
@@ -1,8 +1,10 @@
 <?php
 namespace Psalm\Report;
 
-use function htmlspecialchars;
+use Psalm\Config;
 use Psalm\Report;
+use ReflectionClass;
+use function htmlspecialchars;
 use function sprintf;
 
 class CheckstyleReport extends Report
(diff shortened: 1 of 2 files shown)


muglug:
Hmm. I never really intended for the parent _classes_ to be part of the public API. 

> This change will allow the reporting tools to define the meta data correctly.

What about using the reporting level as a category instead? Each issue has an assigned reporting level.

dsamburskyi:
But reporting level is basically `error`, `info`, etc, right? Or am I looking into wrong place?
$reporting_level for all issues from `testCheckstyleReport()` are identical to what's in `severity` attribute. This won't be very helpful as all issues will fall into just few categories.

On the screenshot below is how the report is looking with existing functionality. There's no way to filter issues by it's category or type. For example, someone wants to look at all UnusedClass errors.
image

I'm Ok to try to implement this in a different way if you think this is worth fixing, just need some guidance on where to look at :)

muglug:
Every issue has an `ERROR_LEVEL` constant that dictates how severe Psalm thinks the issue is: https://github.com/vimeo/psalm/blob/master/src/Psalm/Issue/PossiblyNullIterator.php#L6

caugner:
@dsamburskyi I think it would be cool to use the `ERROR_LEVEL` as the category, probably even more useful than the parent class. What do you think?

src/Psalm/Report/CheckstyleReport.php
```diff
@@ -20,12 +22,21 @@ public function create(): string
                 $issue_data->message
             );
 
+            /** @var class-string $parent_issue */
+            $parent_issue = '\Psalm\\Issue\\' . $issue_data->type;
+            $parent_issue_classname = (new ReflectionClass($parent_issue))
+                ->getParentClass()
+                ->getShortName();
+            $parent_issue_type = Config::getParentIssueType($issue_data->type) ? : $parent_issue_classname;
+            $issue_source = 'Psalm.' . $parent_issue_type . '.' . $issue_data->type;
```
caugner, line 31:
```suggestion
            $issue_source = 'Psalm.ErrorLevel' . $error_level. '.' . $issue_data->type;
```

caugner (changes requested):
LGTM, but it would be nice to use the ERROR_LEVEL.

See: https://github.com/vimeo/psalm/search?q=ERROR_LEVEL+path%3Asrc%2Fpsalm%2FIssue

orklah:
I'll close this as stale, feel free to ping me to reopen

Issue #5405, opened 2021-03-16
BC break in 4.6.3


adamturcsan:
A BC Break has been introduced in a patch version.

https://github.com/vimeo/psalm/commit/205fdd197e914a5377053f9378a7112236a83b50#diff-58b56f2d1a81b22decaddcb6bce7b755662ebf12b8e17e20fd60dc5809787c99

Plugins relying on IssueBuffer::finish are now failing.
https://github.com/Roave/you-are-using-it-wrong/issues/114

adamturcsan:
> Hey @adamturcsan, can you reproduce the issue on https://psalm.dev ?

It's not a user facing issue, so no, I cannot.

weirdan:
Would tagging a new patch version without this change resolve your issue?

PR #5410, opened 2021-03-16
Fix BC break introduced in 4.6.3

weirdan:
Fixes vimeo/psalm#5405

Requires new patch release.

This PR reverts changes to the signature of `IssueBuffer::finish()` and introduces separate method to be used to capture `$_SERVER`


muglug:
Thanks!

PR #5414, opened 2021-03-17
Support template property invariance (fixes #5371)

### LINKED ISSUE #5371: False positive NonInvariantDocblockPropertyType with template.

AndrolGenhald:
I've gone through at least 10 different attempts each with different subtle bugs. I thought I had it this time but I didn't account for `@template-covariant` and I'm at a loss as to how to fix it. Suggestions welcome!


AndrolGenhald:
[This](https://github.com/vimeo/psalm/blob/3046468d1e151a22a19d4c8cb2988ce3cf2633c4/src/Psalm/Storage/ClassLikeStorage.php#L304-L306) confused me a bit. The comment seems to imply that `$template_types` has templates for more than just the current class (ie, parents), but that doesn't seem to be the case. Is it just this way to make it easy to merge `$template_types` from different classes, or is there something I'm missing?

muglug:
> The comment seems to imply that $template_types has templates for more than just the current class (ie, parents), but that doesn't seem to be the case. Is it just this way to make it easy to merge $template_types from different classes, or is there something I'm missing?

It's just that way to make it easier to merge – speaking of which, this is great, thanks!

PR #5416, opened 2021-03-17
Fix composer major.minor version detection (fixes #5415)

### LINKED ISSUE #5415: Composer PHP version detected incorrectly
I have a project requiring `"php": "^7.2.5"`. It looks like Psalm only cares about semver major and minor versions, but the way it detects it is by finding the lowest MAJOR.MINOR version that satisfies the composer required version. Since this project requires 7.2.5, 7.2 does not satisfy that, but 7.3 does, so it returns 7.3. I would expect it to detect 7.2.

https://github.com/vimeo/psalm/blob/3046468d1e151a22a19d4c8cb2988ce3cf2633c4/src/Psalm/Config.php#L2056-L2060

AndrolGenhald:
Fix composer php version detection.


src/Psalm/Config.php
```diff
@@ -2053,8 +2054,10 @@ private function getPHPVersionFromComposerJson(): ?string
             $php_version = $composer_json['require']['php'] ?? null;
 
             if (\is_string($php_version)) {
+                $version_parser = new VersionParser();
+                $min_php_version = $version_parser->parseConstraints($php_version)->getLowerBound()->getVersion();
                 foreach (['5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0'] as $candidate) {
-                    if (Semver::satisfies($candidate, $php_version)) {
+                    if (str_starts_with($min_php_version, $candidate)) {
```
AndrolGenhald, line 2060: Can someone do a sanity check on this? I'm not 100% sure str_starts_with is correct in all cases.

orklah: It's PHP 8 only to begin with :)

AndrolGenhald: I'm pretty sure psalm is using symfony's PHP 8 polyfill which includes this. It's working fine for me on PHP 7.3.

AndrolGenhald: It looks like the phar is failing on circle-ci though, so maybe that does need changed? All the unit tests are working fine though, so maybe the phar isn't being built correctly to include the polyfill?

orklah: well, it's using symfony/console which require symfony/polyfill-php80 so it's an indirect dependancy. Maybe add the real dependancy?

AndrolGenhald: `symfony/console` is a non-dev dependency, which depends on symfony/polyfill-php80, so it seems like this should work and there's something wrong with the phar. Granted, if the polyfill is depended on it should be added as a direct dependency too, there's no guarantee `symfony/console` won't change.

AndrolGenhald: Hmm, doesn't seem to have worked. I'm happy to change it if we don't want the additional dependency, but I would like to know why it's not working.

weirdan: > Can someone do a sanity check on this? I'm not 100% sure `str_starts_with` is correct in all cases.

It'll select wrong version for two-digit minors: https://3v4l.org/jDb9B
I don't think PHP ever had two-digit minor, but theoretically it could happen.

weirdan: Is there any reason you got rid of `Semver::satisfies()`?

orklah: https://github.com/vimeo/psalm/issues/5415#issuecomment-801281452 would be one I guess

weirdan: I think changing the call to `Semver::satisfies($min_php_version, "~$candidate.0")` would fix the issue: e.g. `7.2.5` would satisfy `~7.2.0`, as expected.

AndrolGenhald: > [#5415 (comment)](https://github.com/vimeo/psalm/issues/5415#issuecomment-801281452) would be one I guess

Yep, that.

AndrolGenhald: > I think changing the call to `Semver::satisfies($min_php_version, "~$candidate.0")` would fix the issue: e.g. `7.2.5` would satisfy `~7.2.0`, as expected.

Thanks, I think that'll work! I was stuck thinking about tweaking `$min_php_version`, I didn't even consider changing `$candidate`.

muglug:
Thanks!

Issue #2996, opened 2020-03-19
json_encode cannot return false if JSON_THROW_ON_ERROR is passed


iggyvolz:
https://psalm.dev/r/5208d10bbe

If JSON_THROW_ON_ERROR is passed to json_encode's options, it is not able to return false and is suitable to be used in functions such as __toString.

Behaviour is described at https://www.php.net/manual/en/json.constants.php - may need to check for PHP version >7.3 and check that  JSON_PARTIAL_OUTPUT_ON_ERROR is not set (the wording is a bit ambiguous there)

orklah:
Note: The original code doesn't produce issues anymore. Not because JSON_THROW_ON_ERROR was supported but because the `false` must have been handled as `ignorable`. This means it was not totally fixed (Psalm still think there can be false in there but don't bother reporting it depending on configs). The PR above should fix the core issue

PR #5417, opened 2021-03-17
json_encode does not return false if JSON_THROW_ON_ERROR is in flags

orklah:
This fixes https://github.com/vimeo/psalm/issues/2996

A few caveats:
- JSON_THROW_ON_ERROR  appeared on PHP 7.3. If anyone push the literal 4194304 in PHP 7.2, the return type will be funky
- I didn't use int-mask because there are 29 JSON_* constants and I thought Psalm would not like that. The consequence is that if someone combine flags, The return type will go back to |false

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1123,7 +1123,7 @@ function json_decode(string $json, ?bool $associative = null, int $depth = 512, 
 /**
  * @psalm-pure
  *
- * @return string|false
+ * @return ($flags is 4194304 ? string : string|false)
  *
  * @psalm-flow ($value) -> return
  * @psalm-ignore-falsable-return


AndrolGenhald:
>     * JSON_THROW_ON_ERROR  appeared on PHP 7.3. If anyone push the literal 4194304 in PHP 7.2, the return type will be funky

Would a conditional return type with something like `PHP_MAJOR_VERSION > 7 || PHP_MAJOR_VERSION === 7 && PHP_MINOR_VERSION >= 3` fix that? I haven't used them much so I'm not sure how much logic is supported.

orklah:
Yeah, this is possible, but given you can't use &&, you have to create branches for everything. It'll make a big mess for small gains. Especially if we want to handle more flags in the future

muglug:
Thanks!

PR #5428, opened 2021-03-18
Fix object-like array keys when combining string and automatic keys (fixes #5427).

### LINKED ISSUE #5427: ArrayAnalyzer int_offset broken
For object-like arrays with some keys specified and some not `$int_offset` assigns the wrong key.
https://psalm.dev/r/1066dc35a8
~~It's out of order too (assuming object-like arrays are supposed to retain order).~~ Edit: sorting properties on non-list keyed arrays seems to be intentional.

AndrolGenhald:


AndrolGenhald:
Psalm ran fine locally, is this an actual cache issue or does changing psalm itself mean you have to run `psalm --clear-cache` to get valid results?

AndrolGenhald:
I've fixed the issues, but it brings up a separate issue: should object-like arrays be sorted? Right now, if you change an element returned by `providerValidCodeParse` to swap `assertions` and `error_levels` so `error_levels` comes first, I don't think psalm will complain, but `@dataProvider` (presumably) only cares about order, so it would break.
Of course, for this case if one of the optional keys is specified all of the previous optional keys also need to be specified, which is a bit more complicated.

Also, it seems a bit hacky that a bunch of tests use automatic integer keys for `assertions` and `error_levels`. Should I do a separate pull request to change all of those to use the descriptive keys? And maybe add descriptive keys for `providerInvalidCodeParse` at the same time?

weirdan:
> but `@dataProvider` (presumably) only cares about order, so it would break.

Correct. At least it was that way before named parameters.

Variable-length tuple (that's what we need here, and what is [available in TS](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#optional-elements-in-tuple-types)) currently doesn't seem to be representable in Psalm types. Of course, being able to do something like `@return parameters-of<self, 'testSomething'>` (like [you can do in TS](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype)) would be even better.

muglug:
The only object-like arrays that are order-specific are int-offset lists like `array{int, string, int}`

> Of course, being able to do something like `@return parameters-of<self, 'testSomething'> would be even better.

I think `parameters-of<self::testSomething>` would be more in line with `value-of<...>` but this could be a useful addition for PHPUnit users

weirdan:
> I think parameters-of<self::testSomething> would be more in line with value-of<...> 

It would make sense to accept any valid callable reference there, but I don't think `self::testSomething` is a valid way to reference a method. One day we may be able to store closures into constants (allowing more constructs in static initializers has been discussed recently on internals ML), and it will become ambiguous.

weirdan:
It's also interesting that in TS you can define `Parameters<T>` completely in userland. It's just using a generic type + `infer` keyword.

Issue #2852, opened 2020-02-20
DuplicateArrayKey - String keys are not supported in unpacked arrays


bendavies:
https://psalm.dev/r/b55ef4e286

> muglug: @bendavies mind creating an issue for that? I think it should only warn for array<string, string>

PR #5432, opened 2021-03-19
Fix array spread false negative that crashes (fixes #5431, reverts #2852)

### LINKED ISSUE #5431: Array spread operator false negative
Caused by fix to #2852. It might be slightly annoying to have to annotate with `list<mixed>` or `array<int, mixed>` instead of just `function test(): array {}`, but this is definitely a false negative considering that this code crashes: https://psalm.dev/r/ef5d2a6084 https://3v4l.org/9q0Eq

AndrolGenhald:


muglug:
Thanks!

PR #5433, opened 2021-03-19
Fix bugs with array spread operator (fixes #5421)

### LINKED ISSUE #5421: Variadic parameters and non-empty-list
https://psalm.dev/r/7f75a28f23

In this example `$x` can be empty, but psalm understand this as non-empty-list. It is bug.

AndrolGenhald:
Branched from #5428 since they're somewhat related.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -109,8 +109,12 @@ class ArrayAnalyzer
         }
 
         if ($array_creation_info->all_list) {
-            $array_type = new Type\Atomic\TNonEmptyList($item_value_type ?: Type::getMixed());
-            $array_type->count = count($stmt->items);
+            if (empty($array_creation_info->item_key_atomic_types)) {
+                $array_type = new Type\Atomic\TList($item_value_type ?: Type::getMixed());
+            } else {
+                $array_type = new Type\Atomic\TNonEmptyList($item_value_type ?: Type::getMixed());
+                $array_type->count = count($array_creation_info->property_types);
+            }
 
             $stmt_type = new Type\Union([
                 $array_type,
@@ -199,7 +203,7 @@ class ArrayAnalyzer
             $item_value_type ?: Type::getMixed(),
         ]);
 
-        $array_type->count = count($stmt->items);
+        $array_type->count = count($array_creation_info->property_types);
 
         $stmt_type = new Type\Union([
             $array_type,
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
```diff
@@ -110,8 +109,12 @@ public static function analyze(
         }
 
         if ($array_creation_info->all_list) {
-            $array_type = new Type\Atomic\TNonEmptyList($item_value_type ?: Type::getMixed());
-            $array_type->count = count($stmt->items);
+            if (empty($array_creation_info->item_key_atomic_types)) {
+                $array_type = new Type\Atomic\TList($item_value_type ?: Type::getMixed());
+            } else {
+                $array_type = new Type\Atomic\TNonEmptyList($item_value_type ?: Type::getMixed());
+                $array_type->count = count($array_creation_info->property_types);
```
AndrolGenhald, line 116: Not sure if this is correct, but I'm also not sure what would be correct. `count($stmt-items)` is clearly wrong with unpacked arrays.
Should `count` be renamed to `min_count` as well?

src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
```diff
@@ -200,7 +203,7 @@ public static function analyze(
             $item_value_type ?: Type::getMixed(),
         ]);
 
-        $array_type->count = count($stmt->items);
+        $array_type->count = count($array_creation_info->property_types);
```
AndrolGenhald, line 206: Same as above.

AndrolGenhald:
Rebased onto master to simplify diff now that #5428 is merged.

src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
muglug: `count` is only normally set when it's known — if this any of the items are unpacked arrays then we shouldn't be setting an explicit type

AndrolGenhald: Ok, so `count` is always exact, and if exact count isn't known there's no minimum count or anything, it's just not checked at all.
It would still be good to set it when unpacking literal arrays: psalm.dev/r/fe1e86a493

muglug:
Thanks!

orklah:
> I'm not sure if that's because of the rebase or not.

It was. Github adds this button every time there is a new commit

PR #5437, opened 2021-03-19
Fix false-positive with non-empty object like array (fixes #5436)

### LINKED ISSUE #5436: Object like array with required key not always considered non-empty
https://psalm.dev/r/23869e7fec

AndrolGenhald:


muglug:
Thanks!

PR #5438, opened 2021-03-20
Use appveyor reporter instead of a file upload

weirdan:
Due to the size of the generated PHPUnit report some builds fail on AppVeyor ([example](https://ci.appveyor.com/project/muglug/psalm/builds/38298679)). This PR fixes the issue by using PHPUnit reporter that logs test results to AppVeyor via their API, in real time.


muglug:
Thanks!

Issue #5439, opened 2021-03-20
Order in which taints are added/removed from array items can affect results


mortenson:
I noticed a odd bug, of course immediately after landing #5382 - basically, it looks like if a taint is removed for an array item before another taint item is processed, that other taint item is ignored.

To demonstrate this better, you can checkout the branch `array-sorting-tainting` from https://github.com/mortenson/psalm, then run `./vendor/bin/phpunit tests/Config/PluginTest.php --filter testRemoveTaints`. You'll see that tests pass even though there's an array item that should have triggered an error. Here's the relevant change:

https://github.com/mortenson/psalm/commit/7d929d019e86286ec2a9761276796f5dfe376534

To cause the test to (rightly) fail, change the array key from "a" to "z". This will cause it to be processed later.

So, perhaps the problem here is that the added/removed taints are tied to the array variable, and not to each array item? Not sure yet.

PR #5444, opened 2021-03-20
Make taint source keys unique to the added taints.

mortenson:
This issue addresses #5439 by making sources keyed on the added taints. If you remove this change, the added test should fail to demonstrate the problem.


muglug:
Thanks!

Issue #5445, opened 2021-03-21
Unpacked arguments with string keys should be treated as named arguments


yakimun:
Hello!

As far as I can see, Psalm doesn't treat unpacked arguments with string keys properly.

There are two closely related examples:
* https://psalm.dev/r/d44648978e - `$a` from the array overwrites previously used argument
* https://psalm.dev/r/240dd34dce - the function doesn't have a parameter `$c`

Such mistakes in the code obviously pass standard linter check but lead to fatal errors at runtime.

PR #5446, opened 2021-03-21
Add proper handling of unpacked arguments with string keys

yakimun:
Hello!

These changes solve the problems with the unpacking of arrays with string keys described in #5445.
I'm not sure if any tests are needed here. If they are required, please point me to where they need to be placed, and I will add them.

Fixes: #5445.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
@@ -639,9 +639,81 @@ class ArgumentsAnalyzer
         $matched_args = [];
 
         foreach ($args as $argument_offset => $arg) {
-            if ($arg->unpack && $function_param_count > $argument_offset) {
-                for ($i = $argument_offset; $i < $function_param_count; $i++) {
-                    $arg_function_params[$argument_offset][] = $function_params[$i];
+            if ($arg->unpack) {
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
```diff
@@ -639,9 +639,70 @@ public static function checkArgumentsMatch(
...
+                                        )) {
+                                            // fall through
+                                        }
+                                    }
+
+                                    $matched_args[$candidate_param->name] = true;
+                                    $param_found = true;
+                                    break;
+                                }
+                            }
```
AndrolGenhald, line 689: `$param_found` should be set to `true` if the function has a variadic argument.
Haven't tested it myself yet, but it looks like this would cause a false-positive: https://psalm.dev/r/cac7929f84

AndrolGenhald:
> I'm not sure if any tests are needed here.

You can add the test cases from the issue you opened, maybe in ArgTest.php. Or maybe add a new file for named arguments. I've just been sticking them wherever they seem to fit best.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
yakimun: You are right. I pushed the fix. Thanks!

muglug:
Thanks!

PR #5449, opened 2021-03-22
Add support for object-like non-empty array.

AndrolGenhald:
Object-like arrays are already considered non-empty if they have a required key, but this is useful if all of the keys are optional but at least one of them must be set.

### DIFF
--- a/src/Psalm/Type/Atomic/TKeyedArray.php
+++ b/src/Psalm/Type/Atomic/TKeyedArray.php
@@ -57,6 +57,11 @@ class TKeyedArray extends \Psalm\Type\Atomic
      */
     public $is_list = false;
 
+    /**
+     * @var bool - if this array is required to be non-empty
+     */
+    public $non_empty = false;
+
     public const KEY = 'array';
 
     /**
@@ -94,7 +99,7 @@ class TKeyedArray extends \Psalm\Type\Atomic
         }
 
         /** @psalm-suppress MixedOperand */
-        return static::KEY . '{' . implode(', ', $property_strings) . '}';
+        return ($this->non_empty ? 'non-empty-array' : static::KEY) . '{' . implode(', ', $property_strings) . '}';
     }
 
     public function getId(bool $nested = false): string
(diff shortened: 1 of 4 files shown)


AndrolGenhald:
I believe all of the failed tests are due to https://github.com/vimeo/psalm/commit/9f3f420a5b1785d06b9b9170b5b953a6de04685e.

src/Psalm/Type/Atomic/TKeyedArray.php
```diff
@@ -120,7 +125,7 @@ function ($name, Union $type): string {
         }
 
         /** @psalm-suppress MixedOperand */
-        return static::KEY . '{' .
+        return ($this->non_empty ? 'non-empty-array' : static::KEY) . '{' .
```
AndrolGenhald, line 128: Should I pull this out into a private method? I was thinking about updating `getKey()` but that won't work because it has to be `'array'` since it's used for `Union::$types`. Maybe `getType()`?

muglug: This is fine here, I think.

muglug:
I'm not sure about this.

`non-empty-array<'a'|'b', string>` already works in this scenario

muglug:
(merging/rebasing should fix stuff)

AndrolGenhald:
It doesn't work as well in scenarios like `non-empty-array{a?: Foo, b?: Bar}`, where `!isset($arr['a'])` implies `isset($arr['b'])` and `$arr['b'] instanceof Bar`. I'm fine with not merging this though, you're the one that has to maintain it in the future, and it's not actually needed for what I wanted anyway (better annotation for `Union::$types`, but I recently realized not all of the keys are string literals, so it has to be `&non-empty-array<array-key, Atomic>` anyway, which I _think_ solves the non-empty issue I was running into).
You can close the pull request, but I'll leave the branch around in case it comes up again and you decide it's worthwhile. Assuming there haven't been enough changes that it wouldn't be easier to redo it.

muglug:
Yeah, I think I'm happier with `array{...}&non-empty-array<...>` even though it's more cumbersome, since the thing being described is itself pretty complex.

Issue #5453, opened 2021-03-22
Don't allow unpacking non-iterable


AndrolGenhald:
This should be an InvalidArgument: https://psalm.dev/r/410b622a1f
It's currently allowed because the type _can_ be an iterable, but it should be disallowed if the type can be anything other than iterable.

weirdan:
> This should be an `InvalidArgument`

More precisely, it should be `PossiblyInvalidArgument`.

AndrolGenhald:
I'm not sure I fully understand the difference between `InvalidArgument` and `PossiblyInvalidArgument`.

weirdan:
> Should psalm always emit PossiblyInvalidArgument when it's possible the argument has the correct type?

Yes, if an argument type has an intersection with the required parameter type then it's a possibly invalid call: https://psalm.dev/r/686b176d87

`InvalidScalarArgument`, however, currently doesn't have it's `Possibly*` counterpart, so this can be a bit confusing.

PR #5455, opened 2021-03-22
Improve @no-named-arguments support and variadics.

### LINKED ISSUE #5420: Variadic parameters and list type
Hello!

In PHP 8, it is possible to unpack an array with string keys into the argument list when calling functions.

This variadic function example https://psalm.dev/r/8374b93983 passes analysis without any errors. Annotation shows that the function must return a list, but we return an array containing any keys.

Is there a way to specify that only positional arguments could be passed to a variadic function?
It looks like `@no-named-arguments` should cover this case, but it doesn't work (https://psalm.dev/r/b3faaf9d3f).

Thanks!

### LINKED ISSUE #5454: Prevent argument unpacking iterable with non-array-key key
Iterables have scalar keys, this means when unpacking an iterable when calling a function psalm should show an issue if the key isn't an int or a string.
https://psalm.dev/r/c6f65aebef
https://3v4l.org/Y8Vbu

AndrolGenhald:
Handling of argument unpacking and variadics still needs a pretty big makeover, but this is a good start.
    
Fixes #5420
Improves #5453 (iterable works, array still causes issues)
Fixes #5454

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -310,6 +310,7 @@
             <xs:element name="MoreSpecificImplementedParamType" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="MoreSpecificReturnType" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="MutableDependency" type="PropertyIssueHandlerType" minOccurs="0" />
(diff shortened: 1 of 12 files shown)


AndrolGenhald:
I branched from `9f3f420^` and the tests ran fine locally (though some where skipped).

muglug:
Thanks!

PR #5461, opened 2021-03-23
allow dismissing return value of no-return pure functions

azjezz:
refs https://github.com/azjezz/psl/pull/163

no-return function either:
1. exit
2. throw exceptions
3. run forever

in all of these cases, it's okay not to use the return value.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -1003,7 +1003,20 @@ class FunctionCallAnalyzer extends CallAnalyzer
                 && !$context->inside_conditional
                 && !$context->inside_unset
             ) {
-                if (!$context->inside_assignment && !$context->inside_call && !$context->inside_use) {
+                /**
+                 * If a function is pure, and has the return type of 'no-return',
+                 * it's okay to dismiss it's return value.
+                 */
+                if (
+                    !$context->inside_assignment &&
+                    !$context->inside_call &&
+                    !$context->inside_use &&
+                    !(
+                        $function_call_info->function_storage &&
+                        $function_call_info->function_storage->return_type &&
+                        (array_values($function_call_info->function_storage->return_type->getAtomicTypes())[0] ?? null) instanceof Type\Atomic\TNever
+                    )
+                ) {
                     if (IssueBuffer::accepts(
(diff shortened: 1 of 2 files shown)


weirdan:
And you want to keep such functions pure to be able to call them in pure contexts, right?

azjezz:
exactly.

azjezz:
example

- `no-return`/`pure` : [invariant_violations](https://github.com/azjezz/psl/blob/a23f14859efebad0226b6c7cf4edb826e0ffa192/src/Psl/invariant_violation.php#L16)
- `pure`: [invariant](https://github.com/azjezz/psl/blob/a23f14859efebad0226b6c7cf4edb826e0ffa192/src/Psl/invariant.php#L21)
- `pure`: [log](https://github.com/azjezz/psl/blob/a23f14859efebad0226b6c7cf4edb826e0ffa192/src/Psl/Math/log.php#L18)

azjezz:
currently to avoid this i suppressed all `UnusedFunctionCall` ( https://github.com/azjezz/psl/blob/1.6.x/psalm.xml#L65 ), but i don't really want to do that

azjezz:
also relevant: https://github.com/azjezz/psl/pull/163#discussion_r599985666

azjezz:
build failure is unrelated.

muglug:
I think this is exactly the right approach, thanks!

While traditionally throwing an exception might be considered a side-effect, I think Psalm's notion of purity is safe enough

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
```diff
@@ -1003,7 +1003,20 @@ private static function checkFunctionCallPurity(
...
+                 * it's okay to dismiss it's return value.
+                 */
+                if (
+                    !$context->inside_assignment &&
+                    !$context->inside_call &&
+                    !$context->inside_use &&
+                    !(
+                        $function_call_info->function_storage &&
+                        $function_call_info->function_storage->return_type &&
+                        (array_values($function_call_info->function_storage->return_type->getAtomicTypes())[0] ?? null) instanceof Type\Atomic\TNever
```
muglug, line 1017: I replaced this with `$function_call_info->function_storage->return_type->isNever()`

azjezz: i didn't notice the method existed 😅

PR #5463, opened 2021-03-24
allow dismissing return value of pure functions with by-reference arguments

azjezz:
this is another case where to makes sense to call a pure function without using it's return value

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -1011,6 +1011,7 @@ class FunctionCallAnalyzer extends CallAnalyzer
                 if (!$context->inside_assignment &&
                     !$context->inside_call &&
                     !$context->inside_use &&
+                    !self::callUsesByReferenceArguments($function_call_info, $stmt) &&
                     !(
                         $function_call_info->function_storage &&
                         $function_call_info->function_storage->return_type &&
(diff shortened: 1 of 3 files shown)


azjezz:
converted into draft, this example should trigger an `UnusedFunctionCall` error, with this change, it does not.

```php
<?php

/**
 * @pure
 */
function bar(string $c, string &$str = ""): string
{
  $c .= $str;

  return $c;
}

/**
 * @pure
 */
function baz(): string
{
  $f = "foo";
  bar($f);

  return $f;
}

```

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
```diff
@@ -1034,4 +1035,42 @@ private static function checkFunctionCallPurity(
...
+                $parameter = $parameters[$index] ?? null;
+            }
+
+            if ($parameter && $parameter->by_ref) {
+                return true;
+            }
+        }
+
+        return false;
+    }
```
azjezz, line 1075: something doesn't feel right about this function here, i added as many tests as i could have thought of, all seem to be passing as expected, but need further review.

tests/UnusedCodeTest.php
```diff
@@ -779,6 +779,120 @@ function reverse(string $string): string
                         return strrev($string);
                     }'
             ],
+            'unusedByReferenceFunctionCall' => [
+                '<?php
+                    /**
+                     * @pure
+                     */
+                    function bar(string &$str): string
```
muglug, line 787: Uhh this is definitely not pure though. It may be allowed, but it shouldn't (and I don't like to add tests that future code changes will force me to fix)

azjezz: Personally i don't see the problem with it being pure, as i don't think by-ref arguments are a side effects.

but, if you want to consider functions with by-ref arguments as non pure, i think i can close this PR and work on another to forbid byref in pure context.

muglug: I _sort_ of see your point insofar as byref arguments could be viewed as outputs.

_However_ I think that the return type should always be the primary location for output in pure functions. I don't want to encourage

```php
function some_pure_function(string &$s) : void {}
```
over
```php
function some_pure_function(string $s) : string {}
```

– but I'm fine with 

```php
function some_pure_function(string &$s) : string {}
```

Basically I think a pure function should either _always_ exit, or _always_ return a (non-`void`) value.

LMK if you think that's too-harsh a constraint.

azjezz: i agree with you, personally i don't have any functions that behave either way, but we need to have correct errors in place.

for
```
function some_pure_function(string &$s) : void {}
```

i think it's best to show something like `A pure function must not have a 'void' return type.` ( not sure about the correct issue for this ) 

as for the other, they seem okay to me and there shouldn't be any type of error result from dismissing their result ( in case you only need to the by-reference argument result ) in which this PR fixes.

so i guess we can merge this + i will create another PR to forbid having pure functions with `void` return type.

muglug:
Thanks!

tests/UnusedCodeTest.php
janopae: Sorry for commenting on a decision which has already been work done upon.

But if this isn't allowed: https://psalm.dev/r/cca607305c

I really don't get why this should be allowed: https://psalm.dev/r/a9e67c6ce9

I'd argue that functions that return something *and* have mutated passed-by-reference arguments are more confusing than functions that only do either of those.

PR #5464, opened 2021-03-24
Fix mysqli_real_connect() signature

morozov:
The optional arguments of `mysqli_real_connect()` except for `$flags` are also nullable ([ref1](https://github.com/php/php-src/blob/7e9f6d2a48f8ff6ae458252d395eec1b1d9e6f14/ext/mysqli/mysqli.stub.php#L741-L750), [ref2](https://github.com/php/php-src/blob/da011a312a6c6cd7ff12fe1aa0de1e33fba2f167/ext/mysqli/mysqli_nonapi.c#L81)). `$flags` seems to be nullable as well ([ref](https://github.com/php/php-src/blob/da011a312a6c6cd7ff12fe1aa0de1e33fba2f167/ext/mysqli/mysqli_nonapi.c#L102-L103)) and works as such but it's not documented anywhere, so let's leave it as is.
```php
<?php

$conn   = mysqli_connect('127.0.0.1', 'root', null, null, null, null);
$query  = 'SELECT VERSION()';
$result = mysqli_query($conn, $query);

echo mysqli_fetch_row($result)[0], PHP_EOL;
```
```
$ php -v
PHP 8.0.3 (cli) (built: Mar  6 2021 10:06:45) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.3, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.3, Copyright (c), by Zend Technologies
    with Xdebug v3.0.3, Copyright (c) 2002-2021, by Derick Rethans

$ php mysqli.php
8.0.23
```


muglug:
Thanks!

PR #5473, opened 2021-03-24
Checks the intersection type if the magic method does not exist

jni-:
The original issue we had was with the type `S3Client & MockObject`. Since S3Client has magic methods, calling MockObject methods on it (->expects, ->method, etc) was failing. Using `MockObject & S3Client` would work, because the method is detected on MockObject before checking S3Client. However, an intersection type should still check both, and the order should not matter (I think). 

I think the issue is that when a magic method was detected as missing, it would stop analysis and not validate union or intersection types. The tests I added replicate the problem well, however my confidence in how I fixed it is about 10%. I would need help to validate the solution, I did this in about 15 minutes and didn't understand all the implications. At least it's a starting point! 

Thanks!


muglug:
Thanks! Generally if it passes tests, I'm happy

PR #5477, opened 2021-03-25
Add ibm_db2 constant stubs

### LINKED ISSUE #5466: IBM_DB2 constants are not stubbed
The `ibm_db2` extension is stubbed but its constants aren't: https://psalm.dev/r/5c5902f4dd

I haven't found a way to add a constant to the existing extension stub. Is it currently supported?

morozov:
Closes #5466.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -1734,6 +1734,16 @@ class Config
             $core_generic_files[] = $stringable_path;
         }
 
+        if (\extension_loaded('ibm_db2')) {
+            $ext_ibm_db2_path = dirname(__DIR__, 2) . '/stubs/ext-ibm-db2.phpstub';
+
+            if (!file_exists($ext_ibm_db2_path)) {
+                throw new \UnexpectedValueException('Cannot locate ibm_db2 stub');
+            }
+
+            $core_generic_files[] = $ext_ibm_db2_path;
+        }
+
         if (\extension_loaded('PDO')) {
             $ext_pdo_path = dirname(__DIR__, 2) . '/stubs/pdo.phpstub';
 
--- a/stubs/ext-ibm-db2.phpstub
+++ b/stubs/ext-ibm-db2.phpstub
@@ -1,0 +1,22 @@
+<?php
+
+const DB2_BINARY = 1;
+const DB2_CONVERT = 2;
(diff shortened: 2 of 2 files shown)


src/Psalm/Config.php
```diff
@@ -1734,6 +1734,16 @@ public function visitStubFiles(Codebase $codebase, ?Progress $progress = null):
             $core_generic_files[] = $stringable_path;
         }
 
+        if (!\extension_loaded('ibm_db2')) {
```
morozov, line 1737: I don't really understand why the rest of the conditions look like `if (extension_loaded(/* ... */))`. As I understand, one of the purposes of the stubs is to enable the static analysis of the code that depends on an extension that is unavailable at runtime. Loading the stub only when the extension is loaded doesn't help to solve the issue.

muglug: > one of the purposes of the stubs is to enable the static analysis of the code that depends on an extension that is unavailable at runtime.

Oh, in that case you should instead provide your own stubs. If Psalm were to stub all classes it might miss situations where you're missing an extension.

morozov: I think I don't understand the architecture of stubs in Psalm. There are more or less complete stubs from JetBrains but they are released quite rarely and I want to try to get rid of them given that Psalm and PHPStan provide their own. Having to have 3 sources of stubs, each of which may have its issues is quite a burden.

weirdan: Psalm needs them to provide more detailed types that are unavailable through reflection. I think you can still force-load them in your `psalm.xml`, e.g.:
```
<stubs>
   <file name="vendor/vimeo/psalm/stubs/ext-ibm-db2.phpstub"/>
</stubs>
```

weirdan, line 1737:
```suggestion
        if (\extension_loaded('ibm_db2')) {
```

morozov: > Psalm needs them to provide more detailed types that are unavailable through reflection.

Right, but Psalm also uses the call maps that provide some basic information on the extension API. As a result, it can analyze most of the code that uses the extension (e.g. calls the functions) but not all of it (the usage of constants). From the end-user standpoint, this doesn't look right.

Issue #5465, opened 2021-03-24
OCI-Lob is not stubbed


morozov:
The `oci8` extension is stubbed but specifically the `OCI-Lob` class (or its userland version `OCI_Lob`) isn't: https://psalm.dev/r/0bbdd2e663

morozov:
UPD: actually it [is stubbed](https://github.com/vimeo/psalm/blob/c21fff3a6f0a4c6682e286388ea1fcd606a68ae7/dictionaries/CallMap.php#L9163) but somehow Psalm doesn't recognize the stub.

PR #5478, opened 2021-03-25
Fix oci8 stubs for PHP 8

morozov:
As of https://github.com/php/php-src/commit/74ca2a39a7a5fb403e3f6bf1e3ac77a152114fd5, the `OCI-Lob` and `OCI-Collection` classes are named as `OCILob` and `OCICollection` respectively.

This is related to #5465 but fixes the issue only on PHP 8 and only when the extension is loaded. If I add the difference between PHP 8 and PHP 7.4 to the stubs and run Psalm with `php-version=7.3`, it will start failing on PHP 8. Fixing this issue is out of the scope of this PR.

It looks like apart from stubbing method signatures, Psalm should stub the `OCI-`/`OCI` classes if the runtime version is different from the analyzer version:
```php
if (PHP_VERSION_ID >= 80000 && $codebase->php_major_version < 8) {
    class OCI-Lob { /* ... */}
}

if (PHP_VERSION_ID < 80000 && $codebase->php_major_version >= 8) {
    class OCILob { /* ... */}
}
```


muglug:
Thanks! Please also update the callmap delta! https://github.com/vimeo/psalm/blob/2b1bced09f0dd1a3e72cae86209a3d1353fcf46a/dictionaries/CallMap_80_delta.php

morozov:
@muglug please see the description:

> If I add the difference between PHP 8 and PHP 7.4 to the stubs and run Psalm with `php-version=7.3`, it will start failing on PHP 8.

muglug:
Oh OK sorry. I hate that bloody hyphen.

PR #5488, opened 2021-03-26
Add `@ignore-var` and `@psalm-ignore-var`

sj-i:
This annotation is used to ignore the `@var` annotation written in the same docblock.

## Background
To take advantage of the IDE's auto-completion, you may sometimes want to use explicit `@var` annotations even when psalm can infer the type well enough. This weakens the effectiveness of type checking in many cases since the explicit `@var` annotation overrides the types inferred by psalm.

See this example for more detail.
https://psalm.dev/r/c19d749dbe

## What is changed if this PR is merged
Psalm ignores the `@var` annotation which is co-located with `@psalm-ignore-var`. Then IDEs that don't fully understand complex types like generics can use the manually specified types with `@var` for auto-completion, while psalm can still use its own inferred types for type checking.

## Disclaimer
I'm not confident about my English skills. The sentences may seem odd to you. I know you aren't my English teacher, but if any wordings of the added doc or this PR sounds unnatural, please feel free to correct them. Then I can train myself better.

### DIFF
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -74,7 +74,7 @@ When specifying types in a format not supported by phpDocumentor ([but supported
 
 ### `@psalm-ignore-var`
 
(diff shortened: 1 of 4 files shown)


weirdan:
I would also suggest to file an issue with JetBrains (https://youtrack.jetbrains.com/issues/WI) to implement support for `@phpstorm-` vendor prefix.

muglug:
Thanks, this is great!

cc @ondrejmirtes @TysonAndre

PR #5495, opened 2021-03-28
disallow pure functions to have void return type

azjezz:
as discussed in #5463

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
@@ -188,6 +188,21 @@ class ReturnTypeAnalyzer
             return null;
         }
 
+        if ($return_type && $return_type->isVoid() && $function_like_storage && $function_like_storage->pure) {
+            if (IssueBuffer::accepts(
+                new InvalidReturnType(
+                    $cased_method_id
+                        . ' is expected to return a value, or never return but it has a return type of void.',
+                    $return_type_location
+                ),
+                $suppressed_issues
+            )) {
+                return false;
+            }
+
+            return null;
+        }
+
         if ($return_type
             && $return_type->isNever()
             && !$inferred_yield_types
--- a/tests/ReturnTypeTest.php
+++ b/tests/ReturnTypeTest.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Psalm\Tests;
 
 use const DIRECTORY_SEPARATOR;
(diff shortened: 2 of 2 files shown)


azjezz:
huh, turns out i am shooting my own foot here.

pure + void: https://github.com/azjezz/psl/blob/1.6.x/src/Psl/invariant.php

to be more specific, `invariant` has the return type of `void|no-return`, so there's still a change that it doesn't return.

what we want here is that pure functions with the return type of only "void" should be forbidden.

@muglug WDYT is the solution here?

azjezz:
also, i can't annotate `invariant` as void|no-return ;_;

bdsl:
`void|noreturn` is the same type as `void`. `noreturn` is the bottom type, so making a union type with `noreturn` is like making an addition sum with zero - it doesn't do anything.

azjezz:
yea, but i think there should be a way to annotate it as "possible won't return"

azjezz:
> void|noreturn is the same type as void. noreturn is the bottom type, so making a union type with noreturn is like making an addition sum with zero - it doesn't do anything.

yes, but actually we need to distinguish `void|noreturn` from `void` in this case.

you can have a pure function where the return type is `void` but can be `noreturn` sometimes ( assertion functions, see `Psl\invariant` for example ).

but you can't have a pure function with the return type of `void` ( that would never exist, throw, or run forever ) as that doesn't make any sense.

@Ocramius any thoughts on this?

Ocramius:
### `void` has this weird side-effect of either:

 * returning
    * by performing some impure operation
    * by performing some pure operation
 * throwing
 * exiting
 * running forever

### `noreturn` has the effect of either:

 * throwing
 * exiting
 * running forever

I don't think that `void|noreturn` is what you are looking for though, since a union of the two types just produces `void`. Perhaps what you want is something like `void & !noreturn` never being `@psalm-pure`?

bdsl:
> return type is void but can be noreturn sometimes 

That's like saying the return type is `\DateTimeInterface` but it can return `\DateTimeImmutable` sometimes. The latter (in both cases) is a subtype of the former, say saying it can return void already implies that maybe it can return `noreturn`.

azjezz:
> I don't think that void|noreturn is what you are looking for though, since a union of the two types just produces void. Perhaps what you want is something like void & !noreturn never being @psalm-pure?

yes, hence why i think psalm needs to keep track of whether a function might never return or not.

for example, this code should be invalid:

```php
/**
 * @pure
 */
function my_function(int $a): void
{
  $a++;
}
```

because it's basically dead code, and if you call if from anywhere, the call can be removed safely.

however, this is okay:

```php
/**
 * @pure
 */
function my_other_function(int $a): void
{
  if($a < 3) { throw new \Error('what!!!'); }
}
```

the difference between `my_other_function` and `my_function` is that `my_other_function` might never return ( can return `no-return` at some point ).

we need to forbid declaring pure functions that will never result in an exception, inf loop, or exit.

Ocramius:
I think what you want is that:

 * void + pure = unused
 * void + pure + @throws = ok
 * void + impure = ok

Not sure if there is a mechanism for inferring ***@***.***`.

azjezz:
yea, i think that's the way to go 👍

bdsl:
> keep track of whether a function might never return or not

I think working out whether a function might never return or not is the  [halting problem](https://en.wikipedia.org/wiki/Halting_problem) - so impossible for Psalm to track in a perfect way. But there might still be a useful heuristic that could be found.

Issue #5491, opened 2021-03-27
`class-string|trait-string` does not accept `class-string` 


ngmy:
HI,

In the example below, `class-string|trait-string` does not accept `class-string`:
https://psalm.dev/r/bc07216add

Thank,

weirdan:
Interpreted as `trait-string` only: https://psalm.dev/r/a7deef5790

ngmy:
Thanks, it worked fine with version 4.7.0.

PR #5497, opened 2021-03-29
Do not let string subtypes subsume class-strings

weirdan:
Previously, Psalm would treat unions like `class-string|numeric-string` as `numeric-string`, while the only case when string should subsume`class-string` is when we're combining `class-string` with non-specific `string`.

Fixes vimeo/psalm#5491


muglug:
Thanks!

Issue #5481, opened 2021-03-25
Can't use instanceof after check for class existance with class_exists


ChristophWurst:
I found out this code works just fine https://psalm.dev/r/42d41f391a

But this one doesn't https://psalm.dev/r/05befa391a

As far as I can tell there is no reason to not allow the second one :)

weirdan:
This also fails: https://psalm.dev/r/de623727b9

instanceof analysis will need to consider phantom classes.

PR #5498, opened 2021-03-29
Allow phantom classes in instanceof expression

weirdan:
Fixes vimeo/psalm#5481


weirdan:
Failure seems to be caused by 2d197e3e1265d9f5aa75306bec59cea6d3091043

muglug:
Thanks!

Issue #5472, opened 2021-03-24
Forbid `@template` annotation on closures


weirdan:
> Hack doesn't support templated closures, and I don't think there's much value in Psalm supporting them either. Opens up a whole can of worms.

_Originally posted by @muglug in https://github.com/vimeo/psalm/issues/5471#issuecomment-805830527_

It's currently accepted, but does not produce a useful generic: https://psalm.dev/r/93bb04856b. It would be better to reject it outright.

weirdan:
Expected: `InvalidDocblock: Templated closures are not supported`

Issue #3963, opened 2020-08-08
support generic variables


azjezz:
see: https://psalm.dev/r/5f20a1ae63

The idea is to define a constant for each function within PSL which would allow using that constant as a callable ( invoking the function with the same name as the constant ).

use case ( using PSL ):

before:
```php
/** @var iterable<Collection\VectorInterface> $vectors */
$vectors = getListOfVectors();
$arrays = Iter\map($vectors, fn(Collection\VectorInterface $vector): array => Iter\to_array($vector));
```

after:
```php
/** @var iterable<Collection\VectorInterface> $vectors */
$vectors = getListOfVectors();
$arrays = Iter\map($vectors, Iter\to_array);
```

azjezz:
working example ( https://psalm.dev/r/7fa8479faa ), this example works as `Str\concat` doesn't contain any generic parameters.

azjezz:
closing as https://github.com/php/php-src/pull/6898 fixes this.

```php
/** @var iterable<Collection\VectorInterface> $vectors */
$vectors = getListOfVectors();
$arrays = Iter\map($vectors, Iter\to_array(?));
```

PR #5499, opened 2021-03-29
Forbid template annotation on closures

weirdan:
They don't work properly anyway.

Fixes vimeo/psalm#5472

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php
@@ -455,6 +455,18 @@ class FunctionLikeNodeScanner
                     }
                 }
 
+                if ($stmt instanceof PhpParser\Node\Expr\Closure
+                    || $stmt instanceof PhpParser\Node\Expr\ArrowFunction
+                ) {
+                    if ($docblock_info->templates !== []) {
+                        $docblock_info->templates = [];
+                        $storage->docblock_issues[] = new InvalidDocblock(
+                            'Templated closures are not supported',
+                            new CodeLocation($this->file_scanner, $stmt, null, true)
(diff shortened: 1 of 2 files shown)


weirdan:
Apparently people use them.

/cc: @azjezz

azjezz:
I'm pretty sure hack does support generics in closure ( () => {} syntax, not php closure syntax since that is "deprecated", hence why generics weren't added there ).

azjezz:
so technically, the code that is currently failing in PSL, would pass in Hack if i remove all parameter types, as the type checker would get that it is generic ( even tho not explicitly ), but this would result in an error in psalm.

i personally like having support for generic in closures ( even variables/constants, see https://github.com/vimeo/psalm/issues/3963 )

weirdan:
How do you know it's actually generic and not just untyped?

azjezz:
It's not untyped, hack detects the type automatically depending on the usage of the closure.

e.g:
a

if we add `$c(vec['a', 'b', 'c'], 'c');` under `$c(vec[1, 2, 3], 3);`, it becomes:

b


and it can't be un-typed since hack doesn't allow thing to be un-typed ( at least at the strictest level )

---

so if psalm is to remove generics, it needs to add the ability to automatically detect type of callables without having to specify them based on usage.

azjezz:
hm, actually, looking at it, i think the failed case in psl can be fixed without using generics in closure, however, i still need to specify the type: https://psalm.dev/r/65b9344e48

here's a re-implementation with hack that passes: 

image

azjezz:
okay, all PSL issues have been fixed in https://github.com/azjezz/psl/pull/176, i think the "detect callable type based on usage" feature can be tracked in another issue.

muglug:
@azjezz @weirdan I've updated my fork!

muglug:
Thanks!

PR #5509, opened 2021-03-29
Double directory separator on "Cannot resolve stubfile path"

martingold:
Error message says
```
Problem parsing /Users/m/projects/project/psalm.xml:
  Cannot resolve stubfile path /Users/m/projects/project//stubs/ActiveDataProvider.phpstub
```
Notice the double slash at the second line. The message is correct, the file does not exist but in this case it is caused by different file extension not by psalm using incorrect path.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -961,7 +961,7 @@ class Config
 
                 if (!$file_path) {
                     throw new Exception\ConfigException(
-                        'Cannot resolve stubfile path ' . $config->base_dir . DIRECTORY_SEPARATOR . $stub_file['name']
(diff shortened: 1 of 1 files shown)


AndrolGenhald:
This is going to cause an incorrect message if the path doesn't start with `/`. If the leading `/` on stub paths is ignored then `DIRECTORY_SEPARATOR` would still need to be added if the stub path doesn't start with `/`. Maybe `ltrim` the stub path instead?

martingold:
I checked usage of `$this->base_dir` and i missed it is a public property so it can contain any string.

AndrolGenhald:
Ah, it was the base dir causing problems, that makes more sense.

muglug:
Thanks!

Issue #5508, opened 2021-03-29
Variadic parameter handling 


drupol:
Hello,

This example show the issue: https://psalm.dev/r/6216fb4626

There should not be any `ERROR: DuplicateArrayKey - 25:61 - String keys are not supported in unpacked arrays`.

AndrolGenhald:
See #5500 #5503.
@weirdan Maybe I should have documented this somewhere?

weirdan:
Duplicate of #5500

weirdan:
@drupol you're third today to report this. I wonder what's going on :thinking:

AndrolGenhald:
@weirdan 4.7.0 was released with #5455 11 hours ago. I'd add documentation somewhere, but I'm not really sure if it would help, I think the new behavior is just unexpected because named arguments is still pretty new.

weirdan:
@AndrolGenhald https://github.com/vimeo/psalm/blob/master/docs/running_psalm/issues/DuplicateArrayKey.md is linked from the message, and it'll be updated automatically on psalm.dev, so it's best to document it there.

drupol:
> @drupol you're third today to report this. I wonder what's going on thinking

Sorry mate for this, and thanks for letting me know where I was wrong.

Issue #5503, opened 2021-03-29
Type of variable arguments should be `list`


dsech:
https://psalm.dev/r/107e02f378

The inferred type for `$others` is `array<array-key, int>`, which leads to a `DuplicateArrayKey` when unpacking.

I would say the inferred type for variable arguments should always be `list<>`

weirdan:
It'll have string keys when called with named arguments: https://3v4l.org/SfPvM

You can declare it should not be called with named arguments, and Psalm will infer `list` for `$others` then: https://psalm.dev/r/fc37910e85

Issue #5500, opened 2021-03-29
False MIxedArgumentTypeCoercion for splat arguments


KartaviK:
Argument  1 of SomeClass::method expects array<int, \<Foo\>, parent type array<array-key, Foo> provided

When method is

```php
class SomeClass
{
    public function method(Foo ...$args)
    {
         // code
    }
}
```

$args will always contain array<int, Foo>|list<Foo>

if i try override type with `@param list<Foo> $args` it will work incorrectly

Etherlord:
https://psalm.dev/r/43fe7b3458

https://3v4l.org/bkkbh

weirdan:
It's `array<string, string>` when called with named arguments: https://3v4l.org/AOO6O
Add `@no-named-arguments` if you do not intend to support calling that function with named arguments: https://psalm.dev/r/e1f6a88779

Etherlord:
thanks!

KartaviK:
But i have php version 7.4 that does not have named arguments. Any ideas how psalm can detect it?

weirdan:
@KartaviK add `@no-named-arguments` annotation - it's a future-proof option you can use already. After all, one day you'll probably upgrade.

PR #5511, opened 2021-03-29
Clean up issues list, add documentation to help with named variadic confusion.

AndrolGenhald:
3 issues have been reported (#5500, #5503, #5508) since #5455 was released less than 12 hours ago, it seems there is a great deal of confusion with the new handling of variadic arguments. This is likely because named arguments are still new to most people, and even if they're aware of named arguments they may not consider the interaction with variadics.

This should improve the documentation where issues will pop up due to a missing `@no-named-arguments` annotation.
I also cleaned up issues list to match the issues that actually exist.


docs/running_psalm/issues.md
```diff
@@ -5,8 +5,6 @@
  - [ArgumentTypeCoercion](issues/ArgumentTypeCoercion.md)
  - [AssignmentToVoid](issues/AssignmentToVoid.md)
  - [CircularReference](issues/CircularReference.md)
- - [ComplexFunction](issues/ComplexFunction.md)
- - [ComplexMethod](issues/ComplexMethod.md)
```
weirdan, line 9: Both seem to be thrown.

weirdan (changes requested):

docs/running_psalm/issues.md
AndrolGenhald: Well...perhaps we need to add documentation for those.

muglug:
Thanks!

drupol:
Thanks !

PR #5512, opened 2021-03-29
Add the environment variable for using Shepherd

ngmy:
I want to be able to specify whether or not to use Shepherd by the environment variable.

I use a Composer script to run lint in both my local and CI environments (Github Actions):
```json
"scripts":
    "lint": [
        "@phpcs",
        "@psalm"
    ],
    "phpcs": "@php tools/phpcs ...",
    "psalm": "@php tools/psalm ...",
}
```

I would be happy if it could be specified by the environment variable, so that I don't have to write scripts that only differ in the presence or absence of the `--shepherd` option:
```yaml
- name: Run lint
  env:
      PSALM_SHEPHERD: 1
      run: composer lint
```
or
```yaml
- name: Run lint
  env:
      PSALM_SHEPHERD: 1
      PSALM_SHEPHERD_HOST: host
      run: composer lint
```

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -37,6 +37,7 @@ use function fwrite;
 use function gc_collect_cycles;
 use function gc_disable;
 use function getcwd;
+use function getenv;
 use function getopt;
 use function implode;
 use function in_array;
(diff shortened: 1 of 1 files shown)


weirdan (approved):

weirdan:
Thanks!

Issue #5489, opened 2021-03-26
Invalid RedundantPropertyInitializationCheck for static singleton properties 


danog:
An uninitialized static property cannot be initialized in an instance constructor: https://psalm.dev/r/05e21b46c0

muglug:
I've reverted the initial fix — instead, if it's possible for the property to not be initialised, make it nullable: https://psalm.dev/r/b5888c4b99

I realised the engine treats each those cases differently, but I prefer Psalm to advocate for that safer state of affairs.

This stops Psalm from

muglug:
Sorry, confusingly I'm reapplying @weirdan's change, but have added a second commit that changes the previous error message to one more specific to static properties.

muglug:
Old:

> Property type Singleton::$instance with type Singleton should already be set in the constructor

New:

> Static property type Singleton::$instance with type Singleton has unexpected isset check — should it be nullable?

PR #5525, opened 2021-03-30
Do not consider isset checks on static properties as redundant

weirdan:
Unlike normal properties, static properties do not have a prescribed initialization sequence, so they can always be uninitialized (or unset). Thus `isset()` checks on them are never redundant.

Fixes vimeo/psalm#5489


muglug:
Thanks!

PR #5532, opened 2021-03-30
Marked iterator_to_array as an impure function

vudaltsov:
`iterator_to_array` can be used to traverse a generator with side effects, for instance.

### DIFF
--- a/src/Psalm/Internal/Codebase/Functions.php
+++ b/src/Psalm/Internal/Codebase/Functions.php
@@ -472,7 +472,7 @@ class Functions
             'ldap_set_option',
 
             // iterators
-            'rewind', 'iterator_apply',
+            'rewind', 'iterator_apply', 'iterator_to_array',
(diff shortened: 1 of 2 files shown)


muglug:
Mind fixing the test? It's just a matter of removing `@psalm-pure` from that method

vudaltsov:
@muglug , done!

vudaltsov:
Squashed, rebased, green!

muglug:
Thanks!

someniatko:
I am not sure it's a correct decision to just mark this function as impure. For instance, `foreach` can also be used to traverse a generator with side effects. Should we mark each function using `foreach` on `iterable` an impure one? A better solution could be to separate "pure iterables" and "impure iterables". As temporary measure, until and if a better solution is implemented, i suggest reverting this change.

@vudaltsov @muglug What do you think?

PR #5534, opened 2021-03-30
fix: declare that php 7.3 is minimum level required

mr-feek:
I upgraded to psalm 4 and realized that php 7.3 is now required (per the release notes https://github.com/vimeo/psalm/releases/tag/4.0.0)

> It bumps the minimum requirement to PHP 7.3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^7.1|^8",
+        "php": "^7.3|^8",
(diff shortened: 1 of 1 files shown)


mr-feek:
After digging in more, I realized the php warnings I was getting were due to https://github.com/vimeo/psalm/issues/5259#issuecomment-810593602 , not anything to do with the runtime php version. I guess I'll go ahead and close this

PR #5543, opened 2021-03-31
Marked flock as an impure function

### LINKED ISSUE #3238: Allow flock "LOCK_UN" in UnusedFunctionCall
even though flock may return bool, flock with "LOCK_UN" as 2nd arg always returns true

This means psalm shouldn't give an "UnusedFunctionCall" error in that case

vudaltsov:
Closes #3238


muglug:
Thanks!

PR #5544, opened 2021-03-31
ConcatAnalyzer improvements and non-falsy-string fixes.

### LINKED ISSUE #5483: assert does not constrain to string union to evaluate accessed value type under array key
https://psalm.dev/r/9805391eca

asserting $var to be of value `'x'` does not help psalm to understand that `$a['a' . $var]` accesses `$a['ax']`. Although, psalm knows `$var` is of type `'x'` after assertion.

So for `$a = array{ax: int, ay: string}`, psalm does not know `$a['a' . $var]` is of type `int` (thinks `int|string` instead).

AndrolGenhald:
Deduplicate code.
Improve type inference.
Allow literal type inference when only one side has multiple types (fixes #5483).
Fix invalid type inference with negative int as right operand.

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClosureAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClosureAnalyzer.php
@@ -51,10 +51,12 @@ class ClosureAnalyzer extends FunctionLikeAnalyzer
      */
     public function getClosureId(): string
     {
-        return strtolower($this->getFilePath())
-            . ':' . $this->function->getLine()
-            . ':' . (int)$this->function->getAttribute('startFilePos')
-            . ':-:closure';
+        return strtolower(
+            $this->getFilePath()
+                . ':' . $this->function->getLine()
+                . ':' . (int)$this->function->getAttribute('startFilePos')
+                . ':-:closure'
+        );
     }
 
     /**
--- a/src/Psalm/Internal/Analyzer/ClosureAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClosureAnalyzer.php
@@ -51,12 +51,10 @@ class ClosureAnalyzer extends FunctionLikeAnalyzer
      */
     public function getClosureId(): string
     {
-        return strtolower(
-            $this->getFilePath()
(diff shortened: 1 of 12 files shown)


AndrolGenhald:
I'm tempted to offload most of the analysis here to CastAnalyzer, and since PHP (probably?) handles concatenation by casting everything to string first `InvalidCast` might be more applicable than `InvalidOperand` anyway. The only downside I see is that users would have to update their suppressions.

@muglug @weirdan thoughts?

AndrolGenhald:
Heads up @orklah, since you added the int concatenation to numeric string inference you may want to be aware that this now requires the right operand to be a non-negative int.

src/Psalm/Internal/Analyzer/ClosureAnalyzer.php
```diff
@@ -51,10 +51,12 @@ public function getTemplateTypeMap(): ?array
...
-        return strtolower($this->getFilePath())
-            . ':' . $this->function->getLine()
-            . ':' . (int)$this->function->getAttribute('startFilePos')
-            . ':-:closure';
+        return strtolower(
+            $this->getFilePath()
+                . ':' . $this->function->getLine()
+                . ':' . (int)$this->function->getAttribute('startFilePos')
+                . ':-:closure'
+        );
```
AndrolGenhald, line 59: Just realized this should be fixed instead of changing here. Concatenating int should still result in lowercase string.

orklah:
> 
> 
> Heads up @orklah, since you added the int concatenation to numeric string inference you may want to be aware that this now requires the right operand to be a non-negative int.

oh nice, didn't thought of that one at the time

AndrolGenhald:
I stumbled on some issues with non-falsy-string while fixing newly detected issues in the codebase, so I fixed those as well.

muglug:
Thanks! This is a great addition!

I wonder whether it might be better in the long run to make, "empty", "falsy" and "lowercase" attributes of a string type, with rules about which attributes can intersect. That might make the logic more complex though, not less.

PR #5557, opened 2021-04-01
Provide more detailed return type for get_object_vars()

weirdan:
Handle keyed objects and named objects, including support for property visibility.


muglug:
Thanks, awesome!

Issue #5299, opened 2021-03-01
Modifier to disallow non-documented keys from array shape


ntzm:
I would like to be able to do the following:

https://psalm.dev/r/cdf24c2204

The current behaviour is obviously wanted in most cases, but I would like a modifier that disallows non-documented keys from array shapes, something like `array-shape-strict{foo: string}` or `array!{foo: string}`

muglug:
This was slightly poor design on my part.

We could add a config that would make `array{a: int, b: string}` strict and `array{a: int, b: string}&array` the flexible version. In v5 that config could become the default.

ntzm:
That would work for me

AndrolGenhald:
I've been thinking the best way to do this would be to do something like TypeScript's [index signatures](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures) (and maybe eventually [mapped types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html)) for keyed arrays. Unsealed arrays (I'm say "sealed" and "unsealed" because that's what Psalm [currently calls it](https://github.com/vimeo/psalm/blob/c90cffd382bb7a5d390387d4ad2a02a4935fba5c/src/Psalm/Type/Atomic/TKeyedArray.php#L46). Sealed means no extra keys, unsealed means extra keys are `mixed`.) are currently the default, but in Psalm 5 (or probably 6 at this point) we could have a breaking change and say unsealed arrays must be represented as `array{a: int, [array-key]: mixed}`. We could also think about providing a tool to ease the transition by automatically changing `array{a: int}` to `array{a: int, [array-key]: mixed}`.

@ondrejmirtes I know we want to make sure Psalm and PHPStan are in agreement on this since it's such an important feature. I saw you had some proposals [here](https://github.com/vimeo/psalm/issues/4700#issuecomment-870369305), but I'm not much of a fan of `array{foo: Bar, []}`. I think if we want unsealed arrays to remain the default, a better option would be either `strict-array{foo: Bar}` or `array{foo: Bar, [array-key]: never}`, and I'm leaning towards the second option. Maybe we could have a config option for whether arrays are sealed or unsealed by default? That way `array{foo: Bar}` would use that config option, `array{foo: Bar, [array-key]: mixed}` would always be unsealed regardless of config, and `array{foo: Bar, [array-key]: never}` would always be sealed regardless of config.

If we want to go with index types, there are couple of things to consider.

[...]

Issue #3222, opened 2020-04-25
Feature request: `uppercase-string`


zerkms:
There is a `lowercase-string`, how about adding a symmetrical `uppercase-string`?

If the implementation is just to add thing similar to `lowercase-string` I may try to contribute it.

muglug:
Oh uh ok so it looks like people are using `lowercase-string` in the wild. I'm still unsure whether it should be part of Psalm's type system proper – it has weird interactions with `non-empty-string`

jsiefer:
Hi @muglug 
I was wondering if you could elaborate your thoughts on this? I started using it and could not see potential issues.

PHPStorm is holding back there ticked for supporting it (https://youtrack.jetbrains.com/issue/WI-56043)

PR #5560, opened 2021-04-02
Proposal: Type Constraints

AndrolGenhald:
This is a **very early** rough draft to get feedback on the idea, half of it isn't even implemented.

I've been thinking about trying to generalize some of psalm's types. Instead of having TPositiveInt we could have an int that has arbitrary min and max, written as `int(min=1)` (equivalent to `positive-int`) or `int(min=10, max=20)`.

This could also be useful to simplify and allow more specific string types. Leaving out class/trait and literal strings, we have:

 - `numeric-string`
 - `callable-string`
 - `non-empty-string`
 - `non-falsy-string`
 - `lowercase-string`
 - `non-empty-lowercase-string`
 - `html-escaped-string`

But what about `non-falsy-lowercase-string`, `lowercase-callable-string`, etc?
This could be implemented as `string(non-falsy, lowercase)` and `string(lowercase, callable)` instead.
Some of these might still be easier to have a separate `T...String` class for, but they could still have the constraints since they'd inherit from `TString`.

We could also have `string(minlen=5, maxlen=10)` instead of `non-empty-string`, and `array(minlen=5, maxlen=10)` instead of `non-empty-array` (and list).

Interactions need to be accounted for of course, `int(min=100, max=200) / int(min=5, max=10)` results in `int(min=10, max=40)`, and `string(callable, lowercase) . string(numeric)` results in `string(lowercase)` (numeric string should probably be a separate class that is always lowercase).

I'm not really set on calling this feature "Type Constraints", since Unions in general can be considered type constraints, but I haven't thought of anything better yet. (Edit: maybe Type Attributes?)

### DIFF
--- a/src/Psalm/Exception/InvalidConstraintException.php
+++ b/src/Psalm/Exception/InvalidConstraintException.php
@@ -1,0 +1,6 @@
+<?php
+namespace Psalm\Exception;
+
+class InvalidConstraintException extends \Exception
+{
+}
--- a/src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
@@ -53,6 +53,20 @@ class ScalarTypeComparator
         bool $allow_float_int_equality = true,
         ?TypeComparisonResult $atomic_comparison_result = null
     ) : bool {
+        // I'm thinking about replacing TypeComparators with Atomic::isSupertypeOf and
+        // Atomic::isSubtypeOf, but that can be done separately from type constraints.
+        if ($container_type_part instanceof TInt
+            && $input_type_part instanceof TInt
+        ) {
+            return $container_type_part->isSupertypeOf(
+                $input_type_part,
+                $codebase,
+                $allow_interface_equality,
+                $allow_float_int_equality,
+                $atomic_comparison_result
+            );
+        }
+
         if ($container_type_part instanceof TNonEmptyString
             && get_class($input_type_part) === TString::class
         ) {
--- a/src/Psalm/Internal/Type/ParseTree/ConstrainedTypeTree.php
+++ b/src/Psalm/Internal/Type/ParseTree/ConstrainedTypeTree.php
@@ -1,0 +1,21 @@
+<?php
+namespace Psalm\Internal\Type\ParseTree;
+
+/**
+ * @internal
+ */
+class ConstrainedTypeTree extends \Psalm\Internal\Type\ParseTree
+{
+    /**
+     * @var string
+     */
+    public $type;
+
+    public function __construct(
+        string $type,
+        ?\Psalm\Internal\Type\ParseTree $parent = null
+    ) {
+        $this->type = $type;
+        $this->parent = $parent;
+    }
+}
--- a/src/Psalm/Internal/Type/ParseTree/TypeConstraintTree.php
+++ b/src/Psalm/Internal/Type/ParseTree/TypeConstraintTree.php
@@ -1,0 +1,18 @@
+<?php
+namespace Psalm\Internal\Type\ParseTree;
+
+/**
+ * @internal
+ */
+class TypeConstraintTree extends \Psalm\Internal\Type\ParseTree
+{
+    /**
+     * @var string|null
+     */
+    public $name;
+
+    /**
+     * @var \Psalm\Internal\Type\ParseTree|null
+     */
+    public $value_tree;
+}
--- a/src/Psalm/Internal/Type/ParseTreeCreator.php
+++ b/src/Psalm/Internal/Type/ParseTreeCreator.php
@@ -309,7 +309,8 @@ class ParseTreeCreator
             $this->current_leaf = $this->current_leaf->parent;
         } while (!$this->current_leaf instanceof ParseTree\EncapsulationTree
             && !$this->current_leaf instanceof ParseTree\CallableTree
-            && !$this->current_leaf instanceof ParseTree\MethodTree);
+            && !$this->current_leaf instanceof ParseTree\MethodTree
+            && !$this->current_leaf instanceof ParseTree\ConstrainedTypeTree);
 
         if ($this->current_leaf instanceof ParseTree\EncapsulationTree
             || $this->current_leaf instanceof ParseTree\CallableTree
@@ -334,6 +335,7 @@ class ParseTreeCreator
             || $context_node instanceof ParseTree\KeyedArrayTree
             || $context_node instanceof ParseTree\CallableTree
             || $context_node instanceof ParseTree\MethodTree
+            || $context_node instanceof ParseTree\ConstrainedTypeTree
         ) {
             $context_node = $context_node->parent;
         }
@@ -343,6 +345,7 @@ class ParseTreeCreator
             && !$context_node instanceof ParseTree\KeyedArrayTree
             && !$context_node instanceof ParseTree\CallableTree
             && !$context_node instanceof ParseTree\MethodTree
+            && !$context_node instanceof ParseTree\ConstrainedTypeTree
         ) {
             $context_node = $context_node->parent;
         }
@@ -364,6 +367,16 @@ class ParseTreeCreator
         }
 
         $current_parent = $this->current_leaf->parent;
+
+        if ($current_parent instanceof ParseTree\TypeConstraintTree && $type_token[0] === '=') {
+            if (!$current_parent->value_tree instanceof ParseTree\Value) {
+                throw new TypeParseTreeException('Invalid constraint name');
+            }
+            $current_parent->name = $current_parent->value_tree->value;
+            $current_parent->value_tree = null;
+            $this->current_leaf = $current_parent;
+            return;
+        }
 
         if ($this->current_leaf instanceof ParseTree\MethodTree && $type_token[0] === '...') {
             $this->createMethodParam($type_token, $this->current_leaf);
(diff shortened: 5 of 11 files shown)


orklah:
I would really like having ranges for TInt. This would allow for powerful plugins (which could for example find expired date checks). I thought about that some times ago, but I was wary of introducing an ambiguous case:

For example 
```php
$a = ['0'=>'a', '1'=>'b'];
```
From most precise to less precise `$a` can be : 
- `array{"a", "b"}` / `array{0 => "a", 1=> "b"}` => TKeyedArray
- `non-empty-list<string>` => TNonEmptyList
- `non-empty-array<int, string>` => TArray

but now it could also be: 
- `non-empty-array<int(min=0, max=1), string>`

The thing is `non-empty-array<int(min=0, max=1), string>` is not more or less precise than `non-empty-list<string>` it just have a different meaning because the former does not imply list behaviour (the existence of offset '0' is not guaranteed) and the latter doesn't have range information (there would be no specific error emitted for attempting to check the existence of offset '2')

This means Psalm would have to make some sort of choice for cases like:
`$a = range(0,1);`
Is this a list or is this a constrained type? Each type would come with its shortcomings

AndrolGenhald:
> The thing is `non-empty-array<int(min=0, max=1), string>` is not more or less precise than `non-empty-list<string>` it just have a different meaning...

If I'm understanding you correctly, I think we already have cases like that.
From most to least specific (probably missing a couple):
```
      non-falsy-string
    /                  \
'1' - numeric-string --- non-empty-string - string - mixed
 |                   \                    /        /
  \                    numeric ---------[/]--------
    lowercase-string -------------------/

```

orklah:
Well, the interaction between non-empty-string and lowercase-string is already a weird part of Psalm (https://github.com/vimeo/psalm/issues/3222#issuecomment-619433723) and I think it's for the exact same reason I mentionned

I didn't look at your implementation though (and I'm not sure I have the skills for that) so things may be simpler than in my head 😃

AndrolGenhald:
> Well, the interaction between non-empty-string and lowercase-string is already a weird part of Psalm ([#3222 (comment)](https://github.com/vimeo/psalm/issues/3222#issuecomment-619433723)) and I think it's for the exact same reason I mentionned

I wasn't aware of that, I wonder if that got fixed up with the addition of `non-empty-lowercase-string`? Maybe that issue should be updated.

muglug:
Small nit: `html-escaped-string` will be removed in an upcoming version as it's basically not all that useful now Psalm has taint analysis

## Int ranges

I'd prefer to separate out int ranges into their own thing – PHPStan offers a more machine-readable syntax for int ranges that's still easy for humans to grasp e.g. `int<1,>`.

## String attributes

### Syntax

I'm not sure we need a new syntax to express string attributes. If we did want to express in docblocks this I'd rather we went with something more familiar like intersection types e.g. `string&non-empty&lowercase` — but I still prefer `non-empty-lowercase-string`, because it's just a bit more human readable (whatever its _internal_ representation might become).

### Data structures

Some constraints imply other constraints (or the negation of constraints), but it's not clear how a new datastructure could make those rules easier to implement/use fewer lines of code than are used currently.

For example on a string `callable` implies `non-falsy`, `non-falsy` implies `non-empty`. `callable` and `lowercase` are orthoganal (one implies nothing about the other), `callable` implies '!numeric`, `numeric` implies `!callable` but also `non-empty`.

We'd need to preserve/enshrine rules handling concatenation:

- Concatenating a `callable` string with another string removes `callable` but preserves its `non-falsy`-ness
- Concatenating a `numeric` string with another string removes `numeric` but preserves its `non-empty`-ness
- Concatenating a `non-empty` string with any other string preserves the `non-empty`-ness
- Concatenating a `non-falsy` string with any other string preserves its `non-falsy`ness

We'd also need rules for unioning the types together

If we had some sort of ruleset,  we could have a transition period (before a major version bump) where we shift to using methods on `TString`

`TString::isCallable()`, `TString::isFalsy()`, `TString::isEmpty()`, `TString::isNumeric()`. Those could be implemented on those existing subclasses. 

Literal strings also present a slight issue — current Psalm just treats any literal string as non-callable when unioning types together, because in general figuring out whether a string is callable requires knowing exactly where the string is being used.

AndrolGenhald:
### Ranges

> I'd prefer to separate out int ranges into their own thing – PHPStan offers a more machine-readable syntax for int ranges that's still easy for humans to grasp e.g. `int<1,>`.

Neat, I haven't looked into other static analyzers too much so I wasn't aware there was already a syntax for this. It would indeed be preferable to maintain compatibility wherever possible, unfortunately this causes problems for array size ranges, where `<>` is already used to specify templates. I'd like for array size ranges to be possible too, and it'd be less confusing if it used similar syntax to int ranges. I'll have to think more on this.

Maybe we could treat ranges as templates (at least conceptually, if not in the implementation)? I've noticed `array<string>` is treated as `array<array-key, string>` rather than `array<string, mixed>`, so if too few templates are specified I assume psalm fills them in from right to left. You mentioned `int<1,>` so interpreting that as a template we would want to support empty templates, so you could do `int<1,>` instead of `int<1,PHP_MAX_INT>` or `int<1,null>` (since `int<1>` would be int with max 1 instead of min 1). `array<1, 10,,>` would be `array<array-key, mixed>` with min length 1 and max length 10.

### Strings

> I'm not sure we need a new syntax to express string attributes. If we did want to express in docblocks this I'd rather we went with something more familiar like intersection types e.g. `string&non-empty&lowercase` — but I still prefer `non-empty-lowercase-string`, because it's just a bit more human readable (whatever its _internal_ representation might become).

`non-empty-lowercase-string` is the most readable of those, so I can see your point. I think we could actually do a combination of the two, `non-empty-lowercase-string` seems a bit niche, but `non-empty-string` is probably fairly common, and easier to read than `string&non-empty`. If we want a non-falsy-non-empty-lowercase-string we can do `non-empty-string&non-falsy-string&lowercase-string`.

### Data structures

> Some constraints imply other constraints (or the negation of constraints), but it's not clear how a new datastructure could make those rules easier to implement/use fewer lines of code than are used currently.

I think moving from `TypeComparator`s to `Atomic::isSubtypeOf` and `Union::isSubtypeOf` could allow some things to be simplified. I might try to do a separate pull request next week with some of this.

orklah:
> if too few templates are specified I assume psalm fills them in from right to left

That's not the case. The general rule is that you MUST have exactly the right number of templates.
array content is a special case where if one is omitted, we assume the template describe the content (so the second template)
another exception would be [Generators](https://psalm.dev/docs/annotating_code/type_syntax/object_types/#generators)
who can have up to 4 templates and shorthands notations makes Psalm fills the missing template in arbitrary order

AndrolGenhald:
> > if too few templates are specified I assume psalm fills them in from right to left
> 
> That's not the case. The general rule is that you MUST have exactly the right number of templates.
> array content is a special case where if one is omitted, we assume the template describe the content (so the second template)
> another exception would be [Generators](https://psalm.dev/docs/annotating_code/type_syntax/object_types/#generators) who can have up to 4 templates and shorthands notations makes Psalm fills the missing template in arbitrary order

I'll have to look into that a bit more then, are there any cases where changing it to always fill in from right to left would break things? I know that at the very least you can omit templates entirely and they default to either their `as` type or `mixed`.

orklah:
> are there any cases where changing it to always fill in from right to left would break things

Yep, see the doc for Generators:
-When only one template is present, it's the second one
-When two templates are present, it's the first two
So that means it first fill from the middle to the left and then start from the middle again to go right ><

To my knowledge, you can either not describe any template, in this case, every template will be `empty` (and not mixed) or you specify all of them (otherwise you get this error: https://psalm.dev/r/a651a9f7d0)

AndrolGenhald:
> EDIT: nope, the `empty` part is not true, at least not entirely. I saw `empty` templates before when they were not specified but not sure how to reproduce. In this example they're indeed mixed or the upper bound: https://psalm.dev/r/8f990bdec3

Empty array is the only place I've seen the `empty` type, but I haven't gone looking: https://psalm.dev/r/1f515f51ee

orklah:
> Empty array is the only place I've seen the `empty` type, but I haven't gone looking: https://psalm.dev/r/1f515f51ee

I probably encountered that when working with templates in Collections. For example: https://psalm.dev/r/142c2f6da3 so it's akin to empty arrays. It doesn't seem to apply to this discussion so please ignore what I said :)

AndrolGenhald:
One thing the template style doesn't solve is the issue of non-empty-object-like-array and object-like-array-with-additional-properties. I don't like `array{foo?: string}&array` since intersections like that seem like they should just simplify to `array{foo?: string}`, and `array{foo?: string}|array` doesn't work for the same reason, it's just `array`.

I've already proposed `non-empty-array{foo?: string, bar?: string}`, but for object-like-array-with-additional-properties maybe ellipses could be used? `array{foo?: string, bar?: string, ...}` would be intuitive, and `array{foo?: string, bar?: string, ...}&array<string>` isn't too complicated (additional properties are `string` instead of `mixed`).

One downside to this idea is `array{foo: int, bar: float, ...}&array<string>` doesn't work, since the types don't overlap, you'd need `&array<int|float|string>`, so there's no way to say that the additional properties must all be a specific type if some of the named properties have other types, but I'm not sure how important it is to support that.

orklah:
Makes me think of https://github.com/vimeo/psalm/issues/5299#issuecomment-788070132

AndrolGenhald:
Looking at it again, non-empty-object-like-array should be fine, it just needs fixed (#5561). `array{foo?: string}&non-empty-array` should be fine, and `array{foo: string}&non-empty-array` simplifies to `array{foo: string}` (due to existence of at least 1 required property). So it's just the additional properties thing that needs to be worked out.

orklah:
I kinda forgot this PR existed for a while. I implemented my own version of integer ranges in #6241. If you still want to work on constraints, you'll have to take that into consideration

orklah:
I'm gonna close this because integer ranges were implemented in another PR.

I'm also dubious about the underlying concept of adding constraint to a type, opposed to create a new type.

Adding integer range was a massive work (I probably spent 50+ hours on it) and it had impacts in a lot of modules. From assertions, to combination, intersection, reconciliation etc...

I'm not sure if having a set of optional constraint would play nicely in that. Being able to have a distinct condition where you handle a specific type and you know there is nothing else to handle there is really comfortable and I'm not sure I could have handled juggling with multiple constraints at a time.

PR #5564, opened 2021-04-02
Remove html-escaped-string, utterly unused

muglug:
This might be considered a breaking change, but I very much doubt anyone uses it.


AndrolGenhald:
I was curious why that existed when I noticed it, I assumed it must pre-date taint analysis, or maybe was part of an earlier prototype for taint analysis.

psalm-shepherd:
Psalm found errors in other files

```
examples/plugins/composer-based/echo-checker/EchoChecker.php:48:29 - Psalm\Type\Atomic\TString cannot be identical to Psalm\Type\Atomic\THtmlEscapedString
examples/plugins/composer-based/echo-checker/EchoChecker.php:48:46 - Class or interface Psalm\Type\Atomic\THtmlEscapedString does not exist
src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php:55:9 - Could not verify return type 'bool' for Psalm\Internal\Type\Comparator\ScalarTypeComparator::isContainedBy
src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php:431:48 - Class or interface Psalm\Type\Atomic\THtmlEscapedString does not exist
```

muglug:
Will remove in 5.x

PR #5577, opened 2021-04-05
The return value of generator methods can potentially be null

danog:
The return value of generator methods can potentially be null, if the generator is no longer valid.

### DIFF
--- a/stubs/CoreGenericClasses.phpstub
+++ b/stubs/CoreGenericClasses.phpstub
@@ -20,7 +20,7 @@ interface Traversable {
  */
 class Generator implements Traversable {
     /**
-     * @return TValue Can return any type.
+     * @return ?TValue Can return any type.
      */
     public function current() {}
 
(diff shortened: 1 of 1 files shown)


stubs/CoreGenericClasses.phpstub
```diff
@@ -20,7 +20,7 @@ interface Traversable {
  */
 class Generator implements Traversable {
     /**
-     * @return TValue Can return any type.
+     * @return ?TValue Can return any type.
```
weirdan, line 23: I would probably need [`@psalm-ignore-nullable-return`](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-ignore-nullable-return) as well. Otherwise code like this will fail:
```php
/** @param Generator<int, int> */
function f(Generator $g): void {
   while ($generator->valid()) {
       takesInt($generator->current());
       $generator->next();
   }
}
```

muglug: Yup, good suggestion!

danog: Done! :)

muglug:
Thanks!

PR #5584, opened 2021-04-06
Add ReflectionProperty::getType() and hasType()

franmomu:
This PR tries to do as https://github.com/vimeo/psalm/commit/c97ee5ccdb75c85bc9f6285bfe2cb7a2fb1b50dc for [`ReflectionProperty`](https://www.php.net/manual/en/class.reflectionproperty.php) adding [`ReflectionProperty::hasType()`](https://www.php.net/manual/en/reflectionproperty.hastype.php) and [`ReflectionProperty::getType()`](https://www.php.net/manual/en/reflectionproperty.gettype.php).

Right now: https://psalm.dev/r/b00d652573

### DIFF
--- a/stubs/Reflection.phpstub
+++ b/stubs/Reflection.phpstub
@@ -76,6 +76,18 @@ class ReflectionProperty implements Reflector
      * @return array<ReflectionAttribute<TClass>>
      */
     public function getAttributes(?string $name = null, int $flags = 0): array {}
+
+    /**
+     * @since 7.4
+     * @psalm-assert-if-true ReflectionType $this->getType()
+     */
+    public function hasType() : bool {}
+
+    /**
+     * @since 7.4
+     * @psalm-mutation-free
+     */
+    public function getType() : ?ReflectionType {}
 }
 
 class ReflectionMethod implements Reflector
--- a/tests/AssertAnnotationTest.php
+++ b/tests/AssertAnnotationTest.php
@@ -1509,6 +1509,19 @@ class AssertAnnotationTest extends TestCase
                      */
                     function allIsInstanceOf($value, $class): void {}'
             ],
+            'implicitReflectionPropertyAssertion' => [
+                '<?php
+                    $class = new ReflectionClass(stdClass::class);
+                    $properties = $class->getProperties();
+                    foreach ($properties as $property) {
+                        if ($property->hasType()) {
+                            $property->getType()->allowsNull();
+                        }
+                    }',
+                    [],
+                    [],
+                    '7.4'
+            ],
         ];
     }
 
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -11564,6 +11564,7 @@ return [
 'ReflectionProperty::getName' => ['string'],
 'ReflectionProperty::getType' => ['?ReflectionType'],
 'ReflectionProperty::getValue' => ['mixed', 'object='=>'object'],
+'ReflectionProperty::hasType' => ['bool'],
 'ReflectionProperty::isDefault' => ['bool'],
 'ReflectionProperty::isPrivate' => ['bool'],
 'ReflectionProperty::isProtected' => ['bool'],


franmomu:
I forgot I started this one, so tests fail and the only thing I can think of is that `ReflectionParameter` is immutable, but `ReflectionProperty` is not. Any help? 🤔

caugner:
@franmomu Unfortunately `The logs for this run have expired and are no longer available.`. Would you be able to amend & force-push to re-trigger the checks?

stubs/Reflection.phpstub
```diff
@@ -76,6 +76,17 @@ class ReflectionProperty implements Reflector
...
+    /**
+     * @since 7.4
+     * @psalm-assert-if-true ReflectionType $this->getType()
+     */
+    public function hasType() : bool {}
+
+    /**
+     * @since 7.4
+     */
+    public function getType() : ?ReflectionType {}
```
caugner, line 89: I'm not sure what test fails, but maybe you need to be defensive here and only mention the new (?!) type in the PHPDoc:

```suggestion
     * @return ReflectionType|null
     */
    public function getType() {}
```

caugner: Never mind, just saw now that `ReflectionParameter` uses the same:

https://github.com/vimeo/psalm/blob/f381fba070c9a005359c21209ab85517a544e1e9/stubs/Reflection.phpstub#L112

franmomu:
> @franmomu Unfortunately `The logs for this run have expired and are no longer available.`. Would you be able to amend & force-push to re-trigger the checks?

Sure, done! Feel free to try it, the problem I think it was immutability, `ReflectionProperty` has a `setValue` method so it cannot be immutable, don't know if there is other way.

caugner:
> Sure, done!

Thanks.

```
There was 1 error:
  
1) Psalm\Tests\AssertAnnotationTest::testValidCode with data set "implicitReflectionPropertyAssertion" ('<?php\n                    $c...     }')
   Psalm\Exception\CodeException: MixedMethodCall - src/somefile.php:6:51 - Cannot determine the type of the object on the left hand side of this expression
```

And this is line 6:

```php
$property->getType()->allowsNull();
```

Alright, now I see what you mean. Since `$property` cannot be declared immutable, Psalm doesn't use `assert-if-true` over from `$property->hasType()` to `$property->getType()`. 😦

I wonder if it would help to mark `hasType()` and `getType()` with `@psalm-mutation-free`. But still, even if `$property` is *not* immutable, Psalm should be able to determine that the assertion by `hasType()` is true at least immediately for the next statement.

weirdan:
`MixedMethod` is caused by Psalm thinking you're targeting a PHP version below 7.4. You need to mark the test case as targeting 7.4.

caugner:
> You need to mark the test case as targeting 7.4.

Here's an example (I think):

https://github.com/vimeo/psalm/blob/9a0f9f9c87de437fa4bc7c6772bbea6ca1f28656/tests/Template/ClassTemplateExtendsTest.php#L3442-L3488

tests/AssertAnnotationTest.php
```diff
@@ -1509,6 +1509,16 @@ function foo(array $foos) : array {
...
+            'implicitReflectionPropertyAssertion' => [
+                '<?php
+                    $class = new ReflectionClass(stdClass::class);
+                    $properties = $class->getProperties();
+                    foreach ($properties as $property) {
+                        if ($property->hasType()) {
+                            $property->getType()->allowsNull();
+                        }
+                    }',
+            ],
```
caugner, line 1521: This could/should work:

```suggestion, 
                 [], 
                 [], 
                 '7.4',
            ],
```

franmomu:
> `MixedMethod` is caused by Psalm thinking you're targeting a PHP version below 7.4. You need to mark the test case as targeting 7.4.

> 
> Here's an example (I think):
> 
> https://github.com/vimeo/psalm/blob/9a0f9f9c87de437fa4bc7c6772bbea6ca1f28656/tests/Template/ClassTemplateExtendsTest.php#L3442-L3488

Thanks! done

stubs/Reflection.phpstub
```diff
@@ -76,6 +76,18 @@ class ReflectionProperty implements Reflector
...
+
+    /**
+     * @since 7.4
+     * @psalm-assert-if-true ReflectionType $this->getType()
+     */
+    public function hasType() : bool {}
+
+    /**
+     * @since 7.4
+     * @psalm-mutation-free
```
franmomu, line 88: @caugner I think it works with the `@psalm-mutation-free` you said 👍  (apart from adding the `7.4` check)

caugner: Just to check: Did you need to add mutation-free to make the test pass? So marking the test with 7.4 was not enough?

franmomu: Yes, without: https://psalm.dev/r/5d10995beb
with: https://psalm.dev/r/8e9b30898d

caugner: Thanks, glad to hear it works! This reminds me that I should use that annotation in my own projects as well. 🙂

weirdan (approved):
Thanks, looking good!

PR #5591, opened 2021-04-07
Improve type resolution of class constants.

### LINKED ISSUE #5588: Plus operator in class const not resolving type when referencing parent const
Referencing self consts seem to work fine: https://psalm.dev/r/70beb89706
But referencing parent consts causes the type to become `mixed`: https://psalm.dev/r/cbc8e21d41
Broken for ints too: https://psalm.dev/r/6b0c9ee8fc

Concatenation seems to work ok, though it changes the type to `string` instead of resolving the literal concatenation: https://psalm.dev/r/b67a63d851

AndrolGenhald:
Handle array plus operator (fixes #5588).
Handle array spread operator.
Improve string concatenation to resolve to literal instead of unknown string.

I feel like it should be possible to let ConstantTypeResolver take advantage of the better type analysis in ArrayAnalyzer, ConcatAnalyzer, etc, but it would require a restructuring that's beyond me for the time being.

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
@@ -46,6 +46,8 @@ use function preg_split;
 use const PREG_SPLIT_DELIM_CAPTURE;
 use const PREG_SPLIT_NO_EMPTY;
 use function array_shift;
+use function array_values;
+use function get_class;
 
 class ClassLikeNodeScanner
 {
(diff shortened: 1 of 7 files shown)


src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
```diff
@@ -1161,17 +1163,31 @@ private function visitClassConstDeclaration(
...
+
             if ($const_type) {
                 $existing_constants[$const->name->name] = new \Psalm\Storage\ClassConstantStorage(
                     $const_type,
-                    ClassLikeAnalyzer::VISIBILITY_PUBLIC,
+                    $stmt->isProtected()
+                        ? ClassLikeAnalyzer::VISIBILITY_PROTECTED
+                        : ($stmt->isPrivate()
+                            ? ClassLikeAnalyzer::VISIBILITY_PRIVATE
+                            : ClassLikeAnalyzer::VISIBILITY_PUBLIC),
```
AndrolGenhald, line 1182: Not quite sure if I'm fixing a mistake or introducing a bug here. I _think_ it fixes a mistake, since it looks like these are existing constants for the current class, but none of the tests are affected either way, so I'm not sure exactly what affect this has.

muglug: This is unnecessary — these constants are not actually used, but it's a harmless change so it's fine to keep.

muglug:
Thanks!

PR #5597, opened 2021-04-07
Check oldest ancestor for protected method visibility (fixes #5595)

### LINKED ISSUE #5595: InaccessibleMethod false positive when calling method on sibling
https://psalm.dev/r/dfdf721e08

Since `baz` is originally declared on `FooBar`, `Bar` can access `Foo::baz`.

AndrolGenhald:


src/Psalm/Storage/ClassLikeStorage.php
```diff
@@ -242,6 +242,10 @@ class ClassLikeStorage
     public $appearing_method_ids = [];
 
     /**
+     * Map from lowercase method name to list of declarations in order from parent, to grandparent, to
+     * great-grandparent, etc **including traits and interfaces**. Ancestors that don't have their own declaration are
+     * skipped.
+     *
```
AndrolGenhald, line 248: This seems to be accurate for the cases I looked at, but I haven't checked in depth to make sure.

muglug:
Thanks!

PR #5598, opened 2021-04-07
Numeric can be Falsy

### LINKED ISSUE #5583: False-possitive - Type numeric|true is never falsy
Hopefully I'm not missing something here but `is_numeric($value) || $value === true` doesn't exclude $value from being falsy.

https://psalm.dev/r/b9196c226c

orklah:
This should fix #5583


muglug:
Thanks!

PR #5601, opened 2021-04-08
Strict int/float comparison for conditional types, improve range return type

AndrolGenhald:
Fixes incorrect conditional return type due to int/float comparison.
`int|string` works as expected: https://psalm.dev/r/b87cd02f3a
but `int|float` produces very odd results: https://psalm.dev/r/993cd83eb3

Also improve range return type to be `list<float>` if `$step` is `float` (I didn't find any PHP documentation addressing this but it seems to be empirically true: https://3v4l.org/3tcIK).


muglug:
Thanks!

PR #5603, opened 2021-04-08
Fix race condition creating parser cache directory (fixes #4483)

### LINKED ISSUE #4483: PHP Warning
Occasionally psalm gives me this PHP warning:

```
PHP Warning:  mkdir(): File exists in /app/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ParserCacheProvider.php on line 244
```

AndrolGenhald:
I wasn't able to reproduce the error under normal conditions, but I set a breakpoint on the `mkdir` call, created the directory, then continued, and it seems to handle it correctly.

### DIFF
--- a/src/Psalm/Internal/Provider/ParserCacheProvider.php
+++ b/src/Psalm/Internal/Provider/ParserCacheProvider.php
@@ -23,8 +23,10 @@ use Psalm\Config;
 use function scandir;
 use function serialize;
 use function touch;
+use function trigger_error;
 use function unlink;
 use function unserialize;
+use const E_USER_ERROR;
 use const SCANDIR_SORT_NONE;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Provider/ParserCacheProvider.php
```diff
@@ -402,4 +400,18 @@ private function getParserCacheKey(string $file_name): string
     {
         return md5($file_name) . ($this->use_igbinary ? '-igbinary' : '') . '-r';
     }
+
+    private function createCacheDirectory(string $parser_cache_directory): void
+    {
+        if (!is_dir($parser_cache_directory)) {
```
weirdan, line 406: wouldn't `@mkdir(...)` make more sense? After all, we only want the directory to exist. On the other hand this could just move race-condition elsewhere.

AndrolGenhald: _Technically_ it could be a file for some reason, in which case an error should be thrown, but I wouldn't really be opposed to changing it to `@mkdir(...)`, as it being a file seems incredibly unlikely.

I suppose you're right that this could just move the race condition somewhere else, I hadn't considered that. I haven't really looked into how the caching works, hopefully different processes aren't going to be using the same `$file_cache_key`.

AndrolGenhald: `mkdir` could also fail due to lack of write permission, which also seems unlikely, but more likely than the directory path being a file.

muglug:
Thanks!

PR #5606, opened 2021-04-09
Allow variable keys with array_key_exists

### LINKED ISSUE #5605: array_key_exists with variable literal key does not modify shape of array
https://psalm.dev/r/a3b6b01fb9
As you can see, the shape of $params is not modified to `array{key: string}` inside of the if.

danog:
Fixes #5605 :)


muglug:
Thanks, awesome!

PR #5608, opened 2021-04-09
Allow to use string key for preg_replace_callback callable argument

paxal:
I encountered a wrong signature in `preg_replace_callback`.

The callable argument can have string keys when one uses named capture groups.

Consider the following : 
```php
<?php

echo preg_replace_callback(
    '@Hello, (?<WHO>.*)!@', 
    static function (array $matches): string {
        var_dump($matches); 
        return $matches['WHO']; 
    },
    'Hello, world!'
);
```
on 3v4l https://3v4l.org/7qehR

I tried to fix it by myself, I don't know if I've covered everything :innocent:


src/Psalm/ErrorBaseline.php
```diff
@@ -286,7 +286,7 @@ function (string $extension) : string {
         $xml = preg_replace_callback(
             '/<files (psalm-version="[^"]+") (?:php-version="(.+)"(\/?>)\n)/',
             /**
-             * @param array<int, string> $matches
+             * @param string[] $matches
```
orklah, line 289: I don't think you need to change this, `array<int, string>` should be accepted whenever `string[]` is expected. (`string[]` is an alias of `array<string>` itself an alias of `array<array-key, string>` which is a parent type of `array<int, string>`

tests/FunctionCallTest.php
```diff
@@ -1279,7 +1279,7 @@ function takesInt(int $i) : void {}
                     function foo(string $s) : string {
                         return preg_replace_callback(
                             \'/<files (psalm-version="[^"]+") (?:php-version="(.+)">\n)/\',
-                            /** @param array<int, string> $matches */
+                            /** @param string[] $matches */
```
orklah, line 1282: same here

src/Psalm/ErrorBaseline.php
weirdan: You do need to change this, because `callable(array<array-key, string>)` is a *subtype* of `callable(array<int, string>)` due to parameter contravariance.

caugner (approved):
LGTM

The failing test below is due to the fact that `muglug/psl`'s `replace_with()` function accepts a callable that only expects `array<int, string>` as a parameter, and passes this callable to `preg_replace_callback()`: https://github.com/muglug/psl/blob/da0bb02b7d5024366f4ad8f5498f91b978dca3d2/src/Psl/Regex/replace_with.php#L18

```
ERROR: MixedArgumentTypeCoercion - ../../src/Psl/Regex/replace_with.php:37:56 - Argument 2 of preg_replace_callback expects callable(array<array-key, string>):string, parent type callable(array<int, string>):string provided (see https://psalm.dev/194)
        static fn() => preg_replace_callback($pattern, $callback, $haystack, $limit),
  The type of $callback is sourced from here - ../../src/Psl/Regex/replace_with.php:26:67
function replace_with(string $haystack, string $pattern, callable $callback, ?int $limit = null): string
```

**edit**: I created an [upsream PR](https://github.com/azjezz/psl/pull/219) to fix that.

src/Psalm/ErrorBaseline.php
caugner: Afaik the change is correct, and equivalent to this:

```suggestion
             * @param array<array-key, string> $matches
```

Since the `$search` parameter already uses `string[]` though, I would strongly recommend to use `string[]` as well, to be consistent.

caugner:
@weirdan I think this PR is ready, or are you waiting for a response from orklah?

orklah:
I have nothing to add here :)

caugner (approved):
LGTM

weirdan (approved):

weirdan:
Thanks!

PR #5610, opened 2021-04-09
Draft: Type comparison rewrite

AndrolGenhald:
Well I sure picked a big rabbit hole to charge down.

I'm working on replacing Type Comparators with `TypeNode::isSubtypeOf`, with the goal of making type comparisons easier to understand and (hopefully) fewer lines of code. If you're looking at the diff I recommend scrolling down to `Atomic` first to understand what's going on.

Removing `Scalar` in favor of `TScalar` was done before I came up with `Atomic::SUPERTYPES` so that might end up being reverted (my first attempt was more based on `instanceof` and I wanted subtypes of `TScalar` to be `instanceof TScalar`).

The main stumbling block right now is `TypeComparisonResult`, I haven't been able to figure out the nuance of each of the properties:

- `$scalar_type_match_found` - Not sure I fully understand this, seems to be `true` when coercing a scalar to another scalar, except when one of them is a literal?
- `$type_coerced` - If a type can be coerced to another, `isSubtypeOf` returns false but sets this to `true`.
- `$type_coerced_strict` - I added this to indicate that a type coercion should still allow a strict comparison (ie coercion between string types).
- `$type_coerced_from_mixed` - I think this is just `true` if the type being coerced is `TMixed`? Or maybe if it's `instanceof TMixed`? (Side note, I really think `T[Non]EmptyMixed` should be renamed to `T[Non]FalsyMixed`)
- `$type_coerced_from_as_mixed` - Same as above but if it's a template type that's being coerced (eg `list<mixed>` to `list<int>`)
- `$to_string_cast` - `true` if an object is cast to a string when coercing.
- `$type_coerced_from_scalar` - Not sure I really understand what this is for, and how it's distinct from `$scalar_type_match_found`.
- `$replacement_union_type` - Haven't looked too much at it, looks like it won't matter until I go to rewrite `UnionTypeComparator`.
- `$replacement_atomic_type` - Same as above

### Coercion

This is my understanding of how coercion is supposed to work:

Type `A` can be coerced to `B` if **some but not all** instances of `A` are or can automatically be cast to `B`. This _usually_ means `A` is a supertype of `B`, and in fact many of the `...TypeCoercion` issues say `...expects B,  parent type A provided`, but this message isn't quite correct. `TNonFalsyString` can be coerced to `TNumeric` and vice versa, even though neither is a supertype of the other. Some but not all non-falsy-strings are numeric, and some but not all numerics are non-falsy-strings.


weirdan:
I suspect the most basic operation on types (let's call it *compare* for the lack of better name) that can easily produce all other relations is the following:

Given two types, *A* and *B*, *compare* is an operation that produces 3-tuple {*C*, *D*, *E*} where
* *C* is *A* \\ *B*, `\` denoting a [relative complement](https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement)
* *D* is *A* ∩ *B* (intersection)
* *E* is *B* \\ *A*

Relations can be produced as follows (assuming neither *A* or *B* are bottom types, this should be extendable to those but needs a bit more thinking)

* *A* is a supertype of *B* when *E* = ∅ and *D* ≠ ∅ 
* *A* is a subtype of *B* when *C* = ∅ and *D* ≠ ∅
* *A* is the same as *B* when *C* = ∅, *D* ≠ ∅ and *E* = ∅
* *A* is an invalid argument for parameter of type *B* when *D* = ∅
* *A* is *possibly* invalid argument for parameter of type *B* when *C* ≠ ∅ and *D* ≠ ∅

and so on

AndrolGenhald:
I actually wonder if we should rename Type Coercion to something else, like Possibly Invalid. Strictly speaking type coercion is automatic casting, so maybe it would be best to avoid using the same terminology for the issue of using a supertype when a subtype is expected.

AndrolGenhald:
@weirdan I think I'm following all of that, but are you suggesting that we should actually have a `compare` method that returns `{?Union, ?Union, ?Union}` for C, D, E? I'm not sure that's feasible.

weirdan:
I think we should aim for some minimal (but complete) set of operations on types. The less we have, the easier it is to make it sound. Now what *is* that set is a very good question.

weirdan:
> I actually wonder if we should rename Type Coercion to something else, like Possibly Invalid

There's `ArgumentTypeCoercion`, `InvalidScalarArgument` and `PossiblyInvalidArgument` (and maybe some more) for what I think to be essentially the same error: https://psalm.dev/r/7af46e3ce3

AndrolGenhald:
> There's `ArgumentTypeCoercion`, `InvalidScalarArgument` and `PossiblyInvalidArgument` (and maybe some more) for what I think to be essentially the same error: https://psalm.dev/r/7af46e3ce3

I agree, I think condensing these down to `PossiblyInvalidArgument` and `PossiblyInvalidAssignment` would be a good idea. Is there any value in having separate `Mixed...` issues for possibly invalid templates?

muglug:
> There's ArgumentTypeCoercion, InvalidScalarArgument and PossiblyInvalidArgument (and maybe some more) for what I think to be essentially the same error

Leaving aside `InvalidScalarArgument` (which is a sort of weird byproduct of PHP's type system when `declare(strict_types=0)` is not given) we have the following:

```php
abstract class A {}
final class B extends A {}
final class C extends B {}

function expectsB(B $b) {}

function takesA(A $a) {
  expectsB($a); // ArgumentTypeCoercion
}

function takesBOrC(B|C $b_or_c) {
  expectsB($b_or_c); // PossiblyInvalidArgument
}
```

It's been my experience that `ArgumentTypeCoercion` is _generally_ less likely to signify a bug than `PossiblyInvalidArgument`. Normally `ArgumentTypeCoercion` is the result of a third-party library not accurately specifying its return types.

I'm also saying something which might seem counter-intuitive: even though, in the above example,  any `A` can only be one of `B` or `C`, `A` is still a fundamentally different construct from `B|C`. It's the same logic that treats `mixed` as different from `scalar|resource|object|null`

weirdan:
> I'm also saying something which might seem counter-intuitive: even though, in the above example, any `A` can only be one of `B` or `C`, `A` is still a fundamentally different construct from `B|C`

That's actually quite intuitive. There might be some descendants unknown to Psalm, either because they are outside the codebase being analyzed (e.g. library scenario) or do not exist yet (future extensions).

AndrolGenhald:
> That's actually quite intuitive. There might be some descendants unknown to Psalm, either because they are outside the codebase being analyzed (e.g. library scenario) or do not exist yet (future extensions).

I think you might be missing the point here, it looks like what he's saying applies even if `A` were somehow prevented from being used outside of `B` and `C`.

I think the `mixed` example is easier to understand, `mixed` and `scalar|resource|object|null` mean the same thing as far as what types a variable can have, but they have different meaning to the user. `mixed` means "this variable has no type information", while `scalar|resource|object|null` means "this variable can be anything", which are subtly different.

@muglug disregarding the way it works now, would it make sense to say that a "simple" type that has no intersection or union like `A`, when it's a supertype of the desired type, should give a different error like `InsufficientlyTypedArgument` than `B|C` which should give `PossiblyInvalidArgument`?
Should the same apply outside of objects? What error should I get when trying to use a `string` as `numeric-string`?
What about when using `non-empty-string` as `non-falsy-string`, should it be treated differently since both are psalm types, which hypothetically would make it less likely that the variable was "insufficiently typed"?

muglug:
If you wanted to retire `ArgumentTypeCoercion` it'd have to wait for a major version. Then I think `LessSpecificArgument` or `UnspecificArgument` would be better than `InsufficientlyTypedArgument`.

muglug:
> Removing `Scalar` in favor of `TScalar`

They serve different purposes. `TScalar` represents the explicitly-given type `scalar`. `Scalar` is just the supertype of `TString`, `TBool` etc.

This is similar to how `Atomic` is a supertype to all classes in Psalm's source code, not `TMixed`.

Issue #5607, opened 2021-04-09
Numeric should contain float and int as well as numeric-string


danog:
https://psalm.dev/r/ec474c672b

PR #5611, opened 2021-04-09
Int and Float can be reconciled with Numeric

orklah:
This should fix https://github.com/vimeo/psalm/issues/5607


muglug:
Thanks!

PR #5613, opened 2021-04-10
Add json_last_error to impure function

samsonasik:
The usage can be: 

```bash
<?php

$string = '{\'Organization\': \'Test\'}';
json_decode($string);

var_dump(json_last_error());
```

which `json_last_error()` depends on last `json_decode()` ref https://3v4l.org/E3HKI


muglug:
Thanks!

PR #5617, opened 2021-04-12
Add missing opening PHP tag in the docs

jdreesen:
I assume the `<?php` is needed for the syntax highlighting to work as this is the only example where it's missing and the only one without syntax highlighting on this page.


muglug:
Thanks!

PR #5628, opened 2021-04-14
Fix excluding binary path under Windows

### LINKED ISSUE #5621: No issues locally but issues in GitHub actions and other environments
https://github.com/yiisoft/factory/pull/78/files - there are issues but locally I don't get any.

To reproduce:

1. `clone https://github.com/yiisoft/factory.git`
2. `git checkout -b new-config origin/new-config`
3. `composer update`
4. `./vendor/bin/psalm --no-cache`

```
Scanning files...
Analyzing files...

░

No errors found!

Checks took 1.33 seconds and used 225.092MB of memory
```

Weird thing is that colleagues are getting the errors doing the same actions. Additional info:

1. Psalm 4.7.0@d4377c0baf3ffbf0b1ec6998e8d1be2a40971005
2. PHP 8.0.3, also tried with PHP 7.4.11.
3. Windows 10 with git bash set as shell.

samdark:
Fixes #5621.

### DIFF
--- a/src/command_functions.php
+++ b/src/command_functions.php
@@ -229,7 +229,10 @@ function getPathsToCheck($f_paths): ?array
             /** @var string */
             $input_path = $input_paths[$i];
 
-            if (realpath($input_path) === realpath(dirname(__DIR__) . DIRECTORY_SEPARATOR . 'psalm')
+            if (
+                realpath($input_path) === realpath(dirname(__DIR__, 3) . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'psalm')
+                || realpath($input_path) === realpath(dirname(__DIR__, 3) . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'psalter')
+                || realpath($input_path) === realpath(dirname(__DIR__) . DIRECTORY_SEPARATOR . 'psalm')
                 || realpath($input_path) === realpath(dirname(__DIR__) . DIRECTORY_SEPARATOR . 'psalter')
                 || realpath($input_path) === realpath(Phar::running(false))
             ) {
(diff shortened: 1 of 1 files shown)


src/command_functions.php
```diff
@@ -229,7 +229,10 @@ function getPathsToCheck($f_paths): ?array
             /** @var string */
             $input_path = $input_paths[$i];
 
-            if (realpath($input_path) === realpath(dirname(__DIR__) . DIRECTORY_SEPARATOR . 'psalm')
+            if (
+                realpath($input_path) === realpath(dirname(__DIR__, 3) . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'psalm')
```
weirdan, line 233: Does this mean `vendor/bin/psalm` is not a symlink (or whatever equivalent Windows has for symlinks) on Windows?

samdark: Yes. On Windows Composer, instead of using symlinks, is creating batch-files, such as `psalm.bat` with the following content:

samdark: And bash-specific files are alike with content and adjusted paths. Not symlinks.

weirdan: For future references: when installation is run on Windows or when [`bin-compat`](https://getcomposer.org/doc/06-config.md#bin-compat) flag is set to `full` Composer generates files like this: https://github.com/composer/composer/blob/c3b76a85329c60220588e861eee02c2d0150c5ce/src/Composer/Installer/BinaryInstaller.php#L187

weirdan:
Mind fixing CodeSniffer nitpicks?

weirdan (approved):

samdark:
Fixed.

muglug:
Thanks!

samdark:
Do you plan tagging a release? I'd love to have this fixed in my projects.

// And thank you and the whole team very much for an awesome tool!

PR #5633, opened 2021-04-15
Allow Xdebug handler 2.0

Jean85:
I bumped the constraint since its limitation would prevent me to test PHP-CS-Fixer 3.0: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/composer.json#L21

I went through the [UPGRADE document](https://github.com/composer/xdebug-handler/blob/main/UPGRADE.md) but nothing seems to impede a cross-version support.


muglug:
Thanks!

PR #5654, opened 2021-04-20
adding documentation about @no-named-arguments and allowNamedArgumentCalls

orklah:
I saw a change on a project today where allowNamedArgumentCalls was set to false. I expected this to have the same effect as a `@no-named-argument` project-wide but it didn't seem to work.

After thinking about that, my assumption didn't make sense because without `@no-named-argument` directly in code, Psalm can't enforce the absence of named parameters on external calls.

Those were largely undocumented so I added some details


muglug:
Thanks!

Issue #4963, opened 2021-01-08
Empty array not considered a subtype of the intersection of an object-like array type and an array type


mstilkerich:
Hello,

  I have an associative array whose entries can be from a set of different types, say `string` and `list<string>`. The type is fixed depending on the key. However, some of the keys are computed/dynamic, so I cannot fully define the type as an object-like array.

Example:
```php
$contact = [
    'firstname' => 'John', // <-- always string
    'surname' => 'Doe', // <-- always string
    'email:work' => [ 'john.doe@example.com' ], // <-- email:* keys always have values of list<string>, but the 'work' part of the key is dynamic and not known
];
```

Now my options appear to either type this as an object-like array with all the static keys, like `array{firstname?: string, surname?: string}`. But then for any extra key I guess psalm considers the type to be mixed (actually I just discovered this option by finding from another issue that object-like array definitions allow for extra keys). The other option is to define the array type in a generic way `array<string, string|list<string>>`.

Now what would be great is if I could specify an object-like array type with extra constraints for all the extra keys, like `array{firstname?: string, surname?: string}<string, string|list<string>>`.

So my questions would be:
- Is this may already possible, but not documented?
- If not, is there a suggested best practice to deal with this situation?

(Changing the type is not an option in my case, as it is part of an interface to another project my code interfaces with.)

orklah:
Seems like this shape can exist internally in psalm but it doesn't seem like it can parse it directly in a docblock:
https://psalm.dev/r/d1ed6fd7d1

orklah:
Interesting, this seem to work: https://psalm.dev/r/531651fada

mstilkerich:
Reading through the release notes I just found for release 3.16:

> Allow intersection of object-like arrays and regular arrays
> Psalm now allows the docblock array{foo: string}&array<string, string>, understanding it to mean an array of strings with one explicit known key.

This seems to be just what I'm looking for, will try. Also I have to say that this is a non-intuitive behavior for the type intersection operator.

mstilkerich:
I tried this now, and it seems to generally work, but there appears to be a bug with empty arrays. For the plain object-like array with all-optional keys, an empty array is considered a subtype. But when I add the intersection with the other array type, the empty array is not considered a subtype any more. Seems like a bug to me.

https://psalm.dev/r/342da918bb

(btw I take back my note on non-intuitive behavior - I was still thinking of object-like arrays as being exhaustive; considering they are non-exhaustive, the intersection makes sense)

[...]

PR #5677, opened 2021-04-26
Fix array<empty, empty> not in array{foo?: mixed}<string, mixed>

VincentLanglet:
Trying to fix https://github.com/vimeo/psalm/issues/4963
I tried to modify my vendor and this seems to fix the issue but there is certainly some improvements.

I'm not familiar at all with the different types.
Should I do something with `List` too ? Is there an easier way to check for empty array ?

### DIFF
--- a/src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php
@@ -33,6 +33,18 @@ class ArrayTypeComparator
     ) : bool {
         $all_types_contain = true;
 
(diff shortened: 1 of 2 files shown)


VincentLanglet:
I can't decide if I should create a new file for the test or add one to an existing file ; any thought @muglug ?

muglug:
Merge/rebase master, where you'll find a `tests/TypeComparatorTest.php` with a place to slot yours in.

VincentLanglet:
Thanks @muglug, it works well. Test are failing without my fix. It's ready to be reviewed/improved :)

muglug:
Thanks!

PR #5682, opened 2021-04-27
Kill children on error and check if stream is closed before writing from child.

AndrolGenhald:
I've been trying some changes that happen to crash psalm a lot, and it's leaving child processes hanging taking up all my RAM.

My first though was to check the return from `fwrite`, but that [doesn't work](https://www.php.net/manual/en/function.fwrite.php#96951), so I settled on an `feof` call that seems to correctly detect if the read end of the stream has been closed.

I also killed all children when receiving an error from a child to avoid wasting CPU time, but I figure the `feof` check should be left in case the parent process somehow crashes without killing the children.

### DIFF
--- a/src/Psalm/Internal/Fork/Pool.php
+++ b/src/Psalm/Internal/Fork/Pool.php
@@ -227,7 +227,7 @@ class Pool
         $bytes_to_write = strlen($serialized_message);
         $bytes_written = 0;
 
-        while ($bytes_written < $bytes_to_write) {
+        while ($bytes_written < $bytes_to_write && !feof($write_stream)) {
             // attempt to write the remaining unsent part
             $bytes_written += @fwrite($write_stream, substr($serialized_message, $bytes_written));
 
@@ -354,6 +354,10 @@ class Pool
                                 ($this->task_done_closure)($message->data);
                             }
                         } elseif ($message instanceof ForkProcessErrorMessage) {
+                            // Kill all children
+                            foreach ($this->child_pid_list as $child_pid) {
+                                posix_kill($child_pid, \SIGKILL);
+                            }
                             throw new \Exception($message->message);
                         } else {
                             error_log('Child should return ForkMessage - response type=' . gettype($message));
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Fork/Pool.php
```diff
@@ -354,6 +354,10 @@ private function readResultsFromChildren(): array
                                 ($this->task_done_closure)($message->data);
                             }
                         } elseif ($message instanceof ForkProcessErrorMessage) {
+                            // Kill all children
+                            foreach ($this->child_pid_list as $child_pid) {
+                                posix_kill($child_pid, \SIGKILL);
```
weirdan, line 359: Wouldn't `SIGTERM` be more appropriate?

Also, you need to reap child exit statuses with [`pcntl_waitpid()`](php.net/pcntl_waitpid) lest they become zombies.

weirdan: the code to do that is already there in `Pool::wait()`, but it's not executed when stack is unwinding due to exception. It may make sense to wrap it in `try ... finally` and reap children there.

AndrolGenhald: As long as the parent process dies the zombies should be taken care of by init, is there a reason to handle that ourselves instead?

SIGTERM is probably more appropriate, I just defaulted to SIGKILL because I wanted to be certain the children were killed.

weirdan: That assumes there is init, and it's often absent in container environments. If memory serves PHPStorm plugin runs Psalm in a container - but I don't really know how exactly they do it.

AndrolGenhald: It looks like that's an issue that should be solved by the `--init`option to `docker run`, but I suppose we might as well reap them here in case someone decides to run it in a container without an init process. I'll update it tomorrow.

muglug:
Thanks!

Issue #5687, opened 2021-04-29
Can't suppress issue using class docblock annotation


bdsl:
The [docs](https://psalm.dev/docs/annotating_code/supported_annotations/) say that `@psalm-supppress` "can be used in function docblocks, class docblocks and also inline". I'm trying to use it on a class and it isn't working for me: https://psalm.dev/r/3a20a573a8

Not sure if I'm doing something wrong or this is a psalm bug. It does seem to work if the annotation is moved to the function (https://psalm.dev/r/73b48f7517) , but in my real use case I would need to repeat it several times.

This arose for me because I was working with an existing function that had an inaccurate non-nullable docblock return type. When I changed it to the more accurate nullable return type several PossiblyNull... issues of course appeared in places that used that function.

bdsl:
thanks @weirdan , @muglug

Issue #3357, opened 2020-05-13
Allow class-level @psalm-suppress


AegirLeet:
I recently deprecated a number of related classes. This causes a `DeprecatedClass` error every time one deprecated class references another deprecated class. I'd like to suppress those and adding `@psalm-suppress DeprecatedClass` to every class's class-level DocBlock seemed like the most straightforward approach. Unfortunately, that doesn't seem to have any effect.

There are alternative approaches (`<errorLevel> + <file>` in config, `@psalm-suppress` for every method etc., baseline), but a class-level suppression would be way simpler.

[Example](https://psalm.dev/r/443e68eb21)

Ideally, a `DeprecatedClass` would be generated for `C`, but not for `B`.

eithed:
> I found these snippets:
> 
> https://psalm.dev/r/443e68eb21

This doesn't seem to work

PR #5708, opened 2021-05-03
Inherit class-level suppressions when checking methods

weirdan:
Fixes vimeo/psalm#5687, fixes vimeo/psalm#3357


muglug:
Thanks!

Issue #5722, opened 2021-05-05
Simple arithmetic doesn't seem to work


danog:
https://psalm.dev/r/1419fc3ac0

weirdan:
did you expect `$b: 1`?

danog:
Yep, exactly :)

PR #5723, opened 2021-05-05
Implement literal analysis for all pure built-in string and array functions

danog:
This PR allows inferring the exact return type of a whitelisted subset of pure built-in functions, by actually executing them during static analysis.
This all started when I implemented proper return types for literal values in the ExplodeReturnTypeProvider: after implementing it manually, I realized that a similar but more generic approach could be used for all pure functions, with due precautions: this is the implementation so far.

Basically, when a call to a whitelisted built-in pure function occurs, all and only the literal types of the parameters provided to the function are extracted, and permutated to generate all combinations of possible exact return types.
If any of the parameter unions contains 0 literals, execution does not occur.
Execution is also skipped if we're running on a version of PHP that does not provide the required function (to make this reproducible across all versions, I suggest adding the symfony polyfills as runtime dependency).

The real underlying reason for this MR is an upcoming implementation of recursive static analysis in psalm, which should allow even more insight into complex systems, by similarly recursively replacing parameter types of functions with the exact literal value passed in a specific call, generating full call stacks and analyzing possible paradoxes/other issues of each single call stack.
Of course, this would also require inferring the exact literal return type of built-in functions, which is what this MR provides.

Generics and conditional types provide a similar, yet much less powerful solution: with recursive static analysis, finding bugs will be even easier, and will not require annotating every single function with generics and rewriting the behavior as a generic conditional.

### DIFF
--- a/src/Psalm/Internal/Codebase/Scanner.php
+++ b/src/Psalm/Internal/Codebase/Scanner.php
@@ -14,7 +14,6 @@ use function min;
 use const PHP_EOL;
 use Psalm\Codebase;
 use Psalm\Config;
-use Psalm\Internal\Analyzer\IssueData;
 use Psalm\Internal\ErrorHandler;
 use Psalm\Internal\Provider\FileProvider;
 use Psalm\Internal\Provider\FileReferenceProvider;
--- a/src/Psalm/Internal/Provider/FunctionReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/FunctionReturnTypeProvider.php
@@ -2,7 +2,6 @@
 namespace Psalm\Internal\Provider;
 
 use PhpParser;
-use PhpParser\Node\Arg;
 use Psalm\Codebase;
 use Psalm\CodeLocation;
 use Psalm\Context;
@@ -15,7 +14,6 @@ use Psalm\StatementsSource;
 use Psalm\Storage\FunctionLikeParameter;
 use Psalm\Type;
 use Psalm\Type\Atomic\TArray;
-use Psalm\Type\Atomic\TEmpty;
 use Psalm\Type\Atomic\TFalse;
 use Psalm\Type\Atomic\TKeyedArray;
 use Psalm\Type\Atomic\TList;
@@ -32,6 +30,7 @@ use function strtolower;
 use function is_subclass_of;
 use function function_exists;
 use function is_array;
+use function strlen;
 
 class FunctionReturnTypeProvider
 {
@@ -499,6 +498,7 @@ class FunctionReturnTypeProvider
                 $required <= count($type_args) &&
                 (count($type_args) <= count($callable->params) || $variadic)
             ) {
+                $maxStrLen = \Psalm\Config::getInstance()->max_string_length;
                 foreach ($this->permutateArguments(
                     $callable->params,
                     $type_args,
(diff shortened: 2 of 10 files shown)


danog:
Heh, sometimes I forget that I'm not using phabel))))

weirdan:
This will likely require some limits to be put in place, to make sure performance doesn't drop too much.

orklah:
For safety reason, we should make sure we only execute functions that are declare pure in psalm. It should be a test to make sure we don't add impure functions

danog:
> This will likely require some limits to be put in place, to make sure performance doesn't drop too much.

Of course, I'll add limitations for the pad and range functions.

If you were talking about the recursive static analysis, of course limitations should also be put in place to avoid excessive depth and width of the recursion tree (the first could actually be emitted as an issue on its own).

> For safety reason, we should make sure we only execute functions that are declare pure in psalm. It should be a test to make sure we don't add impure functions

This is already in place, along with a manual whitelist of allowed functions.

muglug:
I'm not sure how I feel about this.

Static analysis should mean static analysis. While Psalm does a few calculations inline (`3 + 4` becomes `7` after @orklah's PR) the general assumption is that no functions (other than the ones Psalm uses to operate) are being executed.

More importantly, I'm not convinced that much code would benefit from this — it's exceptionally rare that `explode` is called with known parameters, for example.

Do you have an example of real-world code that justifies this sort of enhancement?

danog:
>I'm not sure how I feel about this.
>Static analysis should mean static analysis.

To be honest, I too thought of this when writing this module.
However, psalm cannot statically analyze native functions: providing a literal return type for native functions will only improve the overall quality of static analysis: it is basically treating native functions as language operators, whose behavior can always be directly inferred.

About native language operators:
>While Psalm does a few calculations inline (3 + 4 becomes 7 after @orklah's PR) 
That may actually be broken, see #5722 :)

>Do you have an example of real-world code that justifies this sort of enhancement?
The main reason for this PR is the upcoming recursive static analysis module :)

danog:
Fixed all remaining test failures, the remaining failures are caused by #5743, #5742, #5722, and the TypeCombiner not understanding that int|positive-int is actually int :)

orklah:
> TypeCombiner not understanding that int|positive-int is actually int :)

Not sure it's related, but sometimes Psalm's tests doesn't display the most precise syntax. It is possible that `0|positive-int` would end up displayed as `int|positive-int` without being exactly equivalent to `int`

muglug:
> Not sure it's related, but sometimes Psalm's tests doesn't display the most precise syntax. It is possible that 0|positive-int would end up displayed as int|positive-int without being exactly equivalent to int

Yes — you can use `$varname===` instead of `$var` in the test assertions to get the exact type.

muglug:
> The main reason for this PR is the upcoming recursive static analysis module :)

You've mentioned this a couple of times — what real-world bugs will this detect that Psalm cannot?

muglug:
Sorry, would you mind providing example code for that?

danog:
A basic example:
a:php
```php
<?php

function i18n(string $key): string
{
    static $cache = [];
    static $phpCache = null;
    if (array_key_exists($key, $cache)) {
        return $cache[$key];
    }
    $phpCache ??= include 'en_EN.php';
    return $phpCache[$key];
}

echo i18n('init').PHP_EOL;
echo i18n('welcome').PHP_EOL;
```

en_EN.php
```php
<?php

return ['init' => 'Hello!'];
```

The real code is a tad bit more complex (even better suited for semantic analysis), but the core logic is the same.

muglug:
I'm going to close this, I'm afraid — this is not something that I think would be useful to a wide number of Psalm users. Instead I'd encourage you to consider writing your own plugin instead.

orklah:
@danog it would be interesting to check the result of the "check against real project" in the CI to see if the PR would have detected interesting issues in the wild

danog:
Just ran them, and as expected, nothing relevant was found, at least with the default levels.
This was expected, as this PR was supposed to be in preparation of recursive static analysis; I'll continue working on that when I have smore time, will report results (and eventually PRs ;)

PR #5726, opened 2021-05-06
unpack() can return false on error

robchett:


stubs/CoreGenericFunctions.phpstub
```diff
@@ -1163,7 +1163,7 @@ function uniqid(string $prefix = "", bool $more_entropy = false) : string {}
 /**
  * @psalm-pure
  *
- * @return array
+ * @return array|false
```
weirdan, line 1166:
```suggestion
 * @return array|false
 * @psalm-ignore-falsable-return
```

muglug:
Thanks!

PR #5738, opened 2021-05-09
improve array_splice inference

### LINKED ISSUE #5678: Return value of array_splice should be of a list type
This piece of code causes a MixedArgumentTypeCoercion issue:
https://psalm.dev/r/8cb09c981d

Works fine with array_slice though.

orklah:
This should fix #5678

I mainly used array_slice return type provider to adapt it, this will be better than before but there is definitely some improvements to be made:
1) I'm not sure what this does: https://github.com/vimeo/psalm/compare/master...orklah:array_splice?expand=1#diff-5953e521e3d01187b486b9c7a2e7e57ca6c4aa65886e4825c1bc765de64029cbR50
2) The fourth param should be used to infer the return type. Right now I'm only using the first one.
3) There is a reference for the first param that should be treated the same way
4) I only return `list<int>`, `list<string>` or `list<mixed>` to avoid making a really long list. I searched for a method that would widen a type to standard scalars (for example `'foo'|'bar'` would become string) but I couldn't find it


muglug:
Thanks!

PR #5741, opened 2021-05-10
callable-string is always a non-empty-string

danog:


danog:
OK nevermind the root of the issue is actually more complex than I initially thought, see #5742

PR #5749, opened 2021-05-11
fix #4395

### LINKED ISSUE #4395: MixedInferredReturnType when non-existing class is used in the return condition
While the `UndefinedClass` error is ok, the `MixedInferredReturnType` is unexpected, because the code is absolutely valid.

https://psalm.dev/r/83a79939ec

orklah:
This will fix #4395 partially

The issue here is that as soon as Psalm encounters an UndefinedClass, it immediately stops analyzing the rest of the branch. Trying to change this makes psalm reports the same issue (UndefinedClass) on each consecutive usage of the variable.

I tried to tackle this differently by accepting literal scalar values to be "inferred" even when the branch has no inferred types. It will not solve every problem but it will improve the situation for simple cases where a simple scalar (like false, true, null, int or strings) is used for compatibility reasons

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeCollector.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeCollector.php
@@ -20,6 +20,8 @@ class ReturnTypeCollector
      * @param  list<Type\Union>         $yield_types
      *
      * @return list<Type\Union>    a list of return types
+     *
+     * @psalm-suppress ComplexMethod to be refactored
      */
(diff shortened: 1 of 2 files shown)


orklah:
Ouch, complexity threshold reached, this is unlucky. I'm not sure how it's counted and what I can do to remain under threshold

muglug:
Please add a `@psalm-suppress with a comment. Please also add a test!

orklah:
Done!

muglug:
Thanks!

PR #5760, opened 2021-05-13
quote offset value

### LINKED ISSUE #5592: Confusing message when accessing value of value
https://psalm.dev/r/405c78c142

> using offset value of value

This message got me confused. Then I realised the offset is of `'value'`.

IMO the message should get some quotes like 

> using offset value of 'value'

so it's obvious what's only noun and what is the actual offset name.

orklah:
This will fix #5592 

Strangely, in ConstantTest.php, I couldn't add the last `'` without making the test fail. Couldn't understand why.


muglug:
Thanks!

PR #5763, opened 2021-05-13
Mixed contains everything

### LINKED ISSUE #5088: Argument has wrong type 'mixed', expecting ''
https://psalm.dev/r/8a62301736

orklah:
This should fix #5088.

However, I'm not entirely sure this is the right fix.

The ternary here seems to handle two cases: 
- we're in PHP <= 7.3 or we don't have a signature type, then we go in UnionTypeComparator::isContainedByInPhp
- we're in PHP >= 7.4 and we have a signature type, then we go in UnionTypeComparator::isContainedBy

The case in the test is pretty in between because we have a signature type only in the implementation. I don't really know why there are two methods for handling signature and phpdoc, by I kinda have the feeling we enter the wrong one.


muglug:
I think it's good!

Issue #5769, opened 2021-05-14
ArrayFillReturnTypeProvider resolves to NonEmptyList even when count == 0


olsavmic:
https://psalm.dev/r/5a4aeb43e4

ArrayFillReturnTypeProvider looks only on the first and third parameter, skipping the length parameter. The list might be empty if second parameter is set to 0.

PR #5770, opened 2021-05-14
Fix ArrayFillReturnTypeProvider - consider second param

olsavmic:
Resolves: https://github.com/vimeo/psalm/issues/5769

Improves type resolution in other cases as well.


muglug:
Thanks! I fixed failure in analysing Psalm itself (due to a un-specific return type on `count`: https://github.com/vimeo/psalm/commit/5f780e7ef7c71c256d4465f233551f2ade14a75d)

PR #5771, opened 2021-05-14
NullsafeMethodCall makes the variable not null for the rest of the scope

### LINKED ISSUE #5734: False positive: Cannot call method [] on possibly null value when using ?->
psalm should not require using `$x?->b()` when asserted that it's not nullable before in `$x?->a()`

 https://psalm.dev/r/67db6516bb

orklah:
This PR fixes #5734

It adds an assertion !null when using a NullsafeMethodCall on a variable


muglug:
Thanks!

PR #5772, opened 2021-05-14
SmallerOrEqual includes null

### LINKED ISSUE #4358: Null is mistakenly considered greater than zero
https://psalm.dev/r/8e33f66b1e

orklah:
This was a lot of trial and error, but this should fix #4358

I'm still confused by "=", "~" opposed to nothing in assertion. Do you mind explaining that quickly?

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -3807,7 +3807,7 @@ class AssertionFinder
             ) {
                 if ($conditional instanceof PhpParser\Node\Expr\BinaryOp\SmallerOrEqual &&
                     $typed_value_position === self::ASSIGNMENT_TO_RIGHT) {
-                    $if_types[$var_name] = [['=null', '=isset']];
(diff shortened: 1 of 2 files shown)


muglug:
`=` should be used whenever `===` is encountered. It means that the negation of this assertion does not imply the type is not that thing.

e.g. `$foo === $some_string` might give the assertion `$foo -> =string`. If that assertion does not hold it does not necessarily imply that `$foo` is not a string.

`~` is the above but for `==` instead of `===`

orklah:
This is now working but there's something I'm not a fan of.

For the following code:
```php
function test1(int $arg): void {
    echo $arg;
}

function test2(?int $arg): void
{
    if ($arg == null || $arg <= 0) {
        return;
    }

    test1($arg);
}
```

Psalm will emit a `Found a redundant condition when evaluating assertion ((!$arg) || (!$arg) || ($arg is isset))` because of the duplication of assertions. This is indeed redundant given PHP will include null when checking `<= 0` but checking it manually sounds like a good practice...

orklah:
This is buggy here and I fixed that in a future PR so I close this one

PR #5773, opened 2021-05-15
Redundant cast to Int preserve type 

orklah:
This PR conserve the type of an int expression that passes through an int cast. Before this PR, it was widened to `TInt`: https://psalm.dev/r/7f34cf6a8d


muglug:
Thanks!

PR #5774, opened 2021-05-15
Assert intersection type when two variables are compared

### LINKED ISSUE #4687: PossiblyNullArgument inside an if statment with type declarations
I think this is a false positive

https://psalm.dev/r/62cd4847b8

orklah:
This PR aims to fix #4687. When two variables are compared, it asserts the narrowest common type for both variables.

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
@@ -596,10 +596,6 @@ abstract class ClassLikeAnalyzer extends SourceAnalyzer
                 return $emit_issues ? null : true;
 
             case self::VISIBILITY_PROTECTED:
-                if ($appearing_property_class === $context->self) {
-                    return null;
-                }
-
                 if (!$context->self) {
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
```diff
@@ -596,10 +596,6 @@ public static function checkPropertyVisibility(
                 return $emit_issues ? null : true;
 
             case self::VISIBILITY_PROTECTED:
-                if ($appearing_property_class === $context->self) {
```
orklah, line 599: This condition was flagged as redundant by this PR because it was previously checked here: https://github.com/vimeo/psalm/pull/5774/files#diff-218a87ce16087e317b1d9c674f9998fc9617b722fb0eeaa6d09c4d54032d7dbdL559

muglug: Seems fine, it was added many eons ago and I can well believe it's redundant

muglug:
Thanks!

PR #5777, opened 2021-05-16
better inference of string increment

### LINKED ISSUE #4195: Incrementing strings
Hi!

I'm not too proud of this one but:
https://psalm.dev/r/85539c51a5

This can be really useful when working with Excel files to loop over column names. I'm fine with the "Possibly unintended string increment" issue, but `TypeDoesNotContainType` is not ideal here.

Thanks

orklah:
This should fix #4195


muglug:
Thanks!

PR #5780, opened 2021-05-16
Don't migrate virtual nodes

### LINKED ISSUE #5002: psalm-refactor replaces first character in a file with class name
When moving a derived class into another namespace, sometimes the first character of the moved class gets overwritten with the name of the base class.
~~I think this only happens if the base class is in `vendor/` or otherwise excluded in Psalm configuration.~~ Edit: this is not true, please see my updated comment below.
I have run into this issue when moving PHPUnit tests inheriting from `PHPUnit\Framework\TestCase`.

Here's a minimal reproduction repo: https://github.com/pawel-slowik/psalm-move-bug-repro 

#### Steps to reproduce

```
composer install

vendor/bin/psalm-refactor --move 'MyApp\Hello' --into 'MyApp\Foo'
```

#### The error

The opening `<?php` tag has been replaced with `Figlet?php`.

```
git diff HEAD:src/MyApp/Hello.php src/MyApp/Foo/Hello.php
```

```diff
diff --git o/src/MyApp/Hello.php w/src/MyApp/Foo/Hello.php
index f6c831f..c1ad196 100644
--- o/src/MyApp/Hello.php
+++ w/src/MyApp/Foo/Hello.php
@@ -1,6 +1,6 @@
-<?php
+Figlet?php

-namespace MyApp;
+namespace MyApp\Foo;

use Laminas\Text\Figlet\Figlet;
```

I have managed to track the issue up to here:

https://github.com/vimeo/psalm/blob/1afce4d613b6d7f0856a54eb3441947759e96bae/src/Psalm/Internal/Codebase/ClassLikes.php#L1158-L1167

where both `getAttribute` calls return nulls that get converted to integers. That's how we later end up with a `FileManipulation` having `start === 0` and `end === 1`, which overwrites the first character. But I have no idea how to fix it.

I have verified that the issue is also present in the latest version / master branch (`composer require --dev vimeo/psalm:dev-master` and then redo the reproduce steps).

pawel-slowik:
Fixes #5002

### DIFF
--- a/src/Psalm/Internal/Codebase/ClassLikes.php
+++ b/src/Psalm/Internal/Codebase/ClassLikes.php
@@ -30,6 +30,7 @@ use Psalm\Issue\UnusedConstructor;
 use Psalm\Issue\UnusedMethod;
 use Psalm\Issue\UnusedProperty;
 use Psalm\IssueBuffer;
+use Psalm\Node\VirtualNode;
 use Psalm\Progress\Progress;
 use Psalm\Progress\VoidProgress;
 use Psalm\Storage\ClassLikeStorage;
@@ -1198,6 +1199,9 @@ class ClassLikes
         bool $force_change = false,
         bool $was_self = false
     ) : bool {
+        if ($class_name_node instanceof VirtualNode) {
+            return false;
+        }
         $calling_fq_class_name = $source->getFQCLN();
 
         // if we're inside a moved class static method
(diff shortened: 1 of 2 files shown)


weirdan:
Mind fixing CS issues CircleCI complains about?

orklah:
@weirdan they're unrelated to this PR. Also I have fixed them in https://github.com/vimeo/psalm/pull/5777 (that was also unrelated! )

pawel-slowik:
@weirdan since the CS issues are unrelated to this PR, I think I'll just wait until they're fixed in master and rebase.

pawel-slowik:
After rebasing there are no more CS issues.

muglug:
Thanks!

PR #5781, opened 2021-05-17
Fix CS - Fix ArrayFillReturnTypeProvider - consider second param

olsavmic:
Follow up on: #5770 

@muglug My MR unfortunately made it to the master branch with bad code formatting (I suppose that it's required to pass the whole pipeline casually but it was somehow skipped in this case due to manual intervention?)

Let's make the pipeline green again :)


muglug:
Thanks!

PR #5793, opened 2021-05-18
strlen() pure and can return positive-int

still-dreaming-1:
Not all tests were passing locally. We'll what happens here. Hopefully not caused by the changes I made...

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -704,6 +704,17 @@ function htmlspecialchars_decode(string $string, ?int $quote_style = null) : str
 
 /**
  * @psalm-pure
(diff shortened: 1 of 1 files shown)


still-dreaming-1:
At least some of these failures are not from my changes. I'm confused, why are PRs with failures being merged? This makes it hard to know if failures in future PRs are legitimate or where the errors originate from. Especially in an open source project it seems like a bad practice.

muglug:
Yesterday GitHub Actions was down for part of the day, which led to some stuff getting merged with errors.

Your PR also broke two tests — at least one was due to a bug your PR uncovered. I've opened an issue to track that bug: https://github.com/vimeo/psalm/issues/5804

PR #5803, opened 2021-05-19
allow combining TKeyedArray list-like with TList

### LINKED ISSUE #5795: (array)string returns array<int, string> instead of list<string>
This comes from trying to satisfy the new @psalm-var hints in laminas packages.

https://psalm.dev/r/71d8338586

Since the array value type is `string|list<string>`, the cast to array should always return `list<string>` type. There isn't a generic array returned from the cast since the value is always string.

```
/**
  @psalm-var array<string, list<string>>
  */
protected $headers = [];
    
/**
  @psalm-param array<string, string|list<string>> $headers
  */
public function addHeaders(array $headers): void {
    foreach ($headers as $key => $val) {
        $this->headers[$key] = (array)$val;
    }
}
```

orklah:
Fixes #5795

### DIFF
--- a/tests/ArrayAssignmentTest.php
+++ b/tests/ArrayAssignmentTest.php
@@ -797,7 +797,7 @@ class ArrayAssignmentTest extends TestCase
                     $a = null;
                     $a[0][] = 1;',
                 'assertions' => [
-                    '$a' => 'array{0: non-empty-list<int>}',
+                    '$a' => 'array{non-empty-list<int>}',
                 ],
                 'error_levels' => ['PossiblyNullArrayAssignment'],
             ],
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -470,7 +470,6 @@ class AssertionFinder
 
                     return $if_types ? [$if_types] : [];
                 }
-
             }
         }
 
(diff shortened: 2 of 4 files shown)


orklah:
Somehow, the PR changed the format of a test from `array{0: non-empty-list<int>}` to `array{non-empty-list<1>}`:
https://github.com/vimeo/psalm/pull/5803/files#diff-4e6412c7bcc831346dd4bb52849daa61aa21e73e58fa9121ba77f5b7915757e8R800

However, those two notations are equivalent so I made the change in the test assertion

orklah:
I also fixed a CS issue I introduced in a previous PR

tests/ArrayAssignmentTest.php
```diff
@@ -797,7 +797,7 @@ public function offsetSet($offset, $value): void
                     $a = null;
                     $a[0][] = 1;',
                 'assertions' => [
-                    '$a' => 'array{0: non-empty-list<int>}',
+                    '$a' => 'array{non-empty-list<int>}',
```
muglug, line 800: This is more correct — using explicit `int` keys implies the keys don't have to be ordered

muglug:
Ultimately this is slightly not the right fix — see https://github.com/vimeo/psalm/commit/ccd50fdb346eaebd98fd426dee60986914ebeffe

orklah:
Ok, I thought the issue came from combining, but yeah, making every cast a list seems logical. I didn't thought of it at the time because I didn't know of the list behaviour of TKeyedArray. Thanks!

PR #5813, opened 2021-05-22
Add mongodb Cursor and CursorInterface stubs and fix WriteConcern::__construct

franmomu:
Stubs for:

https://www.php.net/manual/en/class.mongodb-driver-cursorinterface.php
https://www.php.net/manual/en/class.mongodb-driver-cursor.php

**UPDATE**: I've also updated `WriteConcern::__construct` based on https://www.php.net/manual/es/mongodb-driver-writeconcern.construct.php

### DIFF
--- a/stubs/mongodb.phpstub
+++ b/stubs/mongodb.phpstub
@@ -1,0 +1,59 @@
+<?php
+
+namespace MongoDB\Driver;
+
+use Iterator;
+use Traversable;
+
+/**
+ * @template-covariant TKey
+ * @template-covariant TValue
+ *
+ * @template-extends Traversable<TKey, TValue>
+ */
+interface CursorInterface extends Traversable
+{
+    /**
+     * @return array<TValue>
+     */
+    public function toArray();
+}
+
+/**
+ * @template-covariant TValue of array|object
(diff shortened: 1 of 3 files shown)


stubs/mongodb.phpstub
```diff
@@ -0,0 +1,56 @@
...
+     * @return array<TValue>
+     */
+    public function toArray();
+}
+
+/**
+ * @template-covariant TValue of array|object
+ *
+ * @template-implements Iterator<int, TValue>
+ * @template-implements CursorInterface<int, TValue>
```
franmomu, line 23: The types are based on 
`array|object`: https://www.php.net/manual/en/mongodb-driver-cursor.current.php
`int`: https://www.php.net/manual/en/mongodb-driver-cursor.key.php

muglug:
You need to add logic in `Config.php` that'll load the stubs if the mongodb extension is enabled

franmomu:
> You need to add logic in `Config.php` that'll load the stubs if the mongodb extension is enabled

Makes sense, thanks!

muglug:
Thanks!

Issue #4996, opened 2021-01-13
`assert($v >= 0)` with `$v : int` is not understood as `positive-int|0`


Ocramius:
https://psalm.dev/r/447ee45ac7

```php
<?php

/** @psalm-return positive-int|0 */
function gimmePositiveOrZero(int $input): int
{
    assert($input >= 0);
    
    return $input;
}


/** @psalm-return positive-int|0 */
function gimmePositiveOrZero2(int $input): int
{
    assert($input === 0 || $input > 0);
    
    return $input;
}
```

```
Psalm output (using commit 1afce4d): 

INFO: LessSpecificReturnStatement - 8:12 - The type 'int' is more general than the declared return type 'int(0)|positive-int' for gimmePositiveOrZero

INFO: MoreSpecificReturnType - 3:19 - The declared return type 'int(0)|positive-int' for gimmePositiveOrZero is more specific than the inferred return type 'int'
```

In practice:

 * `assert($input === 0 || $input > 0)` is understood as `0|positive-int`
 * `assert($input >= 0)` is **not** understood as `0|positive-int`

The workaround for now is obviously to replace `>=` with a boolean expression.

muglug:
Now fixed!

PR #5815, opened 2021-05-22
Add assertions for >=0

orklah:
This is a fix for #4996

This is interesting and it makes some new issues pops. This is due to the new inferrence being `0|positive-int` which contains a literal int and triggers the check for PossiblyUndefinedIntArrayOffset.

Not sure if I should fix the phpdoc in Psalm or make a special exception for `0|positive-int` in this check


muglug:
Mind rebasing this onto 938afc5ae? I think that should solve the issue

orklah:
It's fixed indeed, thanks!

muglug:
Thanks!

PR #5816, opened 2021-05-23
refinement of Possibly undefined int array offset

orklah:
This PR is related to https://github.com/vimeo/psalm/pull/5815

The test I added here does not fail without this PR but it fail with https://github.com/vimeo/psalm/pull/5815. This is the fix that allows accessing a `list` with a `0|positive-int` offset without triggering `PossiblyUndefinedIntArrayOffset` (The same way it's not triggered when accessing an `array` with a non-literal `int`)

Unfortunately the tests fail. The `testEnsureListOffsetExistsAfterArrayPop` does not emit `PossiblyUndefinedIntArrayOffset` anymore but I'm not sure what this test is checking and if I can just remove `$this->expectExceptionMessage('PossiblyUndefinedIntArrayOffset');` from it without altering the test purpose


orklah:
This should be no longer needed, https://github.com/vimeo/psalm/commit/938afc5aeacf9a077ac0e8f0729995f5fd3b96e6 is a similar fix.

PR #5817, opened 2021-05-23
Transform bad offsets

### LINKED ISSUE #4894: Incorrect definition of the array type
https://psalm.dev/r/0bf218383c

1)
```php
$arr1 = [$index => 5];
```
2)
```php
$arr2 = [];
$arr2[$index] = 5;
```
The above two examples do the same thing, but in the second example, the array type is not defined correctly.

orklah:
This fixes #4894 in two ways:
- It transforms any literal float to a literal int instead of just an int on ArrayAnalyzer
- It transforms any bad offset to an acceptable offset in ArrayFetchAnalyzer

I'm not completely satisfied by the fix however, the two objects each detects bad offsets with a different code. I just added the transformation in between. This may be refactored later to extract the common code and use the same detection


muglug:
Thanks!

Issue #4499, opened 2020-11-06
Psalter doesn't add return type on switches


orklah:
For the following example, if we try the fix button, Psalter will add `: array` in signature:
https://psalm.dev/r/fec30a288a

but for this example:
https://psalm.dev/r/5149aedc01
Psalter only adds the return type in the documentation

Thanks

weirdan:
Looks like Psalm thinks that return is conditional. If you add an unconditional return (say, `return null;`) after the switch native return type will be added just fine: https://psalm.dev/r/16ec47c05c

orklah:
@muglug Would this be difficult to solve? I'd like to try but I'm afraid this could prove difficult for me

PR #5819, opened 2021-05-23
Handle the case where a switch return on each case

orklah:
This PR fix https://github.com/vimeo/psalm/issues/4499.

It handles the case where a switch always return. This allows for example Psalter to add the return type automatically


muglug:
Thanks!

PR #5826, opened 2021-05-24
Allow accessing offset 0 on a list

orklah:
This PR is another version of #5816 now that I understand better what's going on.

The tests fail for the same reason as last time and if we delete this test, this is indeed a potential bug Psalm wouldn't detect anymore.

However, the current behaviour is not coherent.
PossiblyUndefinedIntArrayOffset is emitted when an array has an int offset and we access it with a specific offset.
It's not emitted when the array is accessed with a `int` and since yesterday it's not emitted either on `positive-int`. It just seem weird to allow `positive-int` but not `0` on a list due to the nature of the list.

Furthermore, it generates a loot of false positives(or at least issues that people know it's not worth fixing). See below Psalm's own baseline where 99% of issues are PossiblyUndefinedIntArrayOffset and a good amount is [0] access. (I checked, there are 38 and the big majority of them can't be fixed just by a `non-empty-list` somewhere)

So the question is, do we keep the current behaviour or do we change this (and remove the failing test)?

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -36,10 +36,9 @@
     </PossiblyUndefinedIntArrayOffset>
   </file>
   <file src="src/Psalm/Internal/Analyzer/ClassAnalyzer.php">
-    <PossiblyUndefinedIntArrayOffset occurrences="3">
+    <PossiblyUndefinedIntArrayOffset occurrences="2">
       <code>$comments[0]</code>
       <code>$stmt-&gt;props[0]</code>
-      <code>$uninitialized_variables[0]</code>
     </PossiblyUndefinedIntArrayOffset>
   </file>
   <file src="src/Psalm/Internal/Analyzer/CommentAnalyzer.php">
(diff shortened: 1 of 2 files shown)


muglug:
I don't see the inconsistency here? https://psalm.dev/r/8fbb9b168c

orklah:
Well, for the third case, Psalm doesn't complain about accessing the array with `0|positive-int` but it does with `0`?

`0` is way less risky to use on an array than `0|positive-int`, not to mention `int` for which half of its possible values are outright incompatible with a list (negative ints)

muglug:
The `PossiblyUndefinedIntArrayOffset` issue is triggered when using an explicit reference (e.g. `$foo[0]`) where Psalm cannot tell that the offset exists.

Your counter-examples are false-negatives when using a variable reference (e.g. `$foo[$b]`) but in my experience those are much less common as a source of issues.

orklah:
I guess there's a logic, but then, shouldn't those two be ruled out too?
https://psalm.dev/r/90f8658f20

muglug:
No, because in that case `$b` is known and so Psalm treats that as `$a[0]`

orklah:
That still seems weird to me but ok :)

Issue #5831, opened 2021-05-26
Allow adding new errors to baseline.


bdsl:
Currently psalm supports removing errors from the baseline with `--update-baseline` but I don't believe adding new errors is supported. Adding new errors is of course possible running psalm with `--set-baseline=psalm-baseline.xml` but I thin that only works properly if the baseline file is deleted first, and it's a little awkward to use.

Of course in general we want errors to go down not up and we don't want to introduce new buggy code into a codebase, but I think there's a fairly common scenario where adding new errors make sense.

Suppose we have a legacy codebase that looks like this - simplified for brefity:

```
<?php

class PizzaTopping{
	public function __construct(
        public string $name
    ) {}
}

/**
 * @return PizzaTopping
 */
function getPizzaTopping(string $toppingId)
{
    switch ($toppingId) {
        case '1':
            return new PizzaTopping('Black Olive');
        default:
	    return null;
    }
}

function showTopping(string $toppingId): void
{
    echo getPizzaTopping($toppingId)->name;
}

// many more functions that do other things with piza toppings not shown. Some do null checks some don't.
```

Psalm will detect issues on the incorrectly documented getPizzaTopping function, but as this is a large legacy code base we add them to a baseline when we first start using Psalm and move on with our lives.

Some time later we decide we want to fix getPizzaTopping and correctly document it as returning `?PizzaTopping` . At that point Psalm will detect that we should have done a null check in `showTopping` and report an error there. But `showTopping` has been in production for years and it's working fine, and there are dozens of functions like it. It isn't cost effective to fix them all now. So we probably revert the change to the `getPizzaTopping` declaration.

Instead I'm proposing a feature where we would add the issues detected in `showTopping` to the baseline. We don't need to fix it since we we know it's working, but the next time we call `getPizzaTopping` Psalm will insist we do the null check where required.

orklah:
--set-baseline do work well even if there is already a baseline. The only thing is that the end report won't display any issues and everything is added to the baseline silently

Furthermore, it's unfortunate that we have to retype the filename everytime.

A --add-to-baseline that displays the list of new issues added would be wonderful

bdsl:
Right, the list of new issues would be nice to paste into the PR that fixes the `getPizzaTopping` declaration, since the baseline.xml isn't really designed to be easy to read.

bdsl:
[...]

PR #5838, opened 2021-05-26
fix issue reporting when baseline already exists

orklah:
This PR improve the situation on #5831

It fixes an issue where the baseline is read after it was updated, ignoring every issue for the final report. With this fix, only the new issues are displayed


muglug:
Thanks!

Issue #5841, opened 2021-05-27
@psalm-internal don't throw error when used for constructor


vjik:
See example: https://psalm.dev/r/93b9cb3533

bdsl:
`@internal` also doesn't throw error here. It works if you call the constructor directly by name. That's not a solution or even a usable workaround, but it shows that the issue is around tracking that `new` refers to the constructor.

https://psalm.dev/r/af0865496a

bdsl:
Maybe a useful workaround is to make the constructor private and introduce a named constructor (i.e. static method that returns an instance of self). That static method can then be marked as `@internal`, and used as the constructor.

vjik:
> Maybe a useful workaround is to make the constructor private and introduce a named constructor (i.e. static method that returns an instance of self). That static method can then be marked as `@internal`, and used as the constructor.

Yes, this solutions for my case.
But I would like to specify it for the constructor as well :)

PR #5843, opened 2021-05-27
Check whether constructor is internal on `new` call

weirdan:
Fixes vimeo/psalm#5841


bdsl:
@weirdan probably worth adding a test case for the valid code case where `NamespaceAnalyzer::isWithin()` returns true, but otherwise LGTM.

bdsl (approved):

muglug:
Thanks!

PR #5844, opened 2021-05-27
allow psalter to add literal string/int in psalm phpdoc

orklah:
This PR will allow Psalter to be more precise when encountering a small list of literals. I arbitrary chose 3 as the limit to avoid making a mess but this could be changed in the future.


muglug:
Seems like a decent idea, though it might want to be increased in the future. Thanks!

Issue #5782, opened 2021-05-17
Type imports on methods are silently ignored


weirdan:
https://psalm.dev/r/9392781c99

#### Expected

Either no issues or `InvalidDocblock: type imports on methods have no effect, move it to class docblock`

orklah:
More broadly, `@var` on methods or `@param` on attributes should be checked too

weirdan:
Fixed `@var` on methods, but flagging `@param` on properties is harder due to how property docblocks are parsed, so I'm not going to pursue that for now.

PR #5845, opened 2021-05-27
Flag unexpected tags on function docblocks

weirdan:
This makes Psalm flag `@var` and `@psalm-import-type` tags used on methods (the list should also be fairly easy to extend). 

Fixes vimeo/psalm#5782


muglug:
Thanks!

orklah:
Oh, I missed you fixed that, that's awesome! It found some occurences at work

PR #5853, opened 2021-05-28
Union types php7

orklah:
This is a draft for a fix of https://psalm.dev/r/334d9720aa?php=7.1

When clicking fix code on this link, Psalter will add union types even though the php version is set to 7.1.

It happens too when running psalter on psalm itself when the php version running psalm is >8.

It's due to an ambiguity in php versions in Psalm.

There are two php versions:
- One in codebase. It's the PHP version that is fetched directly from PHP_MINOR_VERSION and PHP_MAJOR_VERSION, which means it's the version used to run Psalm
- One in config, through the `getPhpVersion` method. It's the one fetched from the config file, the command line or from composer directly. It's the version we want to check the project with and we don't want psalter to use another version when fixing code.

Unfortunately, it's almost always the first one that's used in the code. This means the behaviour when the code doesn't use the same php version than the server used to run psalm is pretty random.

It have effects in CI, in psalm.dev and even in psalm's own test where the php version we can set is actually the first one, which makes test in this PR fail.

How should we proceed forward on this? Should we try to make a big change and replace almost every usage of the Codebase versions by the one in config?

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
@@ -913,7 +913,6 @@ class ReturnTypeAnalyzer
 
         $allow_native_type = !$docblock_only
             && (int)$php_major_version >= 7
-            && ((int)$php_minor_version >= 8 || $inferred_return_type->isSingle())
             && (
                 $codebase->allow_backwards_incompatible_changes
                 || $is_final
@@ -926,8 +925,8 @@ class ReturnTypeAnalyzer
                     $source->getNamespace(),
                     $source->getAliasedClassesFlipped(),
                     $source->getFQCLN(),
-                    $codebase->php_major_version,
-                    $codebase->php_minor_version
+                    (int)$php_major_version,
+                    (int)$php_minor_version
                 ) : null,
             $inferred_return_type->toNamespacedString(
                 $source->getNamespace(),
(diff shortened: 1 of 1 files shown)


weirdan:
> There are two php versions:

I believe they are synced by the call to `ProjectAnalyzer::setPhpVersion()`: https://github.com/vimeo/psalm/blob/b259296457c0bd247586d61591e9e335bac5c380/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php#L1260-L1281 in `src/psalm.php`: https://github.com/vimeo/psalm/blob/b259296457c0bd247586d61591e9e335bac5c380/src/psalm.php#L642-L652

orklah:
Indeed, then there's something fishy going on. I just tried 
`
        var_dump($codebase->php_major_version.'.'.$codebase->php_minor_version);
        var_dump($project_analyzer->getConfig()->getPhpVersion());
`
on ReturnTypeAnalyzer line 911 and it dumped:
`
string(3) "8.0"
string(3) "7.1"
`

8.0 being my install of PHP and 7.1 being the minimal version required by Psalm in composer.

I'll look further

orklah:
@weirdan thanks for the pointer, turns out psalter.php doesn't do the same thing:

```php
if (isset($options['php-version'])) {
            if (!is_string($options['php-version'])) {
                die('Expecting a version number in the format x.y' . PHP_EOL);
            }

            $project_analyzer->setPhpVersion($options['php-version']);
        }
```

no trace of getPhpVersion to fetch composer version. I'll fix this better in another PR

PR #5855, opened 2021-05-28
fix wrong version used in psalter

orklah:
Following #5853, @weirdan pointed the real issue, the php-version from composer is not pushed to the Codebase instance in psalter.php like it is on psalm.php. This PR fixes that


muglug:
Thanks!

Issue #4523, opened 2020-11-09
Implement nullsafe short-circuiting for methods


ondrejmirtes:
From the RFC: https://wiki.php.net/rfc/nullsafe_operator

> This RFC proposes full short circuiting. When the evaluation of one element in the chain fails the execution of the entire chain is aborted and the entire chain evaluates to null. The following elements are considered part of the chain.

- Array access ([])
- Property access (->)
- Nullsafe property access (?->)
- Static property access (::)
- Method call (->)
- Nullsafe method call (?->)
- Static method call (::)

Playground link: https://psalm.dev/r/4f5df32590
Runtime proof: https://3v4l.org/GqDeo

Line 19 is fine, but line 13 should also indicate nullable result.

Feel free to also check behaviour for ArrayDimFetch, StaticCall, and StaticPropertyFetch.

iluuu1994:
Just noticed the same issue. Somewhat simplified example. https://psalm.dev/r/1a96184835

PR #5856, opened 2021-05-28
Nullsafe calls should not trigger NPE on consecutive calls 

orklah:
This should solve #4523

I'm sure this is not the best way to do that but it seems to do well enough.

One side effect I could foresee is if `baz()` can return null in `$a?->baz()`, then a potential NPE won't be reported on `$a?->baz()->bar()` anymore

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/NullsafeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/NullsafeAnalyzer.php
@@ -90,6 +90,11 @@ class NullsafeAnalyzer
 
         $statements_analyzer->node_data = $old_node_data;
 
+        if ($ternary_type) {
+            //after a nullsafe, the null propagate to consecutive calls but it should not trigger issues
+            $ternary_type->ignore_nullable_issues = true;
+        }
+
         $statements_analyzer->node_data->setType($stmt, $ternary_type ?: Type::getMixed());
 
         return true;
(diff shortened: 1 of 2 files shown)


muglug:
Is this still caught correctly?

```php
<?php

interface Bar {
    public function doBaz(): void;
}
interface Foo {
    public function getBar(): Bar;
}
function fooOrNull(): ?Foo {
    return null;
}
$a = fooOrNull()?->getBar();
$a->doBaz();
```

orklah:
No, it's not caught anymore. After this PR, Psalm would effectively treat any null value from a nullsafe call as "acceptable". 

If we wanted to do better, I guess this would need a new flag on the union `nullsafe_shortcircuit` that would be treated independantly from the null value inside the Union and that would disable NPE only in the current expression. I guess we would have to reset this new flag at the end of each expression.

muglug:
I think this is an AST-sensitive thing, not a type-sensitive thing — we need to add (and remove where necessary) an `->inside_nullsafe` flag on the Context object

muglug:
Oh actually sorry I think your solution is better, and assignments would remove that flag from the type. Icky, but the actual implementation is icky IMO

muglug:
Fixed in f3fc112ba

muglug:
I went with the AST way, which is I think how PHP itself behaves

orklah:
Oh great! Thanks for fixing that :)

PR #5858, opened 2021-05-30
Cli command classes

weirdan:
Over the years Psalm's entrypoints have grown quite large. This PR aims to solve it by moving the entrypoint code to their own classes.

Overview of changes:
* `src/*.php` entrypoints moved to corresponding classes in `src/Psalm/Internal/Cli` folder
* Global functions in `\Psalm` namespace deprecated and moved to `Psalm\Internal\CliUtils` class
* `Psalm\Internal\Cli\Psalm::run()` (originally `src/psalm.php`) split into multiple methods
* `Psalm\Internal\Cli\Psalter::run()` (originally `src/psalter.php`) split into multiple methods


weirdan:
The remaining CI issue is somewhat puzzling. It's only reproducible with `--threads=1` (which is implicit on AppVeyor).

orklah:
There's a lot of code being repeated (sometimes with mistakes) between entry points, for example, the php-version fixed here #5855.

I see you introduced a `initPhpVersion` on Psalm entry point. Maybe this method should be extracted into an EntryPoints class and used by everyone?

weirdan:
@orklah I'd like to defer that to subsequent, more focused, PRs. While I certainly agree with the sentiment, this one is quite big already. Here I aimed for bug-for-bug compatibility, to ease the transition.

orklah:
Well, seems weird you refactored things in methods only in Psalm entrypoint and not on Psalter then.

But apart from that, the PR seems great :)

weirdan:
> Well, seems weird you refactored things in methods only in Psalm entrypoint and not on Psalter then.

Initially I did not intend to do that in this PR, but Psalm was complaining about `ComplexMethod` :smile:. So once it stopped doing that I deemed the PR good enough.

orklah:
Can relate to that :D

muglug:
Thanks!

> Initially I did not intend to do that in this PR, but Psalm was complaining about ComplexMethod 😄. So once it stopped doing that I deemed the PR good enough.

Is this annoying or is it good?

weirdan:
> Is this annoying or is it good?

It's ok, but I did not expect that - the method was no more complex than the closure originally found in `src/psalm.php`, but Psalm did not flag that closure. Also the measure of complexity used is not very intuitive. IIRC, the most complexity reduction came from extracting code like this: https://github.com/vimeo/psalm/blob/38c452ae584467e939d55377aaf83b5a26f19dd1/src/psalm.php#L218-L236
Arguably this does not look the most complex part of that method.

PR #5862, opened 2021-05-30
Add false as return value for mb_convert_encoding

kesselb:
mb_convert_encoding could return false: https://www.php.net/manual/en/function.mb-convert-encoding

https://psalm.dev/r/1887ce60a3


muglug:
Thanks!

PR #5864, opened 2021-05-31
Make toPHP second parameter optional

franmomu:
The second parameter of [`MongoDB\BSON\toPHP` function](https://www.php.net/manual/es/function.mongodb.bson-tophp.php) is optional.

https://psalm.dev/r/1113b68dab


muglug:
Thanks!

PR #5879, opened 2021-06-02
Type negation in generic context

### LINKED ISSUE #5871: Assertion negation doesn't work inside generic context
Without generic context plsam negates assertion correctly: https://psalm.dev/r/54d962f958

With generic context incorrectly: https://psalm.dev/r/6a9d6ed02a

Here psalm tries remove `Valid<T>` from `Union`. But unsuccessfully. Because atomic type has `Valid<T:fn-insidegenericcontext>` as key, not `Valid<T>`.

https://github.com/vimeo/psalm/blob/36d670b9559eae0d5ce067d23e242aa589ea6311/src/Psalm/Internal/Type/NegatedAssertionReconciler.php#L233-L235

Why `fn-insidegenericcontext` is missing ?)

klimick:
fix #5871
My first PR to psalm. Not sure if I did it right.


muglug:
Thanks, looks great!

Issue #5882, opened 2021-06-03
Allow plugins to declare their additional scanners or analyzers


ohader:
It should be possible that plugins can register their custom scanners or analyzers.
Thus, explicitly configuring a plugin is enough...

ohader:
I can kick-start something...

muglug:
Sure, I'd be open to that

muglug:
Sorry clicked wrong button

PR #5883, opened 2021-06-03
[FEATURE] Allow plugins to declare their custom scanner or analyzer

ohader:
Fix: #5882

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,4 @@
 /vendor-bin/*/vendor/
 /tests/fixtures/symlinktest/*
 
-.idea/inspectionProfiles/Project_Default.xml
-.idea/workspace.xml
+.idea/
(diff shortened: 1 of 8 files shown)


ohader:
* Psalm self-test will fail, since new methods are not used - I was reluctant modifying existing `RegistrationInterface` - maybe there might be a new `FileTypeRegistrationInterface` (not sure about the name)
* documentation is missing - will take care
* probably some test cases would be nice as well

muglug:
> Psalm self-test will fail, since new methods are not used

That will be fixed if you add tests!

weirdan:
> I was reluctant modifying existing `RegistrationInterface`

It should be fairly safe to expand that. Third-party code is not expected to implement that interface, plugin authors should only consume it.

ohader:
> > I was reluctant modifying existing `RegistrationInterface`
> 
> It should be fairly safe to expand that. Third-party code is not expected to implement that interface, plugin authors should only consume it.

Nice, just spotted this comment. Will extend the interface later...

.gitignore
```diff
@@ -11,5 +11,4 @@
 /vendor-bin/*/vendor/
 /tests/fixtures/symlinktest/*
 
-.idea/inspectionProfiles/Project_Default.xml
-.idea/workspace.xml
+.idea/
```
ohader, line 14: Sorry for the drive-by fix...

muglug:
Thanks!

PR #5888, opened 2021-06-04
Fix generic destruction in assertion

### LINKED ISSUE #5887: Type assertion destructs generic source with `&&` operator
`$val1` and `$val2` should have `Valid<T:fn-insidegenericcontext as mixed>` type as `$val3`
But actually type for `$va1` and `$val2` is `Valid<T>`

https://psalm.dev/r/48db823f3b

klimick:
fix #5887


muglug:
Thanks!

PR #5897, opened 2021-06-06
Formally deprecate `allowCoercionFromStringToClassConst`

weirdan:
`allowCoercionFromStringToClassConst` is a recognized config attribute, but it has no effect. Previous attempt at removing it has caused a lot of breaks (#3982) so it had to be reverted (#4046).

This time, instead of failing validation we emit a deprecation issue and let Psalm run as usual. This should provide smooth upgrade path.


muglug:
Thanks!

PR #5900, opened 2021-06-07
Enforce `use` sort

weirdan:
It's just a bit easier on the eye this way. This is also automatically fixable with `vendor/bin/phpcbf` so it shouldn't affect authors/contributors too much.


muglug:
Thanks!

Issue #5870, opened 2021-06-01
False-positive DocblockTypeContradiction - `sscanf()` can return array with nulls


fluffycondor:
https://psalm.dev/r/664f29dacd

Expected: No errors
Got: ERROR: DocblockTypeContradiction - 5:5 - Cannot resolve types for $lon - docblock-defined type float|int|string does not contain null

PR #5901, opened 2021-06-07
Fix `sscanf()` signature

weirdan:
Fixes vimeo/psalm#5870

`sscanf()` returns:
* `null` when input string is empty and no out variables provided: https://3v4l.org/2OfgdC
* `list` with nulls for unmatched specifiers: https://3v4l.org/Njvgf
* `null` for out variables corresponding to unmatched specifiers and the number of matched specifiers as a return value: https://3v4l.org/8OZMk
* `-1` when input string is empty and out variables are provided: https://3v4l.org/qq1Al


muglug:
Thanks!

Issue #5903, opened 2021-06-07
Incorrect signature for SessionHandlerInterface


christeredvartsen:
It seems that the signatures for `SessionHandlerInterface` in the callmap does not match the PHP manual. There are some mismatches with regards to both types and variable names.

Class synopsis from [the manual](https://www.php.net/manual/en/class.sessionhandlerinterface.php):

```
SessionHandlerInterface {
  /* Methods */
  abstract public close ( ) : bool
  abstract public destroy ( string $id ) : bool
  abstract public gc ( int $max_lifetime ) : int|bool
  abstract public open ( string $path , string $name ) : bool
  public read ( string $id ) : string|false
  abstract public write ( string $id , string $data ) : bool
}
```

the current signatures found in [the callmap](https://github.com/vimeo/psalm/blob/master/dictionaries/CallMap.php):

```
'SessionHandlerInterface::destroy' => ['bool', 'session_id'=>'string'],
'SessionHandlerInterface::gc' => ['bool', 'maxlifetime'=>'int'],
'SessionHandlerInterface::open' => ['bool', 'save_path'=>'string', 'name'=>'string'],
'SessionHandlerInterface::read' => ['string', 'session_id'=>'string'],
'SessionHandlerInterface::write' => ['bool', 'session_id'=>'string', 'session_data'=>'string'],
```

Here is a link to psalm.dev with a dummy handler implementation with signatures according to the PHP docs where you can see the issue:

https://psalm.dev/r/afd42116f5

weirdan:
Fixed in #5904 (thanks).

PR #5904, opened 2021-06-07
Update signatures according to the PHP manual

christeredvartsen:
Resolves the problems described in #5903.


dictionaries/CallMap.php
```diff
@@ -12043,11 +12043,11 @@
 'SessionHandler::validateId' => ['bool', 'session_id'=>'string'],
 'SessionHandler::write' => ['bool', 'id'=>'string', 'data'=>'string'],
 'SessionHandlerInterface::close' => ['bool'],
-'SessionHandlerInterface::destroy' => ['bool', 'session_id'=>'string'],
-'SessionHandlerInterface::gc' => ['bool', 'maxlifetime'=>'int'],
-'SessionHandlerInterface::open' => ['bool', 'save_path'=>'string', 'name'=>'string'],
-'SessionHandlerInterface::read' => ['string', 'session_id'=>'string'],
-'SessionHandlerInterface::write' => ['bool', 'session_id'=>'string', 'session_data'=>'string'],
+'SessionHandlerInterface::destroy' => ['bool', 'id'=>'string'],
+'SessionHandlerInterface::gc' => ['int|bool', 'max_lifetime'=>'int'],
```
weirdan, line 12047:
```suggestion
'SessionHandlerInterface::gc' => ['int|false', 'max_lifetime'=>'int'],
```
Ref: https://www.php.net/manual/en/sessionhandlerinterface.gc.php
> Returns the number of deleted sessions on success, or `false` on failure.

weirdan (changes requested):

dictionaries/CallMap.php
christeredvartsen: Not sure about this one. The text is interpreted as `int|false`, but the signature itself as presented in the docs uses `int|bool`. Which one should be the deciding one? I guess one could provide a change to the PHP docs for this particular one?

weirdan: Source code is the ultimate truth, and it says `MAY_BE_LONG|MAY_BE_FALSE`: https://github.com/php/php-src/blob/787db3af484a4680b2aa25bd1dcc6a9b09f2c1f2/ext/session/session_arginfo.h#L113-L115

PHP docs have been pretty inaccurate historically, specifically for arg/return types. This situation is improving, but is far from perfect.

christeredvartsen: 👍  Will push the change.

weirdan (approved):

muglug:
Thanks!

PR #5917, opened 2021-06-10
Detect unused return values

muglug:
I think this could be useful?

Here's the associated diff after I ran against Psalm’s codebase: https://github.com/vimeo/psalm/commit/8509999f3fc2afe550c3f9a7a9612007968c5aa6

Triggers when a method has a non-void return type if the method return value is never used.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -381,6 +381,7 @@
             <xs:element name="PossiblyUnusedMethod" type="MethodIssueHandlerType" minOccurs="0" />
             <xs:element name="PossiblyUnusedParam" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="PossiblyUnusedProperty" type="PropertyIssueHandlerType" minOccurs="0" />
+            <xs:element name="PossiblyUnusedReturnValue" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="PropertyNotSetInConstructor" type="PropertyIssueHandlerType" minOccurs="0" />
             <xs:element name="PropertyTypeCoercion" type="PropertyIssueHandlerType" minOccurs="0" />
             <xs:element name="RawObjectIteration" type="IssueHandlerType" minOccurs="0" />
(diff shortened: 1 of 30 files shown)


weirdan:
Assuming I have a method in a library that is not referenced by the library itself, but is intended to be consumed by the library users, would it suggest to make that method void?

orklah:
Yeah, that's pretty cool! PHPStorm has an inspection that does this but only for private methods.

I checked the implementation quickly but I didn't see an exception for fluent classes. It's one of the issues of that inspection in PHPStorm.
Also you might want not to report that when the method is respecting a contract from an interface or a parent class.

@weirdan For libraries this is indeed an issue, but the same can be said for `UnusedMethod` when the library is not tested or the tests are not in the analyzed directories

muglug:
@weirdan this only triggers if the method is detected as having been used somewhere

muglug:
> I checked the implementation quickly but I didn't see an exception for fluent classes.

Added here: https://github.com/vimeo/psalm/pull/5917/commits/543d14d6269a49493b63e2a1e46ba0fbf9c21f8a

weirdan:
> this only triggers if the method is detected as having been used somewhere

Brill, I think that solves most false positives.

> Also you might not want to report that when the method is respecting a contract from an interface or a parent class.

That's a good point. It should behave similar to how unused parameters are currently treated (they are exempt from unusedness checks if dictated by parent class / interface).

Speaking of interfaces, does it also flag unused return values of interface methods? E.g.

```php
<?php

interface I {
    public function work(): bool;
}

function f(I $worker): void {
    $worker->work();
}
```
(assuming `I::work()` is not referenced anywhere else)

muglug:
> does it also flag unused return values of interface methods? E.g.

Yeah that's flagged

muglug:
@orklah I've merged this into master now. If a bunch of classes implement an interface and the interface return value is not used, only the interface is flagged: https://psalm.dev/r/5b81f6ce2a

PR #5927, opened 2021-06-12
Automate dependency updates with Dependabot

weirdan:
Unusable until dependabot/dependabot-core#3888 is fixed

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - directory: /
+    package-ecosystem: composer
+    versioning-strategy: widen
+    schedule:
+      interval: weekly
+      day: friday
+      timezone: UTC
+      time: "22:00"


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - directory: /
+    package-ecosystem: composer
+    versioning-strategy: widen
+    schedule:
+      interval: weekly
+      day: friday
+      timezone: UTC
+      time: "22:00"
```
weirdan, line 10: Please suggest a schedule most convenient to you

orklah:
Oh, I missed that one! 

Seems like your issue in dependabot is still open but the one you flagged as duplicate was closed. Can it be used?

orklah:
Apparently, renovate seems to be cool: https://github.com/renovatebot/renovate, maybe we wouldn't have the same bug?

PR #5930, opened 2021-06-13
infer non-empty-array from count() === 0

### LINKED ISSUE #5869: Recognition of non-empty-array by count()
Recognition of non-empty-array by count()

Even after the conditional branch `count($a) === 0` $a is recognized as an array.
Shouldn't it be recognized as a non-empty-array?

`count($a) ! == 0`, `$a === []` seems to be recognized correctly.

https://psalm.dev/r/38efb9b0b9

orklah:
This fixes #5869

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -374,6 +374,17 @@ class AssertionFinder
             return [];
         }
 
+        $getclass_position = self::hasGetClassCheck($conditional, $source);
+
+        if ($getclass_position) {
+            return self::getGetclassEqualityAssertions(
+                $conditional,
+                $this_class_name,
+                $source,
+                $getclass_position
+            );
+        }
+
         $typed_value_position = self::hasTypedValueComparison($conditional, $source);
 
         //we check for paradoxical assertions now, we'll use $typed_value_position later if not handled by CountEquality
@@ -439,16 +450,7 @@ class AssertionFinder
             return $if_types ? [$if_types] : [];
         }
 
-        $getclass_position = self::hasGetClassCheck($conditional, $source);
 
-        if ($getclass_position) {
-            return self::getGetclassEqualityAssertions(
-                $conditional,
-                $this_class_name,
-                $source,
-                $getclass_position
-            );
-        }
 
         if ($typed_value_position) {
             return self::getTypedValueEqualityAssertions(
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -2555,21 +2555,22 @@ class AssertionFinder
             throw new \UnexpectedValueException('$typed_value_position value');
         }
 
-        if ($var_name && $var_type) {
-            $not_identical = $conditional instanceof PhpParser\Node\Expr\BinaryOp\NotIdentical
-                || ($other_type
-                    && (($var_type->isString() && $other_type->isString())
-                        || ($var_type->isInt() && $other_type->isInt())
-                        || ($var_type->isFloat() && $other_type->isFloat())
-                    )
-                );
+        if ($var_type) {
+            if ($var_name) {
+                $not_identical = $conditional instanceof PhpParser\Node\Expr\BinaryOp\NotIdentical
+                    || ($other_type
+                        && (($var_type->isString() && $other_type->isString())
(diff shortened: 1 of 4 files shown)


orklah:
Ok so the failing test is due to the new `hasCountEqualityCheck` in `scrapeEqualityAssertions` being above the `hasTypedValueComparison` which emits the issue

However I hesitate to change that because in `scrapeInequalityAssertions`, `hasCountEqualityCheck` is above `hasTypedValueComparison` and I find it weird to have two different orders

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -1496,11 +1527,9 @@ protected static function hasLessThanCountEqualityCheck(
     }
 
     /**
-     * @param  PhpParser\Node\Expr\BinaryOp\NotIdentical|PhpParser\Node\Expr\BinaryOp $conditional
```
weirdan, line 1499: Should it actually work for any binary op? E.g. `count($a) + 3`?

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -533,7 +564,7 @@ private static function scrapeInequalityAssertions(
         }
 
         $count = null;
-        $count_inequality_position = self::hasNotCountEqualityCheck($conditional, $count);
+        $count_inequality_position = self::hasCountEqualityCheck($conditional, $count);
```
weirdan, line 567: The var (or method) naming here is quite confusing

orklah:
I converted to draft because I discovered `hasNonEmptyCountEqualityCheck` in scrapeEquality which is absent in scrapeInequality and which should probably have handled the case I fixed, I'll take the time to review why there are two different methods

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
orklah: It's only called with Equal, Identical, NotEqual and NotIdentical but yeah, I'll add this in doc. The old phpdoc was weird so I just deleted, I should have narrowed

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
orklah: Well, the method in itself doesn't check for identical/equal nor NotIdentical/NotEqual but in scrapeEqualityAssertions we check equality and in scrapeInequalityAssertions we check inequality. Do you have a better suggestion?

weirdan: I think it can stay as is, with the updated docblock.

orklah:
Ok so I removed hasNonEmptyCountEqualityCheck  from scrapeEquality: it was not symmetrical with scrapeInequality and it didn't work properly anyway.

However, one test still fail because `hasCountEqualityCheck` handle the assertions before `hasTypedValueComparison` has time to and it doesn't emit the Error anymore.

I tried extracting the code that checks the two inferred types (which is repeated in the two scrape methods) to be able to call it in `hasCountEqualityCheck` but I'm in a slippery slope where Psalm now find me an issue in `src/Psalm/Type/Union.php:495:38` because it thinks `$types` with the type `non-empty-array<string, Atomic>` + `isset($types['null']` is now an array with 1 element and so it can't ever be `count($types) === 2`.

Should I skip the test? Should I try to carry on and modify what it takes to make tests happy? Should I trash this PR? 😄

VincentLanglet:
> Should I trash this PR? 😄

Please no ! It would be great to infer non-empty when checking for a count > 0.

orklah:
> Please no ! It would be great to infer non-empty when checking for a count > 0.

Note that count > 0 should already be supported. The only fix here is count() === 0

VincentLanglet:
> > Please no ! It would be great to infer non-empty when checking for a count > 0.
> 
> Note that count > 0 should already be supported. The only fix here is count() ==(=) 0

Yes indeed.

The issue I have is
```
if (count($foo) === 0) {
    return;
}

// Here it should be non-empty
....
```

orklah:
In the meantime, I recommend $a === [], which is supported too ;)

orklah:
Ok, so tests pass because I skipped one

I can't figure a way to both add my assertions and let Psalm go into `getTypedValueEqualityAssertions` in order to emit the issue while keeping some symmetry between scrapes methods.

orklah:
This is completely broken now. I'll start again from scratch if I try to tackle this again

PR #5931, opened 2021-06-14
Changed some confusing comments and docblocks

elnoro:
Hi, I noticed some inconsistency in the comments in plugin examples. It might be a bit confusing for newcomers. 

1. FunctionCaseChecker description refers to float assignment.
2. PHPDocs define types for arguments that had already been removed here: https://github.com/vimeo/psalm/pull/4881/files

### DIFF
--- a/examples/plugins/StringChecker.php
+++ b/examples/plugins/StringChecker.php
@@ -14,8 +14,6 @@ class StringChecker implements AfterExpressionAnalysisInterface
     /**
      * Called after an expression has been checked
      *
-     * @param  FileManipulation[]   $file_replacements
-     *
      * @return null|false
      */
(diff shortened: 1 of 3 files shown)


orklah (approved):

examples/plugins/StringChecker.php
```diff
@@ -13,10 +13,6 @@ class StringChecker implements AfterExpressionAnalysisInterface
 {
     /**
      * Called after an expression has been checked
-     *
-     * @param  FileManipulation[]   $file_replacements
-     *
-     * @return null|false
```
orklah, line 19: You can leave null|false, it's more narrow than ?bool

elnoro: Thanks for pointing this out. Left null|false here and in the other file

muglug:
Thanks!

Issue #5929, opened 2021-06-12
Emit a different subtype of UnusedVariable when it originates from foreach value


weirdan:
https://psalm.dev/r/fce850227d 

Here `UnusedVariable` is reported as fixable, but Psalm cannot fix it as it's syntactically required.
I propose to emit `UnusedForeachValue` here. It has the following advantages:

1. Can be separately suppressed
2. Does not require changing the variable name (we've been recommending to prefix unused but required variables with underscore)
3. Will not be reported as fixable

weirdan:
/cc: @VincentLanglet - I think this is the way #5493 can be solved.

PR #5932, opened 2021-06-15
Emit separate type of issue when foreach value is unused

weirdan:
Fixes vimeo/psalm#5929

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -452,6 +452,7 @@
             <xs:element name="UnusedClass" type="ClassIssueHandlerType" minOccurs="0" />
             <xs:element name="UnusedClosureParam" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="UnusedConstructor" type="MethodIssueHandlerType" minOccurs="0" />
+            <xs:element name="UnusedForeachValue" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="UnusedFunctionCall" type="FunctionIssueHandlerType" minOccurs="0" />
             <xs:element name="UnusedMethod" type="MethodIssueHandlerType" minOccurs="0" />
             <xs:element name="UnusedMethodCall" type="MethodIssueHandlerType" minOccurs="0" />
--- a/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
@@ -93,7 +93,7 @@ class ForeachAnalyzer
 
         if ($stmt->valueVar instanceof PhpParser\Node\Expr\Variable && is_string($stmt->valueVar->name)) {
             $safe_var_ids['$' . $stmt->valueVar->name] = true;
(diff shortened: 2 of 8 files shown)


src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
```diff
@@ -769,12 +775,32 @@ public function checkUnreferencedVars(array $stmts): void
                 && $this->data_flow_graph instanceof VariableUseGraph
                 && !$this->data_flow_graph->isVariableUsed($assignment_node)
             ) {
-                $issue = new UnusedVariable(
-                    $var_id . ' is never referenced or the value is not used',
-                    $original_location
+                $original_foreach_stmt = NodeFinder::find(
```
muglug, line 778: My worry here is that this is a bunch of unnecessary work for codebases that have grandfathered-in issues on each Psalm run.

What if the `$statements_analyzer` (created afresh for each function analysis) held a list of foreach variable positions? That way the comparison wouldn't have to loop recurse the AST.

config.xsd
```diff
@@ -452,6 +452,7 @@
             <xs:element name="UnusedClass" type="ClassIssueHandlerType" minOccurs="0" />
             <xs:element name="UnusedClosureParam" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="UnusedConstructor" type="MethodIssueHandlerType" minOccurs="0" />
+            <xs:element name="UnusedForeachValue" type="IssueHandlerType" minOccurs="0"></xs:element>
```
orklah, line 455: maybe keep the same  convention of self closing tags

weirdan: oops

src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
weirdan: Yeah, that would work and is actually easy to implement by storing `$stmt->valueVar` locations in `ForeachAnalyzer::analyze()` into `$statements_analyzer`.

weirdan:
Note that as there's a new class (issue) added this requires a new minor (4.8) release.

src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php
```diff
@@ -93,6 +93,10 @@ public static function analyze(
 
         if ($stmt->valueVar instanceof PhpParser\Node\Expr\Variable && is_string($stmt->valueVar->name)) {
             $safe_var_ids['$' . $stmt->valueVar->name] = true;
+            $statements_analyzer->foreach_var_locations['$' . strtolower($stmt->valueVar->name)][] = new CodeLocation(
```
muglug, line 96: why are you normalising the variable name? Variables in PHP are case-sensitive.

weirdan: Good point.

weirdan: Fixed.

muglug:
Thanks! The next release will be 4.8.0, as there'll be a couple of new features to talk about.

PR #5936, opened 2021-06-15
Remove redundant namespaced phpdoc

orklah:
this PR reduce the verbosity of psalm when it doesn't add value
it reduces and `@psalm-return array<array-key, mixed>` to just `@psalm-return array`.

I wanted to push further and remove the whole @psalm-return but the refactor tool depend on it. I also wanted to go further by removing `array<string>` because phpdoc is enough with `string[]` but I couldn't find a good way to do that, It would have needed a `canBeFullyExpressedInPhpdoc` method that doesn't exist and I'm not sure it's worth adding it for now

### DIFF
--- a/src/Psalm/Type/Atomic/GenericTrait.php
+++ b/src/Psalm/Type/Atomic/GenericTrait.php
@@ -100,6 +100,17 @@ trait GenericTrait
             return $value_type_string . '[]';
         }
 
+        $type_params = $this->type_params;
+
+        //no need for special format if the key is not determined and the content can be displayed with phpdoc format
+        if ($this instanceof TArray &&
+            isset($type_params[0]) &&
+            $type_params[0]->isArrayKey()
+        ) {
+            //we remove the key for display
+            unset($type_params[0]);
+        }
+
         $extra_types = '';
 
         if ($this instanceof TNamedObject && $this->extra_types) {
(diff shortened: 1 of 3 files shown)


muglug:
Thanks!

weirdan:
Funny thing, I would want Psalm to flag `array`s without type parameters in docblocks as they are not specific enough to be useful.

orklah:
> Funny thing, I would want Psalm to flag `array`s without type parameters in docblocks as they are not specific enough to be useful.

My issue was this one:

```php
function a() {
    return returnArray();
}
```

If you ask Psalm to add the return type, it will transform it to:
```php
/**
 * @return array
 *
 * @psalm-return array<array-key, mixed>
 */
function a(): array {
    return returnArray();
}
```

It's just 3 times the same type. I can easily use linting rules to remove `@return array` from the phpdoc as redundant, but `@psalm-return array<array-key, mixed>` is another story.

But it gets worse, we have a big legacy codebase, and when psalm emits that sort of thing:
```php
/**
 * @return ((array|int|string|bool)[]|array)[]
 *
 * @psalm-return array<array-key, array<array-key, mixed>|int|string|bool>|array<array-key, mixed>
 */
function a(): array {
    return returnComplexStructure();
}
```

and you have to compare this to the previous type that was just slightly different, you have sudden urges to strangle someone.

I agree `array` is not sufficient in itself, but `@psalm-return array<array-key, mixed>` is no better than `@return mixed[]` so why keep the most verbose syntax?

weirdan:
I wonder whether it's possible to refuse adding `@return array` where no type parameters could be inferred. Like 'Yes, we inferred that this method returns some unspecified array and we won't add it automatically, but you can do that manually'.

Issue #5942, opened 2021-06-16
InvalidArrayOffset accessing arrays with property defined using constants keys


flaviovs:
Change `Foo::C_*` to `int` on the `@property` line and the error goes away.

See: https://psalm.dev/r/2e8b89f692

_EDIT: remove code/error already added by GH bot_

PR #5943, opened 2021-06-16
Allow class constant as offset key

orklah:
This fixes https://github.com/vimeo/psalm/issues/5942

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
@@ -14,6 +14,7 @@ use Psalm\Internal\Type\Comparator\UnionTypeComparator;
 use Psalm\Internal\Type\TemplateInferredTypeReplacer;
 use Psalm\Internal\Type\TemplateResult;
 use Psalm\Internal\Type\TypeCombiner;
+use Psalm\Internal\Type\TypeExpander;
 use Psalm\Issue\EmptyArrayAccess;
 use Psalm\Issue\InvalidArrayAccess;
 use Psalm\Issue\InvalidArrayAssignment;
@@ -42,6 +43,7 @@ use Psalm\Node\VirtualIdentifier;
 use Psalm\Node\VirtualName;
 use Psalm\Plugin\EventHandler\Event\AddRemoveTaintsEvent;
 use Psalm\Type;
+use Psalm\Type\Atomic;
 use Psalm\Type\Atomic\TArray;
 use Psalm\Type\Atomic\TArrayKey;
 use Psalm\Type\Atomic\TClassStringMap;
@@ -450,7 +452,6 @@ class ArrayFetchAnalyzer
         $non_array_types = [];
 
         $has_valid_expected_offset = false;
-        $has_valid_absolute_offset = false;
         $expected_offset_types = [];
 
         $key_values = [];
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
```diff
@@ -2069,4 +2053,61 @@ private static function handleArrayAccessOnString(
...
+                //do nothing
+            } elseif ($atomic_offset_type instanceof Type\Atomic\TNull &&
+                $offset_type->ignore_nullable_issues === true
+            ) {
+                //do nothing
+            } elseif ($atomic_offset_type instanceof Type\Atomic\TString ||
+                $atomic_offset_type instanceof Type\Atomic\TInt ||
+                $atomic_offset_type instanceof Type\Atomic\TArrayKey ||
+                $atomic_offset_type instanceof Type\Atomic\TMixed ||
+                $atomic_offset_type instanceof Type\Atomic\TTemplateParam ||
```
weirdan, line 2101: Shouldn't `$atomic_offset_type->as` be checked here as well? E.g. `@template T of array` and using `T` as an offset is probably wrong.

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
```diff
@@ -2069,4 +2053,61 @@ private static function handleArrayAccessOnString(
...
+                (
+                    $atomic_offset_type instanceof Type\Atomic\TObjectWithProperties
+                    && isset($atomic_offset_type->methods['__toString'])
```
weirdan, line 2104: What about `TNamedObject` with a `__toString()` method?

orklah: fixed

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
orklah: I'm adding for the sake of completeness, however, offset types are checked twice: 
- once against `array-key`
- once against the expected offset type (which could be anything because psalm will technically register a `array<array, string>`

The method you flagged is the first former. This means that you would need to have a source of `array<array, string>` somewhere to be able to give an array offset type to psalm without error.

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
orklah: Well, turns out it's quite easy to add in check against array-key, but really hard against expected type.

It's really deep in the code and I didn't manage to make it work. If anyone wants to try, here's a reproducer https://psalm.dev/r/09a9137459

orklah: oh well, turns out it's not even valid:
https://3v4l.org/PC88H

weirdan: Ah, I assumed it was valid because there was exiting code that allowed `object{__toString()}`. Array keys are really inconsistent in PHP :roll_eyes:

weirdan (approved):
:+1:

muglug:
Thanks!

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
orklah: Yeah, existing code was mine from a few weeks ago, I didn't check at the time...

PR #5950, opened 2021-06-18
Fix wrong rdkafka commitAsync type

danog:


danog:
Also, it seems that the `RdKafka\Consumer::newTopic` and `RdKafka\Producer::newTopic` return types aren't actually used during type resolution: 

```
<?php

/** @var \RdKafka\Consumer $a */
$a = null;

$topic = $a->newTopic('test', null);
/** @psalm-trace $topic */;
```

Result:
```
INFO: Trace - a.php:7:27 - $topic: RdKafka\Topic (see https://psalm.dev/224)
/** @psalm-trace $topic */;
```

Should be:
```
INFO: Trace - a.php:7:27 - $topic: RdKafka\ConsumerTopic (see https://psalm.dev/224)
/** @psalm-trace $topic */;
```

muglug:
Thanks!

Issue #5960, opened 2021-06-20
Unable to create baseline in psalm 4.8.0


NoodlesNZ:
I upgraded a project to psalm 4.8.0 this morning and tried to create a baseline (using `./vendor/bin/psalm --set-baseline=site-baseline.xml`) and it give me this:

```
Writing error baseline to file...
Uncaught Psalm\Exception\ConfigException: site-baseline.xml does not exist or is not readable in /Users/noodles/Developer/web/site/vendor/vimeo/psalm/src/Psalm/ErrorBaseline.php:78
Stack trace:
#0 /Users/noodles/Developer/web/site/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(623): Psalm\ErrorBaseline::read(Object(Psalm\Internal\Provider\FileProvider), 'site-baseline.xm...')
#1 /Users/noodles/Developer/web/site/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(1002): Psalm\Internal\Cli\Psalm::generateBaseline(Array, Object(Psalm\Config), '/Users/noodles/...', NULL)
#2 /Users/noodles/Developer/web/site/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(386): Psalm\Internal\Cli\Psalm::initBaseline(Array, Object(Psalm\Config), '/Users/noodles/...', NULL)
#3 /Users/noodles/Developer/web/site/vendor/vimeo/psalm/psalm(4): Psalm\Internal\Cli\Psalm::run(Array)
#4 {main}
(Psalm 4.8.0@0a57c865499049527b3e1c47f14abea761d94b83 crashed due to an uncaught Throwable)
```

Downgrading to psalm 4.7.2 and re-running the command seemed to work.

weirdan:
Confirmed locally on master

weirdan:
As a workaround, you can generate it using an older version and then re-generate it with 4.8

NoodlesNZ:
Thanks @muglug and @weirdan for your quick responses and resolution :)

muglug:
@weirdan any reason why you reopened this?

weirdan:
@muglug yep, I think the fix provided in the linked PR is better.

PR #5961, opened 2021-06-20
Don't attempt to read baseline before it was generated

weirdan:
Fixes vimeo/psalm#5960

Previous fix would not return generated baseline.

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -618,21 +618,22 @@ final class Psalm
     ): array {
         fwrite(STDERR, 'Writing error baseline to file...' . PHP_EOL);
 
-        try {
-            $issue_baseline = ErrorBaseline::read(
-                new \Psalm\Internal\Provider\FileProvider,
-                $options['set-baseline']
-            );
-        } catch (\Psalm\Exception\ConfigException $e) {
-            $issue_baseline = [];
-        }
-
         ErrorBaseline::create(
             new \Psalm\Internal\Provider\FileProvider,
             $options['set-baseline'],
             IssueBuffer::getIssuesData(),
             $config->include_php_versions_in_error_baseline || isset($options['include-php-versions'])
         );
+
+        try {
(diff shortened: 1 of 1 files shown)


muglug:
Sorry, mind running through what steps weren't happening before that are now?

weirdan:
4.7.3 would write baseline, read it back and then use in `IssueBuffer::finish()`. I'm checking whether that was actually required.

weirdan:
@muglug this prevents a cosmetic issue where issues added to the baseline during current run are still reported:

image

muglug:
Thanks! Not sufficient for a separate release, but this is great.

orklah:
@weirdan This cosmetic issue is exactly what I was trying to achieve in #5838 after discussing it in #5831 :)

When running with a baseline, it become impossible to know what issues have been added

On our project, Psalm takes too long to run to hope adding it in CI, so we have a setup that runs Psalm regularly and saves the errors in baseline. Each time we see new errors, we handle them and report to whoever may be concerned.

Without this, we need to run Psalm once without baseline, takes the issues and send them, then re-run Psalm with the baseline.

Can we revert this change or introduce a new option to return to the described behaviour?

weirdan:
> On our project, Psalm takes too long to run to hope adding it in CI

Even with cache enabled?

weirdan:
@orklah and I thought I botched that during refactoring, because I was comparing with last good release (4.7.3) :laughing: 

Obligatory [xkcd:1172](https://xkcd.com/1172/).

In that workflow of yours, do you output machine-readable data (e.g. `--output-format`/`--report`)?

orklah:
> Even with cache enabled?

Trying to run Psalm without `--no-cache` brings the whole server down in about 15-20 minutes (a run without cache takes 10 minutes). I never could go to the bottom of this issue (I still hope I will one day !)

> Obligatory xkcd:1172.

😄 

> In that workflow of yours, do you output machine-readable data (e.g. --output-format/--report)?

I don't remember the option we use, but we also retrieve in a file the whole list of issues reported by Psalm but the baseline is ignored when doing that so we retrieve every issue

PR #5964, opened 2021-06-21
Make producev parameters optional

danog:


dictionaries/CallMap.php
```diff
@@ -10608,7 +10608,7 @@
 'RdKafka\ProducerTopic::__construct' => ['void'],
 'RdKafka\ProducerTopic::getName' => ['string'],
 'RdKafka\ProducerTopic::produce' => ['void', 'partition'=>'int', 'msgflags'=>'int', 'payload'=>'string', 'key='=>'?string'],
-'RdKafka\ProducerTopic::producev' => ['void', 'partition'=>'int', 'msgflags'=>'int', 'payload'=>'string', 'key='=>'?string', 'headers'=>'?array<string, string>', 'timestamp_ms'=>'?int'],
+'RdKafka\ProducerTopic::producev' => ['void', 'partition'=>'int', 'msgflags'=>'int', 'payload'=>'string', 'key='=>'?string', 'headers='=>'?array<string, string>', 'timestamp_ms='=>'?int', 'opaque='=>'?string'],
```
weirdan, line 10611: `$opaque` is only available with RdKafka 5.0, right? This raises an interesting question that went largely ignored before: should we split callmaps for non-core extensions out into their own files and version them based on the extension version?

PECL extensions have their own version cycle, not bound to PHP release lifecycle. E.g. RdKafka 4.1 supports PHP 5.6 - 7.*, while 5.0 supports an overlapping range of 7.* - 8.*. So when running on 7.4 you could have either 4.1 or 5.0, but we're forced to choose a single signature to put into 7.4 callmap.

@muglug @orklah thoughts?

orklah: If we need to go this far, I guess it would be better to leverage composer and just create stubs plugins for ext libs based on dependancies in composer.json

muglug: I've always felt that the callmap should represent the latest major version of a given dependency. Splintering into separate callmaps for separate extensions would be a maintenance nightmare IMO.

weirdan: Is there actually that much difference between having deltas for N PHP versions and having deltas for N extension versions?

weirdan:
LGTM (as there's no perfect solution for extension versioning issue at the moment).

Issue #5983, opened 2021-06-22
Non-existing class crashes in ForeachAnalyzer


ohader:
https://psalm.dev/r/78e3962d20

Real world example: https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Cache/Traits/RedisTrait.php#L548-L553
> Uncaught InvalidArgumentException: Could not get class storage for predis\connection\aggregate\clusterinterface in /vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:43

Triggered by:
https://github.com/vimeo/psalm/blob/6abce3525ac34f815aaa3b7953739b744ca5c39e/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php#L766-L776

Potential fix (wrap in `try/catch`, return `false` in case subject class does not exist):
https://github.com/vimeo/psalm/blob/6abce3525ac34f815aaa3b7953739b744ca5c39e/src/Psalm/Internal/Codebase/ClassLikes.php#L673

weirdan:
Simplified a bit: https://psalm.dev/r/5671873d21

Requires all of the following:

1. Taint analysis mode
2. Missing class
3. Intersection with `Traversable`

PR #5984, opened 2021-06-22
[BUGFIX] Avoid crash during interface checks on non-existing classes

ohader:
`ClassLikes::classImplements(NonExisting::class, WellKnown::class)`
now returns `false` instead of throwing an exception for non-existing
classes.

Fixes: #5983


muglug (approved):

muglug:
Thanks!

Issue #5975, opened 2021-06-22
PossiblyUnusedReturnValue false-positive with throw


enumag:
https://psalm.dev/r/844d55abf9

PR #5989, opened 2021-06-23
Mark return values in `throw` arguments as used

weirdan:
Fixes vimeo/psalm#5975

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
@@ -140,7 +140,7 @@ class ArgumentAnalyzer
             && !$arg->value instanceof PhpParser\Node\Expr\ConstFetch
         ) {
             $values = \preg_split('//u', $arg_value_type->getSingleStringLiteral()->value, -1, \PREG_SPLIT_NO_EMPTY);
-            
+
             if ($values !== false) {
                 $prev_ord = 0;
 
@@ -940,6 +940,7 @@ class ArgumentAnalyzer
                         || $context->inside_use
                         || $context->inside_assignment
                         || $context->inside_conditional
+                        || $context->inside_throw
                 );
             }
         }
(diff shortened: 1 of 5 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
```diff
@@ -940,6 +940,7 @@ public static function verifyType(
                         || $context->inside_use
                         || $context->inside_assignment
                         || $context->inside_conditional
+                        || $context->inside_throw
```
weirdan, line 943: I'm not sure whether I really needed to add `$context->inside_throw` to _all_ of these places, but no tests broke :man_shrugging:

muglug:
Thanks!

frederikbosch:
[This method](https://github.com/moneyphp/money/blob/master/src/Calculator/BcMathCalculator.php#L50) still errors. Why?
```php
    /** @psalm-pure */
    public static function divide(string $amount, string $divisor): string
    {
        if (bccomp($divisor, '0', self::SCALE) === 0) {
            throw InvalidArgumentException::divisionByZero();
        }

        return bcdiv($amount, $divisor, self::SCALE);
    }
```

`
ERROR: PossiblyUnusedReturnValue - src/Calculator/BcMathCalculator.php:50:69 - The return value for this method is never used (see https://psalm.dev/273)
    public static function divide(string $amount, string $divisor): string
`

weirdan:
@frederikbosch is it used, and if so, where?

orklah:
It seems used at least here: https://github.com/moneyphp/money/blob/2f882263c73d75ddbea49244da5425a9c18289fb/src/Money.php#L265

Also in tests: https://github.com/moneyphp/money/blob/4f9c4fa10f7615a5d7f0c25de6658e0479957be9/tests/Calculator/GmpCalculatorTest.php#L55

and I checked, the `tests` directory is included in Psalm's analysis

frederikbosch:
@orklah See #6021

PR #6000, opened 2021-06-26
Ignore line endings & whitespace in baseline code snippets

### LINKED ISSUE #5979: Ignore line ending characters when matching code issues against baseline
I was puzzling over why my Psalm run was passing locally, but had many errors in CI. The answer was that I had autocrlf turned on in git settings, so my local PHP files were in a different format to those generated in the CI environment. Specifically I had many `<code>` elements in the baseline that ended with `&#13;</code>`

I wonder if it would be worth making Psalm ignore the `&#13;` (carriage return), and also not add it when generating the baseline.

bdsl:
This change makes Psalm trim code snippets read from the baseline, and the snippets from issues found in the current analysis, so that the comparison won't care whether either of them has any carriage return character, or any other type of white space at the start or end.

Should fix #5979


bdsl:
I have tested this manually by applying this patch - the current version of Psalm reports an Error when run on itself with the patch applied. The version from this PR reports 'no errors found'.

I don't think this needs an automated end-to-end test.


`0001-add-carrige-return-to-issue-in-baseline.patch`
```patch
From b0dcbef6bbdfca1b0017002678cefbb3750f94f9 Mon Sep 17 00:00:00 2001
From: Barney Laurance <barney@redmagic.org.uk>
Date: Sat, 26 Jun 2021 22:33:25 +0100
Subject: [PATCH] add-carrige-return-to-issue-in-baseline

---
 psalm-baseline.xml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/psalm-baseline.xml b/psalm-baseline.xml
index 67e00699f..687294a95 100644
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -2,7 +2,7 @@
 <files psalm-version="dev-master@5e8219b613d9f97cd7eeba16e1fea75cc8a808e6">
   <file src="examples/TemplateChecker.php">
     <PossiblyUndefinedIntArrayOffset occurrences="2">
-      <code>$comment_block-&gt;tags['variablesfrom'][0]</code>
+      <code>$comment_block-&gt;tags['variablesfrom'][0]&#13;</code>
       <code>$matches[1]</code>
   </file>
-- 
2.30.2


```

bdsl:
6k !!!!

tests/IssueBufferTest.php
```diff
@@ -54,11 +54,31 @@ public function testFinishDoesNotCorruptInternalState(): void
...
+                new \Psalm\Internal\Analyzer\IssueData(
+                    "error",
+                    0,
+                    0,
+                    "MissingPropertyType",
+                    'Message',
+                    "three.php",
+                    "/path/three.php",
+                    "snippet-3-has-carriage-return" . "\r",
+                    "snippet-3-has-carriage-return" . "\r",
```
bdsl, line 68:
```suggestion
                    "snippet-3-has-carriage-return\r",
                    "snippet-3-has-carriage-return\r",
```

src/Psalm/ErrorBaseline.php
```diff
@@ -273,6 +274,10 @@ function (string $extension) : string {
                 foreach ($existingIssueType['s'] as $selection) {
                     $codeNode = $baselineDoc->createElement('code');
 
+                    /** @todo in major version release (e.g. 5.0.0) replace $selection with trim($selection)
```
weirdan, line 277:
```suggestion
                    /** @todo in major version release (e.g. Psalm 5) replace $selection with trim($selection)
```
We've been marking those things with 'Psalm 5' so that when the time comes they could all be uncovered with a single `grep 'Psalm 5'` run.

src/Psalm/Internal/Analyzer/IssueData.php
```diff
@@ -154,7 +156,7 @@ public function __construct(
         $this->file_name = $file_name;
         $this->file_path = $file_path;
         $this->snippet = $snippet;
-        $this->selected_text = $selected_text;
+        $this->selected_text = trim($selected_text);
```
weirdan, line 159: Trimming here may have far reaching effects - everything that operates on `IssueData` entries is potentially affected, including AfterAnalysis plugins. Could it perhaps be safer to move the trim to where issue is compared to the issue from baseline?

bdsl: Yes, that seems safer - I've moved the `trim`. Tests still pass, including the manual test with applying the patch file. Thanks.

weirdan (approved):

weirdan:
Thanks!

PR #6010, opened 2021-06-28
Fixes #6008, Enabled onSave and onOpen textDocument

### LINKED ISSUE #6008: Language Server textDocumentSync are limited
Is there a reason the language server only sync's change events? Looking at https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/LanguageServer/Server/TextDocument.php I can see the code exists to allow openClose and save

However only `change` is declared as an ability (https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/LanguageServer/LanguageServer.php#L231). This seems to have been the case from the first day this was added?

Since only change is supported static analysis is not run on an open file until a user changes something in that document (or alternatively hover is triggered).

This wouldn't be an issue because maybe you'd only want analysis to run if the file is modified but that breaks because it runs anyways whenever any text is hovered over so why not just enable openClose

Can we enable the other methods? 

```
        "textDocumentSync": {
            "openClose": null,
            "change": 1,
            "willSave": null,
            "willSaveWaitUntil": null,
            "save": null
        }
```

I will submit a PR for this in a moment

tm1000:
Fixes #6008

This enables the already written textDocument methods of `didOpen`, `didClose`,  and `didSave` which means that once the file is opened the LanguageServer is immediately notified.

This allows the document to almost instantaneously show markup (as before it needed a change event for it to function)

These methods were already written but the Language Server never properly notified the client of such. 

Additionally I changed `verboseLog` to a public method so that it can be called from inside textDocument so that I could replace error_log with verboseLog as per the spec (error_log simply outputted right to STDOUT which violates the LSP specification)


src/Psalm/Internal/LanguageServer/Server/TextDocument.php
```diff
@@ -286,6 +299,14 @@ public function completion(TextDocumentIdentifier $textDocument, Position $posit
         return new Success(new CompletionList($completion_items, false));
     }
 
+    /**
+     * The signature help request is sent from the client to the server to request signature
+     * information at a given cursor position.
+     *
+     * @param TextDocumentIdentifier $textDocument
+     * @param Position $position
+     * @return Promise
```
weirdan, line 308:
```suggestion
```
This doesn't add any additional information that was not provided by the function signature alread.

tm1000: Heh. You remind me of my coworker (not a bad thing!). He hates superfluous docblock types! I will remember that for the future :-)

weirdan:
Thanks!

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
weirdan: They tend to get out of sync, and Psalm doesn't flag `@param` entries for params that got removed. Speaking of which, I should probably file that as an issue, as they don't seem to be used in any way.

PR #6014, opened 2021-06-29
Fixes #6013, implement workspace/didChangeWatchedFiles

### LINKED ISSUE #6013: Implement workspace/didChangeWatchedFiles
Implement `workspace/didChangeWatchedFiles` https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didChangeWatchedFiles 

This will resolve https://github.com/psalm/psalm-vscode-plugin/issues/32 because the event will be triggered when files change on the filesystem

tm1000:
This implements `workspace/didChangeWatchedFiles`

We also check to make sure the watch file that was changed is not currently open so this technically depends on #6010 being merged otherwise the files aren't tracked and there will be conflicts

### DIFF
--- a/src/Psalm/Internal/LanguageServer/Server/Workspace.php
+++ b/src/Psalm/Internal/LanguageServer/Server/Workspace.php
@@ -1,0 +1,73 @@
+<?php
+declare(strict_types = 1);
+namespace Psalm\Internal\LanguageServer\Server;
+
+use LanguageServerProtocol\FileChangeType;
+use LanguageServerProtocol\FileEvent;
+use Psalm\Codebase;
+use Psalm\Internal\LanguageServer\LanguageServer;
+
+/**
+ * Provides method handlers for all textDocument/* methods
+ */
+class Workspace
+{
+    /**
+     * @var LanguageServer
+     */
+    protected $server;
+
+    /**
+     * @var Codebase
+     */
+    protected $codebase;
+
+    /** @var ?int */
+    protected $onchange_line_limit;
+
+    public function __construct(
+        LanguageServer $server,
+        Codebase $codebase,
+        ?int $onchange_line_limit
+    ) {
+        $this->server = $server;
+        $this->codebase = $codebase;
+        $this->onchange_line_limit = $onchange_line_limit;
+    }
+
+    /**
+     * The watched files notification is sent from the client to the server when the client
+     * detects changes to files and folders watched by the language client (note although
+     * the name suggest that only file events are sent it is about file system events
+     * which include folders as well). It is recommended that servers register for these
+     * file system events using the registration mechanism. In former implementations clients
+     * pushed file events without the server actively asking for it.
+     *
+     * @param FileEvent[] $changes
+     * @psalm-suppress PossiblyUnusedMethod
+     */
+    public function didChangeWatchedFiles(array $changes): void
+    {
+        foreach ($changes as $change) {
+            $file_path = LanguageServer::uriToPath($change->uri);
+
+            if ($change->type === FileChangeType::DELETED) {
+                $this->codebase->invalidateInformationForFile($file_path);
+                return;
+            }
+
+            if (!$this->codebase->config->isInProjectDirs($file_path)) {
+                return;
+            }
+
+            if ($this->onchange_line_limit === 0) {
+                return;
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/LanguageServer/Server/Workspace.php
```diff
@@ -0,0 +1,73 @@
...
+            if (!$this->codebase->config->isInProjectDirs($file_path)) {
+                return;
+            }
+
+            if ($this->onchange_line_limit === 0) {
+                return;
+            }
+
+            //If the file is currently open then dont analyse it because its tracked by the client
+            if (!$this->codebase->file_provider->isOpen($file_path)) {
```
tm1000, line 68: You know I wonder if this should be

```
if ($this->codebase->file_provider->isOpen($file_path)) {
    return;
}
```

src/Psalm/Internal/LanguageServer/Server/Workspace.php
```diff
@@ -0,0 +1,73 @@
...
+        $this->onchange_line_limit = $onchange_line_limit;
+    }
+
+    /**
+     * The watched files notification is sent from the client to the server when the client
+     * detects changes to files and folders watched by the language client (note although
+     * the name suggest that only file events are sent it is about file system events
+     * which include folders as well). It is recommended that servers register for these
+     * file system events using the registration mechanism. In former implementations clients
+     * pushed file events without the server actively asking for it.
```
weirdan, line 44: This implies server has to register files it wants to watch, but I don't see it anywhere. So I guess this may break eventually.

src/Psalm/Internal/LanguageServer/Server/Workspace.php
weirdan: Both produce same results, so it doesn't matter much

weirdan (changes requested):

src/Psalm/Internal/LanguageServer/Server/Workspace.php
tm1000: Yes. I read into that myself. Right now vscode will send the workspace changes without registering which is good

The bigger issue is that the Felix package Psalm relies on seems to be abandonware (not updated since 2018) so there's no ability to register capabilities. 

At this point it's either fork that project and have psalm reference it's own version. Or fork out only the classes Psalm needs

Besides that if the day comes and vscode stops publishing the events unless psalm registers it doesn't break anything internally at that point we will just need to work on the register call which will hopefully be done before that point

As for now vscode works as is and sends the change regardless

tm1000:
@weirdan it would still be nice to be able to get this merged in even though it will at some point in the future require a register call to function. Without the register call it won't break functionality. 

The reason is that one of the older static analysis tools for php vscode was removed by Microsoft from the marketplace because it was no longer updated and am caused stability issues. The maintainer of that is the same maintainer of the RPC and language server dependencies that Psalm itself uses. 

Namely: https://github.com/felixfbecker/php-language-server

This pr actually brings Psalm almost to feature parity with the extension that was disabled from the marketplace: https://github.com/felixfbecker/vscode-php-intellisense/issues

src/Psalm/Internal/LanguageServer/Server/Workspace.php
weirdan: With the latest release 17 days ago I wouldn't call it abandonware. It would probably suffice to craft a PR and nudge Felix to make a release.

weirdan (approved):

weirdan:
Thanks!

src/Psalm/Internal/LanguageServer/Server/Workspace.php
tm1000: I'm mainly speaking of https://github.com/felixfbecker/php-language-server/issues/807

weirdan: Yeah, but we don't use that.

tm1000: Oh yeah. Right. I don’t know why I thought that was used. The ones that are used are pretty up to date and support php 8. I’ll look into the registry calls for a future pr

Issue #6005, opened 2021-06-28
Fails to infer return type for method using static callable


leflings:
https://psalm.dev/r/2791d5c034

I would expect return type of `combineAll1` to be inferable from the context.

Oddly enough, psalm has no issue with the "callable"-form used in `combineAll2`. (I wonder if this is because Psalm just "gives up" with this form or if it actually understands it?)

I prefer (and believe is php standard?) to supply the callable in the first form `[class, method]` as this is understood by PhpStorm when finding usages and navigating to implementation.

Possible duplicate of https://github.com/vimeo/psalm/issues/5388 though from my (mostly inexperienced) perspective they seem to be two slightly different cases.

weirdan:
Simplified a bit: https://psalm.dev/r/4d5778d2aa

> I wonder if this is because Psalm just "gives up" with this form or if it actually understands it?

Psalm understand that, see the trace produced by the linked snippet.

> Possible duplicate of #5388 though from my (mostly inexperienced) perspective they seem to be two slightly different cases.

They are likely separate, because Psalm has custom code to infer `array_reduce` return type.

PR #6019, opened 2021-06-29
Possible implementation for ArrayReduceReturnTypeProvider support array callables with self and static

elnoro:
Hello! 

This should fix https://github.com/vimeo/psalm/issues/6005

I am not sure why self and static were skipped when handling callables in ArrayReduceReturnTypeProvider, so hopefully this won't break anything - I added a test for it just in case.


weirdan:
Thanks!

PR #6030, opened 2021-07-01
Drop FakeFileProvider from tests folder

weirdan:
And use Psalm\Internal\Provider\FakeFileProvider

Part 2 of #6029


weirdan:
CI failures on master and in this PR are caused by the fix to [PHP bug #81070](https://bugs.php.net/bug.php?id=81070) ([commit](https://github.com/php/php-src/commit/1b3b5c94e52d10eb7a3f69b486a51b3f4d214d4f)).

Issue #6038, opened 2021-07-02
`array_map()` destroys generic part of shape+generic array


weirdan:
https://psalm.dev/r/0c83024b48

**Expected**: `array{0: string}<array-key,string>`

PR #6039, opened 2021-07-03
Keep generic part of shape on array_map return type

weirdan:
Fixes vimeo/psalm#6038


src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayMapReturnTypeProvider.php
```diff
@@ -187,6 +187,8 @@ function (Type\Union $_) use ($mapping_return_type): Type\Union {
                 );
                 $atomic_type->is_list = $array_arg_atomic_type->is_list;
                 $atomic_type->sealed = $array_arg_atomic_type->sealed;
+                $atomic_type->previous_key_type = $array_arg_atomic_type->previous_key_type;
+                $atomic_type->previous_value_type = $array_arg_atomic_type->previous_value_type;
```
weirdan, line 191: This is probably wrong

PR #6040, opened 2021-07-03
Update DateTime and DateTimeImmutable signatures

j4nr6n:
`DateTime::format` and `DateTimeImmutable::format` will always return a string starting in PHP 8.0.

https://psalm.dev/r/b5d797f092
https://www.php.net/manual/en/datetime.format.php#refsect1-datetime.format-changelog


tests/MethodCallTest.php
```diff
@@ -370,6 +373,9 @@ public function method() : self {
                     if (false !== $formatted) {}
                     function takesString(string $s) : void {}
                     takesString($formatted);',
+                'assertions' => [],
+                'error_levels' => [],
+                'php_version' =>  '8.0'
```
weirdan, line 378: Hmm, I'm confused. This test, as it stands, *should* fail on PHP 8.0 with `RedundantCondition`, but it seems it doesn't.

orklah: Shouldn't a line be added in the `old` part of callmap for Psalm to consider what's in `new`? I was always confused by how (or mainly why) it works like that

weirdan: Oh, wait, why you removed my commit, @j4nr6n ?

weirdan: @orklah yeah, the callmap is broken right now in this PR. I fixed that about an hour ago, but my commit was apparently erased by force-push.

weirdan: @orklah callmap deltas work as follows:

`Callmap.php` contains signatures for the latest PHP version (8.0 at the moment, specified in `Psalm\Internal\Codebase\InternalCallmapHandler` constants). When Psalm is set to analyse code for some previous PHP version, it iteratively applies deltas *in reverse order*, until it reaches the version it needs. On every iteration, entries found in `new` section of the delta are removed from callmap and those found in `old` section are inserted.

orklah: oh, I get it! Thanks!

j4nr6n: Sorry @weirdan I didn't intentionally remove any commits. Not sure what you're referring to.

weirdan: https://github.com/vimeo/psalm/pull/6040#event-4973873396 - this is where it happened

j4nr6n: Ok. I see. I restored the changes in the commit that I accidentally wiped out.

weirdan:
Thanks!

Issue #6044, opened 2021-07-04
muglug/phpunit is not installable


weirdan:
Currently CI fails because `muglug/phpunit` is not installable:

```console
$ gh repo clone muglug/phpunit
Cloning into 'phpunit'...
remote: Enumerating objects: 116963, done.
remote: Counting objects: 100% (2167/2167), done.
remote: Compressing objects: 100% (714/714), done.
remote: Total 116963 (delta 1486), reused 1996 (delta 1432), pack-reused 114796
Receiving objects: 100% (116963/116963), 81.41 MiB | 3.17 MiB/s, done.
Resolving deltas: 100% (79023/79023), done.
Updating upstream
remote: Enumerating objects: 5783, done.
remote: Counting objects: 100% (3856/3856), done.
remote: Compressing objects: 100% (960/960), done.
remote: Total 5783 (delta 2879), reused 3797 (delta 2858), pack-reused 1927
Receiving objects: 100% (5783/5783), 3.66 MiB | 4.02 MiB/s, done.
Resolving deltas: 100% (4477/4477), completed with 105 local objects.
From github.com:sebastianbergmann/phpunit
 * [new branch]          master     -> upstream/master
 * [new tag]             6.5.14     -> 6.5.14
[......]
 * [new tag]             9.5.6      -> 9.5.6

$ cd phpunit/
$ composer install
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/php-code-coverage dev-master requires phpunit/php-text-template ^3.0 -> found phpunit/php-text-template[dev-master, 3.0.x-dev (alias of dev-master)] but it conflicts with your root composer.json require (^2.0.3).
    - phpunit/php-code-coverage 10.0.x-dev is an alias of phpunit/php-code-coverage dev-master and thus requires it to be installed too.
    - Root composer.json requires phpunit/php-code-coverage ^10.0 -> satisfiable by phpunit/php-code-coverage[10.0.x-dev (alias of dev-master)].
```

@muglug could you move that repo (and any other used by `test-with-real-projects`) to `psalm` org where I can take care of them?

muglug:
I have renamed it to psalm/endtoend-test-phpunit, which should hopefully avoid ambiguity with the Psalm plugin and any concept of an authoritative fork of phpunit.

[...]

PR #6045, opened 2021-07-04
Use psalm/endtoend-test-phpunit

weirdan:
Fixes vimeo/psalm#6044

Should be faster too, with limited clone depth. I've reset `psalm/endtoend-test-phpunit/master` to `upstream/master` as it worked fine locally without any baseline changes.


weirdan:
@muglug now Psl fails. Mind transferring it as well?

PR #6048, opened 2021-07-04
Update Sublime LSP to match latest version

DannyJJK:
I noticed the current config didn't work on the Sublime Text LSP plugin, I've updated it to the latest version of the config for the plugin which works with Sublime Text 4


weirdan:
Thanks!

PR #6053, opened 2021-07-06
Change return type of ZipArchive::open to `int|true`

### LINKED ISSUE #6052: ZipArchive::open() never returns false
In the [ZipArchive::open documentation](https://www.php.net/manual/en/ziparchive.open.php):
> Returns true on success, or one of the following error codes: [...]

Psalm seems to interpret the return value as `bool|int` instead of `true|int`:
https://psalm.dev/r/13a184ea04

chapa:
As mentioned in https://www.php.net/manual/en/ziparchive.open.php#refsect1-ziparchive.open-returnvalues

Fixes #6052


weirdan:
As mentioned on the linked ticket, `ZipArchive::open()` does return `false`.

PR #6055, opened 2021-07-06
didChangeWatchedFiles should continue not return

tm1000:
didChangeWatchedFiles should continue not return per each watched file in the $changes array


weirdan (approved):

weirdan:
Thanks!

PR #6060, opened 2021-07-07
Pass statement to MethodReturnTypeProviderEvent

VincentLanglet:
I'm trying to solve the following issue on psalm-doctrine: https://github.com/psalm/psalm-plugin-doctrine/issues/77

According to @orklah, this is possible if I can access to the statement from the MethodReturnTypeProviderEvent.

This require to update some signatures.
I added the `$stmt` param at the same position it was done in https://github.com/vimeo/psalm/pull/5333
But maybe it can be improved. For instance, I see that both `$stmt->args` and `$stmt` are now pass to the `getReturnType`.

So maybe @muglug @weirdan, you'll prefer I change `$stmt->args` to `$stmt` in the original signature.
Or I can just pass `$stmt->var`.

Also, only the signature `MethodReturnTypeProvider::getReturnType` is changed ; maybe you'd like I update `FunctionReturnTypeProvider::getReturnType` too (and FunctionReturnTypeProviderEvent ?) in order to have a similar signature.

Of course, I'm opened to better solution in order to solve https://github.com/psalm/psalm-plugin-doctrine/issues/77 (with psalm assert maybe ?)

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
@@ -54,7 +54,7 @@ class FunctionCallReturnTypeFetcher
             $stmt_type = $codebase->functions->return_type_provider->getReturnType(
                 $statements_analyzer,
                 $function_id,
-                $stmt->args,
+                $stmt,
                 $context,
                 new CodeLocation($statements_analyzer->getSource(), $function_name)
             );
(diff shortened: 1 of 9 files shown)


weirdan:
> So maybe  you'll prefer I change $stmt->args to $stmt in the original signature.

Yes, please. Naturally `getCallArgs()` method should stay, and just start returning `$this->stmt->call_args`.

It would be good to keep function type provider signature in sync with method one too.

VincentLanglet:
> > So maybe  you'll prefer I change $stmt->args to $stmt in the original signature.
> 
> Yes, please. Naturally `getCallArgs()` method should stay, and just start returning `$this->stmt->call_args`.
> 
> It would be good to keep function type provider signature in sync with method one too.

Done ! @weirdan :)

weirdan (approved):

weirdan:
Thanks!

VincentLanglet:
Thanks. Is there any strategy about when the new releases are made ?

weirdan:
I think @muglug just does that from time to time.

Issue #6056, opened 2021-07-07
When specifying config file psalm does not find any errors


simPod:
Create psalm config using `vendor/bin/psalm --init src 1`

Running 

```sh
vendor/bin/psalm --config psalm.xml

Scanning files...
Analyzing files...

░

To whom it may concern: Psalm cannot detect unused classes, methods and properties
when analyzing individual files and folders. Run on the full project to enable
complete unused code detection.
------------------------------
No errors found!
------------------------------
```
finds 0 errors

----------

Running 

```sh
vendor/bin/psalm

Scanning files...
Analyzing files...

░░░░░░░E░░░░░░E░░░░░░░░░░░░

...

------------------------------
4 errors found
------------------------------

```

finds some errors. Why is it so? I though I could have multiple psalm configs and specify them using `--config` flag while it seems the flag disables the analysis.

weirdan:
Does it work as expected if you run it as
```
vendor/bin/psalm --config=psalm.xml
```
? (note the equals sign)

simPod:
Oh it does, thanks.

weirdan:
I thought it was how `getopt()` parses that line, but it's not `getopt()`s fault:

```console
$ php -r 'var_dump(getopt("", ["config:"]));' -- --config config.xml
Command line code:1:
array(1) {
  'config' =>
  string(10) "config.xml"
}
```

simPod:
Is there a way to run psalm with xdebug on? I tried to debug it.

orklah:
In Psalm.php there is a specific code that disable xdebug. You can comment the "check" method call IIRC and xdebug will run.

weirdan:
Just run it with `PSALM_ALLOW_XDEBUG` env variable set ([docs](https://psalm.dev/docs/running_psalm/plugins/authoring_plugins/#using-xdebug)):

```console
$ PSALM_ALLOW_XDEBUG=1 psalm
```

simPod:
Even better. The issue is in `CliUtils::getPathsToCheck()`

simPod:
I'll send fix soon.

weirdan:
Going to keep this open as I would like to address this in a more general way.

PR #6062, opened 2021-07-08
Skip argv when --config file.xml is passed

simPod:
I tried to pass a config file using `--config psalm.xml`. Though in help (`-h`) it's specified as `--config=psalm.xml`.

That resulted in incorrect arg parsing and `psalm.xml` was set as a path to check.

I'm no sure we want to support it though. Seems like code heavily relies on `=` being there.

Related to https://github.com/vimeo/psalm/issues/6056

### DIFF
--- a/src/Psalm/Internal/CliUtils.php
+++ b/src/Psalm/Internal/CliUtils.php
@@ -264,6 +264,9 @@ final class CliUtils
                 }
 
                 if (substr($input_path, 0, 2) === '--' && strlen($input_path) > 2) {
+                    if (substr($input_path, 2) === 'config') {
+                        ++$i;
+                    }
                     continue;
                 }
 


weirdan:
I think a better (and more general) fix would be to use third argument to `getopt()`:

```console
$ php -r 'var_dump(getopt("", ["config:"], $rest_offset), $rest_offset, join(" ", array_slice($argv, $rest_offset)));' -- --config config.xml file.php and-a-folder
Command line code:1:
array(1) {
  'config' =>
  string(10) "config.xml"
}
Command line code:1:
int(3)
Command line code:1:
string(21) "file.php and-a-folder"
```
This will make Psalm require files to check to come last in argument list, but that doesn't sound too bad, especially since passing a non-switch/option argument stops further options from being processed by `getopt()` anyway.

With that said, I recognize that similar hacks are already in place, so there's no harm in adding this.

weirdan:
Thanks!

PR #6068, opened 2021-07-09
Make ClassLikeNodeScanner resilient against missing class storage

### LINKED ISSUE #5971: [Bug] Could not get class storage for * | Cache problem when analyzing 1 file
Opening up an upstream issue for https://github.com/psalm/psalm-plugin-laravel/issues/110 

This reproducer is available: https://github.com/psalm/psalm-plugin-laravel/issues/110#issuecomment-731758452 

In the plugin, I can work around the issue by forcing psalm to scan certain classes, but I'm not sure why the cache doesn't have them in the first place.

While debugging, I have narrowed it down to also have something to do with stubs: https://github.com/psalm/psalm-plugin-laravel/issues/110#issuecomment-862622308

Happy to help resolve the issue if someone is familiar with what the problem might be.

Also seems related to https://github.com/vimeo/psalm/issues/4462 as laravel uses `class_alias` to register some facades

caugner:
Fixes https://github.com/psalm/psalm-plugin-laravel/issues/134, and probably fixes #5971 as well.

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
@@ -195,7 +195,11 @@ class ClassLikeNodeScanner
                     $storage->aliases = $this->aliases;
 
                     foreach ($storage->dependent_classlikes as $dependent_name_lc => $_) {
-                        $dependent_storage = $this->codebase->classlike_storage_provider->get($dependent_name_lc);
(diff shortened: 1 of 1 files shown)


caugner:
@weirdan If this is an acceptable fix, this would be very helpful to have in the next Psalm patch release. These "Could not get class storage" crashes have been bugging me and my devs for months.

weirdan (approved):

weirdan:
Thanks!

PR #6069, opened 2021-07-09
Added str_word_count to the stub file (fix #6016)

### LINKED ISSUE #6016: [Q] Associated with InvalidReturnType & InvalidReturnStatement
PHP 8, Level 6

<details>
  <summary>Screenshot</summary>
  
  cats
</details>
Why is this considered an error, what's wrong?

According to the PHP documentation, by default [`str_word_count`](https://www.php.net/manual/en/function.str-word-count.php) will return the value as an int.

Even `str_word_count($this->content, 0)` is also considered an error

elnoro:
Hello

If I need to add any tests, please tell me where.
Thank you!

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -846,10 +846,15 @@ function strrev(string $string): string {}
 /**
  * @psalm-pure
  *
- * @psalm-return (
- *     $format is positive-int
- *     ? array<string>
- *     : int
+ * @param 0|1|2 $format
+ * @return (
+ *     $format is 0 ?
+ *     int : 
+ *     (
+ *         $format is 1 ?
+ *         list<string> :
+ *         array<int, string>
+ *     )
  * )
  */
 function str_word_count(string $string, int $format = 0, string|null $characters = null) {}
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -846,6 +846,17 @@ function strrev(string $string): string {}
 /**
  * @psalm-pure
  *
+ * @psalm-return (
+ *     $format is positive-int
+ *     ? array<string>
+ *     : int
+ * )
+ */
+function str_word_count(string $string, int $format = 0, string|null $characters = null) {}
+
+/**
+ * @psalm-pure
+ *
  * @param string|string[] $pattern
  * @param string|array<string|int|float> $replacement
  * @param string|array<string|int|float> $subject


stubs/CoreGenericFunctions.phpstub
```diff
@@ -843,6 +843,17 @@ function strrchr(string $haystack, string $needle) {}
...
 function strrev(string $string): string {}
 
+/**
+ * @psalm-pure
+ *
+ * @psalm-return (
+ *     $format is positive-int
+ *     ? array<string>
+ *     : int
+ * )
```
weirdan, line 853: Since there's a list of allowed values for `$format` we can be more precise, and flag incorrect values for the `$format` parameter as well
```suggestion
 *
 * @param 0|1|2 $format
 * @return (
 *     $format is 0 ?
 *     int : 
 *     (
 *         $format is 1 ?
 *         list<string> :
 *         array<int, string>
 *     )
 * )
```

weirdan (changes requested):

weirdan:
We generally don't have tests for most of the stubs.

stubs/CoreGenericFunctions.phpstub
elnoro: Thank you for the suggestion! I did not think it was necessary - but it definitely can catch some bugs.

elnoro:
Ok, so it failed on one of the test projects 
The issue reported by Psalm is UnnecessaryVarAnnotation https://github.com/muglug/psl/blob/0.1.x/src/Psl/Str/Byte/words.php#L22. 
Psalm could not figure out the type of the variable before, but it can and therefore it became obsolete. Should I send a PR to the library then?

caugner:
> Should I send a PR to the library then?

That's probably a good idea, but make sure to create the PR [in muglug's fork](https://github.com/muglug/psl), which is used in the "test-with-real-projects".

weirdan:
> That's probably a good idea, but make sure to create the PR in muglug's fork, which is used in the "test-with-real-projects".

It's better to send the PR upstream. For muglug's fork we would just baseline the issue.

weirdan:
> Should I send a PR to the library then?

Feel free to do that, but it's not a prerequisite to get this merged.

caugner:
> It's better to send the PR upstream. For muglug's fork we would just baseline the issue.

Oh, I just saw earlier that the fork's 1.6.x branch is tested, while upstream is already on 1.7.x, so I assumed the fork is not kept in sync with upstream anymore.

PR #6070, opened 2021-07-09
Switch PSL test to Psalm's fork

weirdan:


caugner:
@weirdan If you have access to that PR, it might make sense to change the default branch so that the repository no longer shows `0.1.x` by default.

Issue #4129, opened 2020-09-04
Colorize file/line information in output


Rarst:
Psalm colors code element in snippet that triggered the error.

That is useful, however in any large file it is impractical to locate the error by snippet. For fast IDE lookup the most important information to locate the error is file name and line number.

Also since the length of output varies, this is in different place on screen from error to error, requiring to visually scan monochrome message for it every time.

Can I suggest this part of output is made colored? Probably not red, but blue or something.

Thank you for considering. :)

muglug:
@rarst mind giving an example of what it should look like?

Rarst:
Ok, blue is too dark to be readable in vanilla terminal.

If I just go with more red I get something like this:

![psalm-colorize-location](https://user-images.githubusercontent.com/737584/92505256-800bae80-f20c-11ea-8534-e02add7bd198.png)

caugner:
> Ok, blue is too dark to be readable in vanilla terminal.

Maybe it would make sense to take some inspiration from `yarn outdated` colors:

![grafik](https://user-images.githubusercontent.com/495429/125142879-253b3000-e119-11eb-8921-28446cd25c86.png)

PR #6071, opened 2021-07-09
Remove brackets around issue ref in console output

caugner:
Both GitHub and GitLab treat the closing bracket as part of the URL, so
opening it directly from the log view doesn't currently work.

* GitHub:
  grafik
* GitLab:
  grafik

PS: For GitLab, I [opened a PR](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/65888) to fix this upstream.

### DIFF
--- a/src/Psalm/Report/ConsoleReport.php
+++ b/src/Psalm/Report/ConsoleReport.php
@@ -31,7 +31,7 @@ class ConsoleReport extends Report
             $issue_string .= 'INFO';
         }
 
-        $issue_reference = $issue_data->link ? ' (see ' . $issue_data->link . ')' : '';
+        $issue_reference = $issue_data->link ? ' - see: ' . $issue_data->link : '';
 
         $issue_string .= ': ' . $issue_data->type
             . ' - ' . $issue_data->file_name . ':' . $issue_data->line_from . ':' . $issue_data->column_from
--- a/src/Psalm/Report/GithubActionsReport.php
+++ b/src/Psalm/Report/GithubActionsReport.php
@@ -12,7 +12,7 @@ class GithubActionsReport extends Report
     {
         $output = '';
         foreach ($this->issues_data as $issue_data) {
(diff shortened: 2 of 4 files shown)


src/Psalm/Report/ConsoleReport.php
```diff
@@ -31,7 +31,7 @@ private function format(\Psalm\Internal\Analyzer\IssueData $issue_data): string
             $issue_string .= 'INFO';
         }
 
-        $issue_reference = $issue_data->link ? ' (see ' . $issue_data->link . ')' : '';
+        $issue_reference = $issue_data->link ? ' - see: ' . $issue_data->link : '';
```
caugner, line 34: Hm, I wonder if the `see:` prefix is even necessary?

caugner: Square brackets would also be an alternative: `[see ...]`

caugner: A third option would be to use a gray color for the links to show them less prominent (for environments that don't color links anyway), cf. #4129 (I wonder who's responsible for show the URLs in a nice blue tone)

muglug:
I've added support for trailing brackets in https://psalm.dev's url redirection, so these links should now work. I know it's a little messy, but I _think_ there isn't a strong enough argument for getting rid of the brackets (I quite like them).

If you want to make this tighter while keeping the brackets, one possible alternative would be to add a [Zero-width space](https://en.wikipedia.org/wiki/Zero-width_space) (which are invalid for urls).

caugner:
@muglug Thanks for the redirect fix!

Regarding the zero-width space, I'm not sure this would necessarily help. GitLab is using a regular expression with a negative character range to *exclude* characters, and chances are it's implemented similar in GitHub.

As regards GitLab, my MR has meanwhile been merged and the issue will be fixed in the upcoming release. 🎉

So let's keep it simple and close this PR. 😄

muglug:
Awesome!

PR #6077, opened 2021-07-12
Use ReflectionNamedType instead of ReflectionType (PHP 7.1+) (v2)

lptn:
2nd attempt for “expired” PR https://github.com/vimeo/psalm/pull/5204 Use ReflectionNamedType instead of ReflectionType (PHP 7.1+). Additionally to that PR, I've moved `PropertyType::hasType` to PHP delta 7.4 (see https://3v4l.org/MS1LQ)


caugner:
Here's the failing test:

```
ERROR: PossiblyInvalidArgument - src/Psalm/Internal/Codebase/Reflection.php:380:82 - Argument 1 of Psalm\Internal\Codebase\Reflection::getPsalmTypeFromReflectionType expects ReflectionType|null, possibly different type ReflectionNamedType|ReflectionUnionType provided (see https://psalm.dev/092)
                    $storage->return_type = self::getPsalmTypeFromReflectionType($reflection_return_type);


ERROR: PossiblyInvalidArgument - tests/TypeParseTest.php:996:66 - Argument 1 of Psalm\Codebase::getPsalmTypeFromReflection expects ReflectionType|null, possibly different type ReflectionNamedType|ReflectionUnionType|null provided (see https://psalm.dev/092)
            (string) \Psalm\Codebase::getPsalmTypeFromReflection($reflectionFunc->getReturnType())
```

The corresponding locations are:

https://github.com/vimeo/psalm/blob/e93b37a2253f63b981c13694c8fb9c2e6a5bc7f4/src/Psalm/Internal/Codebase/Reflection.php#L379-L381

and

https://github.com/vimeo/psalm/blob/e93b37a2253f63b981c13694c8fb9c2e6a5bc7f4/tests/TypeParseTest.php#L994-L997

caugner:
So it looks like we need to widen the parameter type here:

https://github.com/vimeo/psalm/blob/6d09418a23399bfc9e99f21946da532f5bad4f90/src/Psalm/Codebase.php#L529-L532

And here:

https://github.com/vimeo/psalm/blob/6d09418a23399bfc9e99f21946da532f5bad4f90/src/Psalm/Internal/Codebase/Reflection.php#L403

**Alternatively**, we could suppress the errors, like it's already been done here (inside `Reflection::getPsalmTypeFromReflectionType()`):

https://github.com/vimeo/psalm/blob/6d09418a23399bfc9e99f21946da532f5bad4f90/src/Psalm/Internal/Codebase/Reflection.php#L409-L427

dictionaries/CallMap_80_delta.php
```diff
@@ -498,6 +498,10 @@
 'read_exif_data' => ['array', 'filename'=>'string', 'sections_needed='=>'string', 'sub_arrays='=>'bool', 'read_thumbnail='=>'bool'],
 'ReflectionClass::getConstants' => ['array<string,mixed>'],
 'ReflectionType::isBuiltin' => ['bool'],
+'ReflectionFunction::getReturnType' => ['ReflectionNamedType|ReflectionUnionType|null'],
+'ReflectionFunctionAbstract::getReturnType' => ['ReflectionNamedType|ReflectionUnionType|null'],
+'ReflectionMethod::getReturnType' => ['ReflectionNamedType|ReflectionUnionType|null'],
+'ReflectionParameter::getType' => ['ReflectionNamedType|ReflectionUnionType|null'],
```
caugner, line 504: I think we should use `ReflectionType` here:

```suggestion
'ReflectionFunction::getReturnType' => ['ReflectionType|null'],
'ReflectionFunctionAbstract::getReturnType' => ['ReflectionType|null'],
'ReflectionMethod::getReturnType' => ['ReflectionType|null'],
'ReflectionParameter::getType' => ['ReflectionType|null'],
```

This should also fix the two Psalm errors.

dictionaries/CallMap_80_delta.php
```diff
@@ -498,6 +498,10 @@
 'read_exif_data' => ['array', 'filename'=>'string', 'sections_needed='=>'string', 'sub_arrays='=>'bool', 'read_thumbnail='=>'bool'],
 'ReflectionClass::getConstants' => ['array<string,mixed>'],
 'ReflectionType::isBuiltin' => ['bool'],
```
caugner, line 500: > **ReflectionType** has become abstract and **ReflectionType::isBuiltin()** has been moved to [ReflectionNamedType::isBuiltin()](https://www.php.net/manual/en/reflectionnamedtype.isbuiltin.php). 

(Source: https://www.php.net/manual/en/class.reflectiontype.php#reflectiontype.changelog)

So I think this line should be updated accordingly:

```suggestion
'ReflectionNamedType::isBuiltin' => ['bool'],
```

caugner (changes requested):
LGTM, but needs some adjustments

dictionaries/CallMap_80_delta.php
lptn: it depends on what we are going to archive:
 a) specify all more specific types
 b) specify fewer types by using parent type

I assume the goal of the psalm is to have as much as possible types information, this is why a) is preferable IMO.

Just for a reference: 
- `ReflectionType` has 2 child classes:
  - `ReflectionUnionType` (methods: `getTypes`)
  - `ReflectionNamedType` (methods: `getName`, `isBuiltin`)

dictionaries/CallMap_80_delta.php
lptn: I like this change but have 2 points:
 1. this one out of the scope of this PR (please understand me currently, [my first PR](#5204) is 5 month old, increasing a scope of this PR also increases a chance for this PR to be lost and outdated as the previous one)
 2. `ReflectionType::isBuiltin` is removed in PHP 8.0. As far as I understand, both `new` and `old` arrays doesn't provide instructions to psalm that the method is removed. `ReflectionNamedType::isBuiltin' => ['bool'],`  record is a valid one, but how to instruct psalm that `ReflectionType::isBuiltin` is removed?


PS: I also see 2 easy wins regarding changed files:
 1. `If they were just updated, the function/method will be present in the 'added' signatures.` uses `added` key, however we use `new`
 2. `dictionaries/CallMap_80_delta.php` doesn't contain indent and thus not really readable

But I've decided to limit a scope of this (and prev) PR

dictionaries/CallMap_80_delta.php
lptn: just created a snippet to test all possible cases for PHP 8.0: https://3v4l.org/QE9Yn

caugner: > I assume the goal of psalm is to have as much as possible types information, this is why a) is preferable IMO.

I get your point, but even when using `ReflectionNamedType|ReflectionUnionType` over `ReflectionType`, you will still need to check the type with `instanceof` first, before you can call any of those three methods you mentioned.

Essentially this is narrowing the type more than the PHP's public API promises, and would prevent Psalm users from handling the case that another `ReflectionType` was returned. On the other hand, this could also be considered a nice side-effect.

My guess is that Psalm tries to stay close to the official PHP signatures, but maybe this is the right moment to ask @muglug or @weirdan for their opinion?

lptn:
@caugner 
>Alternatively, we could suppress the errors, like it's already been done here (inside Reflection::getPsalmTypeFromReflectionType()):

```
ERROR: PossiblyInvalidArgument - src/Psalm/Internal/Codebase/Reflection.php:380:82 - Argument 1 of Psalm\Internal\Codebase\Reflection::getPsalmTypeFromReflectionType expects ReflectionType|null, possibly different type ReflectionNamedType|ReflectionUnionType provided (see https://psalm.dev/092)
                    $storage->return_type = self::getPsalmTypeFromReflectionType($reflection_return_type);


ERROR: PossiblyInvalidArgument - tests/TypeParseTest.php:996:66 - Argument 1 of Psalm\Codebase::getPsalmTypeFromReflection expects ReflectionType|null, possibly different type ReflectionNamedType|ReflectionUnionType|null provided (see https://psalm.dev/092)
            (string) \Psalm\Codebase::getPsalmTypeFromReflection($reflectionFunc->getReturnType())
```


I assume we got these errors because circleci uses php 7.4 where there are no `ReflectionNamedType` and `ReflectionUnionType` (they added in PHP 8.0), so it couldn't realise they extends `ReflectionType`. The simplest option is to suppress these error. Alternative options are more time-consuming and requires a way more knowledge about the package and roadmap that I currently have

dictionaries/CallMap_80_delta.php
caugner: Good points, in doubt I will open a PR separately.

I had a quick look at the [InternalCallMapHandler](https://github.com/vimeo/psalm/blob/6abce3525ac34f815aaa3b7953739b744ca5c39e/src/Psalm/Internal/Codebase/InternalCallMapHandler.php#L361-L402).

I think it should suffice to

1. not mention the removed method in the `dictionaries/CallMap.php`,
2. to add the removed method to the `old` key of the callmap corresponding to the version in which it was removed.

To determine the effective CallMap, Psalm goes backwards from the latest supported version by Psalm to the defined version. Psalm never goes the other way, so removed methods don't have to be declared removed explicitly.

caugner: Sorry, never mind, the `isBuiltin` method is already declared correctly in the `CallMap_80_delta`:

grafik

dictionaries/CallMap_71_delta.php
```diff
@@ -40,6 +40,10 @@
     'session_gc' => ['int'],
     'timezone_identifiers_list' => ['list<string>|false', 'timezoneGroup='=>'int', 'countryCode='=>'?string'],
     'unpack' => ['array', 'format'=>'string', 'string'=>'string', 'offset='=>'int'],
+    'ReflectionFunction::getReturnType' => ['?ReflectionNamedType'],
+    'ReflectionFunctionAbstract::getReturnType' => ['?ReflectionNamedType'],
+    'ReflectionMethod::getReturnType' => ['?ReflectionNamedType'],
+    'ReflectionParameter::getType' => ['?ReflectionNamedType']
```
caugner, line 46: To be consistent, these should go at the top between `DateTimeZone::listIdentifiers` and `SQLite3::createFunction`.

weirdan (changes requested):
I'm not sure whether callmap delta algorithm was misunderstood, but it looks that way to me.

I briefly described it here: https://github.com/vimeo/psalm/pull/6040#discussion_r663330329

> `Callmap.php` contains signatures for the latest PHP version (8.0 at the moment, specified in `Psalm\Internal\Codebase\InternalCallmapHandler` constants). When Psalm is set to analyse code for some previous PHP version, it iteratively applies deltas *in reverse order*, until it reaches the version it needs. On every iteration, entries found in `new` section of the delta are removed from callmap and those found in `old` section are inserted.

Issue #6061, opened 2021-07-08
Psalm does not understand variable initialization order when initialized within array.


simPod:
https://psalm.dev/r/d4185292b9

I'd expect not errors about variable `$timestamp` being not found or unused.

muglug:
Few comments:

- IMO this is bad code and should be rewritten as https://psalm.dev/r/366293c0e1
- the bug is caused by evaluation order — currently all keys are evaluated first, which is wrong. It should be possible instead to do it by key and value

simPod:
Definitely agree the code is not perfect though it's still valid or is it not?

Stumbled upon it in legacy code base where there's complex array structure in tests and the lowest level array is constructed like this. Hard to squeeze foreach inside. I would have to wrap it in a function, call it and unpack `...` or sth like that.

```php
[22=>[
    $timestamp = (new DateTimeImmutable('2017-08-06 00:00:00'))->getTimestamp() => [
        'timestamp' => $timestamp,
    ],
    $timestamp = (new DateTimeImmutable('2017-08-06 01:00:00'))->getTimestamp() => [
        'timestamp' => $timestamp,
    ],
    $timestamp = (new DateTimeImmutable('2017-08-06 02:00:00'))->getTimestamp() => [
        'timestamp' => $timestamp,
    ]
]];
```

->

```php
<?php

[22=>(function() {
    $dates = ['2017-08-06 00:00:00', '2017-08-06 01:00:00', '2017-08-06 02:00:00'];

    $arr = [];

    foreach ($dates as $date) {
        $timestamp = (new DateTimeImmutable($date))->getTimestamp();
        $arr[$timestamp] = ['timestamp' => $timestamp];
    }

    return $arr;
})()]
```

PR #6078, opened 2021-07-12
Analyze array keys first when checking an array

orklah:
This will partly solve #6061. It ensure keys are analyzed before the value of the array, so any assignment here can be used later.

I had to make `$array_creation_info` to be able to replace the whole object with a clone because there is a special handling for unpacking and it has to be after analyzing the value (so after analyzing the keys with this PR) but it also needs the ArrayCreationInfo untouched.

This PR doesn't solve the whole #6061 issue because it doesn't handle the UnusedVariable that's emitted (I don't get how this works yet)


orklah:
superseded by #6081

PR #6081, opened 2021-07-12
Fix #6061 — delay analysis of array value except when unpacking

### LINKED ISSUE #6061: Psalm does not understand variable initialization order when initialized within array.

muglug:
This undoes a mistake made in https://github.com/vimeo/psalm/commit/3f2d57c7a3a69ae1d3b6544db0ae0796c73ae2c9.

cc @orklah — I think this fix is preferable to yours (unless there's something I've missed)

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -232,11 +232,11 @@ class ArrayAnalyzer
         PhpParser\Node\Expr\ArrayItem $item,
         Codebase $codebase
     ) : void {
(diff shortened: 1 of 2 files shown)


orklah:
Right, makes sense and it's more readable. Thanks!

orklah:
Does this also fix the UnusedVariable issue? Can you point me the solution if it does?

muglug:
@orklah 

> Does this also fix the UnusedVariable issue? Can you point me the solution if it does?

Yes, see attached test!

orklah:
> The bug was that values were analysed before keys, which also screws up variable use detection.

I must have messed up somewhere then, My solution still had UnusedVariable issue even though I inverted the value and key analysis.

Thanks for the explanation!

PR #6083, opened 2021-07-12
Always emit InvalidArgument when the parameter is always false

### LINKED ISSUE #5790: InvalidArgument vs PossiblyFalseArgument
A method expecting a `string`, but passed `true` outputs an `InvalidArgument`:

https://psalm.dev/r/47caf98787

Same thing, but now passed `false` outputs a `PossiblyFalseArgument`:

https://psalm.dev/r/0b9b837d1d

In this case `false` is not "possibly" false, it definitely is false.

Is this intentional behavior or a bug?

(I would have expected an `InvalidArgument` error in the second case as well)

orklah:
This fixes #5790


weirdan:
Thanks!

PR #6084, opened 2021-07-13
added dataflow to stringy bitwise op, bitwise not op (fix #5957)

### LINKED ISSUE #5957: UnusedVariable false positive
Hi, for the following code (simplified example from [brick/math](https://github.com/brick/math/blob/master/src/BigInteger.php#L231)):

```php
function randomBits() : string
{
    $bitmask = \chr(0xFF >> 1);

    $randomBytes    = random_bytes(1);
    $randomBytes[0] = $randomBytes[0] & $bitmask;

    return $randomBytes;
}
```

Psalm complains:

> INFO: UnusedVariable - 4:5 - $bitmask is never referenced or the value is not used

https://psalm.dev/r/e436f35229

elnoro:
Hello 

Fixes #5957

It appears addDataFlow was never called for bitwise operators with stringy (hasString() == true) operands.
And I could not find an implementation of "addDataflow" for unary operators, so I added one in BitwiseNotAnalyzer.

Added snippets from the original issue as a test and ran psalm on [brick/math](https://github.com/brick/math) - no more UnusedVariable errors.


weirdan (approved):

weirdan:
Thanks!

PR #6086, opened 2021-07-13
cleanup and refactor AssertionFinder

orklah:
I made a number of refactors, cleanup and documentation in #5930 but it's not finished so we might as well merge what's not broken. It'll improve the code and it will clear the other PR.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -2,6 +2,10 @@
 namespace Psalm\Internal\Analyzer\Statements\Expression;
 
 use PhpParser;
+use PhpParser\Node\Expr\BinaryOp\Equal;
+use PhpParser\Node\Expr\BinaryOp\Identical;
+use PhpParser\Node\Expr\BinaryOp\NotEqual;
+use PhpParser\Node\Expr\BinaryOp\NotIdentical;
 use Psalm\CodeLocation;
 use Psalm\Codebase;
 use Psalm\FileSource;
@@ -28,6 +32,10 @@ use function substr;
 
 /**
  * @internal
+ * This class transform conditions in code into "assertions" that will be reconciled with the type already known of a
+ * given variable to narrow the type or find paradox.
+ * For example if $a is an int, if(count($a) > 0) will be turned into an assertion to make psalm understand that in the
+ * if block, $a is a positive-int
  */
 class AssertionFinder
 {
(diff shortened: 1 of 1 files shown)


orklah:
build failure seem to come upstream https://github.com/sebastianbergmann/type/commit/63aee82dc152c48b7633f57f037914bfd9531878

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -28,6 +32,10 @@
 
 /**
  * @internal
+ * This class transform conditions in code into "assertions" that will be reconciled with the type already known of a
+ * given variable to narrow the type or find paradox.
+ * For example if $a is an int, if(count($a) > 0) will be turned into an assertion to make psalm understand that in the
```
weirdan, line 37: The example looks a bit off, using `count()` on an `int` is kinda unexpected (and is a TypeError in PHP 8).

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -477,7 +485,7 @@ private static function scrapeEqualityAssertions(
     }
 
     /**
-     * @param PhpParser\Node\Expr\BinaryOp\NotIdentical|PhpParser\Node\Expr\BinaryOp $conditional
+     * @param PhpParser\Node\Expr\BinaryOp\NotIdentical|PhpParser\Node\Expr\BinaryOp\NotEqual $conditional
```
weirdan, line 488: I recall there's quite a bit of dead code in this class. It's not caught by Psalm because of param types being too wide.

It could be interesting to implement param counterpart to 'make return types as tight as possible' option, at least for private/protected methods.

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
orklah: Whoops, good catch

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
orklah: I'd kill for that feature :p

I tried everything, I used Psalter to fix MissingParamType while telling Psalm to ignore phpdoc, I tried to tweak the code to make Psalm fix MissingParamType even when the Param is already there but it's a mess every time

orklah: I narrowed some phpdoc, mainly for $conditional. Let's see if there's new issues

weirdan: I think it was `hasPositiveNumberCheck()` and `hasZeroCheck()` methods. They are only ever called from `getGreaterAssertions()`, but have code handling Identical, Smaller, Equal etc.

orklah: Right, I missed quite a few. I'm on it

orklah: Done. There was some dead code that I removed. I wonder if I should leave a note to this PR for future readers that may need to reimplement the code

weirdan:
Thanks!

PR #6089, opened 2021-07-14
Fix usage of gettype in a switch with closed resource

### LINKED ISSUE #6088: Psalm crashes on switch-case statement with gettype call and 'resource (closed)' followed by 'unknown type' cases (misplaced brackets)
Hi!

We have been using Psalm for a while now, and already improved our codebase quite a bit!

However, we ran into a problem while adding Psalm to yet another project: the Psalm runner actually crashes. I managed to find out where the error occurs, and it is really quite specific.

A minimal example can be found here: https://psalm.dev/r/648ae887e8

The error occurs whenever a `switch` on the value of `gettype` has at least two case statements, namely `resource (closed)` and `unknown type`. The `case` statements should be in that order for the error to show, and the `resource (closed)`-case and all the other cases between `case resource (closed):` and `case unknown type:` should not have a statement which exits the `switch`-statement (i.e., a `break`, `throw`, `return` or even `continue`).

So the following fails:

```
<?php

$data = "foo";
switch (gettype($data)) {
	case "resource (closed)":
	case "unknown type":
		return "foo";
}
```

but this does not (https://psalm.dev/r/c95469669e):
```
<?php

$data = "foo";
switch (gettype($data)) {
	case "unknown type":
	case "resource (closed)":
		return "foo";
}
```

In our codebase we can easily change the order of the case-statements and thereby evade this error. I don't know whether this issue maybe signifies a larger bug, so I would like to report this issue nonetheless.

Keep up the good work :)

orklah:
This fixes #6088

It was caused by the absence of a specific code handling this situation that I retrieved from getDebugType(In)?equalityAssertions

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -2381,7 +2381,15 @@ class AssertionFinder
             }
         } else {
             if ($var_name && $var_type) {
-                $if_types[$var_name] = [['!' . $var_type]];
+                if ($var_type === 'class@anonymous') {
+                    $if_types[$var_name] = [['!=object']];
+                } elseif ($var_type === 'resource (closed)') {
+                    $if_types[$var_name] = [['!closed-resource']];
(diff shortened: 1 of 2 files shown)


weirdan:
Thanks!

Issue #6049, opened 2021-07-05
str_pad, str_repeat and non-empty-strings


ciaranmcnulty:
Psalm does not know that str_pad is guaranteed to return a non-empty-string for certain inputs

https://psalm.dev/r/c7f5c10980

If the input is a non-empty-string and the padding is >0, this should always be the case

weirdan:
> If the input is a non-empty-string and the padding is >0, this should always be the case

Actually it's *or*, not *and*

weirdan:
@ciaranmcnulty does the following signature match your expectations: https://psalm.dev/r/ecdcca18ba ?

ciaranmcnulty:
It looks like there is some funny stuff when the length is less than the original string length too

weirdan:
Got an example?

ciaranmcnulty:
Actually that looks fine. There's a third param but it doesn't affect the logic as it has to be a non-empty-string.

str_repeat seems to have a very similar issue / solution

PR #6094, opened 2021-07-14
Detailed return types for `str_pad()` and `str_repeat()`

weirdan:
Fixes vimeo/psalm#6049

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -762,12 +762,25 @@ function str_ireplace($search, $replace, $subject, &$count = null) {}
 /**
  * @psalm-pure
  *
+ * @return ($input is non-empty-string ? non-empty-string : ($pad_length is positive-int ? non-empty-string: string))
+ *
  * @psalm-flow ($input, $pad_string) -> return
  */
 function str_pad(string $input, int $pad_length, $pad_string = '', int $pad_type = STR_PAD_RIGHT): string {}
 
 /**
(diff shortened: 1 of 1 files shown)


stubs/CoreGenericFunctions.phpstub
```diff
@@ -762,13 +762,26 @@ function str_ireplace($search, $replace, $subject, &$count = null) {}
...
+ * @return ($input is non-empty-string ? non-empty-string : ($pad_length is positive-int ? non-empty-string: string))
+ *
  * @psalm-flow ($input, $pad_string) -> return
  */
 function str_pad(string $input, int $pad_length, $pad_string = '', int $pad_type = STR_PAD_RIGHT): string {}
 
 /**
  * @psalm-pure
  *
+ * @todo update $multiplier to be `0|positive-int`
```
orklah, line 774: Did you left this on purpose?

ciaranmcnulty:
str_pad enforces that the pad_string is non-empty - is that something that should be captured as a param?

stubs/CoreGenericFunctions.phpstub
weirdan: Yes. `str_repeat()` throws warnings when given negative int, but Psalm was emitting too many false positive coercion issues (int -> positive int) even on its own codebase.

weirdan:
>  is that something that should be captured as a param?

Yeah, it needs to be typed as `non-empty-string` then.

PR #6099, opened 2021-07-15
Correct callmap entries for XMLWriter

### LINKED ISSUE #6091: Incorrect signature/stub used for XMLWriter::startElementNs
According to the PHP documentation the 1st and 3rd parameters to `XMLWriter::startElementNs()` accept `string|null`; see https://www.php.net/manual/en/xmlwriter.startelementns. However, Psalm complains when `null` is passed as an argument to this method: https://psalm.dev/r/216f3978ac.

Unfortunately I cannot find where Psalm infers these method signatures from; they don't seem to be in the `/stubs` folder. I would be happy to check and correct the method signatures if I know where to find them.

aboks:
Checked and fixed the following, based on the current documentation in the PHP manual:

* capitalization in method names
* parameter names
* parameter types (especially for nullable params)
* return types (especially for `...|false`-types)
* differences between pre- and post-8.0.0-signatures

Fixes #6091


dictionaries/CallMap.php
```diff
@@ -15643,46 +15643,46 @@
...
-'XMLWriter::setIndent' => ['bool', 'indent'=>'bool'],
-'XMLWriter::setIndentString' => ['bool', 'indentstring'=>'string'],
+'XMLWriter::openUri' => ['bool', 'uri'=>'string'],
+'XMLWriter::outputMemory' => ['string', 'flush='=>'bool'],
+'XMLWriter::setIndent' => ['bool', 'enable'=>'bool'],
+'XMLWriter::setIndentString' => ['bool', 'indentation'=>'string'],
 'XMLWriter::startAttribute' => ['bool', 'name'=>'string'],
-'XMLWriter::startAttributeNS' => ['bool', 'prefix'=>'string', 'name'=>'string', 'uri'=>'string'],
-'XMLWriter::startCData' => ['bool'],
+'XMLWriter::startAttributeNs' => ['bool', 'prefix'=>'string', 'name'=>'string', 'namespace'=>'?string'],
```
weirdan, line 15663: I thought you said `$prefix` was nullable?

dictionaries/CallMap.php
```diff
@@ -15643,46 +15643,46 @@
...
 'XMLWriter::startElement' => ['bool', 'name'=>'string'],
-'XMLWriter::startElementNS' => ['bool', 'prefix'=>'string', 'name'=>'string', 'uri'=>'string'],
-'XMLWriter::startPI' => ['bool', 'target'=>'string'],
+'XMLWriter::startElementNs' => ['bool', 'prefix'=>'?string', 'name'=>'string', 'namespace'=>'?string'],
+'XMLWriter::startPi' => ['bool', 'target'=>'string'],
 'XMLWriter::text' => ['bool', 'content'=>'string'],
 'XMLWriter::writeAttribute' => ['bool', 'name'=>'string', 'value'=>'string'],
-'XMLWriter::writeAttributeNS' => ['bool', 'prefix'=>'string', 'name'=>'string', 'uri'=>'string', 'content'=>'string'],
-'XMLWriter::writeCData' => ['bool', 'content'=>'string'],
+'XMLWriter::writeAttributeNs' => ['bool', 'prefix'=>'string', 'name'=>'string', 'namespace'=>'string', 'value'=>'string'],
```
weirdan, line 15676: The types here should probably match `startAttributeNs()`, shouldn't they?

dictionaries/CallMap.php
aboks: Yes, but in the case of `startAttributeNs` it only is as of PHP 8.0: https://www.php.net/manual/en/xmlwriter.startattributens.php#refsect1-xmlwriter.startattributens-changelog. That's why I also added it to the 8.0-delta.callmap: https://github.com/vimeo/psalm/pull/6099/files#diff-5611cd00854405fe61befddd211dacd225d9c479b499ffd7e3c0c7d96eb568edR267

dictionaries/CallMap.php
aboks: You're right that `namespace` should be nullable; I fixed that now. For the prefix, the same thing holds as for `startAttributeNs`: it is only nullable as of PHP 8.0. Although it is not mentioned in the manual, I found this confirmed in the corresponding [bug report](https://bugs.php.net/bug.php?id=79344) and [commit in PHP](http://git.php.net/?p=php-src.git;a=commit;h=aa754ba85e8241c0e59627c359d2c002a5caf8db).

dictionaries/CallMap.php
weirdan: And `CallMap.php` targets PHP 8.0, while delta-8.0 is used to reconstruct callmap for 7.4, delta-7.4 is used to reconstruct callmap for 7.3 and so on. So it should be nullable here.

aboks: Ah, I misunderstood how the callmap deltas work then. I thought they were applied as upward deltas on top of `CallMap.php`, but they are applied as downward deltas. That makes sense. I fixed that now.

I probably got confused because `CallMap.php` already contained the old PHP 7.4 signatures for `xmlwriter_*` (which accept `resource`) instead of the new PHP 8.0 signatures where the `resource` type has been replaced by `XMLWriter`. I also fixed those.

weirdan:
@aboks I've rebased the PR and added a few fixes

weirdan:
Thanks!

aboks:
Great, thank you!

PR #6104, opened 2021-07-15
Detailed return type for `strtoupper`

staabm:


orklah:
Mind adding a nested conditional for `non-falsy-string` too?

It should be checkef before `non-empty-string` because it's more narrow

staabm:
Like that?

orklah:
Perfect, thanks :)

weirdan (approved):

weirdan:
Thanks!

Issue #6075, opened 2021-07-12
Feature: Psalm pseudo-type for `instantiable-object`


boesing:
Hey there,

not sure if thats already possible but is there a way to provide something like this?

https://psalm.dev/r/16e0076766

`Foo::class` is the Interface and thus matches the requirements of the template. But it wont be instantiable because its the interface and thus leads to runtime errors.

What I want to have is something like:

```php
/**
 * @template T of instantiable<Foo>
 */
```

So `instantiable` would match for class names which provide either `final __construct` or anything else which is already verified by `UnsafeInstantiation`.

Is this something which could be achieved?

weirdan:
This actually looks like a bug to me. Not only because we don't know whether the class-string is instantiable, but also because constructor signature is unknown. Even if it's instantiable, it may have required arguments and would still fail at runtime.

muglug:
Yeah, in retrospect making interface names pass a class-string check was a mistake. I think that could be a major-version change.

muglug:
actually it probably doesn't need to be a major-version change

muglug:
This is quite a big refactor, I'll have a look at it this week

muglug:
Actually this is more complex. I’m afraid introducing the change before a major version would induce gnashing of teeth.

I foresee this change happening in two steps: the first very minor step makes `interface-string<X>` a synonym for `class-string<X>`. We should ask library authors to add `interface-string<Foo>` wherever interface strings are acceptable (e.g in templates assertions) for forward-compatibility.

This will help prepare for the next step, which would happen in a major version change, changing the meaning of `class-string` to mean _only_ classes.

boesing:
Sounds reasonable and I'd be fine with that.
 
Shall we move this to a dedicated issue instead? This issue was initially meant to be a feature request which tells psalm that only classes are accepted which are safely instantiable without any `__construct` arguments.

So one can simply do `new $className` without being afraid that it crashes.

But I see the pitfalls, so maybe keeping it to track (even for me) so this can be probably achieved later on. But I'd be fine with closing this as `wont do`. But to have a clear issue ticket regarding the `class-string`, `interface-string` problematic would probably be more clear? What do you think?

Issue #4749, opened 2020-12-01
`class_implements` should return class-string type


staabm:
https://psalm.dev/r/886622aa50

didn't expect 
```
ERROR: ArgumentTypeCoercion - 21:4 - Argument 1 of p expects array<array-key, class-string>, parent type non-empty-array<string, string> provided
```

orklah:
This is easy to fix but I'm not sure what the long term vision is here.

There are two string pseudo-types: `class-string` and `trait-string` and there is a string pseudo-pseudo-type `interface-string` who is interpreted exactly as a `class-string`. This means, `new class_implements(HelloWorld::class)[0]()` would actually be valid code for Psalm.

I'm not sure why there isn't a full `interface-string` so I'm not sure this is a good idea.

PR #6106, opened 2021-07-15
Support interface-string<Foo> as temporary alias of class-string<Foo>

muglug:
Ref #6075 

This allows library authors to hint functions as accepting `interface-string` so things won't break when `class-string` stops accepting interface strings.

## Future work

Transition to using `interface-string` to exclusively mean string class constants (example branch here: https://github.com/vimeo/psalm/commits/interface-string-support)


orklah:
We could add interface-string to the return of class_implements. It would start to solve https://github.com/vimeo/psalm/issues/4749

PR #6116, opened 2021-07-16
radius_send_request returns int|false

fkooman:
https://github.com/LawnGnome/php-radius/blob/master/radius.c#L533


weirdan:
Thanks!

Issue #6111, opened 2021-07-15
obvious WTF UnusedVariable case


williamdes:
https://psalm.dev/r/d4c10c74f0

What is wrong ?, really this one is special 😄 

Source: https://github.com/phpmyadmin/phpmyadmin/blob/cab899d7b5716559d025a0751ec2b8d763bd5111/libraries/classes/Utils/HttpRequest.php#L66

orklah:
simplified:
https://psalm.dev/r/e8c7c7b638

Psalm should consider a variable fetched with global behave similarly to a reference. Any modification made to the variable is "using" the variable. Even if there are no other use of the variable in the current scope

orklah:
Note that if your code use globals, there's a config in your xml file to help Psalm figure out what's in each variable: https://psalm.dev/docs/running_psalm/configuration/#globals
It should help a lot in the search for Mixed variables

williamdes:
> Note that if your code use globals, there's a config in your xml file to help Psalm figure out what's in each variable: https://psalm.dev/docs/running_psalm/configuration/#globals
> It should help a lot in the search for Mixed variables

Thank you so much !

It also triggers without the global: https://psalm.dev/r/fa23212ad8

orklah:
> It also triggers without the global: https://psalm.dev/r/fa23212ad8

Yeah, but without the global, it's legit. You effectively don't use $cfg after so there's no need to push content in it

PR #6129, opened 2021-07-17
Unused global variable after modification

orklah:
This would fix https://github.com/vimeo/psalm/issues/6111#issuecomment-881064978

This basically registers every global variable and prevent emitting UnusedVariable on those.

This means true positives (where UnusedVariable was emitted and the variable was not used) won't work anymore: https://psalm.dev/r/edb1c159ea. Hence the failing test here

However, this bring the behaviour closer to what already exists for references: https://psalm.dev/r/67292ae85a

Going beyond that would mean Psalm would have to record not only variable read access detecting usage but also write access. This seems a big piece of work for detecting a few globals and references unused.

If this is good enough to be merged, I'll remove the failing test and this will become the new behaviour for those cases.

### DIFF
--- a/src/Psalm/Context.php
+++ b/src/Psalm/Context.php
@@ -370,7 +370,7 @@ class Context
     /**
      * @var array<string, bool>
      */
-    public $vars_from_global;
+    public $vars_from_global = [];
 
     public function __construct(?string $self = null)
     {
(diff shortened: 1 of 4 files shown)


weirdan (approved):

weirdan:
> If this is good enough to be merged,

I think it is.

orklah:
Thanks :)

Test is removed

weirdan:
Thanks!

Issue #6117, opened 2021-07-16
Flag accessing static properties with `->`


weirdan:
https://psalm.dev/r/8df7ccd811
https://3v4l.org/iRj67

Expected: `UndefinedProperty`

weirdan:
The opposite should also be checked: https://psalm.dev/r/d79f2332b6

PR #6130, opened 2021-07-17
Flag staticness mismatch

weirdan:
This handles two new cases:
1. Accessing static property with `->` (produces notices and warnings: https://3v4l.org/TiGan)
2. Accessing non-static property with `::` (causes fatal error: https://3v4l.org/IdYSh)

Fixes vimeo/psalm#6117


src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/InstancePropertyAssignmentAnalyzer.php
```diff
@@ -1148,7 +1148,12 @@ private static function analyzeAtomicAssignment(
             $statements_analyzer,
             $context,
             new CodeLocation($statements_analyzer->getSource(), $stmt)
-        )) {
+        )
+        // when methods existance is asserted by a plugin it doesn't necessarily has storage
```
orklah, line 1152: methods should be replaced by properties (and existence is spelled with an e)

muglug (approved):
Looks great!

PR #6132, opened 2021-07-18
remove self analyse from Circle CI jobs

### LINKED ISSUE #6105: CI and Psalm checking itself
Looking at CI running on my PR, I noticed that 2 jobs seem to do the same thing:
https://github.com/vimeo/psalm/blob/master/.circleci/config.yml#L11
and
https://github.com/vimeo/psalm/blob/master/.github/workflows/shepherd.yml#L6

Given that shepherd need to run Psalm on itself in order to work, the CircleCI seem redundant (in addition on being triggered later and using a third party site).

Could we drop the CircleCI one?

orklah:
fix #6105

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,21 @@ jobs:
     executor: php-74
     steps:
       - checkout
+
+      - run: date "+%F" > /tmp/cachekey; cat composer.json >> /tmp/cachekey
+
+      - restore_cache:
+          keys:
+            - composer-v4
+
       - run: composer update
+
+      - save_cache:
+          key: composer-v4
+          paths:
+            - /home/docker/.composer/cache/files
+            - /home/docker/.composer/cache/vcs
+
       - run:
           name: Code Style Analysis with PHPCS
           command: vendor/bin/phpcs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,8 @@ jobs:
   "Code Style Analysis":
     executor: php-74
     steps:
+      - checkout
+      - run: composer update
       - attach_workspace:
           at: /home/docker/project/
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,8 +13,6 @@ jobs:
     steps:
       - checkout
       - run: composer update
-      - attach_workspace:
-          at: /home/docker/project/
       - run:
           name: Code Style Analysis with PHPCS
           command: vendor/bin/phpcs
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -8,41 +8,11 @@ executors:
...
-
-      - persist_to_workspace:
-          root: /home/docker/project/
-          paths:
-            - .
-  "Code Style Analysis":
-    executor: php-74
-    steps:
       - attach_workspace:
           at: /home/docker/project/
```
weirdan, line 17: As it's the first task being executed is there any point of attaching a workspace?

.circleci/config.yml
```diff
@@ -8,41 +8,11 @@ executors:
...
-            - /home/docker/.composer/cache/vcs
-
-      - run:
-          name: Static analysis
-          command: php -dextension=pcntl.so ./psalm --threads=10
-
-      - save_cache:
-          key: psalm-cache-{{ checksum "/tmp/cachekey" }}
-          paths:
-            - /tmp/psalm
```
weirdan, line 37: These `save_cache` calls were probably quite useful

.circleci/config.yml
weirdan: And there's no tasks left that persist anything to workspace.

weirdan: And subsequent tasks attaching a workspace would have it empty, I suppose.

weirdan (changes requested):

.circleci/config.yml
orklah: Seemed to me like they were used between builds. They're not mentioned anywhere else. Should I keep them anyway on the next job?

.circleci/config.yml
orklah: I removed it but to be totally honest, I don't know what I'm doing!

I thought this would be simpler to delete a single job :p

Any idea why the build fails?

.circleci/config.yml
weirdan: I think composer cache should be used by the job that does `composer install`. It speeded up both the `composer install` (it wouldn't need to fetch most of packages afresh) and Psalm itself. Now that the self-check is not performed we can drop psalm cache, but composer cache is still useful.

.circleci/config.yml
weirdan: `persist_workspace` and `attach_workspace` is a way to share files between jobs. The build fails because it can't find the `bin/build-phar.sh` file. And the file is not there because the job neither did the checkout itself, nor attached a workspace that had this file.

Previously, `install-and-self-analyze` would do the checkout and then persist the checked out folder to the workspace that subsequent jobs were using. Now it doesn't happen.

orklah: Thanks, I get it now

weirdan (approved):

weirdan:
Thanks!

Issue #6110, opened 2021-07-15
 `trigger_error` makes Psalm assume the code exits


williamdes:
> This one is similar and psalm thinks that `trigger_error` makes the code exit I suppose: 

https://psalm.dev/r/38c501845b

Ref: #6109

_Originally posted by @orklah in https://github.com/vimeo/psalm/issues/6109#issuecomment-881058195_

orklah:
@muglug @weirdan Do you have an opinion?
Psalm considers that trigger_error stop the execution here https://github.com/vimeo/psalm/blob/0a7d25468900d356ddf2f0e033b95eafbd7bdff8/src/Psalm/Internal/Analyzer/ScopeAnalyzer.php#L107
but not everywhere. (see #6109). What should be the correct behaviour?

williamdes:
Note: we custom handle calls for trigger_error

weirdan:
`trigger_error()` return type (specifically cases when it returns `never`) depends on the error handler behaviour. Since that's external to Psalm and cannot be easily inferred, Psalm should have a sensible default matching default error handler behaviour and let users override that. I see two options here:
1. Let users stub `trigger_error()` and use the stub signature consistently. In practice this means we need to remove hardcoded handling we have currently, and just provide a default, overridable, stub.
2. Add an option that would specify error handler signature, and infer `trigger_error()` return type from that. It would also allow enforcing error handler compatibility.

orklah:
I like the second option. trigger_error has a default behaviour that is valid for a lot of project that don't touch set_error_handler and when the error_handler use the same behaviour as the default one.
That means we could have an option whose default means that Psalm will use the default behaviour of trigger_error. Other values for this option could be "none" where trigger_error never `die` and "always" where it always `die`.

I think this could be handled with a return type provider accessing config, right?

weirdan:
> I think this could be handled with a return type provider accessing config, right?

Sounds good.

>  That means we could have an option whose default means that Psalm will use the default behaviour of trigger_error. Other values for this option could be "none" where trigger_error never die and "always" where it always die.

This sounds sensible, but so is assuming `trigger_error(..., E_USER_ERROR)` never returns :shrug:.

Let's explore the problem space:

[...]

PR #6142, opened 2021-07-19
Trigger error

orklah:
This will solve #6110

I'm marking this as a draft because I'm waiting for the review before adding the documentation on this.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -89,6 +89,7 @@
         <xs:attribute name="reportInfo" type="xs:boolean" default="true" />
         <xs:attribute name="restrictReturnTypes" type="xs:boolean" default="false" />
         <xs:attribute name="limitMethodComplexity" type="xs:boolean" default="false" />
+        <xs:attribute name="triggerErrorExits" type="TriggerErrorExitsType" default="default" />
     </xs:complexType>
 
     <xs:complexType name="ProjectFilesType">
@@ -644,6 +645,14 @@
             <xs:enumeration value="error"/>
             <xs:enumeration value="info"/>
             <xs:enumeration value="suppress"/>
+        </xs:restriction>
+    </xs:simpleType>
+
+    <xs:simpleType name="TriggerErrorExitsType">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="default"/>
+            <xs:enumeration value="none"/>
+            <xs:enumeration value="always"/>
         </xs:restriction>
     </xs:simpleType>
 </xs:schema>
--- a/config.xsd
+++ b/config.xsd
@@ -651,7 +651,7 @@
     <xs:simpleType name="TriggerErrorExitsType">
         <xs:restriction base="xs:string">
             <xs:enumeration value="default"/>
-            <xs:enumeration value="none"/>
+            <xs:enumeration value="never"/>
             <xs:enumeration value="always"/>
         </xs:restriction>
     </xs:simpleType>
--- a/src/Psalm/Internal/Analyzer/ScopeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ScopeAnalyzer.php
@@ -11,7 +11,6 @@ use function array_merge;
 use function array_unique;
 use function array_values;
 use function count;
-use function end;
 use function in_array;
 use function strtolower;
 
--- a/src/Psalm/Internal/Analyzer/ScopeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ScopeAnalyzer.php
@@ -102,19 +102,6 @@ class ScopeAnalyzer
             }
 
             if ($stmt instanceof PhpParser\Node\Stmt\Expression) {
-                if ($stmt->expr instanceof PhpParser\Node\Expr\FuncCall
-                    && $stmt->expr->name instanceof PhpParser\Node\Name
-                    && $stmt->expr->name->parts === ['trigger_error']
-                    && isset($stmt->expr->args[1])
-                    && $stmt->expr->args[1]->value instanceof PhpParser\Node\Expr\ConstFetch
-                    && in_array(
-                        end($stmt->expr->args[1]->value->name->parts),
-                        ['E_ERROR', 'E_PARSE', 'E_CORE_ERROR', 'E_COMPILE_ERROR', 'E_USER_ERROR']
-                    )
-                ) {
-                    return array_values(array_unique(array_merge($control_actions, [self::ACTION_END])));
-                }
-
                 // This allows calls to functions that always exit to act as exit statements themselves
                 if ($nodes
                     && ($stmt_expr_type = $nodes->getType($stmt->expr))
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
@@ -2,6 +2,7 @@
 
 namespace Psalm\Internal\Provider\ReturnTypeProvider;
 
+use Psalm\Internal\Type\TypeCombiner;
 use Psalm\Plugin\EventHandler\Event\FunctionReturnTypeProviderEvent;
 use Psalm\Type;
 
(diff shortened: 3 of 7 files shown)


config.xsd
```diff
@@ -89,6 +89,7 @@
         <xs:attribute name="reportInfo" type="xs:boolean" default="true" />
         <xs:attribute name="restrictReturnTypes" type="xs:boolean" default="false" />
         <xs:attribute name="limitMethodComplexity" type="xs:boolean" default="false" />
+        <xs:attribute name="triggerErrorBehaviour" type="TriggerErrorBehaviourType" default="default" />
```
weirdan, line 92: `behaviour` is a British spelling, while Psalm code usually uses American. However I would suggest to change it to `triggerErrorExits`, as `exits=always` seems clearer than `behavior=always` to me.

src/Psalm/Internal/Analyzer/ScopeAnalyzer.php
```diff
@@ -104,7 +104,6 @@ public static function getControlActions(
             if ($stmt instanceof PhpParser\Node\Stmt\Expression) {
                 if ($stmt->expr instanceof PhpParser\Node\Expr\FuncCall
                     && $stmt->expr->name instanceof PhpParser\Node\Name
-                    && $stmt->expr->name->parts === ['trigger_error']
                     && isset($stmt->expr->args[1])
                     && $stmt->expr->args[1]->value instanceof PhpParser\Node\Expr\ConstFetch
                     && in_array(
```
weirdan, line 109: The whole `if` doesn't make sense without that line, and needs to be dropped. You don't want to enter that `if` for every call to any function where second argument is one of `E_*` constants.

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
```diff
@@ -0,0 +1,67 @@
...
+                        $return_types[] = new Type\Atomic\TTrue();
+                    } elseif ($atomicType->value === E_USER_ERROR) {
+                        $return_types[] = new Type\Atomic\TNever();
+                    } else {
+                        // not recognized int literal. return false before PHP8, fatal error since
+                        $return_types[] = new Type\Atomic\TFalse();
+                    }
+                } elseif ($atomicType instanceof Type\Atomic\TInt) {
+                    // not specific integer. Returning a mix of never and bool would not make sense for Psalm
+                    $return_types[] = new Type\Atomic\TMixed();
```
weirdan, line 55: So that should be `bool` then.

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
```diff
@@ -0,0 +1,67 @@
...
+                } else {
+                    $return_types[] = new Type\Atomic\TMixed();
```
weirdan, line 57: And here as well.

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
```diff
@@ -0,0 +1,67 @@
...
+                }
+            }
+
+            return new Type\Union($return_types);
```
weirdan, line 61: Shouldn't we first simplify `never|bool` to just `bool`, etc? `Union` doesn't do that automatically. There should be an existing method for that, but specific name escapes me at the moment.

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
```diff
@@ -0,0 +1,67 @@
...
+        $statements_source = $event->getStatementsSource();
+        if (isset($call_args[1])
+            && ($array_arg_type = $statements_source->getNodeTypeProvider()->getType($call_args[1]->value))
+        ) {
+            $return_types = [];
+            foreach ($array_arg_type->getAtomicTypes() as $atomicType) {
+                if ($atomicType instanceof Type\Atomic\TLiteralInt) {
+                    if (in_array($atomicType->value, [E_USER_WARNING, E_USER_DEPRECATED, E_USER_NOTICE], true)) {
+                        $return_types[] = new Type\Atomic\TTrue();
+                    } elseif ($atomicType->value === E_USER_ERROR) {
```
weirdan, line 47: Did those constant values ever change? I know `E_ALL` was changed at some time, so other may have been changed as well.

weirdan (changes requested):
Build error is already fixed on master branch.

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
orklah: I started with bool|never but it lead to Psalm reporting an error when trying to return this type. Bool however seem to narrow, given that never is a possibility, don't you think?

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
orklah: Good thinking, I'll check that

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
orklah: > simplify never|bool to just bool

is it really equivalent? when never is a part of the type, it means that sometimes there will be a crash at runtime if we try to return the result

src/Psalm/Internal/Analyzer/ScopeAnalyzer.php
orklah: Right, I made this change too quickly. I'll fix that

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
weirdan: `never` is a bottom type, so `bool|never` is not wider than `bool`, but is identical to it.

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
weirdan: *If* `trigger_error()` returns, its return value is `bool`. If it doesn't (`never`) then there's no return value.

orklah: you're right. I confused errors from psalm with runtime. I'll fix that

weirdan: >  but specific name escapes me at the moment.

I think it's `TypeCombiner::combine()`

config.xsd
muglug: Yeah, much as I wish everything were British I used American spellings (e.g. SwitchAnalyzer) very much on purpose since so many other programming things use American spellings (e.g. color)

config.xsd
```diff
@@ -646,4 +647,12 @@
             <xs:enumeration value="suppress"/>
         </xs:restriction>
     </xs:simpleType>
+
+    <xs:simpleType name="TriggerErrorExitsType">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="default"/>
+            <xs:enumeration value="none"/>
```
weirdan, line 654: I'd suggest to use antonym to `always` instead, which is `never`.

src/Psalm/Internal/Provider/ReturnTypeProvider/TriggerErrorReturnTypeProvider.php
orklah: No mention of them changing. Only E_ALL due to new constants additions: https://www.php.net/manual/en/errorfunc.constants.php

weirdan (approved):

orklah:
Documentation added. It should be good to merge

weirdan (approved):

weirdan:
Thanks!

PR #6146, opened 2021-07-20
fix: Handle DataFlowNodeData in XmlReport

ndench:
Adding `other_references` and `taint_trade` to [IssueData](https://github.com/vimeo/psalm/blame/d8cddd64c0743b633af67824f4fc6708cb805cc7/src/Psalm/Internal/Analyzer/IssueData.php) has caused a regression in XML report generation. LSS\Array2XML expects a nested array of scalars, but it now sometimes contains DataFlowNodeData objects.

I'm not sure of the best way to do this so I just extended the existing way. Feel free to either edit or leave some suggestions.

### DIFF
--- a/src/Psalm/Report/XmlReport.php
+++ b/src/Psalm/Report/XmlReport.php
@@ -2,6 +2,7 @@
 namespace Psalm\Report;
 
 use LSS\Array2XML;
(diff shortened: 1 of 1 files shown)


weirdan:
Thanks!

PR #6155, opened 2021-07-21
Add support for native PHP readonly

muglug:
Updates nikic/php-parser to [4.12.0](https://github.com/nikic/PHP-Parser/releases/tag/v4.12.0), which provides an extra method supporting `readonly`.


weirdan:
Thanks!

PR #6166, opened 2021-07-23
String increment on numeric string

orklah:
This PR allows incrementing on a numeric-string without triggering StringIncrement as this is most probably on purpose. Also, it fixes the resulting type from `non-empty-string` to `float|int`. This is particularly useful when working with numeric array-keys due to PHP shenanigans that cast string to int

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
@@ -387,12 +387,18 @@ class NonDivArithmeticOpAnalyzer
             && $right_type_part instanceof TInt
             && $parent instanceof PhpParser\Node\Expr\PostInc
         ) {
-            $has_string_increment = true;
-
-            if (!$result_type) {
-                $result_type = Type::getNonEmptyString();
+            if ($left_type_part instanceof Type\Atomic\TNumericString) {
+                $new_result_type = new Type\Union([new TFloat(), new TInt()]);
+                $new_result_type->from_calculation = true;
             } else {
-                $result_type = Type::combineUnionTypes(Type::getNonEmptyString(), $result_type);
+                $new_result_type = Type::getNonEmptyString();
+                $has_string_increment = true;
+            }
+
+            if ($result_type) {
+                $result_type = Type::combineUnionTypes($new_result_type, $result_type);
+            } else {
+                $result_type = $new_result_type;
             }
 
             $has_valid_left_operand = true;
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
@@ -390,7 +390,12 @@ class NonDivArithmeticOpAnalyzer
                 $parent instanceof PhpParser\Node\Expr\PreInc
             )
         ) {
-            if ($left_type_part instanceof Type\Atomic\TNumericString) {
+            if ($left_type_part instanceof Type\Atomic\TNumericString || (
+                    $left_type_part instanceof Type\Atomic\TLiteralString &&
+                    //Matches '1', '1.', '-1', '-1.'. Does not match '.1'
+                    preg_match('/^\-?\d+\.?\d*$/', $left_type_part->value)
+                )
+            ) {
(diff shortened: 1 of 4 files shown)


orklah:
(I've also added documentation for TNonEmptyString and TNonFalsyString in the class because I keep confusing myself between the two and I did it again on this PR when I wondered if "hello"++ should result in TNonEmptyString or TNonFalsyString)

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
```diff
@@ -387,12 +387,18 @@ private static function analyzeNonDivOperands(
             && $right_type_part instanceof TInt
             && $parent instanceof PhpParser\Node\Expr\PostInc
```
weirdan, line 388: And what about preincrement? Doesn't it need a similar change?

orklah:
I added pre inc and also pre/post dec. Thanks

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
orklah: I added pre inc and also pre/post dec. Thanks

weirdan: I'd suggest to add a test for that. Something simple, like asserting result of '123'++, ++'123' etc

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
```diff
@@ -385,14 +385,28 @@ private static function analyzeNonDivOperands(
...
+            && (
+                $parent instanceof PhpParser\Node\Expr\PostInc ||
+                $parent instanceof PhpParser\Node\Expr\PreInc
+            )
         ) {
-            $has_string_increment = true;
+            if ($left_type_part instanceof Type\Atomic\TNumericString || (
+                    $left_type_part instanceof Type\Atomic\TLiteralString &&
+                    //Matches '1', '1.', '-1', '-1.'. Does not match '.1'
+                    preg_match('/^\-?\d+\.?\d*$/', $left_type_part->value)
```
weirdan, line 396: > Does not match '.1'

It does work with increment though: https://3v4l.org/updn7

It isn't terribly important to support that, of course.

weirdan: Doesn't increment work on any string that is `is_numeric()`? If so, and given that Psalm doesn't run on PHP 5 where `is_numeric()` also accepted hex-like strings, can't we just replace that with `is_numeric($left_type_part->value)`?

orklah: Yeah, that's why I stated it explicitly, I didn't saw how I could do a regex to support all that and I didn't want to complexify too much. There is little chance for finding this in the wild so it will stay like this for now :)

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
orklah: I did. This led to some more code to support literals.

I removed PreDec and PostDec though because they're handled totally differently in IncDecExpressionAnalyzer and I couldn't really figure why

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
orklah: I guess we can. I stayed out of it because I don't really master all the subtleties of is_numeric but I don't see harm in it as long as we don't try to guess the exact Literal it generates

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
weirdan: > I did. This led to some more code to support literals.

:+1:

weirdan:
Thanks!

PR #6169, opened 2021-07-24
Trigger InvalidOperand when processing two numeric types in strict operands mode

### LINKED ISSUE #6168: arithmetic operations with numeric should trigger InvalidOperand when strictBinaryOperands is true
This can't be reproduced on psalm.dev but here's the snippet anyway:
https://psalm.dev/r/100b60f446

`numeric` can contain int, float or numeric-string, so when strictBinaryOperands  is enabled, doing operations with it should trigger InvalidOperand

orklah:
this PR fixes #6168 where an operation on a `numeric` type did not trigger InvalidOperand even on strictBinaryOperands mode.

It also improve the wording of the error to mention the flag (I was confused about that when CI on PSL failed) and to remove the term `add` that is improper given that other operations trigger the same issue

This PR should break the build because PSL has at least one place where operations on numeric are performed.


weirdan (approved):

weirdan:
Thanks!

PR #6171, opened 2021-07-24
try to speed up appveyor

orklah:
This PR tries to speed up appveyor by removing things or adding configs I stole here and there :) Let's see if that makes a difference


orklah:
great, it made no difference whatsoever. I'm closing this. I may try to migrate on github actions

PR #6172, opened 2021-07-25
Handle malformed parentheses during type parsing phase

### LINKED ISSUE #6159: Psalm crash on `@return (`
https://psalm.dev/r/b566eeeb37

Jack97:
This PR updates the `TypeParser` to throw a `TypeParseTreeException` when unterminated, empty, or unbalanced parentheses are encountered. Previously, empty parentheses would cause Psalm to crash, and the balance was not being verified so `(((((((((((string` would be incorrectly evaluated as `string`.

Closes #6159


weirdan:
Thanks!

PR #6173, opened 2021-07-25
migrate windows CI from appveyor to GA

orklah:
This PR propose ditching Appveyor in favor of Github Actions.

I did a test without cache and GA seem to be slightly worse in execution time so I hope with cache they'll be equivalent.

The big issue we have with Appveyor is that the queue is shared between all PR and branches in the repo. This means when there is a lot of activity, Appveyor queue can get really long. I don't thing GA has this issue.

I'm not really good with yaml files nor with CI builds so it's probable I made mistakes here.


orklah:
Appveyor launched itself here though I deleted the file in the PR. Not sure if it uses the one in master or if I forgot to delete somethin

weirdan:
> Appveyor launched itself here though I deleted the file in the PR.

I think @muglug needs to disable it in repo settings.

weirdan:
> simplify windows ci 

Where there some problems with running tests in chunks?

orklah:
> Where there some problems with running tests in chunks?

Yeah, It uses this file: https://github.com/vimeo/psalm/blob/master/bin/tests-github-actions.sh
And I had no idea how to translate it for windows

.appveyor.yml
```diff
@@ -0,0 +1,4 @@
+build: false
+branches:
+  only:
+    - dear-appveyor-please-dont-build-this-repo
```
orklah, line 4: 😄

weirdan: `¯\_(ツ)_/¯`

weirdan:
Thanks!

PR #6178, opened 2021-07-26
Add false as return type of shell_exec

### LINKED ISSUE #6177: shell_exec can return `false`
Related to https://github.com/phpstan/phpstan/discussions/5376#discussioncomment-1050042
Seems like unlike is saying the php documentation, shell_exec can return `false` https://github.com/php/php-src/pull/7306

But psalm consider the return type to be `string|null`
https://psalm.dev/r/36a3eb5fa8

Where should I update the return type definition ?

VincentLanglet:
The doc is only talking about string|null, but
false is possible when the pipe can't be etablished
@see https://github.com/php/php-src/pull/7306#issuecomment-886489235

Closes #6177


weirdan:
Thanks!

Issue #6183, opened 2021-07-26
Update output provider to differentiate SonarCloud and SonarQube


nils-werner-sonarsource:
At SonarSource, we like what Psalm provides and we are happy to see that there is a default [provider](https://github.com/vimeo/psalm/blob/master/src/Psalm/Report/SonarqubeReport.php) which exports Psalm data into the SonarSource Generic Issues format.
We want to be sure all our PHP users which use SonarQube or SonarCloud get the information that they can load their Psalm results.

This is why we added a dedicated property "sonar.php.psalm.reportPaths" and our documentation [[SonarQube](https://docs.sonarqube.org/latest/analysis/external-issues/), [SonarCloud](https://sonarcloud.io/documentation/enriching/external-issues/)] will be updated accordingly in the coming days.

On Psalm side, we believe it would be great to clarify that SonarSource provides 2 products: SonarQube and SonarCloud.
Today the Psalm CLI help provide only "sonarqube.json" as a potential output format and users of SonarCloud may think that it's not possible to load Psalm results into SonarCloud.

For these reasons, we would like Psalm to support two different output formats: sonarqube.json and sonarcloud.json

PR #6184, opened 2021-07-26
Update output provider to differentiate SonarSource products

nils-werner-sonarsource:
As described in #6183

At SonarSource, we like what Psalm provides and we are happy to see that there is a default [provider](https://github.com/vimeo/psalm/blob/master/src/Psalm/Report/SonarqubeReport.php) which exports Psalm data into the SonarSource Generic Issues format.
We want to be sure all our PHP users which use SonarQube or SonarCloud get the information that they can load their Psalm results.

This is why we added a dedicated property "sonar.php.psalm.reportPaths" and our documentation [[SonarQube](https://docs.sonarqube.org/latest/analysis/external-issues/), [SonarCloud](https://sonarcloud.io/documentation/enriching/external-issues/)] will be updated accordingly in the coming days. [[SonarSource ticket](https://jira.sonarsource.com/browse/MMF-2457)]

On Psalm side, we believe it would be great to clarify that SonarSource provides 2 products: SonarQube and SonarCloud.
Today the Psalm CLI help provide only "sonarqube.json" as a potential output format and users of SonarCloud may think that it's not possible to load Psalm results into SonarCloud.

For these reasons, we would like Psalm to support two different output formats: sonarqube.json and sonarcloud.json


orklah:
If the reports have the same format, maybe refactor to have a common method that generate both? Maybe a common parent between the two Report classes?

src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
```diff
@@ -366,6 +366,7 @@ public static function getFileReportOptions(array $report_file_paths, bool $show
         $mapping = [
             'checkstyle.xml' => Report::TYPE_CHECKSTYLE,
             'sonarqube.json' => Report::TYPE_SONARQUBE,
+            'sonarcloud.json' => Report::TYPE_SONARCLOUD,
```
weirdan, line 369:
```suggestion
            'sonarcloud.json' => Report::TYPE_SONARQUBE,
```

src/Psalm/Internal/CliUtils.php
```diff
@@ -395,7 +395,7 @@ public static function getPsalmHelpText(): string
 
     --output-format=console
         Changes the output format.
-        Available formats: compact, console, text, emacs, json, pylint, xml, checkstyle, junit, sonarqube, github,
+        Available formats: compact, console, text, emacs, json, pylint, xml, checkstyle, junit, sonarqube, sonarcloud, github,
                            phpstorm, codeclimate
```
weirdan, line 399:
```suggestion
        Available formats: compact, console, text, emacs, json, pylint, xml, checkstyle, junit, sonarqube, sonarcloud, github,
                           phpstorm, codeclimate
        
        'sonarqube' format is suitable for both SonarQube and SonarCloud
```

weirdan (changes requested):

weirdan (changes requested):
With the changes suggested above you should be able to drop all the other changes and still get the same result.

src/Psalm/Internal/CliUtils.php
nils-werner-sonarsource: If I understand it correctly we need to keep both Report Types to keep both `--output-format`. We would prefer to offer both names as a format instead of including an extra hint in the help text.

nils-werner-sonarsource (commented):
I have reduced the two report class to a common SonarSourceReport class. As long as there is no difference in the report structure this reduces the overhead, as already suggested by orklah.

nils-werner-sonarsource:
Are there any comments or further suggestions? What conditions still need to be met for the PR to be accepted?

nils-werner-sonarsource:
Any updates for this PR?

weirdan:
Yeah, I decided to opt for help output change instead, basically what I suggested in https://github.com/vimeo/psalm/pull/6184#discussion_r676882171. Hasn't got around to actually implement that though.

PR #6185, opened 2021-07-26
fix wrong wording from #6083

orklah:
I made a mistake in #6083. The $param_type is what's expected in the function, not what's provided


weirdan:
Thanks!

PR #6187, opened 2021-07-26
Inaccurate error with different types in PHPDoc and typehint

### LINKED ISSUE #6128: Inaccurate error with different types in PHPDoc and typehint
Link to the playground: https://psalm.dev/r/616eb6fd60

image

It seems that there should not be a `null|string` type here, since the type in PHPDoc `@return` is just `string`.
This also happens if you set the same types for the parameters of a function or properties of a class.

P.S. Thanks for the great tool! Psalm gives me a lot of inspiration.

orklah:
This fixes #6128

After running the snippet in the original issue, we now get this result:

> ERROR: MismatchingDocblockReturnType - sandbox.php:4:12 - Docblock has incorrect return type 'string', should be 'int|null' (see https://psalm.dev/142)
> * @return string
>
>
>ERROR: InvalidReturnType - sandbox.php:4:12 - The declared return type 'string' for getInt is incorrect, got 'int|null' (see https://psalm.dev/011)
> * @return string

The MismatchingDocblockReturnType is indeed fixed. Note that a new error is emitted that basically sees the error from another point of view. Not sure if that's expected. EDIT: I believe it's expected. I tried and when there are three different types (signature + docblock + code), the two errors are consistent so it's not just a duplicate

I also fixed a bug I made in ReturnTypeCollector a few months ago that was messing with the second error 

I didn't add test because it's changing wording and I don't think we already have tests that do so.

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeCollector.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeCollector.php
@@ -44,7 +44,7 @@ class ReturnTypeCollector
                 } elseif ($stmt->expr instanceof PhpParser\Node\Scalar\String_) {
                     $return_types[] = Type::getString();
                 } elseif ($stmt->expr instanceof PhpParser\Node\Scalar\LNumber) {
-                    $return_types[] = Type::getString();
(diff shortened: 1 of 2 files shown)


weirdan:
Thanks!

PR #6188, opened 2021-07-27
Consider print, and conditionally exit / die impure

### LINKED ISSUE #6136: `print`, `exit`, and `die` effect IO but they are considered pure.
ref: https://psalm.dev/r/f3713595ce

Jack97:
I've updated the `PrintAnalyzer` to raise an `ImpureFunctionCall` issue in a mutation-free context.

I've updated the `ExitAnalyzer`  to do the same thing, but only when a non-integer argument is provided.

Closes #6136


weirdan:
Thanks!

PR #6192, opened 2021-07-27
make array_replace and array_replace_recursive not nullable

### LINKED ISSUE #6186: array_replace_recursive() return type does not contain NULL
`array_replace_recursive()` are documented to return `NULL` "if an error occurs", but Psalm treats it as it could only return an array and not `NULL`:

https://psalm.dev/r/66aaa94d25

also see issues #2925

PHP reference: https://www.php.net/manual/de/function.array-replace-recursive.php

orklah:
This will fix #6186. The only way those function return null is when a wrong parameter is given in input. These kind of errors will be flagged by Psalm and have become a runtime error since PHP8.


weirdan:
Thanks!

Issue #6156, opened 2021-07-21
Taint analysis inaccurate if parent class not defined


mmcev106:
In [this example](https://psalm.dev/r/98bf106bbc), no taints are displayed, even though I suspect they can & should be.  If you uncomment the parent class definition, they will display.

I work in an environment where PHP "modules" are loaded very dynamically (somewhat analogous to WordPress plugins).  It would be safest if we could run a taint analysis on individual modules, without requiring all their dependencies to be defined.  When I comment [this line](https://github.com/vimeo/psalm/blob/4.8.1/src/Psalm/Internal/Analyzer/ClassAnalyzer.php#L339), the taints will display anyway.  I believe this means this is technically do-able, though I'm not sure what the appropriate PR would look like...

If displaying taints regardless of parent class availability is not possible, would it make sense to display the `UndefinedClass` error that is displayed when [running psalm normally](https://psalm.dev/r/0b547e5c43)? As it stands, running a taint-analysis currently reports "No issues!" without any indication that some code was not really scanned due to undefined classes.

orklah:
Running Psalm without some dependancies is weird. For a simple example like this one that doesn't interact with the dependancy, it's ok, but as soon as you have a `$this->callMethodInParent()`, Psalm will be quite lost.

I think the best possible solution for you would be to load stubs for your possibly undefined classes or to run an autoloader that will load your modules as needed

mmcev106:
@orklah, that makes sense.  What are your thoughts on reporting `UndefinedClass` when a dependency is missing during taint analysis?

orklah:
If we go this way, we'll have to add a bunch of other issues to stay consistent. There more than 15 Undefined* issues (like UndefinedInterface, UndefinedTrait, UndefinedMethod) that would most certainly have impact on taint analysis, but it won't stop there. InvalidDocblock for example will break taint analysis if there's taint specific tags in the docblock.

I think if we take this path, we'll probably regret it down the line because we'll have to judge wether an issue should or shouldn't be emitted in taint analysis and we'll have no clear choice.

Fortunately, Undefined* issues are already being reported even at high levels so it should be the first issues to be fixed in a given project so I don't think it will have a big impact. If there's situations were those issue can't be fixed, it would be best to work on that instead.

mmcev106:
Does that mean that it is never recommended to run ONLY the taint analysis without running Psalm normally first (as the taint analysis may incorrectly report no errors)?

weirdan:
[...]

PR #6198, opened 2021-07-28
Clarified when it is safe to run taint analysis

mmcev106:
This is attempt to prevent the confusion I had on what the expectation was around [this issue](https://github.com/vimeo/psalm/issues/6156).  I'm happy to make any changes requested.

### DIFF
--- a/docs/security_analysis/index.md
+++ b/docs/security_analysis/index.md
@@ -48,7 +48,7 @@ You can also [define your own taint sources](custom_taint_sources.md).
 
 ## Taint Sinks
 
-Psalm currently defines a number of different for builtin functions and methods, including `echo`, `include`, `header`.
+Psalm currently defines a number of different sinks for builtin functions and methods, including `echo`, `include`, `header`.
 
 You can also [define your own taint sinks](custom_taint_sinks.md).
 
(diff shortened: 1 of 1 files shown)


docs/security_analysis/index.md
```diff
@@ -2,7 +2,7 @@
 
 Psalm can attempt to find connections between user-controlled input (like `$_GET['name']`) and places that we don’t want unescaped user-controlled input to end up (like `echo "<h1>$name</h1>"` by looking at the ways that data flows through your application (via assignments, function/method calls and array/property access).
 
-You can enable this mode with the `--taint-analysis` command line flag. When taint analysis is enabled, no other analysis is performed.
+You can enable this mode with the `--taint-analysis` command line flag. When taint analysis is enabled, no other analysis is performed.  To [ensure accurate results](https://github.com/vimeo/psalm/issues/6156), Psalm should be run normally prior to taint analysis, and any errors should be fixed.
```
orklah, line 5: I'd use the term `comprehensive` instead of `accurate`. Psalm is accurate in what it reports, we don't want users to think they're doing harm, it's just that the list of vulnerability is possibly not complete

mmcev106: Hmm... From my perspective in this case, Psalm inaccurately implied that all files were being scanned when they were not, and inaccurately reported "No errors found", which lead to a real risk of harm by approving & deploying code that contained vulnerabilities.

Y'all are in charge of course...  If I don't hear back in a couple of days I'll make the change as requested.

mmcev106: The language has been changed as requested.

orklah (approved):

weirdan:
Thanks!

PR #6204, opened 2021-07-29
Class string attributes

### LINKED ISSUE #4871: class-strings are misinterpreted as strings in attribute constructor parameters
When an attribute's constructor takes in a class-string and you pass one in, it misinterprets it as a normal string.

https://psalm.dev/r/18c45b7a8c

### LINKED ISSUE #6197: ArgumentTypeCoercion when using string-class<Interface> in a php attribute
When using a php (8) attribute that expects a single parameter of type `class-string<Foobar>` in its constructor it returns an `ArgumentTypeCoercion`-error. The error incorrectly states I am passing a `parent` class name while in fact it's a `child` class name.

See https://psalm.dev/r/e6976b117f (took the snippet from https://github.com/vimeo/psalm/issues/4871 that might be related to this issue)

orklah:
This fixes #6197 and fixes #4871

However, the original snippets produce a new error `UnusedProperty - sandbox.php:6:20 - Cannot find any references to private property Foo::$className`. I'm not sure if it's a false positive because the property is indeed unused but I couldn't reproduce without attributes involved.

### DIFF
--- a/src/Psalm/Internal/Stubs/Generator/StubsGenerator.php
+++ b/src/Psalm/Internal/Stubs/Generator/StubsGenerator.php
@@ -6,6 +6,7 @@ use PhpParser;
 use Psalm\Internal\Scanner\ParsedDocblock;
 use Psalm\Node\Expr\VirtualArray;
 use Psalm\Node\Expr\VirtualArrayItem;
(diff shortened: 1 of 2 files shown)


weirdan:
Thanks!

PR #6205, opened 2021-07-29
fix typos

orklah:
I saw a typo when making a change but then I forgot about what it was, so when I was searching for it, I found a bunch so I fixed them too :)


weirdan:
Thanks!

PR #6206, opened 2021-07-29
Update return types for get_headers and get_meta_tags

### LINKED ISSUE #6199: get_headers: array<array-key, mixed> can never contain false
https://psalm.dev/r/c9c058d764

pardon, i do myself hard to understand this error, because the [PHP documentation](https://www.php.net/manual/en/function.get-headers.php) clearly says: 

> Returns an indexed or associative array with the headers, or false on failure.

what am I missing here ?

thanks for any help!

Jack97:
Fixes #6199, and I also fixed the incorrect return type for the `get_meta_tags` function.

https://www.php.net/manual/en/function.get-headers
https://www.php.net/manual/en/function.get-meta-tags

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1266,16 +1266,20 @@ function base64_encode(string $data) : string {}
  *
  * @param resource|null $context
  *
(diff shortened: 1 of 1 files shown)


weirdan:
Thanks!

michabbb:
👍 🎉 💪

PR #6207, opened 2021-07-30
Int range

orklah:
This PR adds basic functionality for integer range, as they were added in PHPStan in their latest release.

This will allow Psalm to read them, transforming them back to known types when large enough(positive-int, int) and to perform very basic operations.

This is the first step of many. To have full support, we will have to work on:
- docblock assertions
- combine/intersect
- internal assertions
- for loops with boundaries

at least


weirdan:
Thanks!

PR #6208, opened 2021-07-31
Improve internals of TIntRange

orklah:
The way I designed the internals of TIntRange was dumb. Trying to fit a single string with a meaning along the integer value is not pratical and not elegant. This replace the string value will null.
Can we push that before considering this a BC break of the release from 3 hours ago?


orklah:
(I also fixed an issue I made that would have transformed and int<0, ?> to TPositiveInt but 0 should not be included)

muglug (approved):

weirdan:
@orklah this requires an expedite patch release, right? Otherwise people may start using what's now in 4.9

orklah:
Yeah, this would be great. I'm not too worry about people trying to use it as is a few hours after the release, but if we leave it for days, it might

weirdan:
4.9.1 has just been tagged.

PR #6210, opened 2021-07-31
Assertions for Integer Range

### LINKED ISSUE #6209: Deprecate TPositiveInt for removal in Psalm 5?
Hi,

I want to talk about TPositiveInt and it's future. With the introduction of TIntRange, it has become a redundancy of a more precise type. I checked how it's handled in PHPStan, and it seems like "positive-int" and "negative-int" are just aliases of their Integer Range type and that makes sense.

This would probably help issues like https://github.com/vimeo/psalm/issues/6020 that are really hard to do with the current internal assertions system.

Unfortunately, we can't just remove TPositiveInt as there are plugins (including mine) that uses it, so we have to go through a deprecation process. However, the enhancement I intend to do on TIntRange will probably be made different whether or not we want to keep TPositiveInt in the long term, so I'm creating this issue so we can decide something

orklah:
This is another batch of changes for Integer Ranges

- It deprecates TPositiveInt (and fixes #6209)
- it introduces assertions for Smaller and Greater operations to reduce Integer Ranges
- It allows checks for whether an IntRange is contained into other ints and the other way around 

Unfortunately, right now, it makes tests crash because of this horror: https://github.com/vimeo/psalm/blob/master/tests/TypeReconciliation/TypeAlgebraTest.php#L785

I suppose it's missing some kind of threshold trigger but I'm not sure where I should put it to have minimal impact.
EDIT: I made a blackfire profile of the job (slightly cut so it still runs): https://blackfire.io/profiles/1134ee3f-8529-4296-a980-fa7fb64dd49f/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main()
EDIT: I added a threshold on Algebra::combineOredClauses

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -93,7 +93,7 @@ class AssertionFinder
 
             if ($var_name) {
                 if ($candidate_if_types) {
-                    $if_types[$var_name] = [['&' . \json_encode($candidate_if_types[0])]];
+                    $if_types[$var_name] = [['@' . \json_encode($candidate_if_types[0])]];
                 } else {
                     $if_types[$var_name] = [['!falsy']];
                 }
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -3730,7 +3730,7 @@ class AssertionFinder
             }
 
             if ($var_name) {
-                $if_types[$var_name] = [['+' . $superior_value_comparison]];
+                $if_types[$var_name] = [['=isset'], ['+' . $superior_value_comparison]];
             }
 
             return $if_types ? [$if_types] : [];
@@ -3937,7 +3937,7 @@ class AssertionFinder
             }
 
             if ($var_name) {
-                $if_types[$var_name] = [['-' . $inferior_value_comparison]];
+                $if_types[$var_name] = [['=isset'], ['-' . $inferior_value_comparison]];
             }
 
             return $if_types ? [$if_types] : [];
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -3730,7 +3730,7 @@ class AssertionFinder
             }
 
             if ($var_name) {
-                $if_types[$var_name] = [['=isset'], ['+' . $superior_value_comparison]];
+                $if_types[$var_name] = [['=isset'], ['>' . $superior_value_comparison]];
             }
 
             return $if_types ? [$if_types] : [];
@@ -3937,7 +3937,7 @@ class AssertionFinder
             }
 
             if ($var_name) {
-                $if_types[$var_name] = [['=isset'], ['-' . $inferior_value_comparison]];
+                $if_types[$var_name] = [['=isset'], ['<' . $inferior_value_comparison]];
             }
 
             return $if_types ? [$if_types] : [];
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -93,7 +93,7 @@ class AssertionFinder
 
             if ($var_name) {
                 if ($candidate_if_types) {
-                    $if_types[$var_name] = [['>' . \json_encode($candidate_if_types[0])]];
+                    $if_types[$var_name] = [['&' . \json_encode($candidate_if_types[0])]];
                 } else {
                     $if_types[$var_name] = [['!falsy']];
                 }
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -1583,6 +1583,42 @@ class AssertionFinder
      * @param PhpParser\Node\Expr\BinaryOp\Greater|PhpParser\Node\Expr\BinaryOp\GreaterOrEqual $conditional
      * @return false|int
      */
+    protected static function hasSuperiorNumberCheck(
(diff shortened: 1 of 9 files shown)


src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
```diff
@@ -410,6 +410,48 @@ public static function isContainedBy(
             return false;
         }
 
+        if ($input_type_part instanceof TInt && $container_type_part instanceof TIntRange) {
+            if ($input_type_part instanceof TPositiveInt) {
+                if ($container_type_part->min_bound > 0) {
+                    //any positive int can't be pushed inside a range with a min > 0
+                    if ($atomic_comparison_result) {
```
weirdan, line 417: Shouldn't this be `> 1`? I mean, `positive-int` is `[1, ∞)`, isn't it?

src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
```diff
@@ -410,6 +410,48 @@ public static function isContainedBy(
...
+            if ($input_type_part instanceof TLiteralInt) {
+                $min_bound = $container_type_part->min_bound;
+                $max_bound = $container_type_part->max_bound;
+
+                return
+                    ($min_bound === null || $min_bound <= $input_type_part->value) &&
+                    ($max_bound === null || $max_bound >= $input_type_part->value);
+            }
+
+            //any int can't be pushed inside a range without coercion
```
weirdan, line 446: Unless the range is `int<min,max>`, or is it handled elsewhere?

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -3676,6 +3714,28 @@ private static function getGreaterAssertions(
...
+            } else {
+                $var_name = ExpressionIdentifier::getArrayVarId(
+                    $conditional->right,
+                    $this_class_name,
+                    $source
+                );
+            }
+
+            if ($var_name) {
+                $if_types[$var_name] = [['=isset'], ['+' . $superior_value_comparison]];
```
weirdan, line 3733: Is there any reason why you chose `+`? I think this will eventually bleed out to user-defined assertions and `@psalm-assert +5 $param` is not very intuitive notion for a number that is at most 5.

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -1546,6 +1562,50 @@ private static function reconcileHasArrayKey(
...
+    ) : Union {
+        foreach ($existing_var_type->getAtomicTypes() as $atomic_type) {
+            if ($atomic_type instanceof Atomic\TIntRange) {
+                $existing_var_type->removeType($atomic_type->getKey());
+                if ($atomic_type->max_bound === null) {
+                    $atomic_type->max_bound = (int)$assertion;
+                } else {
+                    $atomic_type->max_bound = min($atomic_type->max_bound, (int)$assertion);
+                }
+                $atomic_type->max_bound = min($atomic_type->max_bound, (int)$assertion);
```
weirdan, line 1599: This looks like it's already assigned in the `if..else..` above.

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
orklah: Damn, I didn't thought about user assertions. I didn't pick ">" because it's used here: https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php#L96 and https://github.com/vimeo/psalm/blob/master/src/Psalm/Type/Reconciler.php#L165. I thought about using `>>` instead but assertions can be negated so ">" in second position is checked too somewhere.

We could replace ">" as it is already used by something else but I'm not sure to understand what's going on with that operator

weirdan: > We could replace ">" as it is already used by something else but I'm not sure to understand what's going on with that operator

Me neither.

src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
```diff
@@ -443,10 +443,12 @@ public static function isContainedBy(
                     ($max_bound === null || $max_bound >= $input_type_part->value);
             }
 
-            //any int can't be pushed inside a range without coercion
-            if ($atomic_comparison_result) {
-                $atomic_comparison_result->type_coerced = true;
-                $atomic_comparison_result->type_coerced_from_scalar = true;
+            //any int can't be pushed inside a range without coercion (unless the range is from min to max)
+            if ($container_type_part->min_bound !== null && $container_type_part->max_bound !== null) {
```
weirdan, line 447: Shouldn't that be OR instead of AND? 'If the range has any non-max/non-min bounds specified TInt won't fit'

orklah: wow, I'm starting to make big mistakes. I have my head full of the stuff, I'll get back to it later. Right now I have build failures because `int<min, 1> + 1` is translated to `int`

weirdan: It does seem that those failures were resolved by this change though.

orklah: You're right. Must be something to do with coercion being allowed in array access or something like that... It's good, it means I'll be able to put arithmetics in a separated chunk of changes

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
orklah: @muglug any insight here as to what's done with the ">" operator in assertions? The goal would be to change it to somewhere less significant so we can use ">" for superior

orklah: I fixed it by replacing it with "@".

weirdan (approved):

weirdan:
Thanks!

PR #6219, opened 2021-08-01
Added missing mb_str_split function

### LINKED ISSUE #6217: mb_str_split signature missing
PHP 7.4 added the [mb_str_split](https://www.php.net/manual/en/function.mb-str-split.php) function to the multibyte string library.

However, it appears to be missing form the CallMap_74_delta.php dictionary, as well as from the main CallMap.php

I believe the signature should be: 

`'mb_str_split' => ['list<string>', 'string'=>'string', 'limit='=>'int', 'encoding='=>'string|null'],`

bitwise-operators:
Added missing mb_str_split() function from PHP 7.4 to dictionaries

Fixes #6217

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -7289,6 +7289,7 @@ return [
 'mb_scrub' => ['string', 'string'=>'string', 'encoding='=>'string'],
 'mb_send_mail' => ['bool', 'to'=>'string', 'subject'=>'string', 'message'=>'string', 'additional_headers='=>'string|array', 'additional_params='=>'string'],
 'mb_split' => ['list<string>', 'pattern'=>'string', 'string'=>'string', 'limit='=>'int'],
+'mb_str_split' => ['list<string>', 'string'=>'string', 'limit='=>'int', 'encoding='=>'string|null'],
 'mb_strcut' => ['string', 'string'=>'string', 'start'=>'int', 'length='=>'?int', 'encoding='=>'string'],
 'mb_strimwidth' => ['string', 'string'=>'string', 'start'=>'int', 'width'=>'int', 'trim_marker='=>'string', 'encoding='=>'string'],
 'mb_stripos' => ['int|false', 'haystack'=>'string', 'needle'=>'string', 'offset='=>'int', 'encoding='=>'string'],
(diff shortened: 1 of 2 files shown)


orklah:
Awesome!

Is there a reason why you called the second parameter "limit"? the doc seem to say "length" https://www.php.net/manual/en/function.mb-str-split.php

dictionaries/CallMap.php
```diff
@@ -7289,6 +7289,7 @@
 'mb_scrub' => ['string', 'string'=>'string', 'encoding='=>'string'],
 'mb_send_mail' => ['bool', 'to'=>'string', 'subject'=>'string', 'message'=>'string', 'additional_headers='=>'string|array', 'additional_params='=>'string'],
 'mb_split' => ['list<string>', 'pattern'=>'string', 'string'=>'string', 'limit='=>'int'],
+'mb_str_split' => ['list<string>', 'string'=>'string', 'limit='=>'int', 'encoding='=>'string|null'],
```
weirdan, line 7292:
```suggestion
'mb_str_split' => ['list<string>', 'string'=>'string', 'length='=>'int', 'encoding='=>'string|null'],
```

dictionaries/CallMap_74_delta.php
```diff
@@ -16,6 +16,7 @@
 return [
 'new' => [
     'gzread' => ['string|false', 'stream'=>'resource', 'length'=>'int'],
+    'mb_str_split' => ['list<string>', 'string'=>'string', 'limit='=>'int', 'encoding='=>'string|null'],
```
weirdan, line 19:
```suggestion
    'mb_str_split' => ['list<string>', 'string'=>'string', 'length='=>'int', 'encoding='=>'string|null'],
```

weirdan (changes requested):

weirdan (approved):

bitwise-operators:
> Is there a reason why you called the second parameter "limit"? the doc seem to say "length" https://www.php.net/manual/en/function.mb-str-split.php

Oops, well spotted. I copy-pasted part of the signature from mb_split() and missed the difference in name.

orklah:
It didn't used to matter but now with PHP 8's named parameter, it does :). Thanks for the PR!

weirdan:
Thanks!

VincentLanglet:
I think it can be improved https://github.com/vimeo/psalm/pull/6324

PR #6229, opened 2021-08-02
Prevent network issues due to xml.xsd

### LINKED ISSUE #6228: `allow_url_fopen=1` now required to run Psalm?
Upgrading from `4.8.1` to `4.9.2` introduced the following error when running `psalm` with `allow_url_fopen=0`:
```sh
▶  psalm                                                                                                  
Uncaught RuntimeException: PHP Error: DOMDocument::schemaValidate(): http:// wrapper is disabled in the server configuration by allow_url_fopen=0 in /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Config.php:707 in /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Internal/ErrorHandler.php:66
Stack trace:                                                                                                                                                                                                                                                                                                   
#0 [internal function]: Psalm\Internal\ErrorHandler::Psalm\Internal\{closure}(2, 'DOMDocument::sc...', '/mnt/ebs/home/d...', 707)
#1 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Config.php(707): DOMDocument->schemaValidate('/mnt/ebs/home/d...')
#2 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Config.php(652): Psalm\Config::validateXmlConfig('/mnt/ebs/home/d...', '<?xml version="...')                                                               
#3 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Config.php(625): Psalm\Config::loadFromXML('/mnt/ebs/home/d...', '<?xml version="...', '/mnt/ebs/home/d...', '/mnt/ebs/home/d...')                                                                                                                    
#4 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Config.php(577): Psalm\Config::loadFromXMLFile('/mnt/ebs/home/d...', '/mnt/ebs/home/d...')                                                                                                                                                            
#5 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Internal/CliUtils.php(490): Psalm\Config::getConfigForPath('/mnt/ebs/home/d...', '/mnt/ebs/home/d...')
#6 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(552): Psalm\Internal\CliUtils::initializeConfig(NULL, '/mnt/ebs/home/d...', 'console', Object(Composer\Autoload\ClassLoader), false)
#7 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(1007): Psalm\Internal\Cli\Psalm::loadConfig(NULL, '/mnt/ebs/home/d...', 'console', Object(Composer\Autoload\ClassLoader), false, Array)                                                                                        
#8 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(216): Psalm\Internal\Cli\Psalm::initConfig('/mnt/ebs/home/d...', Array, 'vendor', Object(Composer\Autoload\ClassLoader), NULL, 'console', false, Array)                                                                        
#9 /mnt/ebs/home/drans/Code/vendor/vimeo/psalm/psalm(4): Psalm\Internal\Cli\Psalm::run(Array)             
#10 {main}                                                                                                                                                                                                          
(Psalm 4.9.2@00c062267d6e3229d91a1939992987e2d46f2393 crashed due to an uncaught Throwable)               
```

Looks like this happens as part of validating the XML. Was that an intentional change?

This is what our config looks like for reference:
```xml
// psalm.xml
<?xml version="1.0"?>
<psalm
    totallyTyped="false"
    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
    errorBaseline="psalm-baseline.xml"
    autoloader="Tests/psalm-load.php"
    resolveFromConfigFile="true"
    hideExternalErrors="true"
    findUnusedVariablesAndParams="true"
>
```

gmessier:
Possible fix #6228


config.xsd
```diff
@@ -5,7 +5,7 @@
     targetNamespace="https://getpsalm.org/schema/config"
     elementFormDefault="qualified"
 >
-    <xs:import namespace="http://www.w3.org/XML/1998/namespace" schemaLocation="http://www.w3.org/2001/03/xml.xsd" />
+    <xs:import namespace="http://www.w3.org/XML/1998/namespace" schemaLocation="xml.xsd" />
```
weirdan, line 8: It would have to be added to PHAR (see `box.json.dist`).

gmessier:
Ok... I though the changes was so small that I didn't have to understand the pipeline and run tests locally... but seems I will need to dig into it.

weirdan:
I suspect file paths in xsd may be resolved relative to cwd, which would be quite a pain to fix for PHAR.

gmessier:
For reference: https://stackoverflow.com/questions/22774425/attempting-to-connect-xml-files-with-xinclude-attribute-xmlbase-error
I still try to find another way to be still be able to include file in another directory, but still not need to load the XML schema from the internet (so be able to load it from the PHAR that seems impossible for now).

gmessier:
The only solution I could find is to allow any attribute to be set in the configuration. XInclude will be able to add the xml:base attribute without breaking the schema validation. It should not be a big problem to allow useless attributes, but it's not to me to take this decision. 
The only other solution I can see is to remove this feature and not be able to include file from other directory. IMO if you can only include files from the same directory it's not so useful. It only "clean" a little bit your configuration by having many files with small chunks of the configuration (by example one for the directories, one for the issuehandlers, etc.)

weirdan:
Thanks!

PR #6231, opened 2021-08-02
Fixes: PHP 8.0: Argument #3 ($offset) must be contained in argument #1 ($haystack)

tm1000:
Fixes crash when using PHP 8.0

See: https://bugs.php.net/bug.php?id=80336&edit=3

(PHP Bug above is not valid but it gives a good look into the issue in Psalm itself)


weirdan (approved):

weirdan:
Thanks!

Issue #6220, opened 2021-08-01
`in_array` assertion with dynamic haystack doesn't reconcile needle type to array's value type


supersmile2009:
Example:
```
function inArrayFromRange(?int $val): int {
    if (!in_array($val, range(0, 4), true)) {
        throw new \Exception();
    }
    
    return $val;
}
```

Errors:
```
ERROR: NullableReturnStatement - 29:12 - The declared return type 'int' for inArrayFromRange is not nullable, but the function returns 'int|null'

ERROR: InvalidNullableReturnType - 24:39 - The declared return type 'int' for inArrayFromRange is not nullable, but 'int|null' contains null
```

I expect `$val` to be `int` type after the assertion.
In all the cases I tested with haystack being `non-empty-list<T>` psalm fails to reconcile the needle type to `T` after the assertion. It works fine with fixed arrays though (e. g. `[0, 1]`)

More examples:
https://psalm.dev/r/53fed31716

orklah:
If you want to take a look, it's handled in AssertionFinder::getInarrayAssertions

PR #6233, opened 2021-08-03
Improve `in_array` assertion to extract assertions for non-literal types

supersmile2009:
Fixes https://github.com/vimeo/psalm/issues/6220

Thanks @orklah for the tip on where to check.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -3429,43 +3429,52 @@ class AssertionFinder
                     || $atomic_type instanceof Type\Atomic\TKeyedArray
                     || $atomic_type instanceof Type\Atomic\TList
                 ) {
+                    $is_sealed = false;
                     if ($atomic_type instanceof Type\Atomic\TList) {
                         $value_type = $atomic_type->type_param;
                     } elseif ($atomic_type instanceof Type\Atomic\TKeyedArray) {
                         $value_type = $atomic_type->getGenericValueType();
+                        $is_sealed = $atomic_type->sealed;
                     } else {
                         $value_type = $atomic_type->type_params[1];
                     }
 
-                    $array_literal_types = \array_filter(
-                        $value_type->getAtomicTypes(),
-                        function ($type) {
-                            return $type instanceof Type\Atomic\TLiteralInt
-                                || $type instanceof Type\Atomic\TLiteralString
-                                || $type instanceof Type\Atomic\TLiteralFloat
-                                || $type instanceof Type\Atomic\TEnumCase;
+                    $assertions = [];
+
+                    foreach ($value_type->getAtomicTypes() as $atomic_value_type) {
+                        $assertion = '';
+                        // If it's not a sealed (fixed, known) array, we can't simply return value types as assertions.
+                        // E. g. in_array($x, ['a', 'b']) is the same as $x === 'a' || $x === 'b',
+                        // which can also be negated correctly.
+                        // However, in_array($x, $y), where y is list<'a'|'b'> doesn't work the same way.
+                        // With positive assertion it has similar meaning: $x is 'a'|'b'.
+                        // But without knowing exact contents of $y, it's not an equality assertion
+                        // threfore it cannot be safely negated.
+                        // If we simply return =string(a) and =string(b) assertions, they will have the same semantics
+                        // as in the first example. So when negated, we will end up with $x !== 'a' && $x !== 'b'.
+                        // That won't work for unknown (not sealed) haystack.
+                        // 'in-array-' prefix is added to distinguish such assertions.
+                        if (!$is_sealed) {
+                            $assertion .= 'in-array-';
                         }
-                    );
-
-                    if ($array_literal_types
-                        && count($value_type->getAtomicTypes())
-                    ) {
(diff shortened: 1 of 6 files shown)


supersmile2009:
The last commit fixes 2 old tests, which used to pass before because `AssertionFinder::getInarrayAssertions()` wasn't treating lists of literals correctly. Since `list<"a"|"b">` doesn't guarantee that either of those values is actually present in the array, the assertions extracted by `getInarrayAssertions()` were not safe. That's what the large comment block added to AssertionFinder::getInarrayAssertions() is about.

By looking at those tests I believe that original intention was to check against a known and fixed list of constants, which is `array{"a", "b"}`. In fact those tests would fail even without this patch if `restrict_return_types` was enabled.

So I wouldn't call it a regression, it's a bug fix rather.

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -3429,43 +3429,52 @@ private static function getInarrayAssertions(
...
+                        // However, in_array($x, $y), where y is list<'a'|'b'> doesn't work the same way.
+                        // With positive assertion it has similar meaning: $x is 'a'|'b'.
+                        // But without knowing exact contents of $y, it's not an equality assertion
+                        // threfore it cannot be safely negated.
+                        // If we simply return =string(a) and =string(b) assertions, they will have the same semantics
+                        // as in the first example. So when negated, we will end up with $x !== 'a' && $x !== 'b'.
+                        // That won't work for unknown (not sealed) haystack.
+                        // 'in-array-' prefix is added to distinguish such assertions.
+                        if (!$is_sealed) {
+                            $assertion .= 'in-array-';
```
orklah, line 3458: Is it possible to fall in `!is-sealed` and in the first block of conditions below? `in-array-=...` would fail I think

If it's not possible, could we rearrange conditions to make that more clear?

orklah:
Indeed, `list<"a"|"b">` doesn't guarantee the presence of one or the other, and the fact that `non-empty-list<"a"|"b">` wasn't used, it didn't even guaranteed the array contained anything at all.

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -1482,7 +1483,7 @@ private static function reconcileInArray(
         Codebase $codebase,
         Union $existing_var_type,
         string $assertion
-    ) : Union {
+    ) : ?Union {
```
orklah, line 1486: In what context does Type::intersectUnionTypes is expected to return null and what happens when we return null here?

orklah:
Seems awesome! Thanks :)

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
supersmile2009: These are the cases where haystack contains type which isn't present in the needle, like:
```
/**
 * @param int|bool $x
 * @param non-empty-list<int|string> $y
 * @return int
 */
function assertInArray($x, $y) {
    if (in_array($x, $y, true)) {
        return $x;
    }

    throw new Exception();
}
```
I missed the fact that some sort of Union instance will have to be returned eventually. I added some tests, which indeed revealed the problem. Fixed by returning TEmpty if intersection is null. Return type was reverted to `Union` (not nullable)

orklah: Great!

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
supersmile2009: Well, technically such case is possible. Haystack of type `list<'a'|'b'|int>` and alike will cause that.
That's why there is a try-catch around `Type::parseString()` in `SimpleAssertionReconciler::reconcileInArray()`. The idea was to leave it as an option for future improvement if anyone ever wants to type-check such weird arrays.

I didn't originally intend to pursue it, but I ended up finding a bunch of cases, which didn't produce satisfactory results, like some unions of scalar types (e. g. `int|bool` needle with `array<int|string>` haystack etc.) triggering somewhat misleading/incorrect errors. So I went down this rabbit hole and in the end there's no more `in-array-=...` case.

I'll push the changes and post more details shortly.

supersmile2009:
1. The change to `Psalm\Type::intersectUnionTypes()`.
It turned out that it couldn't intersect some scalar type unions properly. `(int|string)&(int|bool)` returned null (like no intersection at all, not `null` type) and not `int` as expected. Now it's fixed.
2. `in-array-` assertion now uses slightly different syntax. Type key is simply concatenated with the prefix. The first approach that I took with separate assertion for each type couldn't work. You need to have all the types in a single context to do the intersection. Otherwise if you keep returning Empty type after separate assertion reconciliation, you get wrong errors in some cases. With the new implementation Reconcilers can intersect the types and produce correct errors, like if condition is redundant or if assertion makes no sense, because types are completely different.

orklah:
@supersmile2009 seems like something broke in Psalm. Can you check it?

supersmile2009:
@orklah done. Could you trigger the CI again?

weirdan:
Thanks!

orklah:
This triggered a new false positive in my work codebase: https://psalm.dev/r/aaa65f492c

PR #6238, opened 2021-08-03
don't flag UnusedForeachValue as fixable

### LINKED ISSUE #6234: Doc-block suppress for UnusedForeachValue
https://psalm.dev/r/ce2467c803

```php
<?php

/** @var array<string, int> */
$data = [];

/** @psalm-suppress UnusedForeachValue */
foreach ($data as $k => $v) {
    echo $k;
}
```

I would expect that error on unused foreach value is suppressed, but getting the following:

```
Psalm output (using commit 9d68e2e): 

INFO: UnusedForeachValue - 7:25 - $v is never referenced or the value is not used
 
```

also it's marked as fixable, but fixing is not working as well.

orklah:
This fixes #6234


weirdan:
Thanks!

PR #6239, opened 2021-08-03
Drop Appveyor reporter as we no longer use Appveyor

weirdan:


orklah:
Judging from the tests results, we can conclude that AppVeyor is the single most important piece of Psalm 😄

PR #6240, opened 2021-08-04
refactor arithmetics

orklah:
This PR propose extracting literal arithmetics operations from the big analyzeNonDivOperands


weirdan:
Thanks!

PR #6241, opened 2021-08-04
1> Range arithmetics and assertions

### LINKED ISSUE #4972: Psalm does not detect that values are acceptable to int options
```php
final class C
{
    /**
     * @psalm-var 1|2|3|4|5|6|7|8
     */
    private int $x;

    /**
     * @psalm-var 1|2|3|4|5|6|7|8
     */
    private int $y;

    public static function fromString(string $s): self
    {
        if (strlen($s) !== 2) {
            throw new InvalidArgumentException;
        }

        $x = ord($s[0]) - 96;
        $y = (int) $s[1];

        if ($x < 1 || $x > 8 || $y < 1 || $y > 8) {
            throw new InvalidArgumentException;
        }

        return new self($x, $y);
    }

    /**
     * @psalm-param 1|2|3|4|5|6|7|8 $x
     * @psalm-param 1|2|3|4|5|6|7|8 $y
     */
    private function __construct(int $x, int $y)
    {
        $this->x = $x;
        $this->y = $y;
    }
}
```
```
Psalm output (using commit e299e91): 

ERROR: ArgumentTypeCoercion - 27:25 - Argument 1 of C::__construct expects int(1)|int(2)|int(3)|int(4)|int(5)|int(6)|int(7)|int(8), parent type int provided

ERROR: ArgumentTypeCoercion - 27:29 - Argument 2 of C::__construct expects int(1)|int(2)|int(3)|int(4)|int(5)|int(6)|int(7)|int(8), parent type int provided
```
https://psalm.dev/r/19d5c8f2b9

`if ($file < 1 || $file > 8 || $rank < 1 || $rank > 8) {` results in an exception when the expression evaluates to `false`. Therefore, both `$x` and `$y` fulfill the `int(1)|int(2)|int(3)|int(4)|int(5)|int(6)|int(7)|int(8)` "type".

### LINKED ISSUE #5635: False positive LessSpecificReturnStatement when subtracting from a positive-int
It is difficult to work with positive-int values if there will be any math involved, even basic decrement. I'm not sure if this can be practically "fixed", but it would be nice.

https://psalm.dev/r/1ed8a99bc6

### LINKED ISSUE #6488: Crash with modulo by zero
https://psalm.dev/r/19b89e001a

orklah:
This PR is the continuation of https://github.com/vimeo/psalm/pull/6210. It should fix some issues introduced in it due to assertions.

However, I underestimated this task. There are a few points I don't know how to fix and I'll probably need help for that.

If we need to make a new release, I think #6210 should be reverted to prevent creating false positives in builds. I'll reintegrate in this PR later

EDIT: 
this fixes #5635
this fixes #4972
this fixes #6488

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -1559,18 +1559,12 @@ class AssertionFinder
      * @param PhpParser\Node\Expr\BinaryOp\Greater|PhpParser\Node\Expr\BinaryOp\GreaterOrEqual $conditional
      * @return false|int
      */
-    protected static function hasPositiveNumberCheck(
+    protected static function hasSuperiorNumberCheck(
         PhpParser\Node\Expr\BinaryOp $conditional,
-        ?int &$min_count
+        ?int &$literal_value_comparison
     ) {
-        if ($conditional->right instanceof PhpParser\Node\Scalar\LNumber
-            && $conditional->right->value >= (
-                $conditional instanceof PhpParser\Node\Expr\BinaryOp\Greater
-                ? 0
-                : 1
-            )
-        ) {
-            $min_count = $conditional->right->value +
+        if ($conditional->right instanceof PhpParser\Node\Scalar\LNumber) {
+            $literal_value_comparison = $conditional->right->value +
                 ($conditional instanceof PhpParser\Node\Expr\BinaryOp\Greater ? 1 : 0);
 
             return self::ASSIGNMENT_TO_RIGHT;
@@ -1580,22 +1574,16 @@ class AssertionFinder
     }
 
     /**
-     * @param PhpParser\Node\Expr\BinaryOp\Greater|PhpParser\Node\Expr\BinaryOp\GreaterOrEqual $conditional
+     * @param PhpParser\Node\Expr\BinaryOp\Smaller|PhpParser\Node\Expr\BinaryOp\SmallerOrEqual $conditional
      * @return false|int
      */
-    protected static function hasZeroCheck(
+    protected static function hasInferiorNumberCheck(
         PhpParser\Node\Expr\BinaryOp $conditional,
-        ?int &$zero_count
+        ?int &$literal_value_comparison
     ) {
-        if ($conditional->right instanceof PhpParser\Node\Scalar\LNumber
-            && $conditional->right->value >= (
-            $conditional instanceof PhpParser\Node\Expr\BinaryOp\Greater
-                ? -1
-                : 0
-            )
-        ) {
-            $zero_count = $conditional->right->value +
-                ($conditional instanceof PhpParser\Node\Expr\BinaryOp\Greater ? 1 : 0);
+        if ($conditional->right instanceof PhpParser\Node\Scalar\LNumber) {
+            $literal_value_comparison = $conditional->right->value +
+                ($conditional instanceof PhpParser\Node\Expr\BinaryOp\Smaller ? -1 : 0);
 
             return self::ASSIGNMENT_TO_RIGHT;
         }
@@ -3603,7 +3591,7 @@ class AssertionFinder
      */
     private static function getGreaterAssertions(
         PhpParser\Node\Expr $conditional,
-        ?Codebase $codebase,
+        ?Codebase $_codebase,
         FileSource $source,
         ?string $this_class_name
     ): array {
@@ -3611,12 +3599,10 @@ class AssertionFinder
 
         $min_count = null;
         $count_equality_position = self::hasNonEmptyCountEqualityCheck($conditional, $min_count);
-        $min_comparison = null;
-        $positive_number_position = self::hasPositiveNumberCheck($conditional, $min_comparison);
-        $zero_comparison = null;
-        $zero_position = self::hasZeroCheck($conditional, $zero_comparison);
         $max_count = null;
         $count_inequality_position = self::hasLessThanCountEqualityCheck($conditional, $max_count);
+        $superior_value_comparison = null;
+        $superior_value_position = self::hasSuperiorNumberCheck($conditional, $superior_value_comparison);
 
         if ($count_equality_position) {
             if ($count_equality_position === self::ASSIGNMENT_TO_RIGHT) {
@@ -3672,114 +3658,29 @@ class AssertionFinder
             return $if_types ? [$if_types] : [];
         }
 
-        if ($positive_number_position) {
-            if ($positive_number_position === self::ASSIGNMENT_TO_RIGHT) {
+        if ($superior_value_position) {
+            if ($superior_value_position === self::ASSIGNMENT_TO_RIGHT) {
                 $var_name = ExpressionIdentifier::getArrayVarId(
                     $conditional->left,
                     $this_class_name,
                     $source
                 );
-                $value_node = $conditional->left;
             } else {
                 $var_name = ExpressionIdentifier::getArrayVarId(
                     $conditional->right,
                     $this_class_name,
                     $source
                 );
-                $value_node = $conditional->right;
             }
 
-            if ($codebase
-                && $source instanceof StatementsAnalyzer
-                && ($var_type = $source->node_data->getType($value_node))
-                && $var_type->isSingle()
-                && $var_type->hasBool()
-                && $min_comparison > 1
-            ) {
-                if ($var_type->from_docblock) {
-                    if (IssueBuffer::accepts(
-                        new DocblockTypeContradiction(
-                            $var_type . ' cannot be greater than ' . $min_comparison,
-                            new CodeLocation($source, $conditional),
-                            null
-                        ),
-                        $source->getSuppressedIssues()
-                    )) {
-                        // fall through
-                    }
+            if ($var_name !== null) {
+                if ($superior_value_comparison === 0) {
+                    $if_types[$var_name] = [['positive-numeric', '=int(0)']];
+                } if ($superior_value_comparison === 1) {
+                    $if_types[$var_name] = [['positive-numeric']];
                 } else {
-                    if (IssueBuffer::accepts(
-                        new TypeDoesNotContainType(
-                            $var_type . ' cannot be greater than ' . $min_comparison,
-                            new CodeLocation($source, $conditional),
-                            null
-                        ),
-                        $source->getSuppressedIssues()
-                    )) {
-                        // fall through
-                    }
+                    $if_types[$var_name] = [['=isset'], ['>' . $superior_value_comparison]];
                 }
-            }
-
-            if ($var_name) {
-                $if_types[$var_name] = [[($min_comparison === 1 ? '' : '=') . 'positive-numeric']];
-            }
-
-            return $if_types ? [$if_types] : [];
-        }
-
-        if ($zero_position) {
-            if ($zero_position === self::ASSIGNMENT_TO_RIGHT) {
-                $var_name = ExpressionIdentifier::getArrayVarId(
-                    $conditional->left,
-                    $this_class_name,
-                    $source
-                );
-                $value_node = $conditional->left;
-            } else {
-                $var_name = ExpressionIdentifier::getArrayVarId(
-                    $conditional->right,
-                    $this_class_name,
-                    $source
-                );
-                $value_node = $conditional->right;
-            }
-
-
-            if ($codebase
-                && $source instanceof StatementsAnalyzer
-                && ($var_type = $source->node_data->getType($value_node))
-                && $var_type->isSingle()
-                && $var_type->hasBool()
-                && $zero_comparison > 1
-            ) {
-                if ($var_type->from_docblock) {
-                    if (IssueBuffer::accepts(
-                        new DocblockTypeContradiction(
-                            $var_type . ' cannot be greater than ' . $zero_comparison,
-                            new CodeLocation($source, $conditional),
-                            null
-                        ),
-                        $source->getSuppressedIssues()
-                    )) {
-                        // fall through
-                    }
-                } else {
-                    if (IssueBuffer::accepts(
-                        new TypeDoesNotContainType(
-                            $var_type . ' cannot be greater than ' . $zero_comparison,
-                            new CodeLocation($source, $conditional),
-                            null
-                        ),
-                        $source->getSuppressedIssues()
-                    )) {
-                        // fall through
-                    }
-                }
-            }
-
-            if ($var_name) {
-                $if_types[$var_name] = [[($zero_comparison === 1 ? '' : '=') . 'positive-numeric', '=int(0)']];
             }
 
             return $if_types ? [$if_types] : [];
@@ -3800,10 +3701,11 @@ class AssertionFinder
         $if_types = [];
         $min_count = null;
         $count_equality_position = self::hasNonEmptyCountEqualityCheck($conditional, $min_count);
-        $typed_value_position = self::hasTypedValueComparison($conditional, $source);
-
         $max_count = null;
         $count_inequality_position = self::hasLessThanCountEqualityCheck($conditional, $max_count);
+        //$typed_value_position = self::hasTypedValueComparison($conditional, $source);
+        $inferior_value_comparison = null;
+        $inferior_value_position = self::hasInferiorNumberCheck($conditional, $inferior_value_comparison);
 
         if ($count_equality_position) {
             if ($count_equality_position === self::ASSIGNMENT_TO_LEFT) {
@@ -3855,7 +3757,7 @@ class AssertionFinder
             return $if_types ? [$if_types] : [];
         }
 
-        if ($typed_value_position) {
+        /*if ($typed_value_position) {
             if ($typed_value_position === self::ASSIGNMENT_TO_RIGHT) {
                 $var_name = ExpressionIdentifier::getArrayVarId(
                     $conditional->left,
@@ -3886,6 +3788,28 @@ class AssertionFinder
                 && ($expr_type->getSingleIntLiteral()->value === 0)
             ) {
                 $if_types[$var_name] = [['=isset']];
+            }
+
+            return $if_types ? [$if_types] : [];
+        }*/
+
+        if ($inferior_value_position) {
+            if ($inferior_value_position === self::ASSIGNMENT_TO_RIGHT) {
+                $var_name = ExpressionIdentifier::getArrayVarId(
+                    $conditional->left,
+                    $this_class_name,
+                    $source
+                );
+            } else {
+                $var_name = ExpressionIdentifier::getArrayVarId(
+                    $conditional->right,
+                    $this_class_name,
+                    $source
+                );
+            }
+
+            if ($var_name) {
+                $if_types[$var_name] = [['=isset'], ['<' . $inferior_value_comparison]];
             }
 
             return $if_types ? [$if_types] : [];
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -1570,6 +1570,13 @@ class AssertionFinder
             return self::ASSIGNMENT_TO_RIGHT;
         }
 
+        if ($conditional->left instanceof PhpParser\Node\Scalar\LNumber) {
+            $literal_value_comparison = $conditional->left->value +
+                ($conditional instanceof PhpParser\Node\Expr\BinaryOp\Greater ? 1 : 0);
+
+            return self::ASSIGNMENT_TO_LEFT;
+        }
+
         return false;
     }
 
@@ -1586,6 +1593,13 @@ class AssertionFinder
                 ($conditional instanceof PhpParser\Node\Expr\BinaryOp\Smaller ? -1 : 0);
 
             return self::ASSIGNMENT_TO_RIGHT;
+        }
+
+        if ($conditional->left instanceof PhpParser\Node\Scalar\LNumber) {
+            $literal_value_comparison = $conditional->left->value +
+                ($conditional instanceof PhpParser\Node\Expr\BinaryOp\Smaller ? -1 : 0);
+
+            return self::ASSIGNMENT_TO_LEFT;
         }
 
         return false;
@@ -3674,12 +3688,16 @@ class AssertionFinder
             }
 
             if ($var_name !== null) {
-                if ($superior_value_comparison === 0) {
-                    $if_types[$var_name] = [['positive-numeric', '=int(0)']];
-                } if ($superior_value_comparison === 1) {
-                    $if_types[$var_name] = [['positive-numeric']];
+                if ($superior_value_position === self::ASSIGNMENT_TO_RIGHT) {
+                    if ($superior_value_comparison === 0) {
+                        $if_types[$var_name] = [['=isset'], ['positive-numeric', '=int(0)']];
+                    } elseif ($superior_value_comparison === 1) {
+                        $if_types[$var_name] = [['=isset'], ['positive-numeric']];
+                    } else {
+                        $if_types[$var_name] = [['=isset'], ['>' . $superior_value_comparison]];
+                    }
                 } else {
-                    $if_types[$var_name] = [['=isset'], ['>' . $superior_value_comparison]];
+                    $if_types[$var_name] = [['=isset'], ['<' . $superior_value_comparison]];
                 }
             }
 
@@ -3703,7 +3721,6 @@ class AssertionFinder
         $count_equality_position = self::hasNonEmptyCountEqualityCheck($conditional, $min_count);
         $max_count = null;
         $count_inequality_position = self::hasLessThanCountEqualityCheck($conditional, $max_count);
-        //$typed_value_position = self::hasTypedValueComparison($conditional, $source);
         $inferior_value_comparison = null;
         $inferior_value_position = self::hasInferiorNumberCheck($conditional, $inferior_value_comparison);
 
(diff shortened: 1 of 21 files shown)


weirdan:
I'm planning to cut a new release this weekend. Do you reckon I'll need to revert #6210 ?

orklah:
Yes, please do. I tried to work on this since then and I have obstacles everywhere. Better do it now...

orklah:
I put aside loop features for now, they're pretty complex and they're not required for now.

I also put aside the TPositiveInt deprecation, this PR should have less impact now because I kept TPositiveInt first for now when applicable.

This PR should be ok, except for a test here: https://github.com/vimeo/psalm/blob/3ad3375777b431a559b624d359c9af667de319f1/tests/TypeReconciliation/ConditionalTest.php#L2737

I don't understand why null should be preserved through the conditional given it's always false: https://3v4l.org/K2Ed1

The next step will be to remove TPositiveInt from assertions while still keeping everythink working.

orklah:
Note: please merge https://github.com/vimeo/psalm/pull/6306 before this.

weirdan:
> I don't understand why null should be preserved through the conditional given it's always false

That doesn't make sense for me either, not with `0` anyway.

src/Psalm/Type/Atomic/TIntRange.php
```diff
@@ -70,4 +70,12 @@ public function isPositive(): bool
     {
         return $this->min_bound !== null && $this->min_bound > 0;
     }
+
+    public function contains(int $i): bool
+    {
+        return
+            ($this->min_bound === null && $this->max_bound === null) ||
+            ($this->min_bound === null && $this->max_bound >= $i) ||
+            ($this->max_bound === null && $this->min_bound <= $i);
```
weirdan, line 79: Shouldn't there be another case for when both min and max are specified?

orklah: Thanks!

orklah:
@weirdan can we maybe do a new release and merge this after?

weirdan:
@orklah sure, will do this weekend.

orklah:
the CI is currently failing on phpunit.

this is due to a particular construction like this:
```
function scope(int $a) {
    const A = array('a', 'b');

    $result = A[$i % count(A)];
}
```

Before this PR, Psalm was infering `$i % count(A)` to `int`. Now that the inference has improved, it's infered to `int<-2, 2>`, because we know the `count(A)` will return 2.

However, in Psalm `A[$b]` is tolerated when $b is a positive-int or an int, but it's not when $b is an integer range (this is deliberate as it's one of the way integer range can find new bugs).

The issue in phpunit can be fixed by adding phpdoc to $i that it's a positive-int, this way Psalm will infer the modulo result as `int<0, 2>` and it will be compatible with expected array-keys `0|1|2`.

I think we can merge this now

orklah:
I tried to document as much as I could directly in the code, but I'll make a summary here of what the reason for each modified file is for review and future reference purposes:
- `AssertionFinder.php`: I introduced two big new assertions: `>X` and `<X` to infer ranges. They replaced incomplete previous assertions like `hasPositiveNumberCheck` and `hasZeroCheck`.
- `NonDivArithmeticOpAnalyzer.php`: I allowed Psalm to make literal arithmetic operations between two integer range and between integer range and other int (this is what improved the modulo operation I was talking about above)
- `IntegerRangeComparator.php`: this is an object dedicated to check if an Integer range can be contained by other types. Either in another IntegerRange or even in a bunch of other integer types combined. This is quite a new notion in Psalm to use multiple types to satisfy one
- `ScalarTypeComparator.php`: this checks whether another integer type can be contained by an integer range
- `UnionTypeComparator.php`: this one is the step before the two above files. One piece of code is designed to call `IntegerRangeComparator` to handle the multiple ints that can contain an integer range. The other is a special case needed because until now, Psalm considered two types as "contained by" when their atomics could be identical, but this is not exactly true anymore with integer range (see comment for more info)
- `SimpleAssertionReconciler.php`: it is the file that process what we found in AssertionFinder. It takes the known type for a variable and will try to apply new assertions. I commented some piece of code related to loops because it's a very complex topic we'll have to come back to after this one is merged. Also there are some places where we could emit issues that I intend to add in another PR
- `TypeCombiner.php`: this is the file that is used to make multiple types into one. It remove redundancies and expand existing types instead of just putting them altogether
- `TIntRange.php`: I added some useful functions to recalculate bounds or to check if an integer is contained in the range
- `Union.php`: hasInt is now true if there is range in the union, isAlwaysTruthy can be true for an IntegerRange(I decommented a piece of code I introduced in a previous PR) and I added a useful function to fetch all ranges

orklah:
I'm putting this in draft to resolve #6241 in this PR

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -3672,116 +3751,39 @@ private static function getGreaterAssertions(
...
-                            new CodeLocation($source, $conditional),
-                            null
-                        ),
-                        $source->getSuppressedIssues()
-                    )) {
-                        // fall through
+            if ($var_name !== null) {
+                if ($superior_value_position === self::ASSIGNMENT_TO_RIGHT) {
+                    if ($superior_value_comparison === 0) {
+                        $if_types[$var_name] = [['=positive-numeric', '=int(0)']];
```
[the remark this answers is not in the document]
orklah: first level are ANDed and second level are ORed. That's why positive-numeric and int(0) are on the same group but isset below is on it's own group

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
```diff
@@ -665,6 +665,125 @@ private static function analyzeNonDivOperands(
                 return null;
             }
 
+            if ($left_type_part instanceof Type\Atomic\TIntRange && $right_type_part instanceof Type\Atomic\TIntRange) {
+                if ($parent instanceof PhpParser\Node\Expr\BinaryOp\Div) {
```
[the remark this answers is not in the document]
orklah: It confused me a lot too but there's actually nothing preventing Div to pass through and we indeed get Div here

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
```diff
@@ -665,6 +665,125 @@ private static function analyzeNonDivOperands(
...
+                        $left_type_part->max_bound,
+                        $right_type_part->max_bound,
+                        false
+                    );
+                }
+
+                $min_value = $calculated_min_type !== null ? $calculated_min_type->getSingleIntLiteral()->value : null;
+                $max_value = $calculated_max_type !== null ? $calculated_max_type->getSingleIntLiteral()->value : null;
+
+                $new_result_type = new Type\Union([new Type\Atomic\TIntRange($min_value, $max_value)]);
```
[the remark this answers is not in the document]
orklah: Technically, arithmetics on TIntRange will check for float results and assume min or max when that happens. there's no philosophical difference between TIntRange and TInt or TPositiveInt

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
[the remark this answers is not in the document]
orklah: Actually, there's even an explicit Div call: https://github.com/vimeo/psalm/blob/be73f74b7e5d285b4379538f1a010c34ecc9b33f/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonComparisonOpAnalyzer.php#L118

I'll rename the class for less confusion

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
```diff
@@ -665,6 +665,125 @@ private static function analyzeNonDivOperands(
...
+                    );
+                }
+
+                $calculated_max_type = null;
+                if ($left_type_part->max_bound !== null && $right_type_part->max_bound !== null) {
+                    // when there are two valid numbers, make any operation
+                    $calculated_max_type = self::arithmeticOperation(
+                        $parent,
+                        $left_type_part->max_bound,
+                        $right_type_part->max_bound,
```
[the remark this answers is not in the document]
orklah: I fixed Minus and bit operations

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -1559,47 +1561,106 @@ protected static function hasNotCountEqualityCheck(
...
-            && $conditional->right->value >= (
-                $conditional instanceof PhpParser\Node\Expr\BinaryOp\Greater
-                ? 0
-                : 1
-            )
-        ) {
-            $min_count = $conditional->right->value +
+        $right_assignment = false;
+        $value_right = null;
+        if ($conditional->right instanceof LNumber) {
```
weirdan, line 1571: Note: if we have already analyzed `$right`/`$left` by the time we get here, it's probably better to use the already inferred type, so that it'll work with more complex expressions (e.g. `if ($a >= Limits::MAX + 1)`).

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +893,134 @@ public static function arithmeticOperation(
...
+        if ($parent instanceof PhpParser\Node\Expr\BinaryOp\Div) {
+            //can't assume an int range will stay int after division
+            if (!$result_type) {
+                $result_type = new Type\Union([new Type\Atomic\TInt(), new Type\Atomic\TFloat()]);
+            } else {
+                $result_type = Type::combineUnionTypes(
+                    new Type\Union([new Type\Atomic\TInt(), new Type\Atomic\TFloat()]),
+                    $result_type
+                );
+            }
```
weirdan, line 912: Potential future refactoring: allowing `null` for `Type::combineUnionTypes()` arguments would make these `if () else`s easier to read:
```php
            if (!$result_type) {
                $result_type = new Type\Union([new Type\Atomic\TInt(), new Type\Atomic\TFloat()]);
            } else {
                $result_type = Type::combineUnionTypes(
                    new Type\Union([new Type\Atomic\TInt(), new Type\Atomic\TFloat()]),
                    $result_type
                );
            }

// c.f.

            $result_type = Type::combineUnionTypes(
                new Type\Union([new Type\Atomic\TInt(), new Type\Atomic\TFloat()]),
                $result_type
            );         
```

Alternatively we could have a new method with that semantic.

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +893,134 @@ public static function arithmeticOperation(
...
+            return;
+        }
+
+        if ($parent instanceof PhpParser\Node\Expr\BinaryOp\BitwiseAnd ||
+            $parent instanceof PhpParser\Node\Expr\BinaryOp\BitwiseOr ||
+            $parent instanceof PhpParser\Node\Expr\BinaryOp\BitwiseXor
+        ) {
+            //really complex to calculate
```
weirdan, line 920: Intuitively, for positive ranges it's constrained by `int<0, max(a_max, b_max)>`, but this can be improved later.

weirdan (changes requested):

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +893,134 @@ public static function arithmeticOperation(
...
+            //and the max is the max from first range minus the min from the second
+            $max_operand1 = $left_type_part->max_bound;
+            $max_operand2 = $right_type_part->min_bound;
+        } else {
+            $min_operand1 = $left_type_part->min_bound;
+            $min_operand2 = $right_type_part->min_bound;
+
+            $max_operand1 = $left_type_part->max_bound;
+            $max_operand2 = $right_type_part->max_bound;
+        }
```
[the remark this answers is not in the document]
weirdan: Wolframalpha could be used for quick checks: https://www.wolframalpha.com/input/?i=interval%5B-2%2C2%5D+*+interval%5B-3%2C-2%5D

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
orklah: or we may coalesce with an empty Union and check if it resolve as intended...

Maybe create a new issue on that, I'd prefer avoiding adding this to this PR, it's already big enough

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
orklah: No, it's more complex than that unfortunately: https://3v4l.org/5bgJ9

At least for Or and Xor, the max can be higher than the max of both ranges. Worse, if you have `int<64,64>` and `int<62,64>` the max will be `63|64` so you have to go looking in between the bounds...

But yeah, I figured anyone who could benefit for a better inference here will have better knowledge than me in binary operations! Maybe there's a simple solution here I don't see.

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +893,134 @@ public static function arithmeticOperation(
...
+            }
+            return;
+        }
+
+        if ($parent instanceof PhpParser\Node\Expr\BinaryOp\ShiftLeft ||
+            $parent instanceof PhpParser\Node\Expr\BinaryOp\ShiftRight
+        ) {
+            //really complex to calculate
+            if (!$result_type) {
+                $result_type = new Type\Union([new Type\Atomic\TIntRange(0, null)]);
```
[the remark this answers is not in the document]
orklah: Oh, I was confused by the fact PHP will forbid negative int in the second operand. it's fixed

orklah:
@weirdan could you update our fork of phpunit? It should fix the CI issue

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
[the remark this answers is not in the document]
orklah: oh god, that looks like... Mathematics!

Indeed, my solution is wrong, I'll work on that

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
weirdan: > At least for Or and Xor, the max can be higher than the max of both ranges

True. The result will, however, be of the same (binary) magnitude because none of those operations carry over (that's why they called bitwise). Basically we can throw off most significant `0` bits, fill the remaining with `1`s and call that max. That's for positive ints. Negatives are likely a different story, because they are stored as binary complements.

weirdan:
> could you update our fork of phpunit? It should fix the CI issue

Done.

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
orklah: Indeed, I'll check that. I wonder why it's not what's done in other cases...

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
orklah: Not sure with bits operation usage, but from what I've seen this far is that range operations can lead to false positive quickly (for example, using the range to hit an offset for a known array. If you don't end up with an exact answer, you risk something like `$array[$my_range]` where $my_range is off by a few integers but it's enough to create a false positive.

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
orklah: I did. However, the $source is not always a StatementsAnalyzer (on which we can retrieve type). That means I had to leave the rest of the conditionals for when it's something else, though I didn't check in what cases the type is different.

it may be the reason why Psalm didn't try with already inferred types before, though it seems to work fine so it may be extended to other assertions

orklah: Note that if needed, we could go further. I only checked for Literal Ints but we should be able to use Positive Ints, or other ranges
for example if($range > $other_range) could be asserted to change $range to always be higher than $other_range (and the other way around too)

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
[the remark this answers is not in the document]
orklah: That's tough. The equation doesn't seem to include infinites so it's hard to compute as is.
I'm starting to wonder if an API call to Wolfram wouldn't be the best solution 😄 

I'll work on it more, at least I built every possible case in the test suite, that's progress!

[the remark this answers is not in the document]
orklah: Done :)

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +895,454 @@ public static function arithmeticOperation(
...
+            //this can't happen
+            return;
+        }
+
+        $new_range = TIntRange::convertToIntRange($other_operand);
+        if ($left_is_range) {
+            self::analyzeOperandsBetweenIntRange($parent, $result_type, $range_operand, $new_range);
+        } else {
+            self::analyzeOperandsBetweenIntRange($parent, $result_type, $new_range, $range_operand);
+        }
```
weirdan, line 1041:
```suggestion
        if (!$left_type_part instanceof Type\Atomic\TIntRange) {
            $left_type_part = TIntRange::convertToIntRange($left_type_part);
        }
        if (!$right_type_part instanceof Type\Atomic\TIntRange) {
            $right_type_part = TIntRange::convertToIntRange($right_type_part);
        }
        return self::analyzeOperandsBetweenIntRange($parent, $result_type, $left_type_part, $right_type_part);
```

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +895,454 @@ public static function arithmeticOperation(
...
+        } else {
+            $result_type = Type::combineUnionTypes($new_result_type, $result_type);
+        }
+    }
+
+    /**
+     * @param TIntRange|TInt $left_type_part
+     * @param TIntRange|TInt $right_type_part
+     */
+    private static function analyzeOperandsBetweenIntRangeAndInt(
```
weirdan, line 1017: it probably 'analyze *operation*' rather than 'analyze *operands*'

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +895,454 @@ public static function arithmeticOperation(
...
+        } else {
+            self::analyzeOperandsBetweenIntRange($parent, $result_type, $new_range, $range_operand);
+        }
+    }
+
+    private static function analyzeMulBetweenIntRange(
+        TIntRange $right_type_part,
+        TIntRange $left_type_part,
+        PhpParser\Node\Expr\BinaryOp\Mul $parent,
+        ?Type\Union &$result_type
```
weirdan, line 1048: Could we keep the order of parameters the same across these methods? E.g. `analyzeOperandsBetweenIntRangeAndInt` has it in this order: `$parent, $result_type, $left, $right`

weirdan (changes requested):

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
orklah: Wow, I wrote so many line for so little :D Thanks!

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
orklah: Hmm, I named that function from the caller method that was named `analyzeNonDivOperands` before this PR and now named `analyzeOperands` but Operation seems more fitting indeed. Should I rename everything from Operands to Operation?

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
orklah: Of course, thanks for noticing

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
weirdan: Let's postpone it to some time later.

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +895,454 @@ public static function arithmeticOperation(
...
+    private static function analyzeModBetweenIntRange(
+        TIntRange $right_type_part,
+        TIntRange $left_type_part,
+        ?Type\Union &$result_type
+    ): void {
+        //result of Mod is not directly dependant on the bounds of the range
+        if ($right_type_part->min_bound !== null && $right_type_part->min_bound === $right_type_part->max_bound) {
+            //if the second operand is a literal, we can be pretty detailed
+            if ($right_type_part->min_bound === 0) {
+                $new_result_type = Type::getEmpty();
```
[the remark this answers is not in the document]
orklah: I'm writing this down as a future improvement but this is something we don't do currently, even for Div by 0. I added an Empty case for Mod though, so this PR fixes #6488

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +895,454 @@ public static function arithmeticOperation(
...
+            } else {
+                if ($left_type_part->isPositiveOrZero()) {
+                    if ($right_type_part->isPositive()) {
+                        $max = $right_type_part->min_bound - 1;
```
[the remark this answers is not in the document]
orklah: done!

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +895,454 @@ public static function arithmeticOperation(
...
+                    }
+                    $new_result_type = new Type\Union([new TIntRange(-$max, $max)]);
+                }
+            }
+        } elseif ($right_type_part->isPositive()) {
+            if ($left_type_part->isPositiveOrZero()) {
+                $new_result_type = new Type\Union([new TIntRange(0, null)]);
+            } elseif ($left_type_part->isNegativeOrZero()) {
+                $new_result_type = new Type\Union([new TIntRange(null, 0)]);
+            } else {
```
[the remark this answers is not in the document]
orklah: This seemed pretty complex, so I didn't try. If we fall into this else, this means we have something like
`int<57, 95> % int<min, 18>`. I guess I could put a PositiveInt, because a negative in modulo seem to have the same effective as if it was positive so it would result in a positive integer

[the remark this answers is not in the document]
weirdan: I meant the whole `if..elseif..else` block.

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -867,4 +895,454 @@ public static function arithmeticOperation(
...
+            $result_type = Type::combineUnionTypes($new_result_type, $result_type);
+        }
+    }
+
+    private static function analyzePowBetweenIntRange(
+        TIntRange $left_type_part,
+        TIntRange $right_type_part,
+        ?Type\Union &$result_type
+    ): void {
+        //If Pow first operand is negative, the result will be negative, else it will be positive
```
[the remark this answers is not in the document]
orklah: I totally did. Thanks, this is fixed

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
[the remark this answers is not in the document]
orklah: I guess I could make a case for degenerate right_type_part here
Now that I think about it, if we have a PositiveOrZero range in left, we can pretty much transform it into a int<0, left_max-1>

weirdan (approved):
LGTM. @orklah, feel free to merge.

weirdan:
Congrats on your first merge! 🎆

orklah:
Thanks! 🥳

PR #6244, opened 2021-08-04
Permit null timestamp to date() in PHP 8

### LINKED ISSUE #6243: PHP 8 Bug: Argument 2 of `date()` cannot be null, possibly null value provided
Psalm complains when `null` is passed as the 2nd argument (`$timestamp`) of [`date()`](https://www.php.net/manual/en/function.date.php). As of PHP 8, `$timestamp` is a nullable int whereas it was just an int in prior versions.

## Example

https://psalm.dev/r/52eb795065

## Proposed Solution

Add a callmap for `date()` in https://github.com/vimeo/psalm/blob/master/dictionaries/CallMap_80_delta.php

masonmcelvain:
## What's new
PHP 8 allows the timestamp to be null, but version 7.4 doesn't.
Creates a mapping for date() to allow $timestamp to be null in PHP 8.

Closes #6243

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -1706,7 +1706,7 @@ return [
 'cyrus_connect' => ['resource', 'host='=>'string', 'port='=>'string', 'flags='=>'int'],
 'cyrus_query' => ['array', 'connection'=>'resource', 'query'=>'string'],
 'cyrus_unbind' => ['bool', 'connection'=>'resource', 'trigger_name'=>'string'],
-'date' => ['string|false', 'format'=>'string', 'timestamp='=>'int'],
+'date' => ['string', 'format'=>'string', 'timestamp='=>'?int'],
 'date_add' => ['DateTime|false', 'object'=>'DateTime', 'interval'=>'DateInterval'],
 'date_create' => ['DateTime|false', 'datetime='=>'string', 'timezone='=>'?DateTimeZone'],
 'date_create_from_format' => ['DateTime|false', 'format'=>'string', 'datetime'=>'string', 'timezone='=>'?\DateTimeZone'],
--- a/dictionaries/CallMap_80_delta.php
+++ b/dictionaries/CallMap_80_delta.php
@@ -348,6 +348,7 @@ return [
 'curl_share_init' => ['resource'],
 'curl_share_setopt' => ['bool', 'sh'=>'resource', 'option'=>'int', 'value'=>'mixed'],
 'curl_unescape' => ['string|false', 'ch'=>'resource', 'string'=>'string'],
+'date' => ['string', 'format'=>'string', 'timestamp='=>'int'],
 'date_add' => ['DateTime|false', 'object'=>'DateTime', 'interval'=>'DateInterval'],
 'date_date_set' => ['DateTime|false', 'object'=>'DateTime', 'year'=>'int', 'month'=>'int', 'day'=>'int'],
 'date_diff' => ['DateInterval|false', 'obj1'=>'DateTimeInterface', 'obj2'=>'DateTimeInterface', 'absolute='=>'bool'],
--- a/dictionaries/CallMap_80_delta.php
+++ b/dictionaries/CallMap_80_delta.php
@@ -55,6 +55,7 @@ return [
 'DateTimeImmutable::format' => ['string', 'format'=>'string'],
 'DateTimeImmutable::createFromInterface' => ['self', 'object' => 'DateTimeInterface'],
 'DateTimeZone::listIdentifiers' => ['list<string>', 'timezoneGroup='=>'int', 'countryCode='=>'string|null'],
+'date' => ['string', 'format'=>'string', 'timestamp='=>'?int'],
 'date_add' => ['DateTime', 'object'=>'DateTime', 'interval'=>'DateInterval'],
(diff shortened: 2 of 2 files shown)


dictionaries/CallMap_80_delta.php
```diff
@@ -55,6 +55,7 @@
 'DateTimeImmutable::format' => ['string', 'format'=>'string'],
 'DateTimeImmutable::createFromInterface' => ['self', 'object' => 'DateTimeInterface'],
 'DateTimeZone::listIdentifiers' => ['list<string>', 'timezoneGroup='=>'int', 'countryCode='=>'string|null'],
+'date' => ['string', 'format'=>'string', 'timestamp='=>'?int'],
```
weirdan, line 58: In addition to this, this entry should be in `CallMap.php`. Also, the old signature needs to be present in 'old' section.

masonmcelvain: Added the old signature to this file: https://github.com/vimeo/psalm/pull/6244/commits/81661395ce952b9a66c595db2ada8019aeab7709
Added the new signature to CallMap: https://github.com/vimeo/psalm/pull/6244/commits/b2ba774cb9b08efa16e2d1ccc870911c8939408d

dictionaries/CallMap.php
```diff
@@ -1706,7 +1706,7 @@
 'cyrus_connect' => ['resource', 'host='=>'string', 'port='=>'string', 'flags='=>'int'],
 'cyrus_query' => ['array', 'connection'=>'resource', 'query'=>'string'],
 'cyrus_unbind' => ['bool', 'connection'=>'resource', 'trigger_name'=>'string'],
-'date' => ['string|false', 'format'=>'string', 'timestamp='=>'int'],
+'date' => ['string', 'format'=>'string', 'timestamp='=>'?int'],
```
orklah, line 1709: Callmap.php should contain the signature for the oldest version supported so you should keep |false and not allow nullable. Also the definition in callmap.php must be strictly identical tp tthe 'old' section of the next Callmap (so 8.0 in your case)

bitwise-operators: I think you have this reversed.

Psalm/Internal/Codebase/InternalCallMapHandler loads the current CallMap, and then walks backward from 8.0 through the delta files, removing 'new' entries, and adding 'old' entries until it reaches the version it's analyzing for.

So the main CallMap should have the most recent signature present in the delta files (in this case the 8.0 signature)

masonmcelvain: [date_format](https://www.php.net/manual/en/datetime.format.php) got mapped for php 8, so maybe we can use it as an example.
<details>
<summary>date_format example</summary>

Changelog
image

new section (php 8.0):
https://github.com/vimeo/psalm/blob/019b35850ac31b3919143ea8e202881e863c925f/dictionaries/CallMap_80_delta.php#L61
old section (php 7.4, I think):
https://github.com/vimeo/psalm/blob/019b35850ac31b3919143ea8e202881e863c925f/dictionaries/CallMap_80_delta.php#L353

And the CallMap.php file seems to have the entry from the old section (NOT the php 8 signature).
```php
'date_format' => ['string|false', 'object'=>'DateTimeInterface', 'format'=>'string'],
```
</details>
This is only one example, but it suggests @orklah is right.

What it doesn't clarify is if we should keep including `|false` in the CallMap.php and the old section.

The docblock for 80_CallMap_delta.php says
```
This contains the information needed to convert the function signatures for php 8.0 to php 7.4 
```
so it seems we would want to put the php 7.4 signature in the old section (ie don't include the `|false`).

The confusion: `|false` seems like the oldest supported version so it should be in CallMap.php. But CallMap.php and the old section of 80_CallMap_delta.php need to match exactly.

> Callmap.php should contain the signature for the oldest version supported so you should keep |false and not allow nullable. Also the definition in callmap.php must be strictly identical tp tthe 'old' section of the next Callmap (so 8.0 in your case)

So, do we include `|false`?

orklah: I may be confused here, I never get this right :D Honestly, we could use some unit tests for the callmap consistency

bitwise-operators: > I may be confused here, I never get this right :D Honestly, we could use some unit tests for the callmap consistency

[Way ahead of you mate](https://github.com/vimeo/psalm/pull/6247)! :-D

Like I said, I've had a look at the way Psalm reads the CallMap and Delta files. It loads the main callback and then, if the version needed is older than 8.0, starts from the newest delta file and works its way back. This is the logical approach, as it means that stopping support for an old version can be done by simply removing the old delta files. It's also fastest for the newest PHP version, which I imagine is also a plus.

But it looks like a lot of the changes in the 8.0 delta were not applied to the main CallMap, which is why I wrote that test in the first place. I fixed a lot of the errors, but I'm quite sure a lot of the changes to falsey returns are still missing or incorrect.

Unfortunately, there's no comprehensive list of the functions that received that change with the 8.0 release, and the 'false' return type is often absent in the PHP doc even for functions that (still) have it, so the only real way to check it, is to test it.

orklah: I was confused for a moment by how quick you made these changes after my suggestion after realizing that you posted the PR 3 minutes after my comment and that didn't even left you enough time to write the PR message :D

weirdan (approved):

weirdan:
Thanks!

PR #6251, opened 2021-08-06
Use different psalm cache for different error levels

### LINKED ISSUE #4952: Psalm should reset the cache when changing the error level.
If I run 
```
psalm --error-level=8
```
And get no errors

And then I run
```
psalm --error-level=7
```
I still get no errors

But if I remove the cache, and run again
```
psalm --error-level=7
```
Now I get correctly the psalm errors.

TysonAndre:
Fixes #4952

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -634,6 +634,14 @@ class Config
     }
 
     /**
+     * Computes the hash to use for a cache folder from CLI flags and from the config file's xml contents
+     */
+    public function computeHash(): string
+    {
+        return sha1($this->hash . ':' . $this->level);
+    }
+
+    /**
      * Creates a new config object from an XML string
      * @param  string|null      $current_dir Current working directory, if different to $base_dir
      *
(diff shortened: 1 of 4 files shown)


src/Psalm/Config.php
```diff
@@ -623,7 +623,7 @@ public static function loadFromXMLFile(string $file_path, string $current_dir):
 
         try {
             $config = self::loadFromXML($base_dir, $file_contents, $current_dir, $file_path);
-            $config->hash = sha1($file_contents . \PSALM_VERSION);
+            $config->hash = sha1($file_contents);
```
orklah, line 626: Is there a reason for this? It seems sane to invalidate hash when Psalm is updated

TysonAndre: That was a bug - I initially tried to add cli overrides on this line but realized they weren't available

src/Psalm/Config.php
```diff
@@ -633,6 +633,14 @@ public static function loadFromXMLFile(string $file_path, string $current_dir):
         return $config;
     }
 
+    /**
+     * Computes the hash to use for a cache folder from CLI flags and from the config file's xml contents
+     */
+    public function computeHash(): string
+    {
+        return sha1($this->hash . ':' . $this->level);
```
orklah, line 641: I saw a wrong reuse of the cache between a partial analysis of a single file and a full run where errors due to lack of context on the first partial run was reported on the full run. Shouldn't we use a hash that is composer.lock + psalm.xml + command line to be safe?

weirdan:
Thanks!

Issue #5040, opened 2021-01-18
false positive with non-empty-string and sprintf


ppdeblieck:
When using sprintf to generate a string that string can't be fed into a function where the argument expects a non-empty-string.
A "Argument 1 of func expects non-empty-string, parent type string provided" is thrown.
See: https://psalm.dev/r/55b6f41128

orklah:
There are some edge cases where this is not true when the template only contains flags or empty chars:
https://3v4l.org/3REnd
- Either because params are all empty-strings
- Or because the non-empty-strings params don't match any flag in the template

Getting the right type 100% would require rewriting the string parser and match every flag with every value.

This would be tedious but this would potentially allow emitting Psalm issue for non-matching types for example in`sprintf('%d', '')`

ppdeblieck:
This means the non-empty-string can't be used at all in methods that can get their arguments by sprintf because it'll always fail. The only solution is setting a more generic 'string' type.

If you want to force non-empty Exception messages like in
`throw new CustomException(sprintf('Unrecognized option "%s"', $name));`
(in Symfony "Exception and error message strings must be concatenated using sprintf"; see https://symfony.com/doc/current/contributing/code/standards.html)

How do we annotate this? Or isn't this possible at all?

orklah:
if you put the sprintf result in a variable, you can just add `/** @var non-empty-string $var */` above, Or even `Assert::notEmpty($var)` or similar if you want to be sure

boesing:
Did I miss something? 

Actually, `non-empty-string` is not equivalent with `!empty`. 
Its just that there is at least one character in there.
https://psalm.dev/r/7f4a837969

I've experienced this as well and thus missing this feature. PHPStan somehow parses the 1st argument to the `sprintf` and knows how many "placeholder" are used:
https://phpstan.org/r/9e4475c2-5a37-4908-a558-a0e785e4ebf2

This might be achievable with some sort of `regex` probably.

I've gathered some LoC which I'd expect to be safely treated as `non-empty-string` based on the informations psalm already has:
https://psalm.dev/r/0e712cceb7

When executing, this works as expected: https://3v4l.org/K288q

boesing:
I'd also say, this could be a psalm error because of the usage of `sprintf` without a placeholder (if we are able to verify that there are placeholders in there).

```php
// Same as above, just without the placeholder
nonEmptyString(sprintf(':'));
```

weirdan:
[...]

PR #6252, opened 2021-08-06
Literal float methods to `Union` type

boesing:
Hey there, 

I recently started to write a plugin to handle #5040
I tried to receive the value of a `float` type but then realized that the union type lacks methods for literal float.
In that plugin, i definitely could use `Union#isSingleFloatLiteral` along with `Union#getSingleFloatLiteral` and `Union#hasLiteralFloat`.

This, I'd like to add these methods with this PR. If anything is missing, feel free to hit me up.

### DIFF
--- a/src/Psalm/Type/Union.php
+++ b/src/Psalm/Type/Union.php
@@ -1489,4 +1489,31 @@ class Union implements TypeNode
     {
         return $this->types;
     }
+
+    /**
+     * @return bool true if this is a float literal with only one possible value
+     */
+    public function isSingleFloatLiteral(): bool
+    {
+        return count($this->types) === 1 && count($this->literal_float_types) === 1;
+    }
+
+    /**
+     * @throws \InvalidArgumentException if isSingleFloatLiteral is false
+     *
+     * @return TLiteralFloat the only float literal represented by this union type
+     */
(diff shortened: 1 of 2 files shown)


src/Psalm/Type/Union.php
```diff
@@ -1489,4 +1489,31 @@ public function getChildNodes() : array
...
+
+    /**
+     * @throws \InvalidArgumentException if isSingleFloatLiteral is false
+     *
+     * @return TLiteralFloat the only float literal represented by this union type
+     */
+    public function getSingleFloatLiteral(): TLiteralFloat
+    {
+        if (count($this->types) !== 1 || count($this->literal_float_types) !== 1) {
+            throw new \InvalidArgumentException('Not an int literal');
```
orklah, line 1509: int -> float

src/Psalm/Type/Union.php
```diff
@@ -1489,4 +1489,31 @@ public function getChildNodes() : array
     {
         return $this->types;
     }
+
+    /**
+     * @return bool true if this is a int literal with only one possible value
```
orklah, line 1494: same here :)

boesing:
Not sure why Unit Tests - 6 was cancelled x_O

weirdan:
Thanks!

PR #6253, opened 2021-08-06
fix #5501

### LINKED ISSUE #5501: @psalm-suppress NonInvariantDocblockPropertyType is reported as unused
When running this https://psalm.dev/r/7b36c93015
with `--find-unused-psalm-suppress`

I get 
```
UnusedPsalmSuppress - This suppression is never used (see https://psalm.dev/207)
     * @psalm-suppress NonInvariantDocblockPropertyType
```

Seems like psalm.dev is not running with find-unused-psalm-suppress so I cannot fully reproduce the bug.

orklah:
The issue here was that Psalm registered the suppression with an offset (cursor position) relative to the beggining of the docblock whereas the expected offset was absolute (from the beggining of the file).

This lead me to send the starting position of the docblock to the parse function to make the addition, and then I had to go back to every place that used this offset to substract the docblock position.

I hope I didn't miss one, this was not easy to do.

I'm not sure if it will fix other similar issues or not, and frankly, I'm not even sure to understand why some suppression work and not other (I think this has to do with which type of docblock and where it need to be parsed)

this fixes #5501

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
@@ -506,8 +506,7 @@ class ClassLikeDocblockParser
 
                     $line_parts[1] = preg_replace('/,$/', '', $line_parts[1]);
 
-                    $start = $offset;
-                    $end = $start + strlen($line_parts[0]);
+                    $end = $offset + strlen($line_parts[0]);
 
                     $line_parts[0] = str_replace("\n", '', preg_replace('@^[ \t]*\*@m', '', $line_parts[0]));
 
@@ -534,7 +533,7 @@ class ClassLikeDocblockParser
                             $offset - $comment->getStartFilePos()
                         ),
                         'tag' => $property_tag,
-                        'start' => $start,
+                        'start' => $offset,
                         'end' => $end,
                     ];
                 }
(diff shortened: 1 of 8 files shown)


orklah:
this fixes #5501

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
```diff
@@ -255,21 +266,24 @@ public static function parse(
...
                 }
             }
         }
 
-        $imported_types = array_merge(
-            $parsed_docblock->tags['phpstan-import-type'] ?? [],
-            $parsed_docblock->tags['psalm-import-type'] ?? []
-        );
+        $imported_types = ($parsed_docblock->tags['phpstan-import-type'] ?? []) +
+            ($parsed_docblock->tags['psalm-import-type'] ?? []);
```
weirdan, line 275: Good catch, I believe `array_merge()` was destroying offsets here.

src/Psalm/Internal/Scanner/DocblockParser.php
```diff
@@ -19,7 +19,7 @@
 
 class DocblockParser
 {
-    public static function parse(string $docblock) : ParsedDocblock
+    public static function parse(string $docblock, int $offsetStart) : ParsedDocblock
```
weirdan, line 22: Mind adding a docblock specifying whether absolute file offset or relative docblock offset is expected here?

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
orklah: yeah it was, that was a pain to pinpoint :)

src/Psalm/Internal/Scanner/DocblockParser.php
orklah: Will do!

orklah: done!

weirdan:
Thanks!

PR #6268, opened 2021-08-09
Refined debug_backtrace return type

vudaltsov:
Some explanations:
- `class` and `type` are absent in a simple function call,
- `object` is present only with `DEBUG_BACKTRACE_PROVIDE_OBJECT` option,
- `args` is absent when `DEBUG_BACKTRACE_IGNORE_ARGS` option is used.

Also removed `bool` from `options` type, since it's not there according to [docs](https://www.php.net/manual/en/function.debug-backtrace).


orklah:
There was a time where php.net said this function could take a boolean where it's now flags:
https://web.archive.org/web/20090215184624/http://us3.php.net/debug_backtrace

Seems like it's stopped somewhere between 2009 and 2011 but some devs could be used to that. Conditional on it having the same behaviour as today, I'd vote to keep it possible

vudaltsov:
@orklah , agree, thank you for the information!

weirdan:
Before 5.4 the first argument used to be called `$provide_object`, that's where `bool` comes from. It can't be kept there though, as PHP would throw at runtime with `strict_types` enabled (https://3v4l.org/2WXuL, https://3v4l.org/vORSW), thus Psalm needs to flag it. 

@vudaltsov please revert the revert.

vudaltsov:
@weirdan , thank you! Done.

weirdan:
@vudaltsov you will need to additionally edit either `CallMap_historical` (if the function had this signature before PHP 7.1) or one of the deltas (if there were changes during 7.1-8.0 period). That's why unit tests currently fail here.

weirdan:
Thanks!

Issue #6261, opened 2021-08-08
+ operator over array<int, int> is giving me mixed value


VincentLanglet:
I simplified my code to https://psalm.dev/r/1c3274bd0a
In my real case, I'm adding another array.

I don't understand why the value is considered as possibly mixed.

weirdan:
Funny thing that it works for (any) two arrays, but not for three.

PR #6272, opened 2021-08-09
seal the result of two sealed KeyedArrays

orklah:
This should fix https://github.com/vimeo/psalm/issues/6261

The result of '+' for 2 sealed TKeyedArrays wasn't sealed itself so any additional '+' treated missing keys as "possibly mixed"

I also fixed a missing `<?php` in a test. Hope it won't fail now!


weirdan:
Thanks!

PR #6273, opened 2021-08-09
Constants are non empty strings

### LINKED ISSUE #6266: non empty string constants should be considered as a non empty string
With the option
```
ignoreInternalFunctionFalseReturn="false"
```

In
```
/** @var non-empty-string $foo */
$foo = PHP_EOL;
$bar = explode($foo, $diff);
```
$bar is considered as always an array.

In
```
$bar = explode(PHP_EOL, $diff);
```
$bar is considered as array|false.

So I would say that explode is correctly supporting non-empty-string, but some constant are not considered as non-empty-string. I tried
- PHP_EOL
- DIRECTORY_SEPARATOR
- a custom class constant `public const FOO = 'foo';`

Should these be `non-empty-string` ?
https://psalm.dev/r/377464f1cd

VincentLanglet:
Close #6266

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ConstFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ConstFetchAnalyzer.php
@@ -137,6 +137,9 @@ class ConstFetchAnalyzer
                 case 'PHP_VERSION':
                 case 'DIRECTORY_SEPARATOR':
                 case 'PATH_SEPARATOR':
+                case 'PHP_EOL':
+                    return Type::getNonEmptyString();
+
                 case 'PEAR_EXTENSION_DIR':
                 case 'PEAR_INSTALL_DIR':
                 case 'PHP_BINARY':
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ConstFetchAnalyzer.php
```diff
@@ -155,7 +155,7 @@ public static function getGlobalConstType(
                 case 'PHP_PREFIX':
                 case 'PHP_SAPI':
                 case 'PHP_SYSCONFDIR':
-                    return Type::getString();
```
weirdan, line 158: Not all of those are necessarily non-empty: https://3v4l.org/oSTv2

weirdan: I would suggest to limit the change to those constants you had in your original issue.

VincentLanglet: I found 3 constants possibly empty and fixed them.

VincentLanglet:
@weirdan I followed your advice and only return nonEmptyString for the most useful
PHP_VERSION, DIRECTORY_SEPARATOR, PATH_SEPARATOR and PHP_EOL.

weirdan (approved):

weirdan:
Thanks!

Issue #6263, opened 2021-08-08
Splat in constant array expression produces wrong array shape


weirdan:
https://psalm.dev/r/fb451afec9

Expected: flat list (https://3v4l.org/0IdDo)

orklah:
I took a quick glance at it. The issue is in here: https://github.com/vimeo/psalm/blob/c2ba1fec25f131ea04af7befa93a81ff95bf1797/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php#L1167

We loop over the class constants and we use SimpleTypeInferer::infer to calculate the type. Unfortunately, this method doesn't handle unpacking at all. We should probably take a look at how it's handled in ArrayAnalyzer::handleUnpackedArray and reuse code somehow

weirdan:
@ondrejmirtes PHPStan also suffers from this bug: https://phpstan.org/r/f5a17feb-139f-433d-8503-6d081598e694

orklah:
@weirdan FYI you can dump types in PHPStan, similar to @psalm-trace: https://phpstan.org/r/4b4198b8-b52a-4c72-b185-42b3b5c8dc94

ondrejmirtes:
Please open a bug report in PHPStan's repo, thanks :)

PR #6279, opened 2021-08-10
Refactor: constant array type inference

weirdan:
This brings constant array type inference closer to non-constant analysis, and is a step toward fixing #6263

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
@@ -437,96 +437,19 @@ class SimpleTypeInferer
                 continue;
             }
 
-            $single_item_key_type = null;
-
-            if ($item->key) {
-                $single_item_key_type = self::infer(
-                    $codebase,
-                    $nodes,
-                    $item->key,
-                    $aliases,
-                    $file_source,
-                    $existing_class_constants,
-                    $fq_classlike_name
-                );
-
-                if ($single_item_key_type) {
-                    $array_creation_info->item_key_atomic_types = array_merge(
-                        $array_creation_info->item_key_atomic_types,
-                        array_values($single_item_key_type->getAtomicTypes())
-                    );
-                }
-            } else {
-                $array_creation_info->item_key_atomic_types[] =  new Type\Atomic\TInt();
-            }
-
-            $single_item_value_type = self::infer(
+            if (!self::handleArrayItem(
                 $codebase,
                 $nodes,
-                $item->value,
+                $array_creation_info,
+                $int_offset,
+                $item,
                 $aliases,
                 $file_source,
                 $existing_class_constants,
                 $fq_classlike_name
-            );
-
-            if (!$single_item_value_type) {
+            )) {
                 return null;
             }
-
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -547,4 +413,200 @@ public static function infer(
...
+            $item_value_type = TypeCombiner::combine(
+                $array_creation_info->item_value_atomic_types,
+                null,
+                false,
+                true,
+                30
+            );
+        }
+
+        // if this array looks like an object-like array, let's return that instead
```
orklah, line 477: Maybe a similar check could be done for inferring TList?

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -547,4 +413,200 @@ public static function infer(
...
+            );
+
+            if ($single_item_key_type) {
+                $array_creation_info->item_key_atomic_types = array_merge(
+                    $array_creation_info->item_key_atomic_types,
+                    array_values($single_item_key_type->getAtomicTypes())
+                );
+            }
+        } else {
+            $array_creation_info->item_key_atomic_types[] =  new Type\Atomic\TInt();
```
orklah, line 539: We could probably directly put a TLiteralInt with $int_offset in here I think

orklah (approved):
Seems great. Could there be an opportunity to mutualize the code here with what's in ArrayAnalyzer?

weirdan:
> Could there be an opportunity to mutualize the code here with what's in ArrayAnalyzer?

To share? I thought so, e.g. `handleUnpackedArray()` could technically be shared, at a cost of some additional checks. But there is sufficient number of differences and a lot of what's possible in dynamic array just cannot happen in static arrays. So I think we'll just end up making those two parts similar, but separate.

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
weirdan: You mean, for cases where it crossed the threshold for `can_create_objectlike`?

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
weirdan: Perhaps. That wouldn't be a refactoring in its strict sense though :smile:

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
orklah: I saw multiple cases where can_create_objectlike can be false, so any of those I guess

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
orklah: fair enough :)

Issue #6289, opened 2021-08-11
Wrong implicit integer key inferred in static arrays


weirdan:
https://psalm.dev/r/d7d194894c

**Expected**: key for the "aa" element should be inferred to be 6 (https://3v4l.org/TYqhH)
**Actual**: it's inferred to be 2

PR #6290, opened 2021-08-11
Improve constant array inference

weirdan:
This fixes int offset inference and adds handling for array unpacks in constant arrays. Mostly it's copy/pasted from `ArrayAnalyzer` with further adjustments.

Fixes vimeo/psalm#6263
Fixes vimeo/psalm#6289


orklah (approved):
Seems great!

PR #6306, opened 2021-08-15
Change > assertion back to @ and put back a perf threshold

orklah:
This PR put back some changes that were not part of the core of https://github.com/vimeo/psalm/pull/6210 and were reverted along with the rest in https://github.com/vimeo/psalm/pull/6304


weirdan:
Thanks!

PR #6312, opened 2021-08-15
Fix issues in php 8.0 delta

TysonAndre:
In a last minute change in php 8.0 after the feature freeze,
PhpToken::getAll was changed to tokenize

Also, https://www.php.net/manual/en/function.mb-get-info
has been around since php 4.2 - it's in psalm/dictionaries/CallMap.php
already with the same signature. Same for ReflectionNamedType::isBuiltin

Also, add signatures for getAttributes.

Also, use the same param names for fdiv that are found in reflection in
php 8.0 and 8.1 to support named arguments


weirdan:
I'm going to fix issues in this PR.

TysonAndre:
Oh, CallMap.php is using 8.0 as the base. And mb_get_info should be in CallMap_historical.php. I'd assumed that psalm was still using 7.x as a base

> I'm going to fix issues in this PR.

Will let you do that since you've already started

weirdan:
Yeah, 'historical' is for 7.0 and below, while `CallMap.php` is targeting current version (8.0).

weirdan:
Thanks!

PR #6320, opened 2021-08-16
Fix false positive when passing null to socket_select

### LINKED ISSUE #6318: False positive when passing null to socket_select
```
<?php

$null = null;
$socket = socket_create(AF_INET, SOCK_STREAM, SOL_TCP);
$clients = array($socket);
socket_select($clients, $null, $null, $null, 0);
```

results in `ERROR: NullArgument - 6:39 - Argument 4 of socket_select cannot be null, null value provided to parameter with type int`

However the expected input is `?int` and a null variable can be passed.

```
socket_select(
    ?array &$read,
    ?array &$write,
    ?array &$except,
    ?int $seconds,
    int $microseconds = 0
): int|false
```
https://www.php.net/manual/en/function.socket-select.php

https://psalm.dev/r/28b3494123

Forceu:
Fixes  #6318

According to PHP documentation the function is

```
socket_select(
    ?array &$read,
    ?array &$write,
    ?array &$except,
    ?int $seconds,
    int $microseconds = 0
): int|false
```

however Psalm did not allow `null` for `$seconds`


weirdan (approved):

weirdan:
Thanks!

PR #6321, opened 2021-08-16
Allow type inference on dynamic ClassConstFetch

### LINKED ISSUE #5096: Object instance's string constants cannot be used as array keys with other hard-coded strings
As in my example, using the class constant works, but using the constant directly from an instance does not work.

https://psalm.dev/r/496e0e40cc

orklah:
fix #5096 (almost)

I had to make the class in the example final to keep consistency with the code above. The whole block of code is copy/pasted from above with a few changes that makes it hard to refactor, even more given the code must be placed after analyzing the expression.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
@@ -32,6 +32,10 @@ use function strtolower;
  */
 class ClassConstFetchAnalyzer
 {
+    /**
+     * @psalm-suppress ComplexMethod to be refactored. We should probably regroup the two big if about $stmt->class and
+     * analyse the ::class int $stmt->name separately
+     */
     public static function analyze(
         StatementsAnalyzer $statements_analyzer,
         PhpParser\Node\Expr\ClassConstFetch $stmt,
@@ -475,49 +479,12 @@ class ClassConstFetchAnalyzer
                     $statements_analyzer,
                     $stmt->class,
                     $fq_class_name,
-                    $context->calling_method_id,
-                    false,
-                    false
+                    $context->calling_method_id
                 );
             }
 
             if ($codebase->classlikes->classExists($fq_class_name)) {
                 $fq_class_name = $codebase->classlikes->getUnAliasedName($fq_class_name);
-            }
-
-            if ($stmt->name instanceof PhpParser\Node\Identifier && $stmt->name->name === 'class') {
-                if ($codebase->classlikes->classExists($fq_class_name)) {
-                    $const_class_storage = $codebase->classlike_storage_provider->get($fq_class_name);
-                    $fq_class_name = $const_class_storage->name;
-
-                    if ($const_class_storage->deprecated && $fq_class_name !== $context->self) {
-                        if (IssueBuffer::accepts(
-                            new DeprecatedClass(
-                                'Class ' . $fq_class_name . ' is deprecated',
-                                new CodeLocation($statements_analyzer->getSource(), $stmt),
-                                $fq_class_name
-                            ),
-                            $statements_analyzer->getSuppressedIssues()
-                        )) {
-                            // fall through
-                        }
-                    }
-                }
-
-                $statements_analyzer->node_data->setType($stmt, Type::getLiteralClassString($fq_class_name));
-
-                if ($codebase->store_node_types
-                    && !$context->collect_initializations
-                    && !$context->collect_mutations
-                ) {
-                    $codebase->analyzer->addNodeReference(
-                        $statements_analyzer->getFilePath(),
-                        $stmt->class,
-                        $fq_class_name
-                    );
-                }
-
-                return true;
             }
 
             // if we're ignoring that the class doesn't exist, exit anyway
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
@@ -467,6 +467,18 @@ class ClassConstFetchAnalyzer
                 $fq_class_name = $codebase->classlikes->getUnAliasedName($fq_class_name);
             }
 
+            $moved_class = false;
+
+            if ($codebase->alter_code) {
+                $moved_class = $codebase->classlikes->handleClassLikeReferenceInMigration(
+                    $codebase,
+                    $statements_analyzer,
+                    $stmt->class,
+                    $fq_class_name,
+                    $context->calling_method_id
+                );
+            }
+
             // if we're ignoring that the class doesn't exist, exit anyway
             if (!$codebase->classlikes->classOrInterfaceOrEnumExists($fq_class_name)) {
                 return true;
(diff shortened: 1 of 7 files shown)


orklah:
Note: I tried to refactor. I was pretty happy with the result but there was still a ComplexMethod issue... I don't get what it checks.
The CI was failing anyway, the refactor was broken.

weirdan:
> there was still a ComplexMethod issue... I don't get what it checks.

Intuitively, it's a number of paths through the method, given that extracting a method with a couple of conditionals from the top of the original method makes the most impact on the metrics it reports.

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
```diff
@@ -452,6 +456,255 @@ public static function analyze(
...
+                        if (strtolower($new_fq_class_name) !== strtolower($fq_class_name)) {
+                            $file_manipulations[] = new \Psalm\FileManipulation(
+                                (int) $stmt->class->getAttribute('startFilePos'),
+                                (int) $stmt->class->getAttribute('endFilePos') + 1,
+                                Type::getStringFromFQCLN(
+                                    $new_fq_class_name,
+                                    $statements_analyzer->getNamespace(),
+                                    $statements_analyzer->getAliasedClassesFlipped(),
+                                    null
+                                )
```
weirdan, line 637: Am I mistaken, or would this replace the variable with the class name, thus changing the intent of the code?

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
```diff
@@ -452,6 +456,255 @@ public static function analyze(
...
 
+        if ($stmt->class instanceof PhpParser\Node\Expr\Variable) {
+            $type = $statements_analyzer->node_data->getType($stmt->class);
+
+            $fq_class_name = null;
+            if ($type !== null && $type->isSingle()) {
+                $atomic_type = \array_values($type->getAtomicTypes())[0];
+                if ($atomic_type instanceof TNamedObject) {
+                    $fq_class_name = $atomic_type->value;
+                }
```
weirdan, line 467: We could add support for class-strings here, to be able to infer more type, e.g.: https://3v4l.org/9Kdhd

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
```diff
@@ -616,39 +594,6 @@ public static function analyze(
...
-                                    null
-                                )
-                            );
-                        }
-
-                        $file_manipulations[] = new \Psalm\FileManipulation(
-                            (int) $stmt->name->getAttribute('startFilePos'),
-                            (int) $stmt->name->getAttribute('endFilePos') + 1,
-                            $new_const_name
-                        );
```
weirdan: I think this piece was useful (unlike the one that tried to rename the class).

Renaming `C::CC` to `C::DD` here
```php
final class C { public const CC = 1; }
$c = C::class;
echo $c::CC;
```
should result in
```php
echo $c::DD;
```
but currently results in
```php
echo $c::CC; // should be DD
```

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
orklah: Oh, I get it now. Thanks

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ClassConstFetchAnalyzer.php
orklah: Thanks!

weirdan:
> (almost)

Is there's something missing?

orklah:
Yeah, I explained below that the given example was still giving a `mixed`. This is due to the class in the example not being final.

However, I just checked again and the final thing was only applied for `static::` so I don't think we should apply it when we have the type. I got rid of it so the original example should be completely solved

weirdan:
Hm, I don't think we should allow constant type inference on non-final classes when accessing them dynamically. Consider this code:
```php
<?php

class A
{
    public const C = 1;
}

class B extends A
{
    public const C = "one";
}

function f(A $a): int
{
    return $a::C;
}

f(new B);
```
It passes with flying colors on your branch, but results in runtime errors: https://3v4l.org/F6nbs

Master produces more sensible mixed result: https://psalm.dev/r/97a2a87c7c

weirdan:
Unless constant is inherited from an interface, as those cannot be overridden.

orklah:
I have to think more about that. The thing that bothers me is that in the two tests I added, the type of the class is known with certainty, it can't be a child of the object so I feel like Psalm should know that and not require the class to be final. I'm really not sure Psalm keeps tabs on that sort of thing though and I wonder how hard it could be to add that.

orklah:
I tried introducing a new flag to make Psalm aware of case where the type of the class is known not to be a child. It seems to be working well. I didn't try to cover every case where the type is known, I think it's something we can refine on the go.

I added the flag to TNamedObject and TLiteralClassString but I think it may be useful in TClassString too (in some edge case where we do `::class` on something we don't know the type of). For now I declared two separate properties, but it may have been better with a Trait (I don't really like traits...)

tests/ConstantTest.php
```diff
@@ -314,6 +314,27 @@ class Foo {
...
+            'dynamicClassConstFetchClassString' => [
+                '<?php
+                    class C {
+                        public const CC = 1;
+                    }
+
+                    $c = C::class;
+                    $d = $c::CC;',
+                'assertions' => ['$d===' => '1'],
+            ],
```
weirdan, line 337: It would be great to add couple more tests, to showcase the difference between `can_be_inheritance=true` (or whatever we end up calling it) and the opposite of that.

src/Psalm/Type/Atomic/TNamedObject.php
```diff
@@ -27,17 +27,24 @@ class TNamedObject extends Atomic
      */
     public $was_static = false;
 
+    /**
+     * Whether or not this type can represent a child of the class named in $value
+     * @var bool
+     */
+    public $can_be_inheritance = false;
```
weirdan, line 34: I think the name can be improved. How about something along the lines of `includes_descendants`, or `definite_class`?

tests/ConstantTest.php
orklah: done, I added a case with final and a case without final nor a definite_class to show `mixed` result

src/Psalm/Type/Atomic/TNamedObject.php
orklah: done

tests/ConstantTest.php
```diff
@@ -314,6 +314,51 @@ class Foo {
...
+                    {
+                        public const BAR = "bar";
+                    }
+
+                    /** @var Foo $foo */
+                    $foo = new stdClass();
+                    $_trace = $foo::BAR;',
+                'assertions' => ['$_trace' => 'mixed'],
+            ],
+            'unsafeFinalInferenceClassConstFetch' => [
```
weirdan, line 340: Is it *un*safe?

weirdan (approved):
I think I have one last nitpick - the default value for `$definite_class`. There's quite a few of direct instantiations of both types, and a couple of calls to `getLiteralClassString()` that omit the argument. Historically we mostly treated `TNamedObject` as something that allows descendants - so shouldn't we default that flag to `false` everywhere, for both types, and only enable it where we know we want it?

tests/ConstantTest.php
orklah: What I meant was it was unsafe on the definite_class side of things but final meant you could still make the inferrence, but you're right, it was not clear.

orklah:
Oh, my bad, the wording flipped the whole meaning and I didn't update values. Before, it was true when it could be a child, now it should be true when it can't

src/Psalm/Type/Atomic/TLiteralClassString.php
```diff
@@ -16,9 +16,9 @@ class TLiteralClassString extends TLiteralString
      * Whether or not this type can represent a child of the class named in $value
      * @var bool
      */
-    public $definite_class = false;
+    public $definite_class = true;
 
-    public function __construct(string $value, bool $definite_class = true)
+    public function __construct(string $value, bool $definite_class = false)
```
weirdan: I find different default in constructor signature and property definition a little confusing, even though it shouldn't matter in practice.

weirdan:
This is safe (interface constants cannot be overridden) but is currently not inferred:

```php
interface A
{
    public const C = 1;
}

function f(A $a): int
{
    return $a::C;
}
```

It can be fixed separately though.

orklah:
Yeah, I saw the issue you made. I think this PR is big enough, let's merge that and we'll check that later

PR #6326, opened 2021-08-17
Prevent @property annotations on class properties

### LINKED ISSUE #2164: Prevent @property annotations on class properties
`@property` is valid PHPDoc, but only in docblocks for class elements.

`InvalidDocblock` should be emitted when Psalm encounters something like

```php
class A {
  /** @property string[] */
  public array $arr;
}
```

piporoid:
fixes #2164


weirdan (approved):

weirdan:
Thanks!

PR #6327, opened 2021-08-17
Enable `ClassLikes` to properly detect implementing class aliases

### LINKED ISSUE #6325: Psalm cannot properly infer interfaces when these are declared via `class_alias`
When using `class_alias` to replace an existing interface, psalm is not able to verify that the implementing class implements that interface.

https://psalm.dev/r/d0893a607e

boesing:
Fixes #6325

### DIFF
--- a/src/Psalm/Internal/Codebase/ClassLikes.php
+++ b/src/Psalm/Internal/Codebase/ClassLikes.php
@@ -680,11 +680,9 @@ class ClassLikes
         }
 
         foreach ($class_storage->class_implements as $implementing_interface_lc => $_) {
-            if (!isset($this->classlike_aliases[$implementing_interface_lc])) {
-                continue;
-            }
-
-            $aliased_interface_lc = strtolower($this->classlike_aliases[$implementing_interface_lc]);
+            $aliased_interface_lc = strtolower(
+                $this->getUnAliasedName($implementing_interface_lc)
+            );
 
             if ($aliased_interface_lc === $interface_id) {
                 return true;
--- a/src/Psalm/Internal/Codebase/ClassLikes.php
+++ b/src/Psalm/Internal/Codebase/ClassLikes.php
@@ -26,6 +26,7 @@ use Psalm\Storage\ClassLikeStorage;
 use Psalm\Type;
 use ReflectionProperty;
 
+use function array_keys;
 use function array_merge;
 use function array_pop;
 use function count;
@@ -675,7 +676,23 @@ class ClassLikes
         }
         $class_storage = $this->classlike_storage_provider->get($fq_class_name);
 
-        return isset($class_storage->class_implements[$interface_id]);
+        if (isset($class_storage->class_implements[$interface_id])) {
+            return true;
+        }
+
+        foreach (array_keys($class_storage->class_implements) as $implementing_interface_lc) {
+            if (!isset($this->classlike_aliases[$implementing_interface_lc])) {
+                continue;
+            }
+
+            $aliased_interface_lc = strtolower($this->classlike_aliases[$implementing_interface_lc]);
+
+            if ($aliased_interface_lc === $interface_id) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     public function interfaceExists(
--- a/src/Psalm/Internal/Codebase/ClassLikes.php
+++ b/src/Psalm/Internal/Codebase/ClassLikes.php
@@ -26,7 +26,6 @@ use Psalm\Storage\ClassLikeStorage;
 use Psalm\Type;
 use ReflectionProperty;
 
-use function array_keys;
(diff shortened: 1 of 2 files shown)


tests/Internal/Codebase/ClassLikesTest.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public function testWillDetectClassImplementingAliasedInterface(): void
+    {
+        $this->classLikes->addClassAlias('Foo', 'bar');
+
+        $classStorage = new ClassLikeStorage('Baz');
+        $classStorage->class_implements['bar'] = 'Bar';
+
+        $this->storageProvider->addMore(['baz' => $classStorage]);
```
boesing, line 114: Not sure if there is a better way but this was actually the "easiest" way to provide a `classStorage` to the storage provider.

boesing:
Early on, I was thinking if it would be a thing to have `class_alias` interfaces as part of `class_implements` as well. So that the map wont just contain the FQCN but also the alias.

So in my example I've created in the unit test:

```php
$classStorage = new ClassLikeStorage('Baz');
$classStorage->class_implements = [
    'bar' => 'Bar',
    'foo' => 'Foo',
];
```

But this would somehow require psalm to detect `class_alias` and then iterating over all classes it already infered (and cached) and adding additional details based on all detected `class_alias`. Not sure if this would be doable and thus, I've added the detection to the `ClassLikes` logic instead.

src/Psalm/Internal/Codebase/ClassLikes.php
```diff
@@ -675,7 +676,23 @@ public function classImplements(string $fq_class_name, string $interface): bool
         }
         $class_storage = $this->classlike_storage_provider->get($fq_class_name);
 
-        return isset($class_storage->class_implements[$interface_id]);
+        if (isset($class_storage->class_implements[$interface_id])) {
+            return true;
+        }
+
+        foreach (array_keys($class_storage->class_implements) as $implementing_interface_lc) {
```
weirdan, line 683:
```suggestion
        foreach ($class_storage->class_implements as $implementing_interface_lc => $_) {
```
is marginally faster and is just as clear, IMHO

src/Psalm/Internal/Codebase/ClassLikes.php
```diff
@@ -675,7 +676,23 @@ public function classImplements(string $fq_class_name, string $interface): bool
...
+            if (!isset($this->classlike_aliases[$implementing_interface_lc])) {
+                continue;
+            }
+
+            $aliased_interface_lc = strtolower($this->classlike_aliases[$implementing_interface_lc]);
```
weirdan, line 688: It's probably better to use `ClassLikes::getUnAliasedName()` rather than access `$classlike_aliases` directly. It would likely make fixing #6330 easier.

tests/Internal/Codebase/ClassLikesTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * @var VoidProgress
+     */
+    private $progress;
+
+    /**
+     * @var FileStorageProvider
+     */
+    private $fileStorageProvider;
```
weirdan, line 64: Please use `snake_case` for variables and properties.

tests/Internal/Codebase/ClassLikesTest.php
```diff
@@ -0,0 +1,118 @@
...
+     */
+    private $reflection;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->storageProvider = new ClassLikeStorageProvider();
+        $this->fileReferenceProvider = new FileReferenceProvider();
+        $this->statementsProvider = new StatementsProvider($this->file_provider);
+        $this->providers = new Providers($this->file_provider);
```
weirdan, line 77: After `parent::setUp()` call you already have a fully initialized `ProjectAnalyzer` that is reinitialized for every test method. You can use that to obtain `ClassLikes` and `ClassLikeStorageProvider`:

```suggestion
        $this->classlikes = $this->project_analyzer->getCodebase()->classlikes;
        $this->storage_provider = $this->project_analyzer->getCodebase()->classlike_storage_provider;
```

Or just use those directly in the test method.

tests/Internal/Codebase/ClassLikesTest.php
weirdan: Looks ok to me.

weirdan (changes requested):

tests/Internal/Codebase/ClassLikesTest.php
boesing: Oh nice! Wasn't aware of that. Thanks!

weirdan:
Thanks!

PR #6328, opened 2021-08-17
Fix error message for call to internal method from root namespace

bdsl:
I noticed that the error output linked from https://twitter.com/afilina/status/1425841636282671105 is not grammatical.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallProhibitionAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallProhibitionAnalyzer.php
@@ -56,7 +56,7 @@ class MethodCallProhibitionAnalyzer
                     new InternalMethod(
                         'The method ' . $codebase_methods->getCasedMethodId($method_id)
                             . ' is internal to ' . $storage->internal
-                            . ' but called from ' . $context->self,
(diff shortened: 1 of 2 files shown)


bdsl:
There are probably several other internal code violations that should be fixed - e.g. https://psalm.dev/r/8a680925cf but they can be done separately, and I don't know if it's worth fixing them before I or someone else actually encounters a real case of code that should be internal being onerously called.

weirdan (approved):

weirdan:
> I don't know if it's worth fixing them before I or someone else actually encounters a real case of code that should be internal being onerously called.

If you have time and feel the urge then why not? We don't have any policies like 'must affect N users' or anything like that.

weirdan:
Thanks!

bdsl:
> > I don't know if it's worth fixing them before I or someone else actually encounters a real case of code that should be internal being onerously called.
> 
> If you have time and feel the urge then why not? We don't have any policies like 'must affect N users' or anything like that.

Right, I just mean worth the time for me or anyone else who wants to implement the fix. I might do some point in future, not today. I can report the issue if you think that would be useful in case anyone else wants to implement.

bdsl:
Reported: https://github.com/vimeo/psalm/issues/6329

weirdan:
> I might do some point in future, not today.

Sure, no pressure. We're all volunteers here, there's no obligations of any sort.

Issue #5356, opened 2021-03-09
UnusedPsalmSuppress reported in vendor dir


BenMorel:
I have the following `psalm.xml` config file:

```xml
<?xml version="1.0"?>
...
    <projectFiles>
        <directory name="lib" />
        <directory name="src" />
        <ignoreFiles>
            <directory name="vendor" />
        </ignoreFiles>
    </projectFiles>
</psalm>
```

On first run, the `vendor/bin/psalm --find-unused-psalm-suppress` command returns the following error:

```
ERROR: UnusedPsalmSuppress - vendor/myclabs/php-enum/src/Enum.php:65:33 - This suppression is never used (see https://psalm.dev/207)
            /** @psalm-suppress InvalidCast */
```

I cannot reproduce it on second run. Is this a bug?

muglug:
> I cannot reproduce it on second run. Is this a bug?

Yes, anything that changes dependent on caching is a bug

This is a bit of an edge-case, but still a bug

orklah:
@BenMorel would you mind creating a small repo that replicate this issue?

BenMorel:
@orklah I haven't seen the bug in a while, let's consider it fixed. I'll come back if it appears again!

simPod:
@orklah I've managed to reproduce it here https://github.com/simPod/PsalmBugs/pull/1

PR #6344, opened 2021-08-19
don't report UnusedSuppressions in non project dirs

orklah:
This will fix https://github.com/vimeo/psalm/issues/5356

Thanks to @simPod for the reproducer!


orklah:
Oh, the irony.

Psalm's test cases fake files are not included in project files so this was failing the test. I added a fake file and used it to make the errors go away

weirdan (approved):

weirdan:
Thanks!

PR #6345, opened 2021-08-19
use better solution to check for project dir paths

orklah:
This is a small refactor of an old fix I made in https://github.com/vimeo/psalm/pull/4658. I didn't like the way I did it back then, now I have a better solution, I change it for the best :)


weirdan (approved):

weirdan:
Thanks!

Issue #6300, opened 2021-08-13
Asserts with constant wildcards


weirdan:
https://psalm.dev/r/fc2b99f6eb

weirdan:
When unions are spelled literally (`"a"|"b"|"c"`) it does work for `@psalm-assert`, but doesn't for `@psalm-assert-if-true`: https://psalm.dev/r/d1b97805cd

weirdan:
Duplicate of #5939
Possibly related: #5372
Was once rejected in #2452, but it appears it was reconsidered (as it works for explicit unions with `@psalm-assert`)

boesing:
@orklah it is broken for class-constants. It worked somewhen in an older version (4.7.0 for sure) but stopped working in later versions (havent checked which version broke this).
thats actually why I've created #6529

PR #6347, opened 2021-08-19
Allow unions in `@psalm-assert-if-true`

weirdan:
Refs vimeo/psalm#6300

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -27,6 +27,8 @@ use Psalm\Issue\UnevaluatedCode;
 use Psalm\IssueBuffer;
 use Psalm\Type;
 
+use function array_map;
+use function array_values;
 use function count;
 use function in_array;
 use function is_int;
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -1090,11 +1100,14 @@ protected static function processCustomAssertion(
...
-                        } else {
-                            $if_types[$var_name] = [['!' . $assertion->rule[0][0]]];
-                        }
+                        $if_types[$var_name] = [
+                            array_map(
+                                function ($rule) {
+                                    return $rule[0] === '!' ? (string) substr($rule, 1) : ('!' . $rule);
+                                },
+                                array_values($assertion->rule[0])
+                            )
```
weirdan, line 1109: I'm actually not sure about this. It's supposed to invert the rule, but it effectively turns `"a" | "b" | "c"` into `!"a" | !"b" | !"c"` rather than `!"a" & !"b" & !"c"`

orklah: yeah, that's an issue. This will probably lead to weird edge cases

PR #6375, opened 2021-08-28
Fixing openssl_x509_parse() and openssl_x509_read() Closes #6358

### LINKED ISSUE #6358: Outdated signature of `openssl_x509_parse()`
I'm getting:
> ERROR: InvalidArgument - Argument 1 of openssl_x509_parse expects resource|string, OpenSSLCertificate provided

But `openssl_x509_parse()` has changed its signature to:
```php
openssl_x509_parse(OpenSSLCertificate|string $certificate, bool $short_names = true): array|false
```
See https://www.php.net/manual/function.openssl-x509-parse.php

ThomasLandauer:
Second attempt after https://github.com/vimeo/psalm/pull/6374
Closes #6358
I don't really know what I'm doing here - so please double-check :-)
https://www.php.net/manual/en/function.openssl-x509-parse.php
https://www.php.net/manual/en/function.openssl-x509-read.php


dictionaries/CallMap.php
```diff
@@ -9334,8 +9334,8 @@
 'openssl_x509_export_to_file' => ['bool', 'certificate'=>'string|resource', 'output_filename'=>'string', 'no_text='=>'bool'],
 'openssl_x509_fingerprint' => ['string|false', 'certificate'=>'string|resource', 'digest_algo='=>'string', 'binary='=>'bool'],
 'openssl_x509_free' => ['void', 'certificate'=>'resource'],
-'openssl_x509_parse' => ['array|false', 'certificate'=>'string|resource', 'short_names='=>'bool'],
```
weirdan, line 9337: `short_names` is still optional, so `=` that indicates it has to stay there.

weirdan: Why would `null` be allowed for `short_names`? It results in `TypeError` in PHP 8.

image

weirdan (changes requested):

dictionaries/CallMap.php
ThomasLandauer: Sorry, I was thinking `|null` is the way to indicate it's optional... ;-)

weirdan (approved):

weirdan:
Thanks!

PR #6385, opened 2021-09-01
Add a note about baseline suppression files in the ignoreFiles section

MattyRad:
Hi there, thanks for doing god's work.

Coming from phpstan, I ran into some confusion with ignoring individual files. I figured I'd just take my list of files in `excludes_analyse` from my `phpstan.neon` and pop it into the XML config, but that doesn't appear to be doable. After exploring the docs a bit further, `--set-baseline` is a super smart and easy way kick off psalm usage. It may be helpful to note that in the configuration section itself.

### DIFF
--- a/docs/running_psalm/configuration.md
+++ b/docs/running_psalm/configuration.md
@@ -409,6 +409,8 @@ Contains a list of all the directories that Psalm should inspect. You can also s
 </projectFiles>
 ```
 
+> If you're looking to ignore individual files, consider starting with a [baseline suppression file](https://psalm.dev/docs/running_psalm/dealing_with_code_issues/#using-a-baseline-file) (which can be automatically generated for your existing errors).
+
 #### &lt;extraFiles&gt;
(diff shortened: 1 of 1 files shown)


orklah:
Hi!

The baseline purpose is indeed to kick off Psalm setup, but it doesn't excludes files from the analysis (it just kinda hides the resulting errors when they're already known).

The difference is that every new error to pop up in those files will show up (whereas they wouldn't show if you excluded these from the analysis)

It's a very similar feature PHPStan also offer but it's a separate feature from just ignoring files.

Also, please note that the baseline is already documented in psalm's website here: https://psalm.dev/docs/running_psalm/dealing_with_code_issues/#using-a-baseline-file

MattyRad:
That makes sense, thanks! I'm mainly sharing my personal experience with the docs, I think getting users to that dealing-with-code-issues page is (in my case a least) key to success. I'll close this PR since the changes aren't quite accurate though.

Thanks for your response and maintaining such a great package!

Issue #6317, opened 2021-08-16
False positive `ParadoxicalCondition` when using `in_array()`


corphi:
When given the snippet

```php
function contains(array $list1, array $list2, mixed $element): void
{
    if (in_array($element, $list1, true)) {
    } elseif (in_array($element, $list2, true)) {
    }
}
```

the second condition is mistaken to be identical to the first condition.

See https://psalm.dev/r/e148c5660c

This has been introduced in Psalm 4.9.3.

weirdan:
Also reported by @orklah here: https://github.com/vimeo/psalm/pull/6233#issuecomment-899297077

supersmile2009:
I checked why it happens. Here's a quick summary:
1. Assertion is extracted as `in-array-X`.
2. Inside of `if` clause it's treated as is.
3. For use outside of `if` clause (which includes `elseif` too) it is negated and becomes `!in-array-X`.

Both of those assertions are correctly reconciled by `SimpleAssertionReconciler` (narrows down the type to array value type) and `NegatedAssertionReconciler`(keeps type as is, since `!in-array-X` assertion is inconclusive - variable can still be pretty much anything).
However IfElseAnalyzer as well as the analyzers under `Psalm\Internal\Analyzer\Statements\Block\IfElse` namespace check for paradoxical conditions using the raw (not reconciled) assertions.

One naive solution that I first thought of - just don't negate `in-array` assertion (return no clauses), since such assertion is inconclusive. However negated assertion is still useful in a few cases:
1. post-if scope after negated in_array
```
function assertInArray($x, $y) {
    if (!in_array($x, $y, true)) {
        // clause is negated - resulting in `!in-array-T`
        throw new Exception();
    }
    // Negated clause is negated again, turning it into a much more useful `in-array-T`
    return $x;
}
```
So if we don't return the first negated clause, there won't be the 2nd one.
2. Detecting nonsensical assertions, where there's no type intersection between needle and haystack, e. g. asserting integer variable is not in array of strings.

[...]

PR #6396, opened 2021-09-03
don't infer from in_array with unknown type

### LINKED ISSUE #6333: in_array and TypeDoesNotContainNull
Here is example:
https://psalm.dev/r/274c08c832

orklah:
this should fix #6333


orklah:
It should also fix (or at least diminish the impact of) #6317 by not asserting `in-array-mixed` which doesn't seem useful anyway

orklah:
superseded by #6419

Issue #6382, opened 2021-08-31
Chain static template parameter out of memory


AdeAttwood:
We are getting an out of memory error when we are chaining our `where()`
functions. This I think is coming from `&static` that is getting joined together
when ever we call another `where()`.

With one `where()` we get a type of

```
ActiveQuery<PostSearch&ActiveQuery<PostSearch&static>>&static
```
And then with two

```
ActiveQuery<PostSearch&ActiveQuery<PostSearch&ActiveQuery<PostSearch&ActiveQuery<PostSearch&static>>&static>&ActiveQuery<PostSearch&ActiveQuery<PostSearch&static>>&static>&static
```

I have replicated the error on `psalm.dev` with no errors. If you then add
another `where()` on line 34 you get an out of memory error.

https://psalm.dev/r/311948d849

```php
$query = self::find()
    ->where(['web_ref' => null])
    ->where(['web_ref' => null])
    ->where(['web_ref' => null]);
```

```php
$query = self::find()
    ->where(['web_ref' => null])
    ->where(['web_ref' => null])
    ->where(['web_ref' => null])
    ->where(['web_ref' => null]);
```

AdeAttwood:
@orklah if there is anything I can do to help fix the issue let me know. I have not worked on the internals of psalm before, but If you point me in the right direction, I will have a go.

orklah:
@AdeAttwood I'm not quite sure where to look, but one thing that we already noticed on this is that the `/** @return static */` in phpdoc does not react the same way as `: static` in signature:
https://psalm.dev/r/ce437b0e65

As a workaround, you could declare `: static` if you have control over the ActiveQuery class or create a stub for it.

As for fixing Psalm, it would require investigating why the phpdoc is different than the signature. Maybe try working your way from `ExistingAtomicMethodCallAnalyzer::analyze:208` I think it's the starting point when Psalm calculate the return type for a specific MethodCall.

weirdan:
It appears `static` is being replaced with current object type which also includes several `static`s, so this leads to `static` being multiplied on each step in geometric progression: https://psalm.dev/r/588038dea3

AdeAttwood:
The issue is that the `static` is getting pared differently if you define it in a docblock than when you define it as a return type, as you said in your previous comments. According to psalm they are different types.

I don't  know if we need to forget that fact, there is a memory issue for the moment, I think that is obscuring the issue. I will get what I have into a WIP PR so you can see what I have been hacking on :+1:

[...]

PR #6399, opened 2021-09-04
fix: static return types

AdeAttwood:
When returning static it is acting differently between if it is defined as part
of a docblock or as a return type. The class is getting resolved to "A&static"
when it is returned as a return type, and "static" when resolved as a doc block

```php
abstract class A {
    public function foo(): static {
        return $this;
    }

	/** @return static */
    public function bar() {
        return $this;
    }
}
```

This parses the type in the same way if `static` is defined in the docblock or
in the return type

Ref: #6382


src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php
```diff
@@ -48,8 +48,11 @@ public static function isContainedBy(
             return true;
         }
 
-        $container_has_template = $container_type->hasTemplateOrStatic();
+        if ($input_type->isFormerStaticObject()) {
```
orklah, line 51: Can you explain what that does? I find it weird to be able to assert a type is contained by the other just by looking at one of the types

AdeAttwood: When calculating, static types are now resolved to their class with the `was_static` property true

```php
class A {
    public foo(): static {}
}

class B extends A {}

$b = new B;
$foo = $b->foo(); // $foo == "A&static"

assert($foo instanceof B); // Should work
```
In this situation, `isContainedBy` is called to test that `B` is contained by `$foo` but to psalm its `A&static`, so this will see if it `was_static` and return true because `B::foo` dose contain `A::foo` if the return type of `A::foo` is `static`

Hope that makes sense, its really had to explain in a comment lol.

orklah: when I take your example and run it against your code, there's no difference whether or not the call to isFormerStaticObject is there. Do you have an example of code that is resolved by this method call? I think we need to include an additional check with $container_type to the mix, but I don't see what is the issue being resolved here

AdeAttwood: This was from when I was fixing the test I added in tests/MethodSignatureTest.php:613

orklah: Likewise, I have no errors with or without this condition.

Can you remove it from your PR so we see the impact on every test?

AdeAttwood: That has now been removed :+1:

orklah: Ok, this snippet: https://psalm.dev/r/fcd88c33ed (or MethodSignatureTest with returnStaticParent) fails without your piece of code.

Basically, isContainedBy receives an $input_type `A&static` and a $container_type `B&static`. Those are indeed compatible, but you can't know that by just looking at if `A&static` contains `static`, you have to check that A is a parent of B

AdeAttwood: OK so now I am really stuck. What I have so far is.

```php
if ($input_type->equals($container_type)) {
    return true;
}

if ($input_type->isFormerStaticObject()
 && $codebase->classExtendsOrImplements((string)$container_type, (string)$input_type)
) {
    return true;
}
```

The issue I am having is with the interface test.

```php
abstract class A {
    public function foo(): static {
        return $this;
    }
}

interface I {
    /** @return I */
    public function foo();
}

class C extends A implements I {}
```

When calling `isContainedBy` I am getting `A` for `input_type` and `I` for `container_type`. `A` itself dose not implement `I` but we are checking `C` that dose. In `isContainedBy` I have no access to `C` so cant calculate this, dose this need to be calculated in the function calling `isContainedBy`

orklah: The issue is that this snippet was broken even before your PR: https://psalm.dev/r/2c9b8b1f70

So, trying to have the same behaviour between phpdoc (which is buggy) to another buggy solution is likely to dig up nasty things.

Honestly, I'm quite lost myself here. I'm pretty sure we should be able to find a method that resolve static usages to the actual called class by using some context, but we'll likely hit another wall there.

The solution is less obvious than I thought it would be (unfortunately, it's quite often with Psalm!), so I'd suggest changing strategy and finding where is the runaway memory consumption and try to stop it from happening. It will probably emit strange errors (or no errors when it should) but at least, Psalm will be able to run correctly

orklah:
I'll close this. The solution won't be easy anyway

PR #6400, opened 2021-09-04
Add section on limitations

craigfrancis:
The "Security Analysis in Psalm" page does not cover the Limitations, maybe something like this would give some examples of what developers need to watch out for? (feel free to edit/replace as needed)


weirdan (approved):

weirdan:
Thanks!

PR #6409, opened 2021-09-04
Add documentation on callmap editing

weirdan:


docs/contributing/editing_callmaps.md
```diff
@@ -0,0 +1,81 @@
...
+
+First element of that value is a return type (it also doesn't have a key), and
+subsequent elements represent function/method parameters. Parameter name for an
+optional parameter is postfixed with `=`.
+
+## Delta file format
+
+Delta files (named `CallMap_<PHP major version><PHP minor version>_delta.php`)
+list the changed that happened in the corresponding PHP version. There are
+three section with self-explanatory names: `aded` (for functions/methods that
```
orklah, line 30: typo on `added`

orklah (approved):
Should we bother explaining here the subtleties like `'1` when multiple signature exists, `&` for reference parameters, the fact that Psalm does not follow signature to the letter when param types or return types are the result of developper errors (typically when param became TypeError later) ? Maybe even that functions in callmap are pure by default?

Or maybe keep those out of the simple guide not to afraid contributors?

weirdan:
I intended that as a simple guide indeed. This can be expanded if we observe people confused by subtleties.

PR #6419, opened 2021-09-04
Also support non-literal types in strict in_array() check

### LINKED ISSUE #6317: False positive `ParadoxicalCondition` when using `in_array()`

### LINKED ISSUE #6333: in_array and TypeDoesNotContainNull

TysonAndre:
Modified approach to #6396 by orklah

Fixes #6333
Fixes #6317 

Not very familiar with the approach taken here (in general for type assertions), but `=null` for an assertion seems to not work (strict equality), while `null` does.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -3442,7 +3442,7 @@ class AssertionFinder
                     $assertions = [];
 
                     if (!$is_sealed) {
-                        if ($value_type->getId() !== '') {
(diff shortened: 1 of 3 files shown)


orklah:
Seems to be working 👍 I'll close mine

weirdan:
Thanks!

PR #6420, opened 2021-09-04
Ternary override type when reassigning

### LINKED ISSUE #6384: remove old types when both side of a ternary reassign a variable
https://psalm.dev/r/198b6a7c3a
$_a in this example should not contain null

https://psalm.dev/r/705d3062f6
This edge case could be interesting to check also

PHPStan infers the correct type

orklah:
This fixes #6384.

I feel like TernaryAnalyzer should actually be much more similar to IfElseAnalyzer. I'd bet there is a lot of bugs there that is resolved in the other.

I tried going all the way and creating a set of Virtual statements to push to IfElseAnalyzer but I forgot that there is a slight difference because one is an expression and the other a statement, so you can't just translate one into the other


weirdan:
Thanks!

PR #6423, opened 2021-09-04
2> Count equal 0 assertion

### LINKED ISSUE #5869: Recognition of non-empty-array by count()

### LINKED ISSUE #6393: Return, break, die do not propagate assertions
https://psalm.dev/r/b61522c880

### LINKED ISSUE #6548: Invalid count guessing with optional array indexes
It's possible to declare an array shape with an optional index, in that case the count should assume that the count of items in the array may be either mix or max of array shape, yet it only assumes that the maximum is possible.

https://psalm.dev/r/cbf9b63ac4

### LINKED ISSUE #6551: Arrays checked with `count` should be considered as non-empty-list
Hey,
I encountered some errors of this kind: 
`Argument 1 of max expects non-empty-array<array-key, mixed>, parent type array<array-key, int>`

The thing is that all my calls to the `max` function were not concerned by these errors. The difference was the way I was checking the emptiness of my array. 

https://psalm.dev/r/fa6da48c3f

orklah:
This will fix #6393 and fix #5869. This PR is similar with #5930 but I couldn't finish it at the time because I didn't know about sealed TKeyedArrays and I didn't have IntegerRanges that proved crucial to solve this (this is due to a bug in Psalm, count() was returning a value for non-sealed TKeyedArrays was sometimes incorrect, now an IntegerRange is returned, solving the issue)

EDIT: This will fix #6551
EDIT: This will fix #6548

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -179,7 +179,10 @@
     </PossiblyUndefinedIntArrayOffset>
   </file>
   <file src="src/Psalm/Internal/Analyzer/Statements/UnusedAssignmentRemover.php">
-    <PossiblyUndefinedIntArrayOffset occurrences="3">
+    <PossiblyUndefinedIntArrayOffset occurrences="6">
+      <code>$token_list[$iter]</code>
+      <code>$token_list[$iter]</code>
+      <code>$token_list[$iter]</code>
       <code>$token_list[$iter]</code>
       <code>$token_list[0]</code>
       <code>$token_list[1]</code>
--- a/src/Psalm/Internal/Analyzer/Statements/Block/DoAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/DoAnalyzer.php
@@ -31,6 +31,7 @@ class DoAnalyzer
     ): void {
         $do_context = clone $context;
         $do_context->break_types[] = 'loop';
+        $do_context->inside_loop = true;
 
         $codebase = $statements_analyzer->getCodebase();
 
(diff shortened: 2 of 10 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/NonDivArithmeticOpAnalyzer.php
```diff
@@ -702,17 +821,24 @@ private static function analyzeNonDivOperands(
                     }
 
                     if ($parent instanceof PhpParser\Node\Expr\BinaryOp\Mod) {
-                        if ($always_positive) {
-                            if ($right_type_part instanceof TLiteralInt && $right_type_part->value === 1) {
-                                $result_type = Type::getInt(true, 0);
+                        if ($right_type_part instanceof TLiteralInt) {
+                            $literal_value_max = $right_type_part->value - 1;
```
TysonAndre, line 825:
```suggestion
                            $literal_value_max = abs($right_type_part->value) - 1;
```


```
php > var_export(-5 % -3);
-2
```

orklah: Thanks for the review! However, it would be easier to do it on this PR #6241. This one is based on #6241 so most of the changes are coming from there (including this one :) )

weirdan (approved):

weirdan:
Thanks!

Issue #6427, opened 2021-09-05
Enums: already used value for several cases


weirdan:
psalm.dev/r/108275cc1c?php=8.1

weirdan:
https://psalm.dev/r/108275cc1c?php=8.1

Issue #6426, opened 2021-09-05
Enums: already used case


weirdan:
https://psalm.dev/r/ceaa9c9892?php=8.1

weirdan:
psalm.dev/r/ceaa9c9892

weirdan:
https://psalm.dev/r/ceaa9c9892?php=8.1

PR #6437, opened 2021-09-06
Basic enum checks

weirdan:
* Duplicate cases
* Duplicate case values
* Invalid case values: value on a pure enum case, missing value on a backed enum case, backing type / case type mismatch
* Adds literal expression evaluation, e.g. `enum Mask: int { case Two = 1 << 1; }`

Fixes vimeo/psalm#6426
Fixes vimeo/psalm#6427

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -19,11 +19,12 @@ use Psalm\Internal\Type\TemplateStandinTypeReplacer;
 use Psalm\Issue\DeprecatedClass;
 use Psalm\Issue\DeprecatedInterface;
 use Psalm\Issue\DeprecatedTrait;
+use Psalm\Issue\DuplicateEnumCaseValue;
 use Psalm\Issue\ExtensionRequirementViolation;
 use Psalm\Issue\ImplementationRequirementViolation;
 use Psalm\Issue\InaccessibleMethod;
 use Psalm\Issue\InternalClass;
-use Psalm\Issue\InvalidEnumBackingType;
+use Psalm\Issue\InvalidEnumCaseValue;
 use Psalm\Issue\InvalidExtendClass;
 use Psalm\Issue\InvalidTemplateParam;
 use Psalm\Issue\InvalidTraversableImplementation;
@@ -70,6 +71,8 @@ use function count;
 use function explode;
 use function implode;
 use function in_array;
+use function is_int;
+use function is_string;
 use function preg_match;
 use function preg_replace;
 use function str_replace;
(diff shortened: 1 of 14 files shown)


src/Psalm/Internal/Analyzer/ClassAnalyzer.php
```diff
@@ -2627,5 +2635,119 @@ private function checkEnum(PhpParser\Node\Stmt\Enum_ $enum): void
...
+                            $enum->name->name
+                        )
+                    )) {
+                        continue;
+                    }
+                } elseif (null !== $enum->scalarType && null !== $stmt->expr) {
+                    if (ExpressionAnalyzer::analyze(
+                        new StatementsAnalyzer(
+                            $this,
+                            $type_provider = new \Psalm\Internal\Provider\NodeDataProvider()
```
orklah, line 2682: Can you explain why you had to do that? You couldn't fetch the existing one? Or was there not one yet in the execution?

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
```diff
@@ -2627,5 +2635,119 @@ private function checkEnum(PhpParser\Node\Stmt\Enum_ $enum): void
...
+                        ),
+                        $stmt->expr,
+                        $enum_context,
+                        false,
+                        $global_context,
+                        true
+                    ) === false) {
+                        // xxx: should we rather return here?
```
orklah, line 2690: I think we should. false returns happen when things go pretty wrong,  like UndefinedClass wrong. When that happens, it's no use trying to understand what's going on for the rest of the scope

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
```diff
@@ -2627,5 +2635,119 @@ private function checkEnum(PhpParser\Node\Stmt\Enum_ $enum): void
...
+                        continue;
+                    }
+
+                    $case_type = $type_provider->getType($stmt->expr);
+
+                    if (null === $case_type) {
+                        // xxx: can this even happen?
```
orklah, line 2697: I've seen it happen a lot when building plugins but I had no way to see what happened in the actual analysis so it may have been case of UndefinedClass like above...

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
weirdan: Because it was 2 AM, mostly =)

I probably didn't have to. There should be already an instance of enum storage populated that has `$enum_cases` property, and if that doesn't support cases like https://3v4l.org/YV9Xm/rfc#vgit.master it should be probably made to, using `SimpleTypeInferrer`.

orklah (approved):

PR #6438, opened 2021-09-06
Changed signature of Serializable::unserialize stub

SCIF:
Signature aligned with https://www.php.net/manual/en/serializable.unserialize.php


SCIF:
I can see there are some test crashes but want to know if @orklah and @weirdan are fine with it.

orklah:
There is a concrete difference between phpdoc and types in signature, even in stubs.

To have a compatible implementation, you have to respect types in parent signature according to two rules:
- You can always declare a parameter type with a wider type
- You can always declare a return type with a more narrow type

This means that when you add a param type in signature, the implementation may keep its type empty (because empty is mixed and mixed is wider). However, as soon as you declare a return type, you require that each implementation declare a return type at least equal to what you added. This is why tests fail.

In conclusion, you may keep types in signature for parameters but you have to keep them in phpdoc for return types

SCIF:
@orklah , @weirdan ,
Guys, don't you think, this one is the outdated test and should be deleted?
```            'preventImplementingSerializableWithType' => [
                '<?php
                    class Foo implements \Serializable {
                        public function unserialize(string $serialized) {}
                        public function serialize() {}
                    }',
                'error_message' => 'InvalidReturnType',
            ],
```

Or moved into valid cases :)

orklah:
@SCIF see https://3v4l.org/ZGMmf

Psalm support PHP 7.1 and above. What must be done is add a version number to this test so it's only checked for a specific version

weirdan:
@SCIF 7.4 is still in active support phase, so it's too soon to call that outdated. It's just version-dependent - valid for 8.0+, but invalid for <8.0.

orklah:
however this probably means we need to have stubs that depend on the php version. We already have a Php80.phpstub, but I think Psalm will fail if we have a class in two different stub, so we would need to have a Php7(1|2|3|4).phpstub to contain old signatures

SCIF:
[The test crashed](https://github.com/vimeo/psalm/pull/6438/checks?check_run_id=3524159904) is actually [incorrect](https://3v4l.org/kI1PR).

orklah:
That seem correct, you can get rid of the test

SCIF:
Done :)

orklah (approved):

orklah:
Thanks!

PR #6441, opened 2021-09-06
Force array_walk to not infer input type.

### LINKED ISSUE #6394: ReferenceConstraintViolation seen after calling `array_walk(array|object &rw_array, ...)`
https://psalm.dev/r/5c8bd06393

It seems like array_walk accepts arrays or objects, which leads psalm to infer that the type may be changed to `object` after the function is called?

array_walk_recursive is also affected

```
dictionaries/CallMap.php
423:'array_walk' => ['bool', '&rw_array'=>'array|object', 'callback'=>'callable', 'arg='=>'mixed'],
```

Psalm itself ignores the `rw_` prefix (used by Phan) so I don't think that's related. There aren't any special cases for array_walk at all other than warning about raw object iteration. The lack of special cases is probably the issue, psalm is treating `array_walk` just like any other function accepting references and outputting values in the references.

niconoe-:
Attempt to fix #6394 

I also added a test case but I'm not sure I did it in the right place, so please feel free to move it at its right location.

Have a nice day,
Regards,


weirdan:
Thanks!

PR #6444, opened 2021-09-06
introducing fsync & fdatasync functions in callmap

SMAtaurRahman:
This PR adds callmap for two new functions that are introduced in PHP 8.1

- fsync
- fdatasync

RFC: https://wiki.php.net/rfc/fsync_function


weirdan (approved):

weirdan:
Thanks!

Issue #6439, opened 2021-09-06
Another false positive `ParadoxicalCondition` when using `in_array()`


corphi:
As a follow-up to #6317:

The issue persists when types are used. This snippet triggers the same false positive:

```php
/**
 * @param string[] $list1
 * @param string[] $list2
 */
function contains(array $list1, array $list2, string $element): void
{
    if (in_array($element, $list1, true)) {
    } elseif (in_array($element, $list2, true)) {
    }
}
```

When any of the array types are dropped, no error is reported.

See https://psalm.dev/r/1fc70cf120

orklah:
@TysonAndre that's one thing I was afraid of, and I think this was triggered by adding support for non-literal types in your PR. Is there a specific code that will stop to work correctly without this support?

TysonAndre:
I think my change only affected the sealed array (TKeyedArray, i.e. array literals) and not the branch being referred to, but I may be mistaken

But my change still introduced a bug. For now, the conservative approach of returning the empty array for non-literals would avoid the false positives at the cost of not inferring types.

```php
function contains2(string $a, string $b, $element): void
{
    if (in_array($element, [$a], true)) {
    } elseif (in_array($element, [$b], true)) { // false positive Condition ($element is string) contradicts a previously-established condition ($element is not string)
    }
}
```

The code for handling strict equality in AssertionFinder::scrapeEqualityAssertions may be of use.

Unfortunately,
- The example only works for union types with a single type (literal or non-literal)
- The isSingle check seems to be related to union types only working for getAssertionString with a single type
- `$var_type->getAssertionString() !== $assertion` for atomic types seems useful in avoiding redundant condition warnings, but not paradoxical condition warnings. Type\Union::getAssertionString won't work for variables with more than one type, it only works for union types of a single type, which is why I see isSingle checks elsewhere
- Not sure why that code in scrapeEqualityAssertions is using `~` for approximate equality. Possibly a bug or workaround. Seems to be used in AssertionReconciler::reconcile(). Probably, it's so that negated the condition such as `if ($var === $otherString)` in the else branch isn't treated as a non-string, which happened when I tried to change it.

  Changing it would do a better job of removing mixed, though
- Not sure why getAssertionString() sometimes needs to be prefixed with `=` (strict equality) to be a valid assertion string. I'm guessing there might be special cases for `null` that aren't written for `=null`.

```php
                // both side of the Identical can be asserted to the intersection of both
[...]

PR #6445, opened 2021-09-06
Fix false positive ParadoxicalCondition in negation of in_array

TysonAndre:
For #6439

This turned out to be a better approach than making the negation of `in_array` equal to `mixed`.
`mixed` is changing the type of the value to mixed. Also, double negation wouldn't work in the previously considered approach.

`in-array-x` and `!in-array-x` do not contradict each other (a string might be in an array of strings, but not in another array (such as the empty array)), so prevent the warning from being emitted.

### DIFF
--- a/src/Psalm/Internal/Analyzer/AlgebraAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/AlgebraAnalyzer.php
@@ -12,6 +12,7 @@ use Psalm\IssueBuffer;
 use function array_intersect_key;
 use function array_unique;
 use function count;
+use function preg_match;
 
 /**
  * @internal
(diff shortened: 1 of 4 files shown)


TysonAndre:
This seems to be working as intended - no other changes are planned if there are no comments. This takes advantage of the fact that (by design?) `in-array-T` can be used to indicate that something may be an instance of T, while having a negation that does not imply anything about the type, since the array could be empty or contain an unrelated value that's part of the type T.

```php
<?php
/**
 * @param mixed $element
 */
function contains(int $i, string $s, $element): void
{
    if (in_array($element, [$i, $s, false], true)) {
        /** @psalm-trace $element */  // correctly infers false|int|string
    } elseif ($element === false) {  // Correctly emits ParadoxicalCondition
    } elseif (is_int($element)) {  // Correctly does not warn
    }
}
```

orklah:
Thanks!

PR #6453, opened 2021-09-07
Fix return-type for preg_filter

### LINKED ISSUE #6452: Incorrect RedundantConditionGivenDocblockType / DocblockTypeContradiction
See https://psalm.dev/r/628ac07bfc

I don't see how this is legitimate.. `preg_filter()` returns either a `string` or `null` (if fed with a string)..

tvdijen:
Closes #6452


weirdan:
Thanks!

PR #6462, opened 2021-09-08
type of "$http_response_header" is "list<string>" now

### LINKED ISSUE #6456: Incorrect type for reserved $http_response_header variable: array<array-key, mixed> instead of list<string>
https://psalm.dev/r/ca2f17ab48

According to the PHP documentation, $http_response_header is a list of strings.
https://www.php.net/manual/en/reserved.variables.httpresponseheader.php

SMAtaurRahman:
Fixes #6456 

Let me know if this one requires test cases. I couldn't figure out how/where to write tests.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
@@ -550,6 +550,12 @@ class VariableFetchAnalyzer
             return Type::getInt();
         }
 
+        if ($var_id === '$http_response_header') {
+            return new Type\Union([
+                new Type\Atomic\TList(Type::getString())
+            ]);
+        }
+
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -38,7 +38,6 @@ class VariableFetchAnalyzer
         '$_SESSION',
         '$_REQUEST',
         '$_ENV',
-        '$http_response_header',
```
weirdan, line 41: This is used in other places as well (via `VariableFetchAnalyzer::isSuperGlobal()`) so it's safer to keep it on this list and just override the type, like you did below.

weirdan (changes requested):

weirdan (approved):

weirdan:
Thanks!

PR #6473, opened 2021-09-11
assert both sides of an equality

orklah:
before this PR, 
`if($a === $b)` 
was making sure $a would be restricted with the type of $b but not the opposite.

This is not a big issue with simple types but with positive-int, int-ranges or more complex unions, it can become an issue. This PR makes sure the opposite assertion is made

I hit a bunch of walls and some weird pieces of code while doing this so the result is not optimal.
For example Union::getAssertionString() when applied to Unions with more than one Atomic just returns Atomic::getAssertionString() on the first Atomic from the Union. This has really weird effects because `stdClass|null` returns a `null` assertion that breaks things. Returning an empty string also breaks code so I had to resort to limiting this behaviour to Union with only one atomic

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -3354,11 +3354,23 @@ class AssertionFinder
                 $source
             );
 
+            $other_var_name = ExpressionIdentifier::getArrayVarId(
+                $conditional->right,
+                $this_class_name,
+                $source
+            );
+
             $other_type = $source->node_data->getType($conditional->left);
             $var_type = $source->node_data->getType($conditional->right);
         } elseif ($typed_value_position === self::ASSIGNMENT_TO_LEFT) {
             $var_name = ExpressionIdentifier::getArrayVarId(
                 $conditional->right,
+                $this_class_name,
+                $source
+            );
+
+            $other_var_name = ExpressionIdentifier::getArrayVarId(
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -3279,6 +3291,14 @@ private static function getTypedValueEqualityAssertions(
             } else {
                 $if_types[$var_name] = [['~' . $var_type->getAssertionString()]];
             }
+
+            if ($other_var_name && $other_type && count($other_type->getAtomicTypes()) === 1) {
```
weirdan, line 3295: A comment stating why `Union::isSingle()` is not used here could be useful.

weirdan: Note: the following does not work even when `isSingle()` is used.
```php
/** @var 1|2 $a */
$a = 2;

/** @var 2|3 $b */
$b = 2;

if ($a === $b) {
    /** @psalm-trace $a */; // expected: 2, actual: 2
    /** @psalm-trace $b */; // expected: 2, actual: 2|3
}
```

orklah: Yeah, that's exactly why I count atomics instead. When the Union contains multiple literal of the same type, isSingle still returns true but getAssertionString will return the assertion only for the first Atomic. I'm adding the comment

PR #6474, opened 2021-09-11
4> Loop and Integer range

### LINKED ISSUE #6245: Sum of two `0|positive-int` variables result in `int`
Hey there,

when I add two variables of `0|positive-int`, psalm infers `int` rather than `0|positive-int`.
I guess thats an edgecase, as it works with `positive-int`.

https://psalm.dev/r/750408a26f
https://psalm.dev/r/ca3721fbf8

orklah:
### This PR is dependant on #6241. Please don't merge this before merging the other

This PR reintroduce Int range assertions in loops. This is still a work in progress

EDIT: fix #6245

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Block/LoopAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/LoopAnalyzer.php
@@ -2,6 +2,7 @@
 namespace Psalm\Internal\Analyzer\Statements\Block;
 
 use PhpParser;
+use PhpParser\Node\Expr\Variable;
 use Psalm\CodeLocation;
 use Psalm\Config;
 use Psalm\Context;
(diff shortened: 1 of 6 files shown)


orklah:
When merged, this will fix #6245

src/Psalm/Internal/Analyzer/Statements/Block/LoopAnalyzer.php
```diff
@@ -722,4 +716,60 @@ private static function getAssignmentMapDepth(string $first_var_id, array $assig
...
+    /**
+     * @param PhpParser\Node\Expr[] $post_expressions
+     */
+    private static function analyzePostExpressions(
+        StatementsAnalyzer $statements_analyzer,
+        array $post_expressions,
+        Context $context
+    ): bool {
+        foreach ($post_expressions as $post_expression) {
+            $fake_assignment = null;
```
orklah, line 729: This is a bit weird.

I had to replace the expression in the third argument of for() because otherwise Psalm would do arithmetics on the variable and completely break the assertion from the second argument.

orklah:
I'll close this for now. No matter how I try to tackle this, there's obstacles.
To put it simply (ant to have a memo in case I decide to work on that again in the future) here are the issues:
- Psalm throws a LoopInvalidation sometimes: https://psalm.dev/r/d3c87dac43. However, this is a common practice and Psalm does it on its own code sometimes but does not detect it. Trying to improve inference leads to new errors that makes this issue near useless
- Loops are complex, trying to fix something on one side often generates issues elsewhere that are always hard to debug
- positive-int is a blessing and a curse. It's large enough so it hides a lot of issues in Psalm inference but when we try to replace it by integer ranges, they appear
- Loops are missing a whole module for dealing with literal values and operations. Currently, if Psalm sees an operation in a loop, it will analyze it once and be done with it. We would need an entire duplication of ArithmeticOpAnalyzer who understands loops in order to generate ranges accordingly instead of just literals after a simple operation.
- Psalm can only assert variables. This is an issue with this kind of code (already in tests): https://psalm.dev/r/597bd3d8d2. Even if we fix issue above, this will be an issue because here we would need to understand that $i is decremented until 0 to generate [0, 4]. That involves both assertions and operations so I'm not sure how it would go, but unless we can start by asserting $i is at leat [min, 4], there's no hope.

I may revisit if I become more intelligent, at ease with Psalm or if I get help but now I feel this will only lead to weird false positive I'll be unable to fix down the road

PR #6476, opened 2021-09-12
Always truthy conditions

orklah:
A few weeks ago we introduced alwaysTruthy and alwaysFalsy to Unions. Psalm flags correctly falsy conditions, but not truthy ones. This PR fixes that.

It did find a lot of cases in Psalm but except one, this was wrong phpdoc (I think on purpose). getAtomicTypes is declared as non-empty, but when you removeTypes from an Union, it can become empty.


weirdan:
Thanks!

Issue #6395, opened 2021-09-03
[Help wanted] Listing PHP 8.1 features


orklah:
The first release candidate for PHP 8.1 is out now and the feature freeze is past, We may start looking at the new features and ensuring they're implemented in Psalm. Some were already implemented by Matt but not all.

This thread will regroup every new feature to come and what impact it will have on Psalm.

PHP release notes can be found here: https://github.com/php/php-src/blob/PHP-8.1/UPGRADING.

If you want to help, please respond here with the feature name, a small reproducer of the feature(and the impact it must have on Psalm analysis). For example: 

- [x] **array_is_list** type inference: https://psalm.dev/r/fe157fd547
- [x] **array_is_list** impossible type: https://psalm.dev/r/bb9d8800b1
- [x] **readonly** keyword: https://psalm.dev/r/ec599a194c
- [x] **string key array unpacking** support: https://psalm.dev/r/a25339af39

### Callmap/stub changes

- [x] **array_is_list** function : https://psalm.dev/r/8bef57c13a -> fixed: #6398
- [ ] #6401
- [x] #6402
- [x] #6403
- [x] #6404
- [x] #6405
- [x] #6406
- [x] #6407
- [x] #6408
- [x] #6414
- [x] #6415
- [x] #6416
- [x] #6417
- [x] #6418
- [x] #6422
- [x] #6731

### Parser changes
- [x] #6410
- [x] #6412
- [x] #6413
- [x] #7668

### Enum support
- [x] #6425

I'll update this post with new items as well as progress

niconoe-:
I'm thinking about managing `readonly` in promoted constructors: https://psalm.dev/r/ec599a194c

I'm using [this blog](https://stitcher.io/blog/new-in-php-81) to have a list of new features and deprecated stuff in PHP 8.1. You could probably extract a list from this, as it regroups a lot.

orklah:
@niconoe- thanks, I added your case. Yeah, I know where to find the new features, but creating every repro, thinking of impact on Psalm would take me a long time. It's a good opportunity to make the community participate, even when they lack technical insight of Psalm.

Ideally, I'll then create issues for the simpler cases with an explanation how to do it and encourage people to try and implement the feature

simPod:
Array unpacking with string keys should be supported and new array type inferred

https://psalm.dev/r/a25339af39

orklah:
Thanks @simPod!

weirdan:
[...]

PR #6485, opened 2021-09-14
assert that some type is a list

orklah:
will fix an item of #6395

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -59,7 +59,7 @@ class AssertionFinder
         'is_bool' => ['bool', [Type::class, 'getBool']],
         'is_resource' => ['resource', [Type::class, 'getResource']],
         'is_object' => ['object', [Type::class, 'getObject']],
-        'is_array_list' => ['list', [Type::class, 'getList']],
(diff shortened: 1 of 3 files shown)


weirdan (commented):
I wonder whether a stub would not suffice here. E.g.
```php
/**
 * @psalm-assert-if-true list<mixed> $array
 */
function array_is_list(array $array): bool {}
```

orklah:
Possibly, but string, bool, array, object are all handled here. I think having it here is less confusing.

We could do another PR if we want to migrate everything to stubs though... Or if you want I can migrate all of them in this PR

weirdan:
Consistency is good, so it should be ok the way it is. Just curious, how hard would it be to use FQNs instead of local names? Because right now Psalm doesn't understand namespace-local overrides of global predicates: https://psalm.dev/r/8d7345506e c.f. https://3v4l.org/4Pko7

orklah:
Oh, this is kinda bad (I'd not feel bad about the fool who did override those with another meaning though!). I wonder if this is fixable in assertions, I'll check

orklah:
There's probably a better way to check whether we're using the base function, but I couldn't find it!

orklah:
I'll have to check why CI is failing, I'd not expected that

orklah:
the CI was failing because I didn't handle the namespace fallback. It's now OK

PR #6486, opened 2021-09-14
Update return type for xdebug_get_profiler_filename()

### LINKED ISSUE #6484: xdebug_get_profiler_filename return type
Wasn't sure if this is the sort of thing that should be overridden in the xdebug stub or if it can be pulled from elsewhere...

It looks like psalm assumes a return type of string for xdebug_get_profiler_filename(), but it should be string|false (https://xdebug.org/docs/profiler#xdebug_get_profiler_filename).

https://psalm.dev/r/5ef2b9cf17

ssnepenthe:
Best I can tell this function was introduced in xdebug 2.0 and has had a return type of string|false from the beginning.

Fixes #6484


weirdan (approved):

weirdan:
Thanks!

PR #6494, opened 2021-09-16
Fix wrong template type in `Sequence::copy()`

### LINKED ISSUE #6490: Ds\Sequence<int, mixed> has too many template params, expecting 1
https://psalm.dev/r/bee32f879c

Working with DS Sequence gives 

> Ds\Sequence<int, mixed> has too many template params, expecting 1

I looked into a stub if I see a way how to fix but no idea why this is happening. The definition looks correct.

simPod:
Fixes #6490


weirdan (approved):

weirdan:
Thanks!

Issue #6499, opened 2021-09-17
Psalm don't determine type of object after `instanceof` with `class-string<CLASS>`


vjik:
https://psalm.dev/r/7e3853aa3c

orklah:
simpler reproducer:
https://psalm.dev/r/e11cdd8900

PR #6503, opened 2021-09-17
assert object type from instance of with TClassString

orklah:
this fixes https://github.com/vimeo/psalm/issues/6499


weirdan:
Thanks!

PR #6508, opened 2021-09-20
fixed Uncaught Error: Call to undefined function shell_exec() 

thomasbley:
When php.ini has:
```
disable_functions = "shell_exec,proc_open,passthru,highlight_file,system,popen,exec"
```

I get an error when starting the language server:

```
[ERROR - 10:18:08 AM] Uncaught Error: Call to undefined function shell_exec() in phar:///var/www/src_micro/bin/psalm_4100.phar/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php:943
Stack trace:
#0 phar:///var/www/src_micro/bin/psalm_4100.phar/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(310): Psalm\Internal\Analyzer\ProjectAnalyzer::getCpuCount()
#1 phar:///var/www/src_micro/bin/psalm_4100.phar/src/Psalm/Internal/Cli/LanguageServer.php(213): Psalm\Internal\Analyzer\ProjectAnalyzer->server()
#2 phar:///var/www/src_micro/bin/psalm_4100.phar/src/Psalm/Internal/Cli/Psalm.php(605): Psalm\Internal\Cli\LanguageServer::run()
#3 phar:///var/www/src_micro/bin/psalm_4100.phar/src/Psalm/Internal/Cli/Psalm.php(147): Psalm\Internal\Cli\Psalm::forwardCliCall()
#4 phar:///var/www/src_micro/bin/psalm_4100.phar/psalm(6): Psalm\Internal\Cli\Psalm::run()
#5 /var/www/src_micro/bin/psalm_4100.phar(14): require('...')
#6 {main}
(Psalm 4.x-dev@ crashed due to an uncaught Throwable)
```

Psalm 4.10.0, PHP 8.0.10


orklah:
Does it works after your patch?

I don't know much about retrieving processor numbers, but the only chance left after those shell_exec calls is reading "/proc/cpuinfo". If this fails, Psalm will crash again. Maybe we should make sure to return "1" when shell_exec is disabled?

thomasbley:
I think the problem is that /proc/cpuinfo might not be correct inside containers, e.g.

```
docker run --cpuset-cpus="1" --rm -it debian:buster-slim
root@157a74201501:/# nproc
1
root@157a74201501:/# cat /proc/cpuinfo | grep processor | wc -l
4
```

thomasbley:
maybe cpuset is more accurate, but it gives a range:

```
docker run --cpuset-cpus="1" --rm -it debian:buster-slim
root@157a74201501:/# nproc
1
root@157a74201501:/# cat /proc/cpuinfo | grep processor | wc -l
4
root@157a74201501:/# cat /sys/fs/cgroup/cpuset/cpuset.cpus
1

docker run --rm -it debian:buster-slim
root@173d1fb9c62d:/# cat /sys/fs/cgroup/cpuset/cpuset.cpus
0-3

docker run --cpuset-cpus="0" --rm -it debian:buster-slim
root@9c47555971d3:/# cat /sys/fs/cgroup/cpuset/cpuset.cpus
0
```

thomasbley:
> Maybe we should make sure to return "1" when shell_exec is disabled?

Yes, I think it's the best option.

orklah:
Thanks!

PR #6511, opened 2021-09-20
Update to Phpstorm support

ScottAwesome:
I've added clarifying information that Phpstorm supports psalm out of the box since 2020.3 release, and removed the LSP information.

### DIFF
--- a/docs/running_psalm/language_server.md
+++ b/docs/running_psalm/language_server.md
@@ -29,21 +29,7 @@ This is the config I used:
 
 ## PhpStorm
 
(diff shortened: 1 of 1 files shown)


orklah:
Please keep both options available, IIRC, LSP can do more things than the standard plugin for phpstorm (things like autocomplete).

ScottAwesome:
I re-added the LSP, I set them under separate sub headers so its easy to see the distinction between native phpstorm support and using the LSP.

I did a minor verbiage cleanup too

orklah:
Thanks!

PR #6514, opened 2021-09-21
Add if-this-is

danog:
As per https://github.com/vimeo/psalm/pull/3782#issuecomment-923974334


orklah:
Thanks!

olleharstedt:
Damn, this now works? So it's possible to create a type-safe builder pattern?

olleharstedt:
Next step to make this type-safe is to consider ownership, or non-aliasing.

PR #6518, opened 2021-09-22
improvements of alwaysTruthy/alwaysFalsy

orklah:
this fixes an error I introduced in a previous PR. The continue was not enough to exit both loops


orklah:
I added still more features in this PR, Psalm was unable to detect those `if(rand(0,1) || true){}` as redundant. It's now the case

PR #6519, opened 2021-09-22
Add support for advanced IssueHandler in plugins

gmessier:
Currently a plugin can only define a basic IssueHandler with the Config::setCustomErrorLevel() method.
This pull request add the new Config::setAdvancedErrorLevel() method to be able to define any kind of IssueHandler.

To be able to implement this feature, this pull request also add methods to create FileFilters from an array instead of an XMLElement. The current configuration is highly coupled with XML.  In the future is may be a good idea to reduce this coupling.

Usage example:

```php
Config::getInstance()->setAdvancedErrorLevel('PossiblyUndefinedMethod', [[
            'type' => 'suppress',
            'directory' => [['name' => 'tests']]
        ]]);
```

Tested manually in a project.
I wanted to add some unit tests, but it looks like no test really exist for the `setCustomErrorLevel` method and all `FileFilter` classes. As I'm not familiar with type of test used in Psalm, so I didn't add anything. Don't hesitate if you have suggestions about this.

### DIFF
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -125,7 +125,6 @@
                 <referencedMethod name="Psalm\Codebase::createClassLikeStorage"/>
                 <referencedMethod name="Psalm\Codebase::isVariadic"/>
                 <referencedMethod name="Psalm\Codebase::getMethodReturnsByRef"/>
-                <referencedMethod name="Psalm\Config::setAdvancedErrorLevel"/>
             </errorLevel>
         </PossiblyUnusedMethod>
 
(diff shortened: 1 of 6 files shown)


orklah:
About tests, look for testIssueHandlerWithCustomErrorLevels in ConfigTest. It start from an xml to then check the levels, I guess in your case you'll start with an almost empty xml and then use `setAdvancedErrorLevel`

gmessier:
Seems that the pipeline start to return a new error from the Psalm check:
"Error: The @var array<string, array<int, string>> annotation for $psr4_prefixes is unnecessary (see https://psalm.dev/212)"

It was not there yesterday, it's not related to the code of this PR and another PR has the same error.

orklah:
Build error was fixed in #6526, you can rebase

orklah:
Thanks!

PR #6521, opened 2021-09-22
TCallableString is non-falsy (0 can't be a functionlike name)

orklah:


orklah:
This has the potential of being a BC break, though a very minor one...

weirdan:
I'd postpone it to 5.0 then. I'm curious where one would encounter this?

orklah:
I thought about that when working in Union::isAlwaysTruthy/Union::isAlwaysFalsy. I was thinking about if i needed to include TCallableString to one or the other.

And now, thinking about it, I realize that I made a mistake. TNonEmptyString is not always Truthy (because of '0'), but TNonFalsyString is (not always Truthy).

PR #6529, opened 2021-09-23
Reconcile `class-constant`

### LINKED ISSUE #6300: Asserts with constant wildcards

### LINKED ISSUE #6480: Assert does not provide constant information
Hey there,

I've already commented https://github.com/vimeo/psalm/issues/4617 but the author has problems with properties.

In my case, constants are also affected. I will work on this somewhen in the next days. 

https://psalm.dev/r/88e8252ec3

boesing:
Fixes #6480
Fixes #6300

----

Any help here would be appreciated.

### DIFF
--- a/src/Psalm/Internal/Codebase/ClassConstantByWildcardResolver.php
+++ b/src/Psalm/Internal/Codebase/ClassConstantByWildcardResolver.php
@@ -55,6 +55,6 @@ final class ClassConstantByWildcardResolver
             $matched_class_constant_types[] = $class_constant_storage->type->getAtomicTypes();
         }
 
-        return array_values(array_merge(...$matched_class_constant_types));
+        return array_values(array_merge([], ...$matched_class_constant_types));
     }
 }
--- a/src/Psalm/Internal/Codebase/ClassConstantByWildcardResolver.php
+++ b/src/Psalm/Internal/Codebase/ClassConstantByWildcardResolver.php
@@ -1,0 +1,60 @@
+<?php
+declare(strict_types=1);
+
+namespace Psalm\Internal\Codebase;
+
+use Psalm\Codebase;
+use Psalm\Type;
+use Psalm\Type\Atomic\TMixed;
+
+use function array_merge;
+use function array_values;
+use function preg_match;
+use function sprintf;
+use function str_replace;
+
+/**
+ * @internal
+ */
+final class ClassConstantByWildcardResolver
+{
+    /**
+     * @var \Psalm\Codebase
+     */
+    private $codebase;
+
+    public function __construct(Codebase $codebase)
+    {
+        $this->codebase = $codebase;
+    }
+
+    /**
+     * @return list<Type\Atomic>|null
+     */
+    public function resolve(string $class_name, string $constant_pattern): ?array
+    {
+        if (!$this->codebase->classlike_storage_provider->has($class_name)) {
+            return null;
+        }
+
+        $constant_regex_pattern = sprintf('#^%s$#', str_replace('*', '.*', $constant_pattern));
+
+        $class_like_storage = $this->codebase->classlike_storage_provider->get($class_name);
+        $matched_class_constant_types = [];
+
+        foreach ($class_like_storage->constants as $constant => $class_constant_storage) {
+            if (preg_match($constant_regex_pattern, $constant) === 0) {
+                continue;
+            }
+
+            if (! $class_constant_storage->type) {
+                $matched_class_constant_types[] = [new TMixed()];
+                continue;
+            }
+
+            $matched_class_constant_types[] = $class_constant_storage->type->getAtomicTypes();
+        }
+
+        return array_values(array_merge(...$matched_class_constant_types));
+    }
+}
--- a/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
+++ b/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
@@ -40,6 +40,7 @@ use Psalm\Type\Atomic\TResource;
 use Psalm\Type\Atomic\TScalar;
 use Psalm\Type\Atomic\TString;
 use Psalm\Type\Atomic\TTemplateParam;
+use Psalm\Type\Reconciler;
 use Psalm\Type\Union;
 
 use function assert;
@@ -72,7 +73,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
         bool $negated = false,
         ?CodeLocation $code_location = null,
         array $suppressed_issues = [],
-        int &$failed_reconciliation = 0,
+        int &$failed_reconciliation = Reconciler::RECONCILIATION_OK,
         bool $is_equality = false,
         bool $is_strict_equality = false,
         bool $inside_loop = false
@@ -496,7 +497,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             );
 
             if (empty($existing_var_type->getAtomicTypes())) {
-                $failed_reconciliation = 2;
+                $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
                 return Type::getEmpty();
             }
         }
@@ -710,7 +711,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($positive_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return Type::getEmpty();
     }
@@ -815,7 +816,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($object_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return Type::getMixed();
     }
@@ -907,7 +908,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($string_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return $existing_var_type->from_docblock
             ? Type::getMixed()
@@ -1005,7 +1006,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($int_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return $existing_var_type->from_docblock
             ? Type::getMixed()
@@ -1084,7 +1085,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($bool_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return $existing_var_type->from_docblock
             ? Type::getMixed()
@@ -1159,7 +1160,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($scalar_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return $existing_var_type->from_docblock
             ? Type::getMixed()
@@ -1251,7 +1252,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($numeric_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return $existing_var_type->from_docblock
             ? Type::getMixed()
@@ -1344,7 +1345,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($object_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return $existing_var_type->from_docblock
             ? Type::getMixed()
@@ -1401,7 +1402,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($resource_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return $existing_var_type->from_docblock
             ? Type::getMixed()
@@ -1472,7 +1473,7 @@ class SimpleAssertionReconciler extends \Psalm\Type\Reconciler
             return new Type\Union($iterable_types);
         }
 
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return Type::getMixed();
     }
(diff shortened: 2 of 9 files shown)


tests/TypeReconciliation/ReconcilerTest.php
```diff
@@ -177,4 +176,61 @@ public function providerTestTypeIsContainedBy(): array
             ],
         ];
     }
+
+    /**
+     * @dataProvider constantAssertions
+     */
+    public function testReconciliationOfClassConstantInAssertions(string $assertion, string $expected_type): void
```
boesing, line 183: Is there a more solid way to test this in a more "integrated" way?
This wont ensure that `class-constant(Foo::*)` will be passed to the `AssertionReconciler#reconcile` and thus it may break when something regarding this will change in a later version.

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2507,4 +2517,50 @@ function (Type\Union $t): bool {
...
+    /**
+     * @param   0|1|2    $failed_reconciliation
+     */
+    private static function reconcileClassConstant(
+        Codebase $codebase,
+        string $class_constant_expression,
+        int &$failed_reconciliation
+    ) : Union {
+        if (strpos($class_constant_expression, '::') === false) {
+            $failed_reconciliation = 2;
```
boesing, line 2530: TBH: I have no idea for what this is for

If someone gives some clarification, some `Reconciler::RECONCILIATION_*` constants might make sense. Would be happy to implement these.

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2507,4 +2517,50 @@ function (Type\Union $t): bool {
...
+            }
+
+            $types = $class_constant_storage->type->getAtomicTypes();
+            $type = reset($types);
+            $matched_class_constant_types[] = $type;
+        }
+
+        if ($matched_class_constant_types === []) {
+            $failed_reconciliation = 2;
+            return Type::getMixed();
```
boesing, line 2561: This should only happen when a constant is passed which does not exist

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
orklah: If I'm not mistaken, 0 means no issue, 1 means the reconciliation did not remove any existing type (therefore, it was redundant, like checking is_string when we already have a TString), 2 means the resulting type is empty so there's no possible issue (like checking is_string when we have a TInt)

tests/TypeReconciliation/ReconcilerTest.php
orklah: Not sure I understood what you meant there

tests/TypeReconciliation/ReconcilerTest.php
```diff
@@ -177,4 +176,61 @@ public function providerTestTypeIsContainedBy(): array
...
+    {
+        $this->addFile(
+            'psalm-assert.php',
+            '
+            <?php
+            class Foo
```
orklah, line 189: Can you add a test in a namespace please?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2507,4 +2517,50 @@ function (Type\Union $t): bool {
...
+        if (!$codebase->classlike_storage_provider->has($class_name)) {
+            $failed_reconciliation = 2;
+            return Type::getMixed();
+        }
+
+        $class_like_storage = $codebase->classlike_storage_provider->get($class_name);
+        $matched_class_constant_types = [];
+
+        foreach ($class_like_storage->constants as $constant => $class_constant_storage) {
+            if (!fnmatch($constant_pattern, $constant)) {
```
orklah, line 2545: I'm a little uneasy using a function made for filenames in this context. I think we should already have a similar feature elsewhere without using that

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2507,4 +2517,50 @@ function (Type\Union $t): bool {
...
+                continue;
+            }
+
+            if (! $class_constant_storage->type || !$class_constant_storage->type->isSingle()) {
```
orklah, line 2549: Is there a reason to limit to Single types? You seem to add types in an array and to a big Union later so it should not be an issue to add every AtomicTypes in the array?

tests/TypeReconciliation/ReconcilerTest.php
boesing: What do you mean with "namespace"?
So rather than adding a file with `class Foo`, better add `namespace ReconciliationTest;` or sth. like this in there?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: Is it possible to have a constant of multiple types?
I've thought that a constant can only contain one type and thus I will enforce that here.

tests/TypeReconciliation/ReconcilerTest.php
boesing: In this test, I have to assume that the caller of the class-function has already made its job.
Given the fact that wildcard constants worked in a previous version (I've used that in a project), I think there was no proper test. I think the unit tests are still passing even tho the functionality is gone.

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: Added constants for these and changed code

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: other parts remove the wildcard and then use `strpos`.
Should I do so as well?

```php
$const_name_part = \substr($return_type->const_name, 0, -1);

if ($const_name_part) {
    $matching_constants = \array_filter(
        $matching_constants,
        function ($constant_name) use ($const_name_part): bool {
            return $constant_name !== $const_name_part
                && \strpos($constant_name, $const_name_part) === 0;
        }
    );
}
```

tests/TypeReconciliation/ReconcilerTest.php
orklah: what I meant is that your Foo class is in global namespace, when the class is in a namespace, some part of the code works differently, so yeah, please add a test with a `namespace test;` so that we can see everything works

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
orklah: that doesn't seem a lot better to me. Maybe a small function replacing * with .* in a regex would be better?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
orklah: something like [that](https://3v4l.org/AVdtA) could very well be inferred in Psalm with `float|int`

tests/TypeReconciliation/ReconcilerTest.php
boesing: Okay so I understood you right 👍🏼

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: I've changed the code accordingly

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: done

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2501,10 +2516,55 @@ function (Type\Union $t): bool {
...
-        $failed_reconciliation = 2;
+        $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
         return $assertion === 'empty' && $existing_var_type->possibly_undefined
             ? Type::getEmpty()
             : Type::getMixed();
     }
+
+    /**
+     * @param   0|1|2    $failed_reconciliation
```
orklah, line 2527: while we're on it, care to replace those phpdoc with Reconciler::RECONCILIATION_* everywhere?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2501,10 +2516,55 @@ function (Type\Union $t): bool {
...
+            return $existing_type;
+        }
+
+        [$class_name, $constant_pattern] = explode('::', $class_constant_expression, 2);
+
+        if (!$codebase->classlike_storage_provider->has($class_name)) {
+            return $existing_type;
+        }
+
+        $constant_regex_pattern = sprintf('#^%s$#', str_replace('*', '.*', $constant_pattern));
```
orklah, line 2545: I think a preg_quote could be added somewhere. There's little chance that the provided pattern has a # on it, but I'm not sure what it would do if there is

orklah: Also, could you extract this feature so it could be applied on other usages of constant resolution elsewhere?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: I tried that but psalm failed when doing so. Thats why I've removed it again as I am quite busy ATM and don't want to spent more time on this than necessary.

There are plenty of issues with referencing class-constants using wildcards but I cannot fix them all.

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: Any suggestions on how to name this feature? I really suck when it comes to namings.

orklah: transformPatternToRegex or something like that?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
orklah: No worries then, I'll do it later ;)

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: Oh I see, you don't want to extract the whole `type detection` feature abstracted but only the pattern replacement?

orklah: yeah, it's just to be able to use it where we already resolve the constant but with a simpler method

boesing: Using `preg_quote` will escape wildcards - I could unescape wildcards afterwards but TBH:
as you say - not sure if thats really necessary and its more complicated to unescape than actually preventing an invalid regex which then simply won't match. I'd say, lets tkae the pill. WDYT @orklah?

boesing: I've extracted the class-constant type detection into a dedicated, internal, `ClassConstantByWildcardResolver`

orklah: No worries for preg_quote, we can always go incrementally if we realize later that it's an issue :)

tests/Internal/Codebase/ClassConstantByWildcardResolverTest.php
```diff
@@ -0,0 +1,83 @@
...
+        foreach ($resolved as $type) {
+            self::assertInstanceOf(TLiteralString::class, $type);
+            self::assertTrue($type->value === 'bar' || $type->value === 'baz');
+        }
+
+        $resolved = $this->resolver->resolve('ReconciliationTest\\Foo', 'QOO');
+        self::assertNotNull($resolved);
+        /** @var list<\Psalm\Type\Atomic> */
+        self::assertCount(1, $resolved);
+        /** @var non-empty-list<\Psalm\Type\Atomic> $type */
```
orklah, line 74: Why do you set $type as an array here, just to assert that it's an instance of TLiteralString just below?

boesing: because the "shepherd" build pipeline step yelled at me that accessing an undefined array key would be problematic...

tests/AssertAnnotationTest.php
```diff
@@ -1241,7 +1241,7 @@ function test($a, $b): string {
                         return substr($a, 0, 1) . substr($b, 0, 1);
                     }'
             ],
-            'SKIPPED-convertConstStringType' => [
+            'convertConstStringType' => [
```
boesing, line 1244: I've re-enabled this test case as it is fixed with this PR as well.
Thanks to @orklah for referring to it in #6300

tests/Internal/Codebase/ClassConstantByWildcardResolverTest.php
orklah: Here are ways you should be able to solve this correctly:
```
/** @var non-empty-list<\Psalm\Type\Atomic> $resolved */
$type = $resolved[0];
self::assertInstanceOf(TLiteralString::class, $type);
```
or
```
/** @var \Psalm\Type\Atomic $type */
$type = $resolved[0];
self::assertInstanceOf(TLiteralString::class, $type);
```

in any case, $type can't be an array AND then a class-string, and you shouldn't need to suppress any DocblockTypeContradiction

boesing: Yah nvm, dunno why - I've removed the whole annotations and now it works.
I guess there was some errors due to the `var` annotations.

src/Psalm/Internal/Codebase/ClassConstantByWildcardResolver.php
```diff
@@ -0,0 +1,60 @@
...
+
+            if (! $class_constant_storage->type) {
+                $matched_class_constant_types[] = [new TMixed()];
+                continue;
+            }
+
+            $matched_class_constant_types[] = $class_constant_storage->type->getAtomicTypes();
+        }
+
+        return array_values(array_merge(...$matched_class_constant_types));
```
orklah, line 58: I'm surprised Psalm didn't pick this up. array_merge until 7.4 needed at least a parameter in array_merge. This meant you had to be absolutely sure your array was not empty to be able to use `array_merge(...$array)`. I don't know if this code can end up with an empty $matched_class_constant_types, but even if it could, I don't see how Psalm could know it.

Can you use `array_merge([], ...$matched_class_constant_types)` or confirm that this array can never be empty?

tests/TypeReconciliation/ReconcilerTest.php
```diff
@@ -184,4 +183,70 @@ public function providerTestTypeIsContainedBy(): array
...
+            {
+                const PREFIX_BAR = \'bar\';
+                const PREFIX_BAZ = \'baz\';
+                const PREFIX_QOO = Foo::PREFIX_BAR;
+            }
+            '
+        );
+        $this->project_analyzer->getCodebase()->scanFiles();
+
+        $reconciled = \Psalm\Internal\Type\AssertionReconciler::reconcile(
```
orklah, line 207: If you're trying to transform something like `"foo"|"bar"` into an Union, I think `Type;;parseString` would be more appropriate here

boesing: no I dont want to parse a string. I want to parse the constants and extract their types. Not sure how parseString could be of any help here? Maybe I miss something?

orklah: nope, my mistake, I thought you were passing $expected_type in there.

orklah (approved):

weirdan:
Thanks!

PR #6532, opened 2021-09-25
code trimming

orklah:
mainly small improvements, code readability and small perfs related changes

### DIFF
--- a/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
+++ b/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
@@ -229,7 +229,7 @@ class FunctionDocblockManipulator
                 continue;
             }
 
-            if ($chars[$i] === '\\' || preg_match('/\w/', $char)) {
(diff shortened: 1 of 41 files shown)


src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
```diff
@@ -229,7 +229,7 @@ private function __construct(string $file_path, FunctionLike $stmt, ProjectAnaly
                 continue;
             }
 
-            if ($chars[$i] === '\\' || preg_match('/\w/', $char)) {
+            if ($char === '\\' || preg_match('/\w/', $char)) {
```
weirdan, line 232: @orklah are you sure it's the same as `$chars[$i]`? It looks like `$i` is changed above, after `$char` assignment.

orklah: I think so, every change to $i is followed by a continue; but I agree it's confusing, I'll revert that

PR #6552, opened 2021-09-29
offset the starting position of the function in case of attributes

### LINKED ISSUE #6542: MissingReturnType fix issue
Hi,

I encountered the following issue.
I hope the two snippets make the bug obvious enough and easy to reproduce.

Used version: 4.10.0

**Normal behavior**

```
$ docker compose exec php vendor/bin/psalm --no-cache --alter --issues=MissingReturnType --dry-run src/test.php
Scanning files...
Analyzing files...

░
Altering files...
/srv/app/src/test.php:
--- /srv/app/src/test.php
+++ /srv/app/src/test.php
@@ -3,7 +3,7 @@
 class A
 {
     #[Foo]
-    public function bar()
+    public function bar(): void
     {
     }
 }
```

**Problematic behavior when parenthesis are added on the attribute**

```
$ docker compose exec php vendor/bin/psalm --no-cache --alter --issues=MissingReturnType --dry-run src/test.php
Scanning files...
Analyzing files...

░
Altering files...
/srv/app/src/test.php:
--- /srv/app/src/test.php
+++ /srv/app/src/test.php
@@ -3,7 +3,7 @@
 class A
 {
     #[Foo()]
-    public function bar()
+    void function bar()
     {
     }
 }

```

I'm new with psalm and not yet familiar with the way it works, that's why there's no PR associated.
Just reporting for this time. Hope it's ok :)

orklah:
This fixes #6542

Somehow, PHP-Parser considers that `$stmt->getAttribute('startFilePos')` when a function has an attribute is at the beggining of the attribute.

The method Psalm uses for locating the place where it can add a type in signature is by looking after the parameters when there are some and then check for the first closing parenthesis. If there's no parameter, it looks at the first closing parenthesis from the start of the function.

And this is how it breaks. When there is no parameter and an attribute, the first parenthesis can be inside the attribute. This makes Psalm confused as to where it can add the return types and it derps.

My fix for that was to check for attributes and then consider that the function starts at the end of the last attribute.

There may be more elegant ways to do that but I'm not familiar enough with PHP-Parser API

### DIFF
--- a/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
+++ b/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
@@ -123,6 +123,14 @@ class FunctionDocblockManipulator
         $this->docblock_end = $function_start = (int)$stmt->getAttribute('startFilePos');
         $function_end = (int)$stmt->getAttribute('endFilePos');
 
+        $attributes = $stmt->getAttrGroups();
+        foreach ($attributes as $attribute) {
+            // if we have attribute groups, we need to consider that the function starts after them
+            if ((int) $attribute->getAttribute('endFilePos') > $function_start) {
+                $function_start = (int) $attribute->getAttribute('endFilePos');
+            }
+        }
+
         foreach ($stmt->params as $param) {
             if ($param->var instanceof PhpParser\Node\Expr\Variable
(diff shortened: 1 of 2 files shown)


weirdan (approved):

weirdan:
Thanks!

PR #6553, opened 2021-09-29
Callmap leak

### LINKED ISSUE #4748: $matches sporadic error message PossiblyUndefinedVariable
sometimes we've issues with error messages like:

```
OK ERROR: PossiblyUndefinedVariable - xxxx.php:49:21 - Possibly undefined variable $matches defined in try block (see https://psalm.dev/018)
OK         $x = (int)$matches[1];
```

the code looks similar like this:
https://psalm.dev/r/46e568f7f8

i could not reproduce this issue in psalm.dev - complaining about "PossiblyUndefinedVariable" in this case seems to be sporadic. Some of my co-workers sometimes have the issue.

### LINKED ISSUE #6543: try...finally in one function affecting a completely separate function
On it's own, Psalm passes this function with no issues:

```php
<?php
function two(): mixed
{
    $r = exec('ls', $lines);
    if ($r !== false) {
        return $lines;
    }
    return false;
}
```
(as seen in https://psalm.dev/r/ae1083cb24)

However, if I add another function before it that calls exec() within a try..finally, then it fails:

```php
<?php
function one(string $originalPdf): ?string
{
    try {
        exec('pdfgrep "FOO" -n ' . $originalPdf, $output, $retval);
        if ($retval === 0 && isset($output[0])) {
            return (string)$output[0];
        }
    } finally {
        unlink($originalPdf);
    }

    return null;
}

function two(): mixed
{
    $r = exec('ls', $lines);
    if ($r !== false) {
        return $lines;
    }
    return false;
}
```
(as seen in https://psalm.dev/r/ec585fea79)

Error is:

> INFO: PossiblyUndefinedVariable - 20:16 - Possibly undefined variable $lines defined in try block
 
Note that the issue relates to function `two()` but  the `try` block is in function `one()`.

If I comment out the try block in `one()` - i.e. lines 4 (`try {`), 9 (`} finally {`) and 11 (`}`), then it works as per https://psalm.dev/r/7144ceac33.

From what I can tell the `exec()` call within the try call must use `exec()`'s $output parameter for this issue to occur.

In my actual project the real two functions are in separate methods in separate classes in separate files. 

What am I doing wrong? I didn't expect that wrapping `exec()` in a try...finally in one function to affect a completely different function that also uses `exec()`.

orklah:
This fixes a leak where types from the callmap would be provided to further analysis, if the analysis needed to change the type or flags (in this case, "possibly_undefined_from_try", the callmap itself would change and start providing this new type on each new use.

This fixes #6543
This will most probably fix #4748

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
@@ -969,8 +969,12 @@ class ArgumentsAnalyzer
                     $function_param = $last_param;
                 }
 
-                $by_ref_type = clone $function_param->type;
-                $by_ref_out_type = clone $function_param->out_type;
+                if ($function_param->type) {
+                    $by_ref_type = clone $function_param->type;
+                }
+                if ($function_param->out_type) {
+                    $by_ref_out_type = clone $function_param->out_type;
+                }
 
                 if ($by_ref_type && $by_ref_type->isNullable()) {
                     $check_null_ref = false;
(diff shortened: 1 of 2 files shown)


weirdan (approved):

weirdan:
Thanks!

Issue #6402, opened 2021-09-04
Fileinfo resources have been converted to objects


weirdan:
>  The fileinfo functions now accept and return, respectively, finfo objects instead of resources.

Guide to editing callmaps: [Editing callmaps](https://psalm.dev/docs/contributing/editing_callmaps/).
Example of conversion: https://github.com/vimeo/psalm/pull/6421

tomasnorre:
I would be happy to try doing this. 

How can I verify that I'm done and have done it correctly?

orklah:
There are tests that checks the integrity of the callmap itself so don't be afraid to break things. 
We usually don't add tests for to check for the callmap changes but you totally can if you want. Just add an entry in FunctionCallTest::providerValidCodeParse. You can look for examples where a specific PHP version is set, so when the version is <8.1, fileinfo functions should return a resource (check at the `assertions` array), but for 8.1, they should return objects.

tomasnorre:
Thanks, will give it a spin.

tomasnorre:
Reading the diff, for the example with FTP Functions that is referred. 

I don't quite get why the `Callmap.php` have a change like this. 
```diff
- 'ftp_alloc' => ['bool', 'ftp'=>'resource', 'size'=>'int', '&w_response='=>'string'],
+ 'ftp_alloc' => ['bool', 'ftp'=>'FTP\Connection', 'size'=>'int', '&w_response='=>'string'],
```
I would have expected this only to be in the `CallMap_81_delta.php`. 


as I read the https://github.com/vimeo/psalm/blob/master/docs/contributing/editing_callmaps.md#delta-file-format changes should be done in the respective delta version of the callmap  e.g. (`CallMap_81_delta.php`)and not in the `CallMap.php` itself. 

Can you explain this to me?

orklah:
CallMap.php contains up to date changes.

Then we take delta files to incrementally *remove* newer signature by their old ones. Then, once every delta file has been applied, we should be left with the exact content of CallMap_historical

So yeah, changes must be made in the two files

PR #6566, opened 2021-10-02
Convert finfo functions to use object instead of a resource

tomasnorre:
Resolves: #6402

I'm honestly in doubt how the `finfo` should be written. I don't know if it should be namespaced like the `FTP/Connection` or just `finfo`.

I don't get an indication in the  https://github.com/php/php-src/pull/5987/files/31164d2acd0de1375b24f30175b1c3c67cd56d07

But I'm also not used to read the Source itself, and my PHPStorm I cannot switch to PHP 8.1 to have the auto-complition (jump to source) help me.

Pointers would be a big help. At best a reference to where I see it, then I'll learn from it and not just do that change without knowing why.


tomasnorre:
Could you please add the #hacktoberfest badge to the project, and hopefully later the #hacktoberfest-accepted to this one, currently it says you're not participating

orklah:
No need for namespacing as finfo exist in global namespace (see https://3v4l.org/h2NYD#v8.1rc2)

PR seems great :) Thanks and congrats for your first PR!

>Could you please add the #hacktoberfest badge to the project

I think it should be done soon. Meanwhile, I added the label to this specific PR, does it works?

tomasnorre:
> I think it should be done soon. Meanwhile, I added the label to this specific PR, does it works?

Not yet, but might be cached. So will check later. Will take 14 days to be finally accepted anyway. So no stress.

tomasnorre:
> No need for namespacing as finfo exist in global namespace (see https://3v4l.org/h2NYD#v8.1rc2)

Didn't think about checking it like that.

orklah:
Thanks!

tomasnorre:
Np. Just happy to help and learn. Thanks for enabling and supporting me to help.

Issue #6404, opened 2021-09-04
IMAP resources have been converted to objects


weirdan:
> The IMAP functions now accept and return, respectively, `IMAP\Connection` objects instead of resources

Guide to editing callmaps: [Editing callmaps](https://psalm.dev/docs/contributing/editing_callmaps/).
Example of conversion: https://github.com/vimeo/psalm/pull/6421

tomasnorre:
Could you please assign me to this one too? Then I'll give this a spin too.

PR #6567, opened 2021-10-02
Convert imap functions to use object instead of a resource

tomasnorre:
Resolves: #6404


orklah:
Great! Thanks!

orklah:
Thanks!

Issue #6405, opened 2021-09-04
LDAP resources have been converted to objects


weirdan:
> *  The LDAP functions now accept and return, respectively, `LDAP\Connection` objects instead of `ldap link` resources. 
>  * The LDAP functions now accept and return, respectively, `LDAP\Result` objects instead of `ldap result` resources. 
>  * The LDAP functions now accept and return, respectively, `LDAP\ResultEntry` objects instead of `ldap result entry` resources. 

Guide to editing callmaps: [Editing callmaps](https://psalm.dev/docs/contributing/editing_callmaps/).
Example of conversion: https://github.com/vimeo/psalm/pull/6421

tomasnorre:
Please assign me to this one. PR is WIP

PR #6570, opened 2021-10-02
Convert LDAP resources to objects

tomasnorre:
Resolves: #6405 

I'm honestly in doubt regarding the `'link_identifier'=>'resource'` and `'result_identifier'=>'resources'` if they should be `LDAP\Connection` and `LDAP\Result` respectively?


tomasnorre:
The unit tests are failing with cases like this. 

```
1) Psalm\Tests\Internal\CallMapTest::testOutgoingSignaturesMustMatchMostRecentIncomingSignatures
Outgoing signatures in CallMap_81_delta.php don't match corresponding incoming signatures in CallMap_historical.php
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
     'ldap_first_attribute' => Array (
         0 => 'string|false'
         'ldap' => 'resource'
-        'result' => 'resource'
+        'entry' => 'resource'
     )
     'ldap_first_entry' => Array (...)
     'ldap_first_reference' => Array (...)
````

But I cannot see what I have done wrong reading the code nor the error message. Could some one give me a pointer?

orklah:
Check at the arguments names :). You have to keep consistence all the way from the Callmap.php, through each delta file and when all deltas are applied, you should be left with Callmap_historical. If you changed names, then either the names changed in PHP itself and this should be reflected by a change in a delta file, or the name was wrong in Psalm all the way and you have to change the Callmap_historical too. (note that the names were not important before PHP 8.0 so no need to do research about the internal name before that. If it has not changed in PHP 8.1, change the historical file too)

tomasnorre:
Thx will look into that. How do I confirm "problems" like this? As I have no rendered PHP Manual for 8.1 yet.

orklah:
Looking at the PR in php-src, I remembered they have their own stubs now: https://github.com/php/php-src/pull/6963/files#diff-e4f76e193517f1144ea18d65617548c3b98809ad5b575b8cc38b708de00233ac

You should be able to find what you're looking for in here

orklah:
Same here, could you rebase?

orklah:
Thanks!

Issue #6414, opened 2021-09-04
`CURLStringFile`


weirdan:
https://psalm.dev/r/e86560b6e1

Guide to editing callmaps: [Editing callmaps](https://psalm.dev/docs/contributing/editing_callmaps/).

We need stub added in https://github.com/vimeo/psalm/blob/master/stubs/Php81.phpstub

https://github.com/php/php-src/blob/2378f357870c9c24462389ab57e1b211ea2d79a0/ext/curl/curl_file.stub.php#L30-L37

PR #6574, opened 2021-10-03
Add CURLStringFile class

ishanvyas22:
Fixes https://github.com/vimeo/psalm/issues/6414

### DIFF
--- a/dictionaries/CallMap_81_delta.php
+++ b/dictionaries/CallMap_81_delta.php
@@ -23,7 +23,7 @@ return [
     'imagecreatefromavif' => ['false|GdImage', 'filename'=>'string'],
     'mysqli_fetch_column' => ['null|int|float|string|false', 'result'=>'mysqli_result', 'column='=>'int'],
     'mysqli_result::fetch_column' => ['null|int|float|string|false', 'column='=>'int'],
-    'CURLStringFile::__construct' => ['void', 'data'=>'string', 'postname'=>'string', 'mime'=>'string'],
+    'CURLStringFile::__construct' => ['void', 'data'=>'string', 'postname'=>'string', 'mime='=>'string'],
   ],
 
   'changed' => [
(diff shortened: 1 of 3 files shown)


dictionaries/CallMap_81_delta.php
```diff
@@ -23,6 +23,7 @@
     'imagecreatefromavif' => ['false|GdImage', 'filename'=>'string'],
     'mysqli_fetch_column' => ['null|int|float|string|false', 'result'=>'mysqli_result', 'column='=>'int'],
     'mysqli_result::fetch_column' => ['null|int|float|string|false', 'column='=>'int'],
+    'CURLStringFile::__construct' => ['void', 'data'=>'string', 'postname'=>'string', 'mime'=>'string'],
```
orklah, line 26: The CallMap syntax uses an "=" at the end of an argument name to convey that the argument is optional. You should add that on `mime`

stubs/Php81.phpstub
```diff
@@ -6,6 +6,16 @@ namespace {
     }
 }
 
+namespace {
+    class CURLStringFile {
+        public string $data;
+        public string $postname;
+        public string $mime;
+
+        public function __construct(string $data, string $postname, string $mime = "application/octet-stream") {}
```
orklah, line 15: unless you have specific annotations to add, this should not be needed. The __construct signature is redundant with the CallMap and the properties signatures should be detailed in PropertyMap.php

orklah (changes requested):
Great start! Here's some comments. You'll see the CI failing too because there are changes to be made in CallMap.php too. You'll find more details here: https://psalm.dev/docs/contributing/editing_callmaps/

ishanvyas22:
@orklah, made changes according to your suggestion. Do I need to add properties in `PropertyMap.php` file too?

It seems it will not be needed because I don't see any properties added for `CURLFile` class for example.

orklah:
Properties may be resolved through reflection, best way to be sure would be to run Psalm on 8.1 and check, but no need to go this far. If anyone has a need for those and they're not typed, they'll create an issue.

Thanks for the PR! If you're participating in Hacktoberfest, could you confirm this PR is counted? Otherwise I'll add the specific tag, I just want to check if this works without it.

ishanvyas22:
Yes, it would be great if you add the label :). Otherwise it won't be counted.

PR #6584, opened 2021-10-04
Missing property type on public property is dangerous

### LINKED ISSUE #6583: Psalter can't add types to public properties
https://psalm.dev/r/660a502141

This is totally missed by Psalter and creates a bug. Psalter should skip public properties altogether

orklah:
This fixes #6583 . That's a shame because it's pretty useful but the example in the issue is concerning. Psalm doesn't check at all for external access and it's very easy to break something here. It also avoid adding type in phpdoc if it's the same than signature


orklah:
I need to check a few things before merging, I'm converting to draft

PR #6591, opened 2021-10-05
ReflectionClass::getReflectionConstants has a filter parameter

### LINKED ISSUE #6588: ReflectionClass::getReflectionConstants has one parameter in PHP 8.0+
Starting PHP 8.0, `ReflectionClass::getReflectionConstants` has a `filter` parameter:
https://www.php.net/manual/en/reflectionclass.getreflectionconstants.php

However psalm doesn't know about it and complains:
https://psalm.dev/r/6c07857541

While it's valid code:
https://3v4l.org/d0BE1#v8.0.11
https://phpstan.org/r/b552450b-5acb-4b2a-860e-34b89c370af5

vaceletm:
It was introduced in php 8.0.

Fixes #6588


dictionaries/CallMap_80_delta.php
```diff
@@ -97,6 +97,10 @@
       'old' => ['array<string,mixed>'],
       'new' => ['array<string,mixed>', 'filter='=>'?int'],
     ],
+    'ReflectionClass::getReflectionConstants' => [
+        'old' => ['array<string,mixed>'],
+        'new' => ['array<string,mixed>', 'filter='=>'?int'],
```
orklah, line 102: Seems like there is an incoherence between the signature here `array<string,mixed>` and int CallMap.php `list<ReflectionClassConstant>`

vaceletm:
I believe I managed to change the CallMap as expected however when I tried to write a testcase to assert that, the modifications I made doesn't seems to be taken into account. Maybe there is an intermediate step to run ?

orklah:
No that should be it. Can you commit the test you tried to make so we can check that?

orklah:
Can you check how tests are made in FunctionCallTest? Just add a case here. You'll see that you can specify a php version. I think that's what's missing in your current test.

vaceletm:
You were right. I removed the test example from my commit. Do you want me to add a proper test case for the CallMap change ?
(I didn't manage to sort out if it was expected or not for this kind of contributions)

orklah:
> I didn't manage to sort out if it was expected or not for this kind of contributions

It's not :)

Thanks! We'll let the CI finish and this can be merged

PR #6599, opened 2021-10-06
cleanup - Remove not used infection.json.dist

sasezaki:
, and also delete from .gitattributes.
Also delete changelog.md, .travis.yml line.

Because I was confused that "There is infection.json.dist, so does mutation test running on circleci ?" 😭


orklah:
Thanks!

PR #6609, opened 2021-10-07
Feature/strict bool conditions

### LINKED ISSUE #6607: No error on null in bool condition?
I was wondering why I don't get an error when I use a possible null value in an if condition: https://psalm.dev/r/a70306a460

phpstan also has some trouble with this code, but at least I get an error for line 18:

> Only booleans are allowed in an if condition, bool|null given.

https://phpstan.org/r/bd5371d8-c770-4d68-99fb-b82635436796

Did I miss a configuration?

Dgame:
My first PR as an attempt to close #6607 

@orklah I'm sure I can do better, but I had a fair share of trouble getting my head around the code base. 😅


orklah:
That seems pretty spot on to me!

Maybe that's a personal preference but I'm not really a fan of traits. I'd prefer if you added methods in IfElseConditionalAnalyzer (that's the entry point for all the if()elseif()else() construct).

Also, could you create a new Issue instead of reusing the Cast Issue? That'll help people who want to filter that more precisely

orklah:
Let's call it NonStrictBooleanOperand, it should be clear i think.

About your tests, can you check how they're done in FunctionCallTest for example? Once it's done like that, it's easier to add new test cases. Assertions, Errors, PHP version are already built in in the array shape

Dgame:
> Let's call it NonStrictBooleanOperand, it should be clear i think.

I've called it `NonStrictBoolCondition` but I can change it if you want.

> About your tests, can you check how they're done in FunctionCallTest for example? Once it's done like that, it's easier to add new test cases. Assertions, Errors, PHP version are already built in in the array shape

Sure, I can try that.

kylekatarnls (approved):

orklah:
I'll close this, it was never finished and has been open for years

hockdudu:
I just stumbled upon this pull request. This is a feature I'd like to have as well. Do I see it right, only the two issues mentioned on the comment https://github.com/vimeo/psalm/pull/6609#issuecomment-939480412 must be fixed? Well, other than rebasing the code to the most recent version.

orklah:
I can't be sure, I only saw what was wrong with the CI at the time, I didn't review the actual changes

PR #6614, opened 2021-10-08
feat: Add Fiber stubs

### LINKED ISSUE #6422: Fiber stubs
https://wiki.php.net/rfc/fibers

https://github.com/php/php-src/blob/master/Zend/zend_fibers.stub.php

ishanvyas22:
Closes #6422 

I referred below links for param & return types.
1. https://github.com/php/php-src/blob/master/Zend/zend_fibers.stub.php
2. https://php.watch/versions/8.1/fibers

The [`FiberExit`](https://php.watch/versions/8.1/fibers#FiberExit) class is present in php.watch blog but not in github.com/php/php-src, but I've added it just in case. Do you think it's good idea to keep it? If not I can remove it, please suggest.


orklah:
It should be removed from both files

PR #6622, opened 2021-10-09
fix @return 0

### LINKED ISSUE #6424: @return 0 is ignored
https://psalm.dev/r/b63f4f2a2d

Replace 0 by any integer, positive or not and it's recognized correctly

orklah:
This should fix #6424

Seems like not even Psalm is immune to implicit casting :D


orklah:
Oh, this uncovered an old bug in the test suite. Nice!

PR #6625, opened 2021-10-09
Refined PDO::errorInfo, PDOStatement::errorInfo

vudaltsov:
See https://www.php.net/manual/ru/pdo.errorcode.php and https://www.php.net/manual/ru/pdostatement.errorinfo.php.


vudaltsov:
@orklah, we can do `array{0: string, 1: ?int, 2: ?string, 3?: mixed, 4?: mixed}`, but there might be 6 elements in some drivers...

At least we can say that it's a `non-empty-list` :)

orklah:
Both seem fine to me, pick the one you prefer :)

orklah:
Thanks!

PR #6629, opened 2021-10-09
tweaking with loops

### LINKED ISSUE #6626: Support `for(*;;*)` loop without break condition
https://psalm.dev/r/f870bb2de7
Loop with `for ($i = 0;; ++$i)`.

**Expected**
Psalm sees that the loop has no exit condition in its header, and the only way out is the return statement.
**Actual**
Psalm behaves as if the loop can break, and the function will return null in that case.

I also added examples with `for ($i = 0; true; ++$i)` and `while (true)`.
Apparently for `while (true)` it does understand there is no breaking exit.

orklah:
This brings a little more coherence between the four type of loops Analyzer:
- DoAnalyzer
- WhileAnalyzer
- ForAnalyzer
- ForeachAnalyzer

Some functions used were not the same, isAlwaysTruthy was not used in WhileAnalyzer but was used in ForAnalyzer etc...

This will fix #6626

### DIFF
--- a/tests/Loop/ForTest.php
+++ b/tests/Loop/ForTest.php
@@ -156,6 +156,26 @@ class ForTest extends \Psalm\Tests\TestCase
                         return $data;
                     }'
             ],
+            'InfiniteForLoop' => [
+                '<?php
+                    /**
+                     * @return int
+                     */
+                    function g() {
+                        for (;;) {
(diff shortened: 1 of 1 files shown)


orklah:
Interesting, tests are failing. I'll check if it's legit

PR #6648, opened 2021-10-11
`exit` statuses should be in the range 0 to 254

sasezaki:
from manual.

> If status is an int, that value will be used as the exit status and not printed. Exit statuses should be in the range 0 to 254, the exit status 255 is reserved by PHP and shall not be used. The status 0 is used to terminate the program successfully. 
https://www.php.net/manual/en/function.exit.php

actual PHP behaviors is like this.
```
sasezaki@z:~$ php -r "exit(0);"; echo $?
0
sasezaki@z:~$ php -r "exit(254);"; echo $?
254
sasezaki@z:~$ php -r "exit(255);"; echo $?
255
sasezaki@z:~$ php -r "exit(256);"; echo $?
0
sasezaki@z:~$ php -r "exit(-1);"; echo $?
255
sasezaki@z:~$ php -r "exit(-2);"; echo $?
254
```

---
Note: This is my first change for CallMap,  so if anything missing, please let me know.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ExitAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ExitAnalyzer.php
@@ -14,7 +14,7 @@ use Psalm\Issue\ImpureFunctionCall;
 use Psalm\IssueBuffer;
 use Psalm\Storage\FunctionLikeParameter;
 use Psalm\Type;
-use Psalm\Type\Atomic\TInt;
+use Psalm\Type\Atomic\TIntRange;
 use Psalm\Type\Atomic\TString;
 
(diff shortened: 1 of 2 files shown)


orklah:
Seems correct :)

I'm always a bit cautious about changing parameter types to non-standard types because it has the potential to bother a lot of people but I don't expect exit to be used a lot with dynamic params

weirdan:
Since `exit()` is not really a function, I'd suggest to add a test highlighting new behavior. Callmap could be ignored in this case, and you may need to change `src/Psalm/Internal/Analyzer/Statements/Expression/ExitAnalyzer.php` instead.

Looks good to me otherwise.

PR #6662, opened 2021-10-13
Removed TEmpty

orklah:
This PR removes completely TEmpty and replace it with TNever everywhere.

We probably won't be able to merge that as-is for BC reasons, but it will start the discussion.


weirdan:
Are there any specific benefits, apart from getting rid of one of types?

orklah:
Yeah, we get rid of a confusion because we never know which type we should use.

For example, I used Empty in arithmeticOperation when encountering an operation which threw an error, but I should have used never who emit an issue when assigning or using the value.

Never is the bottom type, empty is just a weird type we have for legacy reason and who sometimes act as a bottom type, sometimes an assertion, sometimes as a placeholder for empty arrays

I wanted to do that for quite some time, but PHPStan introduced their notion of literal empty array `array{}` and suggested to add support for our current annotation `array<empty, empty>` so if we want to change it, now would be a good time...

weirdan:
@orklah, I may have broken something during rebase; mind looking at this?

orklah:
@weirdan nothing major. Matt moved some code from Union to Atomic so my change here has been dropped. 

On Atomic.php line 805, the line should contain `array<never, never>` instead of `array<empty, empty>`. Can you make the change on your side please? I'm afraid I corrupted my clone on this branch somehow. I still have issues merging(or rebasing) after commits are made on branches I have on my side, I try to update but it never seem to feel right...

Anyway, I'll actually remove every occurence of `array<never, never>` to replace them with TArray::isEmptyArray later. (Technically, an array is empty as long as its second template is never. The first one is not required to be never, so it may fix some weird edge cases)

weirdan:
> On Atomic.php line 805, the line should contain array<never, never> instead of array<empty, empty>. Can you make the change on your side please?

Done.

>  I still have issues merging(or rebasing) after commits are made on branches I have on my side, I try to update but it never seem to feel right...

After someone force-pushes your branch, you need to reset your branch to that new head.

> would it be hard to make sure test splitting is the same between windows and linux?

Not hard at all. All we need to do is to use the same scripts on Linux as are currently used on Windows. This will mean we won't have that nice collapsible output that we currently have.

PR #6663, opened 2021-10-13
Allow Symfony 6

loic425:


orklah:
About behat, you'll note that's what they did here:
https://github.com/Behat/Behat/pull/1346/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R30

They introduced symfony 6 in their CI as a dev package to test against before merging.

orklah:
> @orklah Making this available on main branch allow third party packages to work on Symfony 6 support using your "dev-master".

If you have trouble installing Psalm along with other packages, you should use https://github.com/psalm/phar instead, it has no dependencies so it won't bother you with that

loic425:
@orklah Yes, I agree with you. I will work on this soon.

loic425:
@orklah I opened a [PR on Paratest](https://github.com/paratestphp/paratest/pull/639).

loic425:
@orklah 😵‍💫 
https://github.com/paratestphp/paratest/runs/4116716725?check_suite_focus=true

What's next? Do you advice Paratest to use the psalm phar instead?

weirdan:
I didn't realize we had circular dependencies.

weirdan:
I'd be ok with merging and releasing just range extension, without CI changes. Worst case scenario is we have `psalm-plugin` not working in some combination of Psalm and Symfony versions, and that's no big deal in my book.

I don't see a need to test compatibility with every Symfony version on every push.

loic425:
> I'd be ok with merging and releasing just range extension, without CI changes. Worst case scenario is we have `psalm-plugin` not working in some combination of Psalm and Symfony versions, and that's no big deal in my book.
> 
> I don't see a need to test compatibility with every Symfony version on every push.

I reverted the CI changes.

Issue #6673, opened 2021-10-15
Recursive array merging and replacing


ElisDN:
https://psalm.dev/r/cb702c3098

ElisDN:
@orklah Yes, I am doing this.

arjenm:
Here's another code-snippet that I'd expect gets fixed if the types are correct?
https://psalm.dev/r/36f4a82706

weirdan:
> Here's another code-snippet that I'd expect gets fixed if the types are correct? (Otherwise it's a separate bug). [psalm.dev/r/36f4a82706](https://psalm.dev/r/36f4a82706)

It works now: https://psalm.dev/r/36f4a82706

PR #6676, opened 2021-10-15
WIP: Types for `array_merge_recursive` and  `array_replace_recursive`

ElisDN:
For #6673

### DIFF
--- a/tests/ArrayFunctionCallTest.php
+++ b/tests/ArrayFunctionCallTest.php
@@ -264,6 +264,74 @@ class ArrayFunctionCallTest extends TestCase
                         }
                     }',
             ],
+            'arrayMergeRecursiveIntArrays' => [
+                '<?php
+                    $d = array_merge_recursive(["a", "b", "c"], [1, 2, [31, 32]]);',
+                'assertions' => [
+                    '$d' => 'list{string, string, string, int, int, list{int, int}}',
+                ],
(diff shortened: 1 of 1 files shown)


orklah:
Hey @ElisDN what is the status for this PR? Do you still need to work on it? Were you waiting a review?

ElisDN:
@orklah,

Global solution is to make `Type::combineUnionTypes` work recursive with `merge`, `replace` or `default` mode.

For example with additional `$mode` and `$recursive` flags:

```php
$function = $event->getFunctionId(); // like 'array_replace_recursive'

$is_replace_function = mb_strcut($function, 6, 7) === 'replace';
$is_recursive_function = mb_strcut($function, -8) === 'recursive';

$generic_properties[$key] = Type::combineUnionTypes(
    $generic_properties[$key],
    $type,
    $codebase,
    ...,
    $is_replace_function ? MODE_REPLACE : MODE_MERGE, // not MODE_DEFAULT with setting of possible_undefined
    $is_recursive_function
);
```

Is it a good idea?

orklah:
I'm not too keen on changing combineUnionTypes who is very used in Psalm with something that seems so specific. Could you instead call it recursively from the ArrayMergeReturnTypeProvider in a way that does not change its signature?

weirdan:

PR #6677, opened 2021-10-15
Fix get_object_vars() return signature

fluffycondor:
https://psalm.dev/r/ca1f1ecbe0

Current signature is incorrect. An array in the example will contain keys (int)1, (string)'baz', but `array<string, mixed>` is reported.


orklah:
I don't think we should merge that.

Creating numeric only keys through `->{1}` is a PHP weirdness more than a feature. In 99% of cases, this will never be the case and we'll degrade analysis for this.

If get_object_var was more clever and it could read this kind of property exists, then sure, return them as literals, but not like this (PHPStan does the same analysis).

What do you think @weirdan ?

PR #6679, opened 2021-10-16
Improve handling of positive ints by array_filter()

### LINKED ISSUE #5694: array_filter() without 2nd argument treats positive-int as int
Not something really simple but want to report it though.

The behavior of `array_filter()` on lack 2nd argument is determined and replicate's `empty()` check. So https://psalm.dev/r/628170286f should not have possible absent keys. All keys must be preserved for `non-empty-string`, `positive-int`, `true`, `object`, `non-empty-array`

SCIF:
Fixes #5694


SCIF:
> Fortunately, the code here could be much more simple. We have Union::isAlwaysTruthy and Union;;isAlwaysFalsy since a few weeks.

Wow!! So useful! Thanks heaps! Done.

orklah:
Thanks!

Issue #6682, opened 2021-10-17
Providing `iterable` to `array|\Traversable` triggers an `InvalidArgument` error


driehle:
In [doctrine/doctrine-laminas-hydrator](https://github.com/doctrine/doctrine-laminas-hydrator) we are using the function `ArrayUtils::iteratorToArray()` from [laminas/laminas-stdlib](https://github.com/laminas/laminas-stdlib). Though the name of that method suggests to expect an `iterable`, the method actually expects an `array|\Traversable` like this:

```php
    /**
     * @param array|\Traversable  $iterator     The array or Traversable object to convert
     * @return array
     */
    public static function iteratorToArray($iterator, $recursive = true)
    {
        return [];
    }
```

Now in our library, there is a `mixed` argument which is explicitly checked for being an `array` or an instance of `\Traversable`, before being passed to `ArrayUtils::iteratorToArray()`. This looks like this:

```php
    /**
     * @param  mixed $identifier
     */
    private function isNullIdentifier($identifier)
    {
        if ($identifier instanceof \Traversable || is_array($identifier)) {
            ArrayUtils::iteratorToArray($identifier);
        }
    }
```

It seems like Psalm - from the if statement - infers that `$idendifier` is an `iterable`, but then it complains because `iterable` is not an valid input for `ArrayUtils::iteratorToArray()`. The following error raises:

```
ERROR: InvalidArgument - 22:45 - Argument 1 of ArrayUtils::iteratorToArray 
  expects Traversable|array<array-key, mixed>, 
  iterable<array-key|mixed, mixed> provided
```

Probably a bug in Psalm? See https://psalm.dev/r/0e02a5ccf0 for the full example.

orklah:
It is indeed a bug.

From what I could tell, Psalm correctly infer two assertions about `$identifier`: `array` and `Traversable`. It then takes the initial `mixed` type and apply the `array` assertion to it, and it outputs the atomic type `TArray`. It does the same from `mixed` to `TNamedObject` with a value as `Traversable`

Then it pass the two of them to the TypeCombiner, which is the class that outputs 1 type from a list of multiples types. This is what outputs iterable. What puzzle me right now is that the TypeCombiner doesn't have a notion of union or intersection, so I'm not sure how it's actually supposed to do the difference between 
`is_array && Traversable` and `is_array || Traversable`

Issue #4700, opened 2020-11-25
Psalm 5 breaking changes


muglug:
- Redundant casts should become their own issue (currently `RedundantCondition`)

kkmuffme:
Perhaps extend this to redundant intval, floatval,... calls too?

muglug:
Closing, going to make this a minor change instead

weirdan:
* functions in `Psalm` namespace (https://github.com/vimeo/psalm/blob/3ad3375777b431a559b624d359c9af667de319f1/src/command_functions.php) are going to be removed, deprecated in #5858 
* `allowCoercionFromStringToClassConst` config attribute is going to be removed, deprecated in #5897 
* cli scripts in `src` folder are going to be removed, deprecated in #5858

orklah:
- Legacy plugins hooks should be removed too (or is it too soon?)

weirdan:
@orklah They need to be deprecated first, so I went ahead and marked them as such in #5898. We can always postpone the removal and do it in v6, but the earlier we announce deprecation the better.

muglug:
= make array shapes explicit — `array{foo: Bar}` means an array with exactly one element. `array{foo: Bar}&array<string, object>` can be used to denote an array with optional extra elements

weirdan:
> make array shapes explicit

Ideally this would first come with a config switch and v5 would just flip the default, unless you think that's too much work.

orklah:
I'm not sure introducing a config long term on something so fundamental would be beneficial for the project. It will make every future bug report hard to reason about depending on the value of the config.

weirdan:
The following methods were previously deprecated, linked PR marks them for removal in Psalm 5

* `DocComment::parse()`
* `TypeAnalyzer::isContainedBy()`, `TypeAnalyzer::isAtomicContainedBy()`
* `CodeIssue::getLocation()`, `CodeIssue::getFilename()`, `CodeIssue::getMessage()`

weirdan:
Things to be removed can be located by grepping the `src` folder for the `Psalm 5` substring.

VincentLanglet:
> I'm not sure introducing a config long term on something so fundamental would be beneficial for the project. It will make every future bug report hard to reason about depending on the value of the config. 

The format array{foo: string}&array<mixed> is currently not supported by phpstan. So this Bc break would require to have both a phpstan-param and psalm-param annotation for all the array phpdoc.
[...]

PR #6689, opened 2021-10-18
only fuse array & Traversable with phpstorm generics

orklah:
This should fix https://github.com/vimeo/psalm/issues/6682

The change is pretty weird though but I couldn't make sense of it otherwise. It seems this piece of code will only merge Array&Traversable into Iterable if we're *not* allowing PHPStorm generics. In fact, allowing them fix the example: https://psalm.dev/r/ba98d80cff

So I guess it was just a mistake. Seems like no test was covering this neither.

I'm very much thinking about deprecating the "allow PHPStorm generics" and removing it in Psalm 5. PHPStorm now supports standard generics notation so it makes little sense to have this now but this will come back to bite us when we'll try working with intersections. With https://github.com/vimeo/psalm/pull/6672, it's the second fix I make to this config in a week.

### DIFF
--- a/tests/TypeCombinationTest.php
+++ b/tests/TypeCombinationTest.php
@@ -68,7 +68,8 @@ class TypeCombinationTest extends TestCase
             'ArrayAndTraversableNotIterable' => [
                 '<?php declare(strict_types=1);
 
-                    function isNullIdentifier(mixed $identifier): bool
+                    /** @param mixed $identifier */
+                    function isNullIdentifier($identifier): bool
                     {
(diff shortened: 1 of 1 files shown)


weirdan:
> I'm very much thinking about deprecating the "allow PHPStorm generics" and removing it in Psalm 5

:+1: on this. Mind adding it to #4700 ?

orklah:
> I think the idea was to prevent treating `Traversable|float[]` as `iterable<mixed, mixed>`.

It will actually do the opposite if I'm not mistaken. But only if you don't use PHPStorm generics. If you do,  `Traversable|float[]`  will stay as is

Issue #3249, opened 2020-04-27
Errors in overwritten stub reported for function that overwrites


kkmuffme:
When I have stub:

```
/**
 * @param string $arg hello
 */
function some_test( $arg ) {
}
```

but then declare this function in my real code (e.g. bc in original code the stubbed function is wrapped in function_exists or whatever) like:
```
/**
 * @param string $arg hello
 *
 * @return string
 */
function some_test( $arg ) {
    echo $arg;
}
```

Then psalm tells me:
MissingReturnType

bc in the stub there is no return type specified

Since the return type is specified in the "real" function, this is incorrect.

kkmuffme:
no, bc this requires setting a stub in the config file which isn't possible on psalm.dev

muglug:
@kkmuffme could you reproduce in a separate repo?

kkmuffme:
@muglug you can literally just copy the code I posted in OP in a test.php and stubs.php then you will get the issue.

If you still want it, I can create a zip for it though

muglug:
Even better: create a PR with a test that reproduces the bug (there are a bunch of stub tests in `StubTests.php`

PR #6733, opened 2021-10-25
Errors in overwritten stub reported for function that overwrites

kkmuffme:
https://github.com/vimeo/psalm/issues/3249

This is not a fix, but just the code to reproduce the issue as requested.

### DIFF
--- a/tests/StubTest.php
+++ b/tests/StubTest.php
@@ -646,6 +646,42 @@ class StubTest extends TestCase
         $this->analyzeFile($file_path, new Context());
     }
 
+    public function testStubFunctionReturnTypeOverwrite(): void
+    {
+        $this->project_analyzer = $this->getProjectAnalyzerWithConfig(
+            TestConfig::loadFromXML(
+                dirname(__DIR__),
+                '<?xml version="1.0"?>
+                <psalm
+                    errorLevel="1"
+                >
+                    <projectFiles>
+                        <directory name="src" />
+                    </projectFiles>
+
+                    <stubs>
+                        <file name="tests/fixtures/stubs/custom_functions.phpstub" />
+                    </stubs>
+                </psalm>'
+            )
+        );
+
+        $file_path = getcwd() . '/src/somefile.php';
+
+        $this->addFile(
+            $file_path,
+            '<?php
+            /**
+             * @return string
+             */
+            function testReturnTypeOverwrite() {
+                echo "foo";
+            }'
+        );
+
+        $this->analyzeFile($file_path, new Context());
+    }
+
     public function testNoStubFunction(): void
     {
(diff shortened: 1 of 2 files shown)


kkmuffme:
>If you know the parent function will return string, you're free to add that to your stub, at least in phpdoc, this should keep you from getting this

The issue is, that I have 1000s of functions with missing or too general phpdocs.

The issue basically is: why does psalm give me an error for a stub, when the function I declared (non-stub) is correct.

What would need to change is:
If a function is declared in code, it should always take priority over stubs.

orklah:
I made some experiments with function stubbing:
- Psalm will enforce types and numbers of params of the stub instead of those of the declared function
- Psalm will trust phpdoc return type of the stub over the phpdoc return type of the declared function
- Psalm will trust signature return type of the stub over the signature return type of the declared function
- Psalm will trust signature return type of the stub over the phpdoc return type of the declared function
- If the stub has no type (in signature nor in phpdoc) Psalm will infer return type from the stub and trust that

I made limited tests but it appears to work about the same with methods instead of functions, with one exception:
- Psalm will emit an InvalidReturnType when a method doesn't have return statement but the parent stub was inferred as returning string instead of a MissingReturnType like it throws on functions

Overall, the behaviour seem coherent with what I'd expect of stubs, they're made to replace concrete implementation with different/more precise types. I guess we could throw an InvalidReturnType instead of MissingReturnType for functions too.

orklah:
I tried to check, both errors are thrown by ReturnTypeAnalyzer::verifyReturnType. the difference is that this method receive a non-null $return_type (of `string`) for the method and null for the function.

I tried to look further but the way the return type is retrieved is quite different between the two and I didn't find an obvious path to tackle this.

I'm afraid I do not care this much for this bug as it happens for:
- incomplete stubs
- when the function is then found in the code
- with a documented return type for the function that do not match the statements in the function

I won't spend more of my time on this, though if you want to try to work on this, I'll gladly help you and give you pointers.

kkmuffme:
Maybe a simple solution would be to give the actual code priority over any stubs? Perhaps as a configurable option?

So all this:
>Psalm will enforce types and numbers of params of the stub instead of those of the declared function
Psalm will trust phpdoc return type of the stub over the phpdoc return type of the declared function
Psalm will trust signature return type of the stub over the signature return type of the declared function
Psalm will trust signature return type of the stub over the phpdoc return type of the declared function
If the stub has no type (in signature nor in phpdoc) Psalm will infer return type from the stub body and trust that

Would be reversed, with phpdoc of the declared function having priority over stubs.

weirdan:
> Maybe a simple solution would be to give the actual code priority over any stubs? 

Not going to happen. The main use case of stubs in Psalm is to override functions / methods / classes etc.

PR #6777, opened 2021-10-28
Correct pgsql stubs

### LINKED ISSUE #6760: Incorrect stubs for pgsql functions
There are two groups of problems:

First one was introduced in #6671, functions that execute queries are marked as returning `\Pgsql\Connection` when in fact they return `\Pgsql\Result`

https://psalm.dev/r/a29523a02f

Second one is that `pg_escape_identifier` and `pg_escape_literal` are marked as returning `string` only. These may return `false` on error, the easiest way to trigger that is to pass malformed UTF-8 for UTF-8 connection

https://psalm.dev/r/0a608a6b9b

Actually running the above script with a proper connection string results in "This fails" output twice and warnings emitted

sad-spirit:
This should fix #6760

Checked the docs and the source of ext/pgsql so that is what is _really_ accepted and returned (there are differences to what is documented, e.g. `pg_escape_identifier` and `pg_escape_literal` I mentioned in the above issue)


orklah:
Great! Thanks!

PR #6783, opened 2021-10-29
Add a means to switch return types based on PHP_VERSION_ID

sad-spirit:
As suggested by @weirdan in vimeo/psalm#6755 considering return type changes in PHP 8.1

I wouldn't claim that I completely understand all the changes I made, but I used https://github.com/vimeo/psalm/commit/42802e11d1327b99ff9bdd3bafbcd3291c406e91 as a reference.

### DIFF
--- a/tests/Template/ConditionalReturnTypeTest.php
+++ b/tests/Template/ConditionalReturnTypeTest.php
@@ -864,14 +864,26 @@ class ConditionalReturnTypeTest extends TestCase
                     /**
                      * @psalm-return (PHP_VERSION_ID is int<70300, max> ? string : int)
                      */
-                     function getSomething()
-                     {
+                    function getSomething()
+                    {
                         return mt_rand(1, 10) > 5 ? "a value" : 42;
-                     }
+                    }
 
-                     $something = getSomething();
+                    /**
+                     * @psalm-return (PHP_VERSION_ID is int<70100, max> ? string : int)
+                     */
+                    function getSomethingElse()
+                    {
+                        return mt_rand(1, 10) > 5 ? "a value" : 42;
+                    }
+
+                    $something = getSomething();
+                    $somethingElse = getSomethingElse();
                 ',
-                ['$something' => 'int'],
+                [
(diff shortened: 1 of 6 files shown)


tests/Template/ConditionalReturnTypeTest.php
```diff
@@ -859,6 +859,22 @@ function test(string $s): void {
                     }
                     '
             ],
+            'returnTypeBasedOnPhpVersionId' => [
+                '<?php
+                    /**
+                     * @psalm-return (PHP_VERSION_ID is int<70300, max> ? string : int)
```
weirdan, line 865: I would never have thought of using int ranges in that way, that's quite creative :thinking: :bulb:.

tests/Template/ConditionalReturnTypeTest.php
```diff
@@ -859,6 +859,22 @@ function test(string $s): void {
...
+                     function getSomething()
+                     {
+                        return mt_rand(1, 10) > 5 ? "a value" : 42;
+                     }
+
+                     $something = getSomething();
+                ',
+                ['$something' => 'int'],
+                [],
+                '7.2'
```
weirdan, line 876: You may want to add another test with the same code to show up it returns `string` on PHP 7.3+

tests/Template/ConditionalReturnTypeTest.php
sad-spirit: Well, I was _really_ glad to find out that psalm supports ranges, as specifying versions with constants / lists would not be so easy. Ranges aren't quite documented...

That being said, docs on conditional return types don't mention existing special cases of `func_num_args()` and `PHP_MAJOR_VERSION` either.

tests/Template/ConditionalReturnTypeTest.php
sad-spirit: Thanks, added another check to that test with essentially the same function but a range starting from PHP 7.1

weirdan:
Thanks!

PR #6785, opened 2021-10-31
Require one of `release:*` labels on PRs

weirdan:
This PR is also a live demo - you can label / unlabel it and see results in real time.


orklah:
ahah, yes, please add that, I never remember to add them!

PR #6802, opened 2021-11-03
Add pcntl_alarm to the list of impure functions

l-x:
`pcntl_alarm` is currently considered as a pure function, which is not correct.

See https://psalm.dev/r/c51166d9bb


weirdan:
Thanks!

PR #6807, opened 2021-11-03
deprecate exitFunctions in XML

### LINKED ISSUE #6794: deprecate exitFunctions in Psalm 5?
This is an undocumented config that allows adding always exiting functions for Psalm.

However, this is not considered everywhere and it has drawbacks (you have to remember it's there, retrieving them from config, resolving calls to check them...).

Fortunately, since we now have `never`, the whole feature became redundant, as declaring never on phpdoc or on a stub will have the same benefits without drawbacks. 

That's why I'm suggesting to drop this config for Psalm 5

orklah:
this PR deprecate exitFunctions and fix #6794

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -16,7 +16,14 @@
             <xs:element name="mockClasses" type="MockClassesType" minOccurs="0" maxOccurs="1" />
             <xs:element name="stubs" type="StubsType" minOccurs="0" maxOccurs="1" />
             <xs:element name="plugins" type="PluginsType" minOccurs="0" maxOccurs="1" />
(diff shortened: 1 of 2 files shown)


weirdan (changes requested):
Please add it here as well: https://github.com/vimeo/psalm/blob/852f9e87073b9b64e10e2100d63c8272a1159a05/src/Psalm/Config.php#L764-L767

weirdan:
> Please add it here as well:

And while you're at it, it probably makes sense to add `allowPhpStormGenerics` there as well.

weirdan (approved):

PR #6809, opened 2021-11-03
don't taint the result of most binary operations

### LINKED ISSUE #6804: bool comparison should remove tainted input from result
https://psalm.dev/r/132f9272f9

Result of `$taintedInput === $otherValue` shouldn't be marked as tainted.

orklah:
As far as I can tell, only Concat and Coalesce binary ops can pass taint to the result.

To whoever it may concern: this method is called for other Expr, there are other Expr that does not imply tainted results (I'm thinking PostInc for example: https://psalm.dev/r/4e6d6d9566). If someone want to give a hand and list all Expr that could and couldn't pass taint, it would be most helpful and I'd gladly amend this to add them all.

This will fix #6804


weirdan:
> As far as I can tell, only Concat and Coalesce binary ops can pass taint to the result.

What about `+` used on arrays (https://www.php.net/manual/en/language.operators.array.php)?

orklah:
Nice catch, I'll fix that

PR #6810, opened 2021-11-03
Array cast pass taints

### LINKED ISSUE #5550: Typecasting to array breaks taint flow
https://psalm.dev/r/405dfa7e9d

Using the above snippet, SQLi at line 43 was not detected when the tainted variable is typecasted to array.

orklah:
This makes array casts taint sensitive and fix #5550

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
@@ -232,8 +232,7 @@ class CastAnalyzer
                 $type = Type::getArray();
             }
 
-            if ($statements_analyzer->data_flow_graph instanceof \Psalm\Internal\Codebase\VariableUseGraph
-            ) {
+            if ($statements_analyzer->data_flow_graph) {
                 $type->parent_nodes = $stmt_expr_type->parent_nodes ?? [];
             }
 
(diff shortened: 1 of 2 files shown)


weirdan:
The issue with invalid data returned from the data provider should have been caught by Psalm, but somehow it wasn't. Even though one of the elements is missing `error_message` key Psalm thinks the returned array is `non-empty-array<string, array{0: non-empty-string, error_message: non-empty-string}>`. I wonder if there's any threshold after which array elements stop contributing to the inferred shape.

orklah:
>  I wonder if there's any threshold after which array elements stop contributing to the inferred shape.

If I remember correctly, it's this one: https://github.com/vimeo/psalm/blob/64df00baba61af63af4cf42ab1489cc43a959974/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php#L461

weirdan:
Psalm uses at most 101 elements to infer array shape: https://psalm.dev/r/5f192e37bf
And at most 102 elements to infer array value type: https://psalm.dev/r/62d9666fdc
Elements after 102th are ignored: https://psalm.dev/r/6462ff9bcc

This doesn't look right. While it's understandable that we cannot allow unbounded shapes, we probably should still be able to use all elements to infer array value/key type (as we actually process all elements).

Issue #3709, opened 2020-06-29
Taint detection crashes for __toString method inherited from ancestor class


TysonAndre:
See https://psalm.dev/r/a4d70b780f

Expected: Emits TaintedInput, does not crash with `src/Psalm/Internal/Codebase/Methods.php: $storage should not be null for TaintedSubclass::__tostring`

I assume this should load the class and find the method from the ancestor class

TysonAndre:
Thanks!

PR #6813, opened 2021-11-03
don't register taints for numeric variables

### LINKED ISSUE #4872: taint analysis: auto escape for int type hints
I would not expect any error here: https://psalm.dev/r/d7f81c98a6

(It already works for explicit int casts: https://psalm.dev/r/363c53b0e0)

orklah:
this fixes #4872

I was pretty confused for a moment, I expected two fixes: get rid of the taint when passing through params, and when passing through a return type

However, as soon as I fixed the params one, both disappeared. I then realised that echo was treated as a function and it stopped registering taint given they were ints. I searched for sinks that are not functions (or treated as such by psalm) and couldn't find some. Maybe I'm mistaken on this.

Note also that this PR represent a subtle change in behaviour: declaring that a function takes an int as a param or gives int as a return type *in phpdoc* will now remove possible taints. It may be possible to do so only if the type does not come from docblock but it seemed more aligned with Psalm's policy to trust types even in docblock

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/EchoAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/EchoAnalyzer.php
@@ -38,7 +38,7 @@ class EchoAnalyzer
             $expr_type = $statements_analyzer->node_data->getType($expr);
 
             if ($statements_analyzer->data_flow_graph instanceof TaintFlowGraph) {
-                if ($expr_type) {
+                if ($expr_type && $expr_type->hasObjectType()) {
                     $expr_type = CastAnalyzer::castStringAttempt(
                         $statements_analyzer,
                         $context,
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
@@ -1505,6 +1505,14 @@ class ArgumentAnalyzer
             return $input_type;
         }
 
+        // numeric types can't be tainted
+        if ($statements_analyzer->data_flow_graph instanceof TaintFlowGraph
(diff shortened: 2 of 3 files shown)


src/Psalm/Internal/Analyzer/Statements/EchoAnalyzer.php
```diff
@@ -38,7 +38,7 @@ public static function analyze(
             $expr_type = $statements_analyzer->node_data->getType($expr);
 
             if ($statements_analyzer->data_flow_graph instanceof TaintFlowGraph) {
-                if ($expr_type) {
+                if ($expr_type && $expr_type->hasObjectType()) {
```
orklah, line 41: Note: this was done to fix https://github.com/vimeo/psalm/issues/3709 originally but this had the side-effect of turning any int into a `numeric-string`. I was not comfortable skipping taint analysis for a possible string, so I reduced the scope of this fix in order to retrieve the real int type later.

muglug (approved):

src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
```diff
@@ -1505,6 +1505,14 @@ private static function processTaintedness(
             return $input_type;
         }
 
+        // numeric types can't be tainted
+        if ($statements_analyzer->data_flow_graph instanceof TaintFlowGraph
+            && $input_type->isSingle()
+            && ($input_type->isInt() || $input_type->isFloat())
```
staabm, line 1511: Same for bool?

orklah: Probably, and there's surely more! We'll gladly accepts PRs :)

PR #6828, opened 2021-11-05
allow declaring functions and classLikes after a never-returning call

### LINKED ISSUE #6827: don't emit issue for functions or class after a never-returning function call
https://psalm.dev/r/d043c70a10?php=8.1

orklah:
This fixes #6827


weirdan:
Note that it doesn't extend to conditionally-defined symbols: https://3v4l.org/TZrHR

orklah:
No, it's just a side effect of being at top level on global scope. With your example, Psalm will still flag the if as unused

It also doesn't work in this case: https://3v4l.org/THj5q but here, Psalm won't pick up "UnevaluatedCode". I don't think it's a big issue though because it has nothing to do with `never`, it's just that PHP require the function to be above the call (unless it's at top level in global scope)

PR #6830, opened 2021-11-05
Mixed contain every types and Never is contained in every type

orklah:
This adds very fast responses for UnionTypeComparator where container are mixed or input are never because they are the top and bottom type so every type is contained by Mixed and every type contains Never.

This has the potential to make some checks much faster but I mainly did it for clarity when debugging and for consistency of the type system


weirdan:
The opposite could also be useful: `never` contains nothing (except maybe `never`?)

orklah:
Yeah, that's more tricky. Interactions between Void, Never, Empty and Null starts to come into consideration

Issue #6831, opened 2021-11-06
Notations assert-if*  does not support unions in the same way as assert


SCIF:
https://psalm.dev/r/f7ea0b4cf0

SCIF:
@orklah , @weirdan , I added tests in #6832 and I'm really confused now. Initially I tried to simplify tests and use assertions array and found another discrepancy?? Or could you please explain what's wrong with [assertion strings](https://github.com/vimeo/psalm/pull/6832/files#diff-edd0e7b0d24ba2010328c8b718551cffe016b73cea5d1281ee1a792da17f69baR516-R518)?

orklah:
In assertions, you should use `$var===` instead of `$var` to get `class-string|'self'` instead of `class-string|string`

SCIF:
Thanks for that explanation! Could you please advise if there is documentation page explaining this behavior so I could later refer to?

orklah:
I don't think it's documented, at least I don't know where it would be...

weirdan:
It's undocumented as it's just used in the tests and nowhere else.

SCIF:
Here is a quote from tests showing the issue:
```
  1) Psalm\Tests\AssertAnnotationTest::testValidCode with data set "unionOfStringTypes" ('<?php\n                    /*...     }', array('"self"|class-string', '"self"|class-string', '"self"|class-string'))
  Failed asserting that two arrays are identical.
  --- Expected
  +++ Actual
  @@ @@
   Array &0 (
       '$var===' => '"self"|class-string'
  -    '$varTrue===' => '"self"|class-string'
  -    '$varFalse===' => '"self"|class-string'
  +    '$varTrue===' => 'class-string'
  +    '$varFalse===' => 'class-string'
   )
   ```

orklah:
That's actually reassuring, I was convinced this didn't work for standard assertion neither. That means the correct mechanic is there somewhere...

PR #6832, opened 2021-11-06
assert-if-* for unions fixes

SCIF:
Tests for https://github.com/vimeo/psalm/issues/6831 added

### DIFF
--- a/tests/AssertAnnotationTest.php
+++ b/tests/AssertAnnotationTest.php
@@ -437,6 +437,41 @@ class AssertAnnotationTest extends TestCase
 
                     if (!is_int($a)) $a->bar();',
             ],
+            'unionOfStringTypes' => [
+                '<?php
+                    /** @psalm-assert "self"|class-string $b */
+                    function a(string $b): void {
+                        if ("" === $b) { throw new \Exception("");};
+                    }
+
+                    /** @psalm-assert-if-true "self"|class-string $b */
+                    function atrue(string $b): bool {
+                        return $b === "";
+                    }
+
+                    /** @psalm-assert-if-false "self"|class-string $b */
+                    function afalse(string $b): bool {
+                        return $b === "";
+                    }
+
+                    /** @var string $var */;
+                    a($var);
+
+                    /** @var string $varTrue */;
+                    if (!atrue($varTrue)) {
+                        exit;
+                    }
+
+                    /** @var string $varFalse */;
+                    if (afalse($varFalse)) {
+                        exit;
+                    }',
+                [
+                    '$var===' => '"self"|class-string',
+                    '$varTrue===' => '"self"|class-string',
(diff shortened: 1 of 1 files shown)


tests/AssertAnnotationTest.php
```diff
@@ -1474,6 +1555,17 @@ function getPropertyType(\ReflectionProperty $reflectionItem): ?string {
...
+                '<?php
+                    /** @return class-string|"self"|"static"|null */
+                    function getPropertyType(\ReflectionProperty $reflectionItem): ?string {
+                        $type = $reflectionItem->getType();
+                        return ($type instanceof \ReflectionNamedType) && !$type->isBuiltin() ? $type->getName() : null;
+                    }',
+                [],
+                [],
+                '8.0',
+            ],
```
SCIF, line 1568: Disregard this test at the moment pls. It's from another PR, didn't clean. Will do later

tests/AssertAnnotationTest.php
```diff
@@ -1465,7 +1546,7 @@ function preserveContainerAllArrayIterator($value) {
             ],
             'reflectionNameTypeClassStringIfNotBuiltin' => [
                 '<?php
-                    /** @return class-string|null */
+                    /** @return class-string|"self"|int|null */
```
SCIF, line 1549: Disregard pls

tests/AssertAnnotationTest.php
```diff
@@ -437,6 +437,87 @@ function assertIntOrFoo($b) : void {
...
+                    check($var);
+                    check($varTrue);
+                    check($varFalse);
+                    ',
+                [
+                    '$var' => 'class-string|"self"',
+                    '$varTrue' => 'class-string|"self"',
+                    '$varFalse' => 'class-string|"self"',
+                ],
+            ],
```
SCIF, line 520: Tests are identical, but second one checks for assertion statements and it looks like it doesn't work fine by some reason

tests/AssertAnnotationTest.php
```diff
@@ -437,6 +437,87 @@ function assertIntOrFoo($b) : void {
...
+                    exit;
+                }
+
+                /** @var string $varFalse */;
+                if (afalse($varFalse)) {
+                    exit;
+                }
+
+                check($var);
+                check($varTrue);
```
orklah, line 474: This one is not working as you think it is. I believe `$varTrue` is just `class-string` instead of `"self"|class-string` but check() accepts `class-string` as a valid value for `"self"|class-string` so it's not giving you errors. Your assertion test is more trustworthy

SCIF: That's exactly what I tried to test — trace shows `class-string|"self"` for `$var`, but just `class-string` for -if* assertions.

PR #6839, opened 2021-11-06
Added test for NoEnumProperties

weirdan:
This was missed from vimeo/psalm#6836 for some reason.


.php-version
```diff
@@ -1 +0,0 @@
-8.0
```
weirdan, line 1: This should never have been committed in the first place.

orklah (approved):

Issue #6842, opened 2021-11-06
Argument 3 of openssl_sign expects resource|string, OpenSSLAsymmetricKey provided


jirkace:
Function `openssl_sign` [accepts](https://www.php.net/manual/en/function.openssl-sign.php) `OpenSSLAsymmetricKey|OpenSSLCertificate|array|string` for third parameter for PHP8

> 8.0.0 | private_key accepts an OpenSSLAsymmetricKey or OpenSSLCertificate instance now; previously, a resource of type OpenSSL key or OpenSSL X.509 was accepted.

This is new bug in version 4.12.0

jirkace:
> Hey @jirkace, can you reproduce the issue on https://psalm.dev ?

yes :-)

here is it: https://psalm.dev/r/bd9b345da1

weirdan:
> This is new bug in version 4.12.0

Hardly. The types in the callmap entry for this function weren't updated since 2018.

Feel free to propose a PR to update callmaps. We have a guide on doing that here: https://psalm.dev/docs/contributing/editing_callmaps/

jirkace:
Ok, sorry - I understood it as a new bug because it appeared after upgrade from 4.11.2 to 4.12.0. :man_shrugging:

orklah:
@weirdan what changed is https://github.com/vimeo/psalm/pull/6758/files who outputs objects since 4.12.0 and lead to the new error in the function called after

jirkace:
whats is wrong? [there](https://github.com/vimeo/psalm/actions/runs/1431009766) something failed, but I don't know why

orklah:
You can ignore that. All PR requires having a label in order to get merged. I added the label for yours but the check was already made so there was an error at one point but it is now solved

jirkace:
@orklah ok, thank you :-)

PR #6846, opened 2021-11-07
Fixed callmap for function openssl_sign, it changed in PHP 8.0 (modif…

jirkace:
…ied types for param private_key)  #6842


weirdan:
The docs say `array` is also accepted as a private key. For the array format, see https://www.php.net/manual/en/openssl.certparams.php:

```
For private keys, you may also use the syntax `array($key, $passphrase)` where `$key` represents a key specified using the `file://` or textual content notation above, and `$passphrase` represents a string containing the passphrase for that private key 
```

jirkace:
@weirdan I'm sorry about that and thanks for notice. Is `list<string>` correct type notation for that `array` in CallMap? I've updated it PR.

weirdan:
I'm not sure, it may accept `array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}`, and I don't have `openssl` extension installed to test it.

Anyway, `list` would be better replaced with a tuple like shown above, as docs clearly suggest the function expects two-element array, and `list` allows any number of elements (including 0).

jirkace:
@weirdan I've tried locally run snippet of code and it looks like it works with `array{OpenSSLAsymmetricKey, string}` although I understand from the documentation that it can only be there `array{string, string}`

> $key represents a key specified using the file:// or textual content notation

`textual content notation` is `string` only for me :-)

weirdan (approved):

weirdan:
Thanks!

Issue #5798, opened 2021-05-19
Add editor url option


Jean85:
PHPStan just released a patch with an `editorUrl` new options, that makes file names clickable in the terminal, and opens them in your editor of choice: https://github.com/phpstan/phpstan-src/pull/515

This is a huge benefit for a small improvement... Could we think of adding the same here on Psalm? Being able to jump to the exact line highlighted by errors would speed up the workflow by a long shot!

weirdan:
That would better fit a machine-local config, and Psalm doesn't have local configs right now.

PR #6850, opened 2021-11-07
ConsoleReport: Add hyperlinks to open file in editor

gharlan:
`ConsoleReport` adds hyperlinks to the file paths to open them in editor. It uses the file format from `xdebug.file_link_format` (you do not need xdebug extension for this) and falls back to generic `file://...` link.

Terminals not supporting hyperlinks should ignore these escape sequences, see https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#backward-compatibility:

> Any terminal that correctly implements OSC parsing according to ECMA-48 is guaranteed not to suffer from compatibility issues. That is, even if explicit hyperlinks aren't supported, the target URI is silently ignored and the supposed-to-be-visible text is displayed, without artifacts.

psalm output (hover over one link, you can see the target at the bottom):

Bildschirmfoto 2021-11-07 um 14 36 00

taint analysis:

Bildschirmfoto 2021-11-07 um 14 36 32

This fixes https://github.com/vimeo/psalm/issues/5798

### DIFF
--- a/src/Psalm/Report/ConsoleReport.php
+++ b/src/Psalm/Report/ConsoleReport.php
@@ -112,6 +112,12 @@ class ConsoleReport extends Report
      */
     private function getFileReference($data): string
     {
+        $reference = $data->file_name . ':' . $data->line_from . ':' . $data->column_from;
+
+        if (!$this->use_color) {
+            return $reference;
+        }
+
         if (null === $this->linkFormat) {
             // if xdebug is not enabled, use `get_cfg_var` to get the value directly from php.ini
             $this->linkFormat = ini_get('xdebug.file_link_format') ?: get_cfg_var('xdebug.file_link_format')
@@ -119,7 +125,7 @@ class ConsoleReport extends Report
         }
 
         $link = strtr($this->linkFormat, ['%f' => $data->file_path, '%l' => $data->line_from]);
-        $reference = $data->file_name . ':' . $data->line_from . ':' . $data->column_from;
+        // $reference = $data->file_name . ':' . $data->line_from . ':' . $data->column_from;
 
         return "\033]8;;" . $link . "\033\\" . $reference . "\033]8;;\033\\";
     }
--- a/src/Psalm/Report/ConsoleReport.php
+++ b/src/Psalm/Report/ConsoleReport.php
@@ -3,12 +3,19 @@ namespace Psalm\Report;
 
 use Psalm\Config;
 use Psalm\Internal\Analyzer\DataFlowNodeData;
+use Psalm\Internal\Analyzer\IssueData;
 use Psalm\Report;
 
+use function get_cfg_var;
+use function ini_get;
+use function strtr;
 use function substr;
 
 class ConsoleReport extends Report
 {
+    /** @var string|null */
+    private $linkFormat;
+
     public function create(): string
     {
         $output = '';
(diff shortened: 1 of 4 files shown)


weirdan:
Are there any provisions for mapped folders? E.g. Psalm running in a docker container sees files at `/var/www/` but IDE needs to open them from `/home/weirdan/src/project/`.

gharlan:
> Are there any provisions for mapped folders? E.g. Psalm running in a docker container sees files at `/var/www/` but IDE needs to open them from `/home/weirdan/src/project/`.

Hmm, do you have an idea how to solve this?

weirdan:
Thinking out loud here.

This can be handled either by Psalm itself emitting host paths or terminal emulator converting paths in links to their host counterparts. The latter is outside of our control and I have doubts of its universal support anytime soon. This leaves us with the need to implement it ourselves.

The mapping is both machine and project specific: Alice may have her projects in `/home/alice/src/<ProjectName>` while Bob keeps them in `C:\Users\Bob\Documents\<CompanyName>\<ProjectName>`. Some projects will have their container paths rooted in `/var/www/`, some may use another prefix like `/home/web`. This dictates the need for what I would call local project configuration. This deviates from single, committed to git, config file Psalm currently uses. Local mapping specification would have to be .gitignore-d. There are two options currently supported by Psalm:

 * Commit `psalm.xml.dist` and .gitignore `psalm.xml`. Developers working on a project would have to copy `psalm.xml.dist` to `psalm.xml` and add their path mapping there. Pros: easy to understand, familiar approach used by many tools. Cons: developers would have to update their `psalm.xml` when upstream `psalm.xml.dist` changes.
 * Commit `psalm.xml` that uses XInclude to include file containing path mapping. The `<xi:include>` would have to have `<xi:fallback>` element to prevent errors in CI (where local project mapping file is absent). The mapping file would have to be .gitignored. Pros: developers don't need to update their config. Cons: setup is a bit more complicated.

Neither of the options require any additional development as they are already available.

As for the configuration format, I envision something like this:
```xml
<psalm>
  <directoryMapping>
     <directory from="<absolute prefix>" to="<absolute prefix"/>
  </directoryMapping>
</psalm>
```

With that said, it can totally be a separate feature we may add later.

src/Psalm/Report/ConsoleReport.php
```diff
@@ -102,4 +106,21 @@ private function getTaintSnippets(array $taint_trace) : string
...
+        if (null === $this->linkFormat) {
+            // if xdebug is not enabled, use `get_cfg_var` to get the value directly from php.ini
+            $this->linkFormat = ini_get('xdebug.file_link_format') ?: get_cfg_var('xdebug.file_link_format')
+                ?: 'file://%f#L%l';
+        }
+
+        $link = strtr($this->linkFormat, ['%f' => $data->file_path, '%l' => $data->line_from]);
+        $reference = $data->file_name . ':' . $data->line_from . ':' . $data->column_from;
+
+        return "\033]8;;" . $link . "\033\\" . $reference . "\033]8;;\033\\";
```
weirdan, line 124: Perhaps we should disable links when Psalm is being run with '-m' option (you can check it with `$this->use_color`) as that's the only option users have to prevent buggy terminals from displaying ANSI sequences.

weirdan (changes requested):

weirdan (approved):

weirdan:
Thanks!

PR #6852, opened 2021-11-07
Fixed signature for newInstanceArgs for PHP 8.0

### LINKED ISSUE #5136: Incorrect docblock for newInstanceArgs
The ReflectionClass::newInstanceArgs method is currently defined as 
@param array<int, mixed> $args

but for PHP8 it should be instead @param array<array-key, mixed> $args

Here is a perfectly valid example with a false positive:

https://psalm.dev/r/8481bf4b77

jirkace:
Fixes #5136


weirdan:
Thanks!

PR #6873, opened 2021-11-09
load the ReturnTypeWillChange stubs for all versions to allow using it on every PHP version

### LINKED ISSUE #6849: The class ReturnTypeWillChange doesn’t have the Attribute attribute
When I run Psalm 4.12.0 on a codebase that uses the new `ReturnTypeWillChange` attribute, I get a lot of errors regarding that attribute. The message is always:

> The class ReturnTypeWillChange doesn’t have the Attribute attribute

Interestingly enough, this only happens when I run Psalm on PHP 8.1 (8.1.0-rc5 currently). If I use PHP 8.0.12 to run Psalm on the very same codebase, the error is not reported.

When using reflection to analyze `ReturnTypeWillChange` on PHP 8.1, we can see that the class actually has that attribute: https://3v4l.org/H7ngu

### LINKED ISSUE #6854: ReturnTypeWillChange  on PHP 8.0 or earlier
ReturnTypeWillChange  was introduced in PHP 8.1 in order to prevent PHP from emitting deprecated error when it sees return types that are not respected.

This was needed to make sure libraries that need to stay compatible with more than one version of PHP can have a code that doesn't emit deprecation nor errors.

The issue for us is that ReturnTypeWillChange  did not exists in PHP 8.0 so Psalm is reporting it as undefined. While technically true, it's still legit to have this on PHP 8.0 code and it's not an error from user perspective.

Should we make a special case for this and ignore the fact the attribute doesn't exists on PHP 8.0 and earlier?

orklah:
This will fix https://github.com/vimeo/psalm/issues/6854 and https://github.com/vimeo/psalm/issues/6849

Tests are not 100% real because this particular issue depends on php version used to run Psalm as much as version used to analyse but I tested this with 8.0 and 8.1RC5 on my machine and it works

Fixes #6854 
Fixes #6849

### DIFF
--- a/stubs/CoreGenericClasses.phpstub
+++ b/stubs/CoreGenericClasses.phpstub
@@ -494,3 +494,10 @@ final class WeakMap implements ArrayAccess, Countable, IteratorAggregate, Traver
      */
     public function offsetUnset($offset) {}
 }
+
+
+#[Attribute]
+final class ReturnTypeWillChange
+{
+    public function __construct() {}
+}
--- a/stubs/Php81.phpstub
+++ b/stubs/Php81.phpstub
@@ -4,12 +4,6 @@ namespace {
         /** @return non-empty-list<static> */
         public static function cases(): array;
     }
-
-    #[Attribute]
-    final class ReturnTypeWillChange
-    {
-        public function __construct() {}
-    }
 }
 
 namespace FTP {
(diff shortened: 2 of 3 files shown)


derrabus:
Thank you very much for working on this issue!

Issue #6868, opened 2021-11-09
The inferred return type 'null' for foo is more specific than the declared return type 'void'


iluuu1994:
https://psalm.dev/r/ac753d0bfc

orklah:
I have painful memories about this issue. the problem is that TypeCombiner (and another place for template) consider that void in combination with another type must be coerced to null. This is a sound assumption in a lot of case, except when the other types are `never` or `empty` (or possibly another `void`). In this case, the return type is inferred to `never|void` and automatically coerced to null

iluuu1994:
> consider that void in combination with another type must be coerced to null

I see. `never` in a union type has no effect and should likely be removed in the canonicalization process.

Issue #4501, opened 2020-11-06
Conditional no-return


orklah:
I have methods in my codebase where they can perform an exit() based on a parameter. For example, redirect:

https://psalm.dev/r/d980e72509

Could this be supported in conditional return?

muglug:
No, I won't add this – I _strongly_ encourage you instead to refactor your code with a `redirectAndExit` (or, alternatively, `redirectAndContinue`)

weirdan:
> No, I won't add this

@muglug , it's too late :rofl:. It works already: https://psalm.dev/r/a6ab3fc875

orklah:
Oh, so it works but I'm not allowed to have `: void` return hint! Not sure what I'll do then!

Thank you both for your answer anyway!

PR #6874, opened 2021-11-09
fix never combination

orklah:
This should fix https://github.com/vimeo/psalm/issues/6868

And it also fix https://github.com/vimeo/psalm/issues/4501


iluuu1994:
Awesome, thanks @orklah!

PR #6876, opened 2021-11-10
do not treat __halt_compiler() as an expression being executed

### LINKED ISSUE #6875: __halt_compiler() is treated as an expression
https://psalm.dev/r/e29d657d55

`__halt_compiler()` is not really an expression and should not be part of the `UnevaluatedCode` check.

This regressed somewhere between 4.9.3 and 4.12 (which I'm upgrading to ATM)

pilif:
this fixes #6875


orklah:
Thanks!

pilif:
when reading this patch again, it occurred to me that the check for `!($stmt instanceof PhpParser\Node\Stmt\InlineHTML)` that's also being done is probably faulty:

```php
<?php
exit(0);
?>
foo bar baz
```

I would expect this to be flagged as an issue because the text outside of the PHP tags is never emitted this way.

Would you agree? In that case, I'll create a followup issue and PR, though of course this might have some BC implications.

weirdan:
@pilif yeah, please do.

pilif:
See #6878

PR #6878, opened 2021-11-10
detect unevaluated inline-html code after an rearly return

### LINKED ISSUE #6877: psalm does not detect unevaluated inline HTML
this is a bit of an edge-case I came across after fixing #6875

https://psalm.dev/r/cb57223f5c

it could be argued that this should be reported as an error because the inline HTML code after the closing `?>` is never evaluated.

On the other hand, the reason this doesn't currently report an error is an [explicit check](https://github.com/vimeo/psalm/blob/f407ac846ae25f337c94bd4d2bb85fd837482555/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php#L346), so there might be a use-case where this is intentional?

pilif:
this is a followup after #6876

this fixes #6877

### DIFF
--- a/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
@@ -343,7 +343,6 @@ class StatementsAnalyzer extends SourceAnalyzer
             && !$context->collect_initializations
             && !$context->collect_mutations
             && !($stmt instanceof PhpParser\Node\Stmt\Nop)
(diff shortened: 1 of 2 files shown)


orklah:
It may have been done to avoid reporting spaces or line feed after closing PHP tag. I'm inclined to accept this, if someone complains, we'll at least have a use case we'll be able to add in test suite...

pilif:
the check was added explicitly in https://github.com/vimeo/psalm/commit/655039427c2b843d7961ee94b995ff2d76e5bcef, but the commit message doesn't give a reason, so we still don't know whether it was intentional

PR #6880, opened 2021-11-10
Fix --enable-autocomplete=false

### LINKED ISSUE #6879: Disabling autocomplete does not work
`src/Psalm/Internal/Cli/LanguageServer.php` checks for the `$options['enable-autocomplete']` to be a string with the text `'false'`, but it's actually always passed as a bool with value `false`, because it's declared as a long option flag. There is no way for this to accept a value.

The fix would be to declare the argument as having an optional value: `"enable-autocomplete::"`

Currently, no matter what, i cannot disable autocomplete

Nadyita:
Currently, `--enable-autocomplete` is defined as a bool option, so the it's either there or not, but it doesn't accept any arguments. This fix will allow `--enable-autocomplete=false` to work as expected, while still having the default of being enabled.

Closes #6879

### DIFF
--- a/src/Psalm/Internal/Cli/LanguageServer.php
+++ b/src/Psalm/Internal/Cli/LanguageServer.php
@@ -68,7 +68,7 @@ final class LanguageServer
             'tcp:',
             'tcp-server',
             'disable-on-change::',
-            'enable-autocomplete',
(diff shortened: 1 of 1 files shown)


orklah:
Thanks!

PR #6890, opened 2021-11-11
Add ImagickPixel::getColor return type provider

danog:


src/Psalm/Internal/Provider/ReturnTypeProvider/ImagickPixelColorReturnTypeProvider.php
```diff
@@ -0,0 +1,84 @@
...
+        $source = $event->getSource();
+        $call_args = $event->getCallArgs();
+        $method_name_lowercase = $event->getMethodNameLowercase();
+
+        if ($method_name_lowercase !== 'getcolor') {
+            return null;
+        }
+
+        if (!$source instanceof \Psalm\Internal\Analyzer\StatementsAnalyzer) {
+            return Type::getMixed();
```
orklah, line 31: You should return null instead to fallback on callmap type instead of returning Mixed

src/Psalm/Internal/Provider/ReturnTypeProvider/ImagickPixelColorReturnTypeProvider.php
```diff
@@ -0,0 +1,84 @@
...
+                new TKeyedArray([
+                    'r' => new Union([new TIntRange(0, 255)]),
+                    'g' => new Union([new TIntRange(0, 255)]),
+                    'b' => new Union([new TIntRange(0, 255)]),
+                    'a' => new Union([new TIntRange(0, 255)])
+                ])
+            ]);
+        }
+
+        /** @var non-empty-list<Psalm\Type\Union> $types */
```
orklah, line 81: why is that necessary? Type::combineUnionTypeArray should already return a non empty list of Unions

orklah (commented):
Great work <3, see a few comments

src/Psalm/Internal/Provider/ReturnTypeProvider/ImagickPixelColorReturnTypeProvider.php
danog: Welp sorry, copied this directly from the DomNodeAppendChild return type provider :)

src/Psalm/Internal/Provider/ReturnTypeProvider/ImagickPixelColorReturnTypeProvider.php
danog: Psalm is worried that `$types` is a possibly empty `list<Psalm\Type\Union>`, probably due to the loop assertion logic not generating proper assertions for the somewhat complex `$formats` loop+ifs

orklah: oh, misread what was being forced. Please use something like `assert($types !== []);` to check for non-emptiness

orklah:
Thanks! That's great!

Issue #6893, opened 2021-11-11
`ignoreInternalFunctionFalseReturn=false` does not trigger errors for (at least) `base64_decode`


bendavies:
Hi,

This example with `ignoreInternalFunctionFalseReturn="false"` does not trigger an error, but it should.
https://psalm.dev/r/d83a11b4b6

`base64_decode` can return false in this instance because `strict=true`

As an aside, the [return type](https://github.com/vimeo/psalm/blob/a2d6861f12a1c044d3e6c028cc9255b1a5f5fafb/stubs/CoreGenericFunctions.phpstub#L1259) is incorrect, and should be conditional base on `strict=true`, but that should not be the source of this issue.

orklah:
Ok, I get what's going on, stubbed function are not impacted by ignoreInternalFunctionFalseReturn, but rather individual `@psalm-ignore-falsable-return` added on the stub.

This shouldn't be the case and we should respect the config

orklah:
@weirdan I'm not sure we can make the difference between our internal stubs, user files and user stubs once they're in queue for scanning.
This means either we'd need slightly change the meaning of `@psalm-ignore-falsable-return` to be conditional on the config flag. This could impact user code and user stubs, and it would mean the config called `ignoreInternalFunctionFalseReturn` would now apply to not-internal functions too.

Is this something we can do? I'd hate to create a new config and new properties everywhere in code to differentiate internal stubs from code

weirdan:
There's a flag called `$register_stub_files` on the `Codebase` instance that is on when stubs are being scanned. It includes user/plugin stubs as far as I can see, but is off when normal files are scanned.

PR #6895, opened 2021-11-11
make ignore falsable/nullable dependent on the config

orklah:
This would fix https://github.com/vimeo/psalm/issues/6893 by slightly changing the meaning of our configs

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -552,6 +552,11 @@ class Config
      */
     public $trigger_error_exits = 'default';
 
+    /**
+     * @var string[]
+     */
+    public $internal_stubs = [];
+
     protected function __construct()
     {
         self::$instance = $this;
@@ -769,7 +774,9 @@ class Config
         ];
 
         $config->config_issues = [];
-        $attributes = $dom_document->getElementsByTagName('psalm')->item(0)->attributes;
+        $psalm_element_item = $dom_document->getElementsByTagName('psalm')->item(0);
+        assert($psalm_element_item !== null);
+        $attributes = $psalm_element_item->attributes;
         foreach ($attributes as $attribute) {
             if (in_array($attribute->name, $deprecated_attributes, true)) {
                 $line = $attribute->getLineNo();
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php
```diff
@@ -496,8 +498,11 @@ public static function parse(PhpParser\Comment\Doc $comment): FunctionDocblockCo
             $info->no_named_args = true;
         }
 
-        $info->ignore_nullable_return = isset($parsed_docblock->tags['psalm-ignore-nullable-return']);
-        $info->ignore_falsable_return = isset($parsed_docblock->tags['psalm-ignore-falsable-return']);
+        $config = Config::getInstance();
+        $info->ignore_nullable_return = isset($parsed_docblock->tags['psalm-ignore-nullable-return'])
+            && $config->ignore_internal_nullable_issues;
+        $info->ignore_falsable_return = isset($parsed_docblock->tags['psalm-ignore-falsable-return'])
+            && $config->ignore_internal_falsable_issues;
```
weirdan, line 505: I'm not sure it's the best place to have it. Could we perhaps move it to a method on function storage instead?

orklah: Not sure what you make reference to?

weirdan: The parser generates in-memory representation of a docblock. It feels kinda wrong to vary it based on external factors other than the docblock itself.

There's already an ugly hack in `FunctionLikeNodeScanner` that sets parameter flags for functions coming from internal stubs:
https://github.com/vimeo/psalm/blob/85f2083acb7dcb09d636baae3912cdb6bb7e9c2d/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php#L207-L213

Could we perhaps tidy it up and reuse? Scanner does have a reference to codebase so it can access config without resorting to `Config::getInstance()`.

orklah: I listed all the internals stubs in an array on config, that way I can check where the stub is coming from and I don't even need the flag. WDYT?

weirdan: Now that looks much better!

weirdan (approved):

src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php
orklah: I'd be inclined to go one step further and not require the `@psalm-ignore-nullable-return` at all in the stub and adding it automatically where we see false or null coming from the stubs return type. I think that's already what's being done in Callmap.

It would prevent forgetting to add this tag everywhere in the stubs.

You think it's a good idea?

weirdan: Say we have a function in the stub that shouldn't have its nullable return ignored. How would we opt out of that automatic flagging?

orklah: I don't remember we ever had to create a stub to avoid that with CallMap so I guess it doesn't happen. If it did anyway, we would have to make a list of those function somewhere and make it an exception to the code I'll do that will make that will detect null and false and put the flag when needed

weirdan: Ok

orklah: Mmh, for CallMap it seems handled in FunctionCallReturnTypeFetcher and it's more subtle than I though. I'll stay like this for now

PR #6898, opened 2021-11-12
Php version consistency

orklah:
This brings consistency to the way we use php versions. Before that, sometimes we passed them under the X.Y format, sometimes as two int params and sometimes as an array of two ints...

It also get rid of those horrors:
```
if ($codebase->php_major_version < 8 ||
   ($codebase->php_major_version === 8 && $codebase->php_minor_version < 1)
) {
```

and it fixes a couple of bugs like 
`if (\PHP_VERSION_ID < 80100 && $codebase->php_major_version >= 8 && $codebase->php_minor_version >= 1) {`
that would have been a problem in version 9.0

However, it's probably a BC break for plugin users...

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -77,6 +77,7 @@ use function error_log;
 use function explode;
 use function implode;
 use function in_array;
+use function intdiv;
 use function is_numeric;
 use function is_string;
 use function krsort;
@@ -90,8 +91,6 @@ use function strtolower;
 use function substr;
 use function substr_count;
 
-use const PHP_MAJOR_VERSION;
-use const PHP_MINOR_VERSION;
 use const PHP_VERSION_ID;
 
 class Codebase
@@ -304,18 +303,6 @@ class Codebase
      * @var bool
      */
     public $allow_backwards_incompatible_changes = true;
-
-    /**
-     * @var int
-     * @deprecated Removed in Psalm 5, use Codebase::$analysis_php_version_id
-     */
-    public $php_major_version = PHP_MAJOR_VERSION;
-
-    /**
-     * @var int
-     * @deprecated Removed in Psalm 5, use Codebase::$analysis_php_version_id
-     */
-    public $php_minor_version = PHP_MINOR_VERSION;
 
     /** @var int */
     public $analysis_php_version_id = PHP_VERSION_ID;
@@ -534,7 +521,7 @@ class Codebase
     {
         return $this->statements_provider->getStatementsForFile(
             $file_path,
-            $this->php_major_version . '.' . $this->php_minor_version,
+            $this->analysis_php_version_id,
             $this->progress
         );
     }
@@ -2009,5 +1996,20 @@ class Codebase
         );
 
         $this->taint_flow_graph->addSink($sink);
+    }
+
+    public function getMinorAnalysisPhpVersion(): int
+    {
+        return self::transformPhpVersionId($this->analysis_php_version_id % 10000, 100);
+    }
+
+    public function getMajorAnalysisPhpVersion(): int
+    {
+        return self::transformPhpVersionId($this->analysis_php_version_id, 10000);
+    }
+
+    public static function transformPhpVersionId(int $php_version_id, int $div): int
+    {
+        return intdiv($php_version_id, $div);
     }
 }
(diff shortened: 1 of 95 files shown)


src/Psalm/Type.php
```diff
@@ -68,14 +68,14 @@ abstract class Type
      */
     public static function parseString(
         string $type_string,
-        ?array $php_version = null,
+        ?int $analysis_php_version_id = null,
```
weirdan, line 71: We could accept both forms here, and transform the array to integer to avoid the BC break.

src/Psalm/Type/Atomic.php
```diff
@@ -91,15 +91,15 @@ abstract class Atomic implements TypeNode
...
+     * @param int $analysis_php_version_id contains php version when the type comes from signature
+     * @param array<string, array<string, Union>> $template_type_map
+     * @param array<string, TypeAlias> $type_aliases
      */
     public static function create(
         string $value,
-        ?array $php_version = null,
-        array $template_type_map = [],
-        array $type_aliases = []
+        ?int   $analysis_php_version_id = null,
```
weirdan, line 100: ditto

src/Psalm/Type/Atomic.php
```diff
@@ -602,11 +590,10 @@ abstract public function toPhpString(
         ?string $namespace,
         array $aliased_classes,
         ?string $this_class,
-        int $php_major_version,
-        int $php_minor_version
+        int $analysis_php_version_id
     ): ?string;
 
-    abstract public function canBeFullyExpressedInPhp(int $php_major_version, int $php_minor_version): bool;
+    abstract public function canBeFullyExpressedInPhp(int $analysis_php_version_id): bool;
```
weirdan, line 596: This could possibly be changed to `(int $major_or_id, ?int $minor = null)` and the method could interpret first parameter as version id when second is null. The signature would be quite ugly though, but it'll avoid the BC break.

src/Psalm/Type/Atomic/CallableTrait.php
```diff
@@ -136,8 +136,7 @@ public function toPhpString(
         ?string $namespace,
         array $aliased_classes,
         ?string $this_class,
-        int $php_major_version,
-        int $php_minor_version
+        int $analysis_php_version_id
```
weirdan, line 139: ditto

src/Psalm/Internal/Stubs/Generator/StubsGenerator.php
```diff
@@ -300,7 +300,7 @@ public static function getParserTypeFromPsalmType(Type\Union $type): ?PhpParser\
                 || $atomic_type instanceof Type\Atomic\TArray
                 || $atomic_type instanceof Type\Atomic\TIterable
             ) {
-                $identifier_string = $atomic_type->toPhpString(null, [], null, 8, 0);
+                $identifier_string = $atomic_type->toPhpString(null, [], null, 8);
```
weirdan, line 303: this should be `80000` I believe

src/Psalm/Codebase.php
```diff
@@ -271,12 +270,7 @@ class Codebase
     /**
      * @var int
      */
-    public $php_major_version = PHP_MAJOR_VERSION;
-
-    /**
-     * @var int
-     */
-    public $php_minor_version = PHP_MINOR_VERSION;
+    public $analysis_php_version_id = PHP_VERSION_ID;
```
weirdan, line 273: It's possible to keep and deprecate old properties.

weirdan (changes requested):
BC breaks look avoidable. Conversions from version ID to major/minor need to be fixed, and it would be better to move conversions into its own helper class.

orklah:
I did not pursue this as a change in current version because I really hate the change in order to avoid the BC break. This will probably be something only in Psalm 5...

weirdan:
> This will probably be something only in Psalm 5...

I added the Psalm 5 milestone we can use for such PRs / issues.

weirdan (approved):

weirdan:
@orklah mind giving this a once-over? I rebased it, removed deprecated properties and documented BC breaks (in `UPGRADING.md`).

orklah:
I'll take a look tonight

orklah:
That seems correct. Thanks for the documentation and the rebase <3

PR #6899, opened 2021-11-12
don't stop processing for class usage after raising an issue

### LINKED ISSUE #6887: UneavaluatedCode issue causes further processing to stop for detecting class usage
https://psalm.dev/r/1f6cbc8ca8

the class `Foo` in the example is detected as used once the erroneous line 5 is fixed.

IMHO further scanning of the file should not stop at that error because this will lead to some very confusing "Unused*" errors.

I noticed this when upgrading from psalm 4.9.3 to 4.12 where the `UnevaluatedCode` inspection got *much better* (especially with regards to methods flagged `no-return`, but now there are countless instances where psalm flags a method as unused making it very hard to systematically fix newly detected issues.

pilif:
this fixes #6887

### DIFF
--- a/tests/UnusedCodeTest.php
+++ b/tests/UnusedCodeTest.php
@@ -5,7 +5,12 @@ use Psalm\Config;
 use Psalm\Context;
 use Psalm\Internal\Provider\FakeFileProvider;
 use Psalm\Internal\RuntimeCaches;
+use Psalm\IssueBuffer;
 use Psalm\Tests\Internal\Provider;
+
+use function getcwd;
+
+use const DIRECTORY_SEPARATOR;
 
 class UnusedCodeTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


tests/UnusedCodeTest.php
```diff
@@ -99,6 +104,70 @@ public function testInvalidCode($code, $error_message, $error_levels = []): void
         \Psalm\IssueBuffer::processUnusedSuppressions($this->project_analyzer->getCodebase()->file_provider);
     }
 
+    public function testSeesClassesUsedAfterUnevaluatedCodeIssue(): void
+    {
+        $this->project_analyzer->getConfig()->throw_exception = false;
+        $file_path = getcwd() . DIRECTORY_SEPARATOR . 'tests' . DIRECTORY_SEPARATOR . 'somefile.php';
```
pilif, line 110: btw: the other tests in this file use a broken `$file_path` and only work by accident because the project analyzer is configured to throw rather than record the issue.

the other tests use `self::$src_dir_path . 'somefile.php';` which points to the `src ` dir which isn't in the project directory used by the test `project_analyzer` (which is `getcwd() . '/tests'` which causes it to not record any issues.

This took me a while to find, especially because everything is nicely failing silently along the path once you tell the config to not throw exceptions.

orklah:
Great! Thanks!

Issue #6892, opened 2021-11-11
deprecate forbid_echo


orklah:
This is a config that exists in order to emit a ForbiddenEcho issue when using the echo statement.

Now we have ForbiddenCode that checks echo, exit, backticks etc... I believe we could remove this config and Issue for consistency

PR #6902, opened 2021-11-12
deprecate forbidEcho

orklah:
This will fix https://github.com/vimeo/psalm/issues/6892


weirdan (approved):

orklah:
This is failing on PSL 1.9 because they use this config.
PSL 2 stopped using it so it would require retrieving this branch from the main repo and switching the tests on it.

This is not urgent though, so it'll wait for now

PR #6909, opened 2021-11-13
Fix promoted readonly properties

trowski:
This fixes detection of readonly on promoted properties as well as the visibility of promoted readonly properties.

https://psalm.dev/r/0df6074053

https://psalm.dev/r/910eec3285


orklah:
Thanks! That's great!

PR #6910, opened 2021-11-13
Fix mb_convert_encoding to accept any array.

### LINKED ISSUE #6886: Various problems with mb_convert_encoding
https://psalm.dev/r/fb4e03dccd

# :heavy_check_mark: Case 1 – ``string`` input

> INFO: Trace - $case1: false|string

Return type ``false|string`` – everthing as expected, ok.

# :x: Case 2 – ``array<int, string>`` input

> INFO: Trace - $case2: array<int, string>

Return type ``array<int, string>`` – don’t know how to force some error to test and didn’t found out by looking at the PHP source code, but I suspect it also returns ``false`` in case of an error so the correct return type would probably be ``false|array<int, string>`` in that case.

# :x: Case 3 – ``array<string, string>`` input

> ERROR: InvalidCast - array{idx: "abc"} cannot be cast to string
> ERROR: InvalidArgument - Argument 1 of mb_convert_encoding expects string, array{idx: "abc"} provided
> INFO: Trace - 21:2 - $case3: false|string

Psalm seems to expect ``string|array<int, string>`` as type for the first function parameter. A short test shows the functions also works with ``array<string, string>`` as input:

```php
var_dump(mb_convert_encoding(['idx1' => 'abc', 'idx2' => 'abc'], 'ASCII'));
var_dump(mb_convert_encoding([123 => 'abc', 456 => 'abc'], 'ASCII'));
```
results in
```
array(2) {'idx1' => string(3) "abc", 'idx2' => string(3) "abc"}
array(2) {[123] => string(3) "abc", [456] => string(3) "abc"}
```
so array indexes won’t be touched by php.

# :x: Case 4 – ``string|array`` input

> INFO: MixedArgumentTypeCoercion - Argument 1 of mb_convert_encoding expects array<int, string>, parent type array<array-key, mixed>|string provided
> INFO: Trace - $case4: array<int, string>

Return type should be something like ``array<array-key, mixed>|string|false`` instead.

rarila:
Fixes #6886

### DIFF
--- a/dictionaries/CallMap_72_delta.php
+++ b/dictionaries/CallMap_72_delta.php
@@ -7,7 +7,7 @@
  * The 'added' section contains function/method names from FunctionSignatureMap (And alternates, if applicable) that do not exist in php 7.1
  * The 'removed' section contains the signatures that were removed in php 7.2.
  * The 'changed' section contains functions for which the signature has changed for php 7.2.
- *     Each function in the 'changed' section has an 'old' and a 'new' section, 
+ *     Each function in the 'changed' section has an 'old' and a 'new' section,
  *     representing the function as it was in PHP 7.1 and in PHP 7.2, respectively
  *
  * @see CallMap.php
@@ -34,6 +34,7 @@ return [
     'ldap_exop_whoami' => ['string|false', 'ldap'=>'resource'],
     'ldap_parse_exop' => ['bool', 'ldap'=>'resource', 'result'=>'resource', '&w_response_data='=>'string', '&w_response_oid='=>'string'],
     'mb_chr' => ['string|false', 'codepoint'=>'int', 'encoding='=>'string'],
+    'mb_convert_encoding\'1' => ['array', 'string'=>'array', 'to_encoding'=>'string', 'from_encoding='=>'mixed'],
     'mb_ord' => ['int|false', 'string'=>'string', 'encoding='=>'string'],
     'mb_scrub' => ['string', 'string'=>'string', 'encoding='=>'string'],
     'oci_register_taf_callback' => ['bool', 'connection'=>'resource', 'callback='=>'callable'],
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1306,5 +1306,6 @@ function is_a($object_or_class, string $class, $allow_string = false): bool{}
  * @template T of array|string
  * @param T $string
  * @return (T is array<int, string> ? array<int, string> : T is array ? array : string|false)
+ * @psalm-ignore-falsable-return
  */
 function mb_convert_encoding(array|string $string, string $to_encoding, array|string|null $from_encoding = null): array|string|false{}
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1305,6 +1305,6 @@ function is_a($object_or_class, string $class, $allow_string = false): bool{}
 /**
  * @template T of array|string
  * @param T $string
- * @return (T is array<int, string> ? array<int, string> : T is array ? array : string)
+ * @return (T is array<int, string> ? array<int, string> : T is array ? array : string|false)
  */
 function mb_convert_encoding(array|string $string, string $to_encoding, array|string|null $from_encoding = null): array|string|false{}
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1301,3 +1301,10 @@ function get_defined_constants(bool $categorize = false): array {}
  * @return ($allow_string is false ? ($object_or_class is object ? bool : false) : bool)
  */
 function is_a($object_or_class, string $class, $allow_string = false): bool{}
+
+/**
+ * @template T of array|string
+ * @param T $string
+ * @return (T is array<int, string> ? array<int, string> : T is array ? array : string)
+ */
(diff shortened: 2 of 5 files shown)


orklah:
Thanks! Do you know if the shape of the array is preserved?

If it is, we could create a stub in order to make the array shape pass through the function.

something like 
```
/**
 * @template T of array|string
 * @param T $string
 * @return (T is array ? T : string)
 */
```
should do the trick

rarila:
Running tests locally at the moment…
.............................................................  610 / 6499 (  9%)
two hours later… :-))

Hmm, no, it isn’t. Look at the last example in the issue. String keys are translated, objects/recursively referenced values are replaced.

So I guess, no

weirdan:
>  two hours later… :-))

Tests shouldn't be that slow, even with Xdebug enabled (I'd suggest to disable it with, e.g. https://packagist.org/packages/weirdan/run-without-xdebug though).

orklah:
> String keys are translated, objects/recursively referenced values are replaced.

Yeah, string are transformed, but the general shape of the array is kept, no? I mean, if the array was `array<int, string>` it will stay a `array<int, string>` right?

weirdan:
> I mean, if the array was array<int, string> it will stay a array<int, string> right?

Yeah, but `array{€:string}` does not stay that way: https://3v4l.org/djRjL, and you wouldn't be able to express that with templates.

rarila:
@weirdan Yep.

weirdan:
> Tests shouldn't be that slow

It takes about 5 minutes here, using `php-noxdebug vendor/bin/paratest` command.

rarila:
@weirdan gonna try, but I’m already at 44% now – almost done :-)

orklah:
Oh yeah, forgot about literals in array-shapes.

Guess it would need a return type provider in order to recursively strip any string literal.

It may be worth it one day, returning `array` is a sure way to trigger a lot of MixedAssignment just by using the returned array

rarila:
Probably 

/**
 * @template T of array|string
 * @param T $string
 * @return (T is array<int, string> ? array<int, string> : 
 *                T is array                    ? array : 
 *                                                    string)
 */

Would help for most cases

orklah:
Yeah, it would, Mind adding that to the stubs?

rarila:
@orklah ok

orklah:
Note that you have to make a change to CallMap_80 and possibly CallMap_historical for your PR to pass the CI. More is explained here: https://psalm.dev/docs/contributing/editing_callmaps/

rarila:
@orklah CoreGenericFunctions.phpstub? Just add it at the end?

rarila:
@orklah 

How do you like this format:

```
/**
 * @template T of array|string
 * @param T $string
 * @return (T is array<int, string> ? array<int, string> : 
 *          T is array              ? array : 
 *          T is string             ? string :
 *                                    never)
 */
```
I like it that way because it lists all possible options and is quite readable that way.

orklah:
Please don't add never at the end. Listing the allowed params is enough to make Psalm emit an issue. Adding never will just create a whole new set of weird errors that will be irrelevant

rarila:
@orklah ok.

Looking at the other callmaps now…

If I see it right, I'll have to remove the array version from CallMap_historical.php, add it to 'added' CallMap_72_delta.php?!

orklah:
nah, you have to change the signature in CallMap_80 "new" to match with the one in your CallMap file and then, if needed, change the CallMap_historical to match with the CallMap_80 "old" (we start from CallMap and then process each delta file from newer to older untli CallMap_historical)

rarila:
I think I have it, running tests…

dictionaries/CallMap_72_delta.php
```diff
@@ -34,6 +34,7 @@
     'ldap_exop_whoami' => ['string|false', 'ldap'=>'resource'],
     'ldap_parse_exop' => ['bool', 'ldap'=>'resource', 'result'=>'resource', '&w_response_data='=>'string', '&w_response_oid='=>'string'],
     'mb_chr' => ['string|false', 'codepoint'=>'int', 'encoding='=>'string'],
+    'mb_convert_encoding\'1' => ['array', 'string'=>'array', 'to_encoding'=>'string', 'from_encoding='=>'mixed'],
```
orklah, line 37: Oh, nice catch! I didn't know arrays were introduced in 7.2

rarila: At least the manual says so :-)

stubs/CoreGenericFunctions.phpstub
```diff
@@ -1300,4 +1300,11 @@ function get_defined_constants(bool $categorize = false): array {}
...
  * @return ($allow_string is false ? ($object_or_class is object ? bool : false) : bool)
  */
-function is_a($object_or_class, string $class, $allow_string = false): bool{}
\ No newline at end of file
+function is_a($object_or_class, string $class, $allow_string = false): bool{}
+
+/**
+ * @template T of array|string
+ * @param T $string
+ * @return (T is array<int, string> ? array<int, string> : T is array ? array : string)
```
orklah, line 1308: we should add |false everywhere in the return types

rarila: Ah, good catch. But only at the string return.

I still wonder what happens, when you add a string that triggers returning false for a string into an array…

orklah:
Can you add a `@psalm-ignore-falsable-return` on the stub? It will allow people with the config disabled to not handle the false return without getting errors

rarila:
@orklah Here you are :-)

orklah:
Thank you! That's great!

rarila:
No, I have to thank for this great tool!

PR #6920, opened 2021-11-15
Multiple variables in @psalm-trace

rarila:
Psalm did allow multiple variables in ``@psalm-trace`` but failed to add multiple trace issues to the same place.

That way, both cases in this example fail:

https://psalm.dev/r/31262a113f

Additional to fixing this bug, I’ve added the ability to use a comma as separator, so this line is valid also:
```
/** @psalm-trace $a, $b */
```

This is probably more natural for some people.

*Caveat:*

- I don’t know if my fix is the best way to allow multiple traces or if changing the format of ``$emitted_key`` for that case breaks something elsewhere. But no included test fails (at least none related to my changes) and using this codebase in another project worked as expected.
- The additional tests test only half of the addition. I don’t know if is possible to expect multiple errors in the test framework. At least the test asserts that adding two variables doesn’t break getting the first.

### DIFF
--- a/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php
@@ -382,7 +382,7 @@ class StatementsAnalyzer extends SourceAnalyzer
 
             if (isset($statements_analyzer->parsed_docblock->tags['psalm-trace'])) {
                 foreach ($statements_analyzer->parsed_docblock->tags['psalm-trace'] as $traced_variable_line) {
-                    $possible_traced_variable_names = preg_split('/[\s]+/', $traced_variable_line);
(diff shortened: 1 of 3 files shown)


weirdan (approved):
:+1:

weirdan:
> if changing the format of $emitted_key for that case breaks something elsewhere

I think that's unlikely.

> I don’t know if is possible to expect multiple errors in the test framework

Not with `Valid/InvalidCodeTrait`, as those tests rely on Psalm throwing exceptions instead of storing all issues reported. It's possible to arrange a test emitting several issues like it's done in, e.g., `ReportOutputTest`.

weirdan:
Thanks!

pilif:
For checking for multiple errors, see also https://github.com/vimeo/psalm/pull/6899 where I had to do that last week

Issue #6471, opened 2021-09-10
Enum: cases are constants


weirdan:
Psalm should
- [x] flag clashes between constants and cases: https://psalm.dev/r/30e59100be?php=8.1 , https://3v4l.org/fKOs6/rfc#vgit.master
- [x] include cases into const wildcards: https://psalm.dev/r/838acc5287?php=8.1
- [x] Support enum references in class constants: https://psalm.dev/r/d0b73490ba?php=8.1, https://3v4l.org/VlWW3/rfc#vgit.master

PR #6922, opened 2021-11-15
Allow references to enum cases in class constants

weirdan:
Refs #6471


tests/EnumTest.php
```diff
@@ -154,6 +154,26 @@ public function get(): string
                 [],
                 '8.1'
             ],
+            'SKIPPED-wildcardEnum' => [
```
orklah, line 157: skipped?

weirdan: Yeah, it doesn't work yet.

PR #6927, opened 2021-11-15
Invalid SoapClient::__doRequest() response type

### LINKED ISSUE #6901: Invalid SoapClient::__doRequest() signature
The stubs indicate that the `$one_way` variable is an integer. 
Since PHP 8.0 this is a bool.

https://github.com/vimeo/psalm/blob/2d0758ab104b615786a85bda83894b6d26a29266/stubs/soap.phpstub#L261-L268

See https://www.php.net/manual/en/soapclient.dorequest.php

> 8.0.0 | The type of oneWay is bool now; formerly it was int.

veewee:
Fixes #6901

The return type of SoapClient::__doRequest() is actually `?string` instead of `string`.
See

* https://www.php.net/manual/en/soapclient.dorequest.php
* https://github.com/php-soap/ext-soap-engine/issues/3


orklah:
Thanks! Sorry for the delay, I forgot about this PR...

Issue #6928, opened 2021-11-15
Parser fails on type alias of callable with generic array parameter


spawnia:
Hi 👋 

I found that Psalm fails to parse a very specific combination: a type alias that is a callable with a parameter that is a generic array.

```php
@psalm-type TypedArrayHandler callable(array<string>): void
```

https://psalm.dev/r/2a189c7ff9

orklah:
Indeed. You can use this workaround it seems: https://psalm.dev/r/d05cba9b24

spawnia:
Thanks for the feedback. This problem was triggered through a dependency, so I can not change their code.

Is there anything I can do to help with getting this fixed?

orklah:
I have very little experience with Psalm's Parser, but if you want to get your hand in it, the error comes from here: https://github.com/vimeo/psalm/blob/b29737f2b154748430250effdbcf47de8a5904df/src/Psalm/Internal/Type/ParseTreeCreator.php#L143

The parser is trying to create a tree of tokens and it seems it didn't expect what it got. I guess the first step would be to check how it's doing that, what exactly makes it emitting this issue and adapt code until it doesn't :)

spawnia:
Here is a PR with a failing test case: https://github.com/vimeo/psalm/pull/6929

I think the parser is probably fine, `ClassLikeNodeScanner::getTypeAliasesFromComment()` appears to be the culprit. The code in there seems quite complicated, lots of `preg_replace` and the likes going on.

MidnightDesign:
Here's another workaround. Put the generic parameter in a separate type:

Before:
```
/**
 * @psalm-type Resolve callable(string, list<TypeInterface>): TypeInterface
 */
```
After:
```
/**
 * @psalm-type ListOfTypes list<TypeInterface>
 * @psalm-type Resolve callable(string, ListOfTypes): TypeInterface
 */
```

ramsey:
I've encountered this same issue:

https://psalm.dev/r/0f1355a445

I'm able to use the work-around.

weirdan:
No longer reproducible:

* https://psalm.dev/r/0f1355a445
* https://psalm.dev/r/2a189c7ff9

PR #6929, opened 2021-11-15
Add failing test case for parsing a complex type alias

spawnia:
https://github.com/vimeo/psalm/issues/6928

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
@@ -1746,9 +1746,6 @@ class ClassLikeNodeScanner
 
             $type_string = str_replace("\n", '', implode('', $var_line_parts));
 
-            $type_string = preg_replace('/>[^>^\}]*$/', '>', $type_string);
-            $type_string = preg_replace('/\}[^>^\}]*$/', '}', $type_string);
-
             try {
                 $type_tokens = TypeTokenizer::getFullyQualifiedTokens(
                     $type_string,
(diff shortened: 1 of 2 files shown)


orklah:
The error is caused by a preg_replace 
https://github.com/vimeo/psalm/commit/a0866da88e605adaa648e9f1ad37f4990650f55c that was added a few years back to skip extra datas at the end of a line. when there is an `array<...>`, this preg_replace matches then `>` and strip what's after, which in case of a `callable(array<string>): void` is pretty bad.

So this is the regex that needs to change. I'd advise dropping the whole preg_replace and see what fails. We'll then see what was the exact use case that needed this

spawnia:
There should be facilities to handle trailing data after a type definition in PHPDocs, for instance when dealing with `@return`. Perhaps there is a common core that can be extracted and reused?

orklah:
I'm aware of `CommentAnalyzer::sanitizeDocblockType`, not sure if it will work though...

spawnia:
Looks like your idea worked though, it looks like the removal of the offending `preg_replace` did indeed cause a regression in the following test case:

```php
<?php
/**
 * @psalm-type MiddlewareArray = array<
 *     class-string<\Exception>,
 *     array<int, string>
 * >
 *
 * @psalm-type RuleArray = array{
 *     rule: string,
 *     controller?: class-string<\Exception>,
 *     redirect?: string,
 *     code?: int,
 *     type?: string,
 *     middleware?: MiddlewareArray
 * }
 *
 * Foo Bar
 */
class A {}
```

If you have further ideas on how to fix this, feel free to work on my branch and edit it directly.

orklah:
So this is what the regex was needed for. I'd suggest you check what output is coming in the variable that was changed and find some other way to strip the extra datas. Maybe the function I gave you above?

spawnia:
I don't plan to do any more work on this pull request. Feel free to use the test case as the basis for an actual fix.

PR #6932, opened 2021-11-16
session_set_cookie_params signature changed in 8.0

kamil-tekiela:
See https://psalm.dev/r/ebba1bda24 and https://www.php.net/manual/en/function.session-set-cookie-params.php


weirdan:
You will also need to update `CallMap.php` (see https://psalm.dev/docs/contributing/editing_callmaps/ for details).

kamil-tekiela:
Is it correct now?

weirdan:
> Is it correct now?

According to the CI it is.

weirdan:
Thanks!

PR #6946, opened 2021-11-19
expand type aliases when comparing unions

### LINKED ISSUE #6440: Type alias imported via @psalm-import-type cannot be used for constrained template type
Hello,

despite #5205 (it seems) imported types aliases cannot be used for constrained template types as in the following example:
```php
/**
 * @psalm-type Type = string
 */
interface Types {}

/**
 * @template T of string
 */
interface ExampleInterface {}

/**
 * @psalm-import-type Type from Types
 * @implements ExampleInterface<Type>
 */
class ExampleImplementation implements ExampleInterface {}
```
See [https://psalm.dev/r/a85ed8c3d0](https://psalm.dev/r/a85ed8c3d0) for the produced error: `ERROR: InvalidTemplateParam - 18:40 - Extended template param T expects type string, type type-alias(Types::Type) given`.

On the other hand no error arises in case of non-constrained template types - see [https://psalm.dev/r/c9a4ecb87a](https://psalm.dev/r/c9a4ecb87a).

sebkehr:
Fixes #6440.

With the proposed changes type aliases are expanded when comparing union types.

This should allow for type aliases being passed as constrained type parameters.

### DIFF
--- a/src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php
@@ -13,6 +13,7 @@ use Psalm\Type\Atomic\TMixed;
 use Psalm\Type\Atomic\TNull;
 use Psalm\Type\Atomic\TNumeric;
 use Psalm\Type\Atomic\TTemplateParam;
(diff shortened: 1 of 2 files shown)


weirdan:
Thanks!

PR #6949, opened 2021-11-20
Correct missing key-value for stream_get_meta_data function

### LINKED ISSUE #6947: Missing key-value for stream_get_meta_data function
When [`stream_get_meta_data`](https://www.php.net/manual/en/function.stream-get-meta-data.php) is passed an SSL stream it will also have a `crypto` key on it.
The shape in that case should look like this:
```
array(8) {
  'crypto' =>
  array(4) {
    'protocol' =>
    string(7) "TLSv1.2"
    'cipher_name' =>
    string(29) "ECDHE-ECDSA-AES256-GCM-SHA384"
    'cipher_bits' =>
    int(256)
    'cipher_version' =>
    string(7) "TLSv1.2"
  }
  'timed_out' =>
  bool(false)
  'blocked' =>
  bool(true)
  'eof' =>
  bool(false)
  'stream_type' =>
  string(14) "tcp_socket/ssl"
  'mode' =>
  string(2) "r+"
  'unread_bytes' =>
  int(0)
  'seekable' =>
  bool(false)
}
```

For reproduction see here: https://psalm.dev/r/9c3df127fc

EDIT: To my surprise I also do not see this documented in PHP docs. However it is included in the stubs used by Phpstorm.

mallardduck:
This PR addresses the issue I reported here: https://github.com/vimeo/psalm/issues/6947

Fixes #6947

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -13785,7 +13785,7 @@ return [
 'stream_get_contents' => ['string|false', 'stream'=>'resource', 'length='=>'int', 'offset='=>'int'],
 'stream_get_filters' => ['array'],
 'stream_get_line' => ['string|false', 'stream'=>'resource', 'length'=>'int', 'ending='=>'string'],
-'stream_get_meta_data' => ['array{timed_out:bool,blocked:bool,eof:bool,unread_bytes:int,stream_type:string,wrapper_type:string,wrapper_data:mixed,mode:string,seekable:bool,uri:string,mediatype:string}', 'stream'=>'resource'],
+'stream_get_meta_data' => ['array{timed_out:bool,blocked:bool,eof:bool,unread_bytes:int,stream_type:string,wrapper_type:string,wrapper_data:mixed,mode:string,seekable:bool,uri:string,mediatype:string,crypto?:mixed}', 'stream'=>'resource'],
 'stream_get_transports' => ['list<string>'],
 'stream_get_wrappers' => ['list<string>'],
 'stream_is_local' => ['bool', 'stream'=>'resource|string'],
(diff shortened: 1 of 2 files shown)


dictionaries/CallMap.php
```diff
@@ -13785,7 +13785,7 @@
 'stream_get_contents' => ['string|false', 'stream'=>'resource', 'length='=>'int', 'offset='=>'int'],
 'stream_get_filters' => ['array'],
 'stream_get_line' => ['string|false', 'stream'=>'resource', 'length'=>'int', 'ending='=>'string'],
-'stream_get_meta_data' => ['array{timed_out:bool,blocked:bool,eof:bool,unread_bytes:int,stream_type:string,wrapper_type:string,wrapper_data:mixed,mode:string,seekable:bool,uri:string,mediatype:string}', 'stream'=>'resource'],
+'stream_get_meta_data' => ['array{timed_out:bool,blocked:bool,eof:bool,unread_bytes:int,stream_type:string,wrapper_type:string,wrapper_data:mixed,mode:string,seekable:bool,uri:string,mediatype:string,crypto?:mixed}', 'stream'=>'resource'],
```
mallardduck, line 13788: Does psalm support the same logic internally for object-like arrays? Just want to be certain I can indicate this is a contextual return item `crypto?:mixed`.

weirdan: The same as what? If you're asking if you can have nested array shapes, then yes, you can.

Like this:
```
... seekable: bool, uri:string, mediatype:string, crypto?: array{protocol:string, cipher_name: string ... and so on}...
```

weirdan (approved):

weirdan:
Thanks!

Issue #6941, opened 2021-11-18
Static methods declared via `@method static ReturnType methodName()` do not inherit `__callStatic()` purity


Ocramius:
Given following example ( https://psalm.dev/r/c076bbc487 ):

```php
<?php

/**
 * @method self FOO()
 * @method static BAR()
 * @method static BAZ()
 *
 * @psalm-immutable
 */
class MyEnum
{
    const FOO = 'foo';
    const BAR = 'bar';
    const BAZ = 'baz';
    
    /** @psalm-pure */
    public static function __callStatic(string $name, array $params): static
    {
        throw new BadMethodCallException('not implemented');
    }
}

/** @psalm-pure */
function gimmeFoo(): MyEnum
{
    return MyEnum::FOO();
}
```

No errors are produced.

As soon as `@method static self FOO()` is declared (to make it clear that the dynamic method is only available when called statically), we get an error ( https://psalm.dev/r/327af3563f ):

```diff
<?php

/**
- * @method self FOO()
+ * @method static self FOO()
- * @method static BAR()
+ * @method static static BAR()
- * @method static BAZ()
+ * @method static static BAZ()
 *
 * @psalm-immutable
 */
class MyEnum
{
    const FOO = 'foo';
    const BAR = 'bar';
    const BAZ = 'baz';
    
    /** @psalm-pure */
    public static function __callStatic(string $name, array $params): static
    {
        throw new BadMethodCallException('not implemented');
    }
}

/** @psalm-pure */
function gimmeFoo(): MyEnum
{
    return MyEnum::FOO();
}
```

```
Psalm output (using commit 6097e02): 

ERROR: ImpureMethodCall - 26:20 - Cannot call an impure method from a pure context
```

Therefore, I'm wondering if:

 1. `vimeo/psalm` at `4.12.0` removed `__callStatic()` purity from `@method` definitions (worked in `4.10.0`
 2. I'm misusing `@method` by declaring the `static` modifier? :thinking:

orklah:
Probably related to https://github.com/vimeo/psalm/pull/6724, I'll check this later

Issue #3180, opened 2020-04-18
Purity of @methods in phpdocs


vudaltsov:
Psalm currently considers `@method static` to be pure by default. Is it on purpose? Seems to be an error.
Also I think that if `__callStatic` is `@pure`, then all `@method static` should be considered pure.
Same for dynamic methods.

https://psalm.dev/r/6da6503a47

muglug:
> Psalm currently considers @method static to be pure by default. Is it on purpose? Seems to be an error.

Not on purpose

> Also I think that if __callStatic is @pure, then all @method static should be considered pure.
Same for dynamic methods.

I think that's probably sane? But it's also a weird area where there's no obvious correct solution.

vudaltsov:
I personally don't care much about `@method`, I usually have real methods in my code :)

But today I was fixing an issue in `webmozart/assert` and decided to check why for example [`Assert:: nullOrString`](https://github.com/webmozart/assert/blob/master/src/Assert.php#L30) never throws an `ImpureMethodCall`.

zerkms:
@vudaltsov there is a relevant https://github.com/webmozart/assert/pull/126 that I'm trying to get done (not as fast as me and anyone else wanted, but my work + family require my attention too, so my apologies for that :-) )

PR #6953, opened 2021-11-21
check __callStatic purity instead of the pseudoMethod purity

orklah:
In https://github.com/vimeo/psalm/issues/3180, users reported that pseudo methods with `@method static` plus a return type were pure by default.

It was not accurate, it's just that impurity was not checked. In https://github.com/vimeo/psalm/pull/6724, I added code to check purity through __callStatic by checking the pure attribute of the pseudo method that was called.

It was successful in flagging impure calls in pure context but I did not see that the pure calls in pure context were now broken because I used the pure flag of the pseudo method (that is never flagged) instead of the pure flag of __callStatic itself.

This is now fixed and both way works, so this will fix https://github.com/vimeo/psalm/issues/6941

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php
@@ -469,6 +469,7 @@ class AtomicStaticCallAnalyzer
                 true,
                 $context->insideUse()
             )) {
+                $callstatic_storage = $codebase->methods->getStorage($callstatic_id);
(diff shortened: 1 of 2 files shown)


weirdan:
I wonder whether it would work with `usePhpDocMethodsWithoutMagicCall="true"`.

orklah:
> I wonder whether it would work with usePhpDocMethodsWithoutMagicCall="true".

Both tests I added use `__callStatic` so the property does nothing here (I tested that locally)

However, if you have this config and you try to call a static method declared in docblock, you have no way to declare the call as pure but Psalm won't complain about it.

It could be an enhancement

Ocramius:
Thanks @orklah!

PR #6954, opened 2021-11-21
Update $data argument types of file_put_contents

ciaranmcnulty:
The main aim is to exclude boolean false, which would be coerced into '' in the following cases:

```php
file_put_contents('outfile', file_get_contents('invalid_file')); // false instead of string
file_put_contents('outfile', fopen('invalid_file', 'r')); // false instead of resource
file_put_contents('outfile', file('invalid_file')); // false instead of array
```


orklah:

Note that with Psalm's default, there still won't be an error from false due to https://psalm.dev/docs/running_psalm/configuration/#ignoreinternalfunctionfalsereturn

Make sure to put that to false in order to get Psalm to emit errors here!

ciaranmcnulty:
I'm not sure those functions 'rarely return false' TBH

orklah:
Yeah, it's a pretty opinionated take that Psalm took, but as long as it's behind a config, I'm fine with it.

Ideally, Psalm would have made the difference between false values returned because of wrong parameters and false values returned because runtime errors (missing files is a good example here). This way, it would have paved the way for PHP8 TypeErrors and left other false/null values as error by default, but that would have been a big work...

PR #6955, opened 2021-11-22
sleep(0) is valid

BenMorel:
https://3v4l.org/qknjR

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -12187,7 +12187,7 @@ return [
 'sin' => ['float', 'num'=>'float'],
 'sinh' => ['float', 'num'=>'float'],
 'sizeof' => ['int', 'value'=>'Countable|array', 'mode='=>'int'],
-'sleep' => ['int|false', 'seconds'=>'positive-int'],
+'sleep' => ['int|false', 'seconds'=>'0|positive-int'],
 'snmp2_get' => ['string|false', 'hostname'=>'string', 'community'=>'string', 'object_id'=>'string', 'timeout='=>'int', 'retries='=>'int'],
 'snmp2_getnext' => ['string|false', 'hostname'=>'string', 'community'=>'string', 'object_id'=>'string', 'timeout='=>'int', 'retries='=>'int'],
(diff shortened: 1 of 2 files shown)


weirdan:
What's the use-case though? Is it to yield CPU to other processes?

BenMorel:
Our use case is a CLI that accepts an optional startup delay, and we do accept `0` seconds.

Anyway, I think Psalm should align with PHP's behaviour, that complains only for `< 0`?

orklah:
We do have this issue too in our codebase. Our default sleep duration is 0 but can be changed. It feels cheap to make a special condition to avoid calling a sleep instruction that will do nothing

weirdan:
There's a difference between not calling `sleep()` at all and calling it with `0` argument, most noticeable when the system is under load. E.g. running paratest on Psalm test suite with Xdebug enabled (to generate load) and the following script in parallel:
```php
<?php

$s = microtime(true);
$t = $argv[1];
if ($t >= 0) {
    sleep($argv[1]);
}

var_dump('elapsed', number_format(microtime(true) - $s, 6));
```
I get:
```console
$ time php qq.php -1                                                                                     
string(7) "elapsed"
string(8) "0.000018"
user=0.14s system=0.05s cpu=19% total=0.980

$ time php qq.php 0
string(7) "elapsed"
string(8) "0.014775"
user=0.16s system=0.03s cpu=20% total=0.929
```
(LA ~10).

So `0` is not totally useless, as it does produce measurable effect.

weirdan:
Thanks!

orklah:
Is it any different than another "no-action" function call? Let's say strtolower on an empty string?

weirdan:
It is.

Quoting `nanosleep()` man page (`sleep()` is using `nanosleep()` under the hood):
> nanosleep()  _suspends the execution of the calling thread_ until either at least the time specified in *req has elapsed, or [...]
> Furthermore, after the sleep completes, _there may still be a delay before the CPU becomes free_ to once again execute the calling thread.

The net effect of that is that the longer the CPU runqueue, the later your process will be waked up, on average. `sleep(0)` still suspends your process, and it then must wait its turn to use the CPU again.

orklah:
Good to know, thanks!

PR #6963, opened 2021-11-22
Attempt to fix #6937

### LINKED ISSUE #6937: ImplementedReturnTypeMismatch on template trait abstract method
Two template traits used by a class, one trait declares an abstract method, other implements it. When used, the ImplementedReturnTypeMismatch error is reported.

https://psalm.dev/r/07c75f40bc

ptomulik:
Just a blind shot, prehaps completely wrong. But maybe a good starting point for discussion.

### DIFF
--- a/src/Psalm/Internal/Analyzer/MethodComparator.php
+++ b/src/Psalm/Internal/Analyzer/MethodComparator.php
@@ -964,6 +964,15 @@ class MethodComparator
                 $guide_method_storage_return_type,
                 $codebase
             );
+
+            if ($implementer_method_storage->defining_fqcln) {
+                self::transformTemplates(
+                    $implementer_classlike_storage->template_extended_params,
+                    $implementer_method_storage->defining_fqcln,
+                    $implementer_method_storage_return_type,
+                    $codebase
+                );
+            }
         }
 
         if ($implementer_classlike_storage->is_trait) {
--- a/tests/ReturnTypeTest.php
+++ b/tests/ReturnTypeTest.php
@@ -1002,6 +1002,37 @@ class ReturnTypeTest extends TestCase
                         }
                     }'
             ],
+            'returnTypeOfAbstractAndConcreteMethodFromTemplatedTraits' => [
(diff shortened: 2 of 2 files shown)


orklah:
Seems like it solves the issue without adding more, that's already a huge win!

I don't have much experience in traits handling with Psalm, mind explaining a little what you did in there so we can have an idea of what's going on?

ptomulik:
The patch is more an effect of experimenting with the [example from the original issue](https://github.com/vimeo/psalm/issues/6937#issuecomment-971557907) than a consciously elaborated solution. I've observed, than for the class ``Test`` involving two traits as follows

```php
class Test {
    /** @template-use AggregateTrait<int> */
    use AggregateTrait; // implements getValue()
    
    /** @template-use AnotherTrait<int> */
    use AnotherTrait; // declares abstract getValue()
    
    public function __construct() {
        $this->value = 123;
    }
```

the method [compareMethodDocblockReturnTypes](https://github.com/vimeo/psalm/blob/f19155817ee0ec80700ca7120058d76f2b45e130/src/Psalm/Internal/Analyzer/MethodComparator.php#L919) get called twice. In the first call
```
$implementer_classlike_storage->name is "AggregateTrait"
$guide_classlike_storage->name is "AnotherTrait"
$implementer_classlike_storage->is_trait is true
```
And this passes without any problems. But in the second call
```
$implementer_classlike_storage->name is "Test"
$guide_classlike_storage->name is "AnotherTrait"
$implementer_classlike_storage->is_trait is false
```
and this fails. It seems that in the second call psalm treated our ``getValue()`` method as it was implemented by ``Test`` class, not by ``AggregateTrait``. The template parameters were not applied (transformed) and the return type of ``getValue()`` was inferred to be a generic ``T:AggregateTrait`` instead of ``int``. 

What I proposed here is to apply the template parameters (``int``) to ``use AggregateTrait`` in the second call, when the "implementer classlike" is not a trait.

orklah:
Thanks!

PR #6964, opened 2021-11-22
Make UnitEnum name property non-empty

ricardoboss:


orklah:
Thanks, that's great!

ricardoboss:
Hi @orklah, it seems I made a mistake somewhere: https://psalm.dev/r/1bf1de16e0

Any idea why this PR didn't fix this?

orklah:
Yeah, sorry, I should have thought about that, there's special code for handling ->name on enums here: https://github.com/vimeo/psalm/blob/39402c233d4856447615dfccf1d3431eca6ca6c9/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/AtomicPropertyFetchAnalyzer.php#L205

I wonder if we could drop the whole block now that we have a stub... Do you mind trying that? If it doesn't work, then we need to replace Type::getString by a Type::getNonEmptyString()

ricardoboss:
I'll look into it!

ricardoboss:
So I tried flat-out removing the `else` branch, which led to an error in the tests saying "unable to infer return type". I then proceeded to change the line you mentioned to set the type to `Type::getNonEmptyString()` and that worked. Do you want to try to remove the special handling or is this fine as it is?

orklah:
that's fine, thanks for your help!

PR #6987, opened 2021-11-25
fix signature for openssl_open under PHP >= 8.0

pilif:
fixes the false `InvalidArgument` here: https://psalm.dev/r/d84f1672b7


dictionaries/CallMap_80_delta.php
```diff
@@ -941,6 +941,10 @@
       'old' => ['bool', 'mysql'=>'mysqli', 'flags='=>'int', 'name='=>'string'],
       'new' => ['bool', 'mysql'=>'mysqli', 'flags='=>'int', 'name='=>'?string'],
     ],
+    'openssl_open' => [
+      'old' => ['bool', 'data'=>'string', '&w_output'=>'string', 'encrypted_key'=>'string', 'private_key'=>'string|array|resource', 'cipher_algo='=>'string', 'iv='=>'string'],
+      'new' => ['bool', 'data'=>'string', '&w_output'=>'string', 'encrypted_key'=>'string', 'private_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array|string', 'cipher_algo='=>'string', 'iv='=>'string'],
```
weirdan, line 946: Doesn't `$private_key` here have the same range of types as in `openssl_sign`? That one has more detailed type for its `array` union element.

pilif: I took this signature from the [php manual](https://www.php.net/openssl_open), but yes, you're right. I overlooked that psalm of course has more info about arrays. let me fix that.

pilif: oh but wait: Does that signature make sense?

it's currently `array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}` what does that even mean in psalm? using {} implies an object-like array, but then shouldn't there be a : key/value separator?

also: Where did we take the shape of that array from? The [php manual](https://www.php.net/openssl_sign) doesn't actually explain the array case

pilif: and looking at the php source code, I think, this is definitely bogus and should be

`array{0: string, 1: string}`

https://github.com/php/php-src/blob/35786e321b64cb064d9bbdc339836b7b9d1d1800/ext/openssl/openssl.c#L3509

pilif: I also just noticed that while `openssl_pkey_get_private` got the PHP-8 treatment, its alias `openssl_get_privatekey` didn't. and neither did `openssl_pkey_get_public` nor its alias `openssl_get_publickey`.

I'll fix all of those too, but can you first confirm that the currently present definition for the `array` argument is suboptimal? Then I'll fix all of those too.

weirdan: `array{A, B}` is a shorthand for `array{0:A, 1:B}`.

> Where did we take the shape of that array from?

There's a separate chapter on key / certificate params that vaguely documents array shape used: https://www.php.net/manual/en/openssl.certparams.php

@jirkace experimented with objects as the first key element and found it worked: https://github.com/vimeo/psalm/pull/6846#issuecomment-962867211

orklah: > array{A, B} is a shorthand for array{0:A, 1:B}

I believe there is a slight difference
`[1 => 'B', 0 => 'A']` would match the second one but not the first because the first one imply the order of the keys is important

Inside of Psalm, this flag is know as `is_list`, which suggest it has to start from 0 and be in order

weirdan: > I'm not sure this is intended from the PHP side.

IMO that would make sense - you have to provide passphrase somehow.

pilif: But in order to get to that `OpenSSLAsymmetricKey ` instance to pass in the array, you need to already have provided the passphrase to `openssl_pkey_get_private `

weirdan: Perhaps. I'm only superficially acquainted with this extension so I'm inclined to take your word for it.

pilif:
I have force-pushed a new commit which fixes all `openssl_*` signatures for PHP >= 8.0.

In the end, I went with the long signature for that `array` keys argument: Even if some of the arguments as they are allowed now make no sense, they *are* allowed by PHP and we should not complain about valid PHP code, no matter how nonsensical.

One thing: there is [`openssl_pkey_get_details`](https://www.php.net/openssl_pkey_get_details) which returns an associative array whose members depend on the key passed in as the argument and none of the types of these key-dependent values are documented in the manual.

The existing return type was declared as `array|false` and this PR leaves it at that. If somebody wants to go through all of the combinations and figure out their types, that's a substantial additional task which I think should be subject to a future PR as with this PR, at least all `openssl_*` functions are useable with psalm under PHP8, where prior to this PR they weren't.

pilif:
also - wow - what a minefield the `openssl_*` functions are.

stuff like 

Screenshot 2021-11-26 at 09 35 51

or

Screenshot 2021-11-26 at 09 38 23

in functions related to crypto...

dictionaries/CallMap.php
```diff
@@ -9304,60 +9304,60 @@
...
-'openssl_x509_export' => ['bool', 'certificate'=>'string|resource', '&w_output'=>'string', 'no_text='=>'bool'],
-'openssl_x509_export_to_file' => ['bool', 'certificate'=>'string|resource', 'output_filename'=>'string', 'no_text='=>'bool'],
-'openssl_x509_fingerprint' => ['string|false', 'certificate'=>'string|resource', 'digest_algo='=>'string', 'binary='=>'bool'],
-'openssl_x509_free' => ['void', 'certificate'=>'resource'],
+'openssl_verify' => ['-1|0|1|false', 'data'=>'string', 'signature'=>'string', 'public_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string', 'algorithm='=>'int|string'],
+'openssl_x509_check_private_key' => ['bool', 'certificate'=>'OpenSSLCertificate|string', 'private_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string'],
+'openssl_x509_checkpurpose' => ['bool|int', 'certificate'=>'OpenSSLCertificate|string', 'purpose'=>'int', 'ca_info='=>'array', 'untrusted_certificates_file='=>'string'],
```
weirdan, line 9356: `untrusted_certificates_file` is documented as nullable

weirdan (commented):
There's a number of nullability nitpicks, but otherwise this looks really good.

dictionaries/CallMap.php
```diff
@@ -9304,60 +9304,60 @@
...
 'openssl_get_cipher_methods' => ['array', 'aliases='=>'bool'],
 'openssl_get_curve_names' => ['list<string>'],
 'openssl_get_md_methods' => ['array', 'aliases='=>'bool'],
-'openssl_get_privatekey' => ['resource|false', 'private_key'=>'string', 'passphrase='=>'string'],
-'openssl_get_publickey' => ['resource|false', 'public_key'=>'resource|string'],
-'openssl_open' => ['bool', 'data'=>'string', '&w_output'=>'string', 'encrypted_key'=>'string', 'private_key'=>'string|array|resource', 'cipher_algo='=>'string', 'iv='=>'string'],
+'openssl_get_privatekey' => ['OpenSSLAsymmetricKey|false', 'private_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string', 'passphrase='=>'?string'],
+'openssl_get_publickey' => ['OpenSSLAsymmetricKey|false', 'public_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string'],
+'openssl_open' => ['bool', 'data'=>'string', '&w_output'=>'string', 'encrypted_key'=>'string', 'private_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string', 'cipher_algo='=>'string', 'iv='=>'string'],
```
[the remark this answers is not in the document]
pilif: I dug into this in the past and must have forgotten. It was optional until 8.0 and chose the insecure default. 

Since 8.0 it’s non-optional to highlight the insecure default. 

I’ll fix this tomorrow

dictionaries/CallMap.php
```diff
@@ -9304,60 +9304,60 @@
...
+'openssl_csr_export' => ['bool', 'csr'=>'OpenSSLCertificateSigningRequest|string', '&w_output'=>'OpenSSLAsymmetricKey', 'no_text='=>'bool'],
+'openssl_csr_export_to_file' => ['bool', 'csr'=>'OpenSSLCertificateSigningRequest|string', 'output_filename'=>'string', 'no_text='=>'bool'],
+'openssl_csr_get_public_key' => ['OpenSSLAsymmetricKey|false', 'csr'=>'OpenSSLCertificateSigningRequest|string', 'short_names='=>'bool'],
+'openssl_csr_get_subject' => ['array|false', 'csr'=>'OpenSSLCertificateSigningRequest|string', 'short_names='=>'bool'],
+'openssl_csr_new' => ['OpenSSLCertificateSigningRequest|false', 'distinguished_names'=>'array', '&w_private_key'=>'OpenSSLAsymmetricKey', 'options='=>'array|null', 'extra_attributes='=>'array|null'],
+'openssl_csr_sign' => ['OpenSSLCertificate|false', 'csr'=>'OpenSSLCertificateSigningRequest|string', 'ca_certificate'=>'OpenSSLCertificate|string|null', 'private_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string', 'days'=>'int', 'options='=>'array', 'serial='=>'int'],
```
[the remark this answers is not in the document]
pilif: fixed

dictionaries/CallMap.php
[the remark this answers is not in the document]
pilif: fixed, including the `cipher_algo` param

dictionaries/CallMap.php
```diff
@@ -9304,60 +9304,60 @@
...
+'openssl_pkcs12_export_to_file' => ['bool', 'certificate'=>'OpenSSLCertificate|string', 'output_filename'=>'string', 'private_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string', 'passphrase'=>'string', 'options='=>'array'],
 'openssl_pkcs12_read' => ['bool', 'pkcs12'=>'string', '&w_certificates'=>'array', 'passphrase'=>'string'],
-'openssl_pkcs7_decrypt' => ['bool', 'input_filename'=>'string', 'output_filename'=>'string', 'certificate'=>'string|resource', 'private_key='=>'string|resource|array'],
-'openssl_pkcs7_encrypt' => ['bool', 'input_filename'=>'string', 'output_filename'=>'string', 'certificate'=>'string|resource|array', 'headers'=>'array', 'flags='=>'int', 'cipher_algo='=>'int'],
+'openssl_pkcs7_decrypt' => ['bool', 'input_filename'=>'string', 'output_filename'=>'string', 'certificate'=>'OpenSSLCertificate|string', 'private_key='=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string'],
```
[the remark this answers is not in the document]
pilif: fixed.

dictionaries/CallMap.php
```diff
@@ -9304,60 +9304,60 @@
...
+'openssl_pkcs7_encrypt' => ['bool', 'input_filename'=>'string', 'output_filename'=>'string', 'certificate'=>'OpenSSLCertificate|list<OpenSSLCertificate|string>|string', 'headers'=>'array|null', 'flags='=>'int', 'cipher_algo='=>'int'],
 'openssl_pkcs7_read' => ['bool', 'input_filename'=>'string', '&w_certificates'=>'array'],
-'openssl_pkcs7_sign' => ['bool', 'input_filename'=>'string', 'output_filename'=>'string', 'certificate'=>'string|resource', 'private_key'=>'string|resource|array', 'headers'=>'array', 'flags='=>'int', 'untrusted_certificates_filename='=>'string'],
+'openssl_pkcs7_sign' => ['bool', 'input_filename'=>'string', 'output_filename'=>'string', 'certificate'=>'OpenSSLCertificate|string', 'private_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string', 'headers'=>'array|null', 'flags='=>'int', 'untrusted_certificates_filename='=>'string'],
```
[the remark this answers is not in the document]
pilif: fixed

dictionaries/CallMap.php
```diff
@@ -9304,60 +9304,60 @@
...
 'openssl_pkcs7_verify' => ['bool|int', 'input_filename'=>'string', 'flags'=>'int', 'signers_certificates_filename='=>'string', 'ca_info='=>'array', 'untrusted_certificates_filename='=>'string', 'content='=>'string', 'output_filename='=>'string'],
-'openssl_pkey_derive' => ['string|false', 'public_key'=>'mixed', 'private_key'=>'mixed', 'key_length='=>'?int'],
-'openssl_pkey_export' => ['bool', 'key'=>'resource', '&w_output'=>'string', 'passphrase='=>'string|null', 'options='=>'array'],
-'openssl_pkey_export_to_file' => ['bool', 'key'=>'resource|string|array', 'output_filename'=>'string', 'passphrase='=>'string|null', 'options='=>'array'],
-'openssl_pkey_free' => ['void', 'key'=>'resource'],
-'openssl_pkey_get_details' => ['array|false', 'key'=>'resource'],
+'openssl_pkey_derive' => ['string|false', 'public_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string', 'private_key'=>'OpenSSLAsymmetricKey|OpenSSLCertificate|array{OpenSSLAsymmetricKey|OpenSSLCertificate|string, string}|string', 'key_length='=>'?int'],
```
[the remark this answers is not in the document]
pilif: fixed

dictionaries/CallMap.php
pilif: fixed

pilif:
thank you very much for your additional remarks. When I went through the functions, I mostly looked at the `resource`-typed parameters and didn't bother re-checking all the nullabilities. Now everything is up to date with the manual.

The windows tests fail before they even start testing with some missing infrastructure to install PHP. I don't know what's going on there, but I don't think it's related to the updated call map.

weirdan:
> I don't know what's going on there

Seems to be a temporary glitch. I've restarted the job and it seems to be working now.

weirdan:
Thanks!

Issue #6991, opened 2021-11-25
Numeric type hints and casts not considered in taint analysis


ohader:
https://psalm.dev/r/8fa14c18fc

Related aspects:
* #4872
* #6813

ohader:
I've just added some tests to reproduce this behavior in #6992:
* `function fetch(int $id): string { return query("SELECT * FROM table WHERE id=" . $id); }` fails
* `function fetch($id): string { return query("SELECT * FROM table WHERE id=" . (int)$id); }` works

PR #6992, opened 2021-11-25
Add test cases for type hints & type cast using annotated sinks

ohader:
Related: #6991


orklah:
Thanks, I used your tests in #6993

Issue #6994, opened 2021-11-26
Psalm forbids external access to enum constants


weirdan:
https://psalm.dev/r/d47f46d59b?php=8.1
Yet this is allowed by PHP: https://3v4l.org/NTT7F#v8.1rc3

PR #6995, opened 2021-11-26
Drop special handling of constant references on enums

weirdan:
Internally enum cases are constants, and they should be resolved as such.

Fixes vimeo/psalm#6994


orklah:
Yeah, that code seems weird, const don't return enumcases

orklah (approved):

weirdan:
> const don't return enumcases

Well, cases _are_ consts, but `ClassLikes::getClassConstantType()` already deals with that.

I'm a bit confused as the change I did shouldn't have fixed the test I added. Gotta spend some time debugging, so I'm putting this on hold for the time being.

weirdan:
Turns out I actually needed to drop more code and add more tests :laughing:

PR #7003, opened 2021-11-26
Use return type hints instead of php doc "@return void"

rarila:
Just some cosmetics (and of course forced type safetyness:-)

Psalm uses PHP ≥ v7.1 and void return type hints were introduced with 7.1.
Anyway, they are used already here and there.

I did not change testcases and fixtures (now) as I don’t know if there’s some test checking where the return type definition originated from. If anyone knows… :mega: … :ear:

I probably will simply will give it a try and see, if some test goes :boom:

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -119,9 +119,6 @@ class ClassAnalyzer extends ClassLikeAnalyzer
             . '_' . $class->getLine() . '_' . (int)$class->getAttribute('startFilePos');
     }
 
-    /**
-     * @return void
-     */
     public function analyze(
         ?Context $class_context = null,
         ?Context $global_context = null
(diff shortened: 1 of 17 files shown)


orklah:
Thanks! However, some of those could be an issue. Plugins interfaces especially are designed to be implemented. Adding return types would be a BC break.

Can you revert those? If you want, you can submit those again in another PR we will be able to merge for Psalm 5

rarila:
Was already doing that…

But the interesting question is, why didn’t my local test fail?! Makes things a litle bit difficult.

orklah:
Thanks!

rarila:
:relieved:
@orklah Any idea what's different in running tests locally and here?!
I would prefer to upload working PRs instead of doing lengthy fixes afterwards :frowning:

orklah:
Well, tests should be able to run locally, but Psalm make some more checks: it runs on itself to detect errors and it also runs on other projects as well.

The last one rarely fails alone so  it shouldn't be an issue. Running Psalm on itself can be done through `php psalm` in your command line

rarila:
Hmm, but shouldn’t ``composer psalm`` be the same then?!

orklah:
yep, probably! To be honest, I'm so used to test suite not working on window out of the box that I tweak things until I have something that works and I don't bother changing things after that. Thanks for doing so :)

PR #7005, opened 2021-11-26
Make test runnable on plain windows

rarila:
Running ``composer all-tests`` or ``composer psalm`` fails on (standard) Windows as it doesn’t care for the shebang.

[Executing the script](https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts) as in this patch should (hopefully, we’ll see :-) independent of the OS.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -104,10 +104,10 @@
     "scripts": {
         "all-tests": [
             "phpcs",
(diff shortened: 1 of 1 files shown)


orklah:
Great! Thanks!

PR #7006, opened 2021-11-27
Display target PHP version

weirdan:
Historically it was often not quite clear to users what PHP version Psalm assumes, and why. This PR addresses this issue by printing the version and where we got it from right before scanning the files.

Note that psalm/psalm.dev#69 has to merged **first** lest psalm.dev sandbox become unusable.

### DIFF
--- a/src/Psalm/Internal/CliUtils.php
+++ b/src/Psalm/Internal/CliUtils.php
@@ -5,6 +5,7 @@ namespace Psalm\Internal;
 use Composer\Autoload\ClassLoader;
 use Phar;
 use Psalm\Config;
+use Psalm\Internal\Analyzer\ProjectAnalyzer;
 use Psalm\Internal\Composer;
 
 use function assert;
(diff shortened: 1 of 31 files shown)


src/Psalm/Internal/CliUtils.php
```diff
@@ -622,4 +623,25 @@ public static function convertMemoryLimitToBytes(string $limit): int
...
         return (int)$limit;
     }
+
+    public static function initPhpVersion(array $options, Config $config, ProjectAnalyzer $project_analyzer): void
+    {
+        $source = null;
+
+        if (isset($options['php-version'])) {
+            if (!is_string($options['php-version'])) {
+                die('Expecting a version number in the format x.y' . PHP_EOL);
```
orklah, line 633: This check can be dropped here, it seems to be checked again in project analyzer below

orklah (approved):
Oh, that's cool!

src/Psalm/Internal/CliUtils.php
weirdan: It is to avoid MixedAssignment to `$version` variable.

PR #7009, opened 2021-11-27
Fix failing symlink test on Windows

rarila:
At least on Windows 10 ``symlink()`` fails with the error message "_symlink(): Permission denied_", so this unit test failed.

With this PR both error messages are allowed and the test is skipped if neccessary.

Now local unit tests on Windows and those on CI finish identically with no errors or failures and 35 skipped tests.


weirdan:
Thanks!

PR #7010, opened 2021-11-28
Include enum cases in const wildcards

weirdan:
Refs vimeo/psalm#6471


src/Psalm/Internal/Type/TypeExpander.php
```diff
@@ -231,10 +232,6 @@ function ($constant_name) use ($const_name_part): bool {
                         );
                     }
                 } else {
-                    if ($class_storage->is_enum) {
-                        return new Type\Atomic\TEnumCase($return_type->fq_classlike_name, $return_type->const_name);
-                    }
```
weirdan, line 236: It seems the more special handling for enums I drop the better it works :smile:

Issue #6429, opened 2021-09-05
Enums: Implement callmap Enum::from, Enum::tryFrom and Enum::cases


weirdan:
https://psalm.dev/r/3a77b77825?php=8.1

https://wiki.php.net/rfc/enumerations

We need a stub for `BackedEnum`, like this:
```php
interface BackedEnum<T of int|string> extends UnitEnum {
   public static function from(T $value): static;
   public static function tryFrom(T $value): ?static;
   public static function cases(): list<static>; // inherited from `UnitEnum`
}
```
And make Psalm infer that any backed enum implement corresponding type instance (`BackedEnum<int>` or `BackedEnum<string>`), based on backing type.

This is a problem, though, as we currently don't allow static methods to reference class type parameters.

weirdan:
https://psalm.dev/r/3a77b77825?php=8.1

weirdan:
@orklah, @muglug any thoughts on the point of static methods accessing class/interface type params? I know we had quite a few requests for that to be used for userland classes, but now we're facing the need for that for a core interface.

PHPStan allows that (https://phpstan.org/r/130ede95-e35a-4dc0-883b-76e6b2b0550f), so it's one of the points where tools have diverged.

muglug:
I had strong feelings about this when generics was a fairly new feature and where I thought this might be a source of bugs for users coming from languages with reified generics. Now I’m not really bothered — there are enough people around the community who have a basic understanding of generics and could point out common pitfalls, so it might be OK to relax.

orklah:
I think the limitation was added to avoid confusion with this kind of code: https://psalm.dev/r/ea77a57f0f. It makes sense for me that a class level template would be filled differently in different objects and as such, it's weird to see it in a static method.

But your example is correct and there is no good way to make it in Psalm.

In my head, the best way to solve that would be a tag that defines the template on method level: https://psalm.dev/r/9fdff9bdfe

But the added value compared to added complexity is discutable...

ricardoboss:
@weirdan shouldn't these methods also be considered `#[Pure]`?

weirdan:
They should.

PR #7011, opened 2021-11-28
Infer `::from()` and `::tryFrom()` return types on backed enums

weirdan:
Fixes vimeo/psalm#6429

### DIFF
--- a/src/Psalm/Internal/Codebase/Methods.php
+++ b/src/Psalm/Internal/Codebase/Methods.php
@@ -121,8 +121,16 @@ class Methods
             return false;
         }
 
-        if ($class_storage->is_enum && $method_name === 'cases') {
-            return true;
+        if ($class_storage->is_enum) {
+            if ($method_name === 'cases') {
+                return true;
+            }
+
+            if ($class_storage->enum_type
+                && \in_array($method_name, ['from', 'tryFrom'], true)
+            ) {
+                return true;
+            }
         }
 
         $source_file_path = $source ? $source->getFilePath() : $source_file_path;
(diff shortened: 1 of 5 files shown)


src/Psalm/Internal/Codebase/Methods.php
```diff
@@ -686,21 +694,55 @@ public function getMethodReturnType(
...
+                || $original_method_name === 'tryfrom'
+                ) && $source_analyzer
+                && isset($args[0])
+                && ($first_arg_type = $source_analyzer->getNodeTypeProvider()->getType($args[0]->value))
+            ) {
+                $types = [];
+                foreach ($original_class_storage->enum_cases as $case_name => $case_storage) {
+                    if (UnionTypeComparator::isContainedBy(
+                        $source_analyzer->getCodebase(),
+                        \is_int($case_storage->value) ?
```
orklah, line 730: Is it possible to use a constant to initialize a case? If it happens, is_int() will fail. Can't we access the type at this moment?

weirdan: It's possible, and it fails. But not here.

https://3v4l.org/WHXIV#v8.1rc3
https://psalm.dev/r/3547ca91d3

weirdan: Built-in constants work. 
User-defined constants don't: https://3v4l.org/nhGRA#v8.1rc3
Class constants work if the class is defined before the enum: https://3v4l.org/bcYPN#v8.1rc3.
Class constants from autoloaded classes do not work: https://3v4l.org/AjOCi#v8.1rc3

3v4l doesn't have proper 8.1 yet, so it's based on RC3. Some of the above could have been fixed before final release.

weirdan: > Some of the above could have been fixed before final release.

I've just built 8.1 branch, and there are no changes for those cases.

PR #7014, opened 2021-11-28
Fix mysqli_fetch_* stubs

kamil-tekiela:
I have just noticed during testing that I have specified the wrong type in the previous PR. While correcting it, I also decided to provide more specific types. I am not sure if the types for `mysqli_fetch_array()` works this way, but I hope it does.


orklah:
Can you use the occasion to fix the indentation in mysqli stub? :)

kamil-tekiela:
I fixed the indentation and I have also decided to properly specify the type. The type is taken from `fetch_column()` - false. In the case of MYSQLI_NUM it will be a list. In the case of `fetch_all()` it is always a list of one of the three options. 

I have also noticed I forgot to fix `fetch_fields()`

dictionaries/CallMap.php
```diff
@@ -8462,16 +8462,20 @@
 'mysqli_error_list' => ['array', 'mysql'=>'mysqli'],
 'mysqli_escape_string' => ['string', 'mysql'=>'mysqli', 'string'=>'string'],
 'mysqli_execute' => ['bool', 'statement'=>'mysqli_stmt', 'params='=>'list<mixed>|null'],
-'mysqli_fetch_all' => ['array', 'result'=>'mysqli_result', 'mode='=>'int'],
-'mysqli_fetch_array' => ['array|false|null', 'result'=>'mysqli_result', 'mode='=>'int'],
-'mysqli_fetch_assoc' => ['array<string,string>|false|null', 'result'=>'mysqli_result'],
+'mysqli_fetch_all' => ['list<array<array-key,null|int|float|string>>', 'result'=>'mysqli_result', 'mode='=>'MYSQLI_BOTH'],
+'mysqli_fetch_all\'1' => ['list<array<string,null|int|float|string>>', 'result'=>'mysqli_result', 'mode='=>'MYSQLI_ASSOC'],
+'mysqli_fetch_all\'2' => ['list<list<null|int|float|string>>', 'result'=>'mysqli_result', 'mode='=>'MYSQLI_NUM'],
```
weirdan, line 8467: Psalm does not support non-class constants as argument types (they can't be syntactically distinguished from class names), so you'd need to use their values instead.

weirdan (changes requested):

weirdan (approved):

weirdan:
Thanks!

PR #7015, opened 2021-11-28
Drop AppVeyor config

weirdan:


weirdan:
Oh, I see. Only @muglug can actually stop AppVeyor building this repo. Thus the neutering config file needs to be kept until then.

PR #7017, opened 2021-11-29
Disabled coverage (and xdebug) for CI builds

weirdan:
I'd expect to see some performance gains here.


weirdan:
This actually looks quite good indeed.

PR #7019, opened 2021-11-29
Disallow ksorting lists

### LINKED ISSUE #7018: Enhancement: It makes no sense to ksort a list-like keyed array
https://psalm.dev/r/9cc8199536
Similar to my `array_values` issue/MR

danog:
Fixes #7018.
Settled on the same logic used for `array_values`, as lists can only be constructed by appending elements to an array or by using `array_values` itself, in both cases they are sorted by default.


orklah:
I'm not a fan of this one, ksort is hardly a function to transform types, so RedundantCast feels very weird here... However, I couldn't find a more adapted issue and I'm not sure this  case is common enough to create a new issue...

weirdan:
I'd suggest to add `RedundantFunctionCall` for this kind of issues. Also, this looks very useful to me (as are other dead/useless code checks).

orklah:
I was not clear, I meant that I was not a fan of this being classified as RedundantCast but the feature seem fine though.

I'm okay with a RedundantFunctionCall, thinking about it more, this could indeed be reused (a dumb example, but a call to max() with a single value could fit in there).

danog:
Done!

orklah:
Cool! Thanks!

orklah:
@weirdan Can I let you merge that and updating the baseline for phpunit?

weirdan:
Thanks!

PR #7020, opened 2021-11-29
Migrate most IssueBuffer::accepts calls to IssueBuffer::maybeAdd

muglug:
This is not a massive diff, and addresses some (IMO) tech debt in calls to `IssueBuffer::accept`


orklah:
Seems fine to me :)

weirdan:
Thanks!

Issue #7007, opened 2021-11-27
PHP 8.2 has `PHP Error: Creation of dynamic property PhpParser\Node\Stmt\Class_::$namespacedName is deprecated`


TysonAndre:
https://wiki.php.net/rfc/deprecate_dynamic_properties deprecates it in 8.2

This should be using setAttribute('namespacedNameObject') and getAttribute(string $key, $default = null) instead - I'm not familiar with whether the properties are considered to be part of the public API yet.

- Checking for errno=E_DEPRECATED and  `Creation of dynamic property PhpParser\Node` at the start of $errstr is another option

https://github.com/vimeo/psalm/pull/5091 was added in september to set dynamic properties.

orklah:
Oh, I fixed a lot of them here: https://github.com/vimeo/psalm/pull/6656 but seems like I missed some

winstrand:
Still present in v3.4.3
Deprecated: Creation of dynamic property PhpParser\Node\Stmt\PropertyProperty::$fqsen
Deprecated: Creation of dynamic property PhpParser\Node\Stmt\ClassMethod::$fqsen
Deprecated: Creation of dynamic property PhpParser\Node\Stmt\Namespace_::$fqsen
Deprecated: Creation of dynamic property PhpParser\Node\Stmt\Class_::$fqsen

PR #7022, opened 2021-11-29
remove the remaining dynamic property assignment

orklah:
The goal is to fix https://github.com/vimeo/psalm/issues/7007

But reading the issue again, I'm not sure I got it right the first time.

@TysonAndre , can you confirm this is what you had in mind? I was confused by `setAttribute('namespacedNameObject')`, why Object?

(On an unrelated note, the reason why I didn't find those two for #6656 is because there is a @property on PHP-Parser which messed with my detection)

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/SimpleNameResolver.php
+++ b/src/Psalm/Internal/PhpVisitor/SimpleNameResolver.php
@@ -225,11 +225,10 @@ class SimpleNameResolver extends NodeVisitorAbstract
 
     protected function addNamespacedName(Stmt\Class_ $node): void
     {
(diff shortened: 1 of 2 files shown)


TysonAndre (approved):
That's what I had in mind.

I don't expect issues: Looking at this again, `class PhpParser\Node\Stmt\Class_ extends Declaration` and its subclasses are a distinct group of classes from `PhpParser\Node\Name`, so this won't conflict with the existing attribute of the same name used with `PhpParser\Node\Name` in SimpleNameResolver::resolveName.

orklah:
Thanks!

PR #7029, opened 2021-11-30
Enhancing composer scripts

rarila:
Sofar done:

- Added ``phpunit`` script.
- Added ``phpunit-p`` script running paratest.
- Added scripts descriptions displayed e.g. by ``composer run -l``.
- Sorted alphabetically.
- Replaced multiple script runs by references to single tests.

Questions:

- At the moment the naming is mixed between generic names like ``cs`` and tool names like ``phpunit``. Would it be desirable to use generic names everywere or tool names erywhere or do we just don’t care?

- I've added paratest to the scripts and wonder if it should be used as the standard for ``tests`` and ``all-tests`` (``phpunit`` = paratest, ``phpunit-std`` = phpunit) as it’s much faster.

- I’ve noticed that ``php-parallel-lint`` gets installed. Is that still useful? – If yes, we should probably expose it in the scripts section and add it to ``tests`` and ``all-tests``. – If yes, what would be sensible calling options and how should it be named (``lint``, ``syntax-check``, , ``check-syntax``, …)?

- I wonder if it is reasonable to have ``tests`` and ``all-tests``. Would it ux wise be better to just have ``tests`` that runs all tests and single commands for distinctive tasks?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -107,11 +107,11 @@
             "@php ./psalm --find-dead-code",
             "phpunit"
         ],
-        "psalm": "@php ./psalm --find-dead-code",
-        "phpunit": "phpunit",
-        "phpunit-p": "paratest --runner=WrapperRunner",
         "cs": "phpcs -p",
         "cs-fix": "phpcbf -p",
+        "phpunit": "phpunit",
+        "phpunit-p": "paratest --runner=WrapperRunner",
+        "psalm": "@php ./psalm --find-dead-code",
         "tests": [
             "phpcs",
             "phpunit"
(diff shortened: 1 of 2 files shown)


weirdan:
> I’ve noticed that php-parallel-lint gets installed. Is that still useful? 

We use it in CI builds (see #5209 for some context).

> Would it be desirable to use generic names everywere or tool names erywhere or do we just don’t care?

Personally I don't have any opinion on this matter as I don't use those scripts.

rarila:
Ah, thanks for the info. Then I'll gonna add it to the tests to bring them further in line with ci tests.

I had it already two times that local tests were ok and ci failed what IMHO isn’t good.

weirdan:
Thanks!

PR #6853, opened 2021-11-07
Fixed RedisCluster::setOption() argument types

jirkace:
Fixes #4708

PR #7030, opened 2021-11-30
Fix RedisCluster::getOption() and RedisCluster::setOption() argument types

ostrolucky:
See https://github.com/phpredis/phpredis/blob/aac42cd33510030bb62973f8886f4cd98b9b8003/redis_cluster.stub.php#L122

Related https://github.com/vimeo/psalm/pull/6853


orklah:
I may be wrong here but the link you posted points to `Redis`, not `RedisCluster`

If the link is good though, the param name should change too (from name to option)

ostrolucky:
Thx. Fixed the link and renamed name too.

orklah:
link is still wrong :) here's the right one:
https://github.com/phpredis/phpredis/blob/aac42cd33510030bb62973f8886f4cd98b9b8003/redis_cluster.stub.php#L122

In the PR for setOption, the contributor decided to go for `int|string` because he didn't know if it has been a change from older versions or if it was a mistake.

Do you know the answer? If it change between versions of Redis, we should keep retrocompatibility and allow string too, but if it's a mistake, it would be cool to fix getOption in that way too to keep consistency

ostrolucky:
I am not 100% sure, but I think it was a mistake. We never used anything else than int in SncRedisBundle. I think contributor used string because phpstorm stubs are most likely wrong - those specify string. I've changed it now for getOption...

ostrolucky:
Interesting thing that should be also mentioned, is that phpredis is doing something with this type currently in their dev branch https://github.com/phpredis/phpredis/issues/2038. I would assume they broke their types they provide in reflection. But again, it's their dev branch only and this must have been done recently only. To go forward with this PR, I guess we should get and answer there first.

orklah:
(note: the CI failure is unrelated, you can rebase to fix it)

ostrolucky:
So, referenced issue was solved. So indeed it's int. I would say PR is ready.

orklah:
Thanks!

Issue #7028, opened 2021-11-30
PHPUnit output cluttered


rarila:
PHPUnit output gets cluttered like this:

```
[...]
............................................................. 2196 / 6542 ( 33%)
................................................Target PHP version: 8.0 (inferred from current PHP version)
............. 2257 / 6542 ( 34%)
............................................................. 2318 / 6542 ( 35%)
............................................................. 2379 / 6542 ( 36%)
[...]
...S......................................................... 2928 / 6542 ( 44%)
............................................................. 2989 / 6542 ( 45%)
............................................................. 3050 / 6542 ( 46%)
...........................Target PHP version: 8.0 (inferred from current PHP version)
.Target PHP version: 8.0 (inferred from current PHP version)
.Target PHP version: 8.0 (inferred from current PHP version)
.Target PHP version: 8.0 (inferred from current PHP version)
.Target PHP version: 8.0 (inferred from current PHP version)
.............................. 3111 / 6542 ( 47%)
............................................................. 3172 / 6542 ( 48%)
[...]
```

Probably introduced by #7006

Could be that some tests call ``$this->progress->write($this->generatePHPVersionMessage());`` that finally ends up in ``fwrite(STDERR, $message);`` and doesn’t get captured by PHPUnit that way.

PS: don’t use paratest to reproduce as there the output seems to be filtered

weirdan:
Indeed. I wonder why those tests don't use `VoidProgress`.

weirdan:
Ah, they may use it, but `Progress::write()` is not a dummy method, unlike all other. I'll fix it.

PR #7031, opened 2021-11-30
Prevent VoidProgress from outputting anything

weirdan:
Fixes vimeo/psalm#7028


orklah:
I almost feel like the fwrite part should have been in each child. I mean, Progress is basically an interface at this point, it's weird to have some code there

weirdan:
@orklah it seems that way, yes. But making it no-op would be a BC break for possible external descendants. Strictly speaking even this change is a BC break as someone may have relied on `VoidProgress::write()` producing output (and may have extended it as well).

PR #7036, opened 2021-12-01
Throw exception when baseline file is empty

kamil-tekiela:
This should solve the error:

> Uncaught ValueError: DOMDocument::loadXML(): Argument #1 ($source) must not be empty in ...\vimeo\psalm\src\Psalm\ErrorBaseline.php:85


weirdan:
Thanks!

Side note: we should probably annotate `$source` as `non-empty-string` so that Psalm would be able to catch it too.

Issue #7038, opened 2021-12-01
Assertion inside loop does not work


gndk:
Asserting that `mixed` is a `string` (or any other more specific type) does not work inside a loop:

With `if(! is_string)`: https://psalm.dev/r/07d2930bc7
With `assert(is_string())`: https://psalm.dev/r/1d30be66ac

But works as expected when moved outside the loop:

With `if(! is_string)`: https://psalm.dev/r/f0033ef351
With `assert(is_string())`: https://psalm.dev/r/616e96ada3

orklah:
This has to do with the way Psalm record Issues while in loop.

When starting a loop, Psalm calls the method IssueBuffer::startRecording(). This increase by 1 the level of IssueBuffer::$recording_level and will record every issue in IssueBuffer::$recorded_issues instead of adding them to the list of prepared issues.
When leaving a loop, Psalm calls IssueBuffer::clearRecordingLevel()  to retrieve recorded issues and IssueBuffer::stopRecording() to decrement the recording level. It will then call IssueBuffer::bubbleUp() in order to try to emit issues again (if recording level is at 0, it will emit them, if not it will continue recording them).

The core of the issue is that, when Psalm encounters a is_string function, it will try removing any MixedAssignment issue that has been emitted, but given that the issue is not emitted but recorded, IssueBuffer::remove does nothing.

What would need to be done here is to improve IssueBuffer;;remove to allow it to remove recorded issues as well

PR #7040, opened 2021-12-01
document the behavior of methods in IssueBuffer

orklah:
This documents my findings when looking at https://github.com/vimeo/psalm/issues/7038

### DIFF
--- a/src/Psalm/IssueBuffer.php
+++ b/src/Psalm/IssueBuffer.php
@@ -102,8 +102,8 @@ class IssueBuffer
     private static $server = [];
 
     /**
-     * @param   string[]  $suppressed_issues
-     *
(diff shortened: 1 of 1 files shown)


src/Psalm/IssueBuffer.php
```diff
@@ -145,8 +148,11 @@ public static function addUnusedSuppression(string $file_path, int $offset, stri
     }
 
     /**
+     * This will return true if an issue is ready to be added for emission. Reasons for not returning true include:
```
weirdan, line 151: Are you sure it's not the opposite?

orklah: You're right, I'm changing that

PR #7042, opened 2021-12-01
get_class does not return false

kamil-tekiela:
Technically all functions could return false or null when a wrong parameter is provided, but that was undefined behavour in PHP <8 and should not be relied upon.

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -3720,7 +3720,7 @@ return [
 'get_call_stack' => [''],
 'get_called_class' => ['class-string'],
 'get_cfg_var' => ['string|false', 'option'=>'string'],
-'get_class' => ['class-string|false', 'object='=>'object'],
+'get_class' => ['class-string', 'object='=>'object'],
 'get_class_methods' => ['list<string>|null', 'object_or_class'=>'mixed'],
 'get_class_vars' => ['array<string,mixed>', 'class'=>'string'],
 'get_current_user' => ['string'],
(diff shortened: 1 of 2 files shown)


kamil-tekiela:
Ok, it did return false prior to PHP 8. I will change the PR to account for this tomorrow.

weirdan:
Did it ever return class without also producing a warning?

kamil-tekiela:
> Did it ever return false without also producing a warning?

Not to my knowledge. See https://heap.space/xref/PHP-7.4/Zend/zend_builtin_functions.c?r=1e0bc6e3#955

orklah:
not since PHP 5.3:
https://3v4l.org/XBMmE 
https://3v4l.org/clSSM
https://3v4l.org/n7SgA

Psalm is already emitting an error in all those cases. The only case where Psalm does not emit error is: https://psalm.dev/r/781da1e35e but it gives a warning even in PHP 5.2. I say we can get rid of the false return as we already did on such cases for other functions

weirdan:
If we can reliably flag cases where it would return false then we don't need `|false` in the signature.

There's one I can think of: calling `get_class()` without arguments in a non-class context: https://psalm.dev/r/58287e0c4e. I think it could be handled similarly to `__

orklah:
It's not flagged by Psalm but it never worked AND always emitted a warning or worse: https://3v4l.org/TkeS5 (weirdly, it started as a warning, then was an error, again a warning and it's now an error!)

orklah:
That should do it: https://github.com/vimeo/psalm/pull/7043

kamil-tekiela:
What about this PR? Is it ok to just remove `|false` from all versions or should I add delta?

weirdan:
Just dropping `|false` would be perfect.

orklah:
Thanks!

Issue #7047, opened 2021-12-02
4.13 - missing output with ConsoleReport in GHA env


weirdan:
4.12 - https://github.com/loophp/repository-monadic-helper/runs/4396330608?check_suite_focus=true#step:8:88

4.13 - https://github.com/loophp/repository-monadic-helper/runs/4395690254?check_suite_focus=true#step:8:90

/cc: @drupol

drupol:
Maybe @veewee has something to say here?

veewee:
Can you give me some context @drupol ? A bit too limited in time ATM to figure it all out :)

drupol:
Sorry Toon,

I was too hasty as usual.

Basically I was looking for help with Psalm when @weirdan noticed that the output on Github Actions was not visible.

Here's the conversation log:

![image](https://user-images.githubusercontent.com/252042/144478321-e074e687-24f2-4052-8424-1720cbea6fb0.png)

orklah:
Note that console links were introduced in this PR: https://github.com/vimeo/psalm/pull/6850

veewee:
Looks like that PR always adds the control characters instead of validating if those characters are supported as-in Symfony console for example:

https://github.com/symfony/console/blob/ec3661faca1d110d6c307e124b44f99ac54179e3/Output/StreamOutput.php#L94


We call psalm through Symfony/process in GrumPHP without enabling ANSI through cli param, tty or pty.

That most likely is the issue here. 

(Investigating through my phone, so could be horribly wrong :) )

weirdan:
Note that the actual color works, and 4.12 (with colors) worked just fine:

![image](https://user-images.githubusercontent.com/57403/144492284-794c43aa-dc6a-4316-9c8d-fa2c24d36f9b.png)

I guess we need to check whether the output is broken when Psalm is run in GHA env _without_ GrumPHP (as it's likely just passing bytes it receives from the cli tool to its output).

weirdan:
As I suspected, it's a problem with GHA rendering of Psalm's output. I'll look into disabling links in CI envs, they don't seem to be terribly useful there anyway.

@drupol I'd suggest to use `github` output format instead of default `console` that is being used, it'll save you a need to run Psalm again after GrumPHP completes. It's not yet available in GrumPHP though.

@veewee I've sent you a pull request to allow `github` output format (and several more) for Psalm: https://github.com/phpro/grumphp/pull/958

veewee:
Thanks!

PR #7049, opened 2021-12-03
Don't output console links in CI env

weirdan:
They are all but useless there, and are actually causing an issue with GHA output.

Fixes vimeo/psalm#7047

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -310,7 +310,7 @@ final class Psalm
         $progress = self::initProgress($options, $config);
         $providers = self::initProviders($options, $config, $current_dir);
 
-        $stdout_report_options = self::initStdoutReportOptions($options, $show_info, $output_format);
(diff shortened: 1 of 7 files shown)


weirdan:
Test failure is legit, but not being able to rely on superglobals makes it quite a bit harder.

orklah (approved):

drupol:
Thanks for fixing the issue !

Question:Could we consider using https://packagist.org/packages/ondram/ci-detector for checking if it is running in CI ?

orklah:
Yeah, we could consider it if needed. Do you happen to use an exotic CI environment?

drupol:
On my side no, I only use Github and Gitlab.

drupol:
That said, that package support quite a large amount of CI system that I was not even aware of !

orklah:
That was kinda my point, if we weren't aware they existed, it will be the case for most people too :)

If we have a feedback from a user, we'll consider making the change, but I try not to add dependencies for things I'm not sure are needed

PR #7053, opened 2021-12-03
callmap updated with fixed argument names according to PHP 8

SMAtaurRahman:
I've noticed that psalm callmap signatures sometimes contain different argument name than original name used by PHP.  This can cause false alarm when using named arguments.

I tried to fix argument names of simple core functions with this PR.
I detected these with a very ugly piece of code that compares function signatures from callmap with their reflection data.

I didn't touch class methods and extension functions, left out some core functions too because I was too dumb to decipher their signatures.

Let me know if these kind of changes are welcome, If so I will try to invest more time to address them.


orklah:
Yeah, this is welcome, we have some legacy notations in here (like rw_ prefixes that comes from Phan initially but I believe we never used them).

However, changing callmaps in bulks is not always an easy task, especially if there's more than one type of change (for example the removal of the param of curl_version: https://github.com/vimeo/psalm/pull/7053/files#diff-b8eaee1f550652657daf0a771be5e785bdb01e91acd004b4bbb4c41def706713R1696 in PHP 8).

I'd advise to only try to do one type of change in callmap at a time, even if it needs more than one PR.

It's easier to review too!

SMAtaurRahman:
> I'd advise to only try to do one type of change in callmap at a time, even if it needs more than one PR.

Thanks. I'll try to prepare an organized and small PR then.

Issue #6933, opened 2021-11-16
Missing constant name in the error message


kamil-tekiela:
See https://psalm.dev/r/f888fa7a73

When using a constant, Psalm produces a message with a missing variable name. I don't know if the INFO is related to this or not.

orklah:
I'll reopen this because I'm reverting my fix. It was having strange effects

PR #7054, opened 2021-12-03
fix missing case for displaying varId

orklah:
This will fix https://github.com/vimeo/psalm/issues/6933

The message is a little weird because it says `Cannot access value on variable ARR` when it's a constant, but it's not a big deal.

Didn't add test because this is just a wording issue

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ExpressionIdentifier.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ExpressionIdentifier.php
@@ -216,6 +216,10 @@ class ExpressionIdentifier
             }
         }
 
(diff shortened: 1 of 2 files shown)


orklah:
This revealed a deeper bug where, if a constant have a varId, it is affected by assertions a weird way and this snippet https://psalm.dev/r/8c26aa0e98 produce an error because `DIRECTORY_SEPARATOR` somehow became mixed

orklah:
Well, this is working. I'm not sure I like the fact that I had to call an Analyzer inside the Reconciler though, but I couldn't see a cache of global constants and I'm not sure we're supposed to create one...

orklah:
Do you have thoughts about that @weirdan ?

weirdan:
Looks good to me.

PR #7056, opened 2021-12-04
Remove always-false args

muglug:
Noticed this commit: https://github.com/vimeo/psalm/commit/153de33704

This cleans up those unnecessary args


orklah:
Oh nice, it used to confuse me a lot!

Thanks!

PR #7059, opened 2021-12-04
Improve count_chars stub

kamil-tekiela:
This should avoid error:

>ERROR: PossiblyInvalidCast - ....php:34:21 - array<int, int> cannot be cast to string (see https://psalm.dev/190)

and also https://psalm.dev/r/7d587720b5


weirdan:
Thanks!

kamil-tekiela:
@weirdan  I am surprised to see that the changes I suggested here are not working properly. I did some research and it looks single digit values are not working correctly. I will provide PR to fix what I have broken here, but I think there might be some issue in how Psalm treats single digit values.

PR #7060, opened 2021-12-04
Use import instead of fully qualified namespaces

### LINKED ISSUE #7037: Codingstyle – contributor input wanted
I’m a fan of a consistent looking codebase as IMHO makes it easier to read the code.

I think we all know the pros of coding style and also the cons – you need a bit of discipline when writing code. But the good thing is, nowadays we have a lot of good tools (IDEs, phpcs, ...) that makes things a lot of easier.

_So I wanna ask you contributors, what do you think?!_

One thing that annoys me most in Psalms codebase is the mix of fully qualified names and imports in the sourcecode. I’ve even seen imported stuff being used with fqn in the same file. So that would be the first thing I would start with. This will make a big PR for sure, but in the end the code base will profit from it.

rarila:
At least for most cases. Partials left for another PR as they need some more thinking.

All conversions were done carefullly, no tests fail. I hope that no external stuff (Plugins) break.
I don’t know now to review such a monster PR :-(

Fixes #7037


weirdan:
Mind looking into phar issue? See https://github.com/vimeo/psalm/blob/e1d1ef32b6a7ff15ffd517c29c205d67e8dfe48f/.circleci/config.yml#L44-L49 for how it's built.

rarila:
Hmm, we probably should include this in the local tests

rarila:
@weirdan Fixed. But that was mor an educated guess than knowledge. Do you know what happened there? Doesn’t look wrong to me. Some magic with box that does some search-replace with ``\Composer\Autoload\ClassLoader`` perhaps?!

src/Psalm/Internal/Analyzer/Statements/ThrowAnalyzer.php
```diff
@@ -84,6 +84,10 @@ public static function analyze(
         }
 
         if ($stmt instanceof PhpParser\Node\Expr\Throw_) {
+            // TODO: Find bug in Psalm:
+            // Conversion to "Type::getEmpty()" triggers Psalm DeprecatedMethod, which is correct
+            // To be investigated: Why isn’t it triggered with "\Psalm\Type::getEmpty()"?!
+            // @phpcs:ignore SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName
```
weirdan, line 90: Not a bug. See https://github.com/vimeo/psalm/blob/e1d1ef32b6a7ff15ffd517c29c205d67e8dfe48f/psalm-baseline.xml#L256-L260.

Psalm relies on textual matches for it's baseline feature.

rarila: Ah, ok. You’ll never stop learning :)

weirdan:
> Some magic with box that does some search-replace with \Composer\Autoload\ClassLoader perhaps?!

Likely. Should we also revert other similar changes, e.g. in `Cli\LanguageServer`?

src/Psalm/Internal/Codebase/Functions.php
```diff
@@ -548,14 +556,14 @@ public function isCallMapFunctionPure(
...
-            || ($args !== null && \count($args) === 0)
+            || ($args !== null && count($args) === 0)
             || ($function_callable->return_type && $function_callable->return_type->isVoid())
         ) {
             return false;
         }
 
         $must_use = $function_id !== 'array_map'
-            || (isset($args[0]) && !$args[0]->value instanceof \PhpParser\Node\Expr\Closure);
+            || (isset($args[0]) && !$args[0]->value instanceof Closure);
```
weirdan, line 566: It would make sense to alias it to something like `ClosureNode`, to prevent confusion.

rarila:
> Likely. Should we also revert other similar changes, e.g. in `Cli\LanguageServer`?

Yes, to be on the save side. The same code is in LanguageServer, Psalter, and Refactor. I guess when it breaks in Psalm it should do so there, too.

src/Psalm/Internal/Analyzer/Statements/ThrowAnalyzer.php
rarila: so running ``psalm --update-baseline`` should do the trick?!

src/Psalm/Internal/PhpVisitor/CloningVisitor.php
```diff
@@ -22,9 +23,9 @@ public function enterNode(Node $node): Node
                 'comments',
                 array_map(
                     /**
-                     * @return \PhpParser\Comment
+                     * @return Comment
                      */
-                    function (\PhpParser\Comment $c): \PhpParser\Comment {
+                    function (Comment $c): Comment {
```
weirdan, line 28: This docblock is actually useless, let's drop it.

src/Psalm/Internal/Analyzer/Statements/ThrowAnalyzer.php
weirdan: It's `--set-baseline` actually. `--update-baseline` does not add new issues to baseline, it only removes issues that are no longer there.

src/Psalm/Internal/PhpVisitor/CloningVisitor.php
rarila: I think we should’t do now. AFAIK there are more of these and there’s a phpcs sniff for it. Plan was to add this sniff (``SlevomatCodingStandard.TypeHints.ReturnTypeHint``) and do an extra PR then

src/Psalm/Internal/Codebase/Functions.php
rarila: We could for now. Especially PhpParser\* is one of those I’d like to find a way to make them shorter but also more distinguishable – and that in a uniform way throughout the code. But I’s like to make some tests before and then open a new issue to discuss my findings

src/Psalm/Internal/PhpVisitor/CloningVisitor.php
weirdan: Ok then.

src/Psalm/Internal/Analyzer/Statements/ThrowAnalyzer.php
rarila: ok…

rarila: Looks good

rarila: Arghl… nope. For some reason it removes some lines. I already was wondering why, thougth some old stuff. Local tests run fine…

rarila: With some cheating… :-)

weirdan (approved):

weirdan:
Thanks!

rarila:
@weirdan Found the cause of the ``\Composer\Autoload\ClassLoader`` problem:

https://github.com/vimeo/psalm/blob/eb4e9dee8689fd19a2d2a70d12bdb1f4230cb558/scoper.inc.php#L22-L26

PR #7066, opened 2021-12-05
Pretty print arrays in LSP hover

joehoyle:
Currently hovering keyed array definitions is very unreadable. Opening this PR to discuss a good solution. I attached some code, but it's not great. The idea is to have this:

Look like this:

Somewhat inspired by how the TypeScript LSP shows object literals:

---

The implementation here is not that good, ideally I think TKeyedArray should have a `prettyPrint` method perhaps. Also, there's other things to consider, like how to show `array{}` printing when then are embedded in `<list>` or `array<>` etc, where line breaks would go, etc.

This "re-parsing" of the type also is not great, this is because the file-maps (specifically the reference map) flattens type to their string representation. Because I _think_ that's potentially done in other threads, I'm assuming that's why.

Then there's the question of syntax highlighting for `array{}` which would help a lot with readability. The issue ultimately is that we rely on the LSP client to syntax highlght based off `<?php $type` code samples. Given that `array{}` is not valid PHP, then we don't send it as `<?php` prefixed, hence no formatting. One optino could be to actually encode it as `<?php [ "key" => "value" ]` style, though I'm not sure how something like `<?php [ "foo" => string ]` would render.

### DIFF
--- a/src/Psalm/Internal/LanguageServer/Server/TextDocument.php
+++ b/src/Psalm/Internal/LanguageServer/Server/TextDocument.php
@@ -231,17 +231,17 @@ class TextDocument
         }
 
         // Pretty print arrays (only).
-        if ( substr( $symbol_information['type'], 0, 6 ) === 'array{' ) {
-            $union = \Psalm\Type::parseString( $symbol_information['type'] );
+        if (substr( $symbol_information['type'], 0, 6 ) === 'array{') {
+            $union = \Psalm\Type::parseString($symbol_information['type']);
             $types = $union->getAtomicTypes();
             if ( count( $types ) === 1 ) {
                 /** @var \Psalm\Type\Atomic\TKeyedArray */
-                $keyed_array = reset( $types );
+                $keyed_array = reset($types);
 
                 $property_strings = array_map(
                     function ($name, \Psalm\Type\Union $type): string {
-                        if ( is_string( $name ) ) {
-                            $name = "'" . $name . "'";
+                        if (is_string($name) && preg_match('/[ "\'\\\\.\n:]/', $name)) {
+                            $name = '\'' . str_replace("\n", '\n', addslashes($name)) . '\'';
                         }
                         return "\t" . $name . ($type->possibly_undefined ? '?' : '') . ': ' . $type;
                     },
@@ -249,7 +249,7 @@ class TextDocument
                     $keyed_array->properties
                 );
 
-                $symbol_information['type'] = "array{\n" . implode(", \n", $property_strings) . "\n}";
+                $symbol_information['type'] = "<?php array{\n" . implode(", \n", $property_strings) . "\n}";
             }
         }
         $content = "```php\n" . $symbol_information['type'] . "\n```";
--- a/src/Psalm/Internal/LanguageServer/Server/TextDocument.php
+++ b/src/Psalm/Internal/LanguageServer/Server/TextDocument.php
@@ -24,12 +24,14 @@ use UnexpectedValueException;
 
 use function addslashes;
 use function array_keys;
+use function array_map;
 use function count;
 use function error_log;
 use function implode;
 use function is_string;
 use function preg_match;
 use function reset;
+use function str_replace;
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/LanguageServer/Server/TextDocument.php
```diff
@@ -230,6 +230,28 @@ public function hover(TextDocumentIdentifier $textDocument, Position $position):
...
 
+        // Pretty print arrays (only).
+        if ( substr( $symbol_information['type'], 0, 6 ) === 'array{' ) {
+            $union = \Psalm\Type::parseString( $symbol_information['type'] );
+            $types = $union->getAtomicTypes();
+            if ( count( $types ) === 1 ) {
+                /** @var \Psalm\Type\Atomic\TKeyedArray */
+                $keyed_array = reset( $types );
+
+                $property_strings = array_map(
```
orklah, line 241: You may want to use TKeyedArray::getId for that, it will handle escaping better and it will reduce the code here

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
```diff
@@ -230,6 +230,28 @@ public function hover(TextDocumentIdentifier $textDocument, Position $position):
             return new Success(null);
         }
 
+        // Pretty print arrays (only).
+        if ( substr( $symbol_information['type'], 0, 6 ) === 'array{' ) {
+            $union = \Psalm\Type::parseString( $symbol_information['type'] );
```
orklah, line 235: We're currently removing FQCN from the code and replacing them with `use`. Can you do the same here?

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
orklah: Oh, I read the code before the text in PR. This is needed for adding a line feed in there. But my comment on escaping still stand (also, did you quote every key on purpose? I believe Psalm can detect which key really need it).

Also, you may want to keep the usage of $keyed_array::KEY because it is sometimes replace by `callable-array`

orklah:
About syntax highlighting, even if there isn't any issue with `string`, you'll start having them if the type is actually something like `callable(object $e, string $a): void`. So you may want to put the type part in quotes anyway

joehoyle:
For syntax highlighting, I tried just marking it as `<?php`, which actually works ok:



this may depend on the LSP client through I guess.

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
joehoyle: I tried `$keyed_array::KEY` but it seems that's interpreted as mixed value, I get `Right operand cannot be mixedPsalmhttps://psalm.dev/059` when trying that. I'm already checking `array{` is the sub-string of the type, so I think it won't actually be possible in this case.

joehoyle: Added the escaping-- it looks like Psalm has no way to distinguish `array{ 0: true }` versus `array{ '0': true }` (I think PHP supports that?) so we don't need to handle this case for now.

joehoyle:
Latest screenshot:

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
orklah: PHP doesn't support that, any numeric key is automatically casted to int in arrays

weirdan:
Just a thought: since `array{}` and friends are supposed to occur in docblocks, what if we output them as `<?php /** array{...} */` and let the client highlight them the way it sees fit?

weirdan:
Well, I'm worried that highlighting an invalid piece of PHP only works by accident.

joehoyle:
@weirdan yeah that's true. I think the safest thing to do would be to keep it as it is, but with multiline / indenting, and punt a "how to highlight psalm types in comments" to another issue

PR #7071, opened 2021-12-05
Try to provide literal int types when possible (fixes #6966)

### LINKED ISSUE #6966: ArgumentTypeCoercion for typecasted variables
When a variable is typecasted and clearly is a positive int, psalm thinks it's generic int:

https://psalm.dev/r/c1e0a9efd4

ricardoboss:
This PR fixes vimeo/psalm#6966.

I added a return type provider for the `round` function, which will now try to return a literal `float`, which can then be used for further analysis.
The second change involves modifying the `CastAnalyzer` to also return a literal `int` when possible. I also tried to simplify the code a bit.


orklah:
Thanks!

You have to enable the return type provider in FunctionReturnTypeProvider first

orklah:
I made an error in my explanation. round only returns floats:

https://3v4l.org/ETvDv

That shouldn't be an issue however, what you have to do to fix #6966 is check if you have a TLiteralInt or a TLiteralFloat in the num_arg. If it is, you should return a TLiteralFloat with the rounded value of what was in num_arg.

ricardoboss:
I need to implement the CastAnalyzer too. So this is should really be a draft PR.

ricardoboss:
I'm struggling to get the actual values with which the `round` method was called. As you can see, I'm trying to get them with `$call_args[1]->value`, which gives a `PhpParser\Node\Expr`. How do I get the actual value?

orklah:
try a `$statements_analyzer->node_data->getType(...)` on that. It may be a complex expression but Psalm should already have inferred it

weirdan:
E.g. https://github.com/vimeo/psalm/blob/25614ec2bd2f30ebb433267d3fcc2074169b83ea/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayFillReturnTypeProvider.php#L27-L30

orklah:
@ricardoboss is this ready for review?

ricardoboss:
Yes! Sorry I didn't notify you...

orklah:
No worries, we'll check that :)

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -59,37 +58,33 @@ public static function analyze(
                     if (!$maybe_type->from_calculation) {
                         self::handleRedundantCast($maybe_type, $statements_analyzer, $stmt);
                     }
+
+                    if ($maybe_type->isSingleStringLiteral()) {
```
orklah, line 62: That contradicts a previous check `$maybe_type->isInt()` above. This should be extracted from this condition

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -59,37 +58,33 @@ public static function analyze(
...
+                        $valid_int_type = new Type\Union([
+                            new Type\Atomic\TLiteralInt((int)$maybe_type->getSingleStringLiteral()->value),
```
orklah, line 64: you can use `Type;;getInt(false, (int)$maybe_type->getSingleStringLiteral()->value);`

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -59,37 +58,33 @@ public static function analyze(
...
-                $type = $valid_int_type ?? Type::getInt();
+            $type = $valid_int_type ?? Type::getInt();
 
-                if ($statements_analyzer->data_flow_graph instanceof VariableUseGraph
-                ) {
-                    $type->parent_nodes = $maybe_type->parent_nodes ?? [];
-                }
-
-                $statements_analyzer->node_data->setType($stmt, $type);
+            if ($statements_analyzer->data_flow_graph instanceof VariableUseGraph
```
orklah, line 81: Please include this block in the conditional `if($maybe_type)`, that way we can remove the `?? []` and it will make more sense

ricardoboss: This is is impossible, because we need the `$type` variable to be set first.

orklah: I'm only talking about the conditional with VariableUseGraph. It doesn't make use of `$type`

ricardoboss:
@orklah can you take a look again?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
orklah: this needs to be fixed. Your maybe_type is either an int or a string literal, not both

orklah (changes requested):
See comment above. Can you add some tests too?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
ricardoboss: Not sure if we are seeing the same version. In my last commit I changed it to an `elseif`.

orklah: yep, you're right, clicked on the wrong link

orklah:
@ricardoboss do you need some help to design the tests?

Could you also rebase to solve the conflicts?

ricardoboss:
Will do a rebase! I haven't had time to write tests yet. Haven't forgot it either.

orklah:
Sorry for not coming back to this earlier, I don't get notifications for commits so I didn't see this change. I'll take a look. Don't hesitate to ping me or leave a comment when you change things

orklah:
Seems fine. Can you rebase and fix the Code Style error please?

orklah (approved):

ricardoboss:
Can you write a test for this? I would like to see how you test this behavior.

I will do a rebase when I get time to do so

orklah:
For the cast one, you can add a test in ValueTest.php in providerValidCodeParse with the code: `$a = (int)'5';` and in the assertions you can check `$a===` => '5'

For the round one, you can add a test in FunctionCallTest.php in providerValidCodeParse  with the code `$a = round(5.111, 2);` and in the assertions you can check `$a===` => 'float(5.11)'

orklah:
Thanks!

PR #7076, opened 2021-12-05
Fixed generic assertions for list and array

### LINKED ISSUE #5294: User defined assertions works incorrectly for generic iterable types
Lets take a look at this https://psalm.dev/r/12ed2787e9
or this https://psalm.dev/r/a4ab076a44
example.

Works fine. As expected.

But if go deeper and try to assert generic type like `list<int>` then psalm lose types:

psalm-assert-if-true: https://psalm.dev/r/58dbc19d8a
psalm-assert: https://psalm.dev/r/4c25ed702c

klimick:
fix #5294


orklah:
I'm a little out of my depth here and I'm not sure what side effects could appear. However, I tested that each change is indeed needed to fix every test case added, the changes are quite small and seems coherent so I'd go for it...

weirdan:
Thanks!

PR #7077, opened 2021-12-05
Make name property of enum cases return non-empty-strings

ricardoboss:
This is a follow-up for #6964 since the changes made there didn't have the desired outcome.


weirdan (approved):

weirdan:
Thanks!

PR #7089, opened 2021-12-07
abs should always return a positive integer

marcosh:


weirdan:
Thanks!

marcosh:
That was fast, thank you!

Issue #7090, opened 2021-12-07
Wrong type combination: list<positive-int> + int = list<positive-int>


klimick:
https://psalm.dev/r/cc223d445f

PR #7093, opened 2021-12-07
don't stop analyzing array, even if we already have a type and we can't create an object like

orklah:
The goal is to fix https://github.com/vimeo/psalm/issues/7090

But seeing the change I had to do, it'll probably break something elsewhere, Let's run the CI!

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -453,10 +453,6 @@ class ArrayAnalyzer
             }
         }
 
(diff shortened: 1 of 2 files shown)


orklah:
Well, that broke a test but in a goodish way.

the string part was ignored before so that's a fix. The weird thing is that `mixed|string` is usually simplified to `mixed`, that would make the test not broken. However, this seems on purpose because here:
https://github.com/vimeo/psalm/blob/c17af4128ee3abbad8a971f10864a75e0dcca828/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php#L90
We set the "allow_mixed_union" param to true in TypeCombiner. It does what it says and it allows mixed in unions. So, I fixed the test that way.

PR #7094, opened 2021-12-07
Fix count_chars stubs

kamil-tekiela:
I am not sure why it doesn't work properly when they are all separated but it seems to work when they are combined with OR operator. 

I don't know if there are tests I should write for this...


orklah:
Not sure either, never really looked at how Psalm finds the correct signature for a specific case.

I'd like some tests on this is possible FunctionCallTest::providerValidCodeParse would be a fine place for those.

In particular, I'd like to try what Psalm outputs for a call when Psalm doesn't know the literal value of $mode (a simple int for example)

kamil-tekiela:
@orklah  Thanks for the hint. Can you help me a little bit with writing the test properly? I am not very familiar with this syntax.

orklah:
yeah, basically, you have a providerValidCodeParse  and a providerInvalidCodeParse.

The first one is used to check that a snippet does not output errors while the second checks that you correctly emit a specific error.

For your tests, providerValidCodeParse is more on point.

You have to create a new offset in the big array. In the array you have:
- the snippet of code
- the second element is an array called assertions. You can put variable as offsets in there and specify the type of the variable as it should be inferred (Psalm will check that each variable have the right type).
- Then you have an array of errors that will be ignored for this test
- And finally you have the php version used by psalm to analyze this particular code

So, for example https://github.com/vimeo/psalm/blob/8fb2e0f8852b2a74db82b1790465411f94ce4399/tests/FunctionCallTest.php#L1738
This test execute the snippet in PHP 8.1, check that $a and $b have the right types and will not tolerate any error.

(If you see `===` after a variable name in assertions, it's used to tell Psalm you want to check the detailed type (ex: `int` without ===, `6` with it)

orklah:
Can you add a test that shows what happens if you provide `int` to the mode? I'm curious to see what that will do

kamil-tekiela:
> Can you add a test that shows what happens if you provide `int` to the mode? I'm curious to see what that will do

Isn't my second test showing this? I pass 5 which produces a warning saying that it should be 3|4

orklah:
you pass the literal `5`, but I'd like to see what happens if you pass a variable that could contain any int.

For example
```
<?php
function scope(int $mode){
    $a = count_chars("foo", $mode);
}
```

orklah:
Well, the error message is kinda weird (incomplete at least) but I guess I'm fine with it.

Can you fix the test and I think we're good to merge

orklah:
Thanks! That's great!

PR #7097, opened 2021-12-08
Remove useless chunk of code in ScopeAnalyzer

muglug:
This simplifies logic when getting control actions for `finally` blocks.

Found when doing a deep inadvertent audit of the codebase (rewriting large chunks of Psalm in Rust).


orklah:
Great! Thanks!

Issue #5482, opened 2021-03-25
Psalm should load extensions based on composer.json instead of local configuration


AndrolGenhald:
I was thinking about this recently although it hasn't directly affected me yet, but I saw it pop up for someone else so I thought I'd create an issue.

> > one of the purposes of the stubs is to enable the static analysis of the code that depends on an extension that is unavailable at runtime.
> 
> Oh, in that case you should instead provide your own stubs. If Psalm were to stub all classes it might miss situations where you're missing an extension.

_Originally posted by @muglug in https://github.com/vimeo/psalm/pull/5477#discussion_r601032639_

With docker it's possible and probably even common for the local php configuration to be different from what's being used to run the code. If I have the PDO extension enabled locally, but composer doesn't require "ext-pdo" then psalm shouldn't load the PDO extension in case my local docker (or production!) environment doesn't use that extension.

It might also be a good idea to add a configuration option to enable or disable specific extensions in psalm.xml.

AndrolGenhald:
Note to self: the decimal extension can be removed from CI if/when this is implemented (#7103). When running tests all extensions should be enabled.

orklah:
This can be closed now

PR #7103, opened 2021-12-08
Allow operator overloading for Decimal extension (fixes #3938).

### LINKED ISSUE #3938: Operator Overloading
I've been using the [Decimal extension](https://php-decimal.io) which overloads the `+`, `-`, `*`, `/`, `%`, and `**` operators. It would be great if Psalm could support this somehow so I don't have to choose between using the less readable `->add()`, `->mul()`, etc everywhere and littering code with `/** @psalm-suppress */`.

I'm ok with any implementation that works, but I'm personally thinking it would be something like annotating a method with 
```
/** 
 * @psalm-overload-operator +
 * @param self|int|string $value
 * @return self
 */
```
or annotating the class with
```
/**
 * @psalm-overload-plus self|int|string -> self
 */
```

It looks like Psalm already supports the operator overloading used by GMP, so hopefully it wouldn't be too much work to add annotations that expose similar functionality.

AndrolGenhald:


AndrolGenhald:
That `Pull Request Labels` check looks new since the last time I contributed, is there any documentation on what that is? I don't see anything in the [contributing](https://github.com/vimeo/psalm/blob/master/docs/contributing/index.md) documentation.

weirdan:
> is there any documentation on what that is?

It's an internal thing. It makes sure maintainers do not forget to assign PR labels that are required for release notes generation.

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
```diff
@@ -611,6 +612,49 @@ private static function analyzeOperands(
...
+                } else {
+                    if ($statements_source && IssueBuffer::accepts(
+                        new InvalidOperand(
+                            "Cannot add Decimal\\Decimal to {$non_decimal_type->getId()}",
+                            new CodeLocation($statements_source, $parent)
+                        ),
+                        $statements_source->getSuppressedIssues()
+                    )) {
+                        // fall through
+                    }
```
weirdan, line 651: There's a new method called `IssueBuffer::maybyAdd()` with the same parameters, but `void` return. It makes sense to use it here instead of `IssueBuffer::accepts()`.

tests/BinaryOperationTest.php
```diff
@@ -85,6 +85,75 @@ public function testGMPOperations(): void
         $this->assertSame($assertions, $actual_vars);
     }
 
+    public function testDecimalOperations(): void
+    {
+        if (!class_exists('Decimal\\Decimal')) {
+            $this->markTestSkipped('Cannot run test, base class "Decimal\\Decimal" does not exist!');
+        }
```
weirdan, line 92: Would you mind adding this extension to the GHA pipeline? It should be rather easy, see https://github.com/shivammathur/setup-php#heavy_plus_sign-php-extension-support

weirdan (changes requested):

tests/BinaryOperationTest.php
AndrolGenhald: I think it's added, but it doesn't seem to want to run since there are changes requested, so I'm going to go ahead and resolve this to see if it works.

AndrolGenhald: The logs for Unit Tests - 5 includes:
```
  PHPUnit 9.5.10 by Sebastian Bergmann and contributors.
  
  Runtime:       PHP 8.0.13
  Configuration: /home/runner/work/psalm/psalm/phpunit.xml.dist
  Random Seed:   1638987602
  
  ................................................................. 65 / 77 ( 84%)
  ............                                                      77 / 77 (100%)
  
  Time: 00:33.093, Memory: 60.00 MB
  
  OK (77 tests, 100 assertions)
```
So I think this is fixed.

AndrolGenhald: Not sure why the Windows one is failing, the logs show that the extension was installed fine. Is it ok to just disable it on the Windows test? Really I'd like to get around to #5482 sometime and make the test always run whether or not the extension is installed, it doesn't actually need the extension anyway, just the stub. Unfortunately right now, the stub is only loaded if the extension is installed.

.github/workflows/windows-ci.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           php-version: '8.0'
           tools: composer:v2
           coverage: none
+          extensions: decimal
```
weirdan, line 26: It doesn't seem to be available on Windows, hence the CI failure.

tests/BinaryOperationTest.php
weirdan: > Is it ok to just disable it on the Windows test? 

Sure, feel free to do that. I just wanted to have that test run in at least one environment, it doesn't have to be all.

.github/workflows/windows-ci.yml
AndrolGenhald: Odd that it would report:
```
==> Setup Extensions 
√ decimal Installed and enabled 
```
if it didn't actually install it. I think as long as it works on Linux though it's probably fine.

weirdan (approved):

weirdan:
Thanks!

PR #7105, opened 2021-12-08
Fix non-empty-literal-string to behave as subtype of non-empty-string (fixes #7095).

### LINKED ISSUE #7095: ReferenceConstraintViolation with non-empty-literal-string in array
`non-empty-literal-string` should be a subtype of `non-empty-string`, but it still results in a `ReferenceConstraintViolation`: https://psalm.dev/r/2756501f64

This seems specific to `ReferenceConstraintViolation`, it acts correctly when a `non-empty-literal-string` is used as an argument that requires `non-empty-string`: https://psalm.dev/r/f8845a750d

In fact, it seems to be something about using it in an array, just having it as the parameter itself works fine: https://psalm.dev/r/c9ad19355c

AndrolGenhald:


orklah:
Thanks! That's great!

PR #7107, opened 2021-12-08
Enable extensions based on composer.json instead of those loaded at runtime (fixes #5482).

### LINKED ISSUE #5482: Psalm should load extensions based on composer.json instead of local configuration

AndrolGenhald:


src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php
```diff
@@ -17,19 +15,20 @@ public static function getClassLikeNames(): array
...
+            && $config->php_extensions["pdo"]
             && isset($call_args[0])
             && ($first_arg_type = $source->getNodeTypeProvider()->getType($call_args[0]->value))
             && $first_arg_type->isSingleIntLiteral()
         ) {
             $fetch_mode = $first_arg_type->getSingleIntLiteral()->value;
 
             switch ($fetch_mode) {
-                case PDO::FETCH_ASSOC: // array<string,scalar|null>|false
+                case 2: // PDO::FETCH_ASSOC - array<string,scalar|null>|false
```
AndrolGenhald, line 31: I don't really like using literals instead of consts, but I think it's worth it to allow psalm to be used for the PDO extension without having the PDO extension loaded at runtime.

stubs/extensions/pdo.phpstub
```diff
@@ -0,0 +1,153 @@
...
+
+class PDO
+{
+    public const PARAM_NULL = 0;
+    public const PARAM_INT = 1;
+    public const PARAM_STR = 2;
+    public const PARAM_LOB = 3;
+    public const PARAM_STMT = 4;
+    public const PARAM_BOOL = 5;
+    // public const PARAM_STR_NATL = 1073741824; since 7.2
```
AndrolGenhald, line 11: I'm not sure how to make this only available in 7.2, I'd rather not make a whole new stub file, but it doesn't look like psalm supports `@since`? Not immediately necessary anyway.

AndrolGenhald:
I figure there should probably be an XML config to enable/disable extensions as well, but this seems to work fine so far.

src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php
orklah: Yeah, we sometimes have to do it anyway when constant value change between php versions anyway

.github/workflows/ci.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           php-version: '8.0'
           tools: composer:v2
           coverage: none
-          extensions: decimal
+          extensions: :pdo
```
AndrolGenhald, line 84: Is there a way to tell what extensions are loaded on the CI build? It looks like I need to disable soap as well and add a stub for `SoapFault`, but what others?

AndrolGenhald:
I was just thinking about `require-dev`. It would be nice if it were possible to load extensions in `require-dev` for specific namespaces such as tests, but as far as I'm aware there's no way to tell Psalm to only load a stub for use in a specific namespace. This might have to be part of a more general `require-dev` support, it would also be nice to exclude loading `require-dev` libraries when inspecting non-dev code.

AndrolGenhald:
Tests all run fine locally with #7110 merged. I'll see to getting an XML config added and then I think it's good to go.

orklah:
Seems like tests are failing. Maybe due to your last commit?

AndrolGenhald:
Odd, they all run fine locally. Must be a PHP version issue.

AndrolGenhald:
I need to read before going and debugging again...
Those tests fail due to a separate bug fixed in #7110, when that's merged I'll rebase and everything should be fine.

AndrolGenhald:
@orklah anything I can do to streamline review of some of these pull requests? I'm seeing several other pull requests that are going to cause conflicts.

weirdan:
Ideally I'd like to merge CI fix and CS change before all other PRs.

orklah:
@AndrolGenhald Thanks for your PRs, it's really great! I'm having a hard time reviewing everything though!

This one is relatively easy to review, but if you could take the habit of summarize your most complex PR to explain the reason behind the biggest changes and challenges you encountered, it would help a lot :)

AndrolGenhald:
Feel free to liberally point out any areas that aren't explained well enough. In the future I'll try to do multiple commits when moving around and deduplicating code sections, I expect being able to see each step in the refactor would help a lot with reviewing.

AndrolGenhald:
@orklah I believe this is ready to go, just blocked by #7110.

AndrolGenhald:
Currently rebasing on master, I'll have to open a pull request against 4.x to add `loadXdebugStub` to the deprecated attributes list. Doesn't look like we're documenting those in UPGRADING.md?

orklah:
No, we only documented config properties removal in UPGRADING

AndrolGenhald:
@orklah I'm trying to update the baseline on 4.x, but every time I try it's just giving me an empty baseline and saying "No Errors" without actually performing a scan. Any idea what's going on there?

orklah:
try a --clear-cache first, maybe it was corrupted somehow.

If it doesn't work, try running with --debug and --no-cache

AndrolGenhald:
Still not sure what's up with it generating an empty baseline, I'll have to look into it later.
The extra errors showing up were due to duplication of `<file src="src/Psalm/config.php">` in the baseline due to a bad merge when cherry-picking, looks like it just uses the second one instead of merging or throwing an exception.

AndrolGenhald:
When this gets merged I'm planning to open an issue for each extension to move stuff from the callmap to extension stubs. I think I'll also open another issue to create a stub generator based on our discussion at #7153, which should help in the effort. I think for the most part these should be fairly easy issues good for a first-time contributor. Any opinions on the matter @orklah?

orklah:
I agree :) Thanks for your involvement!

AndrolGenhald:
@VincentLanglet That's the intention, but we'd also have to get a stub for memcached added for it to work. I should probably write up some documentation on how to add support for an extension, but it should be fairly straightforward.

The reason I proposed the PR is actually because I [sometimes](https://github.com/vimeo/psalm/issues/5482) have extensions enabled locally where I run Psalm that aren't installed in my docker dev environment and it's caused issues once or twice, but it's also an improvement to be able to have optional extensions like you want, and to be able have Psalm analyze projects requiring extensions that aren't actually installed where Psalm is run.

You still have to be careful and make sure to always check the extension is installed before using it, it's not yet possible to have the stub loaded only for a particular scope, so it's all or nothing.

AndrolGenhald:
@orklah I was just thinking I should add something to UPGRADING.md about this, but I think that's more for plugin authors than end users. We should probably have something in the release notes about this if it's released with Psalm 5 since I imagine a lot of people aren't specifying required extensions in their composer.json like they should. It's necessary for libraries, but for applications it can be overlooked without causing trouble until they upgrade Psalm and get a bunch of errors.

orklah:
Shouldn't we merge the extensions from runtime and those in composer? Maybe remove the runtime part on Psalm 6? It would make the transition easier

AndrolGenhald:
Hmm, I suppose it would make the transition easier, but part of the reason I like using composer.json requirements by default is because you could be running Psalm on a local machine with a bunch of extensions installed, but the places you're actually running the code doesn't have them installed.

Personally I think forcing people to fix their composer.json to include extensions is a feature not a bug, even though it could cause some annoyance when upgrading. It's good practice to have `composer check-platform-reqs --no-dev` in deployment scripts, but it's effectively useless if your requirements aren't correct.

src/Psalm/Config.php
```diff
@@ -955,6 +984,40 @@ private static function fromXmlAndPaths(
...
+        $composer_json_path = Composer::getJsonFilePath($config->base_dir);
+
+        $composer_json = null;
+        if (file_exists($composer_json_path)) {
+            if (!$composer_json = json_decode(file_get_contents($composer_json_path), true)) {
+                throw new UnexpectedValueException('Invalid composer.json at ' . $composer_json_path);
+            }
+        }
+        foreach ($config->php_extensions as $ext => $_) {
+            $config->php_extensions[$ext] = isset($composer_json["require"]["ext-$ext"]);
```
orklah, line 996: Est-ce qu'on peut ajouter require-dev et suggest ici? @VincentLanglet use-case is that he has a project where an extension is optional so it can't be added in require, the only place is in suggest

.github/workflows/ci.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           php-version: '8.0'
           tools: composer:v2
           coverage: none
-          extensions: decimal
+          extensions: none, curl, dom, filter, json, libxml, mbstring, openssl, pcre, phar, reflection, simplexml, spl, tokenizer, xml, xmlwriter
```
orklah, line 84: I'd expect this not to be needed anymore?

psalm-baseline.xml
```diff
@@ -18,6 +18,12 @@
       <code>$symbol_parts[1]</code>
     </PossiblyUndefinedIntArrayOffset>
   </file>
+  <file src="src/Psalm/Config.php">
```
orklah, line 21: This should disappear on rebase, right?

src/Psalm/Config.php
```diff
@@ -198,6 +198,8 @@ class Config
     /**
      * Whether or not to load Xdebug stub
      *
+     * @deprecated going to be removed in Psalm 5
```
orklah, line 201: this should be removed here

src/Psalm/Config.php
AndrolGenhald: I mentioned that earlier [here](https://github.com/vimeo/psalm/pull/7107#issuecomment-989871383). I think it'd be really nice to add better require-dev support in general, but that's outside the scope of this PR.
I think this case is likely uncommon enough that having `<enableExtensions>` and `<disableExtensions>` is enough to work around it. I'll think about it some more, maybe it wouldn't be a bad idea to include require-dev and suggests by default.

src/Psalm/Config.php
```diff
@@ -955,6 +984,40 @@ private static function fromXmlAndPaths(
             $base_dir = $current_dir;
         }
 
+        $composer_json_path = Composer::getJsonFilePath($config->base_dir);
+
+        $composer_json = null;
+        if (file_exists($composer_json_path)) {
+            if (!$composer_json = json_decode(file_get_contents($composer_json_path), true)) {
+                throw new UnexpectedValueException('Invalid composer.json at ' . $composer_json_path);
```
orklah, line 992: Should we revert to runtime extension if we failed to parse composer file?

src/Psalm/Config.php
```diff
@@ -955,6 +984,40 @@ private static function fromXmlAndPaths(
...
+            }
+        }
+        foreach ($config->php_extensions as $ext => $_) {
+            $config->php_extensions[$ext] = isset($composer_json["require"]["ext-$ext"]);
+        }
+
+        if ($config->load_xdebug_stub !== null) {
```
orklah, line 999: Should be removed

.github/workflows/ci.yml
AndrolGenhald: Actually this is desired. It defaults to having a bunch of extensions installed, but if you start the list with "none" it installs only the extensions you specify. This way we're not including any default-installed extensions that we don't want.

stubs/extensions/pdo.phpstub
orklah: Yeah, recurrent issue. It should not be an issue if we start generating stubs as we talked

src/Psalm/Config.php
AndrolGenhald: We do the same thing when getting the PHP version from composer.json.

psalm-baseline.xml
AndrolGenhald: Yeah, I'll rebase in a bit.

.github/workflows/ci.yml
orklah: cool, didn't know about that

src/Psalm/Config.php
orklah: I think we should allow what's in require-dev even outside tests. For example, if one has a logging extension, he may want to run it only on dev, but have code using it within optional checks. But you're right, config tag should be enough

AndrolGenhald:
@orklah is there any plan to version the documentation on psalm.dev? Maybe add psalm.dev/docs/v4 and psalm.dev/docs/v5 or something? Not sure if I should be removing deprecated stuff from the documentation or not.

orklah:
No plan for that no. Changes will be little enough I think and we'll probably advise anyone complaning to update versions anyway...

AndrolGenhald:
Should I add back in the documentation for `loadXdebugStub` then with a deprecation/removed note? I worry about people upgrading from an older version of 4 straight to 5, so they might never get the deprecation notice and see what they're supposed to do to fix it.

orklah:
Thanks!

AndrolGenhald:
@orklah

> Should I add back in the documentation for `loadXdebugStub` then with a deprecation/removed note? I worry about people upgrading from an older version of 4 straight to 5, so they might never get the deprecation notice and see what they're supposed to do to fix it.

orklah:
You can add a note in UPGRADING I guess, I'd just like to avoid keeping things in the xml with no uses

AndrolGenhald:
> You can add a note in UPGRADING I guess, I'd just like to avoid keeping things in the xml with no uses

Oh, sorry I wasn't clear enough, I meant the documentation at psalm.dev/docs. If we're not versioning that people will be using Psalm 5 docs for Psalm 4, so we should probably be careful to avoid removing documentation.

orklah:
I'm fine with that :)

weirdan:
> If we're not versioning that people will be using Psalm 5 docs for Psalm 4, so we should probably be careful to avoid removing documentation.

Should the need arise we'll provide documentation for previous versions on readthedocs.org, so feel free to remove outdated parts from master.

PR #7110, opened 2021-12-09
Fix stub parent class not loaded.

AndrolGenhald:
Fix stub file classes not being correctly referenced as parent classes in non-stub files, preventing checks ensuring property and function types match.

### DIFF
--- a/src/Psalm/Internal/Codebase/Populator.php
+++ b/src/Psalm/Internal/Codebase/Populator.php
@@ -529,7 +529,7 @@ class Populator
         ClassLikeStorage $storage,
         ClassLikeStorageProvider $storage_provider,
         array $dependent_classlikes,
-        string $parent_storage_class,
+        string $parent_storage_class
     ): void {
         $parent_storage_class = strtolower(
             $this->classlikes->getUnAliasedName(
--- a/src/Psalm/Internal/Codebase/Populator.php
+++ b/src/Psalm/Internal/Codebase/Populator.php
@@ -86,6 +86,22 @@ class Populator
             $this->populateClassLikeStorage($class_storage);
         }
 
+        // After loading new classes, go over existing classes with
+        // invalid dependencies to see if any of those have been loaded
+        foreach ($this->classlike_storage_provider->getAll() as $class_storage) {
+            foreach ($class_storage->invalid_dependencies as $dependency => $dependency_type) {
+                if ($this->populateDataFromDependency(
+                    $class_storage,
+                    $this->classlike_storage_provider,
+                    [], // TODO
+                    $dependency,
+                    $dependency_type
+                )) {
+                    unset($class_storage->invalid_dependencies[$dependency]);
+                }
+            }
+        }
+
         $this->progress->debug('ClassLikeStorage is populated' . "\n");
 
         $this->progress->debug('FileStorage is populating' . "\n");
@@ -139,10 +155,17 @@ class Populator
 
         $dependent_classlikes[$fq_classlike_name_lc] = true;
 
-        $this->populateDataFromTraits($storage, $storage_provider, $dependent_classlikes);
+        foreach ($storage->used_traits as $used_trait_lc => $_) {
+            $this->populateDataFromTrait($storage, $storage_provider, $dependent_classlikes, $used_trait_lc);
+        }
 
         if ($storage->parent_classes) {
-            $this->populateDataFromParentClass($storage, $storage_provider, $dependent_classlikes);
+            $this->populateDataFromParentClass(
+                $storage,
+                $storage_provider,
+                $dependent_classlikes,
+                reset($storage->parent_classes)
+            );
         }
 
         if (!strpos($fq_classlike_name_lc, '\\')
@@ -154,9 +177,23 @@ class Populator
             $storage->methods['__construct'] = $storage->methods[$fq_classlike_name_lc];
         }
 
-        $this->populateInterfaceDataFromParentInterfaces($storage, $storage_provider, $dependent_classlikes);
+        foreach ($storage->direct_interface_parents as $parent_interface_lc => $_) {
+            $this->populateInterfaceDataFromParentInterface(
+                $storage,
+                $storage_provider,
+                $dependent_classlikes,
+                $parent_interface_lc
+            );
+        }
 
-        $this->populateDataFromImplementedInterfaces($storage, $storage_provider, $dependent_classlikes);
+        foreach ($storage->direct_class_interfaces as $implemented_interface_lc => $_) {
+            $this->populateDataFromImplementedInterface(
+                $storage,
+                $storage_provider,
+                $dependent_classlikes,
+                $implemented_interface_lc
+            );
+        }
 
         if ($storage->location) {
             $file_path = $storage->location->file_path;
@@ -222,7 +259,7 @@ class Populator
             }
         }
 
-        $this->populateOverriddenMethods($storage);
+        $this->populateOverriddenMethods($storage, $storage_provider);
 
         $this->progress->debug('Have populated ' . $storage->name . "\n");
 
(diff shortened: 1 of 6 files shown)


AndrolGenhald:
It looks like stub files are always loaded last, and changing that causes a bunch of tests to break. Not sure how much work it would take to fix that, although that seems like it would be the preferable solution. I'm guessing it was done this way to make sure actual classes were preferred over stubs?
If that doesn't work, there are some places that use placeholder values that are resolved later, so that might work here. I forget what they're called though, so I'm not sure what to search for to look through how that's implemented.

weirdan:
What are you trying to fix anyway?

> I'm guessing it was done this way to make sure actual classes were preferred over stubs?

It's the opposite, actually. Stubs are loaded last to make sure they _override_ everything they need to override.

AndrolGenhald:
> What are you trying to fix anyway?

`SystemClass::foo` takes `(int, string)`, `Foo::foo` takes `(string, string)`. There should be a `ImplementedParamTypeMismatch` issue.

I noticed this in #7107 after disabling the soap extension, `MethodSignatureTest::testExtendDocblockParamTypeWithWrongDocblockParam` started failing.

weirdan:
> `SystemClass::foo` takes `(int, string)`, `Foo::foo` takes `(string, string)`. There should be a `ImplementedParamTypeMismatch` issue.

The issue is not supposed to be emitted for stubs. Otherwise you would not be able to override wrong signatures coming from libraries/reflection - and it's one of the main problems solved by stubs.

AndrolGenhald:
> The issue is not supposed to be emitted for stubs.

It shouldn't be emitted when checking the stub itself, but shouldn't it be emitted when checking a non-stubbed class that extends a stubbed class?

weirdan:
Ok, gotcha. Yes, it should be emitted in this case.

src/Psalm/Internal/Codebase/Populator.php
```diff
@@ -374,74 +466,73 @@ private function populateOverriddenMethods(
         }
     }
 
-    private function populateDataFromTraits(
+    private function populateDataFromTrait(
```
AndrolGenhald, line 469: I pulled the loops out of these functions because in a previous version I was running them individually on specific dependencies. I later realized that line 295 already reruns the entire thing anyway, but I left the changes here because there were other improvements done as well.

src/Psalm/Internal/Codebase/Populator.php
```diff
@@ -264,15 +285,29 @@ private function populateClassLikeStorage(ClassLikeStorage $storage, array $depe
...
+                                $dependency->location
+                            )
+                        );
+                    }
+
+                    continue;
+                }
+
                 $dependency->populated = false;
+                unset($dependency->invalid_dependencies[$fq_classlike_name_lc]);
```
AndrolGenhald, line 310: This is the important part.

src/Psalm/Internal/Codebase/Populator.php
```diff
@@ -281,8 +316,65 @@ private function populateClassLikeStorage(ClassLikeStorage $storage, array $depe
     }
 
     private function populateOverriddenMethods(
-        ClassLikeStorage $storage
+        ClassLikeStorage $storage,
+        ClassLikeStorageProvider $storage_provider
     ): void {
```
AndrolGenhald, line 321: This was pulled out of `populateDataFromImplementedInterfaces` when I removed the loop from it, so that it could be done afterward. I think it's actually easier to understand this way, one issue that took me a while to figure out is that `$storage->class_implements` may already have had items set elsewhere (from Reflection I believe), so this is actually populating overridden methods that were set elsewhere as well as from this class.

src/Psalm/Internal/Codebase/Populator.php
```diff
@@ -601,255 +691,152 @@ function ($constant) {
         $storage->pseudo_methods += $parent_storage->pseudo_methods;
     }
 
-    private function populateInterfaceDataFromParentInterfaces(
+    private function populateInterfaceData(
```
AndrolGenhald, line 694: This is common code pulled out of `populateInterfaceDataFromParentInterfaces` and `populateDataFromImplementedInterfaces`.

src/Psalm/Internal/Codebase/Populator.php
```diff
@@ -181,10 +181,17 @@ private function populateClassLikeStorage(ClassLikeStorage $storage, array $depe
...
+            $this->populateDataFromTrait($storage, $storage_provider, $dependent_classlikes, $used_trait_lc);
+        }
 
         if ($storage->parent_classes) {
-            $this->populateDataFromParentClass($storage, $storage_provider, $dependent_classlikes);
+            $this->populateDataFromParentClass(
+                $storage,
+                $storage_provider,
+                $dependent_classlikes,
+                reset($storage->parent_classes)
```
orklah, line 193: was there a reason to not loop over parent classes here?

orklah: nevermind, that was the previous behavior, as I saw below

AndrolGenhald: That corresponds with https://github.com/vimeo/psalm/pull/7110/files/8ceb0e2077a07d01932ae350b8edac58b11eb9bf#diff-76daa06f11322c5f544716df14e1c64a90a9ee87e749369914f2a3dbbf2e5f5bL483, it was moved directly without changing. I assume it's fine because `$storage->parent_classes` is supposed to be ordered with the directly extended class first, and that class's storage should already have been populated with its parent data.

AndrolGenhald: That corresponds with https://github.com/vimeo/psalm/pull/7110/files/8ceb0e2077a07d01932ae350b8edac58b11eb9bf#diff-76daa06f11322c5f544716df14e1c64a90a9ee87e749369914f2a3dbbf2e5f5bL483.
It only needs to populate the direct parent, because the direct parent will already have been populated by its parent.

src/Psalm/Internal/Codebase/Populator.php
orklah: So, if I get this part right, the stub reader will try to read a stub, stumbles upon an unknown dependency, so it will add it in invalid_dependencies and invalid_class_storages. Then when it's done, it loops over invalid_class_storages again to check if stubs read later found the dependency. If you find a circular dependency like A -> B and B -> A, you emit an issue.

So what about A -> B, B -> C, C -> A? Would that be detected?

AndrolGenhald: I think so, but I'll update the test.

AndrolGenhald: I guess it doesn't, I'll try to get that fixed in a bit.

AndrolGenhald: That should fix it, nice catch!

orklah:
Seems ok to me. If @weirdan approves, we can merge

AndrolGenhald:
@weirdan Any update on this?

src/Psalm/Storage/ClassLikeStorage.php
```diff
@@ -373,7 +373,7 @@ class ClassLikeStorage
     public $initialized_properties = [];
 
     /**
-     * @var array<string>
+     * @var array<string, true>
```
weirdan, line 376: This is BC break. Good thing that the PR already targets master.

weirdan:
Sorry, I have been busy with v5 changes. I'll review it this week.

src/Psalm/Storage/ClassLikeStorage.php
AndrolGenhald: Oh, good point, I'd completely forgotten about plugins. I'll take more care about that in the future.

AndrolGenhald: I don't think that change is really needed anymore, but unless you object I don't really see a need to change it back. It seems better to me for it to be a set rather than a list, and `array<string, true>` seems like the closest thing PHP has to a native string set without using the ds extension.
Or maybe it'd be better to change it back and methodically update a lot of these types of properties to be more set-like in a separate pull request instead?

orklah:
@AndrolGenhald can you rebase this? We'll merge it like that I think.

AndrolGenhald:
@orklah Done. There are some CS errors locally due to empty files, looks like `TAssertionEmpty` and `TAssertionFalsy` got accidentally undeleted in your description refactor.

orklah:
> TAssertionEmpty and TAssertionFalsy got accidentally undeleted

fixed, sorry

orklah:
Thanks!

PR #7113, opened 2021-12-09
Added support for first-class callables

### LINKED ISSUE #6412: Closures for callables with `myfunc(...)`
https://psalm.dev/r/a33ad9cfa6

https://wiki.php.net/rfc/first_class_callable_syntax

`nikic/php-parser` support for 8.1 is tracked here: https://github.com/nikic/PHP-Parser/issues/801

trowski:
Hopefully the approach here is appropriate. First-class callables are treated as a function call with a single placeholder argument, which necessitates special handling in the call analyzers to swap out the the return value to a closure and skip analyzing function arguments.

Closes #6412.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -20,6 +20,7 @@ use Psalm\Internal\DataFlow\TaintSink;
 use Psalm\Internal\MethodIdentifier;
 use Psalm\Internal\Type\Comparator\CallableTypeComparator;
 use Psalm\Internal\Type\TemplateResult;
+use Psalm\Internal\Type\TypeCombiner;
 use Psalm\Issue\DeprecatedFunction;
 use Psalm\Issue\ImpureFunctionCall;
 use Psalm\Issue\InvalidFunctionCall;
@@ -85,6 +86,7 @@ class FunctionCallAnalyzer extends CallAnalyzer
         $real_stmt = $stmt;
 
         if ($function_name instanceof PhpParser\Node\Name
+            && !$stmt->isFirstClassCallable()
             && isset($stmt->getArgs()[0])
             && !$stmt->getArgs()[0]->unpack
         ) {
@@ -143,6 +145,7 @@ class FunctionCallAnalyzer extends CallAnalyzer
             }
         }
 
+        $is_first_class_callable = $stmt->isFirstClassCallable();
         $set_inside_conditional = false;
 
         if ($function_name instanceof PhpParser\Node\Name
@@ -153,14 +156,16 @@ class FunctionCallAnalyzer extends CallAnalyzer
             $set_inside_conditional = true;
         }
 
-        ArgumentsAnalyzer::analyze(
-            $statements_analyzer,
-            $stmt->getArgs(),
-            $function_call_info->function_params,
-            $function_call_info->function_id,
-            $function_call_info->allow_named_args,
-            $context
-        );
+        if (!$is_first_class_callable) {
+            ArgumentsAnalyzer::analyze(
+                $statements_analyzer,
+                $stmt->getArgs(),
+                $function_call_info->function_params,
+                $function_call_info->function_id,
+                $function_call_info->allow_named_args,
+                $context
+            );
+        }
 
         if ($set_inside_conditional) {
             $context->inside_conditional = false;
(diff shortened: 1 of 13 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
```diff
@@ -235,6 +243,45 @@ public static function analyze(
             );
 
             $config->eventDispatcher->dispatchAfterEveryFunctionCallAnalysis($event);
+
+            if ($is_first_class_callable) {
```
orklah, line 247: Why are you returning early here?

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
```diff
@@ -235,6 +243,45 @@ public static function analyze(
             );
```
orklah, line 243: the AfterEveryFunctionCallAnalysisEvent is the plugin hook. I feel like this should stay at the bottom of the Analyzer.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
```diff
@@ -789,7 +837,7 @@ private static function analyzeInvokeCall(
         $fake_method_call = new VirtualMethodCall(
             $function_name,
             new VirtualIdentifier('__invoke', $function_name->getAttributes()),
-            $stmt->getArgs()
+            $stmt->isFirstClassCallable() ? [new PhpParser\Node\VariadicPlaceholder()] : $stmt->getArgs()
```
orklah, line 840: Please only create Virtual nodes from Psalm. VirtualVariadicPlaceholder may not exist yet, but it should be created

orklah: I wonder if retrieving $stmt->args would be enough here, technically it already return the real VariadicPlaceholder when isFirstClassCallable is true

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
```diff
@@ -58,7 +59,26 @@ public static function fetch(
         $stmt_type = null;
         $config = $codebase->config;
 
-        if ($codebase->functions->return_type_provider->has($function_id)) {
+        if ($stmt->isFirstClassCallable()) {
```
orklah, line 62: Maybe this should be after the return type provider check, would that work? The risk otherwise, you prevent using return type providers when using first class callables (this limits what plugins can do, and even some Psalm internal providers)

orklah:
Wow, that's impressive work!

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
trowski: If I understand correctly, we don't want to use the return type provider as we're not looking for the return type identified by `$function_id`, rather returning a closure instead.

orklah: Oh yeah, of course. We'll have to check that the type providers are correctly called when resolving the closure call then

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
trowski: I wasn't sure about this, but it seemed the code below was generally analyzing a call to the given function, which isn't actually happening. Some guidance on what still needs to be called from the code below would be appreciated.

orklah: you wouldn't go that far anyway, you have another `if ($is_first_class_callable)` just below that create the closure. I just found weird to return before that only when this condition `if ($function_name instanceof PhpParser\Node\Name && $function_call_info->function_id) {` is met and not otherwise

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
orklah: Psalm was using $stmt->args everywhere a few weeks ago. I replaced them all when introducing support for PHP-Parser 4.13: https://github.com/vimeo/psalm/pull/6628

weirdan:
Looks good to me in principle, but it seems it broke closure handling somewhere (hence all the failures).

trowski:
@weirdan I reordered a couple of calls in `ExistingAtomicMethodCallAnalyzer::analyze` I didn't realize were dependent upon each other. After restoring the order and guarding a arg check call if the call being analyzed is a first-class callable, everything appears to be working as it should.

orklah:
Looks good! Let's see how this behave in the wild :D!

PR #7119, opened 2021-12-10
Replace deprecated File Path Utility

rarila:
Replace deprecated [File Path Utility](https://github.com/webmozart/path-util) with suggested [Symfony Filesystem Component](https://github.com/symfony/filesystem)

>Deprecation
>This package has been merged into the Symfony Filesystem Component 5.4. It is not maintained anymore.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,8 +35,7 @@
         "openlss/lib-array2xml": "^1.0",
         "sebastian/diff": "^3.0 || ^4.0",
         "symfony/console": "^3.4.17 || ^4.1.6 || ^5.0 || ^6.0",
-        "symfony/filesystem": "^6.0",
-        "webmozart/path-util": "^2.3"
+        "symfony/filesystem": "^6.0"
     },
     "provide": {
         "psalm/psalm": "self.version"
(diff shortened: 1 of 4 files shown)


rarila:
Ouch: ``symfony/filesystem 5.4`` => ``"php": ">=7.2.5"`` :collision:

I wonder why composer allowed that locally with Psalms restriction to 7.1.
So I probably have to bite the bullet and install PHP 7.1 just for Psalm. :unamused:

I guess we are out of luck here?! :disappointed_relieved:
So when is Psalm 5 planned? :wink:

weirdan:
> So when is Psalm 5 planned?

2022 Q1 ~and it will support PHP 4~

rarila:
No PHP 4 support?! :scream:
I’ll quit!

weirdan:
> So I probably have to bite the bullet and install PHP 7.1 just for Psalm.

[phpenv](https://github.com/phpenv/phpenv) + [php-build](https://github.com/php-build/php-build) combo makes it quite easy, after you have build dependencies installed.

rarila:
Thanks! But I guess it’s less a problem of not being easy (just a download on Windows, and for this purpose the standard php.ini would probably be ok and on Debian a simple ``apt-get install php7.1 php7.1-cli`` would do the job) but still you have to do it. It’s more about having a piece of code lying around which is two years over EOL and which you probably only need in rare cases like this PR.

Anyway, no problem. I`ll keep this PR and simply rebase it for the V5 branch then. Not long until 2022 Q1 :relaxed:

PR #7121, opened 2021-12-10
Fix static::CONST type not being inferred.

AndrolGenhald:
Wish I knew why that check was there in the first place, but it doesn't seem to be necessary.


AndrolGenhald:
Need to remember to clear the psalm cache after making changes to psalm...

AndrolGenhald:
On second thought, that actually makes sense, if it's static, it could be changed by a child.

Issue #5743, opened 2021-05-10
Callable type information is not preserved in class constants


danog:
See https://psalm.dev/r/ac4c096eb4

orklah:
It's actually worse than that. I think @var doesn't work at all above a constant:
https://psalm.dev/r/6e0994afa7

orklah:
Closing this as a duplicate of https://github.com/vimeo/psalm/issues/942

AndrolGenhald:
@orklah Would you reopen this since it's not likely to be fixed with #942?

orklah:
@AndrolGenhald I'm surprised this was not fixed though. Is there something special here?

AndrolGenhald:
We don't currently resolve literal strings to callables: https://psalm.dev/r/13a7a514f3. Not sure if that's an intentional choice or not.

orklah:
Oh yeah, right, forgot the inferred type has priority

AndrolGenhald:
Opened #7547.

Issue #5589, opened 2021-04-06
Support class const covariance checks


AndrolGenhald:
Non-private class constants should be covariant, so a child can't change the type of a parent's const: https://psalm.dev/r/e287a07e3f

Same reasoning as #4184, but since consts can't be changed there's no need for contravariance, hence they should be covariant rather than invariant.

`@var` or `@const` should also be used to specify the type (`@var` seems wrong but [seems to be](https://stackoverflow.com/q/6706051) what's used other places).

orklah:
blocked by https://github.com/vimeo/psalm/issues/942

AndrolGenhald:
I believe properties must be invariant because a parent method can be called on a child class and change a child property, potentially violating its type on the child (eg parent defines property as `array`, child defines it as `array<int>`, parent method assigns `array<string>` to the property). Since constants can't be changed, there's no risk of the parent method changing it and violating the type given in the child.

To quote the first paragraph of the issue I linked:
> Non-private properties in child classes need to be covariant for reading, like function returns and simultaneously contravariant for writing, like function params. That adds up to them needing to be invariant, or have exactly the same type as the property in the parent class.

Consts also need to be covariant for reading, but since they can't be written they have no need for contravariance.

orklah:
Right, I had a slight déja-vu about this discussion :)

I'm not sure I follow your argument that class constants should behave like return types.

They're more related to property types, no? And those are invariants

(I have a possible legit explanation why we should consider them covariant, but I need to make sure I'm not just rationalizing here)

AndrolGenhald:
@orklah My argument is [here](https://github.com/vimeo/psalm/issues/5589#issuecomment-950243787). Since properties can be written there's a risk of passing a child to a function taking the parent as an argument, and that function could change the property to something that matches the parent type but not the child type. Constants can't be written so there's no such risk.

If I'm right, the same also applies to read-only properties, they could be made covariant, but I don't care enough to do it personally.

orklah:
Ok, I arrived to the same conclusion, I'm convinced :)

muglug:
Looks like this is now caught?

AndrolGenhald:
Yes, looks like GitHub stopped recognizing "fixes #..." comments recently.

AndrolGenhald:
@orklah Since you changed literal strings to use single quotes, maybe types should be quoted with double quotes now? https://psalm.dev/r/e287a07e3f

orklah:
[...]

PR #7123, opened 2021-12-10
Support type annotations for class consts (fixes #942).

### LINKED ISSUE #942: `@var` annotation is ignored for class constants
https://getpsalm.org/r/03a2f11ca3
```php
<?php
  /** @param void ...$_p*/
  function type(...$_p): void {}

  class C {
    /** @var int[] */
    const N = [];
  }

  type(C::N);
```
Expected: should complain about `array<mixed,int> provided`
Actual: complains about `array<empty,empty> provided`

There's at least one usage of `@var` on constants in Psalm's tests: https://github.com/vimeo/psalm/blob/0ef71a49cbbc18fc948a4af113eced6eb8a2839a/tests/ConstantTest.php#L66-L69

Additionally, psalm seems to ignore `@var` and const initialization contradiction:
```php
<?php
class C {
  /** @var int */
  public const N = "aaaa"; // should produce issues, but doesn't
}
```

AndrolGenhald:


src/Psalm/Issue/UnresolvableConstant.php
```diff
@@ -0,0 +1,8 @@
+<?php
+namespace Psalm\Issue;
+
+class UnresolvableConstant extends CodeIssue
+{
+    public const ERROR_LEVEL = -1;
```
AndrolGenhald, line 6: Should this be set to something else?

src/Psalm/Storage/ClassConstantStorage.php
```diff
@@ -8,11 +8,27 @@
 
 class ClassConstantStorage
 {
+    use CustomMetadataTrait;
+
+    /**
+     * @var ?CodeLocation
+     */
+    public $type_location;
```
AndrolGenhald, line 16: Isn't used yet, but I feel like it should be. Should I add a psalm-suppress and hope someone gets around to using it, or can someone point me to where it should be used?

tests/ConstantTest.php
```diff
@@ -1482,6 +1513,92 @@ class A {
...
+                         */
+                        public function bar(int $key): string
+                        {
+                            return static::BAR[$key];
+                        }
+                    }
+                ',
+                'error_message' => 'UnresolvableConstant',
+            ],
+            'SKIPPED-keyofSelfConstDoesntImplyKeyofStaticConst' => [
```
AndrolGenhald, line 1584: Would be nice to do, but psalm isn't catching this either: https://psalm.dev/r/4efbbbda75
I thought there was an option to be strict about array keys somewhere, but I'm not seeing it at the moment, and I don't know how I would enable that for the test.

src/Psalm/Issue/UnresolvableConstant.php
orklah: I believe -1 means that the error will be emitted no matter what level the user is on. It's usually reserved to obvious errors or issue behind a specific flag.

I'm not sure I see yet in what scenario this issue could be emitted, but you have to decide a level based on severity

AndrolGenhald:
I just tested #5743 and it seems to work, but if you don't add an argument, it reports "Too few arguments for  - expecting 1 but saw 0". Seems like the `in_array` isn't getting into the message somewhere.

src/Psalm/Issue/UnresolvableConstant.php
weirdan: -1 denotes insuppressible issue, -2 - issues controlled by a flag (like `--find-dead-code`).

So -1 should be reserved for certain fatal errors.

src/Psalm/Storage/ClassConstantStorage.php
orklah: This feels like something that will be used to solve https://github.com/vimeo/psalm/issues/5589

weirdan: I think it could be used for, e.g. highlighting, where issues refer to the constant type as opposed to the constant itself.

tests/ConstantTest.php
```diff
@@ -1482,6 +1513,92 @@ class A {
...
+                        /**
+                         * @return value-of<static::BAR>
+                         */
+                        public function bar(): string
+                        {
+                            return static::BAR[0];
+                        }
+                    }
+                ',
+                'error_message' => 'UnresolvableConstant',
```
weirdan, line 1548: Should this really be `UnresolvableConstant`? I think I'd find it confusing.

src/Psalm/Issue/UnresolvableConstant.php
AndrolGenhald: It should be emitted when `key-of` or `value-of` is used on a constant that can't be resolved, which I think probably should always be emitted, as it's indicative of a typo or other issue with the type annotation, and it's an annotation that's specifically for static analysis. I never use any of the less strict error levels though, so I'm not confident I should be the one deciding.

src/Psalm/Storage/ClassConstantStorage.php
AndrolGenhald: > I think it could be used for, e.g. highlighting, where issues refer to the constant type as opposed to the constant itself.

Yeah, I'm just not sure how to go about finding all the places that should be happening.

tests/ConstantTest.php
weirdan: I believe you might need those two:
```xsd
        <xs:attribute name="ensureArrayIntOffsetsExist" type="xs:boolean" default="false" />
        <xs:attribute name="ensureArrayStringOffsetsExist" type="xs:boolean" default="false" />
```

To enable it you would have to extract the case into a full-fledged test method, like this: https://github.com/vimeo/psalm/blob/3cce69189f748235a613c63d3d2609fe76f079c5/tests/ArrayAccessTest.php#L17-L36

tests/ConstantTest.php
AndrolGenhald: > Should this really be `UnresolvableConstant`? I think I'd find it confusing.

Yeah, perhaps a bit confusing, but I made a note of it in the `UnresolvableConstant` documentation. I didn't really want to have another issue for `IllegalConstant` or something like that. I'm happy to change the name if you can come up with something that intuitively encompasses both usages, or would you like to have a separate issue?

AndrolGenhald:
So here's something I'm not sure of: should we default to the annotated type or the inferred type?
I mainly see two different uses for type annotation:

1. Annotating the type that a constant should always have. The constant may be changed in the future, but it will always have this type.
2. Annotating the type that child classes must be covariant with.

Unfortunately, I could see places where one might want the behavior of 2 without the behavior of 1. Should this be configurable? Should we just go with 1 and say if you annotate it, that's what psalm is going to use?

Contrived illustration of where 2 without 1 may be useful: https://psalm.dev/r/8a9abce968

weirdan:
Is there actually a difference between 1 and 2 though?

AndrolGenhald:
> Is there actually a difference between 1 and 2 though?

In the example I gave it should work fine with 2, but with 1 it will have an issue for `return self::BAR;` because `self::BAR` will be `string` rather than `"bar"`.

AndrolGenhald:
#5743 is broken again after I fixed a mistake. I meant to use the inferred type except when using late static binding, rather than use the annotated type except for with `self::`. I think that should be reopened and fixed separately, depending on how we decide to handle inferred vs annotated type.

My preference is to use the inferred type wherever possible and explicitly opt out of that, because it allows a lot more static analysis, but right now there's no way to opt out and say you want the annotated type to be used for a particular constant.

I'm tempted to say that the fix for #5743 should actually be to allow interpreting `"in_array"` as a callable, but I think that was decided against previously, probably for a good reason. Perhaps we should just leave it closed and say that first class callables in PHP 8.1 are the correct solution?

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
```diff
@@ -510,6 +513,37 @@ public static function getTypeFromValue($value): Type\Union
         }
     }
 
+    /**
+     * Gets the Psalm literal type from a particular value
+     *
+     * @param array|scalar|null $value
+     *
+     */
+    public static function getLiteralTypeFromValue($value): Type\Union
```
AndrolGenhald, line 522: Not sure if this function already exists somewhere and I missed it. Should this go somewhere else? Maybe as `Type::getLiteral`?

tests/ConstantTest.php
AndrolGenhald: Maybe `UnresolvableLiteralConstant`?

weirdan:
I'd say we should follow the principle of least surprise and just use the declared type, and also flag cases where the constant value contradicts the declared type. As far as inheritance is concerned, constant types should be covariant.

Do you see any problems with this approach?

AndrolGenhald:
> I'd say we should follow the principle of least surprise and just use the declared type

Well, `key-of` and `value-of` currently still use the inferred type, and I think it'd be more surprising if they didn't, so at the very least that's going to feel a bit inconsistent.

It also causes annoyances like this: https://psalm.dev/r/fd97d21f4e

> and also flag cases where the constant value contradicts the declared type

I meant to do that and completely forgot about it.

tests/ConstantTest.php
```diff
@@ -586,6 +618,8 @@ class B extends A {
 
                     class Clazz {
                         /**
+                         * @var 0|1
```
AndrolGenhald, line 621: I thought about making it resolve the assignment to `1` instead of `0|1`, but I figure the only reason for a ternary is that the author is planning to change the constant later, in which case I think it might make more sense to force them to use the type annotation. Thoughts?

AndrolGenhald: I changed my mind, I think it would be a good idea to infer this, even though it's redundant.

src/Psalm/Storage/ClassConstantStorage.php
AndrolGenhald: I decided to suppress it for now, I expect to use it in the issues emitted for #5589.

src/Psalm/Storage/ClassConstantStorage.php
```diff
@@ -6,13 +6,32 @@
 use Psalm\Internal\Scanner\UnresolvedConstantComponent;
 use Psalm\Type;
 
+/**
+ * @psalm-suppress PossiblyUnusedProperty
```
AndrolGenhald, line 10: Shouldn't this be at the property level? Should I do a separate pull request for that, or is it intentional?

orklah:
@AndrolGenhald what is the status of this? Are you waiting for another PR, for a review or do you still need to work on it?

AndrolGenhald:
@orklah Ready for review, I just rebased on master and got everything up-to-date.

#7396 might be able to make key/value-of usable for static const arrays.

src/Psalm/Storage/ClassConstantStorage.php
```diff
@@ -60,7 +60,6 @@ class ClassConstantStorage
 
     /**
      * @var list<AttributeStorage>
-     * @psalm-suppress PossiblyUnusedProperty
```
AndrolGenhald, line 44: This actually seems like a false negative that was introduced somewhere.

AndrolGenhald: Actually I guess it's just 08413d7, and the reason it's not a false negative is because it's not marked internal.

AndrolGenhald:
I think the only thing here that might be controversial is what I mentioned [here](https://github.com/vimeo/psalm/pull/7123#issuecomment-991335268), I have it defaulting to using the inferred type whenever possible, so the `@var` annotation is really just used to make sure the literal type matches (useful for complicated nested arrays), and once #7154 is merged it's used to make sure children have covariant constant types.

I can see where some people might want to override the constant type, but that would prevent analysis like [this](https://psalm.dev/r/8a9abce968) where you want to constrain the types a child can override the constant with, but you still want to use the inferred type for other analysis (returning a literal in that example).

I think using the inferred type wherever possible is more useful, and it more closely matches the way psalm handles constants now when it ignores the `@var` annotation. If preventing usage of the inferred type is desired I think we should open a separate issue for that and maybe add some sort of annotation for it.

orklah:
> I have it defaulting to using the inferred type whenever possible

So, just to be sure, I see 3 possible types for a class const.
- the type of the parent const
- the type of the content of the const
- the type of the @var

For me, const should be invariant, like properties, so the type of the parent const must be the same as the child.
However, this can be hard because const content will be literals so unless you inherit with the same exact literal (which makes no sense), Psalm should scream.

I think this is where @var will be useful to define a standard type among all childs/parent (for example, if you have a parent that defined "hello" and a child "hello2", both should be @var'ed to 'string'

So Psalm just need to make sure that the type from the @var is consistent with the content of the const. And then apply invariant checks.

Do I make sense?

AndrolGenhald:
@orklah We discussed this before at #5589, do you want to move the discussion over there? I think it makes more sense for const types to be covariant rather than invariant, see my [last comment](https://github.com/vimeo/psalm/issues/5589#issuecomment-950344124) there with the examples.

orklah:
Ok, now that we settled we're indeed covariant, it makes things easier I think.

As long as the phpdoc type is consistent with the content of the constant, the phpdoc is allowed. Then the phpdoc is always the default because it can be more precise

AndrolGenhald:
> Then the phpdoc is always the default because it can be more precise

Well, not quite, the inferred type is still the default, see [here](https://github.com/vimeo/psalm/pull/7123#issuecomment-1018007679). I think that's the one major unresolved issue here, but I'd rather open a separate issue for that and decide what to do about it later.

orklah:
Ok for handling it later, but then, what's the point of being able to put a @var on a constant if you're always overriding it with the inferred type? Am I missing something?

AndrolGenhald:
@orklah Here's a better example: https://psalm.dev/r/ba711169b3
The `@var` annotation is useful to make sure the constant has a particular form, but using the inferred type is still desired elsewhere.

Or here: https://psalm.dev/r/2d3831ed42
The `@var` annotation makes sure children override the constant with a type that is covariant, but it's still nice to know that `Bar::FOOBAR[1]` is `4` rather than `int`.

orklah:
Ok, fine by me then :)

Thanks for taking the time to explain :p

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
```diff
@@ -512,6 +517,57 @@ public static function getTypeFromValue($value): Union
...
     }
 
+    /**
+     * Gets the Psalm literal type from a particular value
+     *
+     * @param array|scalar|null $value
+     * @throws InvalidArgumentException
+     *
+     */
+    public static function getLiteralTypeFromValue($value, bool $sealed_array = true): Type\Union
```
orklah, line 527: Maybe getLiteralTypeFromScalarValue could be reused or refactored or something?

AndrolGenhald: I overlooked that one, I'll take a look at it in a bit. One thing I'm not too happy about is that allowing `array` here could allow a non-scalar to be passed, but I don't think we have a way to type "possibly deeply nested array with only scalars allowed".

AndrolGenhald:
> Thanks for taking the time to explain :p

No problem, convincing someone is a really good sanity check.

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
AndrolGenhald: Done. I'm half tempted to try adding something like `TNestedArray` where `nested-array<scalar>` would be like `array<scalar|array<scalar|array...>>`, but I'm not sure how useful it would be in practice.

orklah: It would be a hell to add I think. It means you'll have to implement the logic to compare any array to the nested array.
That means:
- can any array be equal to the nested one
- can any array be contained in the nested and the other way around
- nested assertions probably

etc...

orklah:
Thanks!

orklah:
That's weird, PHPUnit test and Psalm itself started to output the same kind of error after I merged. They weren't on CI.

Could you take a look at it to see if they can be fixed?

AndrolGenhald:
@orklah I assume you're referring to `src/Psalm/DocComment.php:21:18: InvalidConstantAssignmentValue: Psalm\DocComment::PSALM_ANNOTATIONS with declared type non-empty-array<int<1, max>, string> cannot be assigned type array{'return', 'param',...`? My bet is something changed between the last CI run and now that didn't cause a conflict but messed up the type comparison. There's no reason I can see that it should think the type is `non-empty-array<int<1, max>, string>`.

orklah:
Mmh, it may actually be one of my PR when I removed positive-int. I'll try to take a look at it

AndrolGenhald:
I think I'm closing in on it, I'm trying to figure out why it's getting the wrong type set in ClassLikeNodeScanner.

AndrolGenhald:
@orklah TypeCombiner.php:1195:
```
value_types['int'] = $all_nonnegative
    ? TIntRange(1, null)
    : TNonspecificLiteralInt();`
```
Should be 0 there, 0 is non-negative.

orklah:
Yeah, I just saw that too. It was new TPositiveInt before my change, it just made it appear. I'll fix that

AndrolGenhald:
Ah, I see, there's a `$had_zero` there to add the literal 0 back into it to give `0|positive-int`, but that got removed as well, so instead of `0|int<1,max>` it just resulted in `int<1,max>`.

orklah:
Thanks btw! You're pretty efficient at finding bugs in Psalm!

AndrolGenhald:
No problem, I enjoy debugging :slightly_smiling_face:

AndrolGenhald:
@orklah is [this](https://psalm.dev/r/3531bf3469) a bug on psalm.dev? When I run it in a test locally the exception is correctly caught and an issue is raised, rather than causing psalm to crash.

PR #7129, opened 2021-12-11
Detect unused properties that are written to inside arrays

muglug:
In my ongoing deep audit of the Psalm codebase I found a property that wasn't being read, and tracked it down to an issue detecting unused properties with array types.

This detects those, and removes similar properties from Psalm's codebase.

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -96,14 +96,6 @@ class Codebase
     public $use_referencing_locations = [];
 
     /**
-     * A map of file names to the classes that they contain explicit references to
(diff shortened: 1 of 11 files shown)


orklah:
That's cool! Thanks!

weirdan:
Those public properties are part of the public API, so they should be deprecated first and then dropped in a major release.

@orklah @muglug I don't think it should be merged just yet.

muglug:
True, but I think the only property that has any chance of being used used is the constant attributes one. It’s your call to revert the second commit, but I’m happy to PR the return of the constant attributes property for a smaller diff.

PR #7130, opened 2021-12-11
Class property issue suppression fixes.

AndrolGenhald:
Fix @psalm-suppress and @psalm-allow-private-mutation being ignored if nothing else is in the docblock.
Fix @psalm-suppress not allowing extra text after the issue name.
Fix PossiblyUnusedProperty and UnusedProperty suppression not working at the property level.
Fix MissingPropertyType suppression not working at the property level.


AndrolGenhald:
There are probably other issues that should be suppressable at the property level, these are just the ones I noticed while I was fixing things.

orklah:
Thanks! That's great!

PR #7134, opened 2021-12-11
Use getSingleAtomic() instead of array_values(getAtomicTypes())[0].

AndrolGenhald:


orklah:
Oh, nice. I think we also have some reset() and array_pop/array_unshift that does that too

AndrolGenhald:
> Oh, nice. I think we also have some reset() and array_pop/array_unshift that does that too

I don't think so. If there are any they have to be assigned to a variable first since those take the argument by reference, so it's a lot harder to find.

orklah:
Well, I usually use regex for that kind of searches.
Try things like
`array_shift\(\$.+type`
`array_pop\(\$.+type`
`reset\(\$.+type`

There's quite a few :)

But no worries if you don't want to do those, we'll do that later

AndrolGenhald:
Actually there is one, I'll fix it in a minute.

orklah:
Thanks!

PR #7135, opened 2021-12-11
Fix type inference of closure params

### LINKED ISSUE #7128: Type inference of closure params
Psalm can infer all closure param types when at least one param is generic: https://psalm.dev/r/6373b867d2
And can't infer types when there are no generic params: https://psalm.dev/r/aeb2cda3aa

What is the reason for this?

klimick:
Fix #7128


orklah:
Thanks!

PR #7136, opened 2021-12-11
Forbid positional arg after named arg

### LINKED ISSUE #7104: Forbid use positional argument after named argument
This code will fail with fatal error: https://3v4l.org/14hEF#v8.0.13
```php
<?php

final class Person
{
    public function __construct(
        public string $name,
        public int $age,
    ) { }
}

new Person(name: '', 0);
```

It would be nice if Psalm report about it:
https://psalm.dev/r/b0ed47490d

klimick:
Closes #7104

### DIFF
--- a/tests/ArgTest.php
+++ b/tests/ArgTest.php
@@ -1,6 +1,7 @@
 <?php
 namespace Psalm\Tests;
 
+use Person;
 use Psalm\Tests\Traits\InvalidCodeAnalysisTestTrait;
 use Psalm\Tests\Traits\ValidCodeAnalysisTestTrait;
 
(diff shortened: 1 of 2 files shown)


tests/ArgTest.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace Psalm\Tests;
 
+use Person;
```
orklah, line 4: Not sure how that will play out, this is not an existing class

klimick: Sorry for my carelessness. Phpstorm auto import)
Fixed.

orklah:
Thanks! That's great!

orklah:
Whoops, Almost merged, but you have still a CS issue. Can you fix that first?

weirdan:
Thanks!

PR #7137, opened 2021-12-11
fix some more expressions that can be replaced by getSingleAtomic

orklah:
cc @AndrolGenhald :)

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
@@ -34,7 +34,6 @@ use function array_merge;
 use function array_pop;
 use function array_values;
 use function count;
-use function current;
 use function get_class;
 
 class CastAnalyzer
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
@@ -61,9 +61,8 @@ class CastAnalyzer
                     }
                 }
 
-                $maybe = $maybe_type->getAtomicTypes();
-
-                if (count($maybe) === 1 && current($maybe) instanceof Type\Atomic\TBool) {
+                if (count($maybe_type->getAtomicTypes()) === 1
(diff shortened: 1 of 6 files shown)


AndrolGenhald:
Ah, `current` is a search I didn't think to do. The others I'm guessing took a lot more manual looking through the code.

I wonder how difficult it would be to write a plugin to detect this issue, might be a good learning exercise. I've often thought it would be nice to be able to refactor more complex issues like this in my own code without so much manual searching.

orklah:
I'm quite used to search things like that, PHPStorm makes this quite easy :)

Plugins are quite easy when you know exactly what you're looking for and it's limited to a single expression. The hard part comes when you have to look for many things at once

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -61,9 +60,8 @@ public static function analyze(
                     }
                 }
 
-                $maybe = $maybe_type->getAtomicTypes();
-
-                if (count($maybe) === 1 && current($maybe) instanceof Type\Atomic\TBool) {
+                if (count($maybe_type->getAtomicTypes()) === 1
```
weirdan, line 63: `count()` could be replaced with `isSingle()`, couldn't it?

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -639,15 +629,13 @@ private static function handleArrayItem(
                 return false;
             }
 
-            $dim_atomic_types = $dim_type->getAtomicTypes();
-
-            if (count($dim_atomic_types) > 1
+            if (count($dim_type->getAtomicTypes()) > 1
```
weirdan, line 632: and this one could be spelled `!$dim_type->isSingle()`

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
orklah: I thought about it too, but isSingle is actually not so straightforward. For example, it will return 1 when two or more literals of the same types are present.

It's usually fine, but I'm not perfectly sure it won't create edge cases

PR #7140, opened 2021-12-12
Fixed compatibility with SoapClient

yethee:
### SoapClient

Some methods can return `null` but currently these cases are not detected by psalm. As instance, `__getLastResponse()` returns `null` when connection was closed before receiving a response.

https://psalm.dev/r/e7364a92c6

If we remove the check for `null`, then can get an error of incompatibility types, at runtime:

```
TypeError: preg_match() expects parameter 2 to be string, null given
```

**Affected methods**:

* [`__getFunctions`](https://www.php.net/manual/en/soapclient.getfunctions.php)
* [`__getLastRequest`](https://www.php.net/manual/en/soapclient.getlastrequest.php)
* [`__getLastRequestHeaders`](https://www.php.net/manual/en/soapclient.getlastrequestheaders.php)
* [`__getLastResponse`](https://www.php.net/manual/en/soapclient.getlastresponse.php)
* [`__getLastResponseHeaders`](https://www.php.net/manual/en/soapclient.getlastresponseheaders.php)
* [`__getTypes`](https://www.php.net/manual/en/soapclient.gettypes.php)

### SoapFault

Value of property `detail` can be mixed.

```xml
<SOAP-ENV:Fault>
  <faultcode>SOAP-ENV:Client</faultcode>
  <faultstring>Bad request</faultstring>
  <detail>
    <ns3:FaultResponse xmlns:ns3="http://direct.yandex.com/api/v5/general">
      <requestId>1010101010101010101</requestId>
      <errorCode>8000</errorCode>
      <errorDetail>Error message</errorDetail>
    </ns3:FaultResponse>
  </detail>
</SOAP-ENV:Fault>
```

```
print_r($fault->detail);

stdClass Object
(
    [FaultResponse] => stdClass Object
        (
            [requestId] => 1010101010101010101
            [errorCode] => 8000
            [errorDetail] => Error message
        )
)
```


orklah:
@veewee You recently submitted changes in soap stubs, do you mind reviewing this PR if you're familiar with the extension?

dictionaries/CallMap.php
```diff
@@ -12266,7 +12266,7 @@
 'SoapFault::getPrevious' => ['?Exception|?Throwable'],
 'SoapFault::getTrace' => ['list<array<string,mixed>>'],
 'SoapFault::getTraceAsString' => ['string'],
-'SoapFault::SoapFault' => ['object', 'faultcode'=>'string', 'faultstring'=>'string', 'faultactor='=>'string', 'detail='=>'string', 'faultname='=>'string', 'headerfault='=>'string'],
+'SoapFault::SoapFault' => ['object', 'faultcode'=>'string', 'faultstring'=>'string', 'faultactor='=>'string', 'detail='=>'mixed', 'faultname='=>'string', 'headerfault='=>'string'],
```
veewee, line 12269: Changing the details is ok, but the other props also don't seem to match the signature of [SoapFault::__construct()](https://www.php.net/manual/en/soapfault.construct.php)

veewee:
@orklah  Seems correct and documented:
https://www.php.net/manual/en/class.soapclient.php

I can imageine that they have always worked like this.

These can return null in non-wsdl mode or when overriding construct and calling these functions withouth the WSDL being parsed:

* __getTypes()
* __getFunctions()

```php
$soap = new SoapClient(null, [
    'location' => 'http://127.0.0.1',
    'uri' => 'http://127.0.0.1',
]);

var_dump($soap->__getFunctions());
// > null
var_dump($soap->__getTypes());
// > null
```


These will return false if trace is disabled (or if the connection got a timeout - you can simulate this with the ini setting `default_socket_timeout`)

* __getLastRequest()
* __getLastRequestHeaders()
* __getLastResponse()
* __getLastResponseHeaders()

```php
$soap = new SoapClient('http://www.dneonline.com/calculator.asmx?wsdl', [
    'trace' => 0,
]);
$result = $soap->Add(['intA' => 1, 'intB' => 2]);

var_dump($soap->__getLastRequest());
// > null
var_dump($soap->__getLastRequestHeaders());
// > null
var_dump($soap->__getLastResponse());
// > null
var_dump($soap->__getLastResponse());
// > null
```

The SoapFault indeed takes a ?mixed details prop.
It looks like the other properties are not fully in sync though.
See : https://www.php.net/manual/en/soapfault.construct.php

dictionaries/CallMap.php
yethee: I have updated the constructor signature according to the documentation. And also the property map has been updated.

weirdan:
test-with-real-projects/PHPUnit failure seems legit. You may want to PR a null check to PHPUnit, but for the purposes of our CI suite we will just baseline that error when it's merged.

dictionaries/CallMap.php
veewee: @orklah  : Do the names of these parameters matter much?
Since It is actually:

```php
#[Pure]
public function __construct(
        array|string|null $code,
        string $string, 
        ?string $actor = null,
        mixed $details = null,
        ?string $name = null,
        mixed $headerFault = null
    ) {
```

See https://github.com/php/php-src/blob/master/ext/soap/soap.stub.php#L38

weirdan: Names do matter, for when Psalm checks calls with named arguments.

yethee: I've updated the signatures for [SoapFault::__construct()](https://github.com/php/php-src/blob/1d54097435bc09a287cb50b6e166eccba19f2925/ext/soap/soap.stub.php#L38) and [SoapFault::SoapFault()](https://github.com/php/php-src/blob/6d5f2ba78d85f24e69eee5c7c3948171b9fc3b05/ext/soap/soap.c#L855-L856).

orklah:
Thanks!

@weirdan can you update the baseline for phpunit?

weirdan:
> @weirdan can you update the baseline for phpunit?

@orklah done.

PR #7141, opened 2021-12-12
in_array returns false in strict mode if types are incompatibles

### LINKED ISSUE #5552: in_array 3rd param true - TypeDoesNotContainType error but not vice versa
https://psalm.dev/r/f3e22a8103

This should give an error for TypeDoesNotContainType, since it's always type string.
It seems to work sometimes (getting the error) but sometimes it doesn't.

Also the vice versa should work:
if Psalm is not sure that the type is always the same, it should require the third parameter to be true.
As currently it can happen that psalm gives an error for TypeDoesNotContainType, the dev removes "true" 3rd arg.
Then some changes happen, e.g. adding int types to the array - but the dev isn't aware of that and suddenly the condition is wrong. In these cases psalm should require true for the 3rd arg.

mathroc:
fixes #5552

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/InArrayReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/InArrayReturnTypeProvider.php
@@ -1,0 +1,82 @@
+<?php
+namespace Psalm\Internal\Provider\ReturnTypeProvider;
+
+use Psalm\Internal\Type\Comparator\UnionTypeComparator;
+use Psalm\Plugin\EventHandler\Event\FunctionReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\FunctionReturnTypeProviderInterface;
+use Psalm\Type;
+use Psalm\Type\Atomic\TArray;
+use Psalm\Type\Atomic\TKeyedArray;
+use Psalm\Type\Atomic\TList;
+
+class InArrayReturnTypeProvider implements FunctionReturnTypeProviderInterface
+{
+    /**
+     * @return array<lowercase-string>
+     */
+    public static function getFunctionIds(): array
+    {
+        return ['in_array'];
+    }
+
+    public static function getFunctionReturnType(FunctionReturnTypeProviderEvent $event): Type\Union
+    {
+        $call_args = $event->getCallArgs();
+        $bool = Type::getBool();
+
+        if (!isset($call_args[0]) || !isset($call_args[1])) {
+            return $bool;
+        }
+
+        $needle_type = $event->getStatementsSource()->getNodeTypeProvider()->getType($call_args[0]->value);
+        $haystack_type = $event->getStatementsSource()->getNodeTypeProvider()->getType($call_args[1]->value);
+
+        if ($needle_type === null || $haystack_type === null) {
+            return $bool;
+        }
+
+        $false = Type::getFalse();
+        $false->from_docblock = $bool->from_docblock = $needle_type->from_docblock || $haystack_type->from_docblock;
+
+        if (!isset($call_args[2])) {
+            return $bool;
+        }
+
+        $strict_type = $event->getStatementsSource()->getNodeTypeProvider()->getType($call_args[2]->value);
+
+        if ($strict_type === null || !$strict_type->isTrue()) {
+            return $bool;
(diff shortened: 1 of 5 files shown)


tests/TypeReconciliation/InArrayTest.php
```diff
@@ -377,24 +377,7 @@ function assertInArray($x, $y) {
 
                         return $x;
                     }',
-                'error_message' => 'RedundantConditionGivenDocblockType - src' . DIRECTORY_SEPARATOR . 'somefile.php:8:30 - Docblock-defined type int for $x is never string',
-            ],
-            'assertNegatedInArrayOfNotIntersectingTypeTriggersMixedReturnStatement' => [
```
mathroc, line 382: would become the same as assertNegatedInArrayOfNotIntersectingTypeTriggersRedundantCondition L317

src/Psalm/Internal/Type/TypeCombiner.php
```diff
@@ -93,10 +92,6 @@ public static function combine(
         bool $allow_mixed_union = true,
         int $literal_limit = 500
     ): Union {
-        if (in_array(null, $types, true)) {
```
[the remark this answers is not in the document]
orklah: Well, It's Psalm job to check for bad types given to methods so I'd say this is not supposed to happen so we can remove it

src/Psalm/Internal/Provider/ReturnTypeProvider/InArrayReturnTypeProvider.php
```diff
@@ -0,0 +1,81 @@
...
+        $call_args = $event->getCallArgs();
+
+        if (!isset($call_args[2])) {
+            return Type::getBool();
+        }
+
+        $third_arg = $call_args[2]->value;
+
+        if (!$third_arg instanceof ConstFetch
+            || strtolower($third_arg->name->parts[0]) !== 'true'
```
weirdan, line 36: This would work with literal `true` only, but is there any reason for that limitation? Can't we fetch the expression type from the node type provider and use that instead?

weirdan (commented):
Not sure about all those changes to expected messages. Shouldn't we mark the return type with `from_docblock` flag when any of the source types are coming from docblock, like this:
```php
$ret = Type::getBool();
$ret->from_docblock = $needle_type->from_docblock || $haystack_type->from_docblock;
return $ret;
```
?

src/Psalm/Internal/Provider/ReturnTypeProvider/InArrayReturnTypeProvider.php
mathroc: > This would work with literal true only, but is there any reason for that limitation? Can't we fetch the expression type from the node type provider and use that instead?

I'm looking at the Psalm internals for the first time, and I forgot to come back to this bit after I discovered the node type provider. I'll change that right away!

> Not sure about all those changes to expected messages. Shouldn't we mark the return type with from_docblock flag when any of the source types are coming from docblock, like this:

Didn't know about that either, I'll try and see what I can do :+1:

orklah:
> Shouldn't we mark the return type with from_docblock flag when any of the source types are coming from docblock

I'm not sure about that. I don't think that's what Psalm does when we use stubs with conditional returns, it seems weird to do that here only

src/Psalm/Internal/Provider/ReturnTypeProvider/InArrayReturnTypeProvider.php
mathroc: I think it's ok now, let me know if anything else can be improved

weirdan:
> I don't think that's what Psalm does when we use stubs with conditional returns, it seems weird to do that here only

Given the following code:
```php
/** @param list<int> $a */
function f(array $a, string $b): void {
   if (in_array($b, $a, true) {}
}
```
What's more appropriate here, `TypeDoesNotContainType` or `DocblockTypeContradiction`?

orklah:
The second one seems more logical. We may want to check how conditional returns are applied and make sure the flag goes through too...

weirdan (approved):
Looks good to me. @orklah, do you think this is good to merge?

orklah:
yep :) Thanks @mathroc !

PR #7143, opened 2021-12-12
Post-assertions in loops

### LINKED ISSUE #7038: Assertion inside loop does not work

klimick:
Closes #7038

### DIFF
--- a/src/Psalm/Issue/CodeIssue.php
+++ b/src/Psalm/Issue/CodeIssue.php
@@ -86,20 +86,23 @@ abstract class CodeIssue
         return $this->message;
     }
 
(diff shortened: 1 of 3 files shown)


weirdan:
Thanks!

PR #7145, opened 2021-12-13
Fix ldap_set_rebind_proc signature

villfa:
Hi,

This PR fixes a false positive about `ldap_set_rebind_proc()`:
```
Argument 2 of ldap_set_rebind_proc expects string, callable provided
```

See:
* Documentation: https://www.php.net/manual/en/function.ldap-set-rebind-proc.php
* phpstorm-stubs: https://github.com/JetBrains/phpstorm-stubs/blob/58c6b9d0308d6371be5b820c7063d1b7f8b9db1c/ldap/ldap.php#L1257


weirdan:
@villfa you need to update `Callmap_historical` as well (it contains signatures as they were in 7.0).

villfa:
Thanks, it's done.

orklah:
Thanks!

PR #7150, opened 2021-12-13
"No errors found!" message is now printed within a nice green block

SMAtaurRahman:
Last week, I introduced psalm+phpstan to some juniors and I noticed they liked visual feedback of phpstan better than psalm.
I have tried to find why is so.

When errors are found, psalm kinda punishes our eyes with all the red texts of error message, but when we fixed them all, it just prints some dull looking texts. I had to guide the juniors multiple time to just ignore all the texts and look for "No errors found!" among them.
Only way to know psalm run status at a glance is to check if terminal has red color text or not.

I think, psalm should have a success page with green (or similar) color which kinda works as a reward for fixing all the errors. We all love the dopamine hit.

So, I've prepared this PR that will print a nice green block when there are no errors to report.

`printSuccessMessage` could be made as more flexible, such as accepting arbitrary message and dynamically calculate padding size. But I think that would increase the complexity unnecessarily  since we are printing a hard-coded string anyway.

Let me know what do you think. Thanks.

### DIFF
--- a/src/Psalm/IssueBuffer.php
+++ b/src/Psalm/IssueBuffer.php
@@ -686,7 +686,7 @@ class IssueBuffer
                     : $error_count . ' errors'
                 ) . ' found' . "\n";
             } else {
-                echo 'No errors found!' . "\n";
+                self::printSuccessMessage();
             }
 
             $show_info = $project_analyzer->stdout_report_options->show_info;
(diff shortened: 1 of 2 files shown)


weirdan:
It's a bit hard to read with the terminal color scheme I'm using (Solarized Dark):

image

SMAtaurRahman:
> It's a bit hard to read with the terminal color scheme I'm using (Solarized Dark):
> 
> image

Yep, I was not so sure about this part as well. Maybe we can use white/black as text color?

weirdan:
Looking great now:
image

weirdan:
Thanks!

PR #7154, opened 2021-12-13
Improve class constant static analysis

### LINKED ISSUE #5589: Support class const covariance checks

### LINKED ISSUE #7108: Add issue for overriding const from interface
Prior to PHP 8.1 overriding a const from an interface is not allowed, psalm should catch this: https://psalm.dev/r/b680b5f7f3 https://3v4l.org/Vbt63

AndrolGenhald:
Add class const covariance support (fixes #5589).
Add check for overriding const from interface in PHP < 8.1 (fixes #7108).
Add check for ambiguous const inheritance.

### DIFF
--- a/docs/running_psalm/issues/AmbiguousConstantInheritance.md
+++ b/docs/running_psalm/issues/AmbiguousConstantInheritance.md
@@ -14,7 +14,6 @@ interface Foo
 interface Bar
 {
     /**
-     * @psalm-suppress OverriddenInterfaceConstant
      * @var non-empty-string
      */
     public const CONSTANT='bar';
--- a/docs/running_psalm/issues/AmbiguousConstantInheritance.md
+++ b/docs/running_psalm/issues/AmbiguousConstantInheritance.md
@@ -1,0 +1,42 @@
+# AmbiguousConstantInheritance
+
+Emitted when a constant is inherited from multiple sources.
+
+```php
+<?php
+
+interface Foo
+{
+    /** @var non-empty-string */
+    public const CONSTANT='foo';
+}
+
+interface Bar
+{
+    /**
+     * @psalm-suppress OverriddenInterfaceConstant
+     * @var non-empty-string
+     */
+    public const CONSTANT='bar';
+}
+
+interface Baz extends Foo, Bar {}
+```
+
+```php
+<?php
+
+interface Foo
+{
+    /** @var non-empty-string */
+    public const CONSTANT='foo';
+}
+
+class Bar
+{
+    /** @var non-empty-string */
+    public const CONSTANT='bar';
+}
+
+class Baz extends Bar implements Foo {}
+```
--- a/src/Psalm/Internal/Analyzer/InterfaceAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/InterfaceAnalyzer.php
@@ -7,11 +7,16 @@ use LogicException;
 use PhpParser;
 use Psalm\CodeLocation;
 use Psalm\Context;
+use Psalm\FileManipulation;
+use Psalm\Internal\Analyzer\Statements\Expression\ClassConstAnalyzer;
+use Psalm\Internal\FileManipulation\FileManipulationBuffer;
 use Psalm\Internal\Provider\NodeDataProvider;
 use Psalm\Issue\ParseError;
 use Psalm\Issue\UndefinedInterface;
 use Psalm\IssueBuffer;
 use UnexpectedValueException;
+
+use function strtolower;
 
 /**
  * @internal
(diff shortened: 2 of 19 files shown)


AndrolGenhald:
Depends on #7110 and #7123, I'll rebase once those are merged. (I actually can't remember why I used #7110 here, that might have been an accident; It could probably be pulled out if necessary).

docs/running_psalm/issues/AmbiguousConstantInheritance.md
```diff
@@ -0,0 +1,42 @@
...
+interface Foo
+{
+    /** @var non-empty-string */
+    public const CONSTANT='foo';
+}
+
+interface Bar
+{
+    /**
+     * @psalm-suppress OverriddenInterfaceConstant
```
AndrolGenhald, line 17: Is there a way to tell the documentation test that this should use PHP 8.1?

src/Psalm/Internal/Analyzer/InterfaceAnalyzer.php
```diff
@@ -140,7 +148,35 @@ public function analyze(): void
...
+                                )
+                            ];
+
+                            FileManipulationBuffer::add(
+                                $this->getFilePath(),
+                                $file_manipulations
+                            );
+                        }
+                    }
+                }
```
AndrolGenhald, line 173: Copied this part from ClassAnalyzer. Haven't tested it, but I assume it's probably fine.

docs/running_psalm/issues/AmbiguousConstantInheritance.md
weirdan: See https://github.com/vimeo/psalm/blob/832fc35d8da6e5bb60f059ebf5cb681b4ec2dba5/tests/DocumentationTest.php#L312-L318

AndrolGenhald:
@orklah Can you rerun CI for this now that the int range issue is fixed? I'm not sure if I'm able to do that myself without pushing something.

orklah:
Done, except for circleCI, I have no dominion there 😄

AndrolGenhald:
Oh, I didn't realize I still had things to fix, I thought I'd handled that already.

orklah:
I was gonna post the same thing as your edit :p we're both idiots then 😄

AndrolGenhald:
@orklah any idea what's up with [this](https://github.com/vimeo/psalm/runs/4943797532?check_suite_focus=true#step:7:201)? I've run into it when testing locally too, but only sometimes.

orklah:
I'm not quite sure, but I think there is something related to the order of the tests

The failing test expects this exception and message: https://github.com/vimeo/psalm/blob/558208e44e8c31477b1b09b8434be202174b1cd6/tests/Config/ConfigTest.php#L1534

most of the time it works, but sometimes, there is a message coming from here:
https://github.com/vimeo/psalm/blob/b924032850df54dd05ffae409a709820fd6e688e/src/Psalm/Internal/ErrorHandler.php#L70

Normally, the error handler is installed here: https://github.com/vimeo/psalm/blob/558208e44e8c31477b1b09b8434be202174b1cd6/tests/Config/ConfigTest.php#L1536

but there's also one here: https://github.com/vimeo/psalm/blob/558208e44e8c31477b1b09b8434be202174b1cd6/tests/Config/ConfigTest.php#L1555

What I don't get is that the exception I would expect most of the time is actually RuntimeException, not ConfigException

AndrolGenhald:
Regardless, this should be ready for review and I think everything works correctly. Hopefully it's a bit easier to understand than some of the others :laughing:

docs/running_psalm/issues/AmbiguousConstantInheritance.md
```diff
@@ -0,0 +1,41 @@
...
+
+interface Bar
+{
+    /**
+     * @var non-empty-string
+     */
+    public const CONSTANT='bar';
+}
+
+interface Baz extends Foo, Bar {}
```
orklah, line 22: Is this an issue when the type are compatibles?

orklah (approved):

docs/running_psalm/issues/AmbiguousConstantInheritance.md
AndrolGenhald: It's an issue even when it's the same exact thing: https://3v4l.org/tjYML

orklah: oh, I thought "ambiguous" was too tame a word for something that broke at compile time, I was wrong!

orklah:
Thanks!

PR #7155, opened 2021-12-14
Convertion of partials to full imports

rarila:
I finally have converted "all" partial to full imports.

I was looking at all cases but I don’t think readability suffers at any case, as mostly they are unambiguous due prefixing or postfixing or just by naming or context.

One exception is the ``PhpParser`` namespace which will need an extra issue for discussion and so is left out here.


rarila:
I guess the error is related to #7153?!

src/Psalm/Codebase.php
```diff
@@ -1124,7 +1136,7 @@ public function getSymbolLocation(string $file_path, string $symbol): ?CodeLocat
 
             $file_contents = $this->getFileContents($file_path);
 
-            return new CodeLocation\Raw(
+            return new Raw(
```
weirdan, line 1139: This is an example of where the readability does suffer. Could be renamed to `RawLocation` though.

weirdan:
I'm halfway through the review, and this is actually not too bad.

weirdan (approved):
Note to self: review ParseTree, UnresolvedConstant and CodeLocation class naming.

src/Psalm/Codebase.php
rarila: Yep, ``Raw`` is quite inexpressively.
The good thing is, it is used only 4 times. Anyway, ``DocblockTypeLocation`` and ``ParseErrorLocation`` are postfixed, so this should really be ``RawLocation``.

But I think this should go into an extra PR.

rarila:
> Note to self: review ParseTree, UnresolvedConstant and CodeLocation class naming.

Those two should probably also be added to that list:
```
Psalm\Internal\LanguageServer\Client\TextDocument // => ClientTextDocument ?!
Psalm\Internal\LanguageServer\Server\TextDocument // => ServerTextDocument ?!
```

src/Psalm/Codebase.php
weirdan: > But I think this should go into an extra PR.

Of course.

orklah:
Thanks!

PR #7157, opened 2021-12-14
ISSUE-5962 Fixed wrong line number for @method annotations

### LINKED ISSUE #5962: Wrong line numbers reported in some cases
I'm not quite sure what triggers it, but here's a minimal test case: https://psalm.dev/r/ed2836fc4b

The error analysis seems correct enough, but the issue is the *location reference*. Psalm reports the issue as being on line 3, column 12. But here line 3 is actually the comment open block, and doesn't *have* 12 characters. I'm pretty sure this is intended to refer to line 4.

This is a particular issue if you're trying to feed the results into Sonarqube, since the scanner entirely crashes when given an invalid column reference.

If you compare to https://psalm.dev/r/76ee877256 it appears it's not an off by one issue, but rather that it's reporting the error in the doc block as belonging to the first line of the docblock. I'd assume that's wrong (it should report the line of the `@method` annotation, not the line the comment starts on), but it's got to be wrong for it to mix a line reference with a column reference from an entirely different line, right?

nowaja:
Resolves #5962

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
@@ -23,6 +23,7 @@ use Psalm\Internal\Type\TypeTokenizer;
 use function array_key_first;
 use function array_shift;
 use function count;
+use function explode;
 use function implode;
 use function in_array;
 use function preg_match;
@@ -182,8 +183,7 @@ class ClassLikeDocblockParser
             }
         }
 
-        if (isset($parsed_docblock->tags['psalm-yield'])
-        ) {
+        if (isset($parsed_docblock->tags['psalm-yield'])) {
             $yield = reset($parsed_docblock->tags['psalm-yield']);
 
             $info->yield = trim(preg_replace('@^[ \t]*\*@m', '', $yield));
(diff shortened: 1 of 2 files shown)


orklah:
Why did you change min php version? Was there a feature you needed?

orklah:
If this is about CI failure on isStatic, it's unrelated, it will be fixed this afternoon

nowaja:
> If this is about CI failure on isStatic, it's unrelated, it will be fixed this afternoon

Yes, i tried if that could be related. Reverted the change. Thanks.

orklah:
Can you explain why you changed something here instead of when we report an issue? It may be the correct way, but that would not have been my first instinct

weirdan:
I'm pretty sure it's a correct approach. We're in foreach over method entries, and the code as it was was creating virtual methods with the _same offset_, which can't be right.

Fixing it at the emit point would likely require reparsing the doc comment at that point, potentially multiple times. Which, if you ask, would be kinda silly and wasteful.

nowaja:
Thank you @weirdan, you were faster. I also tried different solution but exactly what you mentioned had happened. I had to change multiple places, it was really hard to follow. This solution seemed the best.

orklah:
Thanks!

Issue #7057, opened 2021-12-04
Suboptimal discovery of project files


dvz:
The scanning of project files may be suboptimal for projects with a large number of ignored files/directories within project directories:
```xml
<projectFiles>
    <directory name="." />
    <ignoreFiles>
        <directory name=".git" />
        <directory name="other-big-subdirectory" />
    </ignoreFiles>
</projectFiles>
```

This may manifest in a long delay between invoking the CLI and its first output, and may be responsible for the major part of Psalm's execution time when using cache.

- [ ] **Ignored directories iterated over recursively**
  The recursive iteration in:
  - [`FileProvider::getFilesInDir()`](https://github.com/vimeo/psalm/blob/16a61a758e44158c19c779c8b0a1c2143e40d83e/src/Psalm/Internal/Provider/FileProvider.php#L118-L131)
  - [`FileFilter::loadFromArray()`](https://github.com/vimeo/psalm/blob/16a61a758e44158c19c779c8b0a1c2143e40d83e/src/Psalm/Config/FileFilter.php#L186-L210)

  doesn't take ignored locations into account, and all leaves (files) are unnecessarily processed in the loop.

- [ ] **Repetitive discovery of project files**
  The discovery of project files happens multiple times within the same search scope:
  - https://github.com/vimeo/psalm/blob/16a61a758e44158c19c779c8b0a1c2143e40d83e/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php#L295
  - https://github.com/vimeo/psalm/blob/16a61a758e44158c19c779c8b0a1c2143e40d83e/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php#L1070
  - https://github.com/vimeo/psalm/blob/16a61a758e44158c19c779c8b0a1c2143e40d83e/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php#L1102

  To improve performance, the list could be cached.

- [ ] **Psalm cache directory not ignored**
  Psalm's own cache directory is not excluded automatically when located within `projectFiles`, which may be unintuitive. When not ignored, it may significantly contribute to the execution time.

- [ ] **Late filtering during file discovery**
[...]

PR #7161, opened 2021-12-14
Improve project files discovery performance

dvz:
Partially resolves .#7057

### DIFF
--- a/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
@@ -308,7 +308,11 @@ class ProjectAnalyzer
         $file_extensions = $this->config->getFileExtensions();
 
         foreach ($this->config->getProjectDirectories() as $dir_name) {
-            $file_paths = $this->file_provider->getFilesInDir($dir_name, $file_extensions);
+            $file_paths = $this->file_provider->getFilesInDir(
+                $dir_name,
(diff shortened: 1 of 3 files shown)


orklah:
Thanks!

src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
```diff
@@ -308,7 +308,11 @@ public function __construct(
...
         foreach ($this->config->getProjectDirectories() as $dir_name) {
-            $file_paths = $this->file_provider->getFilesInDir($dir_name, $file_extensions);
+            $file_paths = $this->file_provider->getFilesInDir(
+                $dir_name,
+                $file_extensions,
+                [$this->config, 'isInProjectDirs']
+            );
 
             foreach ($file_paths as $file_path) {
                 if ($this->config->isInProjectDirs($file_path)) {
```
weirdan, line 318: With the filter pushed down to the iterator, do we still need to check whether the returned file paths are in project / extra dirs?

dvz: Currently - yes (the iterator returns `true` for all leaves), but the filter can be made more universal: #7201

weirdan: All _first-level_ leaves, that is. But yeah, got you.

Issue #1141, opened 2018-12-19
Symlinked repos in composer must be ignored using full path


muglug:
Composer gives you an option to include packages using a [local path](https://getcomposer.org/doc/05-repositories.md#path), and creates a symlink by default.

Psalm uses `realpath` to get the actual path, so adding `vendor` to `<ignoreFiles />` doesn't mean ignoring these symlinked files.

Fix: block using vendor-provided path name

PR #7163, opened 2021-12-14
Add explicit `resolveSymlinks` option for project directories

dvz:
Partially resolves .#7057

A candidate approach that involves skipping the expensive recursive scanning of project directories (added via https://github.com/vimeo/psalm/commit/07b29e4a2f320ba93938a042eafd23bb13c61ce2) unless enabled for the individual directory (usually the vendor directory, according to #1141).

E.g.:
```xml
<ignoreFiles>
    <directory name=".git" />
    <directory name="other-big-subdirectory" />
    <directory name="vendor" resolveSymlinks="true" />
</ignoreFiles>
```


orklah:
So, for a user that actually uses this composer feature, this would break the analysis because Psalm would not be able to find vendors anymore, correct?

If that's so, we may want to default this to false first. If we can detect the symlink somehow, it would be great to emit a deprecation to warn about the config, and then we switch the config default in Psalm 5

dvz:
If only ignored files are affected (as claimed in the original issue), it's possible that this will result in symlinked paths being unintentionally scanned, but wouldn't reduce coverage nor result in errors.

Otherwise, adding a global option (and passing them to `FileFilter`) may be necessary.

orklah:
Merging your other PR created a conflict here. Mind rebasing?

orklah:
Thanks!

PR #7167, opened 2021-12-15
Fix conflicting array_map return provider fake variables.

### LINKED ISSUE #7164: UndefinedVariable "$__fake_offset_var__"
Thank you for making Psalm such an amazing analyzer.
I might have found a bug but I'm not sure.

https://psalm.dev/r/cccdb532d0
https://psalm.dev/r/c114e494ca

The error goes away when I use an anonymous function as the callable for the inner array_map.

AndrolGenhald:
Fix array_map return provider fake variables conflicting when nesting `array_map` calls (fixes #7164).
Ensure all fake variables are removed from context.

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayFilterReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayFilterReturnTypeProvider.php
@@ -29,6 +29,7 @@ use function array_map;
 use function array_slice;
 use function count;
 use function is_string;
+use function mt_rand;
 use function reset;
 use function spl_object_id;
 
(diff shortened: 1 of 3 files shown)


AndrolGenhald:
As noted in #7164, `$x['$__fake_offset_var__]` was being left in the context by `ArrayMapReturnTypeProvider`. I added a unique id to the fake variables to allow for easy cleanup, and it fixed the main issue as well, which I assume was due to the name being re-used and removed from the context on the inner call while it was still needed for the outer call, or vice versa.

weirdan:
Would you mind adding a test for this case?

AndrolGenhald:
> Would you mind adding a test for this case?

Completely slipped my mind, I have a `.git/info/exclude`d test file that I test things like this with locally and I forgot to copy the test over to a committed file.

orklah:
Can you maybe rename var_id into var_seed or something? var_id is already used a lot in Psalm to refer to actual variable identifiers(or fake ones), so in your case, the var_id would be `__fake_{$fake_var_id}_offset_var__` instead

weirdan:
Thanks!

PR #7168, opened 2021-12-15
Remove superflous exclude pattern in phpcs.xml

rarila:
Just noticed, that this excluded file has moved elsewhere and so this pattern is superflous.


weirdan:
Thanks

PR #7175, opened 2021-12-16
Fix faulty comparison with typed closure

muglug:
When an explicit `Closure(Foo):Bar` type is given this code results in a `MixedArgumentTypeCoercion`.

It should just be an `InvalidArgument`: https://psalm.dev/r/0880d5dfe8

Seen in the course of my ongoing deep audit (rewriting most of the logic in Rust, with changes for Hack).


weirdan:
Thanks!

PR #7179, opened 2021-12-16
Second draft of type comparison rewrite.

AndrolGenhald:
**Very rough draft**, supersedes #5610.
I'll add more description later, I'm just putting this up now to ask specific questions and make sure I'm not going off in the wrong direction.

### DIFF
--- a/tests/ArrayAssignmentTest.php
+++ b/tests/ArrayAssignmentTest.php
@@ -1881,7 +1881,7 @@ class ArrayAssignmentTest extends TestCase
                     takesArray($a);',
                 'error_message' => 'InvalidScalarArgument',
             ],
-            'nonEmptyAssignmentToListElementChangeType' => [
+            'SKIPPED-nonEmptyAssignmentToListElementChangeType' => [ // Should be LessSpecificReturnStatement?
                 '<?php
                     /**
                      * @param non-empty-list<string> $arr
--- a/tests/FunctionCallTest.php
+++ b/tests/FunctionCallTest.php
@@ -1931,7 +1931,12 @@ class FunctionCallTest extends TestCase
                       return $b["a"];
                     }
 
-                    a(["a" => "hello"]);',
+                    function b(string $str): string
+                    {
+                        return $str;
+                    }
+
+                    a(["a" => b("hello")]);',
                 'error_message' => 'InvalidScalarArgument',
             ],
             'objectLikeKeyChecksAgainstDifferentTKeyedArray' => [
--- a/tests/PropertyTypeTest.php
+++ b/tests/PropertyTypeTest.php
@@ -2631,7 +2631,7 @@ class PropertyTypeTest extends TestCase
                     class B extends A {}',
                 'error_message' => 'PropertyTypeCoercion',
             ],
-            'mixedTypeCoercion' => [
+            'SKIPPED-mixedTypeCoercion' => [
                 '<?php
                     class A {
                         /** @var array<int, A> */
@@ -2660,7 +2660,7 @@ class PropertyTypeTest extends TestCase
                     class B extends A {}',
                 'error_message' => 'PropertyTypeCoercion',
             ],
-            'staticMixedTypeCoercion' => [
+            'SKIPPED-staticMixedTypeCoercion' => [
                 '<?php
                     class A {
                         /** @var array<int, A> */
@@ -2790,7 +2790,7 @@ class PropertyTypeTest extends TestCase
                     echo $a->foo;',
                 'error_message' => 'NullPropertyFetch',
             ],
-            'badArrayProperty' => [
+            'SKIPPED-badArrayProperty' => [ // PropertyTypeCoercion?
                 '<?php
                     class A {}
 
@@ -3379,7 +3379,7 @@ class PropertyTypeTest extends TestCase
                     }',
                 'error_message' => 'UninitializedProperty',
             ],
-            'badStaticPropertyDefault' => [
+            'SKIPPED-badStaticPropertyDefault' => [
                 '<?php
                     class TestStatic {
                         /**
--- a/tests/ReturnTypeTest.php
+++ b/tests/ReturnTypeTest.php
@@ -1208,7 +1208,7 @@ class ReturnTypeTest extends TestCase
                       }
                       return $arr;
                     }',
-                'error_message' => 'LessSpecificReturnStatement',
+                'error_message' => 'InvalidReturnStatement',
             ],
             'invalidVoidStatementWhenMixedInferred' => [
                 '<?php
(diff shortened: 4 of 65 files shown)


tests/ArrayAssignmentTest.php
```diff
@@ -1881,7 +1881,7 @@ function takesArray(array $arr) : void {}
                     takesArray($a);',
                 'error_message' => 'InvalidScalarArgument',
             ],
-            'nonEmptyAssignmentToListElementChangeType' => [
+            'SKIPPED-nonEmptyAssignmentToListElementChangeType' => [ // Should be LessSpecificReturnStatement?
```
AndrolGenhald, line 1884: This now causes `LessSpecificReturnStatement - src/somefile.php:9:32 - The type 'non-empty-list<5|string>' is more general than the declared return type 'non-empty-list<string>' for takesList`, which seems correct to me. The type `non-empty-list<5|string>` contains the type `non-empty-list<string>`. Thoughts?

tests/FunctionCallTest.php
```diff
@@ -1931,7 +1931,12 @@ function a($b): int
                       return $b["a"];
                     }
 
-                    a(["a" => "hello"]);',
+                    function b(string $str): string
+                    {
+                        return $str;
+                    }
+
+                    a(["a" => b("hello")]);',
```
AndrolGenhald, line 1939: Non-int literal strings are no longer coercible to int, I think this is an improvement. Without this change it now reports `InvalidArgument`.

tests/PropertyTypeTest.php
```diff
@@ -2790,7 +2790,7 @@ class Foo {
                     echo $a->foo;',
                 'error_message' => 'NullPropertyFetch',
             ],
-            'badArrayProperty' => [
+            'SKIPPED-badArrayProperty' => [ // PropertyTypeCoercion?
```
AndrolGenhald, line 2793: Ignore this, it was failing previously and I thought `B` extended `A`, which does indeed report `PropertyTypeCoercion` already.

tests/ArrayAssignmentTest.php
orklah: I believe LessSpecificReturnStatement is emitted when you return a parent type of what's expected: https://psalm.dev/r/5be7f9f3be
If you return another type than expected, it's InvalidReturnStatement (even if some part of the expected is contained in what's returned.

Think about it without the array. What about this code in strict type: https://psalm.dev/r/3e727ead5c

tests/PropertyTypeTest.php
```diff
@@ -3379,7 +3379,7 @@ public function __construct() {
                     }',
                 'error_message' => 'UninitializedProperty',
             ],
-            'badStaticPropertyDefault' => [
+            'SKIPPED-badStaticPropertyDefault' => [
```
AndrolGenhald, line 3382: Appears to be working now, but at one point I think it was failing with some Coercion issue, which seems more accurate since `1` can be coerced to `bool`.

tests/FunctionCallTest.php
orklah: This would be a big BC break and I'm not sure I see the point. Maybe the term coercion is badly chosen in this case. The actual issue doesn't mention coercion: https://psalm.dev/docs/running_psalm/issues/InvalidScalarArgument/

muglug: `InvalidScalarArgument` originally came about because, in non-strict mode, PHP is more lenient with those scalar type conversions.

If you take the example given in the issue link above, and add a `strict_types` declare you get `InvalidArgument` ([example](https://psalm.dev/r/83cfce18a5)). But that's not universal — docblock types are [treated differently than runtime typehints](https://psalm.dev/r/001ac14f96). That lines up to `InvalidScalarArgument`'s original intent — to warn about likely runtime errors — but it can make some bugs seem less harmful even though they still lead to runtime errors (there's just a level of indirection).

So I think there might be a solid argument for reconsidering `InvalidScalarArgument` (as part of a major version change).

I'll draft a PR with that change for Psalm 5 to see what people think. Might even be worth a Twitter poll to take people's temperature.

tests/ArrayAssignmentTest.php
AndrolGenhald: My mental model for LessSpecific issues was that they should apply whenever the desired type is contained by the used type, but it looks like that's not correct? What's a better way to think about it? Would it be bad to change this in Psalm 5 so that it works that way? It seems like it intuitively makes sense to me.

tests/FunctionCallTest.php
AndrolGenhald: `"1"` is coercible to `int` and would report `InvalidScalarArgument` in this context, the way I did the type comparisons literals like `"hello"` are no longer coercible to `int`. I suppose either way `InvalidScalarArgument` still applies as it is described, but given that the string is definitely not an integer-string I would think it would make more sense to emit `InvalidArgument`.

tests/ReturnTypeTest.php
```diff
@@ -1208,7 +1208,7 @@ function foo(): array {
                       }
                       return $arr;
                     }',
-                'error_message' => 'LessSpecificReturnStatement',
+                'error_message' => 'InvalidReturnStatement',
```
AndrolGenhald, line 1211: This is another issue where a `LessSpecific` issue doesn't work the way I would intuitively expect. Perhaps for non-sealed arrays this would make sense, `array{a: string, b: string}`-unsealed could also have `"c"` set, but Psalm resolves `$arr` to `non-empty-array<"a"|"b", "foo">`, which definitely does not have `"c"` set.

So we have types A = unsealed `array{a: string, b: string, c: string}` and B = `non-empty-array<"a"|"b", "foo">`. A cannot contained by B because B doesn't have `"c"` set. B cannot contain A because A has `"c"` set.

I think perhaps if `LessSpecific` issues are changed to work the way I would inuit them to, maybe we should add a `IncompleteReturnStatement` or something of that nature to indicate that A (return type) and B (return statement type) intersect, but neither contains the other?

tests/ArrayAssignmentTest.php
orklah: I'd have expected that to be a more well defined rule in Psalm but it seems pretty hand coded

I'd say LessSpecific should be emitted when each type in the used type is contained by a type in the desired type without scalar coercion. However, the case where a scalar coercion is performed does not have a specific issue. (ScalarReturnTypeCoercion should exist maybe?)

I'm open to changes here, but it will have to be on a new major...

AndrolGenhald: With this type comparison rewrite I was thinking there would probably have to be _some_ breaking change anyway. To support `Intersection`s in a way that's not convoluted I think it would be good to replace most usages of `Union` with `TypeNode`, which is probably one of the most breaking changes we could do, and it's a heck of a lot of work. This is about the point I got to last time when I realized how much work I'd gotten myself into and rethought it.

PR #7182, opened 2021-12-18
Fixed ignoreTypeStats & useStrictTypes flag value detection for directories in config

SMAtaurRahman:
I've noticed that setting any of `ignoreTypeStats="true"` or `useStrictTypes="true"` flag for directories in config file is ignored due to a bug in `Psalm\Config\FileFilter`

`FileFilter::loadFromXMLElement()` reads these flags and turns them into boolean values, then pass the config array to `FileFilter::loadFromArray()`. loadFromArray reads the flag value again, but this time the function treats the values as string instead of boolean. As a result, it cannot properly detect the value.

I've tried to fix the behavior with this PR.

Note: I wasn't able to understand the purpose of `useStrictTypes=true`. Can you please let me know what difference does it make?

### DIFF
--- a/tests/Config/ConfigTest.php
+++ b/tests/Config/ConfigTest.php
@@ -1483,4 +1483,70 @@ class ConfigTest extends TestCase
         self::assertSame(get_class($analyzerMock), $config->getFiletypeAnalyzers()[$extension] ?? null);
         self::assertNull($expectedExceptionCode, 'Expected exception code was not thrown');
     }
+
+    public function testTypeStatsForFileReporting(): void
+    {
+        $this->project_analyzer = $this->getProjectAnalyzerWithConfig(
+            Config::loadFromXML(
+                (string) getcwd(),
+                '<?xml version="1.0"?>
+                <psalm>
+                    <projectFiles>
+                        <directory ignoreTypeStats="true" name="src/Psalm/Config" />
+                        <directory ignoreTypeStats="1" name="src/Psalm/Internal" />
+                        <directory ignoreTypeStats="true1" name="src/Psalm/Issue" />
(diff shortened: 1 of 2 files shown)


orklah:
useStrictTypes is used to emulate the behavior of `declare(strict_types=1);`. In practice, it makes Psalm a little more strict when it sees scalar coercion (returning a string instead of int for example) or __toString coercion (returning an object with __toString when a string is expected)

tests/Config/ConfigTest.php
```diff
@@ -1483,4 +1483,70 @@ public function pluginRegistersScannerAndAnalyzer(int $flags, ?int $expectedExce
         self::assertSame(get_class($analyzerMock), $config->getFiletypeAnalyzers()[$extension] ?? null);
         self::assertNull($expectedExceptionCode, 'Expected exception code was not thrown');
     }
+
+    public function testTypeStatsForFileReporting(): void
```
orklah, line 1487: tests are failing on windows. I'm pretty sure this has to do with path separators in getcwd()

SMAtaurRahman: Sorry, I've no idea how to approach this issue. I don't have access to windows machine.
`testBarebonesConfig()` also uses `getcwd()`. I wonder how it is working.

Should I try using `dirname(__DIR__, 2)` instead?

orklah: oh, that's not getcwd(), the issue is with `. '/'` at the end of paths. You can replace with ` . DIRECTORY_SEPARATOR`

SMAtaurRahman: Cool, it worked. Thanks.
I should be careful with dir separator from next time.

orklah:
Thanks!

PR #7188, opened 2021-12-19
Use InvalidScalarArgument only when we can be sure PHP attempts coercion

muglug:
This is an alternative to https://github.com/vimeo/psalm/pull/7181

It narrows the occasions where `InvalidScalarArgument` is emitted to only the places where we're sure that PHP will attempt to coerce one scalar type to another. This matches the original intent of `InvalidScalarArgument` — a lesser issue than `InvalidArgument`, for when the PHP runtime may not care.

### DIFF
--- a/docs/running_psalm/issues/InvalidScalarArgument.md
+++ b/docs/running_psalm/issues/InvalidScalarArgument.md
@@ -1,6 +1,10 @@
 # InvalidScalarArgument
 
-Emitted when a scalar value is passed to a method that expected another scalar type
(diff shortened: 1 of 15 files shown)


orklah:
This would make sense. It would be great to document this somewhere. Maybe a comment on the scalar_type_match_found property and an explanation in the Issue page?

orklah:
@weirdan should we merge this now or is it something we will add to Psalm 5?

weirdan:
> is it something we will add to Psalm 5?

That would be my preference, yes.

PR #7192, opened 2021-12-20
allow marking enum cases as deprecated

pilif:
This PR will allow an `@deprecated` annotation to be placed on enum cases which will then cause a `DeprecatedConstant` issue to be raised when the case is being referenced.

I'm unsure as to whether this is the correct way to do things as I'm not too familiar with the psalm internals, so I'm very happy about pointers in how to do this correctly.

~~Also, ideally, I would like to have supported the `#[Psalm\Deprecated]` attribute too, but I don't think the php parser is exposing attributes on enum cases - at least I could not get to them, but maybe I looked in the wrong spot.~~

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
@@ -46,6 +46,7 @@ use Psalm\Issue\InvalidTypeImport;
 use Psalm\Issue\MissingDocblockType;
 use Psalm\Issue\ParseError;
 use Psalm\IssueBuffer;
+use Psalm\Storage\AttributeStorage;
 use Psalm\Storage\ClassConstantStorage;
 use Psalm\Storage\ClassLikeStorage;
 use Psalm\Storage\EnumCaseStorage;
@@ -1350,6 +1351,24 @@ class ClassLikeNodeScanner
                 $case_location
             );
 
+            $attrs = $this->getAttributeStorageFromStatement(
+                $this->codebase,
+                $this->file_scanner,
+                $this->file_storage,
+                $this->aliases,
+                $stmt,
+                $this->storage->name ?? null
+            );
+
+            foreach ($attrs as $attribute) {
+                if ($attribute->fq_class_name === 'Psalm\\Deprecated'
+                    || $attribute->fq_class_name === 'JetBrains\\PhpStorm\\Deprecated'
+                ) {
+                    $case->deprecated = true;
+                    break;
+                }
+            }
+
             $comment = $stmt->getDocComment();
             if ($comment) {
                 $comments = DocComment::parsePreservingLength($comment);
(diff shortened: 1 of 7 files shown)


weirdan:
Attributes are available on enum cases:

```
    public $stmts =>
    array(1) {
      [0] =>
      class PhpParser\Node\Stmt\EnumCase#1183 (4) {
        public $name =>
        class PhpParser\Node\Identifier#1182 (2) {
          public $name =>
          string(1) "A"
          protected $attributes =>
          array(4) {
            'startLine' =>
            int(5)
            'startFilePos' =>
            int(50)
            'endLine' =>
            int(5)
            'endFilePos' =>
            int(50)
          }
        }
        public $expr =>
        NULL
        public $attrGroups =>
        array(1) {
          [0] =>
          class PhpParser\Node\AttributeGroup#1181 (2) {
            public $attrs =>
            array(1) {
              [0] =>
              class PhpParser\Node\Attribute#1180 (3) {
                public $name =>
                class PhpParser\Node\Name\FullyQualified#1179 (2) {
                  public $parts =>
                  array(2) {
                    [0] =>
                    string(5) "Psalm"
                    [1] =>
                    string(10) "Deprecated"
                  }
```

pilif:
thanks. I must have been looking at the wrong spot then (the `attrGroups` I checked was empty).

I'll check later (tomorrow Europe time - I'm on family duty now) and update the PR

pilif:
yes. you were totally right. the attributes are available.

I have pushed a second commit that adds support for the `Psalm\Deprecated` and `JetBrains\PhpStorm\Deprecated` attributes for enum cases.

The patch as shown on GitHub looks a bit crazy because I have extracted the attribute resolution into a new method as it was exactly identical for the property storage too

pilif:
I pushed another commit that updates the docs. However, that fails a test because the documentation samples are checked (very cool) with php 8.0 (not so good).

How should I proceed?

- update the GitHub Action to parse the tests with php-8.1?
- remove the updated code sample?

orklah:
Let's upgrade that to 8.1, it should not be an issue

weirdan:
Wouldn't adding the issue to the list below suffice?

https://github.com/vimeo/psalm/blob/646ba983a0fb09a4ed4946ac2fc47ac192a111d1/tests/DocumentationTest.php#L312-L318

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
```diff
@@ -1345,10 +1346,41 @@ private function visitEnumDeclaration(
...
+                $this->aliases,
+                $stmt,
+                $this->storage->name ?? null
+            );
+
+            foreach ($attrs as $attribute) {
+                if ($attribute->fq_class_name === 'Psalm\\Deprecated'
+                    || $attribute->fq_class_name === 'JetBrains\\PhpStorm\\Deprecated'
+                ) {
+                    $case->deprecated = true;
```
weirdan, line 1367:
```suggestion
                    $case->deprecated = true;
                    break;
```

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
```diff
@@ -1345,10 +1346,41 @@ private function visitEnumDeclaration(
...
+                }
+            }
+
+            $comment = $stmt->getDocComment();
+            if ($comment) {
+                $var_comments = CommentAnalyzer::getTypeFromComment(
```
weirdan, line 1373: This doesn't sound right. Did you mean to use `DocblockParser::parsePreservingLength()`, like it's used for constants here?: https://github.com/vimeo/psalm/blob/646ba983a0fb09a4ed4946ac2fc47ac192a111d1/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php#L1207-L1211

weirdan (changes requested):

pilif:
> Wouldn't adding the issue to the list below suffice? 

yes. you're right. I noticed only after asking about the general PHP update, so that what I've done. In my rebased commits, I have removed the PHP version change and followed your suggestion.

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
pilif: ah! Much better. I wasn't aware of that one.

Fixed.

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
pilif: fixed. thank you

weirdan (approved):

weirdan:
Thanks!

PR #7194, opened 2021-12-20
Added pure annotations to enum functions

### LINKED ISSUE #7086: Mark enum functions as pure
The enum functions introduced with #6429 should be considered pure, as they must never change the state of the enum.

ricardoboss:
Fixes #7086


orklah:
Thanks!

PR #7195, opened 2021-12-20
fix wrong handling of flags in context

orklah:
This PR started when I noticed a weird handling of inside_conditional flag in switch. After looking for more I found a few.

After that, I noticed that we don't always unflag those before leaving an analyzer, mainly when the other analyzer we called returned false.

Not sure this will have any effect whatsoever, but it may fix a few edge cases here and there. It will be less confusing in any case


orklah:
CI is failing for unrelated reason. I filed an issue here: https://github.com/thecodingmachine/docker-images-php/issues/285

muglug (approved):

orklah:
Thanks for the review @muglug  ;)

weirdan:
I wonder if we could simplify this pattern with something like this:
```php
$context->withAlteredFlags(function (Context $context) use (...) {
   $context->inside_conditional = false;
   SomeAnalizer::analyze(..., $context);
});
```
that would undo flag changes performed by the callback.

PR #7197, opened 2021-12-21
fix(stubs): UnitEnum::cases() can return an empty list

azjezz:
e.g:
```
enum Foo: string {}

$cases = Foo::cases();

```

### DIFF
--- a/stubs/Php81.phpstub
+++ b/stubs/Php81.phpstub
@@ -6,7 +6,7 @@ namespace {
 
         /**
          * @psalm-pure
-         * @return non-empty-list<static>
+         * @return list<static>
          */
         public static function cases(): array;
     }


orklah:
Is there a useful use case where we might want to do that?

It seems too bad to remove a more detailed type if it's just for principle and noone will ever use that

azjezz:
I have multiple enums in couple of projects that don't have any cases, manually used as utility classes ( final abstract ), e.g:

```
enum Helper {
  public static function foo(): void {}
}
```

unlike an abstract class, this "class" is non-instantiable *and* final.

azjezz:
so IMHO, the return type is misleading, as this method can return an empty list.

orklah:
I just checked, Psalm seem to already do the complete inference and return literal values. Was there a piece of code that was causing you an issue here?

I'm fine changing the phpdoc of the stub, as long as people that actually uses Enum as they were intended don't have degraded inference

azjezz:
https://psalm.dev/r/0a76e554d7

orklah:
Ok, I guess it's fair when the Enum is not known

weirdan:
@azjezz non-empty-list was actually coming from the `get_enum_cases()` docblock in your snippet.

azjezz:
> @azjezz non-empty-list was actually coming from the get_enum_cases() docblock in your snippet.

yes, which psalm previously failed to report as an error, but now does :)

PR #7201, opened 2021-12-22
Add early file filtering to `FileProvider::getFilesInDir()`

dvz:
Partially resolves .#7057

Extends the use of `callable $directory_filter` added via #7161 to additionally cover individual files, containing the _project/extra/ignored_ check logic within `FileProvider::getFilesInDir()`.


weirdan (approved):
Looks good to me, but we need to fix the CI issue before we can proceed with this.

orklah:
Thanks!

PR #7202, opened 2021-12-22
Remove unused in-array assertions

muglug:
These assertions cannot be parsed, since the `in-array` assertion handler always tries to convert assertions to a type.

This should have a really small performance improvement for array_key_exists checks, but it also just simplifies the code in a pleasing fashion and removes a couple of now-unnecessary `try...catch` blocks.


orklah:
Cool! Thanks!

PR #7203, opened 2021-12-22
flag the context as "inside_conditional" when analyzing leftover cases to prevent emitting unused error

### LINKED ISSUE #6389: wierd behaviour with `switch` statement
With the following snippet of code https://psalm.dev/r/88960656f8 psalm produces some unexpected errors `UnusedFunctionCall` and `ParadoxicalCondition`

orklah:
This should fix #6389

This was due to the analyzing the function call when no [insideUse](https://github.com/vimeo/psalm/blob/97e6511fab6622c1ca434c665dd895e77a0eeca2/src/Psalm/Context.php#L836) was flagged. Psalm emits an Unused error when it sees a pure function call that is not in a "use" context.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Block/SwitchCaseAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/SwitchCaseAnalyzer.php
@@ -320,11 +320,17 @@ class SwitchCaseAnalyzer
             );
 
             if ($new_case_equality_expr) {
+                $was_inside_conditional = $case_context->inside_conditional;
+
+                $case_context->inside_conditional = true;
+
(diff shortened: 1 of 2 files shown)


orklah:
Tests are clearly too long. The change must imply a lot of other checks.

orklah:
Hey @muglug, I have no idea how to proceed from there. I know switch is hard to analyze and it does a lot of background conditions and everything. I have no idea why the tests would be this long after this change and I'm not sure how I can fix that an other way. Would you have a piece of advice here?

orklah:
Oh, the test slowness may not be related to this PR at all.

muglug:
Sometimes CI be like that?

PR #7205, opened 2021-12-23
allow phpcodesniffer-composer-installer plugin

orklah:
Looks like CI has switched on composer 2.2 and it now warns that a plugin is not allowed.

I'm not sure what the source of the plugin is but I guess there is no reason not to allow it, given it was executed before


muglug (approved):

orklah:
Thanks Matt!

I'm merging that, even with the CI hanging as it might be somewhat related given it's due to composer 2.2

PR #7206, opened 2021-12-23
Composer 2.1

orklah:
*This is a draft not meant to be merged for now*

This reverts the CI to run on Composer 2.1. For a mysterious reason, Composer 2.2 makes CI hangs.

https://blog.packagist.com/composer-2-2/ warned that:

> Please beware, the prompt may hang CI/build pipelines if they aren't using the --no-interaction (-n) option, as they should. (PR #10314)

But it should happen at the install stage, not after phpunit finished, and I don't see why it happens only on test 4 and 7

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: '7.1'
-          tools: composer:v2
+          tools: composer:v2.1
           coverage: none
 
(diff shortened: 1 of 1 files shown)


orklah:
Note: the issue should be fixed by https://github.com/composer/composer/issues/10387

weirdan:
Or sebastianbergmann/phpunit#4846, whichever comes first.

orklah:
This is no longer useful

PR #7208, opened 2021-12-23
Workaround for PHPUnit process isolation issue with composer 2.2

weirdan:
This skips composer-generated proxy binaries, thus avoiding the issue introduced by composer 2.2

### DIFF
--- a/bin/tests-github-actions.sh
+++ b/bin/tests-github-actions.sh
@@ -13,7 +13,7 @@ function run {
 
   local -r phpunit_cmd='
 echo "::group::{}";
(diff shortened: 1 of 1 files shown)


orklah:
Yeah, we might as well avoid that if it can cause issues

GrahamCampbell:
This was fixed in Composer 2.2.1.

weirdan:
@GrahamCampbell no, it was fixed after 2.2.1 and is not released yet: https://github.com/composer/composer/compare/2.2.1...main

PR #7210, opened 2021-12-24
Improve CliUtils $argv handling

vstm:
This pull request aims to simplify the CliUtils handling by skipping the first value of the `$argv` array when trying to get the arguments and the paths to be checked.

I have also added a CliUtils test which covers some basic (and one not so basic) use cases of the two methods I changed. Some parts I couldn't test yet (for example when passing `-` as argument, I would have to "mock" STDIN).

This also solves an issue with composer 2.2 when the "bin-dir" config is set, because in with the proxies the value in `$argv[0]` is not covered anymore in the code where psalm "recognices" itself (the part of the code which I removed completely). See this gist on how to reproduce the issue: https://gist.github.com/vstm/40258de6df2991cf9309cd10e716089b

### DIFF
--- a/src/Psalm/Internal/CliUtils.php
+++ b/src/Psalm/Internal/CliUtils.php
@@ -248,68 +248,66 @@ final class CliUtils
             }
         }
 
-        if ($input_paths) {
-            $filtered_input_paths = [];
+        $filtered_input_paths = [];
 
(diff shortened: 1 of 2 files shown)


weirdan (approved):

weirdan:
Thanks!

PR #7211, opened 2021-12-24
Allow XdebugHandler version 3

johnstevenson:
composer/xdebug-handler version 3 provides the same functionality as version 2 but removes support for older PHP versions (< 7.2.25). It will be shipped with the impending Composer 2.3 release.

xdebug-handler version 2 will be supported for the lifetime of  [Composer 2.2 LTS](https://blog.packagist.com/composer-2-2/).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "amphp/byte-stream": "^1.5",
         "composer/package-versions-deprecated": "^1.8.0",
         "composer/semver": "^1.4 || ^2.0 || ^3.0",
-        "composer/xdebug-handler": "^1.1 || ^2.0",
+        "composer/xdebug-handler": "^2.0 || ^3.0",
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Why is support for v1 dropped? That seems unnecessarily restrictive.

orklah:
> Why is support for v1 dropped? That seems unnecessarily restrictive.

Yeah, probably some kind of BC break too I believe

weirdan:
Good point about v1. As of now, there's no reason to drop it. @johnstevenson, mind restoring that part?

johnstevenson:
Will do. Sorry about that

orklah:
Don't be sorry :) We're not here to always be right :) mistakes are expected, thanks for your contribution, it's very welcome!

weirdan:
Thanks!

PR #7215, opened 2021-12-25
add missing samesite array attribute

swiffer:
https://www.php.net/manual/en/function.session-set-cookie-params

When using the second signature, an associative array which may have any of the keys lifetime, path, domain, secure, httponly and samesite. The values have the same meaning as described for the parameters with the same name. The value of the samesite element should be either Lax or Strict. If any of the allowed options are not given, their default values are the same as the default values of the explicit parameters. If the samesite element is omitted, no SameSite cookie attribute is set.


weirdan:
You need to update `Callmap.php` as well, see https://psalm.dev/docs/contributing/editing_callmaps/

swiffer:
@weirdan done

weirdan:
Thanks!

Issue #7219, opened 2021-12-26
Uncaught Exception: Error Call to undefined function str_contains()


voku:
#### example:
Whenever I change the "errorLevel" and try to re-run psalm I see this error via e.g. php 7.4. It's working with php >= 8.0 and if I once ran it with >= 8.0, then I can also run it with e.g. php 7.4 again.

#### error:
`
Uncaught Exception: Error Call to undefined function str_contains()
Emitted in phar:///psalm.phar/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayMapReturnTypeProvider.php:274
Stack trace in the forked worker:
#0 phar:///psalm.phar/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayMapReturnTypeProvider.php(264): Psalm\Internal\Provider\ReturnTypeProvider\ArrayMapReturnTypeProvider::cleanContext()
#1 phar:///psalm.phar/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayMapReturnTypeProvider.php(114): Psalm\Internal\Provider\ReturnTypeProvider\ArrayMapReturnTypeProvider::getReturnTypeFromMappingIds()
#2 phar:///psalm.phar/src/Psalm/Internal/Provider/FunctionReturnTypeProvider.php(164): Psalm\Internal\Provider\ReturnTypeProvider\ArrayMapReturnTypeProvider::getFunctionReturnType()
#3 phar:///psalm.phar/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php(76): Psalm\Internal\Provider\FunctionReturnTypeProvider->getReturnType()
#4 phar:///psalm.phar/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php(139): Psalm\Internal\Analyzer\Statements\Expression\Call\FunctionCallReturnTypeFetcher::fetch()
#5 phar:///psalm.phar/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php(175): Psalm\Internal\Analyzer\Statements\Expression\Call\FunctionCallAnalyzer::analyze()
#6 phar:///psalm.phar/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php(68): Psalm\Internal\Analyzer\Statements\ExpressionAnalyzer::handleExpression()
[...]

PR #7220, opened 2021-12-26
Use `strpos` instead of `str_contains`

weirdan:
Fixes vimeo/psalm#7219 (str_contains was added in PHP 8)

### DIFF
--- a/bin/test-with-real-projects.sh
+++ b/bin/test-with-real-projects.sh
@@ -22,8 +22,8 @@ phpunit)
 	;;
 
 collections)
-	git clone --depth=1 git@github.com:muglug/collections.git
-	cd collections
+	git clone --depth=1 git@github.com:psalm/endtoend-test-collections.git
+	cd endtoend-test-collections
 	composer install
 	"$PSALM" --monochrome --show-info=false
 	;;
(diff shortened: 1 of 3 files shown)


bin/test-with-real-projects.sh
```diff
@@ -22,8 +22,8 @@ phpunit)
 	;;
 
 collections)
-	git clone --depth=1 git@github.com:muglug/collections.git
-	cd collections
+	git clone --depth=1 git@github.com:psalm/endtoend-test-collections.git
```
orklah, line 25: those two could probably removed. AFAIK, they're not in CI anymore...

src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayMapReturnTypeProvider.php
```diff
@@ -505,7 +504,7 @@ public static function getReturnTypeFromMappingIds(
     public static function cleanContext(Context $context, int $fake_var_discriminator): void
     {
         foreach ($context->vars_in_scope as $var_in_scope => $_) {
-            if (str_contains($var_in_scope, "__fake_{$fake_var_discriminator}_")) {
```
orklah, line 508: Psalm did not pick that up? Polyfills are messing with the analysis maybe?

I wonder how hard it could be to add a new config forbidPolyfills that, when enabled would warn about functions removed through deltas but defined elsewhere...

weirdan: > Polyfills are messing with the analysis maybe?

Reflection, likely.

bin/test-with-real-projects.sh
weirdan: `collections` is used:

https://github.com/vimeo/psalm/blob/aa7e400908833b10c0333861f86cd48c510b60eb/.circleci/config.yml#L68-L76

Issue #7223, opened 2021-12-26
Regression: Docblock-defined type 0|int<1, 16> for $length is never int(1)


kamil-tekiela:
In the following example Docblock-defined type 0|int<1, 16> for $length is never int(1) the length can be between 0 and 15 which makes 1 a valid possibility. Psalm started to complain about it in the recent version. 

[Docblock-defined type 0|int<1, 16> for $length is never int(1)](https://psalm.dev/r/b9df4ac358)

orklah:
I fixed only one side of the issue in #7224. The other side is technically a duplicate of https://github.com/vimeo/psalm/issues/6871 but this is very not obvious so I'll keep that one open too.

As I said there: the other side, while easily fixed causes other false positives that I just don't know how to fix yet.

orklah:
This has been fixed in master branch and will be released in Psalm 5

PR #7224, opened 2021-12-26
handle literal equality with integer ranges

orklah:
This will fix a case with literal equality when integer ranges are involved, fixing #7223


weirdan:
Original snippet still fails though

weirdan:
This fixed `DocblockTypeContradiction`, but not the `RedundantCondition` in the `else` branch of the ternary.

orklah:
Oh, good catch. I'll check that

Issue #4535, opened 2020-11-12
Why PossiblyUndefinedStringArrayOffset here


klimick:
https://psalm.dev/r/8434aa7bca

There is way to win `PossiblyUndefinedStringArrayOffset` issue without explicit type annotation like this?

```php
print_r(
    // Is possibly undefined? Why?
    group(getUsers(),
    /**
     * @param array{id: string, group_id: string} $_
     */
    fn(array $_): string => $_['group_id'])
);
```

muglug:
No.

muglug:
Psalm doesn't know that the arguments passed to the closure first result from the `getUsers` call parameters. It can, though, detect a bug when the types disagree: https://psalm.dev/r/2a241d684b

klimick:
@muglug thanks for answer. But.

Why it works fine? https://psalm.dev/r/279fe26579

Functions can't have the same behavior due to design?

muglug:
@klimick it's about when Psalm can infer types – in the latter example it's already inferred that it's working on an array with `group_id`, vs the former where that inference happens within the same set of args.

PR #7228, opened 2021-12-27
Contextual inference for closure param types

### LINKED ISSUE #5820: closure parameter are analysed without the function context ( generics ) when used with userland functions.
The following example shows the issue: https://psalm.dev/r/3e6b2451c3

both map and array_map act the same way, however, psalm complains about missing `@param` type declaration for the closure with user land implementations, but not builtin array_map. 

This results in useless type declaration being added ( see https://psalm.dev/r/15e391e947 ) to satisfy psalm.

### LINKED ISSUE #7209: Contextual inference for Closure param types
It's very painful that psalm can't infer closure parameters from context:
https://psalm.dev/r/310e6ed76e

Of course we can specify type hint. But if type have templates it will be lost:
https://psalm.dev/r/b60a08d22a

Doc blocks preserves templates, but code looks bad:
https://psalm.dev/r/7c9621b9d3

We could write code like this:
```php
/** @var list<Foo<int>> */
$fooList = [];
$mapped = map($fooList, fn(Foo $foo) => [$foo]);

// $mapped is list<Foo<array{int}>>. No doc blocks!
```

From what I've learned, it is relatively easy to implement left-to-right args type resolving for furter closure params inference.
I don't think Psalm needs a right-to-left analysis that was be mentioned in #5820. From my point of view its rare and difficult case.

klimick:
Closes #7209

Also solves my problem from #4535
Partially addresses #5820

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
@@ -231,6 +231,23 @@ class ArgumentsAnalyzer
                 );
             }
 
+            $inferred_arg_type = $statements_analyzer->node_data->getType($arg->value);
+
+            if (null !== $inferred_arg_type && null !== $template_result && null !== $param && null !== $param->type) {
+                $codebase = $statements_analyzer->getCodebase();
+
+                TemplateStandinTypeReplacer::replace(
+                    clone $param->type,
+                    $template_result,
+                    $codebase,
+                    $statements_analyzer,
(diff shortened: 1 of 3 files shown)


klimick:
Appearing `UnnecessaryVarAnnotation` in the psl is reasonably. What should I do in this case?

weirdan:
I'll baseline that error once this PR is merged.

weirdan (approved):
LGTM.

orklah:
Thanks!

orklah (approved):

weirdan:
Thanks!

PR #7230, opened 2021-12-27
code grooming

orklah:
a few improvements each with its own commit. Please tell me if some are not desired, I'll drop the commit


src/Psalm/Internal/Cli/Psalm.php
```diff
@@ -212,7 +212,7 @@ public static function run(array $argv): void
         $include_collector = new IncludeCollector();
         $first_autoloader = $include_collector->runAndCollect(
             // phpcs:ignore SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName
-            function () use ($current_dir, $options, $vendor_dir): ?ClassLoader {
```
weirdan: There were several return types like this in other entrypoint classes. Also, you may want to add a comment explaining why they should be left as they are.

orklah: Thanks, fixed and documented

PR #7233, opened 2021-12-27
more code grooming

orklah:
It's pretty minor refactoring, except for what I removed in ParseTreeCreator. What I removed felt important but I couldn't get how it could be useful and tests seems to prove me right...

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/FilterVarReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/FilterVarReturnTypeProvider.php
@@ -55,8 +55,6 @@ class FilterVarReturnTypeProvider implements FunctionReturnTypeProviderInterface
         ) {
             $filter_type_type = $second_arg_type->getSingleIntLiteral();
 
-            $filter_type = null;
-
             switch ($filter_type_type->value) {
                 case FILTER_VALIDATE_INT:
                     $filter_type = Type::getInt();
--- a/src/Psalm/Internal/Scanner/DocblockParser.php
+++ b/src/Psalm/Internal/Scanner/DocblockParser.php
@@ -101,8 +101,10 @@ class DocblockParser
                 unset($lines[$k]);
             } else {
                 // Strip the leading *, if present.
-                $lines[$k] = str_replace("\t", ' ', $line[$k]);
-                $lines[$k] = preg_replace('/^ *\*/', '', $line[$k]);
+                $text = $lines[$k];
+                $text = str_replace("\t", ' ', $text);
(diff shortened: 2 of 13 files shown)


src/Psalm/Internal/Provider/ReturnTypeProvider/FilterVarReturnTypeProvider.php
```diff
@@ -55,8 +55,6 @@ public static function getFunctionReturnType(FunctionReturnTypeProviderEvent $ev
         ) {
             $filter_type_type = $second_arg_type->getSingleIntLiteral();
 
-            $filter_type = null;
```
weirdan, line 58: I would actually either leave it here or add a default case to the switch below.

src/Psalm/Internal/Scanner/DocblockParser.php
```diff
@@ -101,8 +101,10 @@ public static function parse(string $docblock, int $offsetStart): ParsedDocblock
                 unset($lines[$k]);
             } else {
                 // Strip the leading *, if present.
-                $lines[$k] = str_replace("\t", ' ', $line);
-                $lines[$k] = preg_replace('/^ *\*/', '', $line);
```
weirdan, line 105: It looks like a bug actually, the tab expansion was futile.

weirdan (commented):
> What I removed felt important but I couldn't get how it could be useful

We can leave it for another day then. I imagine it would be quite a pain to put it back half a year later should it actually break something.

src/Psalm/Internal/Provider/ReturnTypeProvider/FilterVarReturnTypeProvider.php
orklah: Did you see this line a little above? https://github.com/vimeo/psalm/pull/7233/files/0b3e1299decdd75fe6cf38acd88391cfea323516#diff-76c8a38ce19c782b153d6a5cbededfe163697de0e37f2f40559f3f07f8eaf318L50

It was just initialized twice

src/Psalm/Internal/Scanner/DocblockParser.php
orklah: Yeah, this was definitely a bug. Forgot to mention it in the PR. I fixed it the way I think it was intended, but I'm not sure what it was designed to do in what case though...

src/Psalm/Internal/Provider/ReturnTypeProvider/FilterVarReturnTypeProvider.php
weirdan: Ok then.

PR #7238, opened 2021-12-29
use more precise type for `PDOStatement->fetchObject`'s `$ctorArgs`

staabm:


stubs/pdo.phpstub
```diff
@@ -12,7 +12,7 @@ class PDOStatement implements Traversable
      *
      * @template T
      * @param class-string<T> $class
-     * @param array $ctorArgs
+     * @param list<mixed> $ctorArgs
```
weirdan, line 15: Just curious whether it supports named arguments in form of `['arg3' => 123, 'param2' => 'zzz']`

staabm: really good point. just tested locally, and thats confirmed named args are supported

PR #7239, opened 2021-12-29
Move from deprecated Composer 1 to Composer 2 runtime api

DeyV:


orklah:
phar build is failing in CI.

It must have something to do with `scoper.inc.php` file, but I was unable to see why PackageVersion worked but not the new package. Can you try to tweak with that to see if you can fix it?

DeyV:
I fixed this problem. 

To full remove dependency of composer/package-versions-deprecated 
release of https://github.com/psalm/psalm-plugin-phpunit/pull/116 is needed

src/Psalm/Internal/CliUtils.php
```diff
@@ -147,8 +147,13 @@ public static function requireAutoloaders(
             exit(1);
         }
 
-        define('PSALM_VERSION', Versions::getVersion('vimeo/psalm'));
-        define('PHP_PARSER_VERSION', Versions::getVersion('nikic/php-parser'));
+        try {
```
orklah, line 150: Please add $version = null; just above the try. Psalm does not like working with possibly undefined variables

orklah:
Thanks!

weirdan:
I think this broke psalm.dev:

```
Psalm runner output: 

Internal Psalm error on line 22 - /src/OnlineChecker.php: Class "PackageVersions\Versions" not found
```

weirdan:
Fixed in https://github.com/psalm/psalm.dev/commit/8c65f63cdeb9c08102067ea398061686237704e7

PR #7240, opened 2021-12-29
mysqli_fetch_object: added missing generic

staabm:


stubs/mysqli.phpstub
```diff
@@ -23,6 +23,9 @@ class mysqli_result implements Traversable
  * @psalm-taint-sink callable $class
  *
  * @template T of object
+ * @template TValue
```
staabm, line 26:
```suggestion
 * @template TValue of object
```

orklah:
I'd expect this template to follow the change you did in the procedural function

https://github.com/vimeo/psalm/pull/7240/files/5361cda062ef45fc7408aa92c15f15124379ade9..457dd2f1c92bc771f62eed2d7d1bde2e02d4f900#diff-6ec175bad64ce89160634d6c7ee7658fbc1944223e6ab59813aeb4ad17a1d032R4

staabm:
Sorry, I don't get what you are telling me.

orklah:
There's two way to get objects from a mysqli_result: https://www.php.net/manual/en/mysqli-result.fetch-object.php

There's an OOP style and Procedural style.

Both are in the stub file you changed, however, you added `of object` only to the procedural function `mysqli_fetch_object` and not in the OOP method `mysqli_result::fetch_object`

I'd expect both to behave similarly

staabm:
IMO both apis - with the PR as is - define the same templates. you need to watch for the `TValue` vs. `T`.

I think both signatures match atm

orklah:
oh, nvm, I was talking about that: https://github.com/vimeo/psalm/pull/7240/files#r776390573 but it isn't in the final result

orklah:
Thanks!

PR #7242, opened 2021-12-30
PHP 8.2: seal all properties configuration

danog:


orklah:
I quite agree with the sentiment, but I'm not sure how PHP 8.2 changed anything here, so I don't see why the default should change based on this version

Basically, PHP 8.2 throw a deprecated when using an undeclared property *without __get or __set*.

Here, you suggest emitting an issue in Psalm even when __get or __set are present based only on @property presence

I'd be up for adding this config with a false default for Psalm 4 and switching it to true for Psalm 5. The default based on PHP version seems weird to me.

(Note: I'm confused by the fact that there is an error on the assignment but not on the fetch here: https://psalm.dev/r/ce07ecb689 given that my config has the `Use PHPDoc methods and properties without magic call.` checked, I'd not expect issue on the assignment either)

VincentLanglet:
> I quite agree with the sentiment, but I'm not sure how PHP 8.2 changed anything here, so I don't see why the default should change based on this version
> 
> Basically, PHP 8.2 throw a deprecated when using an undeclared property _without __get or __set_.
> 
> Here, you suggest emitting an issue in Psalm even when __get or __set are present based only on @Property presence
> 
> I'd be up for adding this config with a false default for Psalm 4 and switching it to true for Psalm 5. The default based on PHP version seems weird to me.

I think the idea was that people shouldn't do deprecated stuff. I don't think that's a bad idea but as a user I would also consider weird to have such impact when bumping my php version. The psalm4/psalm5 BC-break seems more natural.

Also, I might not understand all the changes but
```
public function shouldSealAllProperties(ClassLikeStorage $storage): bool
    {
        return $storage->sealed_properties
            || $this->config->seal_all_properties
            || ($this->php_major_version >= 8 && $this->php_minor_version >= 2);
    }
```
is returning always true for version 8.2 ; which means I cannot override the value with my config.

And speaking about PHP 8.2, this should supports the `#[AllowDynamicProperties]` attributes.

danog:
You're both right, I may have confused this task with a another, different task regarding PHP 8.2 migration, sorry about this.
All fixed now!

PR #7245, opened 2021-12-30
Speed up Windows tests

weirdan:
Enable multiple workers for Windows, just like we do on Linux already.

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -3,13 +3,41 @@ name: Run unit tests on Windows
 on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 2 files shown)


orklah:
That's cool!

bendavies:
very nice. I wondered why I didn't do windows when i originally did #4985.
turns out the tests weren't ran on windows at the time.

orklah:
> turns out the tests weren't ran on windows at the time.

They were, but they ran in AppVeyor until https://github.com/vimeo/psalm/pull/6173. I migrated them on GA but I didn't look up how to translate the splitting script on windows

bendavies:
Ah right you are

PR #7247, opened 2021-12-30
Fixes #7246, wrap getTypeContextAtPosition in try/catch

### LINKED ISSUE #7246: Language Server Crash with Expecting non-empty $root_file_path and $checked_file_path
See #7246 and psalm/psalm-vscode-plugin#107

Specifically if something crashes deep in Psalm we should try to catch most of these crashes in the language server and continue (if possible). This is one such instance where continuing is fine as the error comes from the result of live typing so the code completion (that is erroring) is wrong as the time of typing the letters.

This is really hard to actually give a code reproduction because it relies on Psalm having a live cache and then that cache changing while live typing. You can see this in the reference issue against psalm-vscode-plugin

See attached PR for resolution: #7247

tm1000:
See #7246 and https://github.com/psalm/psalm-vscode-plugin/issues/107

Specifically if something crashes deep in Psalm we should try to catch most of these crashes in the language server and continue (if possible). This is one such instance where continuing is fine as the error comes from the result of live typing so the code completion (that is erroring) is wrong as the time of typing the letters.

This is really hard to actually give a code reproduction because it relies on Psalm having a live cache and then that cache changing while live typing. You can see this in the reference issue against psalm-vscode-plugin

### DIFF
--- a/src/Psalm/Internal/LanguageServer/Server/TextDocument.php
+++ b/src/Psalm/Internal/LanguageServer/Server/TextDocument.php
@@ -6,7 +6,6 @@ namespace Psalm\Internal\LanguageServer\Server;
 
 use Amp\Promise;
 use Amp\Success;
-use Exception;
 use LanguageServerProtocol\CompletionList;
 use LanguageServerProtocol\Hover;
 use LanguageServerProtocol\Location;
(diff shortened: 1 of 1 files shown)


tm1000:
FYI: The Psalm error should be fixed but it's out of the scope of this PR because it was already broken before my code additions

orklah:
Yeah, it's a false positive, and a pretty convoluted one. You should be able to extract the conditional that checks both variable are not null from the try/catch and it should work fine.

tm1000:
@orklah are you talking about this PR or the bugs outside the context of this PR (sorry just a little confused :-) )

orklah:
Sorry, I was talking about that failure in CI: https://github.com/vimeo/psalm/runs/4669003160?check_suite_focus=true#step:4:30

orklah:
Thanks!

Issue #4619, opened 2020-11-19
Assert properties of immutable objects


SCIF:
It's possible to assert properties of `$this`, but apparently it doesn't work for objects: https://psalm.dev/r/9ae22365ff

Issue #6770, opened 2021-10-28
Assertions of magic properties out of scope doesn't work


SCIF:
Test: https://psalm.dev/r/77a0754fcd

I'll fix this issue soon.

AndrolGenhald:
I think `__get` should additionally be required to be mutation free for this, otherwise you could have this: https://psalm.dev/r/45132855a3

PR #7252, opened 2021-12-30
Allow assertions on mutable object properties.

AndrolGenhald:
Follow up to #4619 to allow assertions on mutable properties as well.
All of the functionality is already there, we just have to not prevent it.

### DIFF
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -240,6 +240,8 @@ $b->s = "boo"; // disallowed
 ### `@psalm-mutation-free`
 
 Used to annotate a class method that does not mutate state, either internally or externally of the class's scope.
+This requires that the return value depend only on the instance's properties. For example, `random_int` is considered
+mutating here because it mutates the random number generator's internal state.
 
 ```php
 <?php
@@ -309,7 +311,7 @@ Used to annotate a class where every property is treated by consumers as `@psalm
 abstract class Foo
 {
     public string $baz;
-  
+
     abstract public function bar(): int;
 }
 
@@ -322,7 +324,7 @@ final class ChildClass extends Foo
     {
         $this->baz = $baz;
     }
-  
+
     public function bar(): int
     {
         return 0;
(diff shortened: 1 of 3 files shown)


AndrolGenhald:
Also, thanks @SCIF for the property assertion support, I hadn't noticed that until today!

SCIF:
Hi @AndrolGenhald , thanks for this PR.

I have found another bug https://github.com/vimeo/psalm/issues/6770 which, probably, is fixed by this PR, so could you please add tests of magic properties?

TBH, I didn't notice `rememberPropertyAssignmentsAfterCall` config directive at the time of initial implementation. But I was quite confused whether I should restrict the assertion scope with read-only properties and mentioned this in my initial PR. There was no a discussion about it so I designed a check this (pretty blunt) way. Honouring of `rememberPropertyAssignmentsAfterCall` would be much better way.

AndrolGenhald:
@SCIF I'll take a look at that one too, currently digging through references and trying to add support for that. It doesn't look like Psalm really tracks references themselves yet, just whether or not they exist.

I'm not sure if `rememberPropertyAssignmentsAfterCall` should affect whether or not the assertions work. If the annotation is there I feel like it should work regardless, but I could see where it might be confusing. If `rememberPropertyAssignmentsAfterCall === false` does prevent the assertions from doing anything, the documentation should probably make that clear, or maybe it should just fail with the previous error you had, and change it to mention that config setting?

AndrolGenhald:
After some thought, I think the solution I was working towards (adding a reference list to `Union`) isn't a good idea. I also thought about adding something to `Context`, but that's still going to require a lot of work to make sure the types stay in sync. I think the best solution is to refactor it so that references are stored differently, and actually point to the variable they reference somehow, but that's going to take a lot of work.

In the meantime, I'd prefer to just mark the failing tests as skipped and go ahead and get this merged. With how little most people use references I don't really expect it to be an issue in practice, even though it'd be nice if it worked correctly.

weirdan:
Thanks!

PR #7255, opened 2021-12-30
[Feat]: Initial Support for LSP Code Actions

tm1000:
This adds initial support for Code Actions in the Language Server.

Code actions allow a user in their respective IDE to right click an error and implement a "quickfix".

The initial work only implements the ability to suppress errors. The reasoning for this is because it seems difficult in Psalms current state to query the diff from Psalter while running in the Language server (More investigation needs to be done here.

Here's an example of this working in VSCode using: https://github.com/psalm/psalm-vscode-plugin

https://user-images.githubusercontent.com/564256/147793880-6cb7e2be-60fb-4ca0-947e-047789a9ded7.mov

The next steps after this would be to enable the ability to auto fix issues from Psalter.

### DIFF
--- a/src/Psalm/Internal/LanguageServer/LanguageServer.php
+++ b/src/Psalm/Internal/LanguageServer/LanguageServer.php
@@ -238,7 +238,7 @@ class LanguageServer extends Dispatcher
                     $this->textDocument = new ServerTextDocument(
                         $this,
                         $codebase,
-                        $this->project_analyzer->onchange_line_limit
+                        $this->project_analyzer
                     );
                 }
 
@@ -246,7 +246,7 @@ class LanguageServer extends Dispatcher
                     $this->workspace = new ServerWorkspace(
                         $this,
                         $codebase,
-                        $this->project_analyzer->onchange_line_limit
+                        $this->project_analyzer
                     );
                 }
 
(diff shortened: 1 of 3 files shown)


tm1000:
Psalms assumption:

```
ERROR: InvalidArgument - src/Psalm/Internal/LanguageServer/Server/TextDocument.php:403:25 - Argument 1 of LanguageServerProtocol\WorkspaceEdit::__construct expects array<array-key, LanguageServerProtocol\TextEdit>, non-empty-array<string, array{LanguageServerProtocol\TextEdit}> provided (see https://psalm.dev/004)
                        [
                            $textDocument->uri => [
                                new TextEdit(
                                    $snippetRange,
                                    "{$indentation}/**\n".
                                    "{$indentation} * @psalm-suppress {$issue->type}\n".
                                    "{$indentation} */\n".
                                    "{$issue->snippet}\n"
                                )
                            ]
                        ]
```

Is actually wrong. The issue is poor documentation in https://github.com/felixfbecker/php-language-server-protocol/blob/9d846d1f5cf101deee7a61c8ba7caa0a975cd730/src/WorkspaceEdit.php

I tried to get around this with psalm-suppress to no avail, every time Psalm highlights it as an error, but it is truly not an issue.

tm1000:
@joehoyle thoughts? As you've worked on the LSP in the past as well

weirdan:
Thanks!

PR #7256, opened 2021-12-30
Remove "is not in project" warnings

tm1000:
Remove the `$file_path . ' is not in project'` warnings in the Language Server. All they do is fill up the output buffer and they seem to bother users.

See:

https://github.com/psalm/psalm-vscode-plugin/issues/9


weirdan:
Thanks!

PR #7261, opened 2022-01-01
Improve negated reconciliation logic

muglug:
This just moves `isset` handling to its rightful place, and also removes a few impossible conditionals (`!=isset` is never seen).


orklah:
Thanks!

PR #7266, opened 2022-01-02
Report missing file path properly

simPod:
If the file at the path is missing, `realpath` gives `false`. It makes no sense to try to put it into error message then. It's better to use `prospective_file_path` instead.


weirdan (approved):

weirdan:
Thanks!

PR #7270, opened 2022-01-03
Drop legacy procedural files

weirdan:


orklah:
I believe spl_object_id can be dropped too, as soon as we raise our min PHP version. The polyfill was there for PHP 7.1

weirdan:
> I believe spl_object_id can be dropped too

Yes, I left it for the PR bumping the minimum PHP version.

PR #7271, opened 2022-01-03
uniqid() returns a non-empty-string

MidnightDesign:


orklah:

PR #7275, opened 2022-01-03
Dropped legacy plugin API

weirdan:


orklah:
+81 -1589, that's so cool!

PR #7279, opened 2022-01-03
Trim baseline selection

weirdan:
Refs vimeo/psalm#6000

/cc: @bdsl


bdsl:
Thanks @weirdan . LGTM if master is intended to be 5 series.

weirdan:
> if master is intended to be the 5 series

It is. We branched 4.x a couple of days ago, so master is now 5.x

PR #7281, opened 2022-01-03
Dropped `html-escaped-string` docs

weirdan:


orklah:
You didn't drop THtmlEscapedString at the same time?

weirdan:
Not yet.

weirdan:
It wasn't properly deprecated, so we have to do that in the 4.x branch and drop it in master.

PR #7285, opened 2022-01-03
Dropped `THtmlEscapedString`

weirdan:


orklah:
@muglug do you know if TSqlSelectString is used by anyone? Does it makes sense to keep it?

muglug:
It doesn’t — ditch it!

weirdan:
Oh, it doesn't exist anymore, only its token still remains. I'll drop it.

image

Issue #7283, opened 2022-01-03
TClassConstant doesn't work with param-out


orklah:
I'd expect the snippet here to work: https://psalm.dev/r/1fe7aac5eb

PR #7287, opened 2022-01-03
Identify ints from TClassConstant when checking for LiteralEquality

orklah:
This is the second part of fixing https://github.com/vimeo/psalm/issues/7283

This refactors the int case from handleLiteralEquality because I started using a piece of code multiple times (what's in getCompatibleIntType)

But it mainly add a TClassConstant resolution inside handleLiteralEquality for int cases. This should probably be done for other cases too but I couldn't find a way to make it clean, so this would probably need a refactor at some point.


orklah:
I made mistakes in there. I'm working on it

PR #7293, opened 2022-01-04
Fix parse_url() return type

### LINKED ISSUE #7078: parse_url() may return false for any combination of inputs
```php
if (parse_url('', -1) === false) {}
```
Gives:
```
ERROR: TypeDoesNotContainType - 2:5 - int|null|string does not contain false
```

https://psalm.dev/r/98278bec0b

Compare this with the source if this function:

https://github.com/php/php-src/blob/051b3967a9408fe60f6f9f62816a90193abbbe39/ext/standard/url.c#L346-L350

Note it returns `false` if a URL fails initial checks.

villfa:
Fixes #7078

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/ParseUrlReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/ParseUrlReturnTypeProvider.php
@@ -112,7 +112,7 @@ class ParseUrlReturnTypeProvider implements FunctionReturnTypeProviderInterface
 
                     if ($component_type->isSingleIntLiteral()) {
                         $component_type_type = $component_type->getSingleIntLiteral();
-                        $is_default_component = $component_type_type->value <= 1;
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Provider/ReturnTypeProvider/ParseUrlReturnTypeProvider.php
```diff
@@ -110,22 +109,29 @@ public static function getFunctionReturnType(FunctionReturnTypeProviderEvent $ev
 
                         return $nullable_falsable_int;
                     }
+
+                    if ($component_type->isSingleIntLiteral()) {
+                        $component_type_type = $component_type->getSingleIntLiteral();
+                        $is_default_component = $component_type_type->value <= 1;
```
orklah, line 115: The snippet https://3v4l.org/uD6RR seems to show that any value inferior or equal to -1 returns an array. Values between 0 and 7 returns null and superior or equal to 8 returns an error

orklah:
Thanks!

PR #7299, opened 2022-01-04
Apply literal number separator rector

weirdan:
I opted for customary 3-digit groups, except version id, where 2-digit groups match the version id encoding.


src/Psalm/Type/Atomic/TCallableObject.php
```diff
@@ -26,7 +26,7 @@ public function toPhpString(
         ?string $this_class,
         int $analysis_php_version_id
     ): ?string {
-        return $analysis_php_version_id >= 72000 ? 'object' : null;
```
weirdan, line 29: This is actually a bug, there's no PHP 7.20, and there never will be.

AndrolGenhald:
It it possible to add this to the CS check so we ensure it stays consistent going forward?

weirdan:
Not with the different grouping for version id. I find this difference quite useful, though.

weirdan:
On the second thought, we can mandate at least one separator for values exceeding 1000 without specifying where it should be placed.

src/Psalm/Type/Atomic/TCallableObject.php
orklah: oh nice

orklah (approved):

PR #7301, opened 2022-01-05
change nullable for array signature to be equal to param signature

swiffer:
https://www.php.net/manual/en/function.session-set-cookie-params

Nullable had been set for `domain` in 7.3 but there is no indication in the docs for that.

Reading the docs again I would assume as `domain`, `path`, `secure` and `httponly` are nullable as of 8.0 the same is for the array signature:

> When using the second signature, an associative array which may have any of the keys lifetime, path, domain, secure, httponly and samesite**. The values have the same meaning as described for the parameters with the same name**. The value of the samesite element should be either Lax or Strict. If any of the allowed options are not given, their default values are the same as the default values of the explicit parameters. If the samesite element is omitted, no SameSite cookie attribute is set. 

Follow Up of #7215

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -12069,7 +12069,7 @@ return [
 'session_reset' => ['bool'],
 'session_save_path' => ['string', 'path='=>'string'],
 'session_set_cookie_params' => ['bool', 'lifetime'=>'int', 'path='=>'?string', 'domain='=>'?string', 'secure='=>'?bool', 'httponly='=>'?bool'],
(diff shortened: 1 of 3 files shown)


weirdan:
In the array form, nulls were accepted since 7.3: https://3v4l.org/MZH5n

orklah:
@swiffer is the answer consistent with what you did? Do you have still changes to do or is this ready?

swiffer:
@orklah - pretty clear - sorry letting this take so long to correct: looks like all attributes are nullable in array form since 7.3

https://3v4l.org/UjE1B

orklah:
Thanks!

PR #7304, opened 2022-01-05
Refactor Reconciler::handleLiteralEquality

orklah:
Sorry for this one, it won't be easy to review.

There is multiple goals here, all intertwined:
- fix some mistakes I made in https://github.com/vimeo/psalm/pull/7287 (mainly, the resolution here: https://github.com/vimeo/psalm/pull/7287/files#diff-7ff683939a23ffc517c44a710a76bfa18959b516c0b77b6c27fb241bc748d83dR1291 was flagging int as `has_int` but they did not match with know literals here: https://github.com/vimeo/psalm/pull/7287/files#diff-7ff683939a23ffc517c44a710a76bfa18959b516c0b77b6c27fb241bc748d83dR1332 so there were failures in some cases)
- improve the system by resolving the ClassConst at the very beggining (instead of just in int handling as I've done before)
- modernizing the code here. I suspect the code here was pretty old in Psalm history. There are way of checking things I never saw until now. For example instead of identifying a TLiteralInt and checking its value, the code here: https://github.com/vimeo/psalm/pull/7287/files#diff-7ff683939a23ffc517c44a710a76bfa18959b516c0b77b6c27fb241bc748d83dR1339 just uses the offset of the array of types (that contains the return of Atomic::getKey()) in order to check the equality with the assertion. I replaced that by a `$atomic instanceof TLiteralInt && $atomic->value === $assertion_value`. This is longer but infinitely more clear to read. We have a similar thing here: https://github.com/vimeo/psalm/pull/7287/files#diff-7ff683939a23ffc517c44a710a76bfa18959b516c0b77b6c27fb241bc748d83dR1388 where we just tweak `float(5)` to fetch the "5" value.

I wanted to keep a clear history of commits but this got out of control quickly. So I think I'll smash the majority of later commits together. I kept them for now for easier reading

I'll probably refactor the float case now that I started, to keep consistency. This has high chances of resolving some old bugs, but if I'm honest, it will probably create some (though they should be easier to fix). It will probably change some TypeDoesNotContainType into DocblockTypeContradiction or the other way around because I was more consistent with keeping the origin of the type than before.

I thought about pushing this on Psalm 5 instead, but I fear that if we do that, we'll have to move #7286, #7287 and #7288 too and I'm not sure if we should or if it's easy to do.

(My goal when I started those changes was to add `Reconciler::RECONCILIATION_*` in @param-out. If I knew I'd have to do all that, I may not have started!)

### DIFF
--- a/src/Psalm/Internal/Type/AssertionReconciler.php
+++ b/src/Psalm/Internal/Type/AssertionReconciler.php
@@ -1042,98 +1042,20 @@ class AssertionReconciler extends Reconciler
                 $suppressed_issues
             );
         } elseif ($scalar_type === 'float') {
-            $value = (float) $value;
-
-            if ($existing_var_type->hasMixed() || $existing_var_type->hasScalar() || $existing_var_type->hasNumeric()) {
-                if ($is_loose_equality) {
-                    return $existing_var_type;
-                }
-
-                return new Union([new TLiteralFloat($value)]);
-            }
-
-            if ($existing_var_type->hasFloat()) {
-                $existing_float_types = $existing_var_type->getLiteralFloats();
-
-                if ($existing_float_types) {
-                    $can_be_equal = false;
-                    $did_remove_type = false;
-
-                    foreach ($existing_var_atomic_types as $atomic_key => $_) {
-                        if ($atomic_key !== $assertion) {
-                            $existing_var_type->removeType($atomic_key);
-                            $did_remove_type = true;
-                        } else {
-                            $can_be_equal = true;
-                        }
-                    }
-
-                    if ($var_id
-                        && $code_location
-                        && (!$can_be_equal || (!$did_remove_type && count($existing_var_atomic_types) === 1))
-                    ) {
-                        self::triggerIssueForImpossible(
-                            $existing_var_type,
-                            $old_var_type_string,
-                            $var_id,
-                            $assertion,
-                            $can_be_equal,
-                            $negated,
-                            $code_location,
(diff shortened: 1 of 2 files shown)


orklah:
Note: the commit https://github.com/vimeo/psalm/commit/72fd3de8861e69aa42f21a413e163f0b7a29a28e on master will most certainly be a conflict with this PR. It should not be too bad to resolve I think

muglug:
From a quick skim this all looks good. In my Rust-based fork I took the time to change string-based assertions to an enum (which in Psalm could be done with regular class inheritance)

```rust
pub enum Assertion {
    Any, // used for the null state
    IsType(TAtomic),
    IsNotType(TAtomic),
    Falsy,
    Truthy,
    IsEqual(TAtomic),
    IsNotEqual(TAtomic),
    IsEqualIsset,
    IsIsset,
    IsNotIsset,
    HasStringArrayAccess,
    HasIntOrStringArrayAccess,
    ArrayKeyExists,
    ArrayKeyDoesNotExist,
    InArray(TUnion),
    NotInArray(TUnion),
    HasArrayKey(String),
    NonEmptyCountable(bool),
    EmptyCountable,
    HasExactCount(usize),
    DoesNotHaveExactCount(usize),
}
```

Psalm's version would need a couple more of these like `IsLooseEqual(..)` and `IsNotLooseEqual(..)` to support things that are no longer supported in Hack, but it makes the logic around unpacking assertions quite a bit more clean, and vastly reduces string manipulation.

PR #7313, opened 2022-01-05
replace `array<never, never>` as a way to detect empty arrays by a dedicated method

orklah:
This is a pattern we had on different places but it's not great because technically, `array<array-key, never>` is also an empty array.

### DIFF
--- a/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
+++ b/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
@@ -554,7 +554,7 @@ class SimpleAssertionReconciler extends Reconciler
                 if (!$array_atomic_type instanceof TNonEmptyArray
                     || ($array_atomic_type->count < $min_count)
                 ) {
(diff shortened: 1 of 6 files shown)


src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -554,7 +554,7 @@ private static function reconcileNonEmptyCountable(
                 if (!$array_atomic_type instanceof TNonEmptyArray
                     || ($array_atomic_type->count < $min_count)
                 ) {
-                    if ($array_atomic_type->getId() === 'array<never, never>') {
+                    if (!$array_atomic_type->isEmptyArray()) {
```
weirdan, line 557: You inverted the condition here.

orklah: oh thanks! That explains a lot!

weirdan (approved):

Issue #7305, opened 2022-01-05
First class callables causes crash in some contexts


Riimu:
Despite #6989 and #7196, it seems the first class callables are still causing some crashes. Here is a simple example (https://psalm.dev/r/2cfb6f69bb)

```php
<?php call_user_func(array_map(...), intval(...), ['1']);
```

```
D:\Projects\psalm-bug
λ vendor\bin\psalm
Target PHP version: 8.1 (inferred from current PHP version)
Scanning files...
Uncaught AssertionError: assert(!$this->isFirstClassCallable()) in D:\Projects\psalm-bug\vendor\nikic\php-parser\lib\PhpParser\Node\Expr\CallLike.php:36
Stack trace:
#0 D:\Projects\psalm-bug\vendor\nikic\php-parser\lib\PhpParser\Node\Expr\CallLike.php(36): assert(false, 'assert(!$this->...')
#1 D:\Projects\psalm-bug\vendor\vimeo\psalm\src\Psalm\Internal\PhpVisitor\Reflector\ExpressionScanner.php(171): PhpParser\Node\Expr\CallLike->getArgs()
#2 D:\Projects\psalm-bug\vendor\vimeo\psalm\src\Psalm\Internal\PhpVisitor\Reflector\ExpressionScanner.php(87): Psalm\Internal\PhpVisitor\Reflector\ExpressionScanner::registerClassMapFunctionCall(Object(Psalm\Codebase), Object(Psalm\Storage\FileStorage), Object(Psalm\Internal\Scanner\FileScanner), Object(Psalm\Aliases), 'array_map', Object(PhpParser\Node\Expr\FuncCall), NULL, NULL)
#3 D:\Projects\psalm-bug\vendor\vimeo\psalm\src\Psalm\Internal\PhpVisitor\ReflectorVisitor.php(346): Psalm\Internal\PhpVisitor\Reflector\ExpressionScanner::scan(Object(Psalm\Codebase), Object(Psalm\Internal\Scanner\FileScanner), Object(Psalm\Storage\FileStorage), Object(Psalm\Aliases), Object(PhpParser\Node\Expr\FuncCall), NULL, NULL)
#4 D:\Projects\psalm-bug\vendor\nikic\php-parser\lib\PhpParser\NodeTraverser.php(123): Psalm\Internal\PhpVisitor\ReflectorVisitor->enterNode(Object(PhpParser\Node\Expr\FuncCall))
[...]

PR #7321, opened 2022-01-06
Prevented crashes on `array_map(...)`

weirdan:
Fixes vimeo/psalm#7305

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ExpressionScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ExpressionScanner.php
@@ -125,6 +125,10 @@ class ExpressionScanner
             }
         }
 
+        if ($node->isFirstClassCallable()) {
+            return;
+        }
+
         if ($function_id === 'define') {
             $first_arg_value = isset($node->getArgs()[0]) ? $node->getArgs()[0]->value : null;
             $second_arg_value = isset($node->getArgs()[1]) ? $node->getArgs()[1]->value : null;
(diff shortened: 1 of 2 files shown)


orklah:
maybe add a TODO to signal that we're missing something at that place?

weirdan:
Are we?

orklah:
I didn't look in details what the code here does but I'd expect first class callable to be replaced with a callable or something, no?

weirdan:
With this patch applied, we already do the best we can short of supporting generic closures. For `array_map(...)` Psalm infers `pure-Closure(callable|null, array<array-key, mixed>, array<array-key, mixed>...=):array<array-key, mixed>`. I don't see how it could get any better (without generic closures, that is).

orklah:
Nevermind then 🙂 what do you call generic closures?

weirdan:
E.g. this `Closure<T1 of scalar, T2 of object>(T1, T2):T1|T2`

tm1000:
This actually results in a bug:

```Call to undefined method PhpParser\Node\Expr\FuncCall::isFirstClassCallable()```

That method `isFirstClassCallable` does not exist on `PhpParser\Node\Expr\FuncCall`

orklah:
FuncCall extends CallLike which has isFirstClassCallable method. Mind posting the entire stack along with your installed version of PHP-Parser in a new issue?

tm1000:
@orklah see https://github.com/vimeo/psalm/issues/7378

It does not have that call.

tm1000:
@orklah figured it out: see: https://github.com/vimeo/psalm/issues/7378

Thanks for pointing that out! sorry for the distractions!

Issue #7243, opened 2021-12-30
xi:fallback does not work


weirdan:
`<xi: fallback>` is supposed to kick in when the included file is not available, but Psalm instead throws the following exception:

```
Uncaught RuntimeException: PHP Error: DOMDocument::xinclude(): I/O warning : failed to load external entity "/home/weirdan/src/psalm/psalm/zz.xml" in /home/weirdan/src/psalm/psalm/src/Psalm/Config.php:694 in /home/weirdan/src/psalm/psalm/src/Psalm/Internal/ErrorHandler.php:66
Stack trace:
#0 [internal function]: Psalm\Internal\ErrorHandler::Psalm\Internal\{closure}()
#1 /home/weirdan/src/psalm/psalm/src/Psalm/Config.php(694): DOMDocument->xinclude()
#2 /home/weirdan/src/psalm/psalm/src/Psalm/Config.php(711): Psalm\Config::loadDomDocument()
#3 /home/weirdan/src/psalm/psalm/src/Psalm/Config.php(679): Psalm\Config::validateXmlConfig()
#4 /home/weirdan/src/psalm/psalm/src/Psalm/Config.php(644): Psalm\Config::loadFromXML()
#5 /home/weirdan/src/psalm/psalm/src/Psalm/Config.php(596): Psalm\Config::loadFromXMLFile()
#6 /home/weirdan/src/psalm/psalm/src/Psalm/Internal/CliUtils.php(494): Psalm\Config::getConfigForPath()
#7 /home/weirdan/src/psalm/psalm/src/Psalm/Internal/Cli/Psalm.php(563): Psalm\Internal\CliUtils::initializeConfig()
#8 /home/weirdan/src/psalm/psalm/src/Psalm/Internal/Cli/Psalm.php(1009): Psalm\Internal\Cli\Psalm::loadConfig()
#9 /home/weirdan/src/psalm/psalm/src/Psalm/Internal/Cli/Psalm.php(231): Psalm\Internal\Cli\Psalm::initConfig()
#10 /home/weirdan/src/psalm/psalm/psalm(7): Psalm\Internal\Cli\Psalm::run()
#11 {main}
(Psalm dev-master@1c078136273a669d52d234251ddbae4cd0507d38 crashed due to an uncaught Throwable)
```

when `<xi:include>` is used in `issueHandlers` element:

```xml
        <xi:include href="zz.xml">
            <xi:fallback>
                <MixedAssignment>
                    <errorLevel type="suppress">
[...]

PR #7324, opened 2022-01-06
Issue #7243: Do not throw exception on xinclude with fallbacks

vstm:
This will fix issue #7243, it will suppress all warnings emited by xinclude (they will not be converted to exceptions via ErrorHandler). Errors during xinclude will still be converted to exceptions (for example if there is an invalid xinclude without fallback).

I branched from master - is that still ok, because I've seend there's a new 4.x branch? I can rebase to that if needed.


weirdan:
> I branched from master - is that still ok, because I've seend there's a new 4.x branch? I can rebase to that if needed.

Please do. `master` is now receiving BC breaking changes, and this one is still compatible with the 4.x line.

weirdan:
Thanks!

Issue #7322, opened 2022-01-06
"Cannot treat type object as callable" when implementing __invoke


ricardoboss:
https://psalm.dev/r/0db80c0f40

This only happens when using first-class callables. When I switch out `$classInstance(...)` with `Closure::fromCallable($classInstance)` it works.

villfa:
Here another example to show that is not only related to first-class callables: https://psalm.dev/r/cf384c261f

orklah:
It's handled in FunctionCallAnalyzer::getAnalyzeNamedExpression starting from line 637. We loop over the type in the union in order to detect potential callables. It seems your two examples are not detected as callables. It should not be too difficult to handle them if you want to try

villfa:
Here more examples to illustrate the remaining issues:

* works : https://psalm.dev/r/001d69e1dc
* doesn't work: https://psalm.dev/r/d78ef327e9
* doesn't work: https://psalm.dev/r/001d69e1dc

orklah:
Those are gonna be hard. We don't have a type TClassStringWithProperties...

orklah:
I'll close this as I don't think it's worth implementing. I'd not be opposed to merging an implementation of that but it's a lot of work for such a small use case

AndrolGenhald:
I'll see if it's easier to fix with my type system rewrite, but it looks like that will probably have to wait for Psalm 6 at this point.

PR #7325, opened 2022-01-06
Fix analysis when __invoke() exists

villfa:
Fixes partially #7322


orklah:
That's great! Thanks!

Issue #7323, opened 2022-01-06
False-positive: NoValue


zmitic:
Example: https://psalm.dev/r/5670fbfa74

I have a case when returned Closure must immediately throw exception when called, but under some condition (used random_int for that). Adding `never` to closure returns `InvalidReturnType`.

ricardoboss:
Seems to be an issue with short function notation: https://psalm.dev/r/0651e3f89e

orklah:
> Seems to be an issue with short function notation: https://psalm.dev/r/0651e3f89e

Not exactly. Psalm considers @zmitic's snippet is identical to this: https://psalm.dev/r/d30dcfd371. It's the `return throws ...` that causes an issue. In short closures, the return is implied.

There is also a second issue when trying to document the return type of the short closure: https://psalm.dev/r/69b04c1caf

So, to fix this, Psalm should allow never returning statements in short closures  (calling exit() will do the same) and also allow implicit returns of never returning statements when the return type is `never`

This seems to be also an issue with `void` short closures: https://psalm.dev/r/eb3536c071

orklah:
The case with void is actually correct. PHP will emit a fatal error no matter what the content of the closure is because it will try to return it: https://3v4l.org/sH2Zn

orklah:
oh, actually, the explicit never case is not possible either. This is more weird: https://3v4l.org/JFuMf

PR #7326, opened 2022-01-06
allow short closure to return never

orklah:
This will fix https://github.com/vimeo/psalm/issues/7323

There was 2 bugs:
- Psalm emitted a NoValue when a never returning statement was in the closure. This is due to short closure performing a return implicitly. The ReturnAnalyzer make sure to emit a NoValue when returning a Never. I fixed this by suppressing NoValue in the context of a short closure. It's not the most direct way but I couldn't see how to pass the info that I was in a short closure to the ReturnAnalyzer. Even a flag in Context would have been weird because they seem to be nestable.
- Kinda similarly, whenever there is a return in a functionlike scope and the functionlike declare returning void or never, an InvalidReturnStatement is emitted. I solved that by making sure a `return (never)` is seen as an exit instead of being seen as a return.

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
@@ -464,6 +464,12 @@ abstract class FunctionLikeAnalyzer extends SourceAnalyzer
             $this->track_mutations = true;
         }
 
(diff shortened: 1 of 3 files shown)


orklah:
I'm putting this on hold until we have a response to https://github.com/php/php-src/issues/7900. The change I made is compatible whether this is fixed or not, but I added a comment that would loose relevance if this is fixed

orklah:
I'll merge this as-is, the issue on php-src has been flagged as "feature"

Issue #7329, opened 2022-01-07
Uncaught Error: Class "Composer\InstalledVersions" not found [phar]


thomasbley:
```
psalm.phar -v
Uncaught Error: Class "Composer\InstalledVersions" not found in phar:///home/psalm.phar/src/Psalm/Internal/CliUtils.php:116
Stack trace:
#0 phar:///home/psalm.phar/src/Psalm/Internal/Cli/Psalm.php(188): Psalm\Internal\CliUtils::requireAutoloaders()
#1 phar:///home/psalm.phar/src/Psalm/Internal/IncludeCollector.php(31): Psalm\Internal\Cli\Psalm::Psalm\Internal\Cli\{closure}()
#2 phar:///home/psalm.phar/src/Psalm/Internal/Cli/Psalm.php(189): Psalm\Internal\IncludeCollector->runAndCollect()
#3 phar:///home/psalm.phar/psalm(7): Psalm\Internal\Cli\Psalm::run()
#4 /home/psalm.phar(14): require('...')
#5 {main}
(Psalm (unknown version) crashed due to an uncaught Throwable)
```

Version 4.18, PHP 8.1.1 or 8.0.x

Source code inside phar:

![Screenshot_20220107_110341](https://user-images.githubusercontent.com/941223/148528077-a748b283-f4b7-41d7-85cc-55e00009c37d.png)

villfa:
A new version of Box has been released 4 days ago: https://github.com/box-project/box/releases/tag/3.14.1
It's probably related.

A way to avoid a surprise like that could be to commit the `composer.lock` file in `vendor-bin/box`.

orklah:
Not sure it's related. We merged this https://github.com/vimeo/psalm/pull/7239 in last version. There must be something wrong here, possibly related to our scoper.inc.php file

orklah:
Hey All, can some of you try this phar and tell me if it fixes your issue? https://30363-74386167-gh.circle-artifacts.com/0/build/psalm.phar

orklah:
Note: One of phpdoc's repo is failing due to another error with the phar:
https://github.com/phpDocumentor/TypeResolver/runs/4743694679?check_suite_focus=true#step:7:9

villfa:
I confirm I have the same error:
```
./psalm.phar -v
Uncaught OutOfBoundsException: Package "nikic/php-parser" is not installed in /vendor/composer/InstalledVersions.php:180
Stack trace:
[...]

PR #7337, opened 2022-01-07
replace class name by full FQN for scoper

orklah:
This is an attempt to fix https://github.com/vimeo/psalm/issues/7329


orklah:
I've got to merge this to generate the file it seems.

Issue #7328, opened 2022-01-07
Array assignment type inconsistencies


AndrolGenhald:
https://psalm.dev/r/d45a62e39a
Seems to me like `$a` and `$b` should behave like `$c`.

orklah:
The $b case is an error for Psalm. You should always define your variables before using them. Furthermore, you're in global scope here so there's nothing preventing $b to already have another value.

The $c case is an equivalent form of $a. In fact, $c is more precise than $a (we could probably improve the behaviour of $a in this case to count elements). The difference is that when Psalm see `[0]`, it translate that to an array shape internally and then, it builds upon it once it's created. However, when it sees `$a[] =` on an empty array, it's more common to be on a list behaviour.

Is the $c case causing bugs? I guess it could be changed, but I'm not sure this would be worth it

AndrolGenhald:
> The $b case is an error for Psalm. You should always define your variables before using them. Furthermore, you're in global scope here so there's nothing preventing $b to already have another value.

I'm not entirely convinced it should be. PHP issues no warnings for that code, and while Psalm is more strict than PHP in many cases, I'm not sure it's necessary here. Even if we do want Psalm to be more strict here, I think the way it is now is still incorrect behavior, if `$b` is considered possibly defined is an unresolvable include, then it could also be an `ArrayObject` and the current type is too narrow. If we don't bother considering globals possibly defined somewhere else, then it should be `array{1}`. Either way, for [non-global scope](https://psalm.dev/r/1c3435cde8) I think it should definitely be `array{1}` and should _maybe_ emit `UndefinedVariable` (why is it `PossiblyUndefinedVariable` now?). 

> The $c case is an equivalent form of $a. In fact, $c is more precise than $a (we could probably improve the behaviour of $a in this case to count elements). ... Is the $c case causing bugs? I guess it could be changed, but I'm not sure this would be worth it

The `$c` case is perfect, that's what I think they should all be.

AndrolGenhald:
For context, I'm working on references, and the `$b` case here turns out to be very similar to this: https://psalm.dev/r/422568e244 https://3v4l.org/MnfES

orklah:
[...]

PR #7339, opened 2022-01-07
Add support for references and improve UnusedVariable checks (fixes #7254).

### LINKED ISSUE #7254: Support references
References aren't really supported by Psalm at the moment outside of `ReferenceConstraintViolation` and related issues when using references in arguments. It would be nice to fully support references.

https://3v4l.org/mQ809
https://psalm.dev/r/d79b9b2f0f

See also #7252 (when fixed skipped tests should be unskipped).

AndrolGenhald:
Adds support for references by setting `Context::$vars_in_scope['$reference']` as a reference to `Context::$vars_in_scope['$target']`.
Adding the reference to `$vars_in_scope` basically fixed everything by itself, except for the `Unused*` checks, which required quite a lot of additional work.


src/Psalm/Codebase.php
```diff
@@ -1967,9 +1967,7 @@ public function addTaintSource(
 
         $this->taint_flow_graph->addSource($source);
 
-        $expr_type->parent_nodes = [
-            $source->id => $source,
-        ];
+        $expr_type->parent_nodes[$source->id] = $source;
```
AndrolGenhald, line 1970: This isn't really that closely related to everything else in this PR, but I'm pretty sure it's a correct change. I'm not familiar enough with the taint feature to know how to test it, but I worry that without this change some of the taint stuff might mess up the `parent_nodes` for a reference.

src/Psalm/Context.php
```diff
@@ -41,6 +42,54 @@ class Context
...
+     *
+     * @var array<string, int<0, max>>
+     */
+    public $referenced_counts = [];
+
+    /**
+     * Maps references to referenced variables for the current scope.
+     * With `$b = &$a`, this will contain `['$b' => '$a']`.
+     *
+     * All keys and values in this array are guaranteed to be set in $vars_in_scope.
```
AndrolGenhald, line 56: I'm fairly confident I caught all the places a var_in_scope might be unset, but if this isn't true for some reason I consider that a bug to be fixed.

src/Psalm/Context.php
```diff
@@ -667,7 +753,7 @@ public function removeDescendents(
 
         foreach ($this->vars_in_scope as $var_id => $type) {
             if (preg_match('/' . preg_quote($remove_var_id, '/') . '[\]\[\-]/', $var_id)) {
-                unset($this->vars_in_scope[$var_id]);
+                $this->remove($var_id, false);
```
AndrolGenhald, line 756: Not sure why this wasn't unsetting `$this->vars_possibly_in_scope` before, but it works fine this way. I'm guessing it was just overlooked? Or maybe `$this->vars_possibly_in_scope` is guaranteed to not contain descendents? (side note, that should actually be descendant, descendent is an adjective with a subtly different meaning).

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -235,6 +243,16 @@ public static function analyze(
...
                         );
                     }
+                    $statements_analyzer->node_data->setType($stmt, $stmt_type);
+
+                    if ($assigned_to_reference) {
+                        // Since this variable was created by being assigned to as a reference (ie for
+                        // `$a = &$b` this variable is $b), we need to analyze it as an assignment to null.
+                        AssignmentAnalyzer::analyze($statements_analyzer, $stmt, null, $stmt_type, $context, null);
+
+                        // Stop here, we don't want it to be considered possibly undefined like the array case.
```
AndrolGenhald, line 253: Per #7328 I don't think the array should be PossiblyUndefined either, but I'll think about it and maybe do a separate pull request for that.

src/Psalm/Internal/Analyzer/Statements/GlobalAnalyzer.php
```diff
@@ -79,6 +92,20 @@ public static function analyze(
                         $var,
                         $var_id
                     );
+
+                    if ($statements_analyzer->data_flow_graph instanceof VariableUseGraph) {
+                        if ($global_context !== null && $global_context->hasVariable($var_id)) {
+                            // TODO get global CodeLocation so that global variable can be marked as unused if it is
+                            // unused inside the function. Marking it as a referenced global causes it to be marked
+                            // used if it's used as a global in any function, even if it's unused in that function.
```
AndrolGenhald, line 100: I would love for this to work, but I've spent some time looking and I still have no idea how to find the statement where the global was declared in the global scope.
```
$a = 0; // How do I find this statement so that it can be marked as unused

function foo(): void {
    global $a; // when analyzing this statement
}
```

src/Psalm/Context.php
```diff
@@ -458,6 +502,28 @@ public function update(
...
+     * Updates the list of possible references from a confusing scope,
+     * such as a reference created in an if and then later reused.
+     */
+    public function updateReferencesPossiblyFromConfusingScope(
+        Context $confusing_scope_context,
+        StatementsAnalyzer $statements_analyzer
+    ): void {
+        foreach ($confusing_scope_context->references_in_scope + $confusing_scope_context->references_to_external_scope
+            as $reference_id => $_
+        ) {
```
AndrolGenhald, line 515: I cannot for the life of me figure out how to make phpcs happy here. I'll probably end up using a temporary variable, but this seems like an issue either in our phpcs configuration or in phpcs itself.

tests/FileManipulation/UnusedVariableManipulationTest.php
```diff
@@ -348,14 +348,27 @@ function foo() : void {
                 '<?php
                     function foo() : void {
                         $a = 5;
-                        $b = &$a;
```
orklah, line 351: weird, this test was basically testing nothing...

tests/ReferenceTest.php
```diff
@@ -0,0 +1,305 @@
...
+    use ValidCodeAnalysisTestTrait;
+
+    /**
+     * @return iterable<string,array{string,assertions?:array<string,string>,error_levels?:string[]}>
+     */
+    public function providerValidCodeParse(): iterable
+    {
+        return [
+            'updateReferencedTypes' => [
+                '<?php
```
orklah, line 22: can you test what happens if you change the value of $a after creating the reference in $b?

tests/FileManipulation/UnusedVariableManipulationTest.php
AndrolGenhald: Yep, it was confirming incorrect behavior...
Unfortunately the actual correct behavior doesn't work yet either, but I figure I'll leave it for a future improvement.

tests/ReferenceTest.php
AndrolGenhald: Hmm, I thought I tested that sufficiently elsewhere, but now I'm not so sure.

tests/ReferenceTest.php
```diff
@@ -0,0 +1,305 @@
...
+                    $arr = [1, 2, 3];
+                    foreach ($arr as &$i) {
+                        ++$i;
+                    }
+                    foreach ($arr as &$i) {
+                        ++$i;
+                    }
+                ',
+            ],
+        ];
```
orklah, line 153: can you add a test to see if https://3v4l.org/omrIG makes Psalm crashes and/or create a blackhole?

src/Psalm/Internal/Analyzer/Statements/GlobalAnalyzer.php
orklah: This is basically not solvable:
- the statement could be in another file
- there could be multiple statement in multiple files with multiple types
- even if the statement is in the same file, how do you make sure you executed it when you called the function?
- there could be a side-effect from another function that could change the variable type in another function declared in another file

Global scope is a mess

AndrolGenhald: Yeah, that's what I was afraid of. Guess I'll take out the TODO and related stuff. 🙁

src/Psalm/Codebase.php
orklah: Let's fix that when we have a reported bug then. I see we have multiple instances of both ways of declaring parent_nodes and I'm not sure when to choose which

tests/ReferenceTest.php
orklah: Also this kind of mumbo jumbo: https://3v4l.org/hsBCq

tests/FileManipulation/UnusedVariableManipulationTest.php
orklah: I hate that kind of tests when they're not flagged properly. When you break them, you just have no idea if the test is legit or not

src/Psalm/Context.php
orklah: well, vars_possibly_in_scope is just an array of bools. It's nothing without vars_in_scope so yeah, probably overlooked and never caused issue

tests/FileManipulation/UnusedVariableManipulationTest.php
AndrolGenhald: Agreed, but at least this one was obvious.

src/Psalm/Codebase.php
AndrolGenhald: I was leaning that way originally, but it looks like this only exists for plugins to use, and based on the name it's supposed to **add** a source, not remove all other sources as well.
I think the only time `parent_nodes` should be set to an array rather than setting an offset is when the previous nodes are known to be wrong, such as when assigning a variable.

```
$foo = 'a';
$foo = 'b'; // Should no longer have a parent node for `$foo = 'a'`
```

orklah: I'm unclear on what can cause a variable to have multiple parents though

src/Psalm/Context.php
orklah:
```php
foreach ($confusing_scope_context->references_in_scope
                     + $confusing_scope_context->references_to_external_scope as $reference_id => $_
            ) {
            }
```
This makes it happy for me

src/Psalm/Codebase.php
AndrolGenhald: I'm a bit unsure about that myself. I utilized multiple parents for references, but outside of that I don't know what use it has, other than it probably has something to do with taints.

src/Psalm/Context.php
AndrolGenhald: Of course, one of the few possibilities I didn't try. Honestly though, having a multiple line `foreach` without the `as` on its own line just seems more confusing to me. I prefer the extra variable to that.

orklah: Agree :p

orklah:
Just thinking about it, you removed every unset from $context->vars_in_scope, but we could had some. Would that be a problem?

Should we target Psalm 5 after telling plugin developpers to use your new function?

AndrolGenhald:
> Just thinking about it, you removed every unset from $context->vars_in_scope, but we could had some. Would that be a problem?
> 
> Should we target Psalm 5 after telling plugin developpers to use your new function?

I already targeted Psalm 5 because I removed `Context::$vars_from_global`, but that's another thing to think about.

I'm going to convert this to a draft for now, I'm trying to get assertions working on references, and I discovered that PHP's reference implementation is a broken mess.

src/Psalm/Type/Reconciler.php
```diff
@@ -96,6 +101,32 @@ public static function reconcileKeyedTypes(
             return $existing_types;
         }
 
+        $reference_map = [];
+        if (!empty($existing_references)) {
+            // PHP behaves oddly when passing an array containing references: https://bugs.php.net/bug.php?id=20993
```
AndrolGenhald, line 106: So [this](https://3v4l.org/3s7B5) is fine and makes perfect sense. But then you do [this](https://3v4l.org/G1VRc), which is the exact same thing but with an array item, and everything falls apart. So what you end up with is this [black magic](https://3v4l.org/tNYYj) to work around the problem.

I've come to the conclusion that references in PHP are a half-supported buggy mess, and if you use references in arrays or objects you're asking for trouble.

AndrolGenhald:
I think there's still quite a bit of improvement that could be done, but the basic support is there now and for references that aren't array items or object properties things more or less work as expected. I think I'll leave it here unless someone reports a bug with it, it seems to me that relying on references in PHP for anything more than basic local variable modification is a bad idea, and there are probably much more impactful things that could be worked on.

UPGRADING.md
```diff
@@ -115,6 +115,8 @@
     - `Psalm\Type\Union`
  - While not a BC break per se, all classes / interfaces / traits / enums under
    `Psalm\Internal` namespace are now marked `@internal`.
+ - To remove a variable from the context, Context::remove(). Calling
+   `unset($context->vars_in_scope[$var_id])` can cause problems when using references.
```
AndrolGenhald, line 119: Kind of makes me want `@psalm-external-readonly` that would be like `@psalm-readonly-allow-private-mutation` but instead of restricting modifications to the class itself, it would restrict it to a namespace.

orklah:
Can you describe on https://github.com/vimeo/psalm/blob/master/UPGRADING.md that unset on vars_in_scope is now forbidden? If you see any other BC break, please add them also.

AndrolGenhald:
@orklah That's already in there.

orklah:
Sorry :( I did not look back at the changes.

I'd like to merge that soon, can you do a rebase?

orklah:
Looks like merging the constant one created a lot more conflict. Please ping me after rebase and I'll merge this

AndrolGenhald:
@orklah done. That's the most painful rebase I've had to do in a while, there have been a good number of large changes recently like the `'code'` key in tests.

orklah:
Yep, sorry for that, but it's good news, the project is alive :)

orklah:
Thanks!

PR #7340, opened 2022-01-07
Simplify object comparison

muglug:
The original ObjectComparator code was quite hard to follow. This attempts to make it more clear, fixing a bug or two along the way.


orklah:
Cool! Thanks!

PR #7347, opened 2022-01-08
Fix dir Callmap

### LINKED ISSUE #7345: `dir` has wrong return type Directory|false|null instead of Directory|false
Hi @orklah, 

I run into the following issue
- Psalm consider `dir` to return `Directory|false|null`
- PHPStan consider `dir` to return `Directory|false`

The previous type was introduced by @voku in https://github.com/vimeo/psalm/pull/1541

I did some research, and it seems that the php doc was updated recently from `Directory|false|null` to `Directory|false`.
https://github.com/php/doc-en/pull/903

But I'm not sure which CallMap should be modified then...

VincentLanglet:
Closes #7345

If I understand correctly your message @weirdan 
> But since this was zpp-null, you need to drop null from all callmaps.


weirdan (approved):

weirdan:
> If I understand correctly your message

You absolutely did, thanks!

Issue #7353, opened 2022-01-09
Feature request: merging `issueHandlers` elements


greg0ire:
On `doctrine/orm`, we would like to be able to run Psalm with `doctrine/dbal` 2 and 3 without repeating code.
Currently we are able to do so because the set of errors we get for DBAL 3 is included in the set of errors we get for DBAL 2, so having one file does the trick. The downside is that it breaks `--update-baseline`, which will remove DBAL-2-specific issues.

I've attempted to leverage the include system to have a DBAL-2-specific file, but I got stuck because including a file that has an `issueHandlers` root element seems to result in subsequent `issueHandlers` being ignored.

It would be great to be able to have several `issueHandlers` elements in a single config and have them deep-merged together.

orklah:
@greg0ire If I'm not mistaken, @wouterj's PR fixed part of the error. The other issue was that deep nested merge between two issueHandlers is not done yet, but this was reported here already: https://github.com/vimeo/psalm/issues/6888 so I'm closing this one

Issue #6888, opened 2021-11-11
[bug] XSD allows you to use the same issue handler in config, while only loading the last one


dkarlovi:
Example

```xml
<?xml version="1.0"?>
<psalm xmlns="https://getpsalm.org/schema/config"
       cacheDirectory="var/phpqa/psalm"
       allowStringToStandInForClass="true">
    <projectFiles>
        <directory name="./"/>
    </projectFiles>

    <issueHandlers>
        <PropertyNotSetInConstructor>
            <errorLevel type="info">
                <file name="src/App/Kernel.php"/>
            </errorLevel>
        </PropertyNotSetInConstructor>

        <!-- other stuff -->

        <PropertyNotSetInConstructor>
            <errorLevel type="info">
                <directory name="tests/Behat/Context/"/>
            </errorLevel>
        </PropertyNotSetInConstructor>
    </issueHandlers>
</psalm>
```

This is invalid config, the schema doesn't pick it up.

ThomasLandauer:
My use case:
```xml
<issueHandlers>
    <UnusedParam errorLevel="suppress" />

    <!-- other stuff -->

    <UnusedParam>
        <errorLevel type="suppress">
            <directory name="tests" />
        </errorLevel>
    </UnusedParam>
</issueHandlers>
```
I've added the top `<UnusedParam>` due to https://github.com/psalm/psalm-plugin-doctrine/issues/157, and was wondering why those errors wouldn't disappear...

PR #7354, opened 2022-01-09
Support multiple <issueHandlers> elements

wouterj:
Fixes part of #7353 

This does not handle deep merging issue handler configs. When defining the same issue handler in 2 `<issueHandlers>` elements, the latter overrides the first. This is the same behavior as when defining the same issue handler twice in a single `<issueHandlers>` element.


dkarlovi:
See https://github.com/vimeo/psalm/issues/6888

orklah:
Thanks!

PR #7355, opened 2022-01-09
Fix faulty Config tests

wouterj:
Caught while writing the test for #7354

`realpath()` returns `false` if the path does not exists. This caused these tests to not test what they were expected to (due to `false` being transformed into `""`, which never matches the project paths).


weirdan:
Thanks!

PR #7358, opened 2022-01-09
Rename was_static property to is_static to be more accurate

muglug:
Using `"was"` here doesn't make much sense, because the type being analysed _is_ `static` aka `this`.


orklah:
Great! I remember wondering if there was a meaning behind "was"! Thanks!

PR #7359, opened 2022-01-09
feat(types): add properties-of<T> type

### LINKED ISSUE #3141: Pseudo-types for referencing class properties and methods?
Just throwing out vague idea: there is no implementation nor clear concept, so please feel free to close, but I hope it will be useful to brainstorm this amongst type-junkies.

I was considering adding types for `ReflectionMethod` and `ReflectionProperty`, but it seems to be tricky to do that, due to the lack of a system to reference properties and classes at type-level.

Specifically, parameter `$name` of `ReflectionClass<class-string>#getProperty($name)` could be a `class-property<class-string, non-empty-string, PropertyType of mixe>`, which could be used for type resolution.

Similarly, parameter `$name` of `ReflectionClass<class-string>#getMethod($name)` could be a `ReflectionMethod<class-method<class-string, non-empty-string, MethodType of mixed>>`, which could be used for type resolution, and `class-method<class-string, non-empty-string>` could be a `callable`.

This sort of API is interesting for things like hydrators:

```php
/**
 * @psalm-template TObject of object
 * 
 * NOTE: how do we type `mixed` below here?
 *
 * @psalm-param non-empty-array<
 *   class-property<class-string<TObject>, non-empty-string>,
 *   mixed,
 *   mixed
 * >
 * @psalm-return TObject
 */
function hydrate(array $input, object $object) { /** ... */ }
$object = hydrate(
    [
        'foo' => $bar,
        'baz' => $tab,
    ],
    $object
);
```

Important in the above is that I still don't have an idea of how we could type the `mixed`

Also for assertions:

```php
/**
 * @psalm-assert TObject of object
 * @psalm-param non-empty-array<
 *   class-property<class-string<TObject>, non-empty-string, mixed>,
 *   mixed
 * >
 * @psalm-assert class-method<class-string<TObject>, non-empty-string> $method
 */
function assertMethodExists(string $method, object $object) { /** ... */ }

function findObject() : object {}
function findMethod() : string {}

$object = findObject();
$method = findMethod();

assertMethodExists($object, $method);

$object->$method(); // not sure about parameter count and parameter types
```

Patrick-Remy:
For `ActiveRecord`-patterns in php, there is currently no possibility to type methods like `asArray()` . E.g. for the Yii framework:
```php
/**
 * @property int $a
 * @property bool $b
 */
class Foo extends \yii\db\ActiveRecord {
}

// this will have type `array{a: string, b: string}[]`, as db will return all values as strings in Yii
$foo = Foo::find()->asArray()->all();
```

This PR adds `properties-of<T>` type that currently returns all class property names. There are the variants `public-properties-of`, `protected-properties-of` and `private-properties-of`.

I also fixed an imho bug, that the original class name wasn't proper downpassed in `ReturnTypeAnalyzer` so that `static` wasn't properly resolved in `TypeExpander`.

I aim in a second iteration to return an array-shape form of the class properties, instead of only the names, so that for the above example`properties-of<Foo>` equals `array{a: int}` instead of currently only `'a'`. So that `key-of<class-properties<Foo>>` will pick all property names. Unfortunately currently `key-of` only supports constants.

## TODO
- [x] #7366 
- [x] #7396
- [x] Return associative-array instead of property names via `properties-of`
- [x] Template support for `properties-of`

## SKIPPED
- [ ] `T[k]` usable for arrays, e.g. `(array{a: string})[a] = string` 
- [ ] Add type `TLiteralProperty` for further checks like `property_exists`

Please note, that this is my very first contribution to psalm, so please be aware of any inconveniences. ;)

Partially fixes  #3141


orklah:
Firstly, congrats! This is a lot of work and it's not the easiest way to begin in Psalm!

Secondly, I have some remarks and checks to add:

- Here is a small snippet with a lot of things to check and where we should have answers: https://psalm.dev/r/4971520a7d
- I think it could make sense to have a TLiteralPropertyOf, the same way we have TLiteralClassString for example. This could be used for two things: "Transforming" literal strings into checked properties, for example, after a property_exists check and also resolving literal properties when they are few enough. For example on this: https://psalm.dev/r/9e2eb788b2
- I wonder if we could make TPropertiesOf a child of TNonFalsyString, that way it will behave like a string in most cases
- I think it will be easier to add this into Psalm 5 (so in master). I already see two places in your PR that will create conflicts with changes we made
- Could you submit your fix in another PR (4.x this time)? It should make you a known contributor and you won't need us anymore to run the CI

src/Psalm/Internal/Type/TypeExpander.php
```diff
@@ -299,6 +302,56 @@ function ($constant_name) use ($const_name_part): bool {
             return $return_type;
         }
 
+        if ($return_type instanceof TPropertiesOf) {
+            if ($return_type->fq_classlike_name === 'self' && $self_class) {
```
muglug, line 306: Take the opportunity to break this section out into its own method, which will reduce the overall method complexity

src/Psalm/Internal/Type/TypeExpander.php
```diff
@@ -299,6 +302,56 @@ function ($constant_name) use ($const_name_part): bool {
...
+                $return_type->fq_classlike_name = is_string($static_class_type) ? $static_class_type : $self_class;
+            }
+
+            if (!$codebase->classExists($return_type->fq_classlike_name)) {
+                return $return_type;
+            }
+
+            // Get and merge all properties from parent classes
+            $class_storage = $codebase->classlike_storage_provider->get($return_type->fq_classlike_name);
+            $properties_types = $class_storage->properties;
```
muglug, line 320: You might want to refine these to just instance properties? That might be closer to what you're looking for, since PHP tends to treat static properties pretty differently.

Patrick-Remy: Good catch!

Patrick-Remy:
@orklah I'm done for now and this should be ready for review. I reworked this PR but I think it's clear if you have a look at the tests.

I will initially skip `TLiteralProperty` as it's a bit more work to do, but I will raise another PR for it, once this is merged.

docs/annotating_code/type_syntax/utility_types.md
```diff
@@ -0,0 +1,40 @@
+# Utility types
```
AndrolGenhald, line 1: Where are `key-of` and `value-of` documented? Maybe they should be added here?

Patrick-Remy: I can add this as well in another PR.

Patrick-Remy:
Template/Generic support was easier than I thought! It's done and documented :)

docs/annotating_code/type_syntax/utility_types.md
Patrick-Remy: I added `key-of` and `value-of`, `SomeClass::CONST_*` also could need some documentation, but I think it's not a utility type if you look at typescripts definition (https://www.typescriptlang.org/docs/handbook/utility-types.html)

orklah:
Seems fine to me :)

@ondrejmirtes FYI, this PR introduces a few types in Psalm, namely `properties-of<T>` and variants like `private-properties-of<T>`. Also it extends `key-of<ClassName::CONSTNAME>` and to work with template `value-of<ClassName::CONSTNAME>` to work with templates too

orklah (approved):

Patrick-Remy:
Template Support for `key-of` and `value-of` was introduced in #7359 
I only documented those changes as requested.

orklah:
You linked this PR :)

Patrick-Remy:
@AndrolGenhald do you have some time for review? I'd love to see this in a next alpha of 5.0 🙂

orklah:
Thanks! Sorry, for the delay ;)

aurimasniekis:
@Patrick-Remy thanks for implementing this feature, but I am having some trouble trying to use it. I have tried many different combinations but I really can't seem to find a way to make it work.

P.S. I am trying on `dev-master` branch

Few examples:

```php
<?php

class Test
{
    /**
     * @template T of object
     *
     * @param T                        $object
     * @param key-of<properties-of<T>> $name
     *
     * @return void
     */
    public static function check(object $object, string $name): void
    {
    }
}

class Data
{
    public string $foo = 'foo!';
    public int    $bar = 42;
}

$data = new Data();

Test::check($data, 'aa');
```

```
ERROR: InvalidDocblock - a.php:74:15 - Untemplated key-of param properties-of<T> should be an array in docblock for Test::check (see https://psalm.dev/008)
     * @param key-of<properties-of<T>> $name
```


```php
<?php

class Test
{
    /**
     * @template T of object
     * @template P of properties-of<T>
     *
     * @param T         $object
     * @param key-of<P> $name
     *
     * @return void
     */
    public static function check(object $object, string $name): void
    {
    }
}

class Data
{
    public string $foo = 'foo!';
    public int    $bar = 42;
}

$data = new Data();

Test::check($data, 'foo');
Test::check($data, 'aa');
```

```
ERROR: MismatchingDocblockParamType - a.php:75:15 - Parameter $name has wrong type 'key-of<P>', should be 'string' (see https://psalm.dev/141)
     * @param key-of<P> $name


ERROR: InvalidArgument - a.php:92:20 - Argument 2 of Test::check expects key-of<P:fn-test::check as properties-of<T:fn-test::check as object>>, 'foo' provided (see https://psalm.dev/004)
Test::check($data, 'foo');


ERROR: InvalidArgument - a.php:93:20 - Argument 2 of Test::check expects key-of<P:fn-test::check as properties-of<T:fn-test::check as object>>, 'aa' provided (see https://psalm.dev/004)
Test::check($data, 'aa');
```

Also `key-of<T>` doesn't look like works and just says no issues:

```php
<?php

class Test
{
    /**
     * @template T of array
     *
     * @param T         $input
     * @param key-of<T> $name
     *
     * @return void
     */
    public static function check(array $input, string $name): void
    {
    }
}

$data = ['foo' => 'ddd', 'bar' => 42];

Test::check($data, 'foo');
Test::check($data, 'aa');
```

Tried even providing array shape but no luck.

```
/** @var array{foo: string, bar: int} $data */
```

Patrick-Remy:
Thanks for your examples, these scenarios aren't currently tested, I'll have a look at these and try to figure out what is going wrong there!

aurimasniekis:
No worries, as far as I have tried to debug, I think the main issues is that `key-of` doesn't try to resolve non template values, and for e.g. if value requires to pass it through resolver it doesn't do it.

PR #7363, opened 2022-01-10
PHP 8.1: Report missing typehints in overridden native methods

### LINKED ISSUE #7362: Missing mixed return typehint is not detected by psalm
See https://psalm.dev/r/96c9d21f43, this causes a fatal error on PHP 8.0+:

```
PHP Fatal error:  Declaration of b::json() must be compatible with a::json(): mixed in php shell code on line 8
```

danog:
Allows detection of ReturnTypeWIllChange warnings, also fixes #7362

### DIFF
--- a/src/Psalm/Internal/Analyzer/MethodComparator.php
+++ b/src/Psalm/Internal/Analyzer/MethodComparator.php
@@ -124,7 +124,9 @@ class MethodComparator
             && !$implementer_method_storage->signature_return_type
             && !array_filter(
                 $implementer_method_storage->attributes,
-                fn (AttributeStorage $s) => $s->fq_class_name === 'ReturnTypeWillChange'
+                function (AttributeStorage $s) {
+                    return $s->fq_class_name === 'ReturnTypeWillChange';
+                }
             )
         ) {
             IssueBuffer::maybeAdd(
--- a/src/Psalm/Internal/Analyzer/MethodComparator.php
+++ b/src/Psalm/Internal/Analyzer/MethodComparator.php
@@ -118,6 +118,7 @@ class MethodComparator
 
         if (!$guide_classlike_storage->user_defined
             && $implementer_classlike_storage->user_defined
+            && $codebase->analysis_php_version_id >= 80100
             && ($guide_method_storage->return_type
                 || $guide_method_storage->signature_return_type
             )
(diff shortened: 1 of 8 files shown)


weirdan:
@danog, is there any reason not to add this to 4.x line?

danog:
Not really, no, I just wanted to switch directly to psalm v5 :D

danog:
Hmm, that's weird, I rebased on top of 4.x...

orklah:
There must have been something wrong. isEmpty was only removed in master

danog:
All fixed!

src/Psalm/Internal/Analyzer/MethodComparator.php
```diff
@@ -113,6 +115,29 @@ public static function compare(
             );
         }
 
+        if (!$guide_classlike_storage->user_defined
+            && $implementer_classlike_storage->user_defined
+            && $codebase->analysis_php_version_id >= 80100
```
orklah, line 120: Using 8_01_00 didn't work?

orklah:
Is this related to https://github.com/vimeo/psalm/issues/6395#issuecomment-952049533 ?

src/Psalm/Internal/Analyzer/MethodComparator.php
danog: Turns out it's invalid syntax on PHP 7.1

danog:
Yep, it's related precisely to that.

src/Psalm/Internal/Analyzer/MethodComparator.php
orklah: right, forgot this was a thing and we only did the change on Psalm 5

danog:
>Matt's branch was all about adding types in signature to be able to make the difference I believe?
Yes, but not all native classes are declared as stubs: for example, the `JsonSerializable` class I used in the tests is populated at runtime using reflection.
I could've changed the reflection populator to populate `signature_return_type` instead of `return_type` (and fix existing stubs), but I decided it would be easier (safer?) to just create an explicit check that simply relies on the interaction of a native parent class and a user-defined child class.

Actually, do you think it should be made a separate issue, given that it signals a rather important PHP 8.1 BC semi-break?

orklah:
A different issue could be interesting indeed. About the signature_return_type vs return_type, did PHP add signatures types for every internal function/class/interface or did they do it only for certain cases?

danog:
> About the signature_return_type vs return_type, did PHP add signatures types for every internal function/class/interface or did they do it only for certain cases?

The documentation says *Most* non-final internal methods now require overriding methods to declare a compatible return type, so I just assumed all of them did.

orklah:
@danog is this ready?

danog:
Yep!

orklah:
Thanks!

PR #7366, opened 2022-01-10
fix: pass static class name to ReturnTypeAnalyzer

Patrick-Remy:
As already stated out in PR #7359 and found there:

In these lines the `$fq_classlike_name` is overriden if the return type is declared in a parent class:

https://github.com/vimeo/psalm/blob/549c02d661189029f93ae685db41194a0593d66f/src/Psalm/Internal/Analyzer/ClassAnalyzer.php#L1871-L1875

And then downpassed via `ReturnTypeAnalyzer::verifyReturnType` to as both `$self_class`- and `$static_class_type`-param.
https://github.com/vimeo/psalm/blob/549c02d661189029f93ae685db41194a0593d66f/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php#L424-L428

The bug is currently not really visible, as e.g. for `@return class-string<static>` correctly only `static::class` is allowed and therefor not expected.


Patrick-Remy:
I fixed that I unintentionally commented-out here
https://github.com/vimeo/psalm/pull/7366/files#diff-ec7956a405a9a30d502695776e76d48768f3bb19238e94eee0897b16856dc3acR12

orklah:
Thanks! That's great!

PR #7367, opened 2022-01-10
Fix crash in code action

tm1000:
This PR fixes the following bug introduced into codeAction as a result of the underlying cache being empty when creating new files while using the language server

```
[Error - 11:05:08 PM] Request textDocument/codeAction failed.
  Message: UnexpectedValueException: Expecting /project/myTestFile.php:27:531:-:closure to have storage in /project/myTestFile.php in /project/vendor/vimeo/psalm/src/Psalm/Codebase.php:650
```

It's actually almost the same as #7247 except I didn't add the try/catch to my own code!  🤕 

Additionally this adds a quick speedup to `codeAction` by adding a cache when`emitIssues` is called so that we don't have to run the `IssueBuffer` everytime `codeAction` is called (which in vscode is a lot because every time the mouse or cursor moves a codeAction is sent). Instead we just ask the LanguageServer to give us the previous issues returned from IssueBuffer through the use of the method `getCurrentIssues`


tm1000:
Ugh. This got really messed up ignore!

weirdan:
Please branch off the 4.x branch.

tm1000:
@weirdan did that in my newer pr

PR #7368, opened 2022-01-10
Fix closure to have storage bug in codeAction

tm1000:
This PR fixes the following bug introduced into codeAction as a result of the underlying cache being empty when creating new files while using the language server

```
[Error - 11:05:08 PM] Request textDocument/codeAction failed.
  Message: UnexpectedValueException: Expecting /project/myTestFile.php:27:531:-:closure to have storage in /project/myTestFile.php in /project/vendor/vimeo/psalm/src/Psalm/Codebase.php:650
```

It's actually almost the same as #7247 except I didn't add the try/catch to my own code!  🤕 

Additionally this adds a quick speedup to `codeAction` by adding a cache when`emitIssues` is called so that we don't have to run the `IssueBuffer` everytime `codeAction` is called (which in vscode is a lot because every time the mouse or cursor moves a codeAction is sent). Instead we just ask the LanguageServer to give us the previous issues returned from IssueBuffer through the use of the method `getCurrentIssues`

Sorry for the other PR. I didn't realize you started work on v5 and set the default branch to 4.x

### DIFF
--- a/src/Psalm/Internal/LanguageServer/LanguageServer.php
+++ b/src/Psalm/Internal/LanguageServer/LanguageServer.php
@@ -103,6 +103,11 @@ class LanguageServer extends Dispatcher
      */
     protected $onchange_paths_to_analyze = [];
 
+    /**
+     * @var array<string, list<IssueData>>
+     */
+    protected $current_issues = [];
+
     public function __construct(
         ProtocolReader $reader,
         ProtocolWriter $writer,
@@ -365,6 +370,7 @@ class LanguageServer extends Dispatcher
     public function emitIssues(array $uris): void
     {
         $data = IssueBuffer::clear();
+        $this->current_issues = $data;
 
         foreach ($uris as $file_path => $uri) {
             $diagnostics = array_map(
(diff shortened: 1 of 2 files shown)


tm1000:
test-with-real-projects is failing on the following cause:

```

  Problem 1
    - azjezz/psl is present at version 1.9.x-dev and cannot be modified by Composer
    - php-standard-library/psalm-plugin 2.0.0 conflicts with azjezz/psl <2.0.
    - Root composer.json requires php-standard-library/psalm-plugin ^2.0 -> satisfiable by php-standard-library/psalm-plugin[2.0.0].

You can also try re-running composer require with an explicit version constraint, e.g. "composer require php-standard-library/psalm-plugin:*" to figure out if any version is installable, or "composer require php-standard-library/psalm-plugin:^2.1" if you know which you need.

Installation failed, reverting ./composer.json to its original content
```

So nothing to do with this commit.

muglug (approved):

orklah:
Thanks!

PythooonUser:
Thanks for fixing this :)

PR #7371, opened 2022-01-11
Make links more clear

danog:


weirdan:
https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/ - that section indeed could use a bit more contrast, so I see how links might not stand out. With that said, I'm not too fond of the solution proposed here either.

AndrolGenhald:
> https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/ - that section indeed could use a bit more contrast, so I see how links might not stand out. With that said, I'm not too fond of the solution proposed here either.

Oh, yeah, I probably should have checked the website instead of checking markdown renderers... I can see how that's not super clear. I think the ideal solution would be a css change on the website.

danog:
> I can see how that's not super clear, those links are nearly indistinguishable from the surrounding text. I think the ideal solution would be a css change on the website.

Precisely, I briefly considered doing this, instead, but I personally like the muted color scheme in the site, and (at least to me) links are pretty distinguishable in a normal text-context like in https://psalm.dev/docs/, it's just that this specific page is composed of just links :)

I also considered changing the typehint to a textual description, forcing people to click on it to see the actual typehint (ie `positive-int` => `Positive integer types`)

danog:
@weirdan @orklah Any opinion on this? :)

orklah:
Unfortunately, I have horrible taste in things so my opinion is unhelpful here 😄

weirdan:
I think as long as there's some normal text visible in the view red links are ok. So what we could do instead is to provide textual descriptions here and there (they don't have to be present on every item).

E.g. something like this:

AndrolGenhald:
@weirdan I would prefer a color change, I have trouble telling links apart in the rest of the documentation as well.

weirdan:
@AndrolGenhald please propose the colour to change to then

AndrolGenhald:
@weirdan I'm not exactly the best person to ask about what colors are good for a theme :stuck_out_tongue: all I can tell you is that these colors are hard to tell apart. The standard blue is fine, but @danog likes the muted color scheme without blue links, and I'm guessing others might as well.

How about underlining?
image

Obviously it should exclude the edit icon and navigation, but that shouldn't be difficult. In basic no-css html links are blue and underlined, so if they're going to be different it makes sense to me to at least keep them either blue without underlining or non-blue with underlining.

While we're at it maybe we can add a [dark theme based on system preference](https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#system-preference).

weirdan:
Dotted underline perhaps?

image

AndrolGenhald:
The homepage already has normal underlining:
image
So I'd go with that. We should maybe make those colors consistent as well. I might open a pull request or two later this week.

PR #7372, opened 2022-01-11
Add LiteralKeyUnshapedArray checks

danog:
This has been really useful @ work to enforce shaping of arrays.
This will require some more tweaking to fix tests & psalm itself, in the meantime sending in a draft for review.

Phpstan already does something similar, by directly requiring an array shape in all array typehints (which is also what we've been doing, via a CS rule): https://phpstan.org/r/e02b117c-9c25-414e-ac51-949eb6ee70c8

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -290,6 +290,7 @@
             <xs:element name="LessSpecificClassConstantType" type="ClassConstantIssueHandlerType" minOccurs="0" />
             <xs:element name="LessSpecificImplementedReturnType" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="LessSpecificReturnStatement" type="IssueHandlerType" minOccurs="0" />
+            <xs:element name="LiteralKeyUnshapedArray" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="LessSpecificReturnType" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="LoopInvalidation" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="MethodSignatureMismatch" type="IssueHandlerType" minOccurs="0" />
--- a/docs/running_psalm/error_levels.md
+++ b/docs/running_psalm/error_levels.md
@@ -130,6 +130,7 @@ These issues are treated as errors at level 2 and below.
  - [InvalidDocblockParamName](issues/InvalidDocblockParamName.md)
  - [InvalidFalsableReturnType](issues/InvalidFalsableReturnType.md)
  - [InvalidStringClass](issues/InvalidStringClass.md)
+ - [LiteralKeyUnshapedArray](issues/LiteralKeyUnshapedArray.md)
  - [MissingClosureParamType](issues/MissingClosureParamType.md)
  - [MissingClosureReturnType](issues/MissingClosureReturnType.md)
  - [MissingConstructor](issues/MissingConstructor.md)
--- a/docs/running_psalm/issues.md
+++ b/docs/running_psalm/issues.md
@@ -100,6 +100,7 @@
  - [LessSpecificImplementedReturnType](issues/LessSpecificImplementedReturnType.md)
  - [LessSpecificReturnStatement](issues/LessSpecificReturnStatement.md)
  - [LessSpecificReturnType](issues/LessSpecificReturnType.md)
+ - [LiteralKeyUnshapedArray](issues/LiteralKeyUnshapedArray.md)
  - [LoopInvalidation](issues/LoopInvalidation.md)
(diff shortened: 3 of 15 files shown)


orklah:
Can you describe what that does exactly? This seems quite different from what PHPStan does. Also, this would be *very* opiniated (quite possibly even more than Mixed* issues)

I think this should target Psalm 5 in any case, this would be too big a change for projects just updating their version

danog:
As per the documentation, LiteralKeyUnshapedArray is emitted when a literal key is used on an unshaped array.  

All known array shapes must be annotated using [array shape syntax](/docs/annotating_code/type_syntax/array_types/#object-like-arrays), to allow additional Psalm paradox checks on array keys and types.  

When working with arrays whose shape is not known yet (i.e. user input), a **separate validation function** must be used with a function-level suppress.  

By splitting validation and business logic, Psalm can make more and better typechecks of existing usages of the business logic.

This is indeed very opinionated, and I indeed was unsure whether to put the issue in level 1 along with the `Mixed*` issues, but it has really helped @ work in enforcing more and better phpdocs.

AndrolGenhald:
Is this a bit redundant with [ensureArray*OffsetsExist](https://psalm.dev/docs/running_psalm/configuration/#ensurearraystringoffsetsexist)? I know that config doesn't do as much as I'd like, could that be improved instead? Or maybe only emit the new issue if the relevant setting between those is set?

orklah:
> a PossiblyUndefined*Offset indicates a very local issue, that is solved easily with an isset.

The part were you say it has value is also the part that will annoy people. If a given array can't be documented for whatever reason (including array retrieved from third party libs), users will have no way out. Isset is precisely the usual good practice in that kind of cases.

So, I'd be open in extending ensureArray*OffsetsExist with that check(without the isset part for example), but if you want to keep the whole feature as is, I think this would belong more in a plugin.

I'm kinda undecided though, maybe @weirdan can give us his thoughts

AndrolGenhald:
I don't really mind having this be an optional feature if you want to be really strict about arrays, but the one thing that really makes me uneasy about it is the requirement to use `@psalm-suppress`. I've always viewed suppressing errors as something that should be done if the error is a false-positive or if you just don't have things correctly typed. This feature seems to me like it's changing that meaning to "I want to allow this in this function but nowhere else", which is pretty similar but still seems a little strange to me.

The way I think about `@psalm-suppress` right now, the ideal scenario is not having any instances of it in the codebase, and searching for it can show you all the places things need fixed or psalm is wrong. With this feature that's no longer the case. I think I'd prefer a new tag be introduced.

I definitely think this should go behind a configuration option that defaults to false, and that should be mentioned in the documentation for the issue. It's too opinionated to be the default.

I agree with @orklah that this seems like it might be more suited to a plugin.

danog:
> I don't really mind having this be an optional feature if you want to be really strict about arrays, but the one thing that really makes me uneasy about it is the requirement to use `@psalm-suppress`

I've been thinking about the exact same thing over the last few days, and I intend to switch to another way of handling this issue by introducing a `@psalm-validate` method-level annotation, instead (which possibly also automatically enables `@psalm-pure`).

Again, I know it's really opinionated, but now that I think of it it's not so different from a MixedAssignment issue, which is emitted by default at level 1: it has the same underlying rationale of enforcing typing @ the source.
I'll move it to a level 1 issue and point to master, still hoping it can get in as a default in v5 :D

weirdan:
I prefer something closer to what PHPStan does: forbidding unqualified `array` and other iterables, either as native type (missing the docblock) or `@param array`.

orklah:
I also like this check, but it wouldn't help that much in this case. The goal of this PR is to scream when accessing any literal offset key when the array is not a shape. So even with `array<string, string>`, the phpstan-like check would be happy but not this PR.

Personnally I'm wondering if we should put that in a plugin or into the ensureArray*OffsetsExist check

weirdan:
> or into the ensureArray*OffsetsExist check

That would be my preference, yes.

orklah:
@danog ok so we could add this to the existing check. That implies that isset() will still work, this solves my second worry. Could you try to make the adaptations?

AndrolGenhald:
@orklah I thought the whole point of this PR is to prevent even `isset($arr['possiblyunset'])` outside of a specifically tagged function, which is why I think it's too opinionated.

orklah:
If I understood correctly, the goal was not specifically to forbid `isset($arr['possiblyunset'])` but `$arr['possiblyunset']`, including in isset. What I suggest is making ensureArray*OffsetsExist stricter so it would warn on `$arr['possiblyunset']` but still allows `isset($arr['possiblyunset'])` (which still would allow a way out when the phpdoc is not detailed enough)

PR #7374, opened 2022-01-11
Fix @covers annotation.

AndrolGenhald:
Per the [documentation](https://phpunit.readthedocs.io/en/9.5/annotations.html#covers), `@covers` requires a FQCN.


weirdan:
Thanks!

PR #7377, opened 2022-01-11
Add vscode settings to project

tm1000:
This adds recommended VS Code extensions (namely the supported Psalm one https://github.com/psalm/psalm-vscode-plugin) and additionally sets up settings.json for vscode to read. This gives the ability to use Psalm to analyze itself while in vscode which is helpful (note: the extension recommendations can be ignored so this will not force anything on anybody)

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,6 @@
+{
+    "recommendations": [
+        "getpsalm.psalm-vscode-plugin",
+        "obliviousharmony.vscode-php-codesniffer"
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,5 +3,6 @@
   "psalm.psalmScriptPath": "psalm-language-server",
   "phpCodeSniffer.executable": "vendor/bin/phpcs",
   "phpCodeSniffer.standard": "Custom",
-  "phpCodeSniffer.standardCustom": "phpcs.xml"
+  "phpCodeSniffer.standardCustom": "phpcs.xml",
+  "editor.rulers": [120]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "psalm.configPaths": ["psalm.xml.dist"],
+  "psalm.psalmScriptPath": "psalm-language-server",
+  "phpCodeSniffer.executable": "vendor/bin/phpcs",
+  "phpCodeSniffer.standard": "Custom",
+  "phpCodeSniffer.standardCustom": "phpcs.xml"
+}


AndrolGenhald:
Mind adding `"editor.rulers": [120]` as well? I have that in mine locally because I have a wider default but Psalm has a CS rule for 120.

tm1000:
@AndrolGenhald added!

tm1000:
The three tests that failed would have to be unrelated to this PR

orklah:
I'm not sure about that. I don't think there is precedent for adding files for a specific IDE.

Is it common practice among OSS?

tm1000:
I've seen it in other projects and it only helps. They can be ignored anyways by just not installing the suggested extensions

tm1000:
Additionally I want to emphasize. The settings this would enforce are already strictly enforced by this projects CI policies. It only helps people using vscode to see these errors in their IDE before they run the lint commands manually or before they push up

Really this project needs something like grumpphp that can run cs and psalm as a pre commit hook. It will recent many PRs that fail because people don't run psalm or cs.

orklah:
> It will recent many PRs that fail because people don't run psalm or cs.

Well, I'm the first to use GH CI to test my changes. It just runs so much faster than my own machine that I rarely bother running tests/cs/psalm on my end.

tm1000:
Ah yeah if you draft it the actions will still run

I do think this is useful but in my case I can also just gitignore this folder (or would a pr to ignore that folder be good) so I can still utilize it in my ide.

Patrick-Remy:
At least for me, I copied your files, to use while development of my PRs. It was sometimes a bit tricky, if Psalm restarted while adjusted code was broken, and therefor lead to non-existent errors, but if you keep that in eye, the autocompletion and linting was much more helpful.

tm1000:
@Patrick-Remy I think it's very helpful but I also agree with @orklah in that I see no other projects including something like this (not only for vscode). In my other PRs for Psalm I have just adjusted gitignore to ignore that directory so I dont commit it.

Patrick-Remy:
Not everybody, but some:
- https://github.com/microsoft/TypeScript/tree/main/.vscode
- https://github.com/grafana/grafana/tree/main/.vscode
- https://github.com/vuejs/pinia/tree/v2/.vscode
- https://github.com/vercel/next.js/tree/canary/.vscode

orklah:
Well, makes sense Typescript added it given it's developped by the same company that made the IDE 😄 

Thanks for the contribution but I won't be adding this. While there is value in it, it should not take too long to reconfigure if needed. I wish standard existed for that kind of thing (.editorconfig anyone?), even if the standard is only implemented in a single IDE, at least it doesn't feel like adding special for that IDE.

This would create a precedent for Psalm to allow all kinds of file for other IDE (phpstorm also have lot of those...) making more things to maintain than necessary

tm1000:
All good @orklah !

PR #7379, opened 2022-01-12
Ability to hide warnings in Language Server

tm1000:
As it is currently warnings (eg if you have level set to 4 then levels 1-3 will be warnings in your IDE) are shown in IDEs utilizing language server. Users have requested the ability to disable that (https://github.com/psalm/psalm-vscode-plugin/issues/16)

This PR adds the ability to hideWarnings but it also follows the Language Server Protocol to ask the client what the current configuration is. This allows this setting to be changed without having to restart Psalm Language Server (which you have to do for all other settings because they are CLI flags). https://microsoft.github.io/language-server-protocol/specification#workspace_configuration

### DIFF
--- a/src/Psalm/Internal/LanguageServer/Client/Workspace.php
+++ b/src/Psalm/Internal/LanguageServer/Client/Workspace.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Internal\LanguageServer\Client;
+
+use Amp\Promise;
+use JsonMapper;
+use Psalm\Internal\LanguageServer\ClientHandler;
+
+/**
+ * Provides method handlers for all textDocument/* methods
+ */
+class Workspace
+{
+    /**
+     * @var ClientHandler
+     */
+    private $handler;
+
+    /**
+     * @var JsonMapper
+     * @psalm-suppress UnusedProperty
+     */
+    private $mapper;
+
+    public function __construct(ClientHandler $handler, JsonMapper $mapper)
+    {
+        $this->handler = $handler;
+        $this->mapper = $mapper;
+    }
+
+    /**
+     * The workspace/configuration request is sent from the server to the client to
+     * fetch configuration settings from the client. The request can fetch several
+     * configuration settings in one roundtrip. The order of the returned configuration
+     * settings correspond to the order of the passed ConfigurationItems (e.g. the first
+     * item in the response is the result for the first configuration item in the params).
+     *
+     * @param string $section The configuration section asked for.
+     * @param string|null $scopeUri The scope to get the configuration section for.
+     */
+    public function requestConfiguration(string $section, ?string $scopeUri = null): Promise
+    {
+        /** @var Promise<object> */
+        return $this->handler->request('workspace/configuration', [
+            'items' => [
+                [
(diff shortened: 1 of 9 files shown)


tm1000:
```
Using version ^2.0 for php-standard-library/psalm-plugin
Running composer update php-standard-library/psalm-plugin
Updating dependencies

  Problem 1
    - azjezz/psl is present at version 1.9.x-dev and cannot be modified by Composer
    - php-standard-library/psalm-plugin 2.0.0 conflicts with azjezz/psl <2.0.
    - Root composer.json requires php-standard-library/psalm-plugin ^2.0 -> satisfiable by php-standard-library/psalm-plugin[2.0.0].

You can also try re-running composer require with an explicit version constraint, e.g. "composer require php-standard-library/psalm-plugin:*" to figure out if any version is installable, or "composer require php-standard-library/psalm-plugin:^2.1" if you know which you need.

Installation failed, reverting ./composer.json to its original content.
```

orklah:
The error on CI is fixed.

@weirdan I think this is a BC break. Do you agree?

@tm1000 could this be added on Psalm 5 or is it important for every version?

tm1000:
Yeah @orklah v5 is fine. Should I scope it to that?

orklah:
yeah, if version is not an issue, please target master branch

tm1000:
@orklah is there an issue for what the roadmap of v5 is?

orklah:
There's a discussion here: https://github.com/vimeo/psalm/discussions/7171 but there is no clear roadmap defined.

tm1000:
@orklah ok thanks. That actually helps a lot. 

I think I can slip in a few more features into v5 or maybe at least upgrade amp to whatever version supported 7.4 if it's not already up to date here

orklah:
I'm putting this in draft for now while you change target

weirdan:
> I think this is a BC break

Doesn't look like one to me. From what I understand, older clients would ignore the configuration request (or return an error). What makes you think it's a break?

orklah:
It changes the params in a public function here: https://github.com/vimeo/psalm/pull/7379/files#diff-2102b4f3a518a68952841a4b0e556c2d104fdf35072a295ec2738dad61d0cc0dR225

weirdan:
The class is internal though, so we can change it any way we want.

tm1000:
There aren't any breaking changes but if I apply this to v5 then I can just move more configuration settings into this for IDEs so that we don't have to restart psalm every time. So I'm fine with shooting for more work on the language server in v5

orklah:
if there's no BC break, I'm fine with both :)

tm1000:
Should new features be in v5 moving forward? I'm quite fascinated by what Matt was able to achieve using php and creating a language server out of Psalm and I think there's a lot of way to go as long as you both keep approving my PRs ;) (additionally Matt/Weirden gave me admin on the psalm vscode plug-in as well)

weirdan:
> Should new features be in v5 moving forward?

For now, features can target either. Once v5 is released we will close v4 for new features.

tm1000:
See: https://github.com/tm1000/psalm/tree/feature/upgrade-lsp

Which will make it's way in here once I get unit tests added

tm1000:
Superseded by #8960

PR #7383, opened 2022-01-12
Add test for issues.md to ensure all documented issues are listed.

AndrolGenhald:
We have tests to make sure issues are documented, but I've noticed that when new issues are added, people (myself included) often forget to add them to `issues.md` so they'll show up in the [issues list](https://psalm.dev/docs/running_psalm/issues/).

This should also catch problems like #7373 before they get merged.


docs/running_psalm/issues/ComplexFunction.md
```diff
@@ -0,0 +1,3 @@
+# ComplexFunction
+
+Emitted when a function is too complicated for Psalm to analyze. Consider splitting it up to allow Psalm to analyze it.
```
AndrolGenhald, line 3: This could probably be expanded on, but I don't know enough about the conditions to in which this is emitted to document it correctly. If `Config::$max_graph_size` and `Config::$max_avg_path_length` ever become configurable that would also be good to mention here.

docs/running_psalm/issues/ComplexFunction.md
```diff
@@ -0,0 +1,3 @@
+# ComplexFunction
+
+Emitted when a function is too complicated. Complicated functions should be split up.
```
AndrolGenhald, line 3: This could probably be expanded on, but I don't know enough about the conditions to in which this is emitted to document it correctly. If `Config::$max_graph_size` and `Config::$max_avg_path_length` ever become configurable that would also be good to mention here.

weirdan:
Do you mind adding a script (in `bin/`) to generate the `issues.md` file?

AndrolGenhald:
@weirdan I'm not sure how useful it will be given how infrequently it changes and how simple the updates are, but it's easy enough, so why not.

AndrolGenhald:
Interesting that `composer cs` didn't catch that locally, I'm guessing it's because it's in `bin/`.

weirdan (approved):

weirdan:
Thanks!

PR #7384, opened 2022-01-12
Catch additional InvalidArgumentException

tm1000:
This fixes:

```
[Error - 8:54:14 PM] Request textDocument/codeAction failed.
  Message: InvalidArgumentException: Could not get class storage for app\http\controllers\provisionserver in /usr/src/app/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:46
```

So that the language server does not crash


orklah:
Thanks!

PR #7385, opened 2022-01-13
Add support of template for docblock methods

### LINKED ISSUE #7320: Template return type is not resolved on docblock @method
https://psalm.dev/r/ddebf2928d

Psalm seems to ignore the template on the return type on a docblock method. Same issue occurs using class and interface.

vincent4vx:
See fix #7320

Add the support of template type on return type and parameter of a method declared using `@method` annotation, on class or interface : 

```php
<?php
/**
 * @template T
 * @method T get()
 * @method void set(T $value)
 */
class Container
{
    public function __call(string $name, array $args) {}
}

class A {}
function foo(A $a): void {}

/** @var Container<A> $container */
$container = new Container();
$container->set(new A());
foo($container->get());
```

Current version do not process template types, so calling those pseudo methods will always result to InvalidArgument error.


orklah:
Thanks!

PR #7386, opened 2022-01-13
Ensure that all entries in test arrays have explicit keys

muglug:
This PR serves two purposes:
 - array keys in tests add a bit of extra clarity to what things are for
 - these arrays can be trivially exported for use in other projects

Obviously the second point doesn't benefit Psalm in any way, and if you feel the code change is an annoyance feel free to disregard this.


tests/AnnotationTest.php
```diff
@@ -737,27 +737,27 @@ function foo() : array {
...
                     /**
                      * @return [bad]
                      */
                     function fooBar() {
                     }',
-                [],
-                [
+                'assertions' => [],
+                'ignored_issues' => [
                     'InvalidDocblock' => Config::REPORT_INFO,
```
orklah, line 755: What does having a report option does here?

tests/BinaryOperationTest.php
```diff
@@ -790,32 +848,23 @@ function example(object $foo): string
...
                 'error_message' => 'InvalidOperand',
             ],
-            'differingNumericTypesAdditionInStrictMode' => [
-                '<?php
-                    $a = 5 + 4.1;',
-                'error_message' => 'InvalidOperand',
-                'error_levels' => [],
-                'strict_mode' => true,
-            ],
-            'concatenationWithNumberInStrictMode' => [
```
orklah, line 810: This test was dropped?

orklah (approved):

tests/BinaryOperationTest.php
muglug: No, it was moved up into a custom section to remove the `strict_mode` argument from the trait signature

tests/AnnotationTest.php
muglug: From what I recall it was an edge-case where this slight suppression was necessary to trigger a bug

Issue #4295, opened 2020-10-07
False positive on Exception::getCode


pattisahusiwa:
Only `PDOException` can return a `string` error code, everything else will return `int`.

https://www.php.net/manual/en/class.pdoexception.php
```php
PDOException extends RuntimeException {
    /* Properties */
    public array errorInfo;
    protected string code;
    
    /* Inherited properties */
    protected string message;
    protected int code;
```

https://www.php.net/manual/en/class.exception.php
```php
Exception implements Throwable {
    /* Properties */
    protected string message ;
    protected int code ;
```

**Code to reproduce**
https://psalm.dev/r/e785986fc0

**Expected result**
No Error. Psalm should distinguish `PDOException` from others.

Related commit: 53cd878694d680fb32be4a8d5be35c8eb21c2c64

**Note:**
How to solve this conflict?
https://psalm.dev/r/b44df0b0e1

Another side effect:
https://3v4l.org/Ya32I

morozov:
Another relevant example that uses `Throwable::getCode()`: https://psalm.dev/r/902effb53d.

muglug:
@morozov but `PDOException` is a `Throwable`, so that's not wrong

morozov:
My point is that Psalm incorrectly determines the return type of `Throwable::getCode()` which (unlike `Exception`) is [documented](https://www.php.net/manual/en/throwable.getcode.php#refsect1-throwable.getcode-description) as `int`. Although now I see that the documentation is ambiguous.

hrs-o:
I also think this is a bug.
Will there be a fix?

It is clearly stated that Exception::getCode will return int.
`final public Exception::getCode(): int`
https://www.php.net/manual/en/exception.getcode.php

https://psalm.dev/r/e9a0196ef3


--

```php
<?php

$e = new PDOException("a", 1);
var_dump($e->getCode());

/**
int(1)
 */
```


```php
<?php

$e = new PDOException("a", "a");
var_dump($e->getCode());

/**
Fatal error: Uncaught TypeError: Exception::__construct(): Argument #2 ($code) must be of type int, string given in /src/tmp/test.php:3
Stack trace:
#0 /src/tmp/test.php(3): Exception->__construct('a', 'a')
#1 {main}
thrown in /src/tmp/test.php on line 3
 */
```

weirdan:
> It is clearly stated that Exception::getCode will return int.

Or it's clearly stated that

> Return Values 
[...]

PR #7390, opened 2022-01-13
Add Exception->getCode() return type provider

VincentLanglet:
Close https://github.com/vimeo/psalm/issues/4295


src/Psalm/Internal/Provider/ReturnTypeProvider/ExceptionCodeReturnTypeProvider.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Psalm\Internal\Provider\ReturnTypeProvider;
+
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Type;
+
+class ExceptionCodeReturnTypeProvider implements \Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface
```
orklah, line 8: You must register your returnTypeProvider here: https://github.com/vimeo/psalm/blob/97e6511fab6622c1ca434c665dd895e77a0eeca2/src/Psalm/Internal/Provider/FunctionReturnTypeProvider.php#L111

src/Psalm/Internal/Provider/ReturnTypeProvider/ExceptionCodeReturnTypeProvider.php
```diff
@@ -0,0 +1,38 @@
...
+
+        if ($method_name_lowercase !== 'getcode') {
+            return null;
+        }
+
+        if ($fqcn === 'Exception' || $fqcn === 'Throwable') {
+            return null;
+        }
+
+        if (is_a($fqcn, PDOException::class, true)) {
```
orklah, line 32: Please check with the string instead. I'm not sure there are childs of PDOException out there, and I'd prefer avoid triggering the autoloader when it's not needed

VincentLanglet: What if someone extends the PDOException ?

orklah: Let's worry about that if it happens (though I'd be surprised it does), I had bad surprises before when trying to call the autoloader to perform static analysis (it crashed Psalm when the class did not exist)

If you still want to do it, I think you should be able to check it statically, by fetching the ClassLikeStorage and checking its parents recursively

VincentLanglet:
@orklah Seems like my return type provider is not used. Any idea why ?

src/Psalm/Internal/Provider/ReturnTypeProvider/ExceptionCodeReturnTypeProvider.php
AndrolGenhald: See also #7107, I would like to eventually get to the point where Psalm can analyze most extensions without them being loaded at runtime.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallReturnTypeFetcher.php
```diff
@@ -92,6 +95,16 @@ public static function fetch(
             }
         }
 
+        if ($premixin_method_id->method_name === 'getcode'
+            && $premixin_method_id->fq_class_name !== Exception::class
+            && in_array(Throwable::class, $class_storage->class_implements)) {
+            if ($premixin_method_id->fq_class_name === PDOException::class) {
```
AndrolGenhald, line 101: Please use string so psalm doesn't crash if the PDO extension isn't loaded.

orklah: `::class` doesn't call the autoloader so it shouldn't be an issue

orklah:
@weirdan This will be an important change. If you take a look at test with real projects, you'll see that phpunit for example has a lot of redundant casts now.

This is probably due to Psalm not being able to infer the exact type before, it pushed users into adding cast to please static analyzers, but now it will treat those as redundant.

I talked with @VincentLanglet about this and suggested adding this to Psalm 5, but he made a good point mentioning this PR is the equivalent of PHPStan behaviour right now and for users that like to use both, this difference makes things difficult. (they're forced to suppress everything on one analyzer or the other because behaviour is opposed)

What's your take on this? Can we release that on 4.x?

weirdan:
If we marked the int type generated here with the `from_calculation` flag, casts would be allowed but not required:
https://github.com/vimeo/psalm/blob/33e9210e1a5954ae29d7f18a95ded52edd9da0bf/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php#L71-L75

weirdan:
And then we could drop the flag in Psalm 5

orklah:
Good,idea :)

@VincentLanglet can you apply the changes?

VincentLanglet:
Done @orklah

weirdan (approved):

orklah:
Thanks!

VincentLanglet:
> Thanks!

You're welcome. Any plan for the next release ?

Issue #7198, opened 2021-12-21
Psalm doesn't infer a method return type from an interface docblock


fluffycondor:
https://psalm.dev/r/8eaa0f78f7

If I repeat the same annotation on the class itself, it works perfectly:
https://psalm.dev/r/ee2ed90bdb

PR #7391, opened 2022-01-14
Handle inherited docblock method

vincent4vx:
See: #7198

In current version, psalm will only search for pseudo methods on current class, so if a docblock method is declared on implemented interface, or parent class, it will be ignored, and infered return type or parameters will be ignored.

This PR will add search of the pseudo method and declaring class on ancestors (interfaces + parents) if not defined on the current class. Static and instance calls are handled.

```php
<?php
/**
 * @method int bar()
 */
class A {}

class B extends A {
    public function __call(string $method, array $args) {}
}

function consumeInt(int $i): void {}

$b = new B();

consumeInt($b->bar());
```

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php
@@ -5,6 +5,7 @@ namespace Psalm\Internal\Analyzer\Statements\Expression\Call\StaticMethod;
 use Exception;
 use PhpParser;
 use Psalm\CodeLocation;
+use Psalm\Codebase;
 use Psalm\Context;
 use Psalm\Internal\Analyzer\ClassLikeAnalyzer;
 use Psalm\Internal\Analyzer\ClassLikeNameOptions;
@@ -481,6 +482,12 @@ class AtomicStaticCallAnalyzer
 
         $config = $codebase->config;
 
+        $found_method_and_class_storage = self::findPseudoMethodAndClassStorages(
+            $codebase,
+            $class_storage,
+            $method_name_lc
+        );
+
         if (!$naive_method_exists
             || !MethodAnalyzer::isMethodVisible(
                 $method_id,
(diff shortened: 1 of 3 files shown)


orklah:
Was there a good reason to have the function findPseudoMethodAndClassStorages declared twice? Could we mutualize?

src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php
```diff
@@ -1002,4 +1009,41 @@ public static function handleNonObjectCall(
...
+    private static function findPseudoMethodAndClassStorages(
+        Codebase $codebase,
+        ClassLikeStorage $static_class_storage,
+        string $method_name_lc
+    ): ?array {
+        if ($pseudo_method_storage = $static_class_storage->pseudo_static_methods[$method_name_lc] ?? null) {
+            return [$pseudo_method_storage, $static_class_storage];
+        }
+
+        $ancestors = $static_class_storage->class_implements + $static_class_storage->parent_classes;
```
orklah, line 1034: I'm still not sure what parent_classes really contains, do you?

I'm also surprised because as far as I can tell, inheritance with parent classes worked (as opposed to interfaces): https://psalm.dev/r/be4d8b2941

So I'm not sure if the parent_classes part is needed

vincent4vx:
> Was there a good reason to have the function findPseudoMethodAndClassStorages declared twice? Could we mutualize?

The method declared into `AtomicStaticCallAnalyzer` reads method storage from `pseudo_static_methods`, unlike `MissingMethodCallHandler` which reads from `pseudo_methods`.

So it may be possible to mutualize, but I don't know if the readability will be better...

src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php
vincent4vx: I've to admit it, I don't know what parent_classes exactly contains 😅 
It seems to contains all parents hierarchy (cf: https://github.com/vimeo/psalm/blob/36d5a2a83c1d7b3e2aca1cfe9e9dcc916619e962/src/Psalm/Internal/Codebase/Reflection.php#L104)

It's really weird, because inheritance dot not works for static call. 
Removing `parent_classes` from `MissingMethodCallHandler::findPseudoMethodAndClassStorages()` will works visibly (all tests pass), but it's required for `AtomicStaticCallAnalyzer::findPseudoMethodAndClassStorages()` (tests fails without).

orklah:
Thanks!

PR #7394, opened 2022-01-14
[TypeProvider] Provide argument types during FunctionParamsProviderEvent

veewee:
Discussed in https://github.com/vimeo/psalm/discussions/7212

When adding following code in both a `FunctionParamsProviderInterface` and `FunctionReturnTypeProviderInterface` event handler.

```php
$nodeTypeProvider = $event->getStatementsSource()->getNodeTypeProvider();
$args = $event->getCallArgs();

$type = $nodeTypeProvider->getType($args[0]->value);
```

* The `getFunctionParams` hook always returns `null` on this call.
* The `getFunctionReturnType` hook always returns the detected type on this call. Something like `impure-Closure(string):int`

This PR mostly rearranges some code so that the `ArgumentsAnalyzer` runs  before the `FunctionParamsProviderInterface` hook is called.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
@@ -1291,6 +1291,7 @@ class ArgumentsAnalyzer
             || $arg->value instanceof PhpParser\Node\Expr\FuncCall
             || $arg->value instanceof PhpParser\Node\Expr\MethodCall
             || $arg->value instanceof PhpParser\Node\Expr\StaticCall
+            || $arg->value instanceof PhpParser\Node\Expr\ArrowFunction
             || $arg->value instanceof PhpParser\Node\Expr\New_
             || $arg->value instanceof PhpParser\Node\Expr\Cast
             || $arg->value instanceof PhpParser\Node\Expr\Assign
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
@@ -528,49 +528,56 @@ class FunctionCallAnalyzer extends CallAnalyzer
         $args = $stmt->isFirstClassCallable() ? [] : $stmt->getArgs();
 
         if ($function_call_info->function_exists) {
+            if (!$function_call_info->in_call_map || $function_call_info->is_stubbed) {
+                try {
+                    $function_call_info->function_storage = $function_storage = $codebase_functions->getStorage(
+                        $statements_analyzer,
+                        strtolower($function_call_info->function_id)
+                    );
+
+                    $function_call_info->function_params = $function_call_info->function_storage->params;
+
+                    if (!$function_storage->allow_named_arg_calls) {
+                        $function_call_info->allow_named_args = false;
+                    }
+
+                    if (!$is_predefined) {
+                        $function_call_info->defined_constants = $function_storage->defined_constants;
+                        $function_call_info->global_variables = $function_storage->global_variables;
+                    }
+                } catch (UnexpectedValueException $e) {
+                    $function_call_info->function_params = [
+                        new FunctionLikeParameter('args', false, null, null, null, false, false, true)
+                    ];
+                }
+            } else {
+                $function_callable = InternalCallMapHandler::getCallableFromCallMapById(
+                    $codebase,
+                    $function_call_info->function_id,
+                    $args,
+                    $statements_analyzer->node_data
+                );
+
+                $function_call_info->function_params = $function_callable->params;
+            }
+
             if ($codebase->functions->params_provider->has($function_call_info->function_id)) {
+                ArgumentsAnalyzer::analyze(
+                    $statements_analyzer,
+                    $stmt->getArgs(),
+                    $function_call_info->function_params,
+                    $function_call_info->function_id,
+                    $function_call_info->allow_named_args,
+                    $context
+                );
+
                 $function_call_info->function_params = $codebase->functions->params_provider->getFunctionParams(
                     $statements_analyzer,
                     $function_call_info->function_id,
                     $args,
-                    null,
+                    $context,
                     $code_location
                 );
-            }
-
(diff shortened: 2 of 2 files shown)


orklah:
Can you check why https://psalm.dev/r/ab6674485b  is detected? If you fixed a bug that was making args not being analyzed in existing functions, InvalidArgument should not be here. We must be missing something

veewee:
I'll investigate somewhere next week. 
My best guess is because the ArgumentsAnalyzer is ran again when trying to detect a return type.

orklah:
If it is, then it's running better the second time because it doesn't have false positive, so if we find the difference, we may find our solution

veewee:
@orklah : What I found for now is that the given code runs on the 4.X branch:

* `FunctionCallAnalyzer::analyze()`
* It calls to logic of `FunctionCallAnalyzer::getAnalyzeNamedExpression()` on [L129](https://github.com/vimeo/psalm/blob/de824d67242ba35d28b14473d4e23a474cc6cc04/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php#L129)
  * This is the place where this PR alters code.
  * The ArgumentsAnalyzer on [L500](https://github.com/vimeo/psalm/blob/de824d67242ba35d28b14473d4e23a474cc6cc04/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php#L500) is not being called here, since the function exists
  * It does not try to run a plugin, since the functionId has no registered plugin.
    *  Even if it would, it would not contain
    *  analyzed arguments
  * The `function_params` are injected by the storage on [L546](https://github.com/vimeo/psalm/blob/de824d67242ba35d28b14473d4e23a474cc6cc04/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php#L544)
  * This function returns FunctionCallInfo with the function params
* Next, the arguments analyzer is ran back in the regular `analyze` funciton on line [L175](https://github.com/vimeo/psalm/blob/de824d67242ba35d28b14473d4e23a474cc6cc04/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php#L175) 
  * The detected `function_params` are being passed to the ArgumentsAnalyzer on [L177](https://github.com/vimeo/psalm/blob/de824d67242ba35d28b14473d4e23a474cc6cc04/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php#L175-L183)


So in this case, the `ArgumentsAnalyzer` is ran in a different place (L175) and it contains the `function_params`. The other `ArgumentsAnalyzer` call on L502 is never really triggered and does not contain any `function_params`.

Not providing any `function_params`, to the `ArgumentsAnalyzer` makes it run very different analyzing code.
See [L122](https://github.com/vimeo/psalm/blob/de824d67242ba35d28b14473d4e23a474cc6cc04/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php#L122-L132). Down in the `evaluateArbitraryParam()` method, I've added support for the  additional `ArrowFunction` expression.

So to summarize in short: The ArgumentsAnalyzer is currently never being triggered before the `FunctionParamsProviderInterface` is being triggered. 

I'm feeling a bit too noob in this codebase to make the right decisions at this point. Any help is appreciated :)

orklah:
Ok, so there's two things that troubles me and both can be solved in one go:
- when the function exists, the arguments are analyzed later
- when the function name is an expression(https://github.com/vimeo/psalm/blob/b9d8dd9d3f01ba145582f77b7eecb65d07b280c2/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php#L128), we go through `getAnalyzeNamedExpression` instead of `handleNamedFunction` and only the latter calls the plugins (so something like `$a = 'foo'; $a($arg);` won't ever call the plugins

I think both of these issues could be solved my moving this condition: https://github.com/vimeo/psalm/blob/b9d8dd9d3f01ba145582f77b7eecb65d07b280c2/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php#L526

to here: https://github.com/vimeo/psalm/blob/b9d8dd9d3f01ba145582f77b7eecb65d07b280c2/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php#L189

That way, the plugin will be always called and it should have arguments analyzed.

Can you try that?

orklah:
Thinking about it a little bit more, I think we have a circular issue here...

Analyzing the args require knowing the params, so it's hard to know the args types when retrieving the params.

What this means is:
- either we can make `ArgumentsAnalyzer::analyze` stop depending on params maybe extract whatever uses params in there and put that after calling the plugins
- or we're stuck with current plugin hook and we would need an entire new set of hooks to be able to choose if we cant to be before or after arg populating

Just by curiosity, what was your goal in having args in order to provide params?

veewee:
The goal of the plugin is to finetune variadic params.
So for example: you have following function:

```php
$stages = pipe(
    fn (string $x): int => 2,
    fn (string $y): float => 1.2,
    fn (float $z): int => 23
);
$res = $stages('hello');
```

It passes 'hello' to function 1 and then pipes the results to the other function, resulting in 23.
As you can see, the type of short closure 2 is incorrect : it should be `int -> float`.

It would be handy to have the analyzed types there, because it makes it easier to follow the arguments of the provided closures and to calculate the parameters of the stages closures and the return type.

I am also planning to write a plugin for a `curry` function. Also in this case it would be handy, because the provided arguments are depending of the type of the first provided argument:

```curry('array_map', $someCallback)=> array -> array ```

In this case we need to know from the `array_map` callable what the arguments are so that we can check that argument 2 matches the once from array_map.

veewee:
@orklah This seems to work on master!

So what changed:

* It now first runs the function analysis as it would without a `FunctionParamsProviderInterface` plugin.
* Next it uses those resolved params as input for the `ArgumentsAnalyzer`
* Next it calls the `FunctionParamsProviderInterface::getFunctionParams()` methods
* This one can access the resolved argument types through the `NodeTypeProvider`, making it possible to alter them from within the plugin

orklah:
Seems legit! Thanks!

drupol:
Wooooooooooooooooh !

PR #7396, opened 2022-01-14
feat: make key-of/value-of usable with non-const arrays

Patrick-Remy:
Currently `key-of<T>` and `value-of<T>` helper-types are only usable with class constants. But they can be even more powerful! This PR tries to add the following capabilites:

1. Usage with `array`-like types (includes keyed array, and lists), even with literals
```php
/**
 * @return key-of<list<string>>
 */
/**
 * @return key-of<array{a: string, b: string)>
 */
```
2. Support for unions as `T`-param 
```php
/**
 * @return key-of<array<int, string>|array<float, string>>
 */
```
3. Improve template support of `key-of`, so that correctly not only `array-key` is checked, but instead only methods like `array_keys` can be used to ensure template support.
```php
/**
 * @template T of array
 * @param T $array
 * @return key-of<T>[]
 */
function getKeys($array) {
    return array_keys($array);
}
```
4. Add template support to `value-of` which accepts `array_values` if used correctly.
```php
/**
 * @template T of array
 * @param T $array
 * @return value-of<T>[]
 */
function getValues($array) {
    return array_values($array);
}
```

Unfortunately `@psalm-assert-if-true` seems to not support template types, and even if I fix a bug inside `FunctionLikeDocblockScanner` where template types aren't passed I have no success. So I welcome any suggestions to this! I saw that there is made use of assertions for `array-key-exists` but I didn't find how I can change the type of `$key` so that this test won't fail:
```php
/**
 * @template T of array
 * @param T $array
 * @return key-of<T>|null
 */
function getKey(string $key, $array) {
    if (array_key_exists($key, $array)) {
        return $key;
     }
     return null;
}
```


orklah:
Converting to draft (just for easier maintenance)

Patrick-Remy:
Do you have any idea how I can achieve the described case for `array_key_exists`?

orklah:
I'm not sure I understood the issue exactly, but did you look at this: https://github.com/vimeo/psalm/blob/b633619a2c41f5392d8943ff6876d4963c20ef73/src/Psalm/Internal/Type/AssertionReconciler.php#L278

using array_key_exists will generate an assertion `array-key-exists`. It may be possible to adapt this to be able to change $key.
However, you'll probably start asking questions about this kind of cases: https://psalm.dev/r/e28860fd8a

The array here has int|string keys. You know $key is already a string. What will $key be after you array_key_exists? If you say key-of<$array>, you lost information about $key being a string. If you say string, you lost the information about $key being a valid offset. You'd need something like `string-key-of<$array>`. I think, but this is only a valid solution for keys. Values will be different. And even for keys, if $key was `'a'|'b'` instead of `string`, you're back to square one.

The type system is very complicated and I can only advise that you finish your first PR and all that will be related to it until you start on more

Patrick-Remy:
> The array here has int|string keys. You know $key is already a string. What will $key be after you array_key_exists? If you say key-of<$array>, you lost information about $key being a string. If you say string, you lost the information about $key being a valid offset. 
You're completely right, replacing the type with `key-of<$array>` was a bad idea. 

> You'd need something like `string-key-of<$array>`, I think. But this is only a valid solution for keys, values will be different. And even for keys, if $key was `'a'|'b'` instead of `string`, you're back to square one.
It just would be great, if the snippet would pass with my implementation, as it's completely valid. My current approach was to extend `UnionTypeComparator::isContainedBy` to check if `$container_type` is `TTemplateKeyOf` and there is an assertion `array-key-exists` for the `$input_type` with a relation to the type of `$container_type`, it would be ok. Unfortunately in this context all assertions are gone. So I'm not sure if this is the right context.

Probably this case is something for later. I fixed all other outstanding issues, the original parsing for `TTemplateKeyOf` didn't initialized with the original `TTemplateParam` type, which led to not being replaced and was always `empty`. 
Could you please review again?

orklah:
This will need to target the master branch, removing existing types is a big BC break.

Also, can you fix the errors in shepherd before we review? It's possible you'll see issues that could have a big impact on your existing code. Reviewing is kinda hard, so I'd prefer doing that once everything is fixed!

Patrick-Remy:
Okay, I will rebase to master, do you have some time schedule for the 5.x release?

I'm not quite sure, if the latest shepherd issues are related to my changes. I'll have a look tomorrow.

orklah:
We don't have a release date yet for Psalm 5, we'll make sure to communicate when we will

About the errors on shepherd, I'm pretty sure they're related to your changes. Upstream branch is clean and the errors seem consistent with what you changed

Patrick-Remy:
I think I'm finally done rebasing to `master` and fixing the outstanding shepherd bugs, which were caused by not using `TemplateStandinTypeReplacer::getMostSpecificTypeFromBounds()` to replace the template param.

All failing tests/checks are also currently failing on the `master` branch, if I didn't oversee something.

Ready for review now – hopefully 😅

orklah:
there are errors in shepherd, they are related to a change in test format in master.

We know require tests to be all associative arrays. Can you change the format of the tests you added?

Patrick-Remy:
I already changed the format of my added tests, but there tests in the `master` branch which aren't correct, see:

my shepherd run: https://github.com/vimeo/psalm/runs/4859814032?check_suite_focus=true
latest master shepherd run: https://github.com/vimeo/psalm/runs/4857096400?check_suite_focus=true

orklah:
sorry, this is actually my fault. I'm fixing that

orklah:
fixed :) Thanks for reporting, I forgot to check after the last merge

Patrick-Remy:
Thanks! I just rebased 👍

src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
```diff
@@ -291,6 +289,20 @@ public static function isContainedBy(
...
+        if ($container_type_part instanceof TArrayKey
+            && $input_type_part instanceof TNumeric
+        ) {
+            return true;
+        }
+
+        if ($container_type_part instanceof TArrayKey
+            && ($input_type_part instanceof TInt
+                || $input_type_part instanceof TString
+                || $input_type_part instanceof TTemplateKeyOf)
```
orklah, line 301: This specific line is redundant, it would have been handled line 270

src/Psalm/Internal/Type/TypeExpander.php
```diff
@@ -878,4 +849,85 @@ private static function expandConditional(
...
+                ) ;
+            } else {
+                $new_return_types = array_merge(
+                    $new_return_types,
+                    array_values($type_atomic->type_params[1]->getAtomicTypes())
+                ) ;
+            }
+        }
+
+        if (empty($new_return_types)) {
```
orklah, line 928: please avoid empty, `if(new_return_types !== [])` is more clear

src/Psalm/Internal/Type/TypeParser.php
```diff
@@ -721,16 +718,16 @@ private static function getTypeFromGenericTree(
                 throw new TypeParseTreeException('Union types are not allowed in value-of type');
             }
 
-            if (!$param_union_types[0] instanceof TClassConstant) {
+            if (!$param_union_types[0] instanceof TArray
```
orklah, line 721: couldn't you reuse TKeyOfArray::isViableTemplateType ?

src/Psalm/Type/Atomic/TValueOfArray.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public function canBeFullyExpressedInPhp(int $analysis_php_version_id): bool
+    {
+        return false;
+    }
+
+    public function getAssertionString(bool $exact = false): string
+    {
+        return 'mixed';
```
Patrick-Remy, line 52: I'm not sure about this, I just adopted from old `TValueOfClassConstant`. The default implementation calls `getId()` what should be a return value of this method?

src/Psalm/Type/Atomic/TKeyOfArray.php
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Represents an offset of an array.
+ *
+ * @psalm-type ArrayLikeTemplateType = TClassConstant|TKeyedArray|TList|TArray
+ */
+class TKeyOfArray extends TArrayKey
+{
+    /** @var ArrayLikeTemplateType */
+    public $type;
```
orklah, line 15: So, If I understand this right, whenever you need to create a TKeyOfArray, you store the type of the array at the moment of the TKeyOfArray creation.

Does this mean that if the array is refined after the creation of the TKeyOfArray (for example, with isset on specific keys), the TKeyOfArray will have obsolete information?

Also, Is there a reason to store the entire array? Wouldn't the key types be enough?

src/Psalm/Type/Atomic/TKeyOfArray.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * @param ArrayLikeTemplateType $type
+     */
+    public function __construct(Atomic $type)
```
orklah, line 20: a Single Atomic? Isn't there cases where you have to create a TKeyOfArray based on an Union of two arrays and you need to handle an Union in there?

Patrick-Remy: You're right, this is something I wanted to change later (when all tests pass), and then forgot about it!

src/Psalm/Type/Atomic/TValueOfArray.php
orklah: it should be fine. getAssertionString is used when we want to refine a type from another. returning mixed is like saying there's nothing to refine, I'm not sure we can do a lot more.

Possibly you could infer the actual type of the key or values and if the type is simple enough (like `string`) returning it here, but I doubt this will be very used

tests/KeyOfArrayTest.php
```diff
@@ -0,0 +1,189 @@
...
+                        public function getKey(bool $adams) {
+                            return array_keys(self::FOO);
+                        }
+                    }
+                '
+            ],
+            'keyOfArrayLiteral' => [
+                'code' => '<?php
+                    /**
+                     * @return key-of<array<int, string>>
```
orklah, line 82: Is this something we want to accept? Does it have any concrete use as opposed to returning `int`?

tests/ArrayKeysTest.php
```diff
@@ -0,0 +1,137 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Tests;
+
+use Psalm\Tests\Traits\InvalidCodeAnalysisTestTrait;
+use Psalm\Tests\Traits\ValidCodeAnalysisTestTrait;
+
+class ArrayKeysTest extends TestCase
```
orklah, line 10: Just checked for clarity: all those tests already work on master. Those aren't tests that are solved by the current PR

tests/KeyOfArrayTest.php
Patrick-Remy: I added it, to test if `key-of<properties-of<Foo>>` will work, if I rewrote the related PR to return a keyed-array  instead of only the names.

Additionally this works in TypeScript as well.

tests/ArrayKeysTest.php
Patrick-Remy: Yes, but they are needed to ensure the functionality still works. While developing, I broke many of these and I think it's nice, if somebody else couldn't break it too 😀

tests/KeyOfArrayTest.php
orklah: Fine by me :)

tests/ArrayKeysTest.php
orklah: yeah, this is useful, I Just try to grasp the extent of the change so knowing what snippets are actually better/different than before is useful for me, so I noted it for other reviewers too

tests/Template/KeyOfTemplateTest.php
```diff
@@ -0,0 +1,110 @@
...
+                     * @param T $array
+                     * @return key-of<T>|null
+                     */
+                    function getKey($array) {
+                        return array_key_last($array);
+                    }
+                '
+            ],
+            // Currently not works!
+            // 'acceptsIfArrayKeyExistsFn' => [
```
AndrolGenhald, line 58: Would you mind uncommenting and putting `SKIPPED-` in front of the name? Most of the other failing tests due to missing features are skipped that way, and it makes it much easier to later un-skip all of the skipped tests to see if any have started passing.

tests/KeyOfArrayTest.php
```diff
@@ -0,0 +1,189 @@
...
+     * @return iterable<string,array{code:string,assertions?:array<string,string>,ignored_issues?:list<string>,php_version?:string}>
+     */
+    public function providerValidCodeParse(): iterable
+    {
+        return [
+            'keyOfListClassConstant' => [
+                'code' => '<?php
+                    class A {
+                        const FOO = [
+                            \'bar\'
```
orklah, line 25: Please use either double quotes for strings literals or double quotes for quoting the entire snippet instead of escaping. This is way easier when copy/pasting a snippet down the line

tests/Template/KeyOfTemplateTest.php
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public function providerValidCodeParse(): iterable
+    {
+        return [
+            'acceptsArrayKeysFn' => [
+                'code' => '<?php
+                    /**
+                     * @template T of array
+                     * @param T $array
+                     * @return key-of<T>[]
```
orklah, line 26: seems like putting `value-of<T>[]` here breaks. I'd not have expected that

Patrick-Remy: There is no `TTemplateValueOf` currently.

orklah: oh, right. Would that be easy to add? It feels like TValueOfArray is kinda meaningless on standard arrays without the possibility to combine this with templates

tests/KeyOfArrayTest.php
AndrolGenhald: @orklah I've thought about refactoring tests to use nowdocs for code, unfortunately the indented end identifier requires PHP 7.3. Maybe for Psalm 5?

orklah: I just tried, I'm not really a fan of how PHPStorm handle that. You have to put the `<?php` on a new line and not indented for coloration to work. Then it tells you the end of nowdoc should have the same indentation (so none). That looks more messy at the end

tests/Template/KeyOfTemplateTest.php
Patrick-Remy: Should be easy, I'll give it a try!

src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
Patrick-Remy: There is now also `TValueOfArray::isViableTemplateType()` and used there.

src/Psalm/Internal/Type/TypeExpander.php
Patrick-Remy: Done

src/Psalm/Internal/Type/TypeParser.php
Patrick-Remy: Yes, I adjusted `isViableTemplateType` to handle unions and can now use it there for `$param_union_types`.

src/Psalm/Type/Atomic/TKeyOfArray.php
Patrick-Remy: Done.

tests/Template/KeyOfTemplateTest.php
Patrick-Remy: Done

tests/KeyOfArrayTest.php
Patrick-Remy: I replaced it with double quotes inside.

src/Psalm/Type/Atomic/TKeyOfArray.php
Patrick-Remy: > Does this mean that if the array is refined after the creation of the TKeyOfArray (for example, with isset on specific keys), the TKeyOfArray will have obsolete information?
I'm not sure how that would be possible, I now save the reference to the union to support unions. Does this union get updated on asserts?

orklah: Don't mind the first part, I didn't grasp the change well enough when I wrote that.

About the second part, I'd have said `getGenericArrayType` for TKeyedArray, TInt for TList and type_params[0] for TArray, but it may create issues, I'm not sure

Patrick-Remy:
I think I ran into a psalm bug `$constant_type` is obviously not unused?
https://github.com/vimeo/psalm/runs/4888882053?check_suite_focus=true#step:4:29

https://github.com/vimeo/psalm/pull/7396/files#diff-a5d71d5e11f033fa672c82b7a71184f60e1f5f7ce858548f00ef45426fb28d0fR880-R908

orklah:
It is unused. On the next try catch you either override it in the try or you return without using it in the catch. There is no paths that lead to the initial null being kept.

src/Psalm/Type/Atomic/TKeyOfArray.php
Patrick-Remy: I'm also not sure about it, and for constants the key-of/value-of implementation would still need to be duplicated inside `TypeExpander`.
At least I think it's a bit easier to understand the code like this.

Patrick-Remy:
> It is unused. On the next try catch you either override it in the try or you return without using it in the catch. There is no paths that lead to the initial null being kept.

Thanks, I just forgot that variables in PHP aren't as scoped as I thought :D

tests/Template/KeyOfTemplateTest.php
Patrick-Remy: I'll add this next week.

src/Psalm/Type/Atomic/TKeyOfArray.php
orklah: Nevermind then :)

orklah:
> I just forgot that variables in PHP aren't as scoped as I thought :D

I'm curious, what did you expect then?

Patrick-Remy:
A bit off-topic, but e.g. in JS and Swift this here:

```js
try {
  let foo = 42;
} catch(e) {}

console.log(foo);
```

leads to `foo is not defined`. As variables are block-scoped, not like in PHP.

orklah:
right, but in that case it makes no difference given the return in catch, no?

AndrolGenhald:
@Patrick-Remy Don't know if you care to, but you might be able to add support for stuff like [this](https://psalm.dev/r/3531bf3469) without too much trouble.

Patrick-Remy:
@orklah any idea how to pass this test here?

```php
<?php
                    /**
                     * @template TKey of int
                     * @template TArray of array<TKey, bool>
                     * @param TArray $array
                     * @return list<TKey>
                     */
                    function toListOfKeys(array $array): array {
                        return array_keys($array);
                    }
```

I have no idea why, but while debugging `TArray` has key-type `array-key` instead of expected `int`.

Patrick-Remy:
> @Patrick-Remy Don't know if you care to, but you might be able to add support for stuff like [this](https://psalm.dev/r/3531bf3469) without too much trouble.

@AndrolGenhald sorry, this PR has already been 'exploded' a bit and took a lot of time, should be handled in a separate PR after this has been merged.

orklah:
> @orklah any idea how to pass this test here?

Not sure what you mean, the snippet https://psalm.dev/r/d16778139e seems to work and return `list<1>` on your branch as I'd expect

Patrick-Remy:
> > @orklah any idea how to pass this test here?
> 
> Not sure what you mean, the snippet https://psalm.dev/r/d16778139e seems to work and return `list<1>` on your branch as I'd expect

Are you sure you ran it on my branch (commit `51cab704c4626b8ee55472d82542ea4b031b8960`)? I get:
```
1) Psalm\Tests\KeyOfTemplateTest::testValidCode with data set "keyOfNestedTemplates" ('<?php\n                /**\n ...$a */;')
Psalm\Exception\CodeException: MixedReturnTypeCoercion - src/somefile.php:9:27 - The type 'list<key-of<TArray:fn-array_keys as TArray:fn-array_keys as array<array-key, mixed>>>' is more general than the declared return type 'list<TKey:fn-tolistofkeys as int>' for toListOfKeys
```

It's also already added as a test in `tests/Template/KeyOfTemplateTest.php` as `SKIPPED-`.

Patrick-Remy:
I got it, I didn't implement `replaceWithTemplateArgs` and I didn't add a comparison for `TTemplateKeyOf` also in `AtomicTypeComparator` **before** `TTemplateParam`. Need to refactor a bit tomorrow and then it will be ready for review!

Patrick-Remy:
@orklah It's finally done – could you review again please?

src/Psalm/Internal/Provider/FunctionReturnTypeProvider.php
```diff
@@ -79,7 +78,6 @@ public function __construct()
         $this->registerClass(ArraySpliceReturnTypeProvider::class);
         $this->registerClass(ArrayReverseReturnTypeProvider::class);
         $this->registerClass(ArrayUniqueReturnTypeProvider::class);
-        $this->registerClass(ArrayValuesReturnTypeProvider::class);
```
orklah, line 82: It's redundant now, right?

Patrick-Remy: Right, I removed the return type provider and added it as a stub. 

https://github.com/vimeo/psalm/pull/7396/files#diff-483b988f07809c751ac53f905f9b4dc8b0093e9af15578ee3ac009b9dfe394b5R170-R181

src/Psalm/Type/Atomic/TTemplateKeyOf.php
```diff
@@ -2,12 +2,16 @@
...
+use Psalm\Internal\Type\TemplateResult;
+use Psalm\Type\Atomic;
 use Psalm\Type\Union;
 
 /**
- * Represents the type used when using TKeyOfClassConstant when the type of the class constant array is a template
+ * Represents the type used when using TKeyOfArray when the type of the array is a template
  */
-class TTemplateKeyOf extends TArrayKey
+class TTemplateKeyOf extends Atomic
```
orklah, line 14: why this change? It seems like it would indeed be an array-key?

Patrick-Remy: This caused the bug with nested templates. `TArrayKey` extends from `Scalar`, so if `TTemplateKeyOf` have been compared to a Scalar value the `ScalarTypeComparator` has been called. But if it has been compared to a `TTemplateParam` the `AtomicTypeComparator` has been called. This lead to a duplicate implementation of the `isContainedBy`-logic. Therefor I decided to extend from `Atomic` as there is no benefit from extending from `Scalar`.

orklah: ok, seems fine

orklah:
I'm ready to merge, no last change? :)

Patrick-Remy:
Needs anything to be added to the BC file?

orklah:
Yeah, maybe document the change of inheritance

Patrick-Remy:
I also added the renames as BCs.

orklah:
Thanks!

Patrick-Remy:
Many thanks for your intensive and always fast review! 😊

PR #7402, opened 2022-01-16
Fix method signatures for ReflectionProperty

### LINKED ISSUE #7357: `MethodSignatureMismatch` when extending `ReflectionProperty`
`ReflectionProperty::setValue()` and `ReflectionProperty::getValue()` are polymorphic. If you work on a static property you can omit the first parameter which is the object on which you want to access the property.

The following code is valid imho:

```PHP
lass MyReflectionProperty extends ReflectionProperty
{
    public function getValue(?object $object = null): mixed
    {
        return parent::getValue(...func_get_args());
    }

    public function setValue(mixed $objectOrValue, mixed $value = null): void
    {
        parent::setValue(...func_get_args());
    }
}
```

see https://psalm.dev/r/7de1b5f18a and https://3v4l.org/Ie4oo

However, Psalm logs `MethodSignatureMismatch` errors. If I apply the changes suggested by those errors, I end up with code that would exit with a fatal error:

see https://3v4l.org/mX1AF

derrabus:
Fixes #7357


derrabus:
Sorry, I have no clue how to fix that test. 🙈

AndrolGenhald:
@derrabus See eg #7406, there are delta callmaps for each PHP version. If the signature needs to be changed in all of them, update the historical callmap so it matches, if it needs to be changed only for some specific version onward, add the change to that delta.

orklah:
@derrabus no worries! We have a documentation page dedicated to those changes here: https://psalm.dev/docs/contributing/editing_callmaps/ if you need more informations. But yeah, @AndrolGenhald is right, Callmap.php is the current state of PHP8.1 but we use delta files in order to pinpoint what was the state of every other PHP version until "historical" (which is 7.0), so you need to update those too.

weirdan:

weirdan:
The original issue seems to be already resolved.

PR #7407, opened 2022-01-16
Improve signature of DOMDocument::loadXML()

### LINKED ISSUE #7039: `DOMDocument::loadXML()` should reject empty strings
We should probably annotate `$source` as `non-empty-string` so that Psalm would be able to catch it too.

_Originally posted by @weirdan in https://github.com/vimeo/psalm/issues/7036#issuecomment-983996885_

https://3v4l.org/en00b

https://psalm.dev/r/4bb02d8813

villfa:
Closes #7039


src/Psalm/Config.php
```diff
@@ -696,6 +696,7 @@ private static function loadDomDocument(string $base_dir, string $file_contents)
         $oldpwd = getcwd();
         chdir($base_dir);
 
+        assert($file_contents !== '');
```
orklah, line 699: please document on the param instead, this way Psalm can make sure this is enforced at the source

orklah:
Thanks!

PR #7409, opened 2022-01-16
refactor description of types in Atomic

orklah:
This PR propose to refactor the description of types in Atomic.

I always thought it was messy.

For example:
- In Atomic, getId() called __toString()
- In CallableTrait, __toString() called getId()
- In TCallableString, getId() called getKey()
- In TClassString, __toString() called getKey()
- In TConditional, getKey() called __toString

Aaand I'm already lost. Adding to that the fact that no documentation exists on that, it didn't help.

What I changed:
- __toString is now final in Atomic and calls getId() with default params
- getId() has a new param `$exact` who defaults to true. When true, it displays the most detailed type it can, otherwise, it displays a generic version (it depends widely on each type)
- getId() now calls getKey() by default unless overrided
- I added some documentation

almost every other change is a consequence of changing the things above. Mainly:
- getId() has now two different caches in Union for exact mode and non-exact mode. This may degrade memory a little but it could gain performances (where before, calls to __toString were not cached)
- __toString used to display the non-exact type, now it calls getId() with $exact = true, so wherever I needed, I removed the implicit calls to __toString for a getId(false)
- getId() on TLiteralString was using double quotes to display the types. This is not consistent with the rest of the types (in docblock for example) so I changed that

I'm still not completely satisfied. I think there is still too many redundancy between getId(), getKey(), toNamespacedString, toPhpString. I feel we should be able to go further but it's already a big step forward

@muglug Do you think this goes in the right direction? Maybe you have special insight here? I'm not sure what getId() was supposed to be. I also have difficulties understanding when getKey should be overrided and with what value

### DIFF
--- a/src/Psalm/Type/Atomic/TLiteralString.php
+++ b/src/Psalm/Type/Atomic/TLiteralString.php
@@ -24,15 +24,18 @@ class TLiteralString extends TString
         return 'string(' . $this->value . ')';
     }
 
-    public function getId(bool $nested = false): string
+    public function getId(bool $exact = true, bool $nested = false): string
     {
+        if (!$exact) {
+            return 'string';
+        }
         // quote control characters, backslashes and double quote
         $no_newline_value = addcslashes($this->value, "\0..\37\\\"");
         if (mb_strlen($this->value) > 80) {
-            return '"' . mb_substr($no_newline_value, 0, 80) . '...' . '"';
+            return "'" . mb_substr($no_newline_value, 0, 80) . '...' . "'";
         }
 
-        return '"' . $no_newline_value . '"';
+        return "'" . $no_newline_value . "'";
     }
 
     public function getAssertionString(): string
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -1553,7 +1553,7 @@ class ClassAnalyzer extends ClassLikeAnalyzer
             $message .= ' - consider ' . str_replace(
                 ['<array-key, mixed>', '<never, never>'],
                 '',
-                (string)$suggested_type
+                $suggested_type->getId(false)
             );
         }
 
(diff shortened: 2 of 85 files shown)


orklah:
@muglug do you have an opinion on this? I don't want to botch up this if it had a specific goal I'm missing...

muglug:
Sorry, I think this is a good idea!

muglug:
Occasionally `__toString` has been used for more human-readable things but I don't really think that has value now. `getKey` is mainly used to ensure that arrays are always stored under the same array key.

src/Psalm/Type/Atomic/TLiteralString.php
```diff
@@ -24,15 +24,18 @@ public function getKey(bool $include_extra = true): string
...
+        }
         // quote control characters, backslashes and double quote
         $no_newline_value = addcslashes($this->value, "\0..\37\\\"");
         if (mb_strlen($this->value) > 80) {
-            return '"' . mb_substr($no_newline_value, 0, 80) . '...' . '"';
+            return "'" . mb_substr($no_newline_value, 0, 80) . '...' . "'";
         }
 
-        return '"' . $no_newline_value . '"';
+        return "'" . $no_newline_value . "'";
```
AndrolGenhald, line 38: @orklah Was this an intentional change? It caused one of my tests to fail when rebasing because `$foo = "bar"` now has type `'bar'` instead of `"bar"`.

orklah: This was indeed intended. I mentionned it in the PR
> getId() on TLiteralString was using double quotes to display the types. This is not consistent with the rest of the types (in docblock for example) so I changed that

I didn't thought about plugins, I'll add this in UPGRADE file

orklah: https://github.com/vimeo/psalm/commit/efe9c2bc6d55f532b69c63324e35dd98f8139dfd

PR #7410, opened 2022-01-17
Use objects, not strings, for assertions

muglug:
It's a big addition, but this would make it slightly easier to track where assertions were being used in the codebase, and removes a lot of checks for `$assertion[0] === '!'` etc.


orklah:
This looks really good!

orklah:
Hi @muglug ! Do you think you'll finish this soon? I'd really like having this in the alpha for Psalm 5!

muglug:
Yup, should have unit tests passing today

orklah:
Sorry, my last merge on master broke CI. I'll fix that

src/Psalm/Internal/Type/AssertionReconciler.php
```diff
@@ -1257,21 +1179,15 @@ private static function handleLiteralEqualityWithString(
                     return $existing_var_type;
                 }
 
-                if ($scalar_type === 'class-string'
-                    || $scalar_type === 'interface-string'
-                    || $scalar_type === 'trait-string'
-                ) {
+                if ($assertion_type instanceof TLiteralClassString) {
```
orklah, line 1182: this condition can be dropped (there are a few like that in the file)

orklah:
Could you add some details on UPDGRADE.md to explain that the old Assertion was renamed and other BC breaks introduced?

src/Psalm/Type/Atomic/TClassString.php
```diff
@@ -51,7 +58,9 @@ public function __toString(): string
 
     public function getId(bool $nested = false): string
     {
-        return $this->getKey();
+        return ($this->is_loaded ? 'loaded-' : '')
+            . ($this->is_interface ? 'interface' : 'class')
+            . '-string' . ($this->as === 'object' ? '' : '<' . $this->as_type . '>');
     }
 
     public function getAssertionString(bool $exact = false): string
```
orklah, line 66: every call to getAssertionString is now with $exact to true. Can we drop the param?

orklah:
every call to getAssertionString is now with $exact to true. Can we drop the param?

src/Psalm/Internal/Type/AssertionReconciler.php
muglug: Why can it be dropped?

src/Psalm/Type/Atomic/TCallableObject.php
```diff
@@ -34,7 +34,9 @@ public function canBeFullyExpressedInPhp(int $analysis_php_version_id): bool
         return false;
     }
 
-    public function getAssertionString(bool $exact = false): string
+    public function getAssertionString()
+
+    /** @psalm-mutation-free */: string
```
orklah, line 39: this feels weird

src/Psalm/Internal/Type/AssertionReconciler.php
muglug: Done!

tests/TypeReconciliation/EmptyTest.php
```diff
@@ -473,19 +473,6 @@ function foo(array $r) {
                     }',
                 'error_message' => 'MixedReturnTypeCoercion'
             ],
-            'secondEmptyTwice' => [
```
orklah, line 476: why did you drop this? Seems like it still works on my side

muglug: I migrated the test (in the more common falsiness check form) to `RedundantConditionTest.php`

src/Psalm/Internal/Type/AssertionReconciler.php
orklah: The one on line 1456 can still be dropped :)

orklah:
This makes so many things looks a lot more clearer! Very cool! Thanks :)

src/Psalm/Storage/Assertion/Empty_.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public function isNegation(): bool
+    {
+        return true;
+    }
+
+    public function __toString(): string
+    {
+        return '!non-empty';
```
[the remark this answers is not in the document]
orklah: Somehow it still works? https://psalm.dev/r/bb6d82a6a3 https://psalm.dev/r/90da8dba12

I didn't expect that. What is the system that parses assertion into Assertion objects?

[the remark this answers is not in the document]
AndrolGenhald: @orklah It's the [scanner](https://github.com/vimeo/psalm/blob/2e01e9b7cdefed310ee506abb2bd063ba380f883/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockScanner.php#L594).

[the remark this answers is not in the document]
orklah: Thanks! I never had to tweak here before... Weird, I thought there would have been a direct link between internal assertion names and user assertions ones. But this answer my question! Thanks both!

Issue #7411, opened 2022-01-17
`parent` doesn't account for `@method` annotations?


leighman:
Maybe I'm being stupid but I would not expect `UndefinedMethod` here because of the `@method` annotation. The annotation seems to work when just calling the method.
https://psalm.dev/r/ccfc1a31f2

orklah:
Seems like a bug indeed. @vincent4vx you worked on something very related. Would you mind looking at this?

vincent4vx:
It looks like psalm considers parent::create as static call

orklah:
yeah, I noticed something similar some time ago.

The syntax for an instance method returning static and a static method without return is the same...

We should probably forbid static method not having return type and assume a single "static" is a dynamic method.

Not sure how phpstan handles this

vincent4vx:
PHPStan works as expected : https://phpstan.org/r/15872adb-5dd2-4436-a46a-e07cd5632d1f
The static keyword will be used as return type, like psalm.

Thue current issue is that `parent::xxx()` is analysed by `AtomicStaticCallAnalyzer`, so in case of magic method, it'll check for `__callstatic` method existence, and read the MethodStorage from `pseudo_static_methods`.

This issue do not occurs on instance methods, because static and instance methods are all stored into `methods` property of `ClassLikeStorage`.

orklah:
Your diagnostic seems correct, I fixed the issue by adding handling for `pseudo_methods` when we detect a call with `parent::` but I have an issue with the static return type resolution. I'll push the PR when I find the issue. Thanks!

PR #7414, opened 2022-01-17
Add support of docblock method using parent keyword

vincent4vx:
See: #7411

Currently, when using `parent::`, psalm consider the call as static, so resolve from `pseudo_static_method` instead of `pseudo_method`.
This fix will forward to instance method analyser when `parent::` call is detected, in instance method context, and if the method is not declared.


orklah:
The shepherd error is surprising because the function does not actually exists(removed in phpunit 9) so the suppression should still be used.

On another topic, I didn't expect you'd add a new condition here: https://github.com/vimeo/psalm/pull/7414/files#diff-eb58cff4a0b42fdcb5bb7462a60e099ef97f49c766c10c1086b50f63bc1251f3R653

I was looking at two possibilities on my end:
- adding the pseudo_methods check in this function: https://github.com/vimeo/psalm/blob/5b82082bbdb0d34be99ee537876a894c6230ad55/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php#L489 and going with the flow (this was the solution I tested but I had issues with return types)
- flagging this variable https://github.com/vimeo/psalm/blob/5b82082bbdb0d34be99ee537876a894c6230ad55/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php#L785 to true in order to fallback on dynamic call

Those two seemed more appropriate for this kind of change. I may be wrong here so if you have a legit reason not to use those, go ahead

vincent4vx:
Adding the check of `pseudo_methods` is not enough, because the `__callstatic` method will be checked, which is not expected for `parent::` call, because it should use `__call` instead. Also, in case of undeclared pseudo method, types should be resolved using `__call` and not `__callstatic` (I'll add a test case).

And for the second option, method existence and visibility is checked before, so it's not possible to implements pseudo method here.

Maybe my condition is not at the correct place ?

Interesting behavior on PHP : https://3v4l.org/tfYli
When `__call` is not found, the call will be forwarded to `__callStatic` :shrug:

orklah:
I'm fine with the answer :) 

Can you check why there is an error on shepherd for a non-existing method? After this is fixed, this seems fine to merge

vincent4vx:
Maybe a cache issue ? Merge of branch 4.x has resolve the error.

orklah:
There's no cache on our CI.

And no, it disappeared after a rebase because I merged this: https://github.com/vimeo/psalm/pull/7435

But it's concerning that the error was here in the first place. Can you revert my change on your end to check why the error was here?

vincent4vx:
Oh, it's a bug of my feature...

Because the method do not exists on parent, it will be forwarded to instance method call.
But the static type (i.e. type of $this) will be used instead of the parent one, so the method will exist because it's declared on child with the same name...

So, this code will not raise UndefinedMethod :
```php
<?php
                    class A {}
                    class B extends A {
                        public function foo(): string {
                            return parent::foo();
                        }
                    }
```

orklah:
Thanks!

Issue #7412, opened 2022-01-17
PossiblyUndefinedIntArrayOffset is still emitted after array_key_exists 


danog:
https://psalm.dev/r/23de0efe71

PR #7415, opened 2022-01-17
Array key exists assert both ways

orklah:
This will fix https://github.com/vimeo/psalm/issues/7412

All the code was already there, but it was only using the keys from the array in second param to restrict the type of the first param.

Only when that was failing (the else I dropped), it tried to use the value of the first param to restrict the keys from the array in the second param.

This PR does both, always.


orklah:
Note: This will probably create a conflict on https://github.com/vimeo/psalm/pull/7415

orklah:
I'll put this in draft. I'll probably start this again from scratch on master to avoid conflicts with #7415

orklah:
superseded by #7449

PR #7417, opened 2022-01-18
Contextual type inference for high order function arg

klimick:
Since php doesn't have extension method functionality like C# then pipe function can be a good solution to the situation:

```php
<?php

$greaterThan100 = pipe2(
    [1, 2, 3],
    map(function ($i) {
        return $i > 1 ? $i + 1 : $i + 2;
    }),
    filter(function ($i) {
        return $i > 100;
    })
);
```

Before we can write generic pipe function we should support inference for partially applied closures (`fn($a) => fn($b) => $a + $b`)

This PR tries to implement inference of this kind.
P.S. I tried to do it through the plugin system. It didn't work out.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
@@ -54,6 +54,7 @@ use UnexpectedValueException;
 use function array_map;
 use function array_reverse;
 use function array_slice;
+use function assert;
 use function count;
 use function in_array;
 use function is_string;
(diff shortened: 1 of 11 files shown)


orklah:
Can you add a snippet in test to show the easiest possible code that is solved by this PR? Also could you add some comments in the code describing what you're doing. It will help for the review and also makes easier to read this in the future

klimick:
Done. If there is anything else that needs to be added, please let me know)

orklah:
Sorry, I should have warned before but could you target this on master? We'll start avoiding to land big new features on 4.x to prevent maintenance on old versions once Psalm 5 is out

orklah:
Thanks! @drupol @veewee I think this could interest you

drupol:
OMG !!! I don't have much time these days, but I'm planning to review that and the work of @veewee !

drupol:
I think @Ocramius and @AlexandruGG might be interested as well !

Crell:
Oh my.

I don't know the Psalm codebase well enough, but the generic pipe function already exists: https://github.com/Crell/fp

Includes a whole bunch of related utilities as well.  It's currently very well-audited by PHPStan, but I haven't tried Psalm on it.  (PRs welcome for that.)

veewee:
Hello @klimick 

Nice work on this PR! Will check out the details soon.

Some context from my side:

I am also trying to build a variadic pipe plugin as well and currently ended up with this:
* Plugin: https://github.com/php-standard-library/psalm-plugin/blob/main/src/EventHandler/Fun/Pipe/PipeArgumentsProvider.php
* Function: https://github.com/azjezz/psl/blob/2.0.x/src/Psl/Fun/pipe.php

It works quite well, but only works for callable arguments.

Now I am trying to use the DataTypeProvider in order to get the best type information directly from psalm when calculating the attributes. This is currently not possible yet - but would be the best solution IMO.

See:

* Discussion: https://github.com/azjezz/psl/blob/2.0.x/src/Psl/Fun/pipe.php
* Draft PR (still broken ...) :  https://github.com/vimeo/psalm/pull/7394

I guess you probably also need the type information from the probided variadic pipeline stages to build your plugin.
Care to take a look at it? :)

PR #7420, opened 2022-01-18
Remove forbidEcho

orklah:
This will remove forbidEcho that was deprecated in V4

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -427,11 +427,6 @@ class Config
     /**
      * @var bool
      */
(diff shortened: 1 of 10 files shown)


src/Psalm/Config.php
```diff
@@ -424,11 +424,6 @@ class Config
      */
     public $forbidden_functions = [];
 
-    /**
-     * @var bool
-     */
-    public $forbid_echo = false;
```
weirdan, line 430: Please mention this in `UPGRADING.md`.

Issue #7419, opened 2022-01-18
Generics are not resolved on inherited docblock methods


vincent4vx:
See: https://psalm.dev/r/7d0c3bfaa3

Generics resolution is lost on inherited pseudo methods. Also, generics parameters are lost when use $this return type, so method chaining is not possible.

PR #7430, opened 2022-01-19
Resolve generics of inherited pseudo methods

vincent4vx:
See: #7419 

This PR contains 2 fixes :
- Use real called object type on template resolution
- Detect the class like which actually declare the called pseudo method

For the second fix, pseudo methods are copied to the child class like storage, so the declaring class is missing. That why tracking of the declaring method id is required, and lookup from `parent_classes` is not required.

But, pseudo methods declared on implemented interface are not copied (actual cause of #7198), so resolution of the declaring class like cannot be performed only by a lookup on `declaring_pseudo_method_ids` for this particular case.

Should I copy all pseudo_methods (and also declaring_pseudo_method_ids) from implemented interfaces in this PR, or it should be fixed on another PR ?


orklah:
Small PR is good :) It's easier to review! I'll merge this one for now

PR #7431, opened 2022-01-19
Add configuration option to disable `@psalm-suppress all`

danog:
Personally, I'd completely remove this feature in Psalm v5, it's way too dangerous, baselines+config are enough to suppress all issues in specific files of very legacy codebases.


weirdan:
Thanks!

PR #7432, opened 2022-01-19
Fix Incomplete return type for mb_split() function

### LINKED ISSUE #7421: Incomplete return type for mb_split() function
Function ```mb_split()``` should have a return type of ```array|false``` [per the PHP documentation](https://www.php.net/manual/en/function.mb-split.php). The Psalm call maps only have ```array``` leading to a ```TypeDoesNotContainType``` "list\<string\> does not contain false" error when type checking. See https://psalm.dev/r/0b69779c9f to reproduce the issue.

Changing the call maps from:

```
'mb_split' => ['list<string>', 'pattern'=>'string', 'string'=>'string', 'limit='=>'int'],
```
to:
```
'mb_split' => ['list<string>|false', 'pattern'=>'string', 'string'=>'string', 'limit='=>'int'],
```
should resolve it.

RishiKumarRay:
Fix #7421
Incomplete return type for mb_split() function


orklah:
Thanks @RishiKumarRay ! Congrats for your first PR!

RishiKumarRay:
Thanks @orklah

PR #7442, opened 2022-01-20
Allow config to define thread count

M1ke:
For large projects many systems will not handle multithreading correctly. For example if run in a container, the underlying machine may report 32 threads, but the container has access to 2. Or in large projects where there's a memory limit (e.g. our codebase takes 7GB) running with more than 1 thread, even if possible, consumes all available memory.

Whilst threads can be set on the command line, this isn't always possible. My specific example is the way that PhpStorm runs Psalm, but I imagine there are other tools which may run Psalm off a default config or with their own flags, where a user cannot configure what flags are set. Generally it's worth most flags being settable in a config file anyway, so that behaviour can be locked down - for example in our project for local developers, we have added a set of bash scripts to run Psalm with various flags, and all of them force threads=1.

This PR adds "threads" as a config option. If set, and if lower than the detected threads, it will be applied. If unset or higher than detected threads, existing behaviour is applied. Thus this cannot disrupt existing mechanisms of setting threads to 1.

I've been unable to run tests locally due to a weird bug with composer that's failing to unzip the symyony/filesystem plugin, so I will watch the CI build and make adjustments if required.

### DIFF
--- a/docs/running_psalm/configuration.md
+++ b/docs/running_psalm/configuration.md
@@ -402,8 +402,7 @@ Allows you to hard-code a serializer for Psalm to use when caching data. By defa
         threads="[int]"
 >
 ```
-Allows you to hard-code the number of threads Psalm will use (similar to `--threads` on the command line). This value will only be applied if it is lower than the number of threads Psalm would natively use (i.e. you cannot override other settings that force threads=1 or utilise more threads than detected on the machine)
-
+Allows you to hard-code the number of threads Psalm will use (similar to `--threads` on the command line). This value will be used in place of detecting threads from the host machine, but will be overridden by using `--threads` or `--debug` (which sets threads to 1) on the command line
 
 ## Project settings
 
--- a/docs/running_psalm/configuration.md
+++ b/docs/running_psalm/configuration.md
@@ -402,7 +402,7 @@ Allows you to hard-code a serializer for Psalm to use when caching data. By defa
         threads="[int]"
 >
 ```
-Allows you to hard-code a serializer for Psalm to use when caching data. By default, Psalm uses `ext-igbinary` *if* the version is greater than or equal to 2.0.5, otherwise it defaults to PHP's built-in serializer.
+Allows you to hard-code the number of threads Psalm will use (similar to `--threads` on the command line). This value will only be applied if it is lower than the number of threads Psalm would natively use (i.e. you cannot override other settings that force threads=1 or utilise more threads than detected on the machine)
 
 
 ## Project settings
--- a/docs/running_psalm/configuration.md
+++ b/docs/running_psalm/configuration.md
@@ -396,6 +396,14 @@ Whether or not to allow `require`/`include` calls in your PHP. Defaults to `true
 ```
 Allows you to hard-code a serializer for Psalm to use when caching data. By default, Psalm uses `ext-igbinary` *if* the version is greater than or equal to 2.0.5, otherwise it defaults to PHP's built-in serializer.
 
+#### threads
+```xml
+<psalm
+        threads="[int]"
+>
+```
+Allows you to hard-code a serializer for Psalm to use when caching data. By default, Psalm uses `ext-igbinary` *if* the version is greater than or equal to 2.0.5, otherwise it defaults to PHP's built-in serializer.
+
 
 ## Project settings
 
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -563,6 +563,9 @@ class Config
      */
     public $internal_stubs = [];
 
+    /** @var ?int */
+    public $threads;
+
     protected function __construct()
     {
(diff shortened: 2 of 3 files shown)


docs/running_psalm/configuration.md
```diff
@@ -396,6 +396,14 @@ Whether or not to allow `require`/`include` calls in your PHP. Defaults to `true
 ```
 Allows you to hard-code a serializer for Psalm to use when caching data. By default, Psalm uses `ext-igbinary` *if* the version is greater than or equal to 2.0.5, otherwise it defaults to PHP's built-in serializer.
 
+#### threads
+```xml
+<psalm
+        threads="[int]"
+>
+```
+Allows you to hard-code a serializer for Psalm to use when caching data. By default, Psalm uses `ext-igbinary` *if* the version is greater than or equal to 2.0.5, otherwise it defaults to PHP's built-in serializer.
```
orklah, line 405: this should be updated

M1ke: Oops sorry, missed committing that bit

M1ke: Done

M1ke: Done

M1ke:
> I'd prefer if we could use the config instead of the detection

My concern about that is that the config can then produce unexpected results. I.e. I run on my CI infrastructure usually which has 16 cores, so I configure it to use 14 of them. However one time I try to run locally and I'm honestly not sure what happens if you try and run 14 threads on a machine that has 4 cores. So then you're having to either run in debug mode (limited to 1 thread when maybe you'd prefer to use 2) or having to edit the config file (and remembering not to commit your change)

I guess we could run it in order of:

* Things that force to 1 (debug, CI)
* Command line
* Config file
* Detection

However I think we'd still want detection to be used if the config is set, to make sure it doesn't use too many threads. Which is basically back to the implementation I've done here.

M1ke:
I can see that there's one failing job but not sure how to resolve; I can see the labels section on the PR sidebar but not clear how I add one. Is that something only maintainers can do?

What do you think about how we should handle the order of precedence here?

orklah:
As far as I know, the only risk of running on too many threads is loosing performances.

Either because there is no sense trying to run two threads on the same core or because each thread need some memory and that could sometimes use all the memory the machine has (and then it's either killed by the OS or stopped by php's memory limit)

On the other hand, we know that the detection in CI is limited so it's only a question of time before we have a user that want to fix the detection in the other way...

The last job can only fixed by us, don't worry about it ;)

M1ke:
Cool, so if I modify to this order:

    * Things that force to 1 (debug, CI)
    * Command line
    * Config file
    * Detection

Would you be happy to merge?

AndrolGenhald:
> Actually I realise that exact order would be a breaking change because currently the command line takes precedence, so I'll do:
> 
>    - Command line
>    - Things that force to 1 (debug, CI)
>    - Config file
>    - Detection

Just to give my 2 cents, this is the order I would expect, and I wouldn't expect a command line specifying 50 threads to be limited to 8 because that's what Psalm detects. I can imagine a case where Psalm's runtime is bounded by some sort of latency more than by CPU time. Perhaps someone has their code on a network filesystem with unfortunately high latency, it's possible running Psalm with more threads than cores would improve runtime depending on the file access pattern.

M1ke:
Right, functionality updated per the suggestions, and all green. Need anything else from me or lgtm?

orklah:
Thanks!

M1ke:
Thanks, glad to be of service and as you've seen I've already got my second
(slightly more ambitious) one in! Psalm is great!

M1ke:
@orklah is there a way to know when this makes it into a release? I can see that 4.19 covers commits before and after the date this was merged, but it doesn't seem to have been included. Or have I got my merge target wrong for this?

orklah:
@M1ke your PR targeted the master branch, who is our development branch for the next major version of Psalm (so Psalm 5).

We should have released a first alpha a few days ago but we're kinda late on that. The real version will be released in a few weeks

Your PR didn't contain BC break so it could theoretically have targeted 4.x...

M1ke:
@orklah ok thanks; if I were to submit the same commits as a PR onto 4.x would you be happy to merge?

orklah:
@M1ke I have no objection, but I also have very little experience with git :p

@weirdan what is the standard procedure in this case?

M1ke:
I can attempt a rebase - depends what's changed in the underlying code but I don't think I've touched much that seems to be modified heavily in 5.x

If I can't rebase, I'll write the same content onto 4.x and submit that as a PR. Once merged I can open a PR from 4.x to master which will resolve potential merge conflicts.

weirdan:
> what is the standard procedure in this case?

Rebase and another PR should do the trick, I think.

PR #7443, opened 2022-01-20
fix proc_open stub for php >= 8.0

brainlock:
https://psalm.dev/r/c0ebca79f2

The argument names are defined here:

https://github.com/php/php-src/blob/PHP-8.0.0/ext/standard/basic_functions_arginfo.h#L1773-L1780


orklah:
Thanks! Could you fix param names back to Callmap_historical instead of creating a delta on 8.0? Param names were rarely defined before and it didn't matter anyway so it's better to have some consistency across versions

brainlock:
Sure! I pushed another commit which unifies the argument names across all versions, and fixes the type for the options argument (should be nullable) already in the historical map.

Now the only delta for proc_open is in the 7.4 CallMap, where its type was changed from `string` to `string|array`.

Do you prefer the two commits to be squashed?

orklah:
Thanks! That's great!

Issue #3379, opened 2020-05-16
DX: improve error messages


staabm:
> InvalidArgument: Argument 2 of rex_instance_pool_trait::getInstance expects callable(string...):object|null, Closure(mixed):(null|rex_media&static) provided

The separation of the expected value and the actual value is done by a single `,`. Its really hard to read. I would suggest something like

> InvalidArgument: Argument 2 of rex_instance_pool_trait::getInstance expects callable(string...):object|null but Closure(mixed):(null|rex_media&static) provided

mr-feek:
I do do agree the messaging could be a bit more concise, however I don't think `:` vs. `,` makes much of a difference. I've been contemplating another output format that's more visual based, but don't know exactly what it would look like yet.

The visualizations of the code snippets helps a lot, have you tried enabling them? `--show-snippet=true`

weirdan:
Snippets are nice, but messages need to readable too, for contexts like LSP diagnostics.

weirdan:
Upper case words could provide easy visual anchors

> InvalidArgument: Argument 2 of rex_instance_pool_trait::getInstance EXPECTS callable(string...):object|null GOT Closure(mixed):(null|rex_media&static) instead

mr-feek:
I do agree that might help!

staabm:
> I do do agree the messaging could be a bit more concise, however I don't think `:` vs. `,` makes much of a difference. I've been contemplating another output format that's more visual based, but don't know exactly what it would look like yet.


In my example I changed `,` for the word `but`..

But I agree, as mentioned earlier, there will be better alternatives. 

I also thought about a `diff` formatted style of actual vs expected types

mr-feek:
I think some sort of diff style could be great -- that's what i had in mind.

Also, for the likes of this:

https://psalm.dev/r/1a5fd559cf

Maybe some sort of error message like "does not actually return declared null type" or something could be simpler and easier to understand.

weirdan:
> `does not actually return declared null type`

I think messages for any type contradiction should actually spell out types fully, in predictable format. Psalm actually has those special cases for `null` & `false` in some places, and I always find myself wandering if I passed `?A` where `B` was expected when Psalm in cases like this:
https://psalm.dev/r/b3beff7046

githoober:
Every time I get these errors, I get very discouraged to investigate them as it involves copy-paste them elsewhere and unwrap them to be more easily deciphered
[...]

PR #7445, opened 2022-01-20
Array-like-object comparison message improvement

### LINKED ISSUE #7050: [feature request] Help identify which part of some types are incompatible
When array-objects with more than a handful of keys are incompatible, it can be hard to figure what when wrong, eg:

https://psalm.dev/r/ad7136740e

in the previous exemple, the current error message is :

> ERROR: InvalidArgument - 10:3 - Argument 1 of f expects array{a: int, b: string, c: array<array-key, bool>|null, d: stdClass, e: list<string>}, array{a: 1, b: "asd", c: array{0}, d: stdClass, e: array<empty, empty>} provided

suggestion:

> ERROR: InvalidArgument - 10:3 - Argument 1 of f expects array{c: array<array-key, bool>|null, ...}, array{c: array{0}, ...} provided

M1ke:
I am opening this as a draft because I am unsure how we'd desire the implementation to work, but I feel it's helpful and it turned out not too hard to implement a basic case.

When dealing with a large object-like array it can be difficult to work out which keys are missing, e.g.

> Argument 1 of myFunction() expects array{chicken: bool, fish: bool, pasta: bool, bread: bool, bacon: bool, tea: bool, coffee: bool, beer: bool, onions: bool}, array{chicken: bool, fish: bool, pasta: bool, breead: bool, bacon: bool, tea: bool, coffee: bool, beer: bool, onions: bool} provided (see https://psalm.dev/193)

In this case the eagle-eyed may spot that the second argument has misspelled `bread` as `breead`.

The PR as currently implemented would augment this report as so:

> Argument 1 of myFunction() expects array{chicken: bool, fish: bool, pasta: bool, bread: bool, bacon: bool, tea: bool, coffee: bool, beer: bool, onions: bool}, array{chicken: bool, fish: bool, pasta: bool, breead: bool, bacon: bool, tea: bool, coffee: bool, beer: bool, onions: bool} provided. The differences are in the following keys: bread, breead (see https://psalm.dev/193)

This allows a simple check on the result that is much easier to track.

Currently it is implemented for ArgumentAnalyzer with PossiblyInvalidArgument and InvalidArgument. It will also likely benefit being added to ArgumentTypeCoercion and possibly others - we can discuss that if the basic premise seems worthwhile. I thought I'd open now to allow discussion and to see if this affects any parts of the CI build.

Will fix #7050

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
@@ -60,9 +60,12 @@ use Psalm\Type\Atomic\TLiteralString;
 use Psalm\Type\Atomic\TNamedObject;
 use Psalm\Type\Union;
 
+use function array_diff;
+use function array_keys;
 use function array_merge;
 use function count;
 use function explode;
+use function implode;
 use function in_array;
 use function ord;
 use function preg_split;
@@ -1031,26 +1034,30 @@ class ArgumentAnalyzer
                         $statements_analyzer->getSuppressedIssues()
                     );
                 }
-            } elseif ($types_can_be_identical) {
-                IssueBuffer::maybeAdd(
-                    new PossiblyInvalidArgument(
-                        'Argument ' . ($argument_offset + 1) . $method_identifier . ' expects ' . $param_type->getId() .
-                            ', possibly different type ' . $type . ' provided',
-                        $arg_location,
-                        $cased_method_id
-                    ),
-                    $statements_analyzer->getSuppressedIssues()
-                );
             } else {
-                IssueBuffer::maybeAdd(
-                    new InvalidArgument(
-                        'Argument ' . ($argument_offset + 1) . $method_identifier . ' expects ' . $param_type->getId() .
-                            ', ' . $type . ' provided',
-                        $arg_location,
-                        $cased_method_id
-                    ),
-                    $statements_analyzer->getSuppressedIssues()
-                );
+                $extension = self::extendKeyedArrayComparison($param_type, $input_type);
+
+                if ($types_can_be_identical) {
+                    IssueBuffer::maybeAdd(
+                        new PossiblyInvalidArgument(
+                            'Argument '.($argument_offset+1).$method_identifier.' expects '.$param_type->getId().
+                            ', possibly different type '.$type.' provided'.$extension,
+                            $arg_location,
+                            $cased_method_id
+                        ),
+                        $statements_analyzer->getSuppressedIssues()
+                    );
+                } else {
+                    IssueBuffer::maybeAdd(
+                        new InvalidArgument(
+                            'Argument '.($argument_offset+1).$method_identifier.' expects '.$param_type->getId().
+                            ', '.$type.' provided'.$extension,
+                            $arg_location,
+                            $cased_method_id
+                        ),
+                        $statements_analyzer->getSuppressedIssues()
+                    );
+                }
             }
 
             return null;
@@ -1627,5 +1634,26 @@ class ArgumentAnalyzer
         }
 
         return $input_type;
+    }
+
+    private static function extendKeyedArrayComparison(Union $param_type, Union $input_type): string
+    {
+        // Default value is an empty string
+        $extension = '';
+
+        // Not sure if there's a better or more robust way to do this
+        $first_param_type = $param_type->getAtomicTypes()['array'];
+        $first_input_type = $input_type->getAtomicTypes()['array'];
+        if ($first_param_type instanceof TKeyedArray && $first_input_type instanceof TKeyedArray) {
+            // There's many ways to illustrate this but this is the simplest and provides info
+            // without being too opinionated
+            $extension .= '. The differences are in the following keys: ';
+            $param_keys = array_keys($first_param_type->properties);
+            $input_keys = array_keys($first_input_type->properties);
+            $key_comparison = array_diff($param_keys, $input_keys);
+            $extension .= implode(', ', $key_comparison);
+        }
+
+        return $extension;
     }
 }
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
@@ -1035,7 +1035,7 @@ class ArgumentAnalyzer
                     );
                 }
             } else {
-                $extension = self::extendKeyedArrayComparison($param_type, $input_type);
+                $extension = (string) $param_type->getExtendedComparisonDescription($input_type);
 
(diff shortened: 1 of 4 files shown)


AndrolGenhald:
I like the idea! I'm thinking maybe we should declare your new method as `Union::getExtendedComparisonDescription(Union $other_type): string` or something of that sort? That way it would be usable everywhere and it could be used for other things in the future (`TObjectWithProperties` certainly, and perhaps others).

I think you'll need to add a check that the `Union` is a single `Atomic`, otherwise you might be giving a bad description for passing `array{fooo: int}|int` to a function wanting `array{foo: int}`. The case for comparing `array{fooo: int}|int` to `array{foo: int}|int` is a bit more complicated, so that can probably be left unfinished for now.

M1ke:
Worth noting, I've been using my change during the last hour of my work and it's already made a few Psalm issues for array-like objects easier to debug. Definitely going to be good for workflows that involve decorating array-like objects in legacy projects.

orklah:
It's definitely a good idea. It definitely came up a few time in issues, for example among the last responses to https://github.com/vimeo/psalm/issues/3379

src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
```diff
@@ -1628,4 +1635,34 @@ private static function processTaintedness(
...
         return $input_type;
     }
+
+    private static function extendKeyedArrayComparison(Union $param_type, Union $input_type): string
+    {
+        // Default value is an empty string
+        $extension = '';
+
+        // Not sure if there's a better or more robust way to do this
+        $param_types = $param_type->getAtomicTypes();
```
orklah, line 1645: yeah, something like that:

```php
if ($param_type->isSingle()) {
    $first_param_type = $param_type->getSingleAtomic();
}
```

Then you don't need the isset on ['array'] and you can go on with the instanceof.

You may not want to try handling the Union when there is anything else in it (except maybe null), because you risk having weird results like

`... expects array{...very long list of types...}|string , possibly different type  array{...exact same list...}|int provided. The differences are in the following keys: (nothing)`

src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
```diff
@@ -1628,4 +1635,34 @@ private static function processTaintedness(
...
+        $first_param_type = $param_type->getAtomicTypes()['array'];
+        $first_input_type = $input_type->getAtomicTypes()['array'];
+        if (!$first_param_type instanceof TKeyedArray || !$first_input_type instanceof TKeyedArray) {
+            return $extension;
+        }
+
+        // There's many ways to illustrate this but this is the simplest and provides info
+        // without being too opinionated
+        $extension .= '. The differences are in the following keys: ';
+        $param_keys = array_keys($first_param_type->properties);
```
orklah, line 1661: Sometimes, the difference comes with the fact that some keys are always defined but some are possibly undefined.

This is represented with a `?` after the key, it could be interesting to allow this feature to display that. It adds complexity, but it may be worth it.

For example:
`array{a: string, b?: int}` vs `array{a: string, b: int}` would display those keys as different `b, b?`

src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
AndrolGenhald: > You may not want to try handling the Union when there is anything else in it

@orklah That's what I said :stuck_out_tongue:

Actually, I think if the containing Union has other stuff it's probably fine, but this message should probably only be generated if the compared union is a single atomic.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
AndrolGenhald: Oh, that's a good idea as well!
I think the message needs to be careful to avoid implying that the keys mentioned are the only ones that are potentially wrong. Without testing, I think this will generate a confusing message for `array{a: string}` contained by `array{a: int, b: string}`. I would say maybe the message should be more specific and say "these keys are missing", and just ignore sealed arrays for now (ie extra keys on the child type are fine).

M1ke:
Thanks both, I've finished work for the day but I should be able to sort
these changes tomorrow and likely test it a bit more with my actual
development work. Will message when ready for another look.

M1ke:
I've just pushed to implement:

* Moved the method as a public on `Union`
* Add the same logic when discovering ArgumentTypeCoercion issues
* Handle undefined keys
* Improve logic/fix Psalm errors with regards to pulling the array key vs using `isSingle()`

I think the message logic still needs some work and I'd like to go further to discover mismatches in sub-arrays (probably just one level for now) but that second part may be best in an expansion PR as there's likely a lot more possible conditions as it fans out.

I see the new push has a few failed items so I'll watch and try and resolve those once it all completes.

M1ke:
@orklah one challenge I've had here is testing with a real codebase. What I am doing is copying my changes onto the Psalm version in the vendor directory of a real project, then finding the right error types to run in that codebase. It's handy to see it for real but makes simulating different conditions awkward - and I can't step debug it.

Is there a mechanism in the Psalm development repo to quickly run a detection of one of these error types?

M1ke:
Am adding the same for more/less specific return type though I've not had an obvious test case so can't be sure I've got the comparison the right way round.

M1ke:
Noticed that the system doesn't catch if the types don't match, just if the keys don't. Will rework to ensure it catches that.

AndrolGenhald:
@M1ke You can add a test somewhere (I'm not seeing an obviously applicable place, so maybe a new file) and debug it like `php -d xdebug.mode=debug -dxdebug.start_with_request=yes vendor/bin/phpunit tests/ArgTest.php --filter '@argumentUnpackingLiteral'`

For this particular feature it should probably be matching the error message rather than the issue type, for instance, add the following test (might need tweaked):
```
<?php
/** @psalm-param array{a: string, b: int, c?: float} $_arr */
function foo(array $_arr): void {}
foo(["a" => 1, "d" => 1]);
```
to `providerInvalidCodeParse` with `'error_message' => "The following keys are missing: b."`
This will ensure it's complaining about `"b"` being missing, but ignores `"c"` and `"d"`.

orklah:
@M1ke to test on real project, you'll see that you have a CI build named 'phar-build'. You can go on it, look for an 'Artifact' link and download a phar you can use to run Psalm with

M1ke:
Just to add a note, I am still going to put in more work on this, just need to find some time. It's definitely been helpful just the small change I've done so far, so need to expand to more use cases.

torian257x:
@M1ke  my homie lets gooo. I need this... doesnt need to be perfect imho. Something is still better than what we get currently aka "lmao figure out yourself what is wrong"

PR #7454, opened 2022-01-21
feat: Handle native intersection types

### LINKED ISSUE #6413: Native intersection types
https://psalm.dev/r/229fec5b31

https://wiki.php.net/rfc/pure-intersection-types

`nikic/php-parser` support for 8.1 is tracked here: https://github.com/nikic/PHP-Parser/issues/801

petewalker:
I've raised this against 4.x, but not sure if it's for 4.x or master?

Adds native intersection type handling to psalm, removing the previous `UnexpectedValueException`.

Where an intersection is found in the parse tree, the types are resolved using the existing `Type::intersectUnionTypes` function, which I assume is being used when they're encountered in the existing docblock annotations.

I've added a handful of tests to cover this, but they're certainly not exhaustive. Are there any specific edge cases I should target?

This change feels way too simple... so apologies if I've missed something fundamental..!

Resolves #6413


AndrolGenhald:
I would add a test to make sure `A&(B|C)` is never allowed in PHP type signatures, it probably already works correctly but we definitely want to catch that and make sure it doesn't break in the future.

petewalker:
>I would add a test to make sure `A&(B|C)` is never allowed in PHP type signatures, it probably already works correctly but we definitely want to catch that and make sure it doesn't break in the future.

👍 This triggers a `ParseError` currently anyway

>Also add a test to make sure `A&B` is caught for PHP < 8.1 (looks like that one might need tweaking to make it work).

I can try to do this, but there doesn't seem to be a similar rule for `A|B` in PHP < 8.0? What should do psalm throw in these cases?

orklah:
Please target on master, We try to target new features on master to avoid future maintenance on old branches.

petewalker:
I've retargeted to `master` and added some extra checks for unions in PHP < 8.0 and intersections in PHP < 8.1.

It's meant allowing `TypeHintResolver` to raise issues to the `IssueBuffer`. I'm not sure if that's desirable or not..!

AndrolGenhald:
> I can try to do this, but there doesn't seem to be a similar rule for `A|B` in PHP < 8.0? What should do psalm throw in these cases?

Well that's not great.

> ...added some extra checks for unions in PHP < 8.0 and intersections in PHP < 8.1

That'll be good to have!

petewalker:
Woops... well, that broke a lot of stuff - more to do..!

AndrolGenhald:
@petewalker Try `IssueBuffer::maybeAdd`, that might solve it.

petewalker:
@orklah @AndrolGenhald - I _think_ it's all good to go now

petewalker:
I'm not sure that failing test has anything to do with my change? Unless I'm missing something.

src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php
```diff
@@ -55,6 +74,52 @@ public static function resolve(
             return $type;
         }
 
+        if ($hint instanceof PhpParser\Node\IntersectionType) {
+            $type = null;
+
+            if (!$hint->types) {
+                throw new UnexpectedValueException('bad');
```
orklah, line 81: Please give a proper text to the Exception. I know those are not supposed to happen, but when they end up happening anyway, it's tiresome to find the correct one when it just says "bad" :)

petewalker: That's called lazy copy-pasting! Will do.

tests/Template/ClassTemplateExtendsTest.php
```diff
@@ -4545,6 +4545,8 @@ class Secondary {}
                     '$c3' => 'c3<Secondary, RealE>',
                     '$resultC3' => 'RealE|Secondary',
                 ],
+                'ignored_issues' => [],
+                'php_version' => '8.0',
```
orklah, line 4549: oh, weird, I'd have expected promoted properties to be flagged when not in PHP 8...

petewalker: I didn't look into what this specific test was testing, TBH - I just added PHP 8 because it has a native union type hint.

orklah: yeah, I was checking why this was necessary and I saw promoted properties too, it was just a remark :p

orklah:
Thanks! That looks great!

Issue #4468, opened 2020-11-02
Unexpected `InvalidArgument` with object union


leighman:
https://psalm.dev/r/1c82b5d031

I would expect to be able to pass `Thing` to `foo` but only access things available on the union from inside `foo`.
Am I bringing TS expectations where I shouldn't?

weirdan:
No, the expectations look reasonable.

orklah:
There seem to be two issues here.

Firstly, Thing with an `@property int $id` is not enough to satisfy `object{id: int}`. Only a declared property works: https://psalm.dev/r/4801c66e51

Secondly, the combination of `Thing|object{id:int}` is `object{id: int}` which is very wrong: https://psalm.dev/r/d9a2326d30

Once those two issues fixed, I think this should work

PR #7461, opened 2022-01-22
Take pseudo properties into account for comparing TObjectWithProperties with a TNamedObject

orklah:
This will fix the first issue from https://github.com/vimeo/psalm/issues/4468

EDIT: this doesn't work... yet


orklah:
I'll close this. It's completely broken, my solution is wrong. I have to rethink this

Issue #5849, opened 2021-05-28
Conditional taint escaping not considered


ohader:
https://psalm.dev/r/0d37034313

* code-wise conditional branch not recognized
* thus `htmlspecialchars` is not considered

https://psalm.dev/r/aad2907670

* conditional `@psalm-taint-escape` not resolved correctly
* when used in a chain `first(args) -> process(args, cond)`, escaping is not considered

orklah:
You can now use this syntax: https://psalm.dev/r/4d29ff4c63

orklah:
Oh I get it. I'll reopen since it's a valid issue but I doubt we'll have the resources to work on it unless a contributor is motivated :)

Issue #5101, opened 2021-01-24
Streamline unserialized tainted results


ohader:
* `unserialize($payload)`
  + unsafe, obvious
* `unserialize($payload, ['allowed_classes' => false])`
  + safe, however results in tainted issue
* `unserialize($payload, ['allowed_classes' => [Custom::class]])`
  + more tricky
  + either introduce new annotation e.g. `@psalm-taint-config unserialize allowed_classes`
  + and/or leave it to some custom plugin handling it, by analyzing provided AST part

https://psalm.dev/r/39fff64acc

PR #7468, opened 2022-01-23
[TASK] Detect allowed_classes of unserialize function

ohader:
⚠️ please review & test this change (especially `[FEATURE] Resolve orphaned forward edges in taint graph`) carefully

---

### [TASK] Detect allowed_classes of unserialize function

* adds stub for `unserialize`, replacing call and taint sink maps
* adds new `UnserializeFunctionTainter` resolving `allowed_classes`
* adds additional test cases for `unserialize` handling

### [TASK] Consider removed taints of AddRemoveTaintsEvent directly

---

Related: #5101
Maybe: #5849

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -15120,7 +15120,6 @@ return [
 'unlink' => ['bool', 'filename'=>'string', 'context='=>'resource'],
 'unpack' => ['array|false', 'format'=>'string', 'string'=>'string', 'offset='=>'int'],
 'unregister_tick_function' => ['void', 'callback'=>'callable'],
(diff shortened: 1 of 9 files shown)


ohader:
Wow! I would have expected much more tests to fail - basically due to new `OrphanForwardEdgesWorker`... so either it's "okay" or tests just are not covering more complex taint analysis scenarios.

In any way, while working on the taint flow graph, I realized it really would be worth documenting and explaining the information there. I might take care of that, however I'm not 100% sure whether I understood all concepts and implications yet.

orklah:
Yeah, any documentation is good. I'd even say, bad documentation is sometimes better than no documentation at all.

weirdan:
CS nitpick: please use `$snake_case` for variables and properties.

Issue #7469, opened 2022-01-23
Incorrect type inference with is_array, array_is_list negation in condition


Riimu:
It seem psalm incorrectly infers array type in some cases. The following kind of line produces an unexpected error in psalm:

```php
if (!is_array($value) || !array_is_list($value)) {
```

The produced error:

```
ERROR: TypeDoesNotContainType - 5:30 - Type array<array-key, mixed> for $value is always !list<mixed>
```

Reproduction here: https://psalm.dev/r/63f2cb0d8e

Not entirely sure about the root cause of this one (so may be duplicate of existing issue). It seems however, that inverse of the condition [does not produce an error](https://psalm.dev/r/99aba2a7ab).

PR #7470, opened 2022-01-23
reconcile two arrays by intersecting them

orklah:
This should fix https://github.com/vimeo/psalm/issues/7469

This happened because there wasn't a dedicated code to handle an array existing type along with an array assertion

And given the two share the same getKey() value, it was weirdly handled.

This adds some specific code.

Not completely related to this PR: sometimes, we'll have to wonder what does TypeCombiner::combine really does. It's not the result of an union nor a result of an intersection. It kinda looks like the `refineTypeWithAnother` used somewhere else in the assertion module. I think the name is way better than `combine` and the way it works seems to be better too

### DIFF
--- a/src/Psalm/Internal/Type/NegatedAssertionReconciler.php
+++ b/src/Psalm/Internal/Type/NegatedAssertionReconciler.php
@@ -23,6 +23,8 @@ use Psalm\Type\Atomic\TFalse;
 use Psalm\Type\Atomic\TFloat;
 use Psalm\Type\Atomic\TInt;
 use Psalm\Type\Atomic\TIterable;
+use Psalm\Type\Atomic\TKeyedArray;
+use Psalm\Type\Atomic\TList;
 use Psalm\Type\Atomic\TLiteralFloat;
 use Psalm\Type\Atomic\TLiteralInt;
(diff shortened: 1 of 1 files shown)


AndrolGenhald:
> It's not the result of an union nor a result of an intersection.

If you're referring to combining `array{a: int}` and `array{b: int}` = `array{a: int, b: int}`, that actually is an intersection for unsealed arrays.

I know we haven't decided how to handle sealed arrays yet, but I've recently thought about implementing [mapped types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) for arrays, so an unsealed array in Psalm 4 would be represented as `array{a: int}`, but in Psalm 5 (or maybe 6 at this point) it would be `array{a: int, [array-key]: mixed}`. This makes it much more intuitively obvious that `array{a: int, [array-key]: mixed} & array{b: int, [array-key]: mixed} = array{a: int, b: int, [array-key]: mixed}`, and it also allows for a lot of other cool things.

orklah:
I think the syntax will need to be discussed with Ondrej, we don't want to create two standard on something this important I think

AndrolGenhald:
I'll go comment on 5299, do we have somewhere else we're already coordinating with Ondrej or should I ping him?

Issue #7244, opened 2021-12-30
Missing support for templates resolution in closures/callables


veewee:
* See https://psalm.dev/r/0df3da05ef
* Additional context : https://github.com/vimeo/psalm/discussions/7212

The templates on closures and callables are not resolved into the actual provided types when a `FuncCall` is performed on it.

This results in following issues:

Output:
```
INFO: Trace - 9:32 - $stages: pure-Closure(Closure(T:fn-a::pipe as mixed):T:fn-a::pipe as mixed...):Closure(T:fn-a::pipe as mixed):T:fn-a::pipe as mixed

ERROR: InvalidArgument - 10:21 - Argument 1 expects Closure(T:fn-a::pipe as mixed):T:fn-a::pipe as mixed, "hello" provided

INFO: Trace - 11:30 - $_res: Closure(T:fn-a::pipe as mixed):T:fn-a::pipe as mixed
```

Example code:
```php
<?php
/**
 * @psalm-suppress ForbiddenCode
 */
function test(): void
{
    $a = [A::class , 'pipe'];
    $stages = Closure::fromCallable($a);
    /** @psalm-trace $stages */;
    $_res = $stages('hello');
    /** @psalm-trace $_res */;
}

class A
{
    /**
     * @template T
     *
     * @param Closure(T): T ...$stages
     *
     * @return Closure(T): T
     *
     * @pure
     */
    function pipe(...$stages): callable
    {
        return array_pop($stages);
    }
}
test();
```

orklah:
I think my code example with callable was wrong, here's the fix: https://psalm.dev/r/75a4ef108e

But the conclusion is the same...

veewee:
@orklah 
This problem also applies to first class callable syntax. Which is a bit annoying, since you'dd expect it to resolve the template types.

Is there any entrypoint in code I can look at to check if I can create a fix for this issue?
Or would it be quite complex to implement?


```php
<?php

/**
 * @template T
 * @param T $i
 * @return T
 */
function debug(mixed $i): mixed {
	return $i;
}


$x = debug('hello');
$y = debug(...)($x);


/** @psalm-trace $y */
```

orklah:
I'm not sure :(

But I'd guess the first thing to begin would be to make sure templates are kept through closures:
https://psalm.dev/r/01e2acbe39

This should not be mixed if we want this to work

veewee:
@klimick  : would it (theoretically) be possible to apply the higher order logic for first class callables as well?
[...]

Issue #5471, opened 2021-03-24
polymorphic callables


marcosh:
I would like to define, accept as parameters and return polymorphic functions like

```php
/**
 * @template A
 * @var callable(A): A $id
 */
$id = fn($x) => $x
```

If I try to provide them with types using parameters, I get errors (see https://psalm.dev/r/35c4a4d34e for some examples). It looks like Psalm is not recognising the `@template` declaration

weirdan:
I don't think the example you provided could be made sound. If you instantiate `Foo` with `callable(string):string`, `baz()` will certainly fail.

muglug:
Hack doesn't support templated closures, and I don't think there's much value in Psalm supporting them either. Opens up a whole can of worms.

AndrolGenhald:
`@template` can only be used on classes and functions.
This shows the issue @weirdan pointed out more clearly: https://psalm.dev/r/1d705e3bee

marcosh:
Thanks for your replies.

My snippets actually raised two issues:
- passing a polymorphic function as a parameter, which you discussed in your answers. My example makes sense if you see `$bar` as a function from `A` to `A` for any possible type `A`. These are what are usually called existential types. For example in Haskell you could use them as follows

```haskell
data Poly = Poly (forall a. a -> a)
```

The only value with type `forall a. a -> a` is the identity function.

- the second issue is typing anonymous functions. As per the documentation

```php
/**
 * @template T
 * @psalm-param T $t
 * @return T
 */
function mirror($t) {
    return $t;
}
```

works. Still, if I define the same function as an anonymous functions

```php
/**
 * @template T
 * @var callable(T): T $f
 * @param T $t
 * @return T
 */
$f = fn($t) => $t;
```

the template `T` is not recognised correctly

weirdan:
For the latter, you need docblock to immediately precede the closure: https://psalm.dev/r/0f93b31944. Note, however, that Psalm does not support templated closures as @muglug noted above.

PR #7471, opened 2022-01-23
Function dynamic storage provider

### LINKED ISSUE #7450: `FunctionStorage` was not initialized with `FunctionParamsProviderInterface`
https://github.com/klimick/psalm/commit/1d988add4e765f1e723c313fc574249bd12ba0b1

After #7417 was merged I tried to make a plugin for variadic pipe function (This may look tricky).

It would even work but `FunctionStorage` is not created for functions that are processed by `FunctionParamsProviderInterface`.
Without `FunctionStorage` will not created `TemplateResult` that matters for argument types inference.

After a "quick fix" I caught failing tests.
Can you answer the question why is `FunctionStorage` is not created?

klimick:
Closes #7450
New plugin hook. It allows to create `FunctionStorage` for hard to infer functions by Psalm.
I've added test for `custom_array_map`. `custom_array_map` is similar to native `array_map`.
Of course plugin hook don't cover all cases that covered by Psalm for `array_map`. But it possible. 
That means if we improve the hook, we can remove the weird code inside Psalm for array_map or array_filter.

With new hook we can create `FunctionStorage` for `pipe` or `curry` or `partial` and many other functions from functional world.
I can share plugin hooks for `pipe` and `partialRight` (without hardcoded templates as in #7470) if new hook is legit.
@veewee @orklah


tests/Config/Plugin/Hook/ArrayMapStorageProvider.php
```diff
@@ -0,0 +1,190 @@
...
+        $args_count = count($call_args);
+        $expected_callable_args_count = $args_count - 1;
+
+        if ($expected_callable_args_count < 1) {
+            return null;
+        }
+
+        $last_array_arg = $call_args[$args_count - 1];
+
+        if (ExpressionAnalyzer::analyze($statements_analyzer, $last_array_arg->value, $context) === false) {
```
veewee, line 44: Are the internal classes supposed to be used inside 3rd party plugins?
Because this would be the only way to fetch the datatype on the next line after the if statement.

tests/Config/Plugin/Hook/ArrayMapStorageProvider.php
```diff
@@ -0,0 +1,190 @@
...
+        $input_template = self::createTemplate('TIn');
+
+        // Template type that will be inferred via TemplateInferredTypeReplacer
+        $value_type = new Union([$input_template]);
+
+        $templated_array = new Union([
+            new Type\Atomic\TArray([Type::getArrayKey(), $value_type])
+        ]);
+
+        $template_result = new TemplateResult(
```
veewee, line 104: Same question here: most template methods being used here are in the internal section.
Maybe a good idea to create a wrapper class / function in the plugin or other namespace?

veewee (commented):
Nice!
This opens up a lot of flexibility for those variadics generic cases!

This pr still contains the preanalysed version of the function params hook. Is it the idea to keep it there since we can now use this new hook?

tests/Config/Plugin/Hook/ArrayMapStorageProvider.php
klimick: In the plugins that I wrote before, I was not bothered by this. Of course, it may be that the internal api will change, but I have never had this happen. Maybe we can think about exposing some internal api to pulbic api, but I don't know how to approach it.

tests/Config/Plugin/Hook/ArrayMapStorageProvider.php
klimick: The same. I don't know how it would be better to expose the internal API. Maybe @orklah can assist here.

klimick:
> This pr still contains the preanalysed version of the function params hook. Is it the idea to keep it there since we can now use this new hook?

I would suggest to remove old hook in favor new one.
In my opinion old hook is unuseful against new hook. New hook can do the same and more.

drupol:
Guys, it's amazing to read your comments and the proactivity ! Even if I don't understand anything, I'm pretty sure you're going to make something great that I really want to use ASAP !

tests/Config/Plugin/Hook/ArrayMapStorageProvider.php
orklah: Personally I find the internal API pretty messy but I'm not sure either how to expose it in a way that would make sense. I guess we would start by listing every relevant resource to expose... It's been a while since I tweaked in plugins!

orklah:
> I would suggest to remove old hook in favor new one.

Would old plugins be easy to migrate? If we do it, it's the right moment, we just removed the old plugins interfaces in Psalm 5, it's possible a lot of plugins didn't migrate already

veewee:
> > I would suggest to remove old hook in favor new one.
> 
> Would old plugins be easy to migrate? If we do it, it's the right moment, we just removed the old plugins interfaces in Psalm 5, it's possible a lot of plugins didn't migrate already

It depends which ones are gonna be removed.
The function return type hooks are being used quite often in our plugin and still make sense.
The one for function parameters I don't think is being used commonly. Because you can't really do that much with it at this moment. But I can only speak for my own plugins :)

klimick:
> It depends which ones are gonna be removed.

I suggest to remove `FunctionParamsProviderInterface` and `MethodParamsProviderInterface`.
As I understand: It provides dynamic signature using raw arg nodes.
With `(Function|Method)ParamsProviderInterface` + `(Function|Method)ReturnTypeProviderInterface` we can organize function that return type depends from input args. But it can't provide templates. It can't cover hard to infer functions (native array_map, pipe, curry and other).

On the other side, new hook is very low level. Writing plugin for `custom_array_map` was not easy.
For functions without template params we can leave `FunctionParamsProviderInterface` and `MethodParamsProviderInterface`.

veewee:
> > It depends which ones are gonna be removed.
> 
> I suggest to remove `FunctionParamsProviderInterface` and `MethodParamsProviderInterface`. As I understand: It provides dynamic signature using raw arg nodes. With `(Function|Method)ParamsProviderInterface` + `(Function|Method)ReturnTypeProviderInterface` we can organize function that return type depends from input args. But it can't provide templates. It can't cover hard to infer functions (native array_map, pipe, curry and other).
> 
> On the other side, new hook is very low level. Writing plugin for `custom_array_map` was not easy as it might seem. For functions without template params we can leave `FunctionParamsProviderInterface` and `MethodParamsProviderInterface`.

It might make sense to keep them, for making the upgrade traject easier for maintainers as well.
It gives integrations the flexibility of either taking the more simple plugin approach or have full control with the dynamic one.

orklah:
Ok then let's keep other hooks for now.

Before I dive too deep in the code, I have a few questions:
- does it solves the issue that was created by calling ArgumentsAnalyzer twice?
- can you add some documentation for the new plugin hook and specify when to use the old ones and the new ones
- Is ArrayMapStorageProvider just for test purposes? Is it normal it targets `custom_array_map` only?

klimick:
> * does it solves the issue that was created by calling ArgumentsAnalyzer twice?

I've removed double args analysis just now. This can lead to many problems. If you have not yet understood what the problems are, I can try to describe them.

> * can you add some documentation for the new plugin hook and specify when to use the old ones and the new ones

Of course I can. When PR will ready.
I think we should have hook like this but for class methods. i.e. `MethodDynamicStorageProvider`. What do you think?

> * Is ArrayMapStorageProvider just for test purposes? Is it normal it targets `custom_array_map` only?

Currently just for test. I planned to use it as example for documentation.
If you count it too complicated example I will try to write easy one.

klimick:
I removed internal classes usage and tried to make a public api instead.
- `ArgTypeInferer`: can use we need to infer argument type
- `DynamicTemplateProvider`: template creator. Not all plugin users know about `$defining_class` and what value it should be. And hook can works with many function that enumerated in `getFunctionIds`. It's a little bit difficult pass the right value to 'TTemplateParam'. So creator can create well formed `TTemplateParam`.
- `DynamicFunctionStorage`: I've make it because in the `FunctionStorage` there are useless properties for hook (at my view point)

What do you think? @veewee @orklah

weirdan:
> Before I dive too deep in the code, I have a few questions [...]

Also, how does it interact with cache?

src/Psalm/Plugin/ArgTypeInferer.php
```diff
@@ -0,0 +1,36 @@
...
+        $this->context = $context;
+        $this->statements_analyzer = $statements_analyzer;
+    }
+
+    /**
+     * @return false|Union
+     */
+    public function infer(PhpParser\Node\Arg $arg)
+    {
+        if (ExpressionAnalyzer::analyze($this->statements_analyzer, $arg->value, $this->context) === false) {
```
veewee, line 30: It might be possible that the expression was already analysed - for example inside the `FunctionReturnTypeProvider` callbacks.

Might it be a better idea to first check if the statement analyzer already contains the argument before analysing it?

tests/Config/Plugin/Hook/CustomArrayMapStorageProvider.php
```diff
@@ -0,0 +1,177 @@
...
+    public static function getFunctionIds(): array
+    {
+        return ['custom_array_map'];
+    }
+
+    public static function getFunctionStorage(FunctionDynamicStorageProviderEvent $event): ?DynamicFunctionStorage
+    {
+        $func_call = $event->getExpr();
+
+        if ($func_call->isFirstClassCallable()) {
```
veewee, line 31: This must probably be included in every plugin?
Would it make sense to run any dynamic function providers for first class callables?

tests/Config/Plugin/Hook/CustomArrayMapStorageProvider.php
```diff
@@ -0,0 +1,177 @@
...
+            return null;
+        }
+
+        $template_provider = $event->getTemplateProvider();
```
veewee, line 35: loving these new methods !!! ;)

klimick:
> Also, how does it interact with cache?

I didn't quite understand what you mean.

Dynamically created storage saves only in the `FunctionDynamicStorageProvider` in `$dynamic_storages`.
Because there is cases where `FunctionDynamicStorageProvider` should be called twice for the same expression.
At the second call `FunctionDynamicStorageProvider` returns already computed storage.

As I know Psalm treats each closure as unique and create for them individual storage. Does Psalm cache it?
Dynamically analyzed function has next id:
```php
$dynamic_storage_id = strtolower($statements_analyzer->getFilePath())
            . ':' . $stmt->getLine()
            . ':' . (int)$stmt->getAttribute('startFilePos')
            . ':dynamic-storage'
            . ':-:' . strtolower($function_id);
```
Id from `FunctionDynamicStorageProvider`  has similar to closure id.

These two calls will be analyzed individually and each call will have their id:
```php
custom_array_map(
   fn($i) => $i + 1,
   [1, 2, 3]
);
custom_array_map(
   fn($i) => (string) $i,
   fn($i) => $i + 1,
   ['1', '2', '3']
);
```
There is sense to cache it? And if there is. How?

src/Psalm/Plugin/ArgTypeInferer.php
klimick: Good catch!

tests/Config/Plugin/Hook/CustomArrayMapStorageProvider.php
klimick: > Would it make sense to run any dynamic function providers for first class callables?

What do you think of this case?

```php
/** @var list<int> */
$numbers = [1, 2, 3];

pipe(
    $numbers,
    partial_left(custom_array_map(...), fn($i) => $i + 1)
)
```
Theoretically, it is possible to handle this situation.

veewee: There are some issues here:

`custom_array_map(....)` results in a FunctionCall with first argument of `VariadicPlaceholder`.
The result of this is a closure - of which the type can be determined by the expression analyzer.
This means that the `FunctionDynamicStorageProviderEvent` is triggered for the code `custom_array_map(...)` - but you cannot really do anything on that.


The second issue here, is that there is no support for templates on closures inside psalm yet.
See https://github.com/vimeo/psalm/issues/7244

Since first class callable is just a wrapper around `Closure::fromCallable()` - it means that templates get lost.
See:
https://psalm.dev/r/01e2acbe39


If psalm were able to determine that, in this case, debug returns `Closure(T): T` instead of `Closure(mixed): mixed`, there would never be a need for adding a `FunctionDynamicStorageProviderEvent`.

Since first class callable is going to be used quite often in the future, this feels like a missing part in psalm.
Also scrolled trhough the issues and saw that it is not supported for reasons and that psalm even reports an issue if you add templates to closure docblocks:

"Templated closures are not supported"

https://psalm.dev/r/2da9ff0b52

I think with first class callables, it is a feature that should be working.
More info:

- https://github.com/vimeo/psalm/issues/5472
- https://github.com/vimeo/psalm/issues/5471#issuecomment-805830527

veewee: Ow snap .... on second thought - in this case `custom_array_map()` is dealt with by a plugin as well.
This indeed indicates that we need it to be able to determine a return type in `FunctionDynamicStorageProviderEvent`.

What would you let it return in this case?

klimick: I've removed first class callable support from hook for now.
Current issue is getting bigger and bigger. Let's deal with this in another issue.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
```diff
@@ -562,15 +579,6 @@ private static function handleNamedFunction(
             }
 
             if ($codebase->functions->params_provider->has($function_call_info->function_id)) {
-                ArgumentsAnalyzer::analyze(
```
weirdan, line 565: Are args already analysed at this point?

weirdan:
> I didn't quite understand what you mean.

I meant the file cache Psalm uses to speed up things on subsequent runs. Normally `FunctionStorage` instances are cached as part of `FileStorage`. Having looked at the proposed changes once again, I think the answer is 'it doesn't interact with cache in any way, as it's used during call analysis'. Also, it appears it would shadow the `FunctionStorage` Psalm may generate during the scanning phase. It would be nice to have this documented in the source code itself - say, in `DynamicFunctionStorageProvider` docblock.

Also, would you mind renaming all occurrences of `function dynamic storage` to `dynamic function storage`? I believe it would read better that way.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallAnalyzer.php
veewee: No, not yet .
I introduced a pre-analysis step here so that the type info of the parameters would be available for this plugin.

It was a bad idea. More context, see
https://github.com/vimeo/psalm/issues/7450

tests/Config/Plugin/Hook/CustomArrayMapStorageProvider.php
veewee: Alright, moved over this info to #7244, we can pick it up in another pr / issue.

If you don't trigger the plugin for first class callables, it will fall back to the signature in the dockblocks on the function.
Which will propbably look like this:

```php
/**
 * @param list<mixed> $args
 * @return list<mixed>
 */
function custom_array_map(...$args) {
    //...
}
```

It is ok, but not very usefull for partial function application though.
I can't come up with how you would improve the type from within the plugin - of that's even ever possible.
Because it requires the context of the parent function to see how the of will be used. 
(Basically you want to run the plugin on the result of the first class callable instead of on the function call)

weirdan (approved):

orklah:
Thanks!

drupol:
Excellent work, thanks !

veewee:
Great job!

thomasvargiu:
Will be released in v5?

orklah:
Yeah, this has been targeted on master branch, so with V5 release

PR #7473, opened 2022-01-23
remove TPositiveInt

orklah:
Cool! I finally managed to remove TPositiveInt (I'll make a deprecation on Psalm 4)!

There's just one test I had to skip. (And one I had to dumb down because the nested loops were horribe). I hope I'll be able to refine loops inference in the future to fix such cases


weirdan:
@orklah is `positive-int` still recognized in docblocks?

orklah:
@weirdan yes it is. It would be brutal and unnecessary to remove that. We could even consider adding a negative-int like phpstan

PR #7474, opened 2022-01-23
deprecate TPositiveInt

orklah:
This will deprecate TPositiveInt for removal in Psalm 5

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.9999999.9999999.9999999-dev">
+<files psalm-version="4.x-dev@64206d9c3a9723bba093beb7d1a77ee767332d8a">
   <file src="examples/TemplateChecker.php">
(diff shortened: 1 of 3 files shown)


asgrim:
Quick question on this; is this just an internal thing that's changing, or does this mean using the type `positive-int` is deprecated (e.g. `@param positive-int $someValueThatMustBePositiveInteger`)? If so, what would the migration route be? Maybe there is no sane upper bound for `int<min, max>`, so would something like `int<0, >` (or some other notation of not specifying an upper bound) be a reasonable replacement?

orklah:
Hi Asgrim,

The phpdoc type is not deprecated, it's just now an alias of `int<0, max>`. So yeah, there's no migration route to prepare. Psalm however won't ever suggest positive-int in errors or with Psalter anymore.

`max` is an undefined upper bound so it has no definite value, the same way TPositiveInt was technically not constrained to respect PHP_INT_MAX

asgrim:
I understand - thank you for explaining that, really appreciate it @orklah :metal:

PR #7476, opened 2022-01-24
Enable more checks for native classes on PHP 8

danog:
Coupled with #7475, aims to reach feature parity with [phpstan](https://phpstan.org/r/050dffdd-aaab-431a-87cc-9261a166ce9c) vs [psalm](https://psalm.dev/r/66d598df84).

This PR enables these checks only on 8.1, and like phpstan it doesn't support suppression via the ReturnTypeWillChange attribute, which could possibly cause BC problems when seeking to support PHP 7 (unlike my earlier merged PR which does support ReturnTypeWillChange, but doesn't check for things like mismatched types).

### DIFF
--- a/src/Psalm/Internal/Analyzer/MethodComparator.php
+++ b/src/Psalm/Internal/Analyzer/MethodComparator.php
@@ -203,8 +203,7 @@ class MethodComparator
             );
         }
 
-        if ($guide_classlike_storage->user_defined
-            && ($guide_classlike_storage->is_interface
+        if (($guide_classlike_storage->is_interface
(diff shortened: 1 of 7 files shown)


danog:
All done, the checks are enabled when targeting 8.0+.

orklah:
Thanks for the work. I'm afraid we have some issues about `user_defined` when using stubs (which a lot of users do) that won't allow this to be merged for now...

We had to disable your other related feature about type inheritance in the core because of that stub issue so I feat this will be compromised as well. Please keep this open, if the fix is made in the stubs, we'll be able to merge that then

PR #7484, opened 2022-01-25
Remove mic-drop hack from if analysis

muglug:
Handling of `if` conditions had a path I called "mic drop" that dealt with the following code

```php
if (<expr>) {
  <some code>
  return;
} // no elseif or else
```

While a useful shortcut, this created a bunch of special cases that don't work if an empty ` else {}` is appended to the `if` condition — many examples in `tests/TypeReconciliation/TypeAlgebraTest.php` fail when that empty `else {}` is used.

This PR contains the necessary fixes to make most everything work as expected without the mic drop hack.

### DIFF
--- a/tests/Loop/DoTest.php
+++ b/tests/Loop/DoTest.php
@@ -121,6 +121,7 @@ class DoTest extends TestCase
                             break;
                         }
 
+                        /** @psalm-suppress MixedArgument */
                         foo($a);
                     }
                     while (rand(0,100) === 10);',
--- a/src/Psalm/Context.php
+++ b/src/Psalm/Context.php
@@ -349,7 +349,7 @@ class Context
     /**
      * @var Context|null
      */
-    public $if_context;
+    public $if_body_context;
 
     /**
      * @var bool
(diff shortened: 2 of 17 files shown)


muglug:
The endtoend test failure with PHPUnit is, I think, a true positive. I'll update the baseline in that repo.

tests/Loop/DoTest.php
```diff
@@ -121,6 +121,7 @@ function foo(string $b) {}
                             break;
                         }
 
+                        /** @psalm-suppress MixedArgument */
```
orklah, line 124: It makes $a mixed here?

It seemed correctly inferred to `hello` before

orklah:
Seems great!

While you're in this code, I'd be grateful if you could leave a little comment to explain the role of the different scope and contexts objects for analyzing the if().

I tried to check this code a few times but I get lost with the if_context, outer_scope, post_if_context etc...

tests/Loop/DoTest.php
muglug: Yeah, unfortunately that was dodgy too: https://psalm.dev/r/a1a5253dc3. Since assignments in conditionals are a known pain, I thought this was a reasonable compromise (there are many assignment in conditional tests, and this was the only one affected).

orklah: Seems reasonable

orklah:
Thanks!

PR #7486, opened 2022-01-25
Add missing parameter to Phar::getMetadata()

villfa:
Hi,

An new parameter `$unserializeOptions` has been added in PHP 8.0 to `Phar::getMetadata()` and `PharFileInfo::getMetadata()`  (see https://github.com/php/php-src/commit/0c238ede0).

Documentation:
* https://www.php.net/manual/en/phar.getmetadata.php
* https://www.php.net/manual/en/pharfileinfo.getmetadata.php


orklah:
Thanks!

PR #7487, opened 2022-01-25
fix error with 0 being considered positive

orklah:
fix a bug introduced in https://github.com/vimeo/psalm/pull/7473. When combining ints, we were flagging 0, using TPositiveInt and then check if TPositiveInt to add 0 to the combination. When I removed TPositiveInt, I wrongly assumed this had become unused code

### DIFF
--- a/tests/IntRangeTest.php
+++ b/tests/IntRangeTest.php
@@ -704,6 +704,18 @@ class IntRangeTest extends TestCase
                     '$_arr===' => 'non-empty-array<int<0, max>, int<0, max>>',
                 ],
             ],
+            'noErrorPushingBigShapeIntoConstant' => [
+                'code' => '<?php
+                        class DocComment
+                            {
+                                private const PSALM_ANNOTATIONS = [
+                                    "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+                                    "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""
+                                ];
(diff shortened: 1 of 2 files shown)


tests/IntRangeTest.php
```diff
@@ -704,6 +704,18 @@ function doAnalysis(): void
...
+                        class DocComment
+                            {
+                                private const PSALM_ANNOTATIONS = [
+                                    "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+                                    "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""
+                                ];
+                            }',
+                'assertions' => [
+                ],
+            ],
```
AndrolGenhald, line 718: Does this test actually fail before the fix? I tried the same thing with the actual values from DocComment but for some reason it passed fine.

orklah: it does: https://psalm.dev/r/1529295950

orklah:
Psalm bot, please fetch this for posterity: https://psalm.dev/r/1529295950

tests/IntRangeTest.php
orklah: Although I noticed that it started failing after 52 values in there. DocComment must have around 55, so maybe you didn't include them all?

AndrolGenhald: @orklah Oh I see what happened, I used `--filter` for my test, but I forgot `paratest` doesn't support that, you have to use `phpunit` directly. I saw the success message and didn't notice it said 0 tests...

PR #7488, opened 2022-01-25
Bugfix: "TypeDoesNotContainNull: on return value from dom_import_simplexml"

### LINKED ISSUE #6151: TypeDoesNotContainNull: on return value from dom_import_simplexml()
When using the `dom_import_simplexml()` function, psalm gives a `TypeDoesNotContainNull` warning when performing a null check on the return value. Psalm still indicates the return type to be `DOMElement|false` while the actual return type is `DOMElement|null`.

Example: https://psalm.dev/r/42bcd39f32

KevinVanSonsbeek:
Fixes #6151


KevinVanSonsbeek:
I seem to have branched from the wrong branch.

PR #7489, opened 2022-01-25
Bugfix/#6151 typedoesnotcontainnull on return value from dom import simplexml

### LINKED ISSUE #6151: TypeDoesNotContainNull: on return value from dom_import_simplexml()

KevinVanSonsbeek:
Fixes #6151

### DIFF
--- a/dictionaries/CallMap_80_delta.php
+++ b/dictionaries/CallMap_80_delta.php
@@ -310,7 +310,7 @@ return [
       'new' => ['string', 'object'=>'DateTimeInterface', 'format'=>'string'],
     ],
     'dom_import_simplexml' => [
-        'old' => ['DOMElement|false', 'node'=>'SimpleXMLElement'],
+        'old' => ['DOMElement|null', 'node'=>'SimpleXMLElement'],
         'new' => ['DOMElement', 'node'=>'SimpleXMLElement'],
     ],
     'explode' => [
--- a/dictionaries/CallMap_80_delta.php
+++ b/dictionaries/CallMap_80_delta.php
@@ -309,6 +309,10 @@ return [
       'old' => ['string|false', 'object'=>'DateTimeInterface', 'format'=>'string'],
       'new' => ['string', 'object'=>'DateTimeInterface', 'format'=>'string'],
     ],
+    'dom_import_simplexml' => [
+        'old' => ['DOMElement|false', 'node'=>'SimpleXMLElement'],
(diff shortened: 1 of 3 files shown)


orklah:
Thanks!

I believe the false was not possible even before PHP 8, Am I mistaken on this?

KevinVanSonsbeek:
> I believe the false was not possible even before PHP 8, Am I mistaken on this?

That actually is true yeah. I'll change the callmaps when i get home later today.

dictionaries/CallMap_80_delta.php
```diff
@@ -309,6 +309,10 @@
       'old' => ['string|false', 'object'=>'DateTimeInterface', 'format'=>'string'],
       'new' => ['string', 'object'=>'DateTimeInterface', 'format'=>'string'],
     ],
+    'dom_import_simplexml' => [
+        'old' => ['DOMElement|null', 'node'=>'SimpleXMLElement'],
+        'new' => ['DOMElement', 'node'=>'SimpleXMLElement'],
+    ],
```
AndrolGenhald, line 315: Was a null return possible in PHP < 8.0 when using correct parameters? If the only time null is returned is because the parameters are invalid and Psalm correctly catches that we can exclude the null return value (it's basically there because PHP before 8.0 was really lenient on some function calls and returned null instead of throwing an error for incorrect arguments).

KevinVanSonsbeek: From looking at the old source code, it could return null if the supplied nodetype was invalid. (This behavior was changed in https://github.com/php/php-src/commit/8fef83dd3ceb6105f40ab4b0748a5360f1d6adf2)

```php
if (nodep && nodeobj && (nodep->type == XML_ELEMENT_NODE || nodep->type == XML_ATTRIBUTE_NODE)) {
	DOM_RET_OBJ((xmlNodePtr) nodep, &ret, (dom_object *)nodeobj);
} else {
	php_error_docref(NULL, E_WARNING, "Invalid Nodetype to import");
	RETURN_NULL();
}
```

orklah: Fine by me then :)

orklah:
Thanks!

PR #7490, opened 2022-01-25
Fix crash due to unresolvable constant.

AndrolGenhald:
#7123 introduced a crash whenever the new `UnresolvableConstant` issue is supposed to be added. It wasn't caught by tests because the tests end as soon as the first issue is raised, and the issue is raised before it crashes.

Not the prettiest fix, but at least it works. I'm not super happy with it, but I can't think of a better way at the moment.


AndrolGenhald:
Oh, I forgot trailing commas are still disallowed in argument lists until PHP 8.0 :slightly_frowning_face:

AndrolGenhald:
Actually maybe it'd be better to pass an optional `CodeLocation` into expandUnion and raise the issue there? The reason I added the exception is because `expandUnion` doesn't have access to the `CodeLocation`.

orklah:
Thanks!

PR #7492, opened 2022-01-26
Ensure all template parameters are specified in all extended|used classlikes

danog:
For feature parity with phpstan.

This PR kind of renders useless the `template_extended_count` property, replacing it with the more complete `template_type_extends_count`, IDK if it's worth deprecating here and removing in v5.

PR #7494, opened 2022-01-26
Deprecate template_extended_count

danog:
As per #7492


orklah:
Please run with --set-baseline=psalm-baseline.xml to add the deprecated warnings to the baseline. You'll have to run it again on master to remove them

PR #7503, opened 2022-01-28
Fix failing case for const analyzer.

AndrolGenhald:
Adding an extra interface causes the constant storage to be set to null, so the analysis doesn't finish properly. This should fix that case, as well as enable assertions on CI runs so that failing assertions can be caught sooner.

Thanks @muglug for noticing! I'm running tests with assertions enabled locally from now on!


orklah:
Thanks!

PR #7504, opened 2022-01-28
InternalTaintSinkMap: Add `getimagesize` as SSRF sink

mal-tee:
Hi, 

I hope this was the right place. I wasn't able to figure out which sinks go into `stubs/` and which into this map. :)

Demo of the sink:

image


weirdan:
Thanks!

PR #7506, opened 2022-01-28
Fix side effect when loading config

AndrolGenhald:
Loading the config currently sets `libxml_use_internal_errors(true)` and never changes it back. This fixes that side effect and also ensures that xml errors from `loadDomDocument` get thrown as `ConfigException` rather than `RuntimeException`.


weirdan (approved):

orklah:
Oh cool :)

Issue #6871, opened 2021-11-09
Using equality to narrow down int range


still-dreaming-1:
https://psalm.dev/r/e825568c19

orklah:
This is handled in https://github.com/vimeo/psalm/blob/b7834689e53eda8bd1675ef3edebe4ff4ed548cc/src/Psalm/Internal/Type/NegatedAssertionReconciler.php#L394

The method does not check integer ranges into account. I actually tried to change that a week or two ago but as usual when I try to change things around integers, it makes loops goes crazy.

The comment I posted here is relevant https://github.com/vimeo/psalm/pull/6474#issuecomment-958198889

I should try to work around issues, but these tries takes a lot of time and it's pretty depressing to see it was for nothing when the whole CI fails

orklah:
I tried to tackle this again but hit the same wall as always. If you're curious, here's the branch:
https://github.com/orklah/psalm/tree/negatedLiterals

and the test suite:
https://github.com/orklah/psalm/runs/4169544216?check_suite_focus=true

In the test suite, there is one monstruous code that is actually extracted directly from Psalm code that would take me weeks to understand and reason about:
https://psalm.dev/r/19a33b692f

Currently, this only works because Psalm reverts everything that moves to an int or positive-int who are two very coercable types (you can put those types pretty much everywhere without Psalm complaining it's too large) but as soon as you try converting that to literal ints or int ranges, Psalm start to freak out :)

Fixing every loop inferrence issue could help with that, but I don't think I'll have the patience to do that alone and even then, I'm not even sure it is possible to compute this kind of loop without starting to actually execute the thing turn by turn in order to see what can happen

still-dreaming-1:
Thanks. I guess we will have to stick to alternative types of integer comparisons for now.

https://psalm.dev/r/fa2b700f26

Issue #6797, opened 2021-11-02
Non-negative integer assertion not working


danog:
https://psalm.dev/r/bc9c74c3d2

orklah:
This is is a limitation due to Psalm being unable to negate `lesser than 0` into `0|positive-int`. This is is the kind of issue that can be fixed by integer ranges pretty easily but they're hard to introduce because loops are pretty complex things and trying to be more precise in it is a problem so we have to go slowly

weirdan:
There are no loops here though.

orklah:
No, but removing positive-int as an assertion solve this but break loops :p

Issue #6778, opened 2021-10-29
Using if to narrow down int range


still-dreaming-1:
This issue cropped up for me in the code I work on after updating to the new release. Here is a simplified version of the issue.
https://psalm.dev/r/7db09cea09

I also tried using int ranges instead, but the issue remains.
https://psalm.dev/r/9ccc816956

It's not actually a problem for me as I was able to resolve it by simplifying the code. I'm just reporting this in case it is helpful since it seemed to work before.
https://psalm.dev/r/4f1bcd3b7f

orklah:
I tried to solve this but it's hard.

I'm not even exactly sure what is going on so I'll left some context (possibly for future - more knowledgeable - me).
Analysing `||` seems hard. Psalm somehow transform `$a === 0 || $a > 0` into three assertions at one point `!=int(0)`, `=isset`, `positive-numeric`. This is correct and it ends up in AssertionReconciler. After a fix I made and that I gonna PR, AssertionReconciler manages to reduce this to `int<1, 1>|int<1, max>` which is mostly correct. But then I lost track of what is done.
It seems at some point Psalm loose the info or replace it with less precise type

I also note PHPStan has the same exact issue: https://phpstan.org/r/1f64d5d9-7f82-4cf9-aa31-d3033a0adb49 so it may be something really deep

still-dreaming-1:
I'm not sure if this is helpful or not, but here is a similar scenario that might help you think about it a different way. I can also create a new issue for this one if you think that is better.

https://psalm.dev/r/f754e62ef2

orklah:
This is different, please create a new issue

PR #7511, opened 2022-01-28
improve literal inequality with ranges and rework GreaterThan/LessThan assertions

orklah:
I'm pretty happy with this one. I started by fixing the literal inequality reconciliation that I couldn't do before when TPositiveInt was still here. However, there was snippets I expected to be fixed by this change but weren't so I went a little deeper.

I discovered that a previous PR that introduced filtering of null value through (>/</>=/<=) operators added an isset assertion in addition to the integer comparison. However, Psalm can reconcile two assertions but struggles with negation of two assertions so it made reconciliation difficult in some cases.

I reworked the entire thing, I used the new Assertions object system as well as the two new assertions (IsGreaterThanOrEqualTo/IsLessThanOrEqualTo) to have a much nicer system (that allows me to stop considering that `>=5` is `>4`) and I introduced a function to know if the assertions filters the null value. That way I can use only one assertion and reconcile both.

It fixes https://github.com/vimeo/psalm/issues/6797
It fixes https://github.com/vimeo/psalm/issues/6871
It fixes https://github.com/vimeo/psalm/issues/6778
It fixes https://github.com/vimeo/psalm/issues/4358


orklah:
Github broke the linked issue feature it seems...

PR #7517, opened 2022-01-29
Fix empty array

orklah:
This fixes two oddities  in the code that became visible after I introduced isEmptyArray on a previous PR.

First, on ArrayFetchAnalyzer.php we were setting two variables from_empty_array and from_mixed_array after we already stated that the array was empty. Those variables had only one possible value so I removed them

Second, on SimpleNegatedAssertionReconciler.php, the original code on line 530 was `$array_atomic_type->getId() !== 'array<never, never>'` that means that the only way not to enter the next conditional was to have an empty array but the next conditional checks `$array_atomic_type instanceof TKeyedArray` which is not compatible with `array<never, never>`, making it dead code.

However, I'm not sure my fixes are optimal.

On the first, the code was clearly meant for handling more than empty arrays, so maybe I should have dropped the isEmptyArray in conditional and that would have worked better

On the second, I just placed the dead condition up in the code but I didn't even read the code or tested it, so not sure it will work right.

I'm opening this not to forget about the issue and to open a discussion if someone has hints

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php
@@ -1099,10 +1099,7 @@ class ArrayFetchAnalyzer
     ): void {
         $has_array_access = true;
 
-        if ($in_assignment
-            && $type instanceof TArray
-            && $type->isEmptyArray()
-        ) {
+        if ($in_assignment && $type instanceof TArray) {
             $from_empty_array = $type->isEmptyArray();
 
             if (count($key_values) === 1) {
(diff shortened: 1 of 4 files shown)


orklah:
After a lot of digging:
- The code on SimpleNegatedAssertionReconciler was correct. Doing it before has no negative side effect it seems. The specific test I made to enter the code shows that it's capable of seeing a redundant comparison between a never empty keyed array and `[]` but I couldn't find a way to make it emit the issue because you have to have a $code_location and I'm not sure why there's none for my test. However, the redundancy is emitted from somewhere else already in my test so it's good.

- The code in ArrayFetchAnalyzer was another story. The code was clearly prepared to handle more than what it was doing so I tried to remove the IsEmptyArray check that was restricting it. It made a few tests broke:
    - one because the constructed keyed array didn't handle class-strings so I had to replace literals by Atomics in order to convey the class-string-ness
    - one because when TypeCombiner combine two keyed arrays, it looses the fact that both of them are not empty, and also the fact that both of them had key types and value types.
    - one because the test was broken and my fix revealed it

One thing especially that I make a note here in case it cause issues in the future: https://github.com/vimeo/psalm/pull/7517/files#diff-4f029c4356416f8c892d88d24cc3e7a2d389e74a3fadc739ad0970ffedae233dR740
This code was broken. It counted the number of possibly undefined properties to fill the min_count instead of the number of definitely defined properties. So I inverted the callback

PR #7521, opened 2022-01-29
Add command to generate stubs for PHP extensions.

### LINKED ISSUE #7513: Create tool to generate stubs for an extension
Now that extensions can be supported without being loaded at runtime, to fully support an extension we'll need complete stubs for it. To aid in this effort, it would be fantastic to have a tool to generate a complete stubfile for an extension.

Stubs will often need to be modified manually after generation for additional accuracy (for instance [`PDOStatement::fetch`](https://www.php.net/manual/en/pdostatement.fetch.php) should take `PDO::FETCH_*, PDO::FETCH_ORI_*, int` instead of `int, int, int`), but this tool should simplify the task significantly.

AndrolGenhald:
Fixes #7513.

I think extension stubs will require enough manual modification that using a different stubfile for each version will require too much duplicated effort. I'm thinking we should do something like this, using xml as an example:

Add a `stubs/extensions/reflection-generated/xml` directory. Stubs under the `reflection-generated` won't ever be loaded, they're just there for diffing.

When initially adding support, run `./psalm --generate-extension-stub xml >stubs/extensions/reflection-generated/xml/xml-7.4.0.phpstub` (using the extension's version, not PHP's version, but many extensions track the PHP version). Then copy that generated stub to `stubs/extensions/xml.phpstub` and modify it as necessary.

When updating an extension to support a new version, run `./psalm --generate-extension-stub xml >stubs/extensions/reflection-generated/xml/xml-8.1.0.phpstub`. Then the reflected stubs can be diffed with `diff stubs/extensions/reflection-generated/xml/xml-{7.4.0,8.1.0}.phpstub`, and that diff can be used as a starting point for what modifications will be needed in `stubs/extensions/xml.phpstub`.

Thoughts, criticisms, improvements?

### DIFF
--- a/src/Psalm/Internal/Codebase/Reflection.php
+++ b/src/Psalm/Internal/Codebase/Reflection.php
@@ -54,11 +54,15 @@ class Reflection
      */
     private static $builtin_functions = [];
 
+    /**
+     * @var array<string, Union>
+     */
+    private static $builtin_constants = [];
+
     public function __construct(ClassLikeStorageProvider $storage_provider, Codebase $codebase)
     {
         $this->storage_provider = $storage_provider;
         $this->codebase = $codebase;
-        self::$builtin_functions = [];
     }
 
     public function registerClass(ReflectionClass $reflected_class): void
@@ -419,6 +423,16 @@ class Reflection
         return null;
     }
 
+    /**
+     * @param array|scalar|null $constant_type
+     */
+    public function registerConstant(string $constant_name, $constant_type): void
+    {
+        self::$builtin_constants[$constant_name] = new Union([
+            ConstantTypeResolver::getLiteralTypeFromScalarValue($constant_type),
+        ]);
+    }
+
     public static function getPsalmTypeFromReflectionType(?ReflectionType $reflection_type = null): Union
     {
         if (!$reflection_type) {
@@ -540,8 +554,17 @@ class Reflection
         return self::$builtin_functions;
     }
 
+    /**
+     * @return array<string, Union>
+     */
+    public function getConstants(): array
+    {
+        return self::$builtin_constants;
+    }
+
     public static function clearCache(): void
     {
         self::$builtin_functions = [];
+        self::$builtin_constants = [];
     }
 }
--- a/src/Psalm/Internal/Codebase/Reflection.php
+++ b/src/Psalm/Internal/Codebase/Reflection.php
@@ -24,6 +24,7 @@ use ReflectionNamedType;
 use ReflectionParameter;
 use ReflectionType;
 use ReflectionUnionType;
+use Throwable;
 use UnexpectedValueException;
 
 use function array_map;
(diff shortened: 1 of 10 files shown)


src/Psalm/Internal/Codebase/Reflection.php
```diff
@@ -540,8 +554,17 @@ public function getFunctions(): array
         return self::$builtin_functions;
     }
 
+    /**
+     * @return array<string, Union>
+     */
+    public function getConstants(): array
+    {
+        return self::$builtin_constants;
```
AndrolGenhald, line 562: Seems weird that this class is half static and half not.

src/Psalm/Internal/ExtensionStubGenerator/Command/GenerateExtensionStubCommand.php
```diff
@@ -0,0 +1,116 @@
...
+        // filter out parent classlikes that aren't defined by this extension.
+        $classlike_storages = array_intersect_key(
+            $providers->classlike_storage_provider->getAll(),
+            array_flip(array_map("strtolower", $extension_reflection->getClassNames())),
+        );
+
+        foreach ($extension_reflection->getFunctions() as $function_reflection) {
+            $function_name = $function_reflection->getName();
+            assert($function_name);
+            $reflection->registerFunction($function_name);
```
AndrolGenhald, line 93: `registerClass` takes a `ReflectionClass` but `registerFunction` take a string. Maybe I should fix it to take a `ReflectionFunction`?

src/Psalm/Internal/ExtensionStubGenerator/Command/GenerateExtensionStubCommand.php
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        /** @var mixed $const_type */
+        foreach ($extension_reflection->getConstants() as $const_name => $const_type) {
+            if (!is_array($const_type) && !is_scalar($const_type) && $const_type !== null) {
```
weirdan, line 97: Constants can be objects, so I believe extensions could expose such constants too and thus this check needs to be dropped.

AndrolGenhald:
I'd like to have a documented way to add support for an extension that's as simple as possible, but I also think we might end up removing this if we set up a better way to generate stubs with a separate project. 

Would you be ok merging this if I add a note [here](https://github.com/vimeo/psalm/pull/7521/files#diff-20b00de237986e634ae4c11c7c5daada342880a20c8ab7bde29e91cc92b30a04R1343) that the command is experimental and could be removed at any time? It would be nice to have this in here, but also nice to be able to remove it before Psalm 6 if our process improves.

orklah:
Sorry, didn't had time to look at this one for now. Will do soon :)

AndrolGenhald:
@orklah You can hold off for a bit, I'm trying to see how much work it is to fix that other generator I found. I realized while I was generating some stubs that this doesn't account for inheritance at all, and there are probably other issues too.

orklah:
Ok :)

Please ping me if you had specific questions. 

On the long run, it could be great to try and replace one extension stub by one generated by the generator, just to check if it's legit or if it has big issues

AndrolGenhald:
You can review the documentation I added, I'll probably keep that even if we go with a separate project for the generator. I'm currently planning to keep extra reflection-generated stubs for easy diffing, but the stub that psalm actually uses will usually have to be modified from the generated one.

AndrolGenhald:
Closing in favor of #7577.

PR #7525, opened 2022-01-30
Fix for Exception->getCode return type provider

VincentLanglet:
@orklah, I found one case when the solution we implements doesn't work.

It's when you're working with an interface extending `Throwable`.
`getCode` is supposed to return `int`, but psalm doesn't understand it because 
```
in_array(Throwable::class, $class_storage->class_implements)
```
return false.


orklah:
Thanks!

PR #7526, opened 2022-01-30
[TASK] Add AfterFunctionLikeAnalysisEvent::getFunctionlikeStorage

ohader:
AfterFunctionLikeAnalysisEvent's method `getClasslikeStorage` actually
returned the current `FunctionLikeStorage`. This change streamlines the
naming and adds corresponding `getFunctionlikeStorage` method.

Releases: 5.0

---
4.x deprecation: #7532

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -209,3 +209,5 @@
    - `Psalm\PluginRegistrationSocket::getAdditionalFileExtensions` was removed
    - `Psalm\PluginRegistrationSocket::addFileExtension` was removed
    - :information_source: migration possible using `Psalm\PluginFileExtensionsSocket`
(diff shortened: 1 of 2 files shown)


orklah:
Please commit the deprecation on 4.x and its removal in master

But sorry for the name, I believe it was a mistake I made 1 and a half year ago :(

ohader:
> Please commit the deprecation on 4.x and its removal in master

Shall the `4.x` change also include the new method? Thus, people already could switch to use the new method in when running a `4.x` version still. `master` branch would then really only contain the removal and the change to `UPGRADING.md`. Correct?

> But sorry for the name, I believe it was a mistake I made 1 and a half year ago :(

Don't worry. I just was upgrading our code-base for Psalm v5 and stumbled over that `@todo` annotation in our code... 😃

orklah:
As long as you can add the full feature without BC break on 4.x, go for it yeah, then just remove the old on master

weirdan:
Thanks!

Issue #7528, opened 2022-01-30
Allow to intercept adding issue in IssueBuffer


ohader:

ohader:
Merged...

PR #7530, opened 2022-01-30
[FEATURE] Allow to intercept adding issue in IssueBuffer

ohader:
This change introduces new `BeforeAddIssueEvent` which is invoked
from `IssueBuffer::add`, which allows to collect and intercept code
issue in a generic way.

Resolves: #7528

### DIFF
--- a/src/Psalm/Plugin/EventHandler/BeforeAddIssueInterface.php
+++ b/src/Psalm/Plugin/EventHandler/BeforeAddIssueInterface.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Plugin\EventHandler;
+
+use Psalm\Plugin\EventHandler\Event\BeforeAddIssueEvent;
+
+interface BeforeAddIssueInterface
+{
+    /**
+     * Called before adding a code issue.
+     *
+     * @param BeforeAddIssueEvent $event
+     * @return null|bool $event How and whether to continue:
+     *  + `null` continues with next event handler
+     *  + `true` stops event handling & keeps issue
+     *  + `false` stops event handling & ignores issue
+     */
+    public static function beforeAddIssue(BeforeAddIssueEvent $event): ?bool;
+}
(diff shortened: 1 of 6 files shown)


orklah (approved):

src/Psalm/Plugin/EventHandler/Event/BeforeAddIssueEvent.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Plugin\EventHandler\Event;
+
+use Psalm\Issue\CodeIssue;
+
+class BeforeAddIssueEvent
```
ohader, line 9: should be `final`

ohader: ✅

src/Psalm/Plugin/EventHandler/Event/BeforeAddIssueEvent.php
```diff
@@ -0,0 +1,36 @@
...
+     * @var CodeIssue
+     */
+    private CodeIssue $issue;
+
+    /**
+     * @var bool
+     */
+    private bool $fixable;
+
+    public function __construct(CodeIssue $issue, bool $fixable)
```
weirdan, line 21:
```suggestion
    /** @internal */
    public function __construct(CodeIssue $issue, bool $fixable)
```

src/Psalm/Plugin/EventHandler/BeforeAddIssueInterface.php
```diff
@@ -0,0 +1,21 @@
...
+declare(strict_types=1);
+
+namespace Psalm\Plugin\EventHandler;
+
+use Psalm\Plugin\EventHandler\Event\BeforeAddIssueEvent;
+
+interface BeforeAddIssueInterface
+{
+    /**
+     * Called before adding a code issue.
```
weirdan, line 12:
```suggestion
     * Called before adding a code issue.
     *
     * Note that event handlers are called in the order they were registered, and thus
     * the handler registered earlier may prevent subsequent handlers from running by
     * returning a boolean value.
```

weirdan (changes requested):

weirdan (approved):

weirdan:
Thanks!

Issue #7518, opened 2022-01-29
False positive MethodSignatureMustProvideReturnType with stubs.


VincentLanglet:
In 4.19, the `MethodSignatureMustProvideReturnType` error was introduced.

In the following project https://github.com/sonata-project/EntityAuditBundle/pull/474, I now have errors like:
```
MethodSignatureMustProvideReturnType - src/Collection/AuditedCollection.php:127:21 - Method SimpleThings\EntityAudit\Collection\AuditedCollection::add must have a return type signature! (see https://psalm.dev/282)
    public function add($element)
```

Removing the doctrine plugin
```
<pluginClass class=“Weirdan\DoctrinePsalmPlugin\Plugin”/>
```
solve my issue.

So it would means that the Collection stub provided by the plugin is considered as a native php interface.

Related to https://github.com/vimeo/psalm/pull/7363
cc @weirdan @orklah @danog

orklah:
@bendavies made a similar claim a while ago https://github.com/psalm/psalm-plugin-symfony/issues/146

> This is causing all service classes to be marked as not user_defined

It could be the same bug that is now revealed

Should we disable this feature/Issue and release a new bugfix version until we can find and fix the issue?

bendavies:
I understand the issue here now and have a way forward. I'll try this weekend

orklah:
Cool! Glad to hear that!

VincentLanglet:
As a quick fix, shouldn't the `MethodSignatureMustProvideReturnType` rule be temporary disabled in a patch release ?

bendavies:
well, i spent a bit of time on this over the weekend and the bug i was looking at as part of https://github.com/psalm/psalm-plugin-symfony/issues/146 is no longer reproducable. 

there still is a bug here which i discussed with @orklah . i think the way forward is understood. @orklah could probably give a better summary as he understands the psalm internals better than me

caugner:
After upgrading from 4.17 to 4.19, we're seeing unexpected `MethodSignatureMustProvideReturnType` issues as well:

![grafik](https://user-images.githubusercontent.com/495429/151783573-cd2657d9-670f-4d89-b29e-2d46bb9486cd.png)

```php
# app/Http/Requests/Customer/IndexCustomerRequest.php
<?php

namespace App\Http\Requests\Customer;

use App\Models\Customer;
use App\Models\Team;
use App\Models\User;
use Illuminate\Foundation\Http\FormRequest;

/**
[...]

PR #7539, opened 2022-01-31
Revert "PHP 8.1: Report missing typehints in overridden native methods"

weirdan:
Reverts vimeo/psalm#7363


orklah:
@tm1000, @weirdan realized removing this would create a BC break and possibly break even more things for people who, for example, started to suppress this issue or possibly plugins that may have already adapted their code.

We'll think about what we'll do, disabling the feature seem a possible way

tm1000:
@orklah is there some place you two discuss these things? I originally went to the PR but did not see anything. In open source optics are everything so a long as the user is aware of why their PR was closed its fine but I couldn't find anywhere the reasons for reversion before your comment.

weirdan:
@tm1000, I didn't realize you were asking about the reason to revert the original PR. It's been discussed here: #7518.

PR #7542, opened 2022-01-31
Fix object constant inference

### LINKED ISSUE #7541: `use const Foo\foo;` infer const type as mixed
It seems Psalm already support const with new expression: https://psalm.dev/r/a273debeb2

But if I import const from other file Psalm infers it as mixed.

file1.php:
```php
<?php

namespace Foo;

final class Foo {}
const foo = new Foo();
```

file2.php:
```php
<?php

namespace Bar;

use const Foo\foo;

$_foo = foo;
/** @psalm-trace $_foo */ //$_foo is mixed
```

klimick:
Fix #7541

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
@@ -55,8 +55,7 @@ class SimpleTypeInferer
         Aliases $aliases,
         FileSource $file_source = null,
         ?array $existing_class_constants = null,
-        ?string $fq_classlike_name = null,
-        bool $const_inference = false
+        ?string $fq_classlike_name = null
     ): ?Union {
         if ($stmt instanceof PhpParser\Node\Expr\BinaryOp) {
             if ($stmt instanceof PhpParser\Node\Expr\BinaryOp\Concat) {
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -55,7 +55,8 @@ public static function infer(
         Aliases $aliases,
         FileSource $file_source = null,
         ?array $existing_class_constants = null,
-        ?string $fq_classlike_name = null
+        ?string $fq_classlike_name = null,
+        bool $const_inference = false
```
orklah, line 59: why did you need to add a flag here?

orklah:
Test failure is unrelated, but it looks like you're on an outdated revision

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
klimick: I mistakenly thought that adding `New_` would break the existing inference.

klimick:
> Test failure is unrelated, but it looks like you're on an outdated revision

I've rebase main branch. Tests is ok now.

orklah:
Thanks!

PR #7554, opened 2022-02-01
Ensure `trait_exists()` always returns `bool`

### LINKED ISSUE #7478: `trait_exists()` returns nullable value, according to psalm stubs
See https://github.com/Roave/BetterReflection/pull/983#discussion_r790908170
See https://psalm.dev/r/c41a43805d

In `vimeo/psalm:4.18.1`, `trait_exists()` is declared as `trait_exists(string): ?bool`:  it should perhaps return `bool` at all time (not nullable).

```php
<?php

function traitDoesIndeedExist(string $traitName): bool
{
    return trait_exists($traitName);
}
```

Produces

```
Psalm output (using commit c7d938b): 

ERROR: NullableReturnStatement - 5:12 - The declared return type 'bool' for traitDoesIndeedExist is not nullable, but the function returns 'bool|null'

ERROR: InvalidNullableReturnType - 3:51 - The declared return type 'bool' for traitDoesIndeedExist is not nullable, but 'bool|null' contains null
 
```

Definition seems to be imprecise in https://github.com/vimeo/psalm/blob/f1c4b62f5c65e887892e768dbaff6207ded14cee/dictionaries/CallMap.php

/cc @voku

Ocramius:
Fixes #7478

As discussed in the upstream issue, `trait_exists()` always returns `bool`: while
it can return `null` when the arguments passed to it do not match (either no arguments, or
3 or more arguments), we do not support that scenario, as that already doesn't respect the
type signature of this function.

We cut to the point: always make it `bool`, which is the scenario that works under healthy
operational conditions.

Ref: https://github.com/Roave/BetterReflection/pull/983#discussion_r790908170
Ref: https://psalm.dev/r/c41a43805d
Ref: https://github.com/vimeo/psalm/issues/7478#issuecomment-1020330351
Ref: https://github.com/vimeo/psalm/issues/7478#issuecomment-1020337712
Ref: https://3v4l.org/XpHmh

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -14855,7 +14855,7 @@ return [
 'trader_wclprice' => ['array', 'high'=>'array', 'low'=>'array', 'close'=>'array'],
 'trader_willr' => ['array', 'high'=>'array', 'low'=>'array', 'close'=>'array', 'timePeriod='=>'int'],
 'trader_wma' => ['array', 'real'=>'array', 'timePeriod='=>'int'],
-'trait_exists' => ['?bool', 'trait'=>'string', 'autoload='=>'bool'],
+'trait_exists' => ['bool', 'trait'=>'string', 'autoload='=>'bool'],
 'Transliterator::create' => ['?Transliterator', 'id'=>'string', 'direction='=>'int'],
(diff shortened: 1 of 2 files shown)


orklah:
Thanks!

PR #7563, opened 2022-02-02
Fixes #7560 manually look for DOMDocument

### LINKED ISSUE #7560: Uncaught Error: Class "DOMDocument" not found
## Summary
I use the VSCode extension for running Psalm inside the editor. When opening a PHP Symfony project, Psalm crashes immediately multiple times and shuts down.

The error is as follows:
```
[INFO] Starting language server
[INFO] Psalm Version was detected as 4.19.0
[ERROR] Uncaught Error: Class "DOMDocument" not found in /path/to/project/vendor/vimeo/psalm/src/Psalm/Config.php:707
Stack trace:
#0 /path/to/project/vendor/vimeo/psalm/src/Psalm/Config.php(734): Psalm\Config::loadDomDocument()
#1 /path/to/project/vendor/vimeo/psalm/src/Psalm/Config.php(697): Psalm\Config::validateXmlConfig()
#2 /path/to/project/vendor/vimeo/psalm/src/Psalm/Config.php(661): Psalm\Config::loadFromXML()
#3 /path/to/project/vendor/vimeo/psalm/src/Psalm/Internal/CliUtils.php(491): Psalm\Config::loadFromXMLFile()
#4 /path/to/project/vendor/vimeo/psalm/src/Psalm/Internal/Cli/LanguageServer.php(255): Psalm\Internal\CliUtils::initializeConfig()
#5 /path/to/project/vendor/vimeo/psalm/psalm-language-server(7): Psalm\Internal\Cli\LanguageServer::run()
#6 {main}
(Psalm 4.19.0@a2ad69ae4f5ab1f7d225a8dc4e2ec2d9415ed599 crashed due to an uncaught Throwable)
```

When running Psalm via CLI, `$ vendor/bin/psalm`, everything works normal.

Which additional information do you need?

### System Info
-   PHP version: PHP 8.1.2 (cli) (built: Jan 24 2022 10:43:09) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.2, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.2, Copyright (c), by Zend Technologies
-   Psalm version: 4.19.0

### User Settings
```
phpExecutablePath: "/usr/bin/php"
phpExecutableArgs: ["-dxdebug.remote_autostart=0","-dxdebug.remote_enable=0","-dxdebug_profiler_enable=0"]
psalmVersion: null
psalmScriptPath: "vendor/vimeo/psalm/psalm-language-server"
psalmScriptArgs: []
disableAutoComplete: false
maxRestartCount: 5
unusedVariableDetection: false
enableVerbose: false
connectToServerWithTcp: false
enableUseIniDefaults: false
logLevel: "INFO"
analyzedFileExtensions: [{"scheme":"file","language":"php"},{"scheme":"untitled","language":"php"}]
configPaths: ["psalm.xml","psalm.xml.dist"]
hideStatusMessageWhenRunning: true
```

tm1000:
Manually look for DOMDocument. See #7560

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -704,6 +704,10 @@ class Config
      */
     private static function loadDomDocument(string $base_dir, string $file_contents): DOMDocument
     {
+        if(!class_exists('DOMDocument')) {
+            throw new ConfigException('DOMDocument class not found. Have you installed the php-xml module?');
+        }
+
         $dom_document = new DOMDocument();
 
         // there's no obvious way to set xml:base for a document when loading it from string


AndrolGenhald:
I was thinking a better solution might be to do `extension_loaded` checks when Psalm is started.

tm1000:
> I was thinking a better solution might be to do `extension_loaded` checks when Psalm is started. It would also be good to ensure the other required extensions are loaded and give a better error message for those too.

Yes but if you look at the original issue we have to assume Psalm was installed and the extension WAS loaded as they stated that it worked fine from the other CLI entry point, just not language server entry point. 

I'm up to checking both but I still think DOMDocument has to be checked through class_exists as a fallback

AndrolGenhald:
> Yes but if you look at the original issue we have to assume Psalm was installed and the extension WAS loaded as they stated that it worked fine from the other CLI entry point, just not language server entry point.

If the extension was loaded wouldn't the class have to exist though? I was under the impression that somehow the extension was loaded when running from CLI but not when running the language server from vscode.

KevinVanSonsbeek:
> I was thinking a better solution might be to do `extension_loaded` checks when Psalm is started. It would also be good to ensure the other required extensions are loaded and give a better error message for those too.

This is probably the most elegant and reliant solution. Also means it would be easier to find the checks if they're at a consistent check location.

AndrolGenhald:
@tm1000 [Mystery solved](https://github.com/vimeo/psalm/issues/7560#issuecomment-1028642504), they were actually different environments. I think the `extension_loaded` check should catch the problem and ensure that `DOMDocument` exists without the `class_exists` check.

tm1000:
Yup you are right

So is the location I'm looking the best place to check the extension loaded or is there a better place to check for all extensions

tm1000:
It looks like composer 2.0 already does this on startup through autoload.php unless I'm mistaken 

https://getcomposer.org/doc/07-runtime.md#platform-check

AndrolGenhald:
> So is the location I'm looking the best place to check the extension loaded or is there a better place to check for all extensions

@orklah?

I would say as early as possible, maybe somewhere in `Psalm\Internal\Cli::run()`? We definitely want to make sure the check is there for all the ways Psalm can be launched, CLI, language server, Psalter, etc.

> It looks like composer 2.0 already does this on startup through autoload.php unless I'm mistaken

Oh, I wasn't aware that was done at runtime, I've only used that through the CLI interface.
"The default value is php-only which only checks the PHP version."
Maybe we can change that to check extensions at runtime too?

tm1000:
The issue with relying on composer (and yes there is a setting we can put into composer.json to check the extensions) is that it looks to be a 2.0 feature only

AndrolGenhald:
I think we could require `"composer-runtime-api": "^2.0"` without causing any problems. I believe it's required anyway due to dependencies, but I could be wrong about that.

tm1000:
@AndrolGenhald 

That was actually added about a month ago and then removed by @weirdan in a revert (see: https://github.com/weirdan/psalm/commit/a4878ac2b5b7b0d4271e79eb0b8411320580fa2d)

I don't think we can require that at this time. Best check might just to be to do it ourselves.

weirdan:
> Maybe we can change that to check extensions at runtime too? 

That option goes into `config` entry that works for root packages only, so whatever we set it to would get ignored when people install psalm with `"require" : { "vimeo/psalm": "..." }`

tm1000:
@weirdan good point.

weirdan:

weirdan:
Superseded by #8782

PR #7565, opened 2022-02-02
Fix bullet indentation in documentation

AndrolGenhald:
MkDocs requires 4 spaces for bullet indentation.

Searched with `/^( *)[*+-].*\n\1 {1,3}[*+-]/`. The language_server.md change is apparently not needed, I'm guessing since the non-indented bullets have a space in front of them as well it only requires 3 spaces, but that seems inconsistent, so I changed it anyway in case that's a bug in the markdown parser that might get fixed in the future.


orklah:
Thanks!

PR #7574, opened 2022-02-03
Fix extension capitalization issue, add extensions to version message.

AndrolGenhald:
Adds a list of enabled and supported extensions, and enabled but unsupported extensions to the PHP version message that Psalm outputs. This should help with confusion resulting from #7107.

Also fixes capitalization, previously requiring `"ext-PDO"` wouldn't enable the extension due to the capitalization difference.


orklah:
Thanks!

PR #7577, opened 2022-02-04
Add extension documentation.

### LINKED ISSUE #7513: Create tool to generate stubs for an extension

AndrolGenhald:
Fixes #7513.

Psalm's stub generator had enough issues that I decided to fork https://github.com/sasezaki/php-extension-stub-generator and get it working again.

I added documentation on how to add/improve support for extensions, and some generated stubs for the extensions I have installed.

### DIFF
--- a/docs/contributing/adding_an_extension.md
+++ b/docs/contributing/adding_an_extension.md
@@ -1,0 +1,42 @@
+# Adding an extension
+
+Psalm tries to support most extensions with stubs so that projects requiring extensions can be analyzed without having
+those extensions installed at runtime. When an extension isn't already supported by Psalm, if it is loaded at runtime
+Psalm will use reflection to support it as best as possible, but it's always better to have stubs. In addition to
+allowing analysis without having the extension installed, stubs can be more accurate. The gmp and decimal extensions
+have a lot of functions that use `numeric-string` instead of `string`, and many extension like Ds, DOM, and others use
+templates. Most extension stubs can be improved by using Psalm features that aren't possible from reflection alone.
+
+## Using the stub generator
+
+We try to make adding support for an extension as easy as possible by providing a simple way to generate stubs with
+reflection, but keep in mind that reflection can miss a lot of useful information:
+
+ - Which `@throws` tags should be added to a method
+ - Psalm specific types like `numeric-string`, `int<0, max>`, or `non-empty-list`.
+ - [Templates](../annotating_code/templated_annotations.md)
+
+To use the extension stub generator, download the latest phar at
+https://github.com/AndrolGenhald/php-extension-stub-generator/releases and run `php-extension-stub-generator.phar dump
+[extension-name]`. Open a pull request to https://github.com/AndrolGenhald/php-extension-stub-generator to add the stub,
+or open a GitHub issue there and copy the output into the issue or a Gist.
+
+## Updating stubs
+
+Generated stubs are kept for each version of an extension at
+https://github.com/AndrolGenhald/php-extension-stub-generator/tree/master/generated_stubs.
+
+When adding a new extension, simply copy the generated stub to `stubs/extensions/[extension-name-lowercase].phpstub` and
+update as necessary with Psalm types. Alternatively, if the extension provides its own stub somewhere that includes
+things like `@throws` tags or other features reflection isn't able to provide, use that as a base, but make sure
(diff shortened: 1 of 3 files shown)


docs/contributing/adding_an_extension.md
```diff
@@ -0,0 +1,44 @@
...
+
+We try to make adding support for an extension as easy as possible by providing a simple way to generate stubs with
+reflection, but keep in mind that reflection can miss a lot of useful information:
+
+ - Which `@throws` tags should be added to a method
+ - Psalm specific types like `numeric-string`, `int<0, max>`, or `non-empty-list`.
+ - [Templates](../annotating_code/templated_annotations.md)
+
+To use the extension stub generator, download the latest phar at
+https://github.com/AndrolGenhald/php-extension-stub-generator/releases and run `php-extension-stub-generator.phar dump
```
AndrolGenhald, line 20: Does someone want to create psalm/php-extension-stub-generator and move this there?

AndrolGenhald:
To add a bit more clarity here, the intention is that after this is merged, I'll open an issue for each of the extensions here to update the actual stubfile that Psalm uses. I figured it'd be best to do a separate PR for each extension. See `adding_an_extension.md` for how I expect this process to work, if I don't explain it satisfactorily there I'd rather update the documentation than explain in the PR comments.

docs/contributing/adding_an_extension.md
orklah: We'd have to ask Matt I think. Let's do that when it's finished

stubs/extensions/reflection-generated/dom-8.1.2.phpstub
```diff
@@ -0,0 +1,1019 @@
+<?php
+// Generated with PHP version 8.1.2 dom version 20031129
```
AndrolGenhald, line 2: I named the file dom-8.1.2 because I'm pretty sure this dom version is misleading and the dom extension has actually changed along with php. I'm not sure why it doesn't show the php version like the other php-bundled extensions, but I'm guessing that's a bug on their end.

docs/contributing/adding_an_extension.md
AndrolGenhald: Well, I think it is finished, more or less :stuck_out_tongue:
I was thinking this would be merged as mostly a documentation/process update.

AndrolGenhald:
@orklah I'm wondering if we should keep the reflection-generated stubs in a separate repo, there are getting to be quite a number of them and they're not directly needed by Psalm anyway. Do you have any preference there? Maybe I should stick them in the stub generator repo?

orklah:
well, you'll end up requiring them anyway and always in the latest version. Also I'm not sure I want to know what will happen the day someone will try using Psalm 5 with stubs made for Psalm 7 :p

But I'm undecided, I don't have much experience on advantages or drawbacks of making separate repos

AndrolGenhald:
Well the generated stubs aren't used directly, I'm just using them to simplify seeing what's changed between versions. Some of the extensions I'm taking one of the generated stubs as a base and applying stuff from the CallMap/PropertyMap/existing stub on top of that, and I'm using `@since` to support multiple versions so there's only going to be one stub that's actually used. The reflection-generated stubs I'm just keeping around for historical/diffing purposes, so Psalm doesn't really need them at all.

I tried to explain this as best I could in `adding_an_extension.md`.

orklah:
Yeah, I need to start the review first :p

AndrolGenhald:
I went ahead and removed them. Should be ready for review.

orklah:
Mind rebasing? There seems to be a few build errors that I would not expect.

Otherwise this seems good

@weirdan we might want to create a new repo on psalm org for the stub generator. Is this something you can do?

weirdan:
@orklah I can, but I'll need the name for the repo.

orklah:
I think we can keep the current name `php-extension-stub-generator`

Issue #7512, opened 2022-01-29
Improve versioning tag support


AndrolGenhald:
We currently support `@since` tags for requiring PHP versions on functions and methods, but the support is currently very limited and has some issues:

 - No support for properties
 - No support for consts
 - Declarations for later versions are required to come after declarations for earlier versions
 - No support for package or extension versions, only PHP versions
 - No way to distinguish between package/extension versions and PHP versions ([this](https://psalm.dev/r/265ca1743a?php=7.4) should not fail)
 - No way to indicate a declaration is removed after a certain version

We should add support to properties and consts, and produce the same results regardless of the order of redeclarations.

I think we should use `@since x.y[.z]` to specify package/extension version and `@since PHP x.y` to specify PHP version. Package versions can be ignored for now since it's harder to support, but extension versions shouldn't be too hard to do when using the extension stubs (some extensions track PHP versions so it doesn't matter, but some don't).

I haven't seen a standard tag for removal at a version, so I propose we use `@until PHP x.y`.

A declaration should be ignored when:
 - A `@since PHP x.y` tag exists and the minimum PHP version is less than x.y
 - An `@until PHP x.y` tag exists and the maximum PHP version is greater than x.y

This would ensure the code is correct for all PHP versions, not just the minimum PHP version.

We'll have to add `$minimum_php_version_id` and `$maximum_php_version_id` to `Codebase` to allow this, and we'll probably want to deprecate `$analysis_php_version_id` to be removed in Psalm 6.

It would also be nice to warn about conflicting declarations:
```
// composer.json has PHP ^7.0|^8.0

class Foo
{
    /** @since PHP 7.0 */
    public static function bar(string $arg): void {}

    /** @since PHP 8.0 */
    public static function bar(): void {}
}

Foo::bar("baz"); // This function is actually impossible to call correctly without first checking the PHP version
```

But I think for now detecting version checks and only loading the applicable declarations for that scope is too complicated.
For now when multiple declarations exist for supported versions, we'll probably just want to stick with the declaration with the lowest version.

orklah:
It could be interesting to check phpstorm stubs, that phpstan read and parse. I believe they have a similar feature so we could have a compatible syntax.

> We'll have to add $minimum_php_version_id and $maximum_php_version_id to Codebase to allow this, and we'll probably want to deprecate $analysis_php_version_id to be removed in Psalm 6.

[...]

PR #7590, opened 2022-02-05
Improve support for @since tag, along with various adjacent improvements

### LINKED ISSUE #7588: Crash when parsing invalid docblock on class const
https://psalm.dev/r/a5d51c4caa

I'll be fixing this as a side effect of some other changes.

AndrolGenhald:
Add support for `@since` on class constants and properties.
Improve `@since` on functions.
Fix #7588.
Add DuplicateProperty issue.

This is a breaking change, `@since` tags that target PHP versions are now required to use the format `@since PHP-x.y.z` instead of `@since x.y.z`. This should allow `@since` tags for library versions to live alongside those for PHP versions without confusion.


AndrolGenhald:
I still want to do more improvements for `@since` I think we'll want it on classes themselves as well in addition to the other ideas I have in #7512, but I think this is enough to get most of the extension stubs updated so we can remove them from the CallMap.

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
```diff
@@ -386,14 +389,14 @@ public function start(PhpParser\Node\Stmt\ClassLike $node): ?bool
             }
         }
 
-        foreach ($node->getComments() as $comment) {
-            if (!$comment instanceof PhpParser\Comment\Doc) {
+        foreach ($node->getComments() as $doc_comment) {
+            if (!$doc_comment instanceof PhpParser\Comment\Doc) {
```
AndrolGenhald, line 393: I didn't actually mean to change `$comment` to `$doc_comment` and `$comments` to `$parsed_docblock` across the entire class, but after I did I figured it'd be better if it was consistent, so I kept it.

src/Psalm/Internal/Scanner/DocblockParser.php
```diff
@@ -258,4 +262,27 @@ private static function resolveTags(ParsedDocblock $docblock): void
...
+
+    /**
+     * @throws DocblockParseException if the @since PHP tag is malformed
+     */
+    public static function parseSincePhpVersion(ParsedDocblock $docblock): ?int
+    {
+        if (!isset($docblock->tags['since'])) {
+            return null;
+        }
+        $since = trim(reset($docblock->tags['since']));
```
AndrolGenhald, line 274: In the future this should be changed to a loop in case a non-PHP `@since` tag comes first.

orklah:
I'm not a big fan of the PHP- syntax. Isn't it a currently used format? For example https://github.com/JetBrains/phpstorm-stubs/blob/c228fd014d23903cb59c788d53b3ffa90f6b916c/FFI/FFI.php#L255

And I think PHPStan must use the same format since they parse phpstorm stubs

If we must, we could have a @psalm-since with that format, but it's probably overkill. Just stop interpreting the tag if the format is unexpected?

AndrolGenhald:
> I'm not a big fan of the PHP- syntax.

It's not ideal, but with the way it works right now [this](https://psalm.dev/r/265ca1743a?php=7.4) is a problem. We could limit it to exclude user code, but that prevents nice stuff like [this](https://psalm.dev/r/5aa3bb7cfb) in a library.

Since the current syntax makes it impossible to know what version the `@since` tag is referring to, I think if we keep it we need to be much more conservative about where we analyze it, but I'm not even sure how we would do that. We could allow it on core stubs and extensions, but what about PECL extensions that don't track the PHP version?

I almost proposed a compromise of allowing the `PHP-` syntax but still supporting the `@since 8.0` syntax in certain places, but now I'm not even sure we can safely tell what places the non `PHP-` syntax should be allowed. It doesn't really matter for core stubs since we can change those to use either, but if we're going to fix it to stop breaking [this](https://psalm.dev/r/265ca1743a?php=7.4) I'm not sure where we _can_ allow it...

orklah:
@weirdan WDYT about changing the parsing of @since ?

It would make Psalm behave differently than phpstan but would make sure it's really directed at the php version.

I'm quite mixed about that personally

AndrolGenhald:
The main problem is [this](https://psalm.dev/r/265ca1743a?php=7.4), do you know if PHPStan handles that correctly? I [tried it](https://phpstan.org/r/43dec731-11c8-4459-8634-c90aa0ccba4b) but I can't tell if there's a way to tell PHPStan to use a specific version there, so idk if it handles it correctly or not.

I'm actually a bit surprised this hasn't caused issues already, but I'm guessing that's because we limit it to `/^[4578]\.\d(\.\d+)?$/` and there aren't too many projects that both have a version number that high and also use `@since` tags.

AndrolGenhald:
@orklah Any further thoughts on this? I think there has to be a way to distinguish between PHP version, extension version, and user library version, and this is the only way I can think of to do this.

### Alternatives

#### Use namespaces

PHP historically sticks way too much stuff in the global namespace, as do users, so this is basically impossible to do correctly.

#### Treat stubs differently

ie user code would be considered to use `@since` with the user project version, stubs would use PHP version. This would prevent user stubs from using the user's project version, and it doesn't allow for extension versions in stubs either.

#### Only use PHP/extension version for built in stubs

Users wouldn't be able to use `@since` with their own stubs for built in or extension defined things. This seems like it's probably less of a downside than the others and maybe worth considering, but it's still a downside, especially when there are extensions that Psalm doesn't directly support.

---

Are there any other ways to accomplish this that I haven't considered?

orklah:
Well, I don't think we discussed it already: what if we used a different tag than @since altogether? That way, any @since in user code would be ignored instead of trying to assume what it means. Using another keyword would allow phpstan to pick it up effortlessly instead of messing with their own stubs. Maybe something like @php-from, @php-to and @lib-from, @lib-to ?

orklah:
that's exactly my issue :)

They're using `@since` for php versions. If we try to introduce a new standard `@since-php` for php versions and `@since` for something else entirely it will be very confusing.

Either we use with the same syntax and meaning (that would mean keeping whatever is after `@since` as a php version not prefixed) and we need to have a special syntax for libraries or extensions, otherwise, switching keyword entirely would probably be easier (having an alias is better than having a single symbol with multiple meaning between tools)

PR #7595, opened 2022-02-05
Finalize issues

weirdan:


orklah (approved):

pilif:
This is indeed a bit of a shame. 

In our code-base, we have some legacy patterns we want to discourage and I wrote a plug-in to find those and throw a deprecation warning. 

But I’m reporting an anonymous subclass of the issue and set the URL to point to our wiki with a more detailed explanation. 

If you need to make these final, please consider letting plugins provide URLs and message overrides.

weirdan:
Is there any reason you're not extending `PluginIssue` instead?

https://github.com/vimeo/psalm/blob/b06fb932d486eb7165bbea6625583506e4b2cfd1/src/Psalm/Issue/PluginIssue.php#L5

pilif:
Because the pattern is about a deprecation, so using the native deprecation made sense with regards of configuration how to deal with deprecations in which directories. 

Basing the issue in PluginIssue produces a secondary hierarchy of issues which needs a separate configuration.

PR #3967, opened 2020-08-09
specify requirement phpdocumentor/reflection-docblock:^4.3.4 || ^5

sasezaki:
specify requirements for the sake of travis "low" deps fails.

ダウンロード (4)
  - https://travis-ci.org/github/vimeo/psalm/jobs/716250055

----

"low" deps would install phpdocumentor/reflection-docblock with `4.0.0`. but, reflection-docblock 4.0.0. has bug `Uninitialized string offset: 0 `.
  - see https://github.com/phpDocumentor/ReflectionDocBlock/issues/111

---

 - Currently, psalm requirements `"php": "^7.1.3|^8"` 
 - phpspec/prophecy:1.9.0, 1.10.0 requirements `php:"^5.3|^7.0"`
 - phpspec/prophecy:1.9.0, 1.10.0 requirements `"phpdocumentor/reflection-docblock": "^2.0|^3.0.2|^4.0|^5.0",`

But, prophecy upper version drop support php 7.1.
 - phpspec/prophecy:1.11.0 requirements `php:"^7.2"`,
 - phpspec/prophecy:1.11.0 requirements `"phpdocumentor/reflection-docblock": "^5.0",`.
and reflection-docblock:5 also requires `"php": "^7.2"`

So, specify requirement phpdocumentor/reflection-docblock:^4.3.4 || ^5 .

PR #7612, opened 2022-02-09
No longer require declare reflection-docblock:^5 as dependency

sasezaki:
see https://github.com/vimeo/psalm/pull/3967
prophecy upper version always require phpdocumentor/reflection-docblock:^5.0.


weirdan:
Thanks!

Issue #7581, opened 2022-02-04
phpredis stubs wrong again?


kkmuffme:
https://psalm.dev/r/c49242e965

I think since 4.2.0, phpredis stubs seem to be completely wrong and reporting lots of false positive issues
(had that same issue some time ago, but was working fine until recently)

orklah:
4.20 shouldn't have changed on this matter, however, psalm.dev runs on dev-master which contains this change: https://github.com/vimeo/psalm/pull/7107

And because Redis is neither on composer.json nor in the config for psalm.dev, stubs are simply not loaded.

But again, it should not be an issue on an install of 4.20.0. Are you sure you have this issue?

kkmuffme:
Yes, I'm sure:
>vimeo/psalm                                    4.20.0

And I get the exact same errors as on psalm.dev for the above code.
Didn't get these errors before.

AndrolGenhald:
Should have realized it when I read this earlier, it can't be #7107 because redis isn't in the supported extensions list, we don't have a stubfile for it.

orklah:
When I look at redis official stubs, I see:
zAdd expects an int as second param, you sent an array
https://github.com/phpredis/phpredis/blob/552f41a44f0687c221e5d4adf9ee387479bb78a1/redis.stub.php#L488

zpopmin was apparently added a few months ago and I think we never had definitions for it:
https://github.com/phpredis/phpredis/blame/552f41a44f0687c221e5d4adf9ee387479bb78a1/redis.stub.php#L502

mget expects an array, you passed a string:
https://github.com/phpredis/phpredis/blob/552f41a44f0687c221e5d4adf9ee387479bb78a1/redis.stub.php#L267

mget indeed returns something else than an array, in their stub, contrary to what we have in Psalm, but we have this wrong (or old?) definition since 2016

I think you may have just baselined all these errors and they somehow get out of the baseline when upgrading to another version.

You're welcome to fix the definitions in Psalm :)

AndrolGenhald:
Or, if you want to be a guinea pig, you could try following the instructions [here](https://github.com/vimeo/psalm/pull/7577/files#diff-a8f5996f44efa5bcb7120605c7af05f11887a00b3743ee7a0e886afcab1584a1) to help Psalm support the extension with the new design, which will allow analyzing the extension even when it's not present when running Psalm. Fair warning though, it's not merged yet and the process may end up changing a bit.

kkmuffme:
[...]

PR #7614, opened 2022-02-09
add phpredis stubs

kkmuffme:
Fixes https://github.com/vimeo/psalm/issues/7581

Source: https://github.com/phpredis/phpredis/blob/e93be87023fffbf9213e8f47066f6dd646d8ba5a/redis.stub.php
=> these are not fully correct though (return types are partly wrong, params are partly wrong).

However there is a pending PR with correct stubs by the phpredis maintainer with correct stubs:
https://github.com/phpredis/phpredis/pull/2015
So I used the stubs from that PR as they are correct: https://github.com/phpredis/phpredis/blob/e93be87023fffbf9213e8f47066f6dd646d8ba5a/redis.stub.php

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -2009,6 +2009,12 @@ class Config
             $this->internal_stubs[] = $ext_decimal_path;
         }
 
+        // phpredis
+        if (extension_loaded('redis')) {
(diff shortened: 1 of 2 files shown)


AndrolGenhald:
For 4.x you'll have to add an `extension_loaded` check around [here](https://github.com/vimeo/psalm/blob/4.x/src/Psalm/Config.php#L1976) so that the stub is actually used.

orklah:
Thanks!

PR #7621, opened 2022-02-09
add ReflectionIntersectionType stub

orklah:
This stub was missing, the goal is to fix current CI failure

### DIFF
--- a/stubs/Php81.phpstub
+++ b/stubs/Php81.phpstub
@@ -56,6 +56,13 @@ namespace {
          */
         public function getBackingValue(): int|string;
     }
+
+    class ReflectionIntersectionType extends ReflectionType {
+        /**
+         * @return non-empty-list<ReflectionType>
+         */
+        public function getTypes() {}
+    }
 }
 
 namespace FTP {


orklah:
Ok, I understand what's going on.

First, the issue is due to our phpunit being updated by Matt yesterday, that's what changed.

But then, the issue come from phpunit requiring PHP 8.0 while still using ReflectionIntersectionType which is a PHP 8.1 class only.

However, this is not an issue because it's used in an instanceof, that doesn't emit PHP error in this case, but it's not handled by Psalm currently. PHPStan does not like that either: https://phpstan.org/r/9c8f7205-b3c0-40e7-818c-b555e2b55206

I guess we should just baseline those, it's quite an uncommon pattern and it deserve to be warned against in most cases

orklah:
Internals are pretty useful too, that's why we have stubs of them. However, we just don't load stubs for higher versions when dealing with lower versions, hence the CI failure here

sebastianbergmann:
> the issue come from phpunit requiring PHP 8.0

No stable version of PHPUnit requires PHP 8.0.

orklah:
@sebastianbergmann Thanks for your input :) we don't use a stable version however, we're hooked on your master branch (frozen on a fixed version of whenever we refreshed it)

right now, it needs PHP 8.0 or more: https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L25 but use a symbol introduced in 8.1, causing the error here

PR #7622, opened 2022-02-09
consistency in AssertionFinder

orklah:
This doesn't do much, it only fixes a small phpdoc error and rearrange a few methods and params to have a good consistency between scrapeEqualityAssertions and scrapeInequalityAssertions

There's a few things that are not directly order changes:
- The call to getTrueEqualityAssertions didn't pass the inside_conditional value contrary to getTrueInequalityAssertions, getFalseEqualityAssertions and getFalseInequalityAssertions so I added it
- The count_equality_position check had been put after an early return when we were not dealing with a StatementsAnalyzer. I did that on an earlier PR because I needed the node provider for retrieving types. However, I only needed it for handling paradoxicalAssertions, so I put that back above, like the count_inequality_position check
- I added the check for paradoxicalAssertions in the count_inequality_position check. (it will possibly find some weird conditions in user code)

Nothing big but it will be cleaner for future changes

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -1782,7 +1782,7 @@ class AssertionFinder
     }
 
     /**
-     * @param Identical|Equal|Smaller|SmallerOrEqual|NotIdentical|NotEqual $conditional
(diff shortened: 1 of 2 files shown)


orklah:
Cool, I was able to find a bug too. We were transforming array-key into string as soon as we were seeing an Int assertion, but assertion could have been int ranges or literal int

PR #7623, opened 2022-02-09
Finalize types nodes

weirdan:


orklah:
Does removing final a BC break in itself?

orklah (approved):

weirdan:
There's no such tag, unfortunately.

Issue #7567, opened 2022-02-03
DOMNodeList::item() gives DOMElement instead of DOMNode


phptest2:
https://psalm.dev/r/8def2b8419 gives

ERROR: [UndefinedMethod](https://psalm.dev/022) - 12:15 - Method DOMNode::getElementsByTagName does not exist

https://3v4l.org/hF7Rc
gives DOMElement instead of DOMNode

I guess https://www.php.net/manual/de/domnodelist.item.php is also wrong with the return type.

Previous versions did not give the ERROR here.

Psalm 5.0.0-alpha1, PHP 8.0.15.

phptest2:
https://github.com/php/doc-en/issues/1375

orklah:
Ok, so while any DomNodeList could contains other DomNodes apart from DomElement, it's clearly documented on our part here that it's not the case for the one returned by `getElementsByTagName` : https://github.com/vimeo/psalm/blob/4.x/stubs/DOM.phpstub#L9

It's definitely a regression.

I think this has to do with https://github.com/vimeo/psalm/pull/7107 but I'm still unclear how the dom extension could not be loaded while Psalm still manages to run. I'll keep digging

cc @AndrolGenhald

AndrolGenhald:
Definitely #7107, I'll work on the extension stuff more next week. After #7515 is done this will result in some stronger error since Psalm won't think `DOMDocument` exists at all unless the extension is enabled. The problem right now is that it's still using reflection to analyze the extension even though the extension isn't enabled, but I need to generate complete stubs to ensure the extension is fully supported without reflection before #7515 can be fixed.

PR #7632, opened 2022-02-11
Adjust for composer-require-checker check.

sasezaki:
_To avoid conflict, I do PR into master branch_

* Added Composer\Audoload\ClassLoader into `bin/composer-require-checker-config.json` , it is soft dependency symbol.
* DOMElement instead of DomElement - https://php.net/DOMElement
  to avoid composer-require-check reports false-positive,
    just did - `find src/ -type f -exec sed -i -e 's/DomElement/DOMElement/g' {} \;`

### Before
```
$ composer-require-checker --config-file=$(pwd)/bin/composer-require-checker-config.json
ComposerRequireChecker 3.8.0@537138b833ab0f9ad72b667a72bece2a765e88ab
The following 2 unknown symbols were found:
+-------------------------------+--------------------+
| Unknown Symbol                | Guessed Dependency |
+-------------------------------+--------------------+
| Composer\Autoload\ClassLoader |                    |
| DomElement                    |                    |
+-------------------------------+--------------------+
```

### After
```
composer-require-checker --config-file=$(pwd)/bin/composer-require-checker-config.json
ComposerRequireChecker 3.8.0@537138b833ab0f9ad72b667a72bece2a765e88ab
There were no unknown symbols found.
```
* Also composer-require-checker v4 check passed.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -764,7 +764,6 @@ class Config
 
         $psalm_nodes = $dom_document->getElementsByTagName('psalm');
 
-        /** @var DOMElement|null */
         $psalm_node = $psalm_nodes->item(0);
 
         if (!$psalm_node) {
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -7,7 +7,7 @@ use Composer\Semver\Constraint\Constraint;
 use Composer\Semver\VersionParser;
 use DOMAttr;
 use DOMDocument;
-use DomElement;
+use DOMElement;
 use InvalidArgumentException;
 use JsonException;
 use LogicException;
@@ -764,7 +764,7 @@ class Config
 
         $psalm_nodes = $dom_document->getElementsByTagName('psalm');
 
-        /** @var DomElement|null */
+        /** @var DOMElement|null */
         $psalm_node = $psalm_nodes->item(0);
 
         if (!$psalm_node) {
@@ -853,7 +853,7 @@ class Config
     }
 
     private static function processDeprecatedElement(
-        DomElement $deprecated_element_xml,
+        DOMElement $deprecated_element_xml,
         string $file_contents,
         self $config,
         string $config_path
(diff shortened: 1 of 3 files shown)


src/Psalm/Config.php
```diff
@@ -764,7 +764,7 @@ private static function validateXmlConfig(string $base_dir, string $file_content
 
         $psalm_nodes = $dom_document->getElementsByTagName('psalm');
 
-        /** @var DomElement|null */
+        /** @var DOMElement|null */
         $psalm_node = $psalm_nodes->item(0);
```
sasezaki, line 768: oh, good catch! 😄

sasezaki: removed at 04ea5c0

sasezaki: > https://github.com/vimeo/psalm/runs/5150670385?check_suite_focus=true

I'm currently wondering about this psalm report.

>  Error: src/Psalm/Config.php:767:18: UnnecessaryVarAnnotation: The @var DOMElement|null annotation for $psalm_node is unnecessary (see https://psalm.dev/212)

https://psalm.dev/r/7bd042f521

In this case, inline phpdoc var should be set to suppress error.

sasezaki: > removed at https://github.com/vimeo/psalm/commit/04ea5c01c74690ccf9585c637e819b50ae0900a7

I did reset this,  just now.

sasezaki: related : https://github.com/vimeo/psalm/issues/7567

sasezaki: so, `getElementsByTagName()` returns `DOMNodeList<DOMElement>` is intended behavior.
https://github.com/vimeo/psalm/blob/11e60fa261c682627e6b85300dd6f424b87520b1/stubs/extensions/dom.phpstub#L9

Ok, I would remove this `/** @var DOMElement|null */`

AndrolGenhald:
Can we add this check to composer.json scripts and CI?

orklah:
Thanks!

PR #7633, opened 2022-02-11
Allow config to define thread count (4.x)

M1ke:
Applying the same commits from https://github.com/vimeo/psalm/pull/7442 to the 4.x branch


M1ke:
As mentioned here (https://github.com/vimeo/psalm/pull/7442#issuecomment-1026056111) if this gets merged to 4.x I am happy to resolve any conflicts with master, but I believe it should fit in directly.

M1ke:
I can see that the check `ci/circleci: test-with-real-projects` fails and it's hard to see what in this branch would cause this - nothing in this work would change how Psalm carries out its analysis, and it will only apply the thread limit if the config file is modified.

orklah:
failures in real projects are indeed unrelated :)

orklah:
Thanks!

PR #7641, opened 2022-02-11
Extension stubs

AndrolGenhald:
Add extension stubs and remove all stubbed declarations for extensions from callmap and propertymap, and disable reflection for supported extensions.
This allows Psalm to properly show issues for `UndefinedClass`, `UndefinedFunction`, etc when an extension is not enabled.


AndrolGenhald:
#7590 should be merged first. Despite the passing tests, things might break for end-users if the stubs don't correctly interpret the `@since` tags for version changes.

The failure for PHPUnit is by design, suggested and require-dev extensions should be added to the `<enableExtensions>` config.

For the backward compatibility checker, I'm not quite sure what to do. It doesn't matter for now, but after Psalm 5 is released we'll want to be able to continue adding more extensions. Should we just add a comment that there's no compatibility guarantee for the keys of that type and suppress the error? Is the error only there because `Config` isn't `final` or is it considered a breaking change even if it is `final`?

src/Psalm/Internal/PhpVisitor/TraitFinder.php
```diff
@@ -71,6 +78,10 @@ public function getNode(): ?PhpParser\Node\Stmt\Trait_
 
         try {
             $reflection_trait = new ReflectionClass($this->fq_trait_name);
+
+            if ($this->config->isReflectionFromSupportedExtension($reflection_trait)) {
+                return null;
+            }
```
AndrolGenhald, line 84: I'm not too sure about this one, but I don't know of any extensions that define traits. I suppose if it's not working correctly it can be fixed if/when such an extension is added.

AndrolGenhald:
Note to self: `@since` tags targeting PECL extension versions should be changed to use the prefix like `@since [ext-name]-x.y`.

src/Psalm/Config.php
```diff
@@ -2172,7 +2188,11 @@ public function getPredefinedConstants(): array
 
     public function collectPredefinedConstants(): void
     {
-        $this->predefined_constants = get_defined_constants();
+        $predefined_constants = get_defined_constants(true);
+        foreach (self::SUPPORTED_EXTENSIONS as $ext_name) {
+            unset($predefined_constants[$ext_name]);
```
orklah, line 2193: What's the point of that?

src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php
```diff
@@ -593,12 +593,10 @@ public static function isContainedBy(
                         return true;
                     }
 
-                    if ($codebase->methods->methodExists(
-                        new MethodIdentifier(
-                            $input_type_part->value,
-                            '__tostring'
-                        )
-                    )) {
+                    if ($codebase->methods->methodExists(new MethodIdentifier($input_type_part->value, '__tostring'))
```
orklah, line 596: I feel like this would belong in another PR

tests/PropertyTypeTest.php
```diff
@@ -717,22 +717,28 @@ class Foo {
...
+                ',
                 'assertions' => [
                     '$owner' => 'DOMDocument|null',
                 ],
+                'ignored_issues' => [],
+                'php_version' => '7.3', // Not needed, only here because required_extensions has to be set
+                'required_extensions' => ['dom'],
             ],
-            'propertyMapHydration' => [
+            'SKIPPED-propertyMapHydration' => [ // Property map is no longer used for DOM.
```
orklah, line 729: maybe we could keep that test active with another property?

src/Psalm/Config.php
AndrolGenhald: Related to #7647, that's where constants that exist at runtime are loaded. This is basically removing any constants that come from a supported extension so that they'll instead be supplied by the stub.

stubs/extensions/xdebug.phpstub
```diff
@@ -1,43 +1,66 @@
 <?php
 
-const XDEBUG_TRACE_APPEND = 1;
```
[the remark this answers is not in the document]
AndrolGenhald: Looks like they exist in [3.1.2](https://github.com/AndrolGenhald/php-extension-stub-generator/blob/master/generated_stubs/xdebug-3.1.2.phpstub) but not [3.0.0](https://github.com/AndrolGenhald/php-extension-stub-generator/blob/master/generated_stubs/xdebug-3.0.0.phpstub). I don't remember exactly, but I'm guessing I was thinking it'd be better to leave them out until we support `@since` for extensions. On the other hand, there's some functions in here with `@since 3.1`, so apparently I wasn't consistent in that.

src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php
AndrolGenhald: Ehh, maybe. I sort of considered it a minor fix to the extension's behavior, I added it because I removed the `__toString` from the stub since the function doesn't actually exist.

src/Psalm/Config.php
AndrolGenhald: eg if you're running PHP 8.1 and have PDO enabled but you only require PHP "^8.0" this allows Psalm to warn you that [`PDO::FETCH_DEFAULT`](https://github.com/vimeo/psalm/pull/7641/files/8c4da1902a4134b2e5ce1c22a7af39f814f1d20e#diff-e91509795683dcfbba2e373ab4c4488dfce559643e06766ae954ccb19c2cad79R21-R22) doesn't exist.

stubs/extensions/xdebug.phpstub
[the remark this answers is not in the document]
orklah: Mainly I'm worried about projects depending on those in any version. They're gonna have undefined constants if we remove that

src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php
orklah: ok, fine by me :)

stubs/extensions/xdebug.phpstub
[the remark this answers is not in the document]
AndrolGenhald: Yeah, I'll add it back and try to add support for `@since [extension-version]-x.y.z`. I think it's fine to show those undefined constant warnings for `"ext-xdebug": "*"` as long as it works correctly for `"ext-xdebug": "^3.1"`.

PR #7642, opened 2022-02-11
skip a test because we removed the feature for now

orklah:
Not sure why this test doesn't explode on 4.x but we removed this feature for now due to shenanigans with stubs from plugins and this makes master branch fail


weirdan:
> why this test doesn't explode on 4.x

Because it's absent from that branch :man_shrugging:

AndrolGenhald:
I did a bit of work to allow `mixed` in stubs for PHP 7.4 on #7641, this test is still failing there though, I didn't look into it much.

PR #7648, opened 2022-02-12
Organize extension requirements in composer.json

sasezaki:
continuation of #6844 and #7632.

* Declare as "ext-filter" is require obviously.
* Added ext-pcntl, ext-posix to composer.json suggest property.
  * I added as "In order to run psalm-language-server".  If it is not correct, please pointed it.
* use PDOException literal instead of ::class to avoid taken as dependency
  * composer-require-checker guessed dependency "ext-PDO". But only use of `PDOException::class`.

### DIFF
--- a/bin/composer-require-checker-config.json
+++ b/bin/composer-require-checker-config.json
@@ -18,8 +18,7 @@
     "pcntl",
     "posix",
     "filter",
-    "curl",
-    "PDO"
+    "curl"
   ],
   "scan-files" : []
 }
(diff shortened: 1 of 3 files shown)


AndrolGenhald:
>     * Added ext-pcntl, ext-posix to composer.json suggest property.
>       
>       * I added as "In order to run psalm-language-server".  If it is not correct, please pointed it.

I believe they're used for "threads" to speed up analysis, I'm not sure if they're related to the language server or not.

PR #7659, opened 2022-02-13
Correct name of ErrorException::__construct $line param

### LINKED ISSUE #7658: Parameter $line does not exist on function ErrorException::__construct
The `ErrorException` constructor takes a parameter called `line`, but Psalm disagrees.

Source code: https://github.com/php/php-src/blob/0b524a56f5ed1fa878a7f1f402643eb0aa52b09b/Zend/zend_exceptions_arginfo.h#L51-L58
3v4l example (runs fine): https://3v4l.org/E6A1i
Psalm example (fails type checking): https://psalm.dev/r/acebce1efa

```php
<?php

new ErrorException(line: 1);
```

The other parameters (`message`, `severity`, `filename`) are all known to Psalm; only `line` appears to be problematic.

chloekek:
Fixes #7658.

Thanks to @weirdan for the help!

Do I have to add a regression test?


weirdan:
No, it's fine as it is

weirdan:
Thanks!

PR #7660, opened 2022-02-13
Update return type for mysqli::connect_error

morozov:
Steps to reproduce: https://psalm.dev/r/62a046ceec ([documentation](https://www.php.net/manual/en/mysqli.connect-error.php)).

The [procedural stub](https://github.com/vimeo/psalm/blob/11e60fa261c682627e6b85300dd6f424b87520b1/dictionaries/CallMap.php#L8449) is already correct.


weirdan:
This could go to 4.x actually

weirdan:
Thanks!

Issue #7654, opened 2022-02-12
Backed enums not correctly returning as instances of enums


davidbyoung:
Sorry, the title is a little vague, but when I am setting a variable to a backed enum value, and then checking if it's an instance of that backed enum, Psalm gets a little tripped up.  In my example below, `$intCode` will always be an integer.

https://psalm.dev/r/6953255922

weirdan:
But ultimately `enum(Code::Ok)` and `Code` should be collapsed into `Code` when used in a union.

PR #7662, opened 2022-02-13
Enum assertions

orklah:
This will fix https://github.com/vimeo/psalm/issues/7654

It doesn't adress the collapsing issue, but it makes sure the assertions are correctly reconciled

### DIFF
--- a/src/Psalm/Internal/Provider/StatementsProvider.php
+++ b/src/Psalm/Internal/Provider/StatementsProvider.php
@@ -155,10 +155,6 @@ class StatementsProvider
 
             $existing_statements = $this->parser_cache_provider->loadExistingStatementsFromCache($file_path);
 
-            if ($existing_statements && !$existing_statements[0] instanceof PhpParser\Node\Stmt) {
-                $existing_statements = null;
-            }
-
             $existing_file_contents = $this->parser_cache_provider->loadExistingFileContentsFromCache($file_path);
 
             // this happens after editing temporary file
--- a/tests/EnumTest.php
+++ b/tests/EnumTest.php
@@ -405,6 +405,35 @@ class EnumTest extends TestCase
                 'ignored_issues' => [],
                 'php_version' => '8.1',
             ],
+            'EnumCaseReconciliation' => [
+                'code' => '<?php
+                    enum Code: int
+                    {
+                        case Ok = 0;
+                        case Fatal = 1;
+                    }
+
+                    function foo(): Code|null
+                    {
+                        return null;
+                    }
+
+                    $code = foo();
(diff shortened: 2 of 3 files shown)


src/Psalm/Internal/Provider/StatementsProvider.php
```diff
@@ -155,10 +155,6 @@ public function getStatementsForFile(
 
             $existing_statements = $this->parser_cache_provider->loadExistingStatementsFromCache($file_path);
 
-            if ($existing_statements && !$existing_statements[0] instanceof PhpParser\Node\Stmt) {
```
orklah, line 158: I made a check that went a little too far so I had to revert it, but it saw this as redundant. I couldn't find the reason for this, it seems redundant indeed

tests/EnumTest.php
```diff
@@ -405,6 +405,35 @@ function foo(): int|Code|null
...
+
+                    $code = foo();
+                    $code1 = null;
+                    $code2 = null;
+                    if($code instanceof Code){
+                        $code1 = $code;
+                    }
+                    if(!$code instanceof Code){
+                        $code2 = $code;
+                    }',
```
weirdan, line 429: Does it actually test reconciliation with `Enum|Enum::Case`? It doesn't seem like an individual enum case would ever appear in this test case.

orklah: what I wanted to test was the enum case not leaking to the negated if like in here: https://github.com/vimeo/psalm/issues/7654#issuecomment-1037378155

The test would be cleaner without null values, (it showed Code|Code::Ok and never on the other var before) but assertions doesn't work if the var appears only in an if

weirdan: I mean, this test passes on 5.0a1 (I just tested that). So it doesn't cover changes introduced in this PR.

orklah: Ugh :( I'll try to design a better test then!

weirdan: What you could do instead is to abort (exit or throw) in the `if` body, and then assert the type as it was changed below the `if`. E.g.
```php
[
            'EnumCaseReconciliation' => [
                'code' => '<?php
                    enum Code: int
                    {
                        case Ok = 0;
                        case Fatal = 1;
                    }

                    function foo(): Code|null|int
                    {
                        return null;
                    }

                    $code = foo() ?? Code::Ok;
                    if (!$code instanceof Code) throw new RuntimeException;
                    ',
                'assertions' => [
                    '$code===' => 'Code',
                ],
                'ignored_issues' => [],
                'php_version' => '8.1',
            ],
]
```

PR #7666, opened 2022-02-13
More class const improvements.

### LINKED ISSUE #7668: Support for final constants
https://psalm.dev/r/867638ab8b?php=8.1

AndrolGenhald:
Add InvalidClassConstType issue as alternative to LessSpecificClassConstType when type isn't contravariant.
Handle final class consts (#6395).
Use double quotes for types in class const issues.

Fixes #7668


src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php
```diff
@@ -721,38 +724,89 @@ public static function analyze(
...
+            if ($const_storage->stmt_location !== null) {
+                // Check final in PHP < 8.1
+                if ($codebase->analysis_php_version_id < 8_01_00 && $const_storage->final) {
+                    IssueBuffer::maybeAdd(
+                        new ParseError(
+                            "Class constants cannot be marked final before PHP 8.1",
+                            $const_storage->stmt_location,
+                        ),
+                        $const_storage->suppressed_issues
+                    );
```
AndrolGenhald, line 807: Not sure if there's a better issue for this, I went with ParseError because that's what it is if you use some other invalid keyword.

weirdan: `ParseError` is good.

weirdan:
Can you add a test highlighting Psalm behaviour in regards to non-final interface constant inheritance? They were forbidden in PHP 5.0-8.0, and are allowed in PHP 8.1+: https://3v4l.org/PYZFB

AndrolGenhald:
> Can you add a test highlighting Psalm behaviour in regards to non-final interface constant inheritance? They were forbidden in PHP 5.0-8.0, and are allowed in PHP 8.1+: https://3v4l.org/PYZFB

[Way ahead of you](https://psalm.dev/r/684e1851bf?php=8.0), that was added in #7154.

weirdan:
Thanks!

Issue #7672, opened 2022-02-14
Incorrect type for RuntimeException::getCode()


AndrolGenhald:
https://psalm.dev/r/512ac9662c https://3v4l.org/RBCW0

@VincentLanglet re #7390

I suggest copy/pasting https://github.com/vimeo/psalm/blob/06ce3adba5dca8491ba144256d0abccc060f7984/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallReturnTypeFetcher.php#L99 and doing a `RuntimeException::class` check as well.

Also, PDOException should return `int|string` (or maybe `0|string` but I'm not sure): https://3v4l.org/9vu2u

VincentLanglet:
I followed PHPStan on this https://phpstan.org/r/42089d28-dbaa-4b67-ac00-154372ce55a8
But indeed, it need to be fixed.

orklah:
PDOException is a child of RuntimeException? :o

What a mess :( ...

VincentLanglet:
I just opened an issue https://github.com/phpstan/phpstan/issues/6651 and a PR https://github.com/phpstan/phpstan-src/pull/1018 on PHPStan. I'll hope it will be reviewed quickly to be sure phpstan is going in the same direction.

Do you want to make the fix on psalm @AndrolGenhald or should I ?

AndrolGenhald:
You can go ahead, I'm in the middle of rewriting TryAnalyzer.

VincentLanglet:
I made https://github.com/vimeo/psalm/pull/7673. But I would recommend to wait for a response from the phpstan-side in order to copy the behavior.

PR #7673, opened 2022-02-14
PDOException extends RuntimeException and can use int code errors

VincentLanglet:
Closes https://github.com/vimeo/psalm/issues/7672

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallReturnTypeFetcher.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallReturnTypeFetcher.php
@@ -29,6 +29,7 @@ use Psalm\Type\Atomic\TGenericObject;
 use Psalm\Type\Atomic\TNamedObject;
 use Psalm\Type\Atomic\TTemplateParam;
 use Psalm\Type\Union;
+use RuntimeException;
 use Throwable;
 use UnexpectedValueException;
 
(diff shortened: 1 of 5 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallReturnTypeFetcher.php
```diff
@@ -97,16 +98,14 @@ public static function fetch(
 
         if ($premixin_method_id->method_name === 'getcode'
             && $premixin_method_id->fq_class_name !== Exception::class
+            && $premixin_method_id->fq_class_name !== RuntimeException::class
+            && $premixin_method_id->fq_class_name !== PDOException::class
```
orklah, line 102: Please use "PDOException" as a string. We don't want to introduce a false dependency when it's not needed

VincentLanglet: It was already used with `$premixin_method_id->fq_class_name === PDOException::class`, you said it was ok for FQCN: https://github.com/vimeo/psalm/pull/7390#discussion_r784799089

orklah: Yeah, but I believe we went back on another PR, let me check that

orklah: nvm, this was for constants for allowing to use them without having the extension: https://github.com/vimeo/psalm/pull/7107/files#r765332473

It shouldn't be an issue for ::class, forget I said anything :)

AndrolGenhald: It was changed to a string in #7648.

orklah:
@VincentLanglet what's the status of this? It seems to still be in draft on your side

VincentLanglet:
> @VincentLanglet what's the status of this? It seems to still be in draft on your side

The PR is ready and I made the same on the Phpstan. I was just waiting an answer/merge from ondrej on the Phpstan PR before. In order to be sure that both psalm and phpstan will have the same behavior.

VincentLanglet:
Seems like the PR https://github.com/phpstan/phpstan-src/pull/1018 will be accepted on phpstan

orklah:
Thanks!

PR #7686, opened 2022-02-17
Add `@psalm-check-type` and `@psalm-check-type-exact`.

AndrolGenhald:
I initially added these as part of my TryAnalyzer rewrite to allow testing complicated `finally` types like this:
```php
$foo = 1;
try {
        $foo = 2;
} catch (Exception $_) {
        $foo = 3;
} finally {
        $bar = $foo;
        /** @psalm-check-type-exact $bar = 1|2|3 */;
}
/** @psalm-check-type-exact $bar = 2|3 */;
```
Using the `'assertions'` in tests doesn't work since the type is different inside the `finally`.

I decided to extract the new annotation from the rest of my changes and do a separate pull request since I think others may find it useful, and it should be much easier to review than the entire TryAnalyzer rewrite.


weirdan (approved):

orklah:
That's so freaking great! I wanted something like that for ages!

PR #7688, opened 2022-02-17
Rewrite TryAnalyzer

### LINKED ISSUE #5123: False positive for undefined variable
https://psalm.dev/r/1fcfcf14b9

What's interesting is that it's not reproducible if not inside a class method.

### LINKED ISSUE #5700: UnusedVariable false positive with thrown and caught exceptions
If a variable is modified just before an exception is thrown, and then caught in the same context and the variable used, psalm emits an UnusedVariable issue.

https://psalm.dev/r/95027a9b22
https://3v4l.org/9UkSq

### LINKED ISSUE #5967: False positive for undefined variable in try-catch-finally
This is probably related to #5123, but a more general case:  
https://psalm.dev/r/17c2685cb3

In all possible code paths, the variable will be defined.

### LINKED ISSUE #5973: Incorrect PossiblyUndefinedVariable after try/catch statement
Hi, in this case `$data` is always defined but a `PossiblyUndefinedVariable` error is triggered.

https://psalm.dev/r/6fda415fad

### LINKED ISSUE #6286: Invalid PossiblyUndefinedVariable when try/catch/finally
With this code
```
function func(string $name): void {
    try {
    }
    catch (Throwable $e) {
        if ($name === 'stuff') throw $e;
        echo $e->getCode();
    }
    finally {
    }
}
```

I get and error for line `echo $e->getCode();`:
```
INFO: PossiblyUndefinedVariable - 9:14 - Possibly undefined variable $e defined in try block
```
Funny thing, if I remove `if ($name === 'stuff') throw $e;` or if I remove `finally`, the error disappears...

Link: https://psalm.dev/r/e9822c86ed

### LINKED ISSUE #6342: False positives after "continue" or rethrown exception
When we define a variable within a try catch within a loop and then either continue the loop or rethrow the catched exception, psalm thinks the variable might be unset.

I fail to see any path that could lead to the variable being undefined.

Reproducer for global scope:
https://psalm.dev/r/fd6cc07cc7

Reproducer for function scope:
https://psalm.dev/r/fd6cc07cc7

### LINKED ISSUE #6659: False positive UnusedVariable when wrapping catched exception
When an exception is catched and overridden in the catch block, psalm thinks the variable is unused (might be related to #5700).

https://psalm.dev/r/24f4faa0bd

To make the issue even worse, this can NOT be supressed:
https://psalm.dev/r/9b5d2bd80c

### LINKED ISSUE #7249: Flaw in the co-dependant variables types
In the example here: https://psalm.dev/r/9bc618221f

Psalm doesn't detect that $type_context can't be null because we're in a scope where $completion_data is null and we already stated that both can't be null.

Psalm manages to make it work if the conditional is outside the try/catch: https://psalm.dev/r/ab2b600564.

This is pretty low priority because it's an edge case of a complex module and the code, while correct, is a little confusing anyway.

Found by: https://github.com/vimeo/psalm/pull/7247

AndrolGenhald:
As usual I bit off way more than I intended. It turns out try/catch/finally is actually really complicated...

Before reviewing I strongly recommend looking through the `unionAssignmentsWithMultipleNestedTry` test line by line and fully understanding it.

Fixes #5123, fixes #5700, fixes #5967, fixes #5973, fixes #6286, fixes #6342, fixes #6659, fixes #7249. 

The one thing that really annoys me about this change is that the `finally` analysis assumes that there will never be an uncaught exception thrown in a `catch`:

```
$foo = -1;
try {
    maybeThrow();
    $foo = 1;
} catch (Exception $_) {
    maybeThrow();
    $foo = 2;
} finally {
    if ($foo < 1) {} // False positive
    takesPositiveInt($foo); // False negative
}
```

If anyone strongly objects to this I'll change it back, I just hate to throw away all that work :/.

I'm personally fairly ok with this since I don't think throwing in a `catch` with a `finally` after it is too common of a pattern anyway, and even when it does happen I'd be surprised if these false negatives/positives caused a problem. It also helps prevent other false positives as well:

```
try {
    maybeThrow();
    $foo = 1;
} catch (Throwable $_) {
    $foo = 2;
} finally {
    return $foo; // If we assume `catch`es can throw anywhere, $foo is possibly undefined
}
```


AndrolGenhald:
Shepherd failure is waiting for #7684.

AndrolGenhald:
Could someone baseline the PHPUnit cast to int errors? I almost didn't notice the extra errors that showed up there from this PR. I'll look tomorrow to see if they're anything to worry about.

src/Psalm/Internal/Analyzer/Statements/UnusedAssignmentRemover.php
```diff
@@ -45,7 +47,28 @@ public function findUnusedAssignment(
...
+                    $original_location->raw_file_start + 1,
+                    $original_location->raw_file_start + 1 + strlen($assign_stmt->var->name),
+                    "_",
+                );
+            } else {
+                $file_content = $codebase->file_provider->getContents(
+                    $original_location->file_path
+                );
+                $end_of_exception = $original_location->raw_file_start - 1;
+                for (; ctype_space($file_content[$end_of_exception]); --$end_of_exception);
```
orklah, line 63: this is to handle multiple exception on a single catch, right?

tests/TryCatchTest.php
```diff
@@ -369,13 +368,6 @@ class_alias(
                         }
                     }'
             ],
-            'suppressUndefinedVarInFinally' => [
```
orklah, line 372: does this cause an issue?

src/Psalm/Internal/Analyzer/Statements/UnusedAssignmentRemover.php
AndrolGenhald: That's to remove all whitespace. `catch (Exception|Error $e)` becomes `catch (Exception|Error)`, but also
```php
catch (Exception|Error
    $e)
```
becomes
```php
catch (Exception|Error)
```
instead of
```php
catch (Exception|Error
    )
```
I imagine it's not too relevant most of the time, but maybe there are multiple lines for some very long exception names. I also noticed a couple of places where we have 2 spaces between the exception and the variable, so it fixes that as well.

tests/TryCatchTest.php
AndrolGenhald: I think I removed that thinking it was testing a false negative that I fixed, but I'm not sure why I thought that. It seems to run fine now, I'll re-add it.

AndrolGenhald:
[This](https://psalm.dev/r/fc93066322) is currently failing with `RedundantCondition - src/somefile.php:12:30 - Type Exception for $e is never null`, I'm wondering if it's related to the `$vars_in_scope` not being cloned issue. Maybe I'll end up fixing that here after all...

AndrolGenhald:
Still waiting on #7684, otherwise I think this is good to go. The two remaining issues for PHPUnit are sort of like [this](https://psalm.dev/r/ea3ae076d5), so I think they can be ignored.

src/Psalm/Internal/Analyzer/Statements/Block/TryAnalyzer.php
```diff
@@ -118,396 +109,543 @@ public static function analyze(
...
-                $catch_context->removeDescendents(
-                    $catch_var_id,
-                    $catch_context->vars_in_scope[$catch_var_id],
-                    $catch_context->vars_in_scope[$catch_var_id],
-                    $statements_analyzer
-                );
+        $fq_catch_classes = [];
 
-                $catch_context->vars_possibly_in_scope[$catch_var_id] = true;
+        assert(!empty($catch->types));
```
orklah, line 260: maybe this should be part of PropertyMap or php-parser stubs? It would make more sense

src/Psalm/Internal/Analyzer/Statements/BreakAnalyzer.php
```diff
@@ -16,7 +16,7 @@
 class BreakAnalyzer
 {
     public static function analyze(
-        StatementsAnalyzer $statements_analyzer,
+        StatementsAnalyzer $_statements_analyzer,
```
orklah, line 19: we're on master, if it's not used anymore, we could remove it

src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php
```diff
@@ -1048,7 +1048,7 @@ private function createStorageForFunctionLike(
                     $docblock_info = null;
                     try {
                         $docblock_info = FunctionLikeDocblockParser::parse($doc_comment);
-                    } catch (IncorrectDocblockException|DocblockParseException $e) {
```
orklah, line 1051: nice!

tests/DocumentationTest.php
```diff
@@ -65,6 +65,7 @@ class DocumentationTest extends TestCase
      * annotations that we don’t want documented
      */
     private const INTENTIONALLY_UNDOCUMENTED_ANNOTATIONS = [
+        '@psalm-check-type', // Used internally for testing try-catch-finally, not sure if we want to support it
```
orklah, line 68: should this be left here?

AndrolGenhald: Nope, that was before I pulled it out into the separate PR, guess when I merged that in it got left there.

tests/TryCatchTest.php
```diff
@@ -142,7 +141,7 @@ function test(): string {
 
                     try {
                         $var = test();
-                    } catch (Exception $e) {
+                    } catch (Throwable $e) {
```
orklah, line 144: If I get this right, a catch(Exception) is no longer seen as a "catch all" part?

It's technically true, but do we want to add potentially lots of false positive for a few true positives?

src/Psalm/Internal/Analyzer/Statements/BreakAnalyzer.php
AndrolGenhald: I thought about it, but I wondered if it'd be better left for consistency since all of the calls [here](https://github.com/AndrolGenhald/psalm/blob/try-analyzer-rewrite/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php#L533:L580) are otherwise the same though. It's all internal I guess, so might as well remove it.

tests/TryCatchTest.php
AndrolGenhald: Actually I need to go through and undo most of those, I'd planned to and forgot about it. `Throwable` is the "catch all" instead of `Exception`, but I changed those back on a previous attempt where I handled it incorrectly. The "catch all" thing only actually matters for `finally`.

tests/Traits/ValidCodeAnalysisTestTrait.php
```diff
@@ -70,21 +70,29 @@ public function testValidCode(
         $this->analyzeFile($file_path, $context);
 
         $actual_vars = [];
-        foreach ($assertions as $var => $_) {
+        foreach ($assertions as $expr => $_) {
             $exact = false;
 
+            $var = $expr;
```
weirdan, line 76: Is there any reason to alias this? `$expr` doesn't seem to be used in the loop body.

AndrolGenhald: Nope, that must be a leftover change from when I was figuring out how to get it working.

weirdan:
One possible pattern that employs both `finally` and exception re-throw in `catch` is nested transactions, e.g. https://psalm.dev/r/b176fbdb47

tests/TryCatchTest.php
```diff
@@ -491,6 +491,613 @@ function foo() : void {
...
+                'code' => '<?php
+                    $foo = 1;
+                    try {
+                        $foo = 2;
+                    } catch (Exception $_) {
+                        unset($foo);
+                    }
+                ',
+                'assertions' => [
+                    '$foo?===' => '1|2',
```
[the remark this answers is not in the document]
orklah: I think it's assumed to consider that an exception could happen at any moment. Maybe not on a simple case like this, but if you replace `$foo = 2;` by `$foo[] = 2;` you get an error that could be caught.

tests/TryCatchTest.php
```diff
@@ -491,6 +491,613 @@ function foo() : void {
...
+                    $foo = 1;
+                    $bar = "baz";
+                    try {
+                        extract([]); // extract overrides all variables with `mixed`.
+                    } catch (Exception $_) {
+                    }
+                ',
+                'assertions' => [
+                    '$foo' => 'int|mixed',
+                    '$bar' => 'mixed|string',
```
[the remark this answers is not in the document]
orklah: We could probably add a TypeCombiner::combine to collapse those

tests/TryCatchTest.php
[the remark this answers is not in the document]
weirdan: I mean, assignment either happened and `foo` is `2` or didn't (due to exception) and `foo` is unset, so to me it looks like it should result in a possibly unset 2 => `'$foo?===' => '2'`.

AndrolGenhald:
> One possible pattern that employs both `finally` and exception re-throw in `catch` is nested transactions, e.g. https://psalm.dev/r/8e926f8c41

That one should be fine, it only becomes an issue if the `finally` assumes something happened in the `catch` but the `catch` might throw an exception before that happens.

tests/TryCatchTest.php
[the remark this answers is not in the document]
AndrolGenhald: You're right, I'll have to take a look at that. If an uncaught exception is thrown execution won't continue, if a caught exception is thrown `$foo` is unset, and if no exception is thrown `$foo` is `2`. I think the issue is that right now it only considers a variable overridden by the `catch` if an assignment happens, it doesn't realize that the `unset` also lets us take only the last value from the `try`.

tests/TryCatchTest.php
```diff
@@ -491,6 +491,613 @@ function foo() : void {
...
+                    $foo = 1;
+                    try {
+                        $foo = 2;
+                    } catch (Exception $_) {
+                        unset($foo);
+                        return;
+                    }
+                ',
+                'assertions' => [
+                    '$foo===' => '1|2',
```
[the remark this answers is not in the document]
AndrolGenhald: Similar to above, returning `catch`es should allow us to assume the `try` completed successfully. I'm not quite sure why this one isn't working though, I thought I'd accounted for that.

tests/TryCatchTest.php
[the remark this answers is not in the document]
orklah: > the idea was to allow better completion

I think it's just a bug. Given the sheer amount of Unions created, it would probably be a horrible place to put an automatic combining there for performances, but IMHO, we should never end up with this kind of types, it's just the symptom of a wrong handling somewhere

theodorejb:
Is it still planned to finish this? It would be nice for all the related issues to be fixed (I just ran into #5700).

AndrolGenhald:
I should have some time again next month, I'll try to rebase it and see what issues pop up, I think there was a thorny one with the `if` analysis since I ended up redoing a significant portion of that to fix some scope issues with possibly undefined variables.

PR #7690, opened 2022-02-18
Throw exception if file_put_contents failed

dmitryuk:
If `psalm.xml` is not writable, commands `vendor/bin/psalm-plugin enable|disable %some_plugin` will always response
```
 [OK] Plugin enabled 
``` 

with warning:
```
Warning: file_put_contents(/var/www/psalm.xml): Failed to open stream: Permission denied in /var/www/vendor/vimeo/psalm/src/Psalm/Internal/PluginManager/ConfigFile.php on line 146
```
For disabled warnings (`error_reporting=E_ERROR`) a coder will success without writing the result.
The PR throws an exceptions this case

### DIFF
--- a/src/Psalm/Internal/PluginManager/ConfigFile.php
+++ b/src/Psalm/Internal/PluginManager/ConfigFile.php
@@ -143,6 +143,9 @@ class ConfigFile
             }
         }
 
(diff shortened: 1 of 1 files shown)


weirdan:
Thanks!

PR #7699, opened 2022-02-18
Fix issues with int range unpacking and with min/max

### LINKED ISSUE #1820: query re: issue with max
This is possibly a little more complicated than I thought, because `max()` accepts & processes a lot more than numerics o_O

I've got an issue whereby I'm treating `max()` as `max(T, ...T):T`, with what I would be hoping would be detected for `T as int`.

https://psalm.dev/r/a181337b14

```php
<?php
$files = [
  __FILE__,
  __FILE__,
  __FILE__,
  __FILE__,
];

$a = array_map('filemtime', $files);
$b = array_map(
  function (string $file) : int {
    return filemtime($file);
  },
  $files
);

$A = max(filemtime(__FILE__), ...$a);
$B = max(filemtime(__FILE__), ...$b);

echo date('c', $A), "\n", date('c', $B);
```
> Psalm output (using commit 5bb28cc): 
> 
> INFO: MixedAssignment - 17:1 - Cannot assign $A to a mixed type
> 
> INFO: MixedAssignment - 18:1 - Cannot assign $B to a mixed type
> 
> INFO: MixedArgument - 20:16 - Argument 2 of date cannot be mixed, expecting int
> 
> INFO: MixedArgument - 20:37 - Argument 2 of date cannot be mixed, expecting int

### LINKED ISSUE #7697: Literal array with int ranges is broken
https://psalm.dev/r/f460152b78

Love these sorts of error messages :laughing:
At least it's clearly a bug so it's less likely to confuse people.

AndrolGenhald:
Fixes #1820, fixes #7697.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -93,6 +93,28 @@ class ArrayAnalyzer
             );
         }
 
+        // if this array looks like an object-like array, let's return that instead
+        if ($array_creation_info->can_create_objectlike
+            && $array_creation_info->property_types
+        ) {
+            $object_like = new TKeyedArray(
+                $array_creation_info->property_types,
+                $array_creation_info->class_strings
+            );
+            $object_like->sealed = true;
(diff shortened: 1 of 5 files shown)


weirdan:
Thanks!

PR #7700, opened 2022-02-20
Docker image build 

Legion112:
Add docker image build
https://github.com/vimeo/psalm/discussions/6602#discussioncomment-2196704

- [ ] add GitAction automatic build for image
- [ ] Refactor to use phar to minimize image size
- [ ] setup credentials for psalm on docker hub

### DIFF
--- a/entrypoint
+++ b/entrypoint
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+/psalm/psalm "$@"
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+vendor
+vendor-bin
+composer.lock
+vendor-bin/box/vendor
+vendor-bin/box/composer.lock
+.circleci
+.github
+assets
+bin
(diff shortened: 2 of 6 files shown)


entrypoint
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
weirdan, line 1: Please move docker stuff into its own folder, e.g. `docker`. It'll be easier to export-ignore it in package distribution.

Legion112:
Move to another request. Have some problem with branch name.
https://github.com/vimeo/psalm/pull/7701

entrypoint
Legion112: This one impossible to move. Must be in the root. 
.dockerignore

Legion112: https://github.com/vimeo/psalm/pull/7701/commits/957001bfdec4bef2855fdc19dbd16fbe8920abf8

PR #7701, opened 2022-02-20
Add image build for psalm for 4.x version

Legion112:
Add docker image build
https://github.com/vimeo/psalm/discussions/6602#discussioncomment-2196704

- [x] add GitAction automatic build for image
- [ ] Refactor to use phar to minimize image size (in another PR)
- [ ] setup credentials for psalm on docker hub @weirdan  
- [x] move docker staff to separate directory
- [ ] Submit official image build (This should be done in another PR)
https://docs.docker.com/docker-hub/official_images/#creating-a-docker-official-image


.github/workflows/build-image.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
Legion112, line 17: I would need your help for setup those secrets @weirdan.

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Log in to Docker Hub
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: fearofcode/psalm
```
Legion112, line 22: This should be change to
```suggestion
          images: vimeo/psalm
```
I leave it as is to be able to test it on my docker hub account, when we are ready to merge need to apply the suggestion. 
To test the image you could run `docker run --rm -it -v "${PWD}":/app -w /app fearofcode/psalm -c psalm.xml.dist`

Legion112: Seem like creating an official image would require a lot of effort and time now.
I think it's better to publish those images under `vimeo/psalm` first and after to try to submit a request for an official image to the Docker community. 
I have never done that but looks like it's stored in a separate repository like [PHP](https://github.com/docker-library/php).

ciaranmcnulty: You could use ${{ env.GITHUB_REPOSITORY }} here if it makes testing easier

docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM composer as composer
+WORKDIR /app
+COPY ./ /app
```
ciaranmcnulty, line 3: What's actually needed to composer install? It may just be the `composer.{lock, json}` in which case restricting it will cause this layer to re-run less

docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM composer as composer
```
ciaranmcnulty, line 1: I'd name the stage something different so it doesn't collide with the official image

docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+WORKDIR /app
+COPY ./ /app
+RUN composer install --no-dev --no-progress --optimize-autoloader --no-interaction
+
+FROM php:8.1-cli-alpine
+WORKDIR /app
+COPY --from=composer /app/ /psalm/
+COPY docker/entrypoint /usr/bin/entrypoint
+ENTRYPOINT ["/usr/bin/entrypoint"]
```
ciaranmcnulty, line 10: Add a newline here at the end I think

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,31 @@
+name: Build image
```
ciaranmcnulty, line 1: I'm not clear what the difference between the two workflows is?

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Extract metadata (tags, labels) for Docker
```
ciaranmcnulty, line 18: What are the default tags here? You'll want the semver stuff at least

docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM composer as composer
+WORKDIR /app
+COPY ./ /app
+RUN composer install --no-dev --no-progress --optimize-autoloader --no-interaction
+
+FROM php:8.1-cli-alpine
+WORKDIR /app
+COPY --from=composer /app/ /psalm/
+COPY docker/entrypoint /usr/bin/entrypoint
```
ciaranmcnulty, line 9: I think it'd be good practice to validate the install:

```
/usr/bin/entrypoint --version
```

ciaranmcnulty (commented):
Looks mostly good to me, I do think the PHAR would make more sense, as you could pick it up from the other workflow and we'd get definite consistency between the phar+image

weirdan:

orklah:

PR #7708, opened 2022-02-21
Config threads: add to config.xsd

M1ke:
Resolves a bug caused by https://github.com/vimeo/psalm/pull/7633

That work didn't add to the `config.xsd` file - this was missed as the test config wasn't altered to use threads, and the unit tests for config don't attempt a validation of a file against the xsd file.

If possible this could go out in a point release as it will prevent Psalm running if `threads` is used in the config.


M1ke:
Also spotted XML not loading correctly anyway, so it's also not respecting the threads value. Now fixed also.

weirdan (approved):

weirdan:
Thanks!

PR #7692, opened 2022-02-18
Analyze attribute statements instead of constructing virtual statements.

AndrolGenhald:
Fixes #7685.

My main worry here is that the code locations might get messed up somehow. I added the line/column numbers to the existing tests, but is there a way to test the ending column of the message in addition to the starting column?

PR #7713, opened 2022-02-21
Use current context when analyzing attributes

### LINKED ISSUE #7710: False positive NonStaticSelfCall when attributes are used
Issue:

https://psalm.dev/r/7f3d9feda6

But this is legit usage, take a look at:

https://3v4l.org/faA2D#v8.1.3

#### Update:

Suppressing this error still leaves ``MixedArgument`` problem. Version 4.21.0 works normally, only ``dev`` fails.

AndrolGenhald:
Fixes #7710 (broken in #7692).


orklah:
Thanks!

PR #7715, opened 2022-02-22
Fix first-class callable in loop

trowski:
Found another case of a first-class callable failing due to hitting the assertion in PHP-Parser `CallLike::getArgs()`, this time when used within a loop when not directly assigned to a variable (e.g., used as a parameter to another function).


orklah:
Thanks!

PR #7718, opened 2022-02-22
Fix #3036: make argument $read of internal PHP function stream_socket nullable.

### LINKED ISSUE #3036: Inconsistent types for stream_select
https://psalm.dev/r/9bb31457dc

`$read` isn't reported as possibly null, while `$write` is reported as possibly null. These should be consistent. As long as they're arrays before, they should still be arrays afterwards, I guess.

niconoe-:
According to documentation: https://www.php.net/manual/en/function.stream-select.php

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -13811,7 +13811,7 @@ return [
 'stream_notification_callback' => ['callback', 'notification_code'=>'int', 'severity'=>'int', 'message'=>'string', 'message_code'=>'int', 'bytes_transferred'=>'int', 'bytes_max'=>'int'],
 'stream_register_wrapper' => ['bool', 'protocol'=>'string', 'class'=>'string', 'flags='=>'int'],
 'stream_resolve_include_path' => ['string|false', 'filename'=>'string'],
-'stream_select' => ['int|false', '&rw_read'=>'?resource[]', '&rw_write'=>'?resource[]', '&rw_except'=>'?resource[]', 'seconds'=>'?int', 'microseconds='=>'?int'],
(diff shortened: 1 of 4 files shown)


niconoe-:
TBH, I don't really know how to use this function, and I though just mapping the CallMap with the PHP doc would be sufficient...

To me, it can take NULL and it can return NULL : https://3v4l.org/4QKQt

My question is "can it take NULL but out as an array of streams?" and "can it take an array of streams but out as NULL ?".
On this, I can't answer.

orklah:
The fact that it can return null is enough to say Psalm should check against this, so this seem legit.

Can you add checks in Psalm code then?

orklah:
No test needed, but Psalm uses the function internally and your changes makes it fail on CI

Can you check error in Psalm in the "Run Shepherd" job? It should point you to specific pieces of codes where "null" case should be handled

niconoe-:
Ok, thanks.

I casted the usage of it in a `foreach` to an array (`(array)null === []` so it will just don't go into the foreach block).

Also, I upated the historical CallMap as it provoked an error in the unit tests executed.

The error on the job "Pull Request Labels / label", on the other hand, looks like out of my hands, unless you can help me figure out what I've done wrong?

orklah:
That's weird that this error pops up...

muglug:
@niconoe- could you add the stub for `stream_select` suggested by @AndrolGenhald instead?

AndrolGenhald:
@niconoe- You can basically just copy/paste the function from [this](https://psalm.dev/r/50ec07381c) into CoreGenericFunctions.phpstub somewhere and change the name and whatever other tweaks are needed so it's correct.

niconoe-:
> @niconoe- You can basically just copy/paste the function from [this](https://psalm.dev/r/50ec07381c) into CoreGenericFunctions.phpstub somewhere and change the name and whatever other tweaks are needed so it's correct.

Done.

Thank you for the help :+1:

orklah:
Thanks!

PR #7746, opened 2022-03-02
Register openssl_sign function to impure functions

samsonasik:
openssl_sign has $signature parameter that by reference that can re-used, eg:

```php
        $pkeyid = "some_public_key";
	openssl_sign("blah", $signature, $pkeyid);
 
        return $signature;
```


orklah:
Thanks!

Issue #7630, opened 2022-02-10
Add strict init/config option


AndrolGenhald:
Psalm has a lot of configs that default to being lenient, but some users wish to use Psalm to keep their code to the strictest standards. Even if a user is aware that many config options default to being lenient, it's a chore to go through all of them and figure out what they should be.

It would be nice to provide an option to initialize a new project with strict defaults to help with this.

Even better, we could add a strict mode config that changes default values for other configs unless they are explicitly overridden in the config file. This way newly added configs will still use their strict mode setting even for a project initialized with a previous version of Psalm, and it simplifies adoption (no need to initialize an empty project and copy over all the changed settings).

PR #7758, opened 2022-03-05
DateInterval::createFromDateString can be false

### LINKED ISSUE #7747: TypeDoesNotContainType - DateInterval does not contain false
https://psalm.dev/r/07a940ed95

This is a mistake, because [`DateInterval::createFromDateString`](https://www.php.net/manual/ru/class.dateinterval.php) can return `false`. 

```php
public static DateInterval::createFromDateString(string $datetime): DateInterval|false
```

Ne-Lexa:
Fixes #7747

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -1762,7 +1762,7 @@ return [
 'DateInterval::__construct' => ['void', 'spec'=>'string'],
 'DateInterval::__set_state' => ['DateInterval', 'array'=>'array'],
 'DateInterval::__wakeup' => ['void'],
-'DateInterval::createFromDateString' => ['DateInterval', 'time'=>'string'],
+'DateInterval::createFromDateString' => ['DateInterval|false', 'time'=>'string'],
 'DateInterval::format' => ['string', 'format'=>'string'],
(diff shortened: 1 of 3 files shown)


orklah:
Thanks! Can you also do this change in Callmap_historical? There should be a match between Callmap (who is the current state) and another Callmap file (who'll show when this particular signature was introduced)

Ne-Lexa:
@orklah, changes made.

orklah:
Thanks!

orklah:
And congrats for your first PR!

boesing:
Was this possibly reverted?

Does not create an issue in 4.26 nor in latest 5.x commit:
https://psalm.dev/r/49ce1d4e62

AndrolGenhald:
@boesing [Still works](https://psalm.dev/r/760586cee1), that's just [ignoreInternalFunctionFalseReturn](https://psalm.dev/docs/running_psalm/configuration/#ignoreinternalfunctionfalsereturn) being annoying.

The fact that there are so many config values that most people never look at, and the amount of work it takes to go through all of them manually, is what motivated me to open #7630, but I've been short on time lately. If you or someone else wants to go through the options and decide what value is "strictest" for each of them I would guess the actual implementation of it should be relatively simple.

PR #7759, opened 2022-03-06
Allow null to ArrayAccess::offsetSet $offset param

franmomu:
`$offset` can be null: https://www.php.net/manual/en/arrayaccess.offsetset.php#refsect1-arrayaccess.offsetset-notes

related to https://github.com/doctrine/collections/pull/296#discussion_r820219863

**Update**: There was a test checking the `NullArgument` error 🤔

### DIFF
--- a/stubs/CoreGenericClasses.phpstub
+++ b/stubs/CoreGenericClasses.phpstub
@@ -101,7 +101,7 @@ interface ArrayAccess {
      * Offset to set
      * @link http://php.net/manual/en/arrayaccess.offsetset.php
      *
-     * @param TKey $offset The offset to assign the value to.
+     * @param TKey|null $offset The offset to assign the value to.
      * @param TValue $value The value to set.
      * @return void
      *
--- a/tests/ArrayAssignmentTest.php
+++ b/tests/ArrayAssignmentTest.php
@@ -1735,6 +1735,24 @@ class ArrayAssignmentTest extends TestCase
                 [],
                 '8.1'
             ],
+            'allowsArrayAccessNullOffset' => [
+                '<?php
+                    /**
+                     * @template-implements ArrayAccess<int, string>
+                     */
+                    class C implements ArrayAccess {
+                        public function offsetExists(int $offset) : bool { return true; }
+
+                        public function offsetGet($offset) : string { return "";}
+
+                        public function offsetSet(int $offset, string $value) : void {}
+
+                        public function offsetUnset(int $offset) : void { }
+                    }
+
+                    $c = new C();
+                    $c[] = "hello";'
+            ],
         ];
     }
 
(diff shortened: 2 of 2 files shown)


orklah:
I'm not sure about that. If the class declares a `ArrayAccess<int, string>`, it seems normal you're warned about sending null as an offset. I agree PHP allow it, but the template in this case should prevent it

orklah:
However, you should be free to add null in your template if you want to allow this

derrabus:
`null` has to be allowed. If you perform an append operation on an instance of your array access implementation, PHP itself will call `offsetSet()` with `null` as `$offset`.

https://3v4l.org/so8RC

derrabus:
I'd like to elaborate my point a bit more. For me, `ArrayAccess` is a contract provided by the PHP runtime. The methods on that interface are called by the PHP runtime. It is a bit unfortunate that the operations "set a key/value pair" and "append a value" are mapped to the same method, but well, that's how it is.

A very simple example of an `ArrayAccess` implementation that just wraps an array:

https://psalm.dev/r/c1f0e09e4f

Two things happen here:
* The code actually checks if an append operation is performed which is a pretty sane thing to do. But Psalm complains about this very check. The developer is lured into writing fragile code.
* The actual append operation is marked with an error because `null` is not a valid offset.

Now, your solution is to add `null` to the union of allowed keys. Let's do that:

https://psalm.dev/r/54d8e8ec18

The errors that we see now are bogus because Psalm now complains if you don't take into account `null` as a possible value for `$offset` in all the other methods. However, PHP will never call those methods with `null`.

So in other words, you can't do it right. If you want to build an `ArrayAccess` implementation that actually supports appending values, you have no choice but to work around false-positive Psalm errors.

orklah:
You have a point.

Then should we say the offset from `offsetSet` is unrelated to the template completely?

because as you said, the implementation behind may be anything. Here we interpret any null value as an append that will make an integer key, but what about appending on a `ArrayAccess<string, string>` or worse, an `ArrayAccess<MyObject, string>`?

Maybe TKey doesn't belong here at all?

derrabus:
> Then should we say the offset from `offsetSet` is unrelated to the template completely?

No, I believe the propsed change is correct. The expected type is `TKey|null`:

* `TKey` when setting a value for a given key, e.g. `$myObject[42] = 'foo';`
* `null` on append operations, e.g. `$myObject[] = 'foo';`

> because as you said, the implementation behind may be anything. Here we interpret any null value as an append that will make an integer key, but what about appending on a `ArrayAccess<string, string>` or worse, an `ArrayAccess<MyObject, string>`?

It's up to the implementation to chose an appropriate key. The contract does not say that I have to perform an array append. For instance, the key could also be a UUID and an append operation would then generate a new UUID for the element that is about to be added.

However, you are right that there are scenarios where determining the key is not possible and thus the append operation would cause an error. This is the case for PHP's `SplObjectStorage` for instance: https://3v4l.org/dtGgE

If we merged this PR, it wouldn't be possible anymore to tell the static analyzer that the append operation is **not** supported on my custom `ArrayAccess` implementation. For `SplObjectStorage` for instance, Psalm complains when performing an append operation which is probably good because that operation is doomed to fail.

So, it looks like, there is no good solution. 😞

orklah:
Let's try this, we'll see if there are issue, but if necessary, we can just drop the TKey as a param. All it should do is drop the check that makes sure all values sent are consistent with the template, but as we saw, you could send any value to offsetSet and make the implementation change the param before adding to the array so it could make sense...

PR #7760, opened 2022-03-06
Fix #7750: array_column type inference

### LINKED ISSUE #7750: array_column type inference
`array_column([['a' => 1], ['a' => 2]], null, 'a')` should infer a return type of `array<int, array{a: int}>`.

mtk3d:
My proposition of fixing this problem https://github.com/vimeo/psalm/issues/7750
I was trying to keep the similar tests together, so this change looks a bit unclear in diff.

That's my first contribution to the psalm, so sorry if something is wrong 😄

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayColumnReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayColumnReturnTypeProvider.php
@@ -77,6 +77,7 @@ class ArrayColumnReturnTypeProvider implements FunctionReturnTypeProviderInterfa
         }
 
         $value_column_name = null;
+        $value_column_name_is_null = false;
         // calculate value column name
         if (($second_arg_type = $statements_source->node_data->getType($call_args[1]->value))) {
             if ($second_arg_type->isSingleIntLiteral()) {
@@ -84,6 +85,7 @@ class ArrayColumnReturnTypeProvider implements FunctionReturnTypeProviderInterfa
             } elseif ($second_arg_type->isSingleStringLiteral()) {
                 $value_column_name = $second_arg_type->getSingleStringLiteral()->value;
             }
+            $value_column_name_is_null = $second_arg_type->isNull();
         }
 
         $key_column_name = null;
@@ -114,7 +116,7 @@ class ArrayColumnReturnTypeProvider implements FunctionReturnTypeProviderInterfa
                 }
                 //array_column skips undefined elements so resulting type is necessarily defined
                 $result_element_type->possibly_undefined = false;
-            } elseif ($value_column_name === null) {
+            } elseif ($value_column_name_is_null) {
                 $result_element_type = new Union([$row_shape]);
             } else {
                 $result_element_type = Type::getMixed();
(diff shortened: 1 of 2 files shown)


alex-dev:
Why `new Union($row)`  rather than just `$row` itself?

mtk3d:
You're right @alex-dev, fixed 😃

src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayColumnReturnTypeProvider.php
```diff
@@ -114,6 +114,8 @@ public static function getFunctionReturnType(FunctionReturnTypeProviderEvent $ev
                 }
                 //array_column skips undefined elements so resulting type is necessarily defined
                 $result_element_type->possibly_undefined = false;
+            } elseif ($value_column_name === null) {
+                $result_element_type = $row_shape;
```
orklah, line 118: $result_element_type should be a Union, $row_shape is an Atomic (a part of a Union). `new Union([$row_shape]);` should be ok

mtk3d: So my reasoning was right after all 😄 
Thank you I fixed that 💪

src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayColumnReturnTypeProvider.php
```diff
@@ -114,6 +114,8 @@ public static function getFunctionReturnType(FunctionReturnTypeProviderEvent $ev
                 }
                 //array_column skips undefined elements so resulting type is necessarily defined
                 $result_element_type->possibly_undefined = false;
+            } elseif ($value_column_name === null) {
```
orklah, line 117: One last thing, the fact that value_column_name is null here doesn't mean the second arg is null. It just means it wasn't a single Integer not a single String.
(For example, this test will fail `$d = array_column([["a" => 1], ["a" => 2], ["a" => 3]], rand(0,1)? 'a': 'b', "b");` with your code)

Could you create a new var value_column_name_is_null and set to true on line 86 when `$second_arg_type->isNull()` is true?

src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayColumnReturnTypeProvider.php
alex-dev: I don't understand why? `$result_element_type ` should be `$row_shape`. How is wrapping it in a single element union useful?

AndrolGenhald: @alex-dev This is a minor weirdness in the way Psalm handles types, right now everything is a `Union` which has one or more `Atomic`s. I've been slowly working on changing it so that types can be `Atomic`s or `Union`s or `Intersection`s, but it's a long endeavor.

src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayColumnReturnTypeProvider.php
mtk3d: Sure
I fixed this, and also add a test for that case

src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayColumnReturnTypeProvider.php
alex-dev: I see. Thank for the explanation.

orklah:
That's great! Thanks!

Congrats for your first PR!

PR #7775, opened 2022-03-11
Improve int range parsing

### LINKED ISSUE #7734: Psalm treats any string literal as a "max" in an int range
https://psalm.dev/r/192f071f25

I think it would be better to emit `InvalidDocblock` on line 7.

fluffycondor:
Fixes #7734

### DIFF
--- a/src/Psalm/Internal/Type/TypeParser.php
+++ b/src/Psalm/Internal/Type/TypeParser.php
@@ -863,6 +863,10 @@ class TypeParser
                 return new TPositiveInt();
             }
 
+            if (is_int($min_bound) && is_int($max_bound) && $min_bound > $max_bound) {
+                throw new TypeParseTreeException("Min bound can't be greater than max bound");
+            }
+
             return new TIntRange($min_bound, $max_bound);
         }
 
--- a/src/Psalm/Internal/Type/TypeParser.php
+++ b/src/Psalm/Internal/Type/TypeParser.php
@@ -864,7 +864,7 @@ class TypeParser
             }
 
             if (is_int($min_bound) && is_int($max_bound) && $min_bound > $max_bound) {
-                throw new TypeParseTreeException("Min bound can't be greater than max bound");
+                throw new TypeParseTreeException("Min bound can't be greater than max bound, int<" . $min_bound . "," . $max_bound . "> given");
             }
 
             return new TIntRange($min_bound, $max_bound);
--- a/src/Psalm/Internal/Type/TypeParser.php
+++ b/src/Psalm/Internal/Type/TypeParser.php
@@ -837,12 +837,12 @@ class TypeParser
 
             if ($param0_union_types[0] instanceof TNamedObject
                 && $param0_union_types[0]->value !== TIntRange::BOUND_MIN) {
-                throw new TypeParseTreeException('Unknown named object as a min boundary');
+                throw new TypeParseTreeException('Incorrect named object as a min boundary');
             }
 
             if ($param1_union_types[0] instanceof TNamedObject
                 && $param1_union_types[0]->value !== TIntRange::BOUND_MAX) {
-                throw new TypeParseTreeException('Unknown named object as a max boundary');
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Type/TypeParser.php
```diff
@@ -830,19 +830,24 @@ private static function getTypeFromGenericTree(
...
-                throw new TypeParseTreeException("max bound for int range param can't be 'min'");
+
+            if ($param0_union_types[0] instanceof TNamedObject
+                && $param0_union_types[0]->value !== TIntRange::BOUND_MIN) {
+                throw new TypeParseTreeException('Unknown named object as a min boundary');
+            }
+
+            if ($param1_union_types[0] instanceof TNamedObject
+                && $param1_union_types[0]->value !== TIntRange::BOUND_MAX) {
+                throw new TypeParseTreeException('Unknown named object as a max boundary');
```
AndrolGenhald, line 845: I would change these messages, [here](https://psalm.dev/r/30ef078e0f) `Foo` isn't an unknown named object, it's known it's just incorrect.

src/Psalm/Internal/Type/TypeParser.php
```diff
@@ -858,6 +863,10 @@ private static function getTypeFromGenericTree(
                 return new TPositiveInt();
             }
 
+            if (is_int($min_bound) && is_int($max_bound) && $min_bound > $max_bound) {
+                throw new TypeParseTreeException("Min bound can't be greater than max bound");
+            }
```
AndrolGenhald, line 868: [Wow](https://psalm.dev/r/4cdae92b1d), that check didn't exist before? :laughing:

I'm pretty sure the `CodeLocation` used for this error is going to be the entire type rather than just the int range, so this might be slightly confusing for something like `array<array-key, int<5, 3>>`, maybe mention `int<$min_bound, $max_bound>` in the message?

orklah: Guilty as charged! (In fact, there's a few other places an IntegerRange could end up with inconsistent bounds due to bad checks). I briefly considered making TIntegerRange immutable in order to set up this check in constructor, but that was making a lot of code needlessly complex

AndrolGenhald: @orklah otoh if it was immutable [this](https://github.com/vimeo/psalm/issues/7697) difficult to debug issue would never have happened. I'd love for the type system to become immutable eventually, but I haven't really looked too hard at what it would take to make that happen.

orklah: IIRC, the worst would be Union manipulation. You won't be able to change an Atomic state just by having the instance, you have to actually know its key and replace it inside the Union

src/Psalm/Internal/Type/TypeParser.php
fluffycondor: Fixed!

src/Psalm/Internal/Type/TypeParser.php
fluffycondor: > this might be slightly confusing for something like array<array-key, int<5, 3>>, maybe mention int<$min_bound, $max_bound> in the message

You're right. Done!

AndrolGenhald (approved):
LGTM!

orklah:
Thanks!

orklah:
@fluffycondor the previous code was suboptimal and your enhancement made it apparent. Turns out `$param1_union_types[0]->value` contains the current namespace along with `min` or `max`.

For example, `Psalm\max`. It showed when we tried to merge this onto master (who's using more integer ranges in Psalm code itself).

This made parsing of `int<0, max>` fail when we're in a namespace. Could you take a look at it and pushing a fix? :)

AndrolGenhald:
I think the best way to go about this would be to use `$parse_tree->children` directly when doing the `BOUND_MIN/MAX` comparison, but I haven't looked into it too deeply.

Issue #7771, opened 2022-03-10
JsonException Psalm\Internal\Clause


ohader:
> Uncaught JsonException: Malformed UTF-8 characters, possibly incorrectly encoded in src/Psalm/Internal/Clause.php:124
https://github.com/vimeo/psalm/blob/0a81f8cd6cfb4d6db37b9d29c0e7b51d5af0f498/src/Psalm/Internal/Clause.php#L124

using `vimeo/psalm dev-master` (a0739b1), probably related to 0a81f8cd6cfb4d6db37b9d29c0e7b51d5af0f498

#### The payload of `$possibility_strings` at that time is

```
array(1) {
  ["$signature"]=>
  array(1) {
    [0]=>
    string(18) "!~string(��ࡱ�)"
  }
}
```

and the parsed file at that time was (similar to) https://github.com/PHPOffice/PHPExcel/blob/1.8.0/Classes/PHPExcel/Shared/OLE.php

#### Stack Trace

```
Stack trace:
#0 src/Psalm/Internal/Clause.php(124): json_encode(Array, 4194304)
#1 src/Psalm/Internal/Algebra/FormulaGenerator.php(436): Psalm\Internal\Clause->__construct(Array, 4922980, 4922980, false, true, true, Array)
#2 src/Psalm/Internal/Analyzer/Statements/Block/IfElseAnalyzer.php(135): Psalm\Internal\Algebra\FormulaGenerator::getFormula(4922980, 4922980, Object(PhpParser\Node\Expr\BinaryOp\NotEqual), 'PHPExcel_Shared...', Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(Psalm\Codebase))
#3 src/Psalm/Internal/Analyzer/StatementsAnalyzer.php(534): Psalm\Internal\Analyzer\Statements\Block\IfElseAnalyzer::analyze(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(PhpParser\Node\Stmt\If_), Object(Psalm\Context))
#4 src/Psalm/Internal/Analyzer/StatementsAnalyzer.php(212): Psalm\Internal\Analyzer\StatementsAnalyzer::analyzeStatement(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(PhpParser\Node\Stmt\If_), Object(Psalm\Context), Object(Psalm\Context))
[...]

PR #7790, opened 2022-03-15
Avoid json_encode when serializing non-UTF8 literals

ohader:
`\Psalm\Internal\Clause::$hash` basically holds a hash on arbitrary input literals, used for later comparison. Using `json_encode` fails when dealing with non-UTF8 literals, which has been replaced by plain PHP `serialize`.

Resolves: #7771

### DIFF
--- a/src/Psalm/Internal/Clause.php
+++ b/src/Psalm/Internal/Clause.php
@@ -15,13 +15,11 @@ use function array_map;
 use function array_values;
 use function count;
 use function implode;
-use function json_encode;
 use function ksort;
 use function md5;
 use function reset;
+use function serialize;
 use function substr;
-
-use const JSON_THROW_ON_ERROR;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


ohader:
> **q.e.d.** https://github.com/vimeo/psalm/runs/5547432542?check_suite_focus=true#step:7:291

orklah:
Thanks for the quick fix!

You may want to rebase on 4.x though. Master branch will be for Psalm 5 which has no stable release yet so you'd get your fix quicker on 4.x :)

ohader:
Just spotted `array<string, non-empty-array<string, Assertion>> $possibilities` - probably it should stay `json_encode`, but with some additional handling, e.g. `bin2hex`.

orklah:
I'll close this. #7791 will be merged on 4.x and when I'll merge 4.x onto master, it will result in a conflict that I'll fix by keeping the code made on 4.x (instead of the json_encode with the flag to throw errors)

PR #7791, opened 2022-03-15
Avoid json_encode when serializing non-UTF8 literals

ohader:
`\Psalm\Internal\Clause::$hash` basically holds a hash on arbitrary input literals, used for later comparison. Using `json_encode` fails when dealing with non-UTF8 literals, which has been replaced by plain PHP `serialize`.

Resolves: #7771

### DIFF
--- a/tests/TypeReconciliation/ConditionalTest.php
+++ b/tests/TypeReconciliation/ConditionalTest.php
@@ -2839,6 +2839,15 @@ class ConditionalTest extends TestCase
                         echo "false";
                     }',
             ],
+            'BinaryDataIsPassedSeeIssue7771' => [
+                'code' => '<?php
+                    function matches(string $value): bool {
+                        if ("\xD0\xCF\x11\xE0\xA1\xB1\x1A\xE1" !== $value) {
+                            return false;
+                        }
+                        return true;
+                    }'
+            ],
         ];
     }
 
--- a/tests/TypeReconciliation/ConditionalTest.php
+++ b/tests/TypeReconciliation/ConditionalTest.php
@@ -2839,8 +2839,8 @@ class ConditionalTest extends TestCase
                         echo "false";
                     }',
             ],
-            'BinaryDataIsPassedSeeIssue7771' => [
-                'code' => '<?php
+            '#7771: non-UTF8 binary data is passed' => [
+                '<?php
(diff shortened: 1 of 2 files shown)


ohader:
This change does not actually crash in the 4.x branch - errors in `json_encode` are just ignored, results casted to `string` and "hashed" - always to `d41d8cd98f00b204e9800998ecf8427e`, which is `md5('')`. Thus, in 4.x it is a fix, but not having the same severity as in master branch.

tests/TypeReconciliation/ConditionalTest.php
```diff
@@ -2839,6 +2839,15 @@ function b(B $_b): void {
                         echo "false";
                     }',
             ],
+            'BinaryDataIsPassedSeeIssue7771' => [
+                'code' => '<?php
```
orklah, line 2843: tests format changed between 4.x and master. removing `'code' => ` should fix the build

orklah:
> This change does not actually crash in the 4.x branch

This is interesting because it means those hash are returned as false on 4.x. Meaning whatever use we make of those hash could be completely bogus (for example when two weird encoded string need to be hashed)

This is a potential bug revealed by @weirdan applying a rector rule here: https://github.com/vimeo/psalm/pull/7303 

/cc @TomasVotruba :)

TomasVotruba:
I'd go with `Nette\Utils\Json`. The native function cannot be trusted :)

ohader:
Unfortunately `Nette\Utils\Json` does not provide the require magic, it's also just `json_encode` in the end.. 😉 
→ https://github.com/nette/utils/blob/master/src/Utils/Json.php#L48

Seems I need to check JSON specs carefully on that topic, since in JavaScript the following seems to be fine...
```js
JSON.stringify(String.fromCharCode(255))
```

orklah:
If serialize is working fine, I'm good with it :)

AndrolGenhald:
There is [`JSON_INVALID_UTF8_IGNORE`](https://3v4l.org/HC3ZO), but it [appears](https://3v4l.org/dMbtA) to just be stripping invalid UTF-8 characters.

ohader:
ℹ️ Merged to `master` branch in 20351c6667e4f4e2443ea9aca6d5cde610aaaf8d

PR #7798, opened 2022-03-17
Add return shape for Throwable::getTrace

ciaranmcnulty:
This is based on the documentation for debug-backtrace:
   https://php.net/debug_backtrace

Experimentation that shows the `object` field is not populated for exceptions:
   https://3v4l.org/iQoni

And validation on psalm.dev:
   https://psalm.dev/r/b35a8df2f3

Should I also update the stubs?
What tests should I add?


ciaranmcnulty:
Updated to allow optional `args` (in the case of an included file)

https://psalm.dev/r/11247acbc9

orklah:
Seems great!

You'll have to change Callmap_Historical too for tests to pass

ciaranmcnulty:
@orklah should I update /stubs too? I'm not sure the connection between the two things

AndrolGenhald:
Yes, stubs should be updated too. CallMap is there because it's much faster to parse for simple functions, but stubs exist to add annotations that don't work in the CallMap, so the stubs override the CallMap.

orklah (approved):

ciaranmcnulty:
I rebased onto 4.x

I'm not sure what's happened there with Appveyor?

ciaranmcnulty:
Closed/reopened to try and nudge appveyor :/

orklah:
Appveyor was removed from our CI ages ago, but sometimes on rebase it pops again, not sure why. As long as the rest of the CI runs, it's good

orklah:
Thanks!

dictionaries/CallMap.php
```diff
@@ -346,7 +346,7 @@
 'ArgumentCountError::getLine' => ['int'],
 'ArgumentCountError::getMessage' => ['string'],
 'ArgumentCountError::getPrevious' => ['?Throwable'],
-'ArgumentCountError::getTrace' => ['list<array<string,mixed>>'],
+'ArgumentCountError::getTrace' => ['list<array{file:string,line:int,function:string,class?:string,type?:\'::\'|\'->\',args?:array<mixed>}>'],
```
staabm, line 349: per https://github.com/phpstan/phpstan-src/pull/914#issuecomment-1007370554 `file` needs to be nullable

staabm: and `class` can be narrowe to `class-string`

AndrolGenhald: `line` is optional as well apparently.

PR #7801, opened 2022-03-18
Fix disabling of appveyor auto build

gndk:
This is an attempt to fix the unexplained appveyor builds that @orklah mentioned here: https://github.com/vimeo/psalm/pull/7798#issuecomment-1072311603

According to the [appveyor.yml reference](https://www.appveyor.com/docs/appveyor-yml/) (search for "disable automatic builds"), this is the correct way to disable builds.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-build: false
+build: off
 branches:
   only:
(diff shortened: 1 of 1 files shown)


orklah:
Thanks! Let's see if the issue is solved with that :)

PR #7812, opened 2022-03-23
Fix int-mask-of doc example

Seldaek:
I hope this is correct.. it seems to be off in a few places, so it made me doubt.. but from the code it seems like a correct fix.


noplanman:
There's also a double `with` in there.

orklah:
Thanks!

PR #7824, opened 2022-03-28
Fix ftp_fget/_fput/_nb_fget/_nb_fput stream arg type

mrardon:
Here is the how psalm behaves with `ftp_fget`: https://psalm.dev/r/cebd1f1758

I believe the error received is incorrect per PHP docs: ```Argument 2 of ftp_fget expects FTP\Connection, resource provided```. 

Based on PHP's documentation for `ftp_fget` the Argument 2 is still a `resource` in 8.1: https://www.php.net/manual/en/function.ftp-fget.php

The `$stream` argument types were set to `FTP\Connection` types when they should be `resource` types. I see the PR where this was introduced: https://github.com/vimeo/psalm/pull/6421 and it appears  `ftp_fget`/`ftp_fput`/`ftp_nb_fget`/`ftp_nb_fput` all have the same issue. This PR updates the call maps of those 4 functions by changing the `FTP\Connection $stream` back to a `resource`.


orklah:
Thanks!

PR #7828, opened 2022-03-30
Psalm considers most readline functions as pure, but they have side effects

mstilkerich:
Psalm considers most of the readline functions as pure, but all readline functions have side effects.

This leads to spurious UnusedFunctionCall errors as can be seen here:
https://psalm.dev/r/eb33a31483

This PR makes psalm treat all readline functions as impure.


AndrolGenhald:
Interesting that that's implemented by blacklisting impure functions rather than whitelisting pure functions. I guess there are probably a lot more pure functions than impure functions in the library, but considering a pure function impure isn't as bad of a failure as considering an impure function pure.

LGTM!

orklah:
I can merge this, but as @AndrolGenhald said, you better target 4.x for that. If you don't care, ping me and I'll merge anyway :)

mstilkerich:
Oops, changed the target but now there is tons of commits. Shall I create a new PR that is also based on 4.x instead of master?

mstilkerich:
Thanks, I force pushed the commit based on 4.x. One of the CI checks fails, but it seems unrelated to my change.

orklah:
Thanks!

PR #7829, opened 2022-03-30
Fix namespaced min/max int range keyword issue introduced in #7775

AndrolGenhald:
Figured I'd better get around to this so we can do a new 4.x release without breaking a bunch of stuff.


orklah:
Oh, awesome, it was on my todolist for a week or two but I couldn't find the time? I'll check that tonight

AndrolGenhald:
> it was on my todolist for a week or two but I couldn't find the time

Yeah same :laughing: I was secretly hoping fluffycondor would come back and fix it up so I could ignore it.

orklah:
Cool! Thanks!

Issue #7825, opened 2022-03-28
Incompatible types found for T (Stub&ProxyQueryInterface is not in ProxyQueryInterface&Stub)


VincentLanglet:
Hi @orklah, I got a pretty weird error messages.

In https://psalm.dev/r/f5569873e0, the error
```
Incompatible types found for T (Stub&ProxyQueryInterface is not in ProxyQueryInterface&Stub)
```
shouldn't be reported IMHO.

orklah:
Yeah, this check could probably be improved: https://github.com/vimeo/psalm/blob/3a298d028e949516920cdd54204cf0d98066d14f/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php#L1088

VincentLanglet:
> Yeah, this check could probably be improved:
> 
> https://github.com/vimeo/psalm/blob/3a298d028e949516920cdd54204cf0d98066d14f/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php#L1088

I dunno if it would be easier but what if types were alphabetically sorted ?
Even if some phpdoc is `Stub&ProxyQueryInterface`, psalm would save the type as `ProxyQueryInterface&Stub`.

orklah:
For now, intersections are not just an array of objects. It's a property `$extra_types` added by `HasIntersectionTrait`. So, in one case you have a TNamedObject `stub` with an extra `ProxyQueryInterface` and in the other case, you got a `ProxyQueryInterface` with an extra `stub`. This should probably be reworked to have a real Intersection between types, but this is not a small task

AndrolGenhald:
> It's a property `$extra_types` added by `HasIntersectionTrait`

Oh right, that's why it's broken. I was wondering, because I know we sort most things like this.

> This should probably be reworked to have a real Intersection between types, but this is not a small task

No kidding, hopefully I can find some time to work on my attempt again but there are a few other things that need done first :confused:

VincentLanglet:
WDYT of this solution: https://github.com/vimeo/psalm/pull/7838 @orklah @AndrolGenhald ?

PR #7838, opened 2022-04-02
Fix Incompatible types found for T (Stub&ProxyQueryInterface is not in ProxyQueryInterface&Stub)

VincentLanglet:
I tried this on my project and
- my issue was solved.
- I saw no extra errors.

Closes https://github.com/vimeo/psalm/issues/7825

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php
@@ -1137,18 +1137,30 @@ class CallAnalyzer
                     } else {
                         foreach ($lower_bounds as $lower_bound) {
                             if ($lower_bound->equality_bound_classlike === null) {
-                                if (!in_array($lower_bound->type->getId(), $equality_types, true)) {
-                                    IssueBuffer::maybeAdd(
-                                        new InvalidArgument(
-                                            'Incompatible types found for ' . $template_name . ' (' .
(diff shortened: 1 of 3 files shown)


orklah:
I'm not sure about that one. It fixes the issue because the check was made by checking if the type display is the same, so when you fix the display, it just sweep the dust under the rug, but the issue with the suboptimal check is still there and the structural issue with intersections not behaving the same way when turned the other way around is still there too

AndrolGenhald:
Yeah, I agree with @orklah, I know we use IDs for type equality checks sometimes but it doesn't seem like a great idea. There's also `Union::equals` but that doesn't do what you'd expect either. I think for now the best solution is to change it to 2 `UnionTypeComparator::isContainedBy` checks, checking that each type is contained by the other.

VincentLanglet:
> I think for now the best solution is to change it to 2 `UnionTypeComparator::isContainedBy` checks, checking that each type is contained by the other.

I updated the PR then

VincentLanglet:
Is it ok like this @orklah ? :)

orklah:
Looks good!

Can you add your snippet in tests please?

VincentLanglet:
> Looks good!
> 
> Can you add your snippet in tests please?

Sure ! ... but in which file ?

orklah:
ClassTemplateTest should be fine :)

VincentLanglet:
> ClassTemplateTest should be fine :)

Done :)

orklah:
Thanks!

Issue #7843, opened 2022-04-05
`@psalm-assert lowercase-string` on a given `non-empty-string` does not infer a `non-empty-lowercase-string`


Ocramius:
See https://psalm.dev/r/c808c16b27

```php
<?php

/** @psalm-assert lowercase-string $input */
function assertLowerCase(string $input): void { throw new \Exception($input . ' irrelevant'); }

/** 
 * @param non-empty-string $input
 * @return non-empty-lowercase-string
 */
function makeLowerNonEmpty(string $input): string
{
    assertLowerCase($input);

    return $input;
}
```

```
Psalm output (using commit cab6f33): 

ERROR: [TypeDoesNotContainType](https://psalm.dev/056) - 12:5 - Cannot resolve types for $input - non-empty-string does not contain lowercase-string

INFO: [LessSpecificReturnStatement](https://psalm.dev/129) - 14:12 - The type 'lowercase-string' is more general than the declared return type 'non-empty-lowercase-string' for makeLowerNonEmpty

INFO: [MoreSpecificReturnType](https://psalm.dev/070) - 8:12 - The declared return type 'non-empty-lowercase-string' for makeLowerNonEmpty is more specific than the inferred return type 'lowercase-string'
```

Few things are a bit weird here:

 1. `@psalm-assert lowercase-string` reports `Cannot resolve types for $input - non-empty-string does not contain lowercase-string`, indicating that they somehow cannot be intersected?
 2. the type is then not restricted, but instead replaced with `lowercase-string`

PR #7844, opened 2022-04-05
fix refining lowercase string and non-empty-string together

orklah:
This should fix https://github.com/vimeo/psalm/issues/7843 (on Psalm 5 though...)

For the explanation: Psalm assertions works well when one of the two type is a super type of another (isContainedBy method). When that happens, Psalm just takes the narrower type and goes on. Issues arise when none of the type is completely included in another. (A great example is the condition just above with `int<0, 10>` and `int<5, 15>` which should result in `int<5, 10>` but need to be handled separately).

This is the same with non-empty-string and lowercase-string, the intersection of both is possible within Psalm's type system (non-empty-lowercase-string) but none is completely included in the other


orklah:
Failure is unrelated, I'll have to check that later

Ocramius (commented):
Nice!

PR #7851, opened 2022-04-07
Document `@psalm-ignore-variable-*`

jrmajor:

PR #7849, opened 2022-04-07
Document `@psalm-yield`

jrmajor:
Fixes #5397

PR #7853, opened 2022-04-09
Remove `@psalm-generator-return` annotation

jrmajor:
While trying to shrink `DocumentationTest::WALL_OF_SHAME` (#7849, #7851), I stumbled upon `@psalm-generator-return` annotation.

The only references to it I could find are the aforementioned wall of shame, `DocComment::PSALM_ANNOTATIONS` and `GeneratorTest`:

https://github.com/vimeo/psalm/blob/eeb12bc285b3ebfeac7f6f8328a2917a2b0a1225/tests/GeneratorTest.php#L56-L69

I tried adding assertions to this test, and inferred types don't seem to be affected by `@psalm-generator-return` presence (I checked `fooFoo()` return type and `fooFoo()->getReturn()`).

The oldest reference to this annotation I can find is d71d439e25a44d5c2196187e84f0f7c38ed7ec5d (late 2016).

My guess is that at that point Psalm did not support `TSend` and `TReturn` for generators yet, so this annotation was meant to be a replacement for the latter, which ended up not being implemented.

I'm pretty sure that this annotation is no-op and can be safely removed. The only effect of this change should be throwing an unknown annotation error, that's why it's targeted at `master`.

### DIFF
--- a/src/Psalm/DocComment.php
+++ b/src/Psalm/DocComment.php
@@ -24,7 +24,7 @@ final class DocComment
         'assert', 'assert-if-true', 'assert-if-false', 'suppress',
         'ignore-nullable-return', 'override-property-visibility',
         'override-method-visibility', 'seal-properties', 'seal-methods',
-        'generator-return', 'ignore-falsable-return', 'variadic', 'pure',
+        'ignore-falsable-return', 'variadic', 'pure',
         'ignore-variable-method', 'ignore-variable-property', 'internal',
         'taint-sink', 'taint-source', 'assert-untainted', 'scope-this',
         'mutation-free', 'external-mutation-free', 'immutable', 'readonly',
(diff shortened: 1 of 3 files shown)


AndrolGenhald:
> The only effect of this change should be throwing an unknown annotation error, that's why it's targeted at master.

This sort of thing should be safe to do on 4.x. We should probably make it more clear, but Psalm provides no compatibility guarantees as far as issue messages, issues detected, etc between minor versions (and indeed a lot of the minor releases include newly detected issues and fixed false positives).

Major changes should target 5.x since there are enough changes there to make larger features difficult to merge, and anything breaking compatibility with the plugin API (including any non-internal classes) needs to target 5.x because we _do_ (try to) keep the plugin API compatible in minor releases.

jrmajor:
> Psalm provides no compatibility guarantees as far as issue messages, issues detected, etc between minor versions

I know, but deleting an annotation (even if it's a broken one) feels like something more suitable for a major version :)
I can rebase this PR onto `4.x` if you prefer that.

orklah:
Thanks!

PR #7859, opened 2022-04-11
Resolve #7855 remove false from return type, in php8.0 and further ve…

### LINKED ISSUE #7855: Invalid inferred return type of `DateTimeImmutable::getTimestamp`
Hello I've come across an issue while constructing an array with timestamps used as keys, psalm is reporting `false|int` can't be used as array key while `getTimestamp` always returns an `int`.

e.g. https://psalm.dev/r/c45a9a0a0f

dkemper:
…rsions there is a ValueError returned instead


AndrolGenhald:
It looks like you missed `DateTimeImmutable` in CallMap.php. Could you also check the other methods and see if they need updated? It looks like at least `diff` and `setTime` have the same issue, and there might be one or two more that I missed.

orklah:
Thanks!

PR #7861, opened 2022-04-12
Disable filepath formatting as a link for Drone CI's output

white43:
https://docs.drone.io/pipeline/environment/reference/

It seems, default console output was changed some time ago. It had been clear and easy-to-read, but then it turned into something like this:

```
ERROR: MissingReturnType - ]8;;file:///path/to/File.php#L9\src\File.php:9:30]8;;\ - Method App\File::canBuild does not have a return type (see https://psalm.dev/050)
        abstract public function canBuild();
```

Using `--long-progress` CLI option doesn't help, unfortunately.


drupol:
FYI, this package might be helpful: https://github.com/OndraM/ci-detector

PR #7866, opened 2022-04-13
Fix openssl_csr_export() signature

mpesari:
The documentation for [openssl_csr_export()](https://www.php.net/manual/en/function.openssl-csr-export.php) was [updated](https://github.com/php/doc-en/commit/3a2e96fcc0bd83370c3a4318aa773783a87110f8) recently, although it does not seem to be live yet.

This fixes the 2nd parameter `$output` of `openssl_csr_export()` signature in Psalm.

Example of what this should fix (there should be no errors): https://psalm.dev/r/03755de763


orklah:
Thanks!

PR #7867, opened 2022-04-13
Documented the allowMissingFiles attribute for ignoreFiles

mmcev106:
Let me know if you'd prefer this be documented differently in any way.


orklah:
Could you maybe add a description of what the attribute does? Seems clear from the name, but it never hurts to have it spelled out :)

mmcev106:
Done.  How does that look?

orklah:
Great! Thanks!

Issue #7873, opened 2022-04-16
performance: parsing files multiple times


vognev:
I've noticed, that analyzing code which is using tcpdf library (example here - https://github.com/vognev/psalmbug) 
is taking significant time.

with debug logging enabled it is seen that single `tcpdf.php` is parsed more than 150 times
https://github.com/vognev/psalmbug/blob/master/debug.log

still not fully understanding if that is really needed, but my question is about if it is possible to reduce analysis time in this case?
maybe by using/introducing volatile internal cache for reusing parsing results within single analysis run

vognev:
by quickly hacking `src/Psalm/Internal/Provider/StatementsProvider.php` in following way:

```
        $file_content_hash = md5($version . $file_contents);

        if (!$this->parser_cache_provider
            || (!$config->isInProjectDirs($file_path) && strpos($file_path, 'vendor'))
        ) {
            static $__cache = [];

            if (array_key_exists($file_content_hash, $__cache)) {
                return $__cache[$file_content_hash];
            }

            $progress->debug('Parsing ' . $file_path . "\n");

            $has_errors = false;

            $stmts = self::parseStatements($file_contents, $php_version, $has_errors, $file_path);

            return $__cache[$file_content_hash] = $stmts ?: [];
        }
```

makes analysis run in about 12 seconds

maybe you can accept a PR containing this small fix or suggest how to implement properly that kind of volatile cache (LRU maybe)

AndrolGenhald:
Wow, I'm surprised no one has noticed that before! We'll need to make sure the cache is appropriately invalidated for the language server, and make sure it doesn't have any weird interactions with the existing cache, but that's a fantastic improvement!

vognev:
guess can be closed

PR #7876, opened 2022-04-18
cache statements even without persistent parser cache

vognev:
Attempt to improve performance issue reported in #7873

### DIFF
--- a/src/Psalm/Internal/Provider/StatementsProvider.php
+++ b/src/Psalm/Internal/Provider/StatementsProvider.php
@@ -56,6 +56,11 @@ class StatementsProvider
     private $file_storage_cache_provider;
 
     /**
+     * @var ?VolatileCacheProvider
+     */
+    private $volatile_cache_provider;
+
+    /**
      * @var array<string, array<string, bool>>
      */
     private $unchanged_members = [];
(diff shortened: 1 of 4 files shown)


orklah:
Looks great!

Can you take a look at failures on the "Run Shepherd / build (pull_request)" job? (You can ignore the two first errors, I need to fix those)

orklah:
Great! Thanks!

M1ke:
I am wondering if this is the change responsible for a change of behaviour I am seeing in 4.23.

When running 4.22 on our codebase (6840 PHP files) the result comes back with

```
Checks took 177.73 seconds and used 2,468.652MB of memory
Psalm was able to infer types for 89.8237% of the codebase
```

Whereas running 4.23 returns

```
Checks took 177.55 seconds and used 3,392.045MB of memory
Psalm was able to infer types for 89.8159% of the codebase
```

So a ~35% increase in RAM usage. Not suggesting this is definitely the cause, but from the items listed as released in 4.23 this does appear the obvious place where RAM usage might increase. @vognev I am interested if you saw any increased RAM usage when debugging after the fix (as I know you had used debug logs when investigating what needed fixing)?

vognev:
in my testing, for my codebase (885 files analyzed), we benefited a lot in total time which inspection taken:

```
4.22

Checks took 425.14 seconds and used 517.665MB of memory
Psalm was able to infer types for 82.2386% of the codebase
```

```
55a45a9

Checks took 47.48 seconds and used 1,181.505MB of memory
Psalm was able to infer types for 82.2576% of the codebase
```


```
4.23

Checks took 48.66 seconds and used 1,181.594MB of memory
Psalm was able to infer types for 82.2576% of the codebase
```

yes, used memory amount increased, but caching always needs some memory
on another hand, cache added in `55a45a9`, could contain some means
for controlling it's max size, but i've not found reliable and efficient way for
measuring single variable memory footprint

M1ke:
I have a thought about using a pre-pass to assess what cache hits are likely, and only caching things that will be hit more than once, but potentially that'd just confuse things.

I think I need to see the impact of this on the psalm 5 alpha before I think about whether I'd like to offer a change to disable the caching via a config option.

vognev:
well, i can propose only following improvement then - https://github.com/vognev/psalm/commit/cc4d6350cd9b6b2e8a149523518e90ab796f6a93

it will cache up to 32 heaviest statements, still not memory-constrained though

M1ke:
That looks pretty sharp, I'll run it on our codebase and see what it can
manage. Thanks!

PR #7887, opened 2022-04-22
improve LSP

### LINKED ISSUE #7881: Language Server (didSave): $textDocument->text always seems to be empty
Related to: https://github.com/psalm/psalm-vscode-plugin/issues/149

https://github.com/vimeo/psalm/blob/2724c1dba965c0ad67736af2bcccc642f27c2b09/src/Psalm/Internal/LanguageServer/Server/TextDocument.php#L106

This prevents Psalm from performing analysis when saving the file.

ging-dev:
fixes #7881

### DIFF
--- a/src/Psalm/Internal/LanguageServer/Server/TextDocument.php
+++ b/src/Psalm/Internal/LanguageServer/Server/TextDocument.php
@@ -6,7 +6,6 @@ namespace Psalm\Internal\LanguageServer\Server;
 
 use Amp\Promise;
 use Amp\Success;
-use InvalidArgumentException;
 use LanguageServerProtocol\CompletionList;
 use LanguageServerProtocol\Hover;
 use LanguageServerProtocol\Location;
@@ -27,7 +26,6 @@ use Psalm\Internal\Analyzer\ProjectAnalyzer;
 use Psalm\Internal\LanguageServer\LanguageServer;
 use UnexpectedValueException;
 
-use function array_combine;
 use function array_values;
 use function count;
 use function error_log;
@@ -92,8 +90,9 @@ class TextDocument
      * The document save notification is sent from the client to the server when the document was saved in the client
      *
      * @param TextDocumentItem $textDocument the document that was opened
+     * @param ?string $text the content when saved
      */
-    public function didSave(TextDocumentItem $textDocument): void
+    public function didSave(TextDocumentItem $textDocument, ?string $text): void
     {
         $file_path = LanguageServer::uriToPath($textDocument->uri);
 
@@ -103,7 +102,7 @@ class TextDocument
 
         // reopen file
         $this->codebase->removeTemporaryFileChanges($file_path);
-        $this->codebase->file_provider->setOpenContents($file_path, $textDocument->text);
+        $this->codebase->file_provider->setOpenContents($file_path, (string) $text);
 
         $this->server->queueFileAnalysis($file_path, $textDocument->uri);
     }
@@ -119,10 +118,6 @@ class TextDocument
         $file_path = LanguageServer::uriToPath($textDocument->uri);
 
         if (!$this->codebase->config->isInProjectDirs($file_path)) {
-            return;
-        }
-
-        if ($this->project_analyzer->onchange_line_limit === 0) {
             return;
         }
 
@@ -266,8 +261,6 @@ class TextDocument
      */
     public function completion(TextDocumentIdentifier $textDocument, Position $position): Promise
     {
-        $this->server->doAnalysis();
-
         $file_path = LanguageServer::uriToPath($textDocument->uri);
         if (!$this->codebase->config->isInProjectDirs($file_path)) {
             return new Success([]);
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/LanguageServer/Server/TextDocument.php
```diff
@@ -92,8 +90,9 @@ public function didOpen(TextDocumentItem $textDocument): void
      * The document save notification is sent from the client to the server when the document was saved in the client
      *
      * @param TextDocumentItem $textDocument the document that was opened
+     * @param ?string $text the content when saved
      */
-    public function didSave(TextDocumentItem $textDocument): void
+    public function didSave(TextDocumentItem $textDocument, ?string $text): void
```
ging-dev, line 95: null won't happen in this case, because `includeText = true`
https://github.com/vimeo/psalm/blob/e91e33c39c2d15f326f91da7906744714b387a65/src/Psalm/Internal/LanguageServer/LanguageServer.php#L265
Should I keep it for consistency with the language server protocol specification?
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#didSaveTextDocumentParams

orklah:
@tm1000 Would you mind taking a look? I don't know LSP well enough to judge

src/Psalm/Internal/Cli/LanguageServer.php
```diff
@@ -276,14 +291,31 @@ function () use ($current_dir, $options, $vendor_dir): ?\Composer\Autoload\Class
...
-        $providers = new Providers(
-            new FileProvider,
-            new ParserCacheProvider($config),
-            new FileStorageCacheProvider($config),
-            new ClassLikeStorageCacheProvider($config),
-            new FileReferenceCacheProvider($config),
-            new ProjectCacheProvider(Composer::getLockFilePath($current_dir))
-        );
+        if (isset($options['no-cache']) || isset($options['i'])) {
+            $providers = new Providers(
```
tm1000, line 295: Unless something has changed. When I tried this last it totally breaks the LSP. Cache must remain on. A better way to do this is provide the ability to clear cache on boot.

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
tm1000: Yea. It should match the LSP spec. Just because include text is true does not mean a manage server will respect that

tm1000 (changes requested):
My biggest issue with this is I know from my own work on the LSP if you disable cache by just setting file provider that the LSP will eventually break while in use. This is because psalm relies on its own cache to reference other files. If there is no cache when working with a large code base you will start getting random issues or no issues at all. 

Instead a clear cache on boot option is better. Which essentially forces a re-indexing. 

The rest looks pretty good and matches similarly to the work I have been doing in my upcoming PR

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
ging-dev: This will help psalm to analyze the file on save, even if we use `disable-on-change`

src/Psalm/Internal/Cli/LanguageServer.php
ging-dev: I use LSP on some projects that are too small to detect problems with cache. I agree with using cache, but I really want `no-file-cache` as default (`ParserCacheProvider($config, false)`. Every time the didChange event fires, Psalm does a file analysis and caches the files in the project, which actually happens all too often. I hope you will have no problem turning this off.

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
```diff
@@ -121,23 +121,16 @@ public function didChange(VersionedTextDocumentIdentifier $textDocument, array $
             return;
         }
 
-        if ($this->project_analyzer->onchange_line_limit === 0) {
-            return;
-        }
```
ging-dev, line 127: didChange will never be called when `onchange_line_limit=0` because
https://github.com/vimeo/psalm/blob/e91e33c39c2d15f326f91da7906744714b387a65/src/Psalm/Internal/LanguageServer/LanguageServer.php#L269

src/Psalm/Internal/LanguageServer/Server/TextDocument.php
```diff
@@ -121,23 +121,16 @@ public function didChange(VersionedTextDocumentIdentifier $textDocument, array $
...
-
-        if (count($contentChanges) === 1 && $contentChanges[0]->range === null) {
-            $new_content = $contentChanges[0]->text;
-        } else {
-            throw new UnexpectedValueException('Not expecting partial diff');
+        foreach ($contentChanges as $change) {
+            $this->codebase->addTemporaryFileChanges($file_path, $change->text);
         }
```
ging-dev, line 126: Copy from Phan

tm1000:
The issue with removing the cache still exists in this PR and I believe overall that's a bad idea from my own testing. This week I'll run this pr through a few tests of my own to see how it performs.

tm1000:
For reference the work I've been doing is here: https://github.com/tm1000/psalm/tree/feature/upgrade-lsp

The only reason this is not a PR yet is I'm working on writing unit tests so that the LSP is my reliable through community contributions.

ging-dev:
> The issue with removing the cache still exists in this PR and I believe overall that's a bad idea from my own testing. This week I'll run this pr through a few tests of my own to see how it performs.

I will revert to disabling cache, but will keep the `--no-file-cache` option. I hope you won't have any problems with this setting:

```php
$providers = new Providers(
    new FileProvider,
    new ParserCacheProvider($config, false),
    new FileStorageCacheProvider($config),
    new ClassLikeStorageCacheProvider($config),
    new FileReferenceCacheProvider($config),
    new ProjectCacheProvider(Composer::getLockFilePath($current_dir))
);
```

`new ParserCacheProvider($config, false)` This prevents cache overwriting after each analysis.

image

Constantly writing data to the cache is generally not good for me as it harms the SSD. Also, clearing cache on boot is also not a good idea, we should just clear it in case it causes Psalm to behave incorrectly.

ging-dev:
@tm1000 you're right, I've noticed that Psalm actually works incorrectly when cache is missing. PR is ready!

tm1000 (approved):

orklah:
Thanks!

PR #7890, opened 2022-04-23
disussion#7862: adds first test case

BruceGitHub:
This PR poin to implements the idea from this 
https://github.com/vimeo/psalm/discussions/7862


tests/ReportOutputTest.php
```diff
@@ -1323,4 +1327,96 @@ private function toUnixLineEndings(string $output): string
...
+        }
+    }
+
+    /**
+     * @return Generator<string, array<string>>
+     */
+    public function outputProvider(): Generator
+    {
+        $file_contents = <<<'EOT'
+<?php
```
AndrolGenhald, line 1360: Since the master branch requires PHP 7.4, I'd prefer the indented heredoc syntax. Not sure if @orklah has any thoughts on that.

orklah: I never use heredoc so I don't really know, but yeah, if we can indent that, it's probably for the best

BruceGitHub: Ok!

BruceGitHub: this right ? 
https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc

AndrolGenhald: Yep, just indent everything including `EOT`.

BruceGitHub: I'm here!
I'm busy during this period, but I worked on it...

BruceGitHub: Ok, I elaborate first crappy crappy version!
The idea is:

1. Detect from the "issue->message" two arrays, requested and provided 
to perform this operation I create a class "PrettyDetectArray->detect"

This is a complex operation because the message is a "random" payload! 
The next idea might add some sort of meta-data, to the point where detecting the issue. 
In this case, the process is simple: format and compare... (next in work...) 

2. Format one and second array 
to perform this operation I create a class "PrettyFormat->format"

3.  Compare one and second array
to perform this operation I create a class "PrettyCompare->compare"

I think to improve the concrete code, to make it more clean and readable 

push coming

AndrolGenhald: > Detect from the "issue->message" two arrays, requested and provided to perform this operation I create a class "PrettyDetectArray->detect"

This seems unnecessarily complicated and will probably cause issues if I do something like `class Myarray` `Myarray{key: 'value'}`. I think it would probably be easier and better to do something like add a `$types` list to issues and have the message be a format string. You could even be more generic and support multiple rendering categories in the future:

```<?php
new SomeIssue('The type "{provided}" doesn't match the expected "{expected}"', ["provided" => $provided_type, "expected" => $expected_type], $suppressed_issues);
```

On the other hand having to track the type in the issue requires more refactoring, so idk. If we do decide to do a refactor requiring a change to issue messages, we might as well try to be consistent about the order of "Foo provided, expected Bar" vs "Expected Foo, found Bar" everywhere.

Really nice to get this improvement though!

AndrolGenhald: Thinking about it a bit more, I think if we went the route of having the issues track the actual `Union` types, we'd probably want to make the message generic like `'The provided type doesn't match the expected type'`, and then when it's rendered that can be converted to either `'The provided...type. Provided: array{...}; Expected: array{...}` or the new pretty format based on config and output type.

If you've already gotten the type detection mostly working though, that might be easier to go with for now until things get closer to finished.

BruceGitHub: Nice !
>Really nice to get this improvement though!
I'm ready to go in this direction for me is the best! 

I could create a intermediate task/issue to doing this work, then complete this ...

BruceGitHub: Can you explain the second sentence?
> Thinking about it a bit more, I think if we went the route of having the issues track the actual Union types, ...
like structure with the "tokens" ?

orklah:
Sorry for not triggering CI earlier, I have some backlog on replying to issues and PR. Don't hesitate to ping me if needed ;)

BruceGitHub:
> Sorry for not triggering CI earlier, I have some backlog on replying to issues and PR. Don't hesitate to ping me if needed ;)

I will try some spyke

orklah:
The error on windows looks very much like a line feed issue.

The approach seems fine to me :)

Can't wait to add this!

BruceGitHub:
I will try to simulate windows env woth linux ...

BruceGitHub:
> I did not understand if the last commit passed all tests or if some random failure occurs

small recap: 
- ci: I did not understand well the problem yet
- src: I found a defect in the algorithm and I working on it to resolve

orklah:
If you want, find a little something to PR so I can merge it. You won't be bothered by manual approval on our end anymore after that ;)

AndrolGenhald:
> If you want, find a little something to PR so I can merge it. You won't be bothered by manual approval on our end anymore after that ;)

☝️

There are plenty of issues marked as easy or good first issue, I generally don't bother with the easy ones that I don't personally care about so that there are good ways for new contributors to get started.

BruceGitHub:
> > If you want, find a little something to PR so I can merge it. You won't be bothered by manual approval on our end anymore after that ;)
> 
> point_up
> 
> There are plenty of issues marked as easy or good first issue, I generally don't bother with the easy ones that I don't personally care about so that there are good ways for new contributors to get started.

OK, so I start with an issue with this label? 
some suggestion or preference?

orklah:
https://github.com/vimeo/psalm/issues/8062 should be fairly easy for example

PR #7894, opened 2022-04-23
Fix return type of `ldap_search`

paulfedorow:
The return type of `ldap_search` since PHP 8.1 is `LDAP\Result|false` rather than `LDAP\Connection|false`: https://www.php.net/manual/en/function.ldap-search.php

This PR changes `CallMap.php` and `CallMap_81_delta.php` to reflect this.


orklah:
Thanks!

PR #7904, opened 2022-04-26
Alter order of baseline option loading to permit updating custom baseline

### LINKED ISSUE #7903: Can't update baseline with custom baseline
```
psalm --use-baseline="./psalm-baseline.xml" --update-baseline
```

The above command looks like it should utilise the named file as a baseline, and update it. Instead it outputs:

> Cannot update baseline, because no baseline file is configured

At the end of the Psalm run. For a start, this check could be done before the Psalm run (which may take some time), but more specifically this behaviour should work - if a custom baseline can be used, it should also be possible to be updated.

M1ke:
Resolves #7903

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -1029,10 +1029,6 @@ final class Psalm
             $issue_baseline = self::generateBaseline($options, $config, $current_dir, $path_to_config);
         }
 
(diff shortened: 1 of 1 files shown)


M1ke:
I'm confused by the Shepherd fail - the fail lines don't seem related to the change I've made

orklah:
Thanks! Sorry for CI failures :(

M1ke:
Thanks for the merge!

PR #7910, opened 2022-04-29
[draft] Allow templated first class callables

veewee:
WIP 

This PR aims to fix the first class callable issue in #7244.

The idea is to set a new 'forwarding_to' prop on a `TClosure`.
This prop will be used during analyzing to do a `VirtualFunctionCall` to the forwarded function, with the arguments provided to the first class callable function.

So in this case:

```php

/**
 * @template T
 * @param T $param
 * @return T
 */
function debug($param)
{
    return $param;
}

$firstClass = debug(...);
$actualResult = $firstClass('x');

```

Template on `$firstClass` will be inferred like this during a function call:

* firstClass = `Closure(T): T` with `forward_to: debug` and args `['x']`
* fetch result type from `VirtualFunctionCall` to `debug('x')`
* Make sure to Untemplate first class callable arguments afterwards (to avoid other type errors)
* Mark `'x'` as the result type of the function call `firstClass('x')`

Feel free to point me in the right direction, because I am just trying out some things :)

Todo:

- [ ] write + fix test cases
- [x] first class callable methods
- [x] first class callable static methods
- [ ] first class invokables
- [ ] first class callable arrays - (static) method calls
- [ ] first class callable strings
- [x] Closure::fromCallable
- [ ] cleanup
- [ ] Nested lookups?
- [ ] how does it behave inside `pipe` combinator

### DIFF
--- a/refs/firstclass.php
+++ b/refs/firstclass.php
@@ -1,0 +1,23 @@
+<?php declare(strict_types = 1);
+
+/**
+ * @template T
+ * @param T $param
+ * @return T
+ */
+function debug($param)
+{
+    return $param;
+}
+
+$firstClass = debug(...);
+$actualResult = $firstClass('x');
+
+/** @psalm-trace $firstClass, $actualResult */
+
+$boom = $actualResult/2;
+
+/** @psalm-suppress ForbiddenCode */
+var_dump($boom);
+
+
--- a/refs/firstclassmethod.php
+++ b/refs/firstclassmethod.php
@@ -1,0 +1,26 @@
+<?php declare(strict_types = 1);
+
+class X {
+    /**
+     * @template T
+     * @param T $param
+     * @return T
+     */
+    public function debug($param)
+    {
+        return $param;
+    }
+}
+
+
+$firstClass = (new X)->debug(...);
+$actualResult = $firstClass('x');
+
+/** @psalm-trace $firstClass, $actualResult */
+
+$boom = $actualResult/2;
+
+/** @psalm-suppress ForbiddenCode */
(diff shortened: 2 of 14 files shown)


klimick:
Looks like a working solution. Unfortunately for me, hard to say what's drawbacks are followed with it.

klimick:
Somewhere @weirdan mentioned that something like this requires the new type `Closure<T>(T): T` (or more params: `Closure<T, B of int>(T, B, string): list<T|B|string>`)

```php
/**
 * @return Closure<T>(T): T
 */
function getId(): Closure
{
    return fn($v) => $v;
}

getId()('foo'); // type is 'foo'


/**
 * @template T
 * @param T $v
 * @return T
 */
function id(mixed $v): mixed
{
    return $v;
}

$fn = id(...); // type is Closure<T>(T): T
$fn('foo'); // type is 'foo'
```

But again. I don't know what the difference between these two solutions.

veewee:
That is what I was trying to do first, but the current analyzers are not ready for that: the closure signature is loaded from memory and templates don't get inferred. Closires do accept template types (internally), but nothing happens with them.

According to the code I've seen and the issues in the issue tracker, it looks like it will be very hard to get that working. Frankly a bit above my competence at this moment.. It would be the best solution though 🙂

thomasvargiu:
@veewee @klimick Any news on this? It would be awesome

veewee:
I haven't been able to find a good solution for this problem so far... The changes were from 9 months ago and a lot has changed since then. Hopefully anyone else has more luck implementing this feature!

Issue #7927, opened 2022-05-05
False positive DocblockTypeContradiction on a key saying that the key is always non-empty


white43:
I've found several cases of the same false positive error. Pay attention to the last part with `isset` instead of `empty`, since somehow it works asexpected.

https://psalm.dev/r/24b8f380fb

white43:
`git bisect` point to this commit 5f17ea9b63785d12af05363f3148ec4e8e660aa2

white43:
I'm not sure if it's a good solution, but setting `possibly_undefined` to `true` for non-existent array properties in `getValueForKey` does help; all tests are still green. If that's ok, I can open a PR. :smile: 

https://github.com/vimeo/psalm/blob/4.x/src/Psalm/Type/Reconciler.php#L685-L688

orklah:
Please do :)

PR #7933, opened 2022-05-08
Setting possibly_undefined to true for non-existent array properties

white43:
See https://github.com/vimeo/psalm/issues/7927

Would you like me to add some tests for this PR?

### DIFF
--- a/src/Psalm/Type/Reconciler.php
+++ b/src/Psalm/Type/Reconciler.php
@@ -741,6 +741,7 @@ class Reconciler
                                 if ($existing_key_type_part->previous_value_type) {
                                     $new_base_type_candidate = clone $existing_key_type_part->previous_value_type;
                                     $new_base_type_candidate->different = true;
+                                    $new_base_type_candidate->possibly_undefined = true;
                                 } else {
                                     return null;
                                 }
--- a/tests/TypeReconciliation/RedundantConditionTest.php
+++ b/tests/TypeReconciliation/RedundantConditionTest.php
@@ -906,6 +906,14 @@ class RedundantConditionTest extends TestCase
                         assert(!!$p);
                     }',
             ],
+            'noDocblockTypeContradictionOnTheSecondEmpty' => [
+                'code' => '<?php
+                    /** @var stdClass[] $a */
+                    $a = [];
+
+                    if (empty($a["a"]) || empty($a["b"])) {
+                    }',
+            ],
         ];
     }
 


orklah:
Could you target master instead please? It seems to work well with the build but I'm not sure if it could have side effects.

And yes, please add a test to show that you can use empty on a unknown key without errors

orklah:
The PossiblyInvalidArgument  error is pretty weird. `$options['tcp']` was already checked on line 247 and is supposed to be a string. Not sure why the error pops up now...

white43:
Yeah, that's really odd. After switching to master I can't run static analysis on my laptop. I'll need to dig into it, but it's highly likely caused by my changes.

white43:
It took a while to narrow down it to a snippet for further debugging. I'm not familiar much with Psalm innards, so next two snippets seem weird to me. Somehow in the second case Psalm doesn't see `tcp` property at the line with `@psalm-trace`. It has nothing in common with the issue; it's just yet another observation.

https://psalm.dev/r/fb786bf3e0
https://psalm.dev/r/3e37ee3f51

orklah:
Oh, I think I get it. The negated array_key_exists transforms the array into an array shape. Then, when the !string only helps when we have an array shape (it's ignored otherwise). That's a nice bug, but I don't get why it shows up with your change yet...

white43:
I haven't finished debugging yet. Psalm is a complicated thing. :smile: 
At this moment I managed to narrow down the problem to this block:

https://github.com/vimeo/psalm/blob/f960d71b7e8dccb0815bce28a22a33ce5410e361/src/Psalm/Type/Reconciler.php#L294-L296

In my branch `$result_type->isNever()` is `true` while in master branch `$result_type` contains `TNull`. So in my case `$options['tcp']` won't be added to `$changed_var_ids` here

https://github.com/vimeo/psalm/blob/f960d71b7e8dccb0815bce28a22a33ce5410e361/src/Psalm/Type/Reconciler.php#L320

... and won't be merged with `$outer_context` later.

white43:
Ok, I got it. The reason why I see `never` is here:
https://github.com/vimeo/psalm/blob/f960d71b7e8dccb0815bce28a22a33ce5410e361/src/Psalm/Internal/Type/SimpleNegatedAssertionReconciler.php#L93-L95
As far as I understand, it covers cases like `!isset($something_possibly_undefined)`, but I don't understand why on earth we return `never`.

orklah:
It kinda makes sense to me. If you have this code
```
if(!isset($var)){
    //the type of $var is *NEVER* here
}
```

PR #7935, opened 2022-05-09
Infer object shape when array or scalar is cast to object

### LINKED ISSUE #7916: Infer object shape when array is cast to object
When an array is cast to an object, Psalm only sees the type as `stdClass` instead of `object` with the same shape as the array. This prevents detecting if the object shape is invalid:

https://psalm.dev/r/5584585af9

## Expected output

```
ERROR: InvalidReturnStatement - The inferred type 'object{notstatus: 'failed'}' does not match the declared return type 'object{status: string}' for foo
```

## Actual output

No issues!

### LINKED ISSUE #7934: Detect reduntant object casts
When an array is cast to an array, Psalm detects and reports the redundant cast. However, this doesn't appear to work for `(object)` casts.

https://psalm.dev/r/39198b19f3

## Expected output
RedundantCast - 4:12 - Redundant cast to object

## Actual output
No issues!

theodorejb:
Also detect redundant object casts.

Fixes #7916, fixes #7934

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
@@ -183,22 +183,42 @@ class CastAnalyzer
         }
 
         if ($stmt instanceof PhpParser\Node\Expr\Cast\Object_) {
-            $was_inside_general_use = $context->inside_general_use;
-            $context->inside_general_use = true;
-            if (ExpressionAnalyzer::analyze($statements_analyzer, $stmt->expr, $context) === false) {
-                $context->inside_general_use = $was_inside_general_use;
-
+            if (!self::checkExprGeneralUse($statements_analyzer, $stmt, $context)) {
                 return false;
             }
-            $context->inside_general_use = $was_inside_general_use;
 
-            $type = new Union([new TNamedObject('stdClass')]);
+            $permissible_atomic_types = [];
+            $all_permissible = false;
 
-            $maybe_type = $statements_analyzer->node_data->getType($stmt->expr);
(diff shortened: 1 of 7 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -183,22 +183,44 @@ public static function analyze(
...
+            if ($stmt_expr_type = $statements_analyzer->node_data->getType($stmt->expr)) {
+                if ($stmt_expr_type->isObjectType()) {
+                    self::handleRedundantCast($stmt_expr_type, $statements_analyzer, $stmt);
+                }
+
+                $all_permissible = true;
+
+                foreach ($stmt_expr_type->getAtomicTypes() as $type) {
+                    if ($type instanceof Scalar) {
+                        $objWithProps = new TObjectWithProperties(['scalar' => new Union([$type])]);
```
orklah, line 204: It's a nice touch to handle this, but do we really want to?

This is a very weird behaviour by PHP and, IMHO, rarely intended by any developper.

We could throw a InvalidCast or a PossiblyInvalidCast when encountering a scalar in an object cast, what do you think?

theodorejb: It's definitely weird behavior by PHP, and probably rarely intended by devs. But I still feel like Psalm should infer the correct type when possible rather than throwing an error, especially when it's easy to do so. Scalars cast to an array are also quite uncommon, but Psalm does correctly infer the type for these.

orklah:
Thanks!

Issue #7607, opened 2022-02-08
`implode` of non-empty-list should always inferred as `non-empty-string`


boesing:
Hey guys,

as of now, `join`/`implode` usage on `non-empty-list`/`non-empty-array` of strings does not lead to a `non-empty-string`.
What would be the best way of implementing this or do you think that this should actually not be done by this project?

Thanks!

https://psalm.dev/r/86de0748e7

orklah:
Slight edge case: https://3v4l.org/CdIJ8

Which means the correct reproducer should be: https://psalm.dev/r/21307e4197

Then, what need to be changed is this:
https://github.com/vimeo/psalm/blob/3cce69189f748235a613c63d3d2609fe76f079c5/stubs/CoreGenericFunctions.phpstub#L571

Currently, there is no possibility to return a non-empty-string when the separator is an empty string. You could improve the conditional by adding a nested conditional in order to check that the array is a non-empty-array of non-empty-strings. In this case only you can tell the implode will return a non-empty-string

hirokinoue:
Hi, I'll try this issue if I can. I think the code might meet the expectation above. 

```
 * @return (
 *     $separator is non-empty-string
 *     ? ($array is non-empty-array
 *         ? ($array is array<literal-string|literal-int>
 *             ? ($separator is literal-string ? non-empty-literal-string : non-empty-string)
 *             : non-empty-string
 *         )
 *         : string)
 *     : ($array is non-empty-array
 *         ? ($array is array<non-empty-literal-string|non-empty-string>
 *             ? ($array is array<non-empty-literal-string> ? non-empty-literal-string : non-empty-string)
 *             : string
 *         )
 *         : string)
 * )
```

But one remaining problem is the return type of `implode('', [1])` is string instead of non-empty-literal-string maybe because of somewhere around ScalarTypeComparator. I guess it is not easy (at least for me currently) and could be sperate issue from this issue.

Even so, the code above still improve the inference of implode(). Is it alright to create a PR?
Or I may have overlooked something. Any advice is welcome. Thanks in advance.

orklah:
Please do :) 

If we can't seem to meet the goal with a conditional return, we can always add a ReturnTypeProvider instead, it doesn't take a lot of Psalm knowledge but it's more powerful than the conditional return syntax

hirokinoue:
> Please do :)

Thanks! I created a PR.

> we can always add a ReturnTypeProvider

Thank you for the advice, it'll take some time but I'll check it.

PR #7967, opened 2022-05-14
Improve @return annotation for implode() so that it can handle non-empty-array of non-empty-strings case

hirokinoue:
Fixes: #7607


AndrolGenhald:
Could you mention that this is for `implode` in the title? PR titles go in the changelog and the current title is a bit unclear.

hirokinoue:
Thanks! I fixed the title.

orklah:
Thanks!

Issue #7956, opened 2022-05-13
array_unique 2nd parameter may restrict the array items type


mathroc:
Hi,

the [array_unique() function](https://www.php.net/array_unique) has an optional 2nd parameter flag, which is one of the following value:

* `SORT_REGULAR`
* `SORT_NUMERIC`
* `SORT_STRING` (default value)
* `SORT_LOCALE_STRING`

if the flag is one of `SORT_LOCALE` or `SORT_LOCALE_STRING`, the array should only contain strings or values that can be casted to string. I'm not sure exactly what the definition should be, maybe: `int|float|string|bool|\Stringable|null`

Here we can see it fail : https://3v4l.org/lJsCp

and here we can see that Psalm does not find any error: https://psalm.dev/r/2135f5de25

---

edit: I realized that the `SORT_NUMERIC` also restrict the item type, probably something like `int|float|string|null`

staabm:
I am wondering whether the assumptions holds, as passing a array with a single `stdClass` objects seems to work just fine https://3v4l.org/bYH89

mathroc:
It's true that if there is only one element it's fine (because it's not compared to anything, it can only be unique already)

but put 2 elements and this fail. I would argue that calling `array_unique` with an array knowing it has only 1 element is useless, so even though Psalm will raise an issue in this case I don't think it's worth it trying to avoid it

weirdan:
The linked pull request doesn't seem to have fixed the original snippet: https://psalm.dev/r/2135f5de25 (still shows no errors)

othercorey:
@weirdan Is there a way to annotate an array with just 1 element?

weirdan:
Not if you don't know the keys.

weirdan:
@othercorey but we don't have to be that specific. Flagging all `array<array-key,object>` going into `array_unique(..., SORT_REGULAR)` is fine.

othercorey:
It's simpler than that. All elements must be stringable if there is more than 1. I'm just adjusting the callmap and adding tests.

PR #7981, opened 2022-05-17
Improve array_unique callmap stubs

mathroc:
see #7956


mathroc:
thank you, I finally got some time to fix it, looks like it needs a label to be fully green but I can't do that

orklah:
Could you drop `associative-array` and just keep `array`?

It seems to be a weird old type that is only in callmap and it doesn't seem to expect `<>` after it so it may not work as expected

mathroc:
> Could you drop `associative-array` and just keep `array`?
> 
> It seems to be a weird old type that is only in callmap and it doesn't seem to expect `<>` after it so it may not work as expected

done, should I have keep it for the `0` flag? (as it doesn't have type parameters)

orklah:
It seem to be exactly handled the same as `array`, except for special handling in `array` so you can keep it like that. I think it's a type that existed at one point in Phan to denote `array<string, ...>` but it doesn't even have this meaning in Psalm

orklah:
Thanks!

PR #7982, opened 2022-05-17
Bugfix/#7912 class APCuIterator does not exist

### LINKED ISSUE #7912: Class, interface or enum named APCuIterator does not exist
See: https://psalm.dev/r/8daba04c8c

Ref: https://www.php.net/manual/en/apcuiterator.construct.php

KevinVanSonsbeek:
Fixes #7912

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -2051,6 +2051,11 @@ class Config
             $this->internal_stubs[] = $ext_phpredis_path;
         }
 
+        if (extension_loaded('apcu')) {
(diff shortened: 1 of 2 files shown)


ricardoboss (approved):

orklah:
Looks great. @AndrolGenhald can you check if we missed something? I have a vague memory that we have a list of supported extensions somewhere?

AndrolGenhald:
I think this is all that's needed for Psalm 4, but it'll probably have issues when merging to master. I might try to open a PR to merge 4.x to master later while undoing part of this and including the new extension stuff.

It's still unclear to me why the CallMap doesn't work in this case, but my personal goal is to remove extensions from the CallMap anyway 🤷

orklah:
Thanks!

PR #7988, opened 2022-05-19
Fix conflict of option -c with shell completion

GromNaN:
Symfony Console 5.4+ provided shell completion, with 2 new default commands. `completion` to dump the script and `_complete` to return the values. The 2nd command defines an option with shortcut '-c' that conflict with the default option in the `psalm-plugin` manager. By moving this option back to the 3 commands that use it, the completion can work.

Only command and option names can be completed for now. It would be useful if the `pluginName` could be autocompleted (using interactive questions or shell completion).


orklah:
We require `"symfony/console": "^3.4.17 || ^4.1.6 || ^5.0 || ^6.0",`

and you said completion changed in 5.4. Is this change backward compatible?

GromNaN:
Yes, that's 2 new commands that are registered by default from 5.4. Here, the application needs the changes I proposed to support this new feature, but nothing is broken by the update.

AndrolGenhald:
I think @orklah was asking if your PR is backward compatible, especially since we allow `symfony/console` versions older than 5.4. It looks to me like it probably shouldn't be an issue though, although I haven't actually tested it. I assume you've actually run each of the commands with the `-c` flag before and after your change?

GromNaN:
So not doubt, `Command::addOption` exists since [Symfony 2]( https://github.com/symfony/symfony/blob/v2.0.0/src/Symfony/Component/Console/Command/Command.php#L323).

I tested each command.

orklah:
Thanks!

PR #7992, opened 2022-05-20
Fix class const issue when using floats declared in future consts (fixes #7973).

### LINKED ISSUE #7973: Invalid suggestion for `psalm --alter` and strange const error
image

There's a lot to unpack here... Any idea why any of this is happening? I don't understand the error message (I have not declared any type for the const, it should infer the type from the value) and the `--alter` suggestion doesn't even work.

AndrolGenhald:
- Calculate literal type for float arithmetic instead of only for int arithmetic
 - Fix copy/paste fail causing InvalidConstantAssignmentValue to be marked as fixable


AndrolGenhald:
I could have sworn there was a PHP warning for doing [bitwise operations on floats](https://3v4l.org/fBIgZ), but apparently not... Do we want to add an issue for that to Psalm?

orklah:
Thanks!

PR #7994, opened 2022-05-21
Adds support for fixing missing throws doc block

### LINKED ISSUE #7918: Add support for auto fixing missing throws doc block errors
Related to 
`checkForThrowsDocblock` option, it would be nice to auto add throws docblock to functions. 
Use case: 
Better understanding of how the function can fail for the caller, often times it is useful then to refactor the code to not catch the exception at the higher level and re throw it to provide more context

aszenz:
Fixes #7918

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
@@ -65,6 +65,8 @@ use function array_keys;
 use function array_merge;
 use function array_search;
 use function array_values;
+use function assert;
+use function class_exists;
 use function count;
 use function end;
 use function in_array;
@@ -723,6 +725,7 @@ abstract class FunctionLikeAnalyzer extends SourceAnalyzer
             }
 
             if (!$is_expected) {
+                assert(class_exists($possibly_thrown_exception));
                 $missingThrowsDocblockErrors[] = $possibly_thrown_exception;
                 foreach ($codelocations as $codelocation) {
                     // issues are suppressed in ThrowAnalyzer, CallAnalyzer, etc.
@@ -737,8 +740,7 @@ abstract class FunctionLikeAnalyzer extends SourceAnalyzer
             }
         }
 
-        if (
-            $codebase->alter_code
+        if ($codebase->alter_code
             && isset($project_analyzer->getIssuesToFix()['MissingThrowsDocblock'])
         ) {
             $manipulator = FunctionDocblockManipulator::getForFunction(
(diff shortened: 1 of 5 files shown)


orklah:
Cool! That looks nice!

If you target 4.x, you must keep compatibility with PHP 7.1 though. Can you check this and the other CI failure?

Could you also add a test case with two different exceptions, to check the docblock rendering?

tests/FileManipulation/ThrowsBlockAdditionTest.php
```diff
@@ -0,0 +1,91 @@
...
+                    function foo(string $s): string {
+                        if("" === $s) {
+                            throw new \Exception();
+                        }
+                        return $s;
+                    }',
+                '<?php
+                    /**
+                     * @throws InvalidArgumentException|DomainException
+                     * @throws Exception
```
aszenz, line 77: i preserve the existing throws annotation and add a new one based on inference

src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
```diff
@@ -719,6 +725,8 @@ public function analyze(
             }
 
             if (!$is_expected) {
+                assert(class_exists($possibly_thrown_exception));
```
aszenz, line 728: should i keep this assert or remove it, it makes sense for throws keyword to only contain classes which exist

tests/FileManipulation/ThrowsBlockAdditionTest.php
```diff
@@ -0,0 +1,123 @@
...
+                        if("" === $s) {
+                            throw new \Exception();
+                        }
+                        return $s;
+                    }',
+                '7.4',
+                ['MissingThrowsDocblock'],
+                true,
+            ],
+            'doesNotAddDuplicateThrows' => [
```
aszenz, line 89: This validates that the fix only adds exceptions which are not listed in the current throws doc block

tests/FileManipulation/ThrowsBlockAdditionTest.php
aszenz: So the fix will always add a new throws annotation which can have `Exception1|Exception2` syntax, but not modify existing one

src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
aszenz: ideally this check would not be needed, but that means modifying the return type for the function `getUncaughtThrows` which would be a bigger change

orklah: Please drop it. It's risky because classes could be stubbed or inserted through a Plugin so you can't guarantee they'll exists for Psalm to autoload

orklah:
Seems fine to me, if you mark this as ready, I'm up to merge it :)

aszenz:
@orklah Can this be merged or is it something to do still

orklah:
Sorry :( I didn't get a notification on the status change! Thanks!

Issue #7996, opened 2022-05-22
Use faster hashing algorithm for cache


kkmuffme:
Currently all places uses sha1 algorithm

When tracing psalm, I can see that ClassLikeStorageCacheProvider.php `private function getCacheHash` is extremely slow and makes up 20-25% of the whole psalm run time just with sha1 calls.

To avoid hash conflicts, we'd need at least a 64bit, better a 128bit hash (sha1 is 160bit, but that's overkill) https://stackoverflow.com/a/14210379

Checking https://php.watch/articles/php-hash-benchmark we can see that xxh128 is about 20 times faster than sha1, and md4 is ~2 as fast as sha1.

We could change all `sha1(` calls to something like:
`version_compare( phpversion(), '8.1', '>=' ) ? hash('xxh128', $data ) : hash( 'md4', $data )`

This improves psalm run time by about 20% in PHP8.1 case and about 10% for older PHP versions

I'm happy to provide a PR, just want to discuss it beforehand.

kkmuffme:
No, as this is an internal issue

orklah:
We have a few md5 as well.

From the profiling I made a while ago, I didn't see a big hash issue, but even then, if we can improve that, I'm all in!

PR #7997, opened 2022-05-22
change cache hash type for better performance

kkmuffme:
Fixes https://github.com/vimeo/psalm/issues/7996

Only changed the cache hash, as that is where profiling showed slowness.
Other hashes not changed, to avoid breaking anything.

Question: this is a breaking change for the cache directory (= any existing caches do not work anymore after this change is implemented). This needs to be documented in the release notes I guess? (I'd need help for that)


kkmuffme:
@orklah tests pass now. There seems to be a bug in Windows PHP, as just putting the condition separately makes the test pass - see my commit "debug CI" which fixes this.

Please add release:internal label and merge.

AndrolGenhald:
Does Psalm currently have a process to automatically clean up old cache files? Afaict the only risk should be that we waste a bunch of space on old cached stuff that can never be used.

src/Psalm/Internal/Provider/ClassLikeStorageCacheProvider.php
```diff
@@ -111,7 +113,8 @@ public function getLatestFromCache(
 
     private function getCacheHash(?string $file_path, ?string $file_contents): string
     {
-        return sha1(($file_path ? $file_contents : '') . $this->modified_timestamps);
+        $data = ($file_path ? $file_contents : '') . $this->modified_timestamps;
+        return version_compare(phpversion(), '8.1', '>=') ? hash('xxh128', $data) : hash('md4', $data);
```
weirdan, line 117: If that's a hot path, a check like `PHP_VERSION_ID > 80100` would make more sense.

kkmuffme: Thanks for the feedback.
Given the current codebase of psalm, this is quite a microptimization (this change makes the psalm analysis 0.03% faster than version_compare as per profiler).

I'm happy to adjust this though. Can I just add it to the existing PR? or do I need to create a new one?

orklah: I don't think you can add on this one but maybe add this to the new one you created fo igbinary?

Issue #8008, opened 2022-05-25
Performance: only load files once


kkmuffme:
FileProvider.php function getContents

When logging the `$file_path` directly before:

```
return (string)file_get_contents($file_path);
```

I can see the same files getting opened multiple times between 2-100s of times. 90% of file_get_contents calls here are duplicates, which means psalm could be 4% faster if that didn't happen.

We check if the file was opened before, but this is useless, since it's always empty:

```
if (isset($this->open_files[$file_path_lc])) {
            return $this->open_files[$file_path_lc];
}
```

Changing `$open_files` to a static and using `self::$open_files` should fix this.
I tried this - however weirdly enough, $open_files still is always empty.

Any idea why this is happening though? (I can provide a PR with the static, however it has the same issue)

kkmuffme:
No, this is an psalm internal issue.

PR #8011, opened 2022-05-25
Performance only load files once

kkmuffme:
Fixes https://github.com/vimeo/psalm/issues/8008

Speeds up psalm between 5-15%.

Also fixes an issue on Unix systems where A.php and a.php were treated equal, even though they are different files, since unix file names are case sensitive.

### DIFF
--- a/src/Psalm/Internal/Provider/FakeFileProvider.php
+++ b/src/Psalm/Internal/Provider/FakeFileProvider.php
@@ -4,7 +4,6 @@ namespace Psalm\Internal\Provider;
 
 use function microtime;
 use function strpos;
-use function strtolower;
 
 class FakeFileProvider extends FileProvider
 {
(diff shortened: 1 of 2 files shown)


AndrolGenhald:
Can you double check memory usage? I'd expect the parsed structures to be much bigger than file contents, but this will definitely increase memory usage and it wouldn't be bad to double check. I assume this code is bypassed for the language server, but that would be good to verify as well.

kkmuffme:
The memory usage depends on what you are checking, as the memory usage is directly proportional to the number and size of files analyzed.

If real world tests show that memory consumption is too high, we could offer an option to 
a) limit the number of files that stay "cached" or 
b) only "cache" specific types (e.g. stubs)
c) exclude certain files from "cache"

I tested it on the WP core repo and the memory usage was only slightly higher than before.

AndrolGenhald:
Yeah, that's what I expected, just wanted to confirm with an actual test.

I do want to double check how the language server works though. If the language server relies on this class then the caching could completely break it, unless it's only used to get initial file state in which case the cache is just wasting memory.

kkmuffme:
Could someone merge this?

orklah:
Thanks!

PR #8022, opened 2022-05-26
Improve handling of unsupported references (fixes #8018).

### LINKED ISSUE #8018: Something weird.
Please rename this issue to something more appropriate, once you figure out what it is. I didn't know what to name it.

```php
/**
* @psalm-type Test = string
*/
class Foo
{
    /** @var array<string, Test> */
	private array $stuff = [];
    
    /**
    * @param non-empty-list<string> $param
    * @return Test
    */
    public function foo(array $param)
    {
        $arr = &$this->stuff[$param[0]];	// This seems to cause Psalm to blackout.
        
        /** @var Test */
        $x = $arr;
        
        // This trace doesn't work, and $x's type appears to be mixed at this point.
        /** @psalm-trace $x */
        return $x;
    }
}
```
https://psalm.dev/r/f81173405a

Converting `$param` to string, works: https://psalm.dev/r/718c05c153
Same if we remove the reference on the offending line: https://psalm.dev/r/6ebda981ae

AndrolGenhald:
Fixes the reference analysis to report references it's unable to analyze and continue as if they weren't references, rather than aborting analysis altogether.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php
@@ -5,7 +5,6 @@ namespace Psalm\Internal\Analyzer\Statements\Expression;
 use PhpParser;
 use PhpParser\Node\Expr\ArrayDimFetch;
 use PhpParser\Node\Expr\PropertyFetch;
-use PhpParser\Node\Expr\Variable;
 use Psalm\CodeLocation;
 use Psalm\CodeLocation\DocblockTypeLocation;
 use Psalm\Codebase;
@@ -82,7 +81,6 @@ use Psalm\Type\Atomic\TNonEmptyArray;
 use Psalm\Type\Atomic\TNonEmptyList;
 use Psalm\Type\Atomic\TNull;
 use Psalm\Type\Union;
-use RuntimeException;
 use UnexpectedValueException;
 
 use function array_filter;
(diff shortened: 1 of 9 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php
```diff
@@ -981,44 +975,20 @@ public static function analyzeAssignmentRef(
...
-            $lhs_location
-        );
+        if (!$stmt->var instanceof ArrayDimFetch && !$stmt->var instanceof PropertyFetch) {
+            // If left-hand-side is an array offset or object property, usage is too difficult to track,
+            // so it's not registered as an unused variable (this mirrors behavior for non-references).
+            $statements_analyzer->registerVariableAssignment(
+                $lhs_var_id,
+                $lhs_location
+            );
+        }
```
AndrolGenhald, line 985: It turns out the bug this pull request fixes actually affected this method, causing everything after `$context->vars_in_scope[$lhs_var_id] = &$context->vars_in_scope[$rhs_var_id];` to be ignored.

After it was fixed, that same line was reported as an unused variable, so I finally gave up and just decided that putting a reference in an object or an array is a bad idea anyway, and it's not worth it to track usage in that case (this is basically increasing the scope of the change in #7684 from only ignoring unused references on `$this` to ignoring unused references on any object or array).

orklah:
Thanks!

PR #8028, opened 2022-05-29
Fix syntax check failures

weirdan:
They were caused by installing packages with `--ignore-platform-reqs`
which brought PHP 8.1 autoloaded packages and caused fatal errors when
running php-parallel-lint. Instead of ignoring platform requirements we
now remove packages that are incompatible with PHP 7.1 (phpunit and its
dependents).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-composer-
 
+      - name: Drop incompatible packages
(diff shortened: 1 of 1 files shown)


weirdan:
Shepherd issue is fixed in #8029

orklah (approved):

orklah:
There's a build failure but I can't pinpoint what went wrong

weirdan:
> There's a build failure but I can't pinpoint what went wrong

I think that's the one that's fixed in #8029

PR #8029, opened 2022-05-29
Run Shepherd with 8.0

weirdan:
Because installing packages with 8.1 and checking them as if we're
running 7.1 just doesn't work.

### DIFF
--- a/.github/workflows/shepherd.yml
+++ b/.github/workflows/shepherd.yml
@@ -8,6 +8,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: shivammathur/setup-php@v2
+      with:
+        php-version: '8.0'
 
     - name: Install dependencies
       run: composer install --prefer-dist --no-progress --no-suggest


weirdan:
PHP syntax is fixed in #8028

orklah:
Do we still have a CI running on 7.1 somewhere after that? I think I remember seeing a few failures due to using incompatible syntax on 4.x

weirdan:
Psalm detects the minimum PHP version supported by reading `composer.json` and that's 7.1 for 4.x, regardless of the actual runtime version that Composer uses during dependency resolution.

> Do we still have a CI running on 7.1 somewhere after that? 

I'm not sure if I understand the question. There is a job running with 7.1, the syntax check one.

Funny thing though, the latest Psalm 4.x is not installable on 7.1 because Composer requires dev dependencies to be compatible, even if they are not installed. And PHPUnit 9.0 is not compatible with 7.1

weirdan:
>  Psalm 4.x is not installable on 7.1

Not installable with composer, that is, and it seems it never was. Phar should still run ok.

weirdan:
All green now, and should make the 4.x build green when merged. @orklah please merge once approved.

PR #8032, opened 2022-05-29
Precise stat return type

VincentLanglet:
From https://www.php.net/manual/fr/function.stat.php

See also https://github.com/phpstan/phpstan-src/pull/996

I basically copied the types from phpstan.


VincentLanglet:
Hi @orklah, i'm not against some help because I don't understand the failures :s

orklah:
You can ignore parse errors, seems like it's on our side. However, there are failing tests because you only updated Callmap.php. This file contains the most up-to-date information for the current tracked version(8.1), but it should always match with the file that introduced that definition. (For example, if a function was created in 7.4, you should add it in Callmap.php and Callmap74.php).

In your case, I think those function always existed and returned about the same thing, so you need to update Callmap_historical too

VincentLanglet:
> In your case, I think those function always existed and returned about the same thing, so you need to update Callmap_historical too

Done :)

orklah:
Thanks!

PR #7319, opened 2022-01-06
Applied ArraySpreadInsteadOfArrayMergeRector

weirdan:
Blocked by #7296

PR #8044, opened 2022-06-02
Fix various array spread issues.

### LINKED ISSUE #7296: Spread in array incorrectly produces non-empty list
https://psalm.dev/r/d8a8d758f6?php=8.1
The resulting array can still be empty: https://3v4l.org/1LZ7Y

### LINKED ISSUE #7297: Spreading an array should produce array in PHP 8.1 rather than list
https://psalm.dev/r/d8a8d758f6?php=8.1

AndrolGenhald:
- Correctly infer `array` and `list` instead of `non-empty-array` and `non-empty-list` (fixes #7296)
 - Add support for spreading string keys (fixes #7297).
 - Show issue when trying to unpack non-iterable
 - Show issue when trying to unpack iterable with non-array-key key
 - Re-added invalid PHP 8.0 tests removed in #6613
 - Unpacked lists with known keys will be inferred as eg `array{0: int, 1: int}<int<0, max>, int>` now but will still be treated as lists
 - Unpacked arrays with known keys will now be inferred as eg `array{a: string, b: string}<int, int>` instead of `array<int|string, int|string>`

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -588,7 +588,7 @@ class ArrayAnalyzer
 
                 $array_creation_info->can_create_objectlike = false;
 
-                if (!UnionTypeComparator::canBeContainedBy(
+                if (!UnionTypeComparator::isContainedBy(
                     $codebase,
                     $iterable_type->type_params[0],
                     Type::getArrayKey(),
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -8,12 +8,15 @@ use Psalm\Codebase;
 use Psalm\Context;
 use Psalm\Internal\Analyzer\Statements\ExpressionAnalyzer;
 use Psalm\Internal\Analyzer\StatementsAnalyzer;
+use Psalm\Internal\Codebase\ConstantTypeResolver;
 use Psalm\Internal\Codebase\TaintFlowGraph;
 use Psalm\Internal\Codebase\VariableUseGraph;
 use Psalm\Internal\DataFlow\DataFlowNode;
+use Psalm\Internal\Type\Comparator\UnionTypeComparator;
 use Psalm\Internal\Type\TypeCombiner;
 use Psalm\Issue\DuplicateArrayKey;
 use Psalm\Issue\InvalidArrayOffset;
+use Psalm\Issue\InvalidOperand;
 use Psalm\Issue\MixedArrayOffset;
 use Psalm\Issue\ParseError;
 use Psalm\IssueBuffer;
@@ -24,9 +27,7 @@ use Psalm\Type\Atomic\TArrayKey;
 use Psalm\Type\Atomic\TBool;
 use Psalm\Type\Atomic\TFalse;
 use Psalm\Type\Atomic\TFloat;
-use Psalm\Type\Atomic\TGenericObject;
 use Psalm\Type\Atomic\TInt;
-use Psalm\Type\Atomic\TIterable;
 use Psalm\Type\Atomic\TKeyedArray;
 use Psalm\Type\Atomic\TList;
 use Psalm\Type\Atomic\TLiteralClassString;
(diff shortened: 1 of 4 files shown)


AndrolGenhald:
Also unblocks #7319.

src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
```diff
@@ -127,7 +106,7 @@ public static function analyze(
             $item_key_type = null;
         }
 
-        if ($array_creation_info->item_value_atomic_types) {
+        if (!empty($array_creation_info->item_value_atomic_types)) {
```
orklah, line 109: Could we use anything but empty? I very much dislike this function (it make sense sometimes, but rarely)

src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
```diff
@@ -139,19 +118,39 @@ public static function analyze(
...
         }
 
         if ($array_creation_info->all_list) {
-            if (empty($array_creation_info->item_key_atomic_types)) {
+            if ($array_creation_info->can_be_empty) {
                 $array_type = new TList($item_value_type ?? Type::getMixed());
             } else {
                 $array_type = new TNonEmptyList($item_value_type ?? Type::getMixed());
-                /** @psalm-suppress InvalidPropertyAssignmentValue */
-                $array_type->count = count($array_creation_info->property_types);
```
orklah, line 154: This won't cause issues?

AndrolGenhald: It actually causes issues if it's there because `property_types` is guaranteed to be empty at that point since any array with `property_types` set will be turned into an array shape (I only noticed because Psalm warned me...). I'm also not convinced it wasn't buggy before, it seems like it could probably have been setting `count` even when it wasn't exact.

src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
AndrolGenhald: Oh, yeah, bad habit because I'm used to similar functions specifically for arrays/vecs/lists in other languages, but I'm also not a fan of implicit conversion of array to bool either, I'll try to remember to do `count(...) === 0` in the future.

AndrolGenhald: Looks like there are a lot of `empty` usages around that could probably be changed, would it be worthwhile to add some sort of `bannedFunctions` Psalm config, or do other tools handle that well enough that we don't want to? For stuff like this you don't want to ban it entirely, you just want a warning that you can suppress, so disabling the function in PHP isn't possible.

orklah: I removed some a while ago when it seemed safe: https://github.com/vimeo/psalm/pull/4211/files

There's indeed a few more that popped up since then. I actually have a Psalm plugin for getting rid of those somewhere !

I guess we could add it to forbidden functions yeah: https://psalm.dev/docs/running_psalm/issues/ForbiddenCode/

AndrolGenhald: Aha! I _thought_ we had something like that. Looks like that's not documented on the configuration page, so that's why I missed it.

tests/ArrayAssignmentTest.php
```diff
@@ -1543,8 +1648,113 @@ function test(): array {
...
+
+                    /**
+                     * @return array<int, mixed>
+                     */
+                    function foobar(Foo $foo): array
+                    {
+                        return [...$foo];
+                    }
+                ',
+            ],
```
AndrolGenhald, line 1690: I think I meant to go back and look at how this should work, it seems wrong doesn't it? I know we allow omitting template params for arrays, but I'm not sure if we do for traversable. If we do though, I would think it would unpack into `array<array-key, int>` instead of `array<int, mixed>`.

AndrolGenhald: That's only about forcing templates to be explicitly defined, this is about `Traversable` allowing either 1 or 2 templates so that you can omit the key, and this doesn't handle it correctly since it's using the single template as the key instead of as the value.

AndrolGenhald: Turns out I accidentally used `canBeContainedBy` instead of `isContainedBy`, it actually was being resolved as `list<int>` which is a subtype of `array<int, mixed>` so it went unnoticed.

src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
```diff
@@ -493,21 +491,27 @@ private static function analyzeArrayItem(
...
             ) {
                 $array_creation_info->property_types[$item_key_value] = Type::getMixed();
             } else {
                 $array_creation_info->can_create_objectlike = false;
+                $array_creation_info->item_key_atomic_types = array_merge(
+                    $array_creation_info->item_key_atomic_types,
+                    array_values($key_type->getAtomicTypes())
+                );
+                $array_creation_info->item_value_atomic_types[] = new TMixed();
             }
```
AndrolGenhald, line 515: This is actually untested and I couldn't find a way to test it.

orklah:
I'll wait after merging 4.x on master to merge this, it'll probably be easier for you to resolve the conflict

AndrolGenhald:
Rebased, should be gtg now, @orklah not sure if you're done reviewing.

orklah:
Thanks!

PR #8049, opened 2022-06-02
Fix `TypeCombiner::combine` to not modify TIntRange arguments.

AndrolGenhald:
This keeps coming up in obscure places, hopefully this fixes it once and for all. I would reeeaaally love to have an immutable type system at some point...


orklah:
> I would reeeaaally love to have an immutable type system at some point...

Agreed, though I'm afraid about performances (this will result in a loooot of unnecessary cloning

PR #8050, opened 2022-06-02
Improve array-shape list reconcilation (fixes #8046).

### LINKED ISSUE #8046: array_is_list makes the type less specific.
https://psalm.dev/r/ffab419792

The **test1** function passes the test successfully.

Errors are generated for the **test2** function because ```array_is_list``` turns the ```$arr``` type into a ```non-empty-list<array<array-key, string>|string>```.

AndrolGenhald:
Improve reconciliation of array shape with list to keep the array shape.

### DIFF
--- a/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
+++ b/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
@@ -78,6 +78,8 @@ use function assert;
 use function count;
 use function explode;
 use function get_class;
+use function is_int;
+use function ksort;
 use function min;
 use function strpos;
 
@@ -358,6 +360,7 @@ class SimpleAssertionReconciler extends Reconciler
         ) {
             return self::reconcileList(
                 $assertion,
+                $codebase,
                 $existing_var_type,
                 $key,
                 $negated,
@@ -2000,6 +2003,7 @@ class SimpleAssertionReconciler extends Reconciler
      */
     private static function reconcileList(
         Assertion $assertion,
+        Codebase $codebase,
         Union $existing_var_type,
         ?string $key,
         bool $negated,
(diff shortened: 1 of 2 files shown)


AndrolGenhald:
Blocked by #8049.

tests/AssertAnnotationTest.php
```diff
@@ -2055,6 +2055,27 @@ function makeLowerNonEmpty(string $input): string
...
+
+                    /** @var array{0: int, 1: bool, 2: string} */
+                    $list = [1, true, "string"];
+                    assert(is_list($list));
+
+                    function takesString(string $_str): void {}
+
+                    takesString($list[2]);
+                ',
+                'assertions' => ['$list===' => "array{0: int, 1: bool, 2: string}<int<3, max>, mixed>"],
```
orklah, line 2077: Could you just check what is the resulting type if you have a `array{0: int, 1: bool, 2: string}<int<3, max>, mixed>` for $var and you do a `$var[] = 'a'`?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2021,20 +2025,53 @@ private static function reconcileList(
...
-                }
+            } elseif ($type instanceof TKeyedArray) {
+                if ($type->is_list) {
+                    $array_types[] = $type;
+                } else {
+                    $did_remove_type = true;
+                    $type = clone $type;
+                    $min_unset_list_key = 0; // Minimum list key not explicitly set
+                    ksort($type->properties);
+                    foreach ($type->properties as $prop_key => $prop_value) {
```
orklah, line 2043: Oh yeah, I guess that's a better way of fixing that than just adding the count property

tests/AssertAnnotationTest.php
AndrolGenhald: It becomes `non-empty-list<bool|int|mixed|string>`, which seems a bit unfortunate since that's strictly less specific than the previous type... Would you prefer if I tried to fix that here or leave it for a future PR?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2021,20 +2025,53 @@ private static function reconcileList(
...
+                        if (!is_int($prop_key)) {
+                            if ($prop_value->possibly_undefined) {
+                                unset($type->properties[$prop_key]);
+                            } else {
+                                // Can't reconcile, type is removed
+                                continue 2;
+                            }
+                        } elseif ($prop_key === $min_unset_list_key) {
+                            ++$min_unset_list_key;
```
AndrolGenhald, line 2052: Actually I was just thinking about this, if we do `array_is_list(...)` on `array{0: int, 1?: int, 2?: int, 3: int}` the `1` and `2` keys should no longer be `possibly_undefined`. I'll go fix that shortly.

SCIF:
Not blocked anymore :)

AndrolGenhald:
Yeah, this is next on my todo list now that #8044 is merged, just haven't had a lot of time lately :stuck_out_tongue:

PR #8051, opened 2022-06-02
Fix possibly empty array shape appearing non-empty (fixes #8048).

### LINKED ISSUE #8048: False negative when using possibly empty array shape as non-empty-list
https://psalm.dev/r/4250172a68

AndrolGenhald:


src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayMergeReturnTypeProvider.php
```diff
@@ -214,6 +214,15 @@ public static function getFunctionReturnType(FunctionReturnTypeProviderEvent $ev
         $inner_value_type = null;
 
         if ($inner_key_types) {
+            /**
```
orklah, line 217: well it's not incorrect but Psalm miss the subtleties to show the type is truthy

AndrolGenhald: I keep forgetting that array shapes are non-exhaustive 🤦
It's interesting that it resolves `$inner_value_type` differently since it's doing `bool ? list<atomic> : list<atomic>` instead of `bool ? array{0: TInt} : list<atomic>`.

It's not that Psalm is missing that it's truthy though, it's missing that because it's truthy it's a non-empty-list.

AndrolGenhald: Fixed on #8050 since it's sort of related to that too and reconciliation has changed significantly on master.

PR #8052, opened 2022-06-02
Simplify `ForeachAnalyzer::getKeyValueParamsForTraversableObject`.

AndrolGenhald:
Depends on #8044.


orklah:
@AndrolGenhald the blocker is no longer blocking. Should we merge this after conflicts are resolved?

AndrolGenhald:
I'll take a look at it soon, I think this was a pretty good simplification reusing other code instead of duplicating a bunch of complicated functionality, so as long as the tests pass it should be ok to merge.

Issue #7863, opened 2022-04-12
Could not get class storage for forwardcompatauthenticatortrait


yuriimatsak:
Getting an error when trying to analize
It does not matter if i ran with flag --no-cache or without

```
Uncaught Exception: InvalidArgumentException Could not get class storage for lexik\bundle\jwtauthenticationbundle\security\authenticator\forwardcompatauthenticatortrait
Emitted in /code/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:46
```

Could be reproduced here 
https://psalm.dev/r/10017f26ce

kkmuffme:
Same for me. I ran with `--no-cache` and `--threads=1` too, to ensure it's not one of those causing it.

Full trace:

```
Uncaught InvalidArgumentException: Could not get class storage for cacheable_request_trait in /opt/composer/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:46
Stack trace:
#0 /opt/composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/TraitAnalyzer.php(66): Psalm\Internal\Provider\ClassLikeStorageProvider->get('Cacheable_Reque...')
#1 /opt/composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php(597): Psalm\Internal\Analyzer\TraitAnalyzer::analyze(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(PhpParser\Node\Stmt\Trait_), Object(Psalm\Context))
#2 /opt/composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php(206): Psalm\Internal\Analyzer\StatementsAnalyzer::analyzeStatement(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(PhpParser\Node\Stmt\Trait_), Object(Psalm\Context), NULL)
#3 /opt/composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/Statements/Block/IfElse/IfAnalyzer.php(68): Psalm\Internal\Analyzer\StatementsAnalyzer->analyze(Array, Object(Psalm\Context))
[...]

PR #8054, opened 2022-06-04
store origin location by ID to speed up psalm by up to 75%

kkmuffme:
In some instances with particularly nested/bad code, psalm can spend minutes on a single file (and it's includes) with getOriginLocations being called over and over again on the same thing.
This runtime caches the locations in a private variable, to speed up psalm up to 75% (worst file I tested)

Similar to https://github.com/vimeo/psalm/pull/8011

### DIFF
--- a/src/Psalm/Internal/Codebase/VariableUseGraph.php
+++ b/src/Psalm/Internal/Codebase/VariableUseGraph.php
@@ -18,6 +18,9 @@ class VariableUseGraph extends DataFlowGraph
     /** @var array<string, DataFlowNode> */
     private $nodes = [];
 
+    /** @var array<string, list<CodeLocation>> */
+    private $origin_locations_by_id = [];
+
     public function addNode(DataFlowNode $node): void
     {
         $this->nodes[$node->id] = $node;
@@ -94,6 +97,10 @@ class VariableUseGraph extends DataFlowGraph
      */
     public function getOriginLocations(DataFlowNode $assignment_node): array
     {
+        if (isset($this->origin_locations_by_id[$assignment_node->id])) {
+            return $this->origin_locations_by_id[$assignment_node->id];
+        }
+
         $visited_child_ids = [];
 
         $origin_locations = [];
@@ -127,6 +134,8 @@ class VariableUseGraph extends DataFlowGraph
 
             $child_nodes = $new_parent_nodes;
         }
+
+        $this->origin_locations_by_id[$assignment_node->id] = $origin_locations;
 
         return $origin_locations;
     }


kkmuffme:
This is ready to be merged :)

orklah:
Could you add the psalm report with a before/after on a big enough project so we see the difference in memory consumption please?

kkmuffme:
With `2455 errors found` in the codebase:

From:
>Checks took 242.62 seconds and used 1,059.082MB of memory

To:
>Checks took 66.33 seconds and used 1,059.087MB of memory

orklah:
can you make sure to run without cache?

I find this suspicious, VariableUseGraph is an object that is kept all along analysis. Caching things in here shoud be visible in the memory usage

kkmuffme:
Did some more testing, and it seems this issue arises from the use of psalm plugins where CodeLocation is used.
(these are not comparable to the previous comment, as it was run on a faster instance, but I did everything from scratch.)

I tried it with the base config (no stubs, no plugins,...) for the codebase I tested above + another bigger code base now.
Also I tested with `--no-cache` and `--threads=1`

**BASE CONFIG**
From:
>Checks took 27.32 seconds and used 402.501MB of memory
>Checks took 67.72 seconds and used 1,838.005MB of memory

To:
>Checks took 27.27 seconds and used 402.501MB of memory
>Checks took 67.85 seconds and used 1,838.005MB of memory

I tried it on my biggest codebase available (11k files), but there psalm has another bug https://github.com/vimeo/psalm/issues/7863 so it failed.

It seems like there is **no impact on performance and memory consumption with the default psalm config** at all.
How this is possible? `getOriginLocations` never gets called with psalm base config.

**CUSTOM CONFIG WITH PLUGINS**
Again without cache, single threaded.

From:
>Checks took 207.57 seconds and used 857.046MB of memory

To:
>Checks took 46.30 seconds and used 859.465MB of memory

Memory consumption increased by 0.3%, speed increased by ~78%

orklah:
IIRC, VariableUseGraph is used for two things: taint detection and unused variable detection. If you base config don't use unused var detection nor taint detection, it may not be called at all.

Then, maybe some plugins or some config in your config file enabled one of the two feature and made use of VariableUseGraph

kkmuffme:
Yeah, thanks. I don't manage the psalm configs/plugins for our team, just doing devops performance work.

While there is an increase in memory consumption that is linear to the number of files analyzed (and to some extent to what is in those files), the increase in performance is completely worth it in my opinion.
I tried a few additional codebases now, and in all cases the speed increase surpassed the increased memory consumption manifold.

orklah:
Thanks!

PR #8055, opened 2022-06-04
Performance: cut the selected_text from snippet

kkmuffme:
instead of from full text

50% faster than cutting from full text, improves psalm performance up to 3% depending on file length and number of errors in file


kkmuffme:
This is ready to be merged :)

orklah:
Thanks

Issue #7902, opened 2022-04-26
Missing param type causes false positive for missing type for all subsequent param/return


kkmuffme:
https://psalm.dev/r/af8681be3d shows a `MissingReturnType` error even though the return type is there.
When making the @param valid, the `MissingReturnType` error goes away.
https://psalm.dev/r/3027253976

https://psalm.dev/r/4011d3af99 shows `MissingParamType` for `$hello` even though it's there
but:
https://psalm.dev/r/ebd7f96328 correctly only reports `MissingParamType` only for `$bar`

Summary:
when specifying a @param without a type, this will break all subsequent param/return types and falsely report an error.

AndrolGenhald:
Error recovery during docblock parsing is a bit harder than it seems because we support multi-line tags for things like [array shapes](https://psalm.dev/r/62de1ac672). I was worried an `@` in an array shape key would completely mess things up, but it looks like we [require](https://psalm.dev/r/2a972a82e6) that to be [quoted](https://psalm.dev/r/26fb84f6b4) already anyway, so this could probably be fixed without too much trouble.

kkmuffme:
It's not really an issue with matching here though - psalm correctly parses it (and outputs the error)

I think it boils down to the fact that in case the param is invalid, FunctionLikeDocblockParser::parse just skips that param in further handling, thus causing all kinds of follow up issues.
e.g. [params] will only include the valid ones, which means the "params" array count does not match the actual number of @param comments specified.

orklah:
Any error in parsing a docblock emits an InvalidDocblock and will prevent any data from being read from said docblock.

Even previously parsed tags are just ignored. This could probably be improved

kkmuffme:
That would really be necessary, since this causes lots of false positives, e.g. the return type is not missing at all, just the @param is invalid. e.g. in this very basic case 2/4 errors are false positives due to this behavior.

kkmuffme:
That's a good input, however in some instances it cannot be fixed (e.g. bc it's a library, and I'm not going to fix 100s of files, where there is some issue in the docblock, but the @return is fine).
But bc I call functions from this library, I don't get accurate return types, since psalm chooses to ignore the whole invalid docblock.

At least changing it for "Badly-formatted @param" extremely helpful. Right now it throws (and therefore returns from the loop)

orklah:
Well, feel free to propose a PR to improve on that then, even if fixing the source could help every user of this library :)

kkmuffme:
[...]

PR #8056, opened 2022-06-05
fix false positives for partially invalid phpdoc

kkmuffme:
Fix 2 cases specified in https://github.com/vimeo/psalm/issues/7902

Can someone please help me fix the tests, since I don't know how to mock the CodeLocation in FunctionLikeDocblockParserTest.php?

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php
@@ -3,12 +3,15 @@
 namespace Psalm\Internal\PhpVisitor\Reflector;
 
 use PhpParser;
+use Psalm\CodeLocation;
 use Psalm\DocComment;
 use Psalm\Exception\DocblockParseException;
 use Psalm\Exception\IncorrectDocblockException;
 use Psalm\Internal\Analyzer\CommentAnalyzer;
 use Psalm\Internal\Scanner\FunctionDocblockComment;
 use Psalm\Internal\Scanner\ParsedDocblock;
+use Psalm\Issue\InvalidDocblock;
+use Psalm\IssueBuffer;
 
 use function array_shift;
 use function array_unique;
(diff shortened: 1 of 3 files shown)


orklah:
if you make the test extends Psalm\Tests\TestCase, you should be able to do basically this: https://github.com/vimeo/psalm/blob/20351c6667e4f4e2443ea9aca6d5cde610aaaf8d/tests/TestCase.php#L140. FileAnalyzer can be given to CodeLocation and only needs a ProjectAnalyzer that you can retrieve on the properties

kkmuffme:
@orklah sorry, could you pinpoint me on how to proceed? I updated everything for the test cases, however I still don't get how to get the CodeLocation (see my @todo comment in the last commit)

orklah:
just go with new CodeLocation and give it what it asks for. One of those thing is a FileAnalyzer that you can retrieve like I explained earlier

kkmuffme:
That is the exact question. CodeLocation wants:

```
FileSource $file_source,
PhpParser\Node $stmt,
```

I have nothing of that sort available in those tests, since I only have a php docblock string.
How would `new FileAnalyzer( ` help? I don't have a file either, just this php docblock string.

orklah:
FileAnalyzer is a FileSource so you can give it to CodeLocation.

For the Node, you can do a new String_('randomString');, it should work

kkmuffme:
@orklah all mocked now except `PhpParser\Node`, which causes the tests to fail. Please advise on how to mock this here, so tests pass.

orklah:
you can do a `new String_('randomString');`, it should work

kkmuffme:
@orklah But it doesn't :-) I tried all variations of that and none worked unfortunately.

orklah:
String_ must obviously be a child of Node, so it's in PHPParser, not in global namespace: PhpParser\Node\Scalar\String_

Don't you have an IDE to help you with that?

kkmuffme:
I'm devops not a PHP developer and I don't have this setup on my local computer. Working on SSH terminal. That's why I cannot run tests but always have to use the Github pipeline to test the tests.

Anyway, all tests pass now. It's ready to be merged :-)

orklah:
Thanks!

PR #8069, opened 2022-06-08
Fix generic object comparison to use template constraint as default.

### LINKED ISSUE #8068: class-string false positive for constrained template
https://psalm.dev/r/b2a12349ff

Since `Foo`'s template is constrained to `string` and `bar` takes `class-string<Foo>`, it should default that template to `class-string<Foo<string>>` and `Foo::class` should satisfy that.

Related to #7913 (which is for the case where `bar` takes something like `class-string<Foo<non-empty-string>>` which is currently impossible to satisfy), but this is a special case that should be easy to fix.

AndrolGenhald:
Fixes #8068.

### DIFF
--- a/src/Psalm/Internal/Type/TemplateStandinTypeReplacer.php
+++ b/src/Psalm/Internal/Type/TemplateStandinTypeReplacer.php
@@ -33,6 +33,7 @@ use Psalm\Type\Atomic\TTemplateParamClass;
 use Psalm\Type\Union;
 use Throwable;
 
+use function array_fill;
 use function array_keys;
 use function array_merge;
 use function array_search;
@@ -42,6 +43,7 @@ use function count;
 use function in_array;
 use function reset;
 use function strpos;
+use function strtolower;
 use function substr;
 use function usort;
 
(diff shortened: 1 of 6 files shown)


AndrolGenhald:
This will conflict with #8044 when both are merged to master, this one is correct.

src/Psalm/Internal/Type/TemplateStandinTypeReplacer.php
```diff
@@ -1134,7 +1136,21 @@ public static function getMappedGenericTypeParams(
...
+
+            $container_class = $container_type_part->value;
+
+            if (strtolower($input_type_part->value) === strtolower($container_type_part->value)) {
+                $input_type_params = $class_storage->getClassTemplateTypes();
+            } elseif (!empty($class_storage->template_extended_params[$container_class])) {
+                $input_type_params = array_values($class_storage->template_extended_params[$container_class]);
+            } else {
+                $input_type_params = array_fill(0, count($class_storage->template_types ?? []), Type::getMixed());
+            }
```
AndrolGenhald, line 1152: This will conflict with https://github.com/vimeo/psalm/pull/8044 when both are merged to master, this one is correct.

orklah:
Thanks!

PR #8080, opened 2022-06-09
Skip native classlike aliases

### LINKED ISSUE #8079: Native classlike alias failure
Psalm fails to correctly populate the classlike storage when some native extension like https://github.com/jbboehr/php-psr creates aliased classlikes, as seen in this simple example: https://paste.daniil.it/psalm.tar.xz

danog:
This pull request fixes reflection issues with extensions like https://github.com/jbboehr/php-psr which define class aliases.
With this pull request, native classlike aliases are simply ignored, and only the original native classlikes are loaded using reflection.

This fixes #8079 by letting Psalm fallback onto the userland polyfills, instead.


orklah:
Thanks!

PR #8098, opened 2022-06-13
Fix nasty parameter storage corruption bug

danog:
This fixes a very nasty parameter storage corruption bug, which lead to very hard to reproduce, order-dependent false positives and false negatives in our codebase.


danog:
I'm actually already using psalm v5, so I don't mind getting it merged into master :)

orklah:
Thanks!

PR #8114, opened 2022-06-15
number_format: allow third and forth parameters to be nullable

### LINKED ISSUE #8111: number_format erroneously report null argument error
Even if PHP code is valid and has same result in PHP 7 and PHP 8: https://3v4l.org/c566Y
Psalm reports an error: https://psalm.dev/r/d3ab8d92ee

Same issue exists also for the 4th parameter.

drealecs:
This should allow in number_format both decimal_separator and thousands_separator parameters to be nullable in PHP 7 and PHP 8.
In PHP 7 there will be two forms:
```
['string', 'num'=>'float|int', 'decimals='=>'int']
['string', 'num'=>'float|int', 'decimals'=>'int', 'decimal_separator'=>'?string', 'thousands_separator'=>'?string']
```
While in PHP 8 only one form will remain as:
```
['string', 'num'=>'float|int', 'decimals='=>'int', 'decimal_separator='=>'?string', 'thousands_separator='=>'?string']
```
This should fix #8111

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -9027,7 +9027,7 @@ return [
 'nsapi_response_headers' => ['array'],
 'nsapi_virtual' => ['bool', 'uri'=>'string'],
 'nthmac' => ['string', 'clent'=>'string', 'data'=>'string'],
-'number_format' => ['string', 'num'=>'float|int', 'decimals='=>'int', 'decimal_separator='=>'string', 'thousands_separator='=>'string'],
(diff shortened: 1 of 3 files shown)


drealecs:
Do you think some tests would be good here?
I could look a bit into how we could write some, both covering PHP 7 and PHP 8, if needed. But I could do that probably only starting next week.

orklah:
Seems good as is :) Thanks!

PR #8118, opened 2022-06-16
Complete type aliases for int type 

BruceGitHub:


BruceGitHub:
Beside the test, Which other point I should to change
@andreybolonin

BruceGitHub:
I'm busy, but soon 
I will return back... to complete this task

PR #8140, opened 2022-06-22
Update DOM stub and add separate SimpleXML stub

AndrolGenhald:


stubs/extensions/simplexml.phpstub
```diff
@@ -0,0 +1,118 @@
...
+ */
+class SimpleXMLIterator extends SimpleXMLElement implements RecursiveIterator, Countable
+{
+    public function count(): int {}
+
+    public function rewind(): void {}
+
+    public function valid(): bool {}
+
+    public function current(): SimpleXMLIterator {}
```
AndrolGenhald, line 78: Technically incorrect, it can return null until PHP 8.1, but I don't think calling `current()` directly is very common, so it's probably preferable to trade the very uncommon false negative for much more common false positives.

Related to php/doc-en#1640

tests/ArrayAccessTest.php
```diff
@@ -936,7 +936,9 @@ function foo(SimpleXMLElement $s) : SimpleXMLElement {
             'simpleXmlArrayFetchChildren' => [
                 'code' => '<?php
                     function iterator(SimpleXMLElement $xml): iterable {
-                        foreach ($xml->children() as $img) {
+                        $children = $xml->children();
+                        assert($children !== null);
+                        foreach ($children as $img) {
```
AndrolGenhald, line 941: This was a false negative: https://3v4l.org/rqvMc

orklah:
Thanks!

Issue #8132, opened 2022-06-21
Immutable storage [breaking change proposal]


danog:
Opening this issue to allow discussion of possible breaking changes in psalm v5 to allow the immutability of types stored inside storages.

I've been encountering a good number of parallelism bugs, that are actually execution order bugs caused by corruption of some part of function, method, class or argument storages.
To debug this, in my fork I have "frozen" the contents of all storages after scanning, by recursively rendering immutable all stored Unions, using a flag that triggers exceptions when using mutator methods.

While this has helped me a lot in finding bugs (which led to bugfix PR #8131), another bug I manually  found and fixed in #8098 isn't detected by rendering Unions immutable, and is probably caused by a reassignment of the `as` property in a `TTemplate*` Atomic type, which is obviously not detected even if the Union in the `as` itself is also marked immutable.

This is more problematic, and highlights a problem with the struct-like nature of `Atomic` types: lacking encapsulation, one can't easily detect the re-assignment of properties in `Atomic` types.

The breaking change I'm proposing to solve this issue (and to help detect misuse through static analysis) is to render all `Atomic` classes `@immutable`, and create a separate `AtomicMutable` class hierarchy which _is_ mutable, with methods to convert between the two, sharing common methods between the two using traits; the same could be done for Unions, using MutableUnions instead.

This would allow statically detecting misuse of types (including in plugins, etc), and even at runtime if we were to require PHP 8.1 for psalm v5 to use `readonly` properties.

danog:
cc'ing maintainers @orklah @weirdan @muglug :)

orklah:
IIRC, we discussed making Unions immutable (and Atomics would follow) once with @AndrolGenhald. I think it's a very good idea but It may be a massive work: Psalm change properties a lot on Unions and Atomics...

I'm not sure I get the appeal of having both Mutable and Immutable classes though

danog:
I've gone ahead with a draft refactoring, currently settling more on always-immutable Atomics and Unions and a mutable MutableUnion.

weirdan:
I'd like to avoid `Mutable*` hierarchy if at all possible.

AndrolGenhald:
I've wanted this for a long time, but it's a lot of work. I've been thinking a good way to refactor it would be to add a `VariableStorage` similar to `PropertyStorage` and move a lot of the non-type related stuff from `Union` to that. It would be good to have the type system be not so tightly coupled with the rest of the code.

[...]

PR #8143, opened 2022-06-22
Begin immutable refactoring

### LINKED ISSUE #8310: Type with templated closure is not inferred correctly
Turns out I actually fixed this bug in my immutable refactoring, couldn't initially figure out why was upstream returning this weird result :)
https://psalm.dev/r/a2a2c35578

danog:
As per #8132, fixes #8310


danog:
Settling on a simpler structure based on ImmutableUnions to avoid churn.

src/Psalm/Internal/Algebra/FormulaGenerator.php
```diff
@@ -153,7 +153,6 @@ public static function getFormula(
                         $redefined = false;
 
                         if ($var[0] === '=') {
-                            /** @var string */
```
orklah, line 156: This is only possible because you set composer to PHP 8.1. We decided that Psalm must still be able to run on PHP 7.4 for its version 5.

You should revert the change on composer to still allow 7.4, otherwise this won't be merged in Psalm 5 (and will probably won't be released soon: allowing only PHP8.1 would mean denying two third of our users)

danog: I'll revert this, but still, I see nothing wrong with restricting psalm to 8.1, since it's a development tool that is only ever required as a development dependency and runs mainly in CI jobs, where the PHP version isn't as important and doesn't require changes to production code.

orklah: I don't have stats for that, but there is a lot of Psalm runs that are made manually by developpers locally. Once you take that into account, you have to factor in every developer config. For example, for a dev on windows running Wamp to run their application on their laptop, installing a new PHP version just for Psalm may be a real pain.

Psalm may also need to use reflection on particular extensions that may only be available for certain PHP versions. 

We made a big leap to 7.4 for Psalm 5 (remember Psalm 4 supported PHP 7.1. It was initially 7.3 but users complained). It still represents about 6% of Psalm installs to this day that we're gonna left behind. We feel it's acceptable, but we're not ready to dump 66% (not even if we could say that they can easily change their CI jobs)

danog: That's understandable (especially the windows part), I'll revert this here and keep using 8.1 in my own fork, I really want to use readonly properties to catch eventual runtime corruptions :)

danog:
Almost done here, the only strange thing left is the assertion testcase failure, they're caused by https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php#L766, which relies on an inconsistent internal state of `Union` after template substitution, which the old code caused (the `types` property is correctly populated but the `literal_*_types` properties are empty), it's really strange that we're relying on this behavior and I'm not really sure what to do here, any help? :)

danog:
Actually hang on, it seems like the `allowCanBeNotSameAfterAssertionScalar` unit test itself is wrong: 
```php
<?php
                    namespace Bar;

                    /**
                     * Asserts that two variables are the same.
                     *
                     * @template T
                     * @param T      $expected
                     * @param mixed  $actual
                     * @psalm-assert !=T $actual
                     */
                    function assertNotSame($expected, $actual) : void {}

                    $c = 4;
                    $d = rand(0, 1) ? 4 : 5;
                    assertNotSame($d, $c);

                    function foo(string $a, string $b) : void {
                        assertNotSame($a, $b);
                    }
```

It's basically asserting that `$c != $d`, aka `4 != 4|5` aka `4 != 4 || 4 != 5`, and the first assertion is indeed impossible

danog:
The other unit test failure seems actually like a (partial) improvement over the previous behavior, so I'm actually thinking of correcting the testcase to account for the new behavior

orklah:
I don't get how the unit test is wrong. You should be able to replace `assertNotSame($d, $c);` by `assert($d !== $c);`. It's a valid thing to check

danog:
> I don't get how the unit test is wrong. You should be able to replace `assertNotSame($d, $c);` by `assert($d !== $c);`. It's a valid thing to check

I feel there is a slight distinction (https://psalm.dev/r/05628e8554):
- `assert($expected !== $actual)`: psalm errors if there is no intersection between the types of  `$actual` and `$expected`
- `assertNotSame($expected, $actual)` psalm errors if the type of `$actual` doesn't contain the type of `$expected`

Or at least, that's the current behavior with this MR.
The previous behavior was:
- `assert($expected !== $actual)`: psalm errors if there is no intersection between the types of  `$actual` and `$expected`
- `assertNotSame($expected, $actual)`:
  - if `$expected` is a union type, psalm errors if there is no intersection between the types of  `$actual` and `$expected`
  - otherwise, psalm errors if the type of `$actual` doesn't contain the type of `$expected`

I'm not sure what should be the correct behavior, given that to obtain the previous behavior we relied on an inconsistent state of union, possibly caused by a mistake/bug.

Could anyone clarify on what was the actual, intended, behavior, and if it relied on that inconsistent state of Union?

danog:
I think this is what was @boesing actually meant when using `isSingle`, fixed the testcase.

boesing:
Just as a sidenote, AFAIR the isSingle check was there before I added my stuff. I just refactored it a bit. Probably worth going through the history to find the initial commit which added the check.

I was super confused by these test cases as well when I worked on adding my code. I somehow managed to change the code so it works for the old test cases and my new code.

danog:
This should do the trick nicely I think :)

danog:
All done!

danog:
I'll continue refactoring in a separate MR, I still have to implement atomic immutability (at least for nested Union properties) and switch to @readonly properties in storages.

src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php
```diff
@@ -742,17 +742,21 @@ private static function analyzeNestedArrayAssignment(
 
             $is_last = $i === count($child_stmts) - 1;
 
+            $temp = $child_stmt_dim_type ?? Type::getInt();
```
orklah, line 745: Could you change the variable in something more descriptive please?

src/Psalm/Internal/Analyzer/Statements/Expression/BitwiseNotAnalyzer.php
```diff
@@ -59,6 +60,7 @@ public static function analyze(
                         new TLiteralInt(~$type_part->value) :
                         new TInt;
 
+                    // todo:immutable this makes no sense?
```
orklah, line 63: is this a part of what you intend to fix in a next PR?

orklah (commented):
A few things:
- do we need bustCache in Union still? IIRC, it's used to rerender the internal id when the union changed. It shouldn't do that anymore.
- Could we make a check for all `clone $union` everywhere and get rid of them? it shouldn't be needed anymore
- Could you document this change in README.md. It's actually a big change for plugin authors and they'll need to be aware of this

@weirdan @muglug @AndrolGenhald this is a big fundamental change. I could use your help reviewing this :)

AndrolGenhald:
> do we need bustCache in Union still? IIRC, it's used to rerender the internal id when the union changed. It shouldn't do that anymore.

The scope of this PR is actually smaller than I expected. The `Atomic`s in the `Union` are mutable and can still change so that'll still be required when that happens. Some calls might be removable now though, if they haven't been already.

> Could we make a check for all clone $union everywhere and get rid of them? it shouldn't be needed anymore

It's probably still needed in most cases since `Atomic`s are still mutable.

orklah:
Oh right, didn't thought about that

src/Psalm/Internal/Analyzer/Statements/Expression/BitwiseNotAnalyzer.php
danog: Actually now that I look at it, that code makes perfect sense, removed the comment :)

danog:
>do we need bustCache in Union still? IIRC, it's used to rerender the internal id when the union changed. It shouldn't do that anymore.

As atomics are still mutable (they won't be in my next PR), bustCache still has some possible uses.

>Could we make a check for all clone $union everywhere and get rid of them? it shouldn't be needed anymore

Same thing about atomics + even after immutable atomics, there __might__ be some usecases to cloning types, since I'm still not sure whether to mark as `@readonly` **all** properties of Unions and Atomics (not just those with type Union/Atomic).

>Could you document this change in README.md. It's actually a big change for plugin authors and they'll need to be aware of this

Done!

orklah:
Seems great. Let's wait a few days to see if Matt or Bruce have time to chime in and we'll be good to go :)

Thanks for the massive work!

orklah:
Alright, looks like we didn't get more comments.

I'm ready to merge, I just need to make sure you're ready to do the last part (otherwise, it won't make much sense as is) :)

orklah:
I wonder if this is a change big enough to consider delivering V5 soon and merge for a V6 that we would start immediately after.

All things considered, we probably should have released V5 long ago and started a V6 already. Given how we lack long term vision, we should probably deliver major version more often to avoid piling big changes together...

AndrolGenhald:
> I wonder if this is a change big enough to consider delivering V5 soon and merge for a V6 that we would start immediately after.

FWIW I was hoping to get a V6 relatively soon after V5 anyway (or at least sooner than V5 is after V4) with the major type system changes I've been wanting to do (and never managed to get fully working), but it's probably not going to be until next year till that I can devote a lot of time to that again.

danog:
>I'm ready to merge, I just need to make sure you're ready to do the last part (otherwise, it won't make much sense as is) :)

I'm actually already working on it on the immutable_readonly branch!

orklah:
Yeah, September is as good a date as any, we just need to lock a date and keep it. I'll leave this unmerged for now, it may be hard to merge incoming 4.x changes in the meantime otherwise

muglug:
I think this is a fine v5 change if you want to target September. Sorry I haven’t been more active, but I wanted to say earlier: when it comes to these sorts of big changes, I rely on Psalm and Psalm’s test suite to detect regressions.

Over Psalm’s history I have done many big refactors, and very few of them broke some downstream consumer’s pipeline.

orklah:
Thanks for feedback Matt! Did you had issues with plugin authors on previous majors?

This is definitely a change that will affect them directly

muglug:
No — I think if the intent of the refactoring is sound, people generally “get it”

PR #8147, opened 2022-06-23
Fix signatures of ldap_* functions

othercorey:
There were many functions that had LDAP\Connection instead of LDAP\Result or LDAP\ResultEntry.


othercorey:
I couldn't find these functions in the documentation or php source stubs. Removing them caused other errors in historical maps:

```php
'ldap_control_paged_result' => ['bool', 'link_identifier'=>'resource', 'pagesize'=>'int', 'iscritical='=>'bool', 'cookie='=>'string'],
'ldap_control_paged_result_response' => ['bool', 'link_identifier'=>'resource', 'result_identifier'=>'resource', '&w_cookie'=>'string', '&w_estimated'=>'int'],
```

AndrolGenhald:
https://www.php.net/manual/en/function.ldap-control-paged-result.php
https://www.php.net/manual/en/function.ldap-control-paged-result-response.php

They need to be removed in the 8.0 delta.

orklah:
Thanks!

PR #8148, opened 2022-06-23
Add ReflectionProperty properties

BenMorel:
See: https://www.php.net/manual/en/class.reflectionproperty.php

### DIFF
--- a/dictionaries/PropertyMap.php
+++ b/dictionaries/PropertyMap.php
@@ -450,6 +450,10 @@ return [
     'reflectionparameter' => [
         'name' => 'string',
     ],
+    'reflectionproperty' => [
+        'name' => 'string',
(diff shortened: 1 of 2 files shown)


BenMorel:
I noticed other classes that are missing properties, I'll probably take care of them as well in another PR. Question: in the case of `ReflectionFunction::$name`, inherited from `ReflectionFunctionAbstract`, do I need to put the property in both classes? If I'm not mistaken, this is the case with the CallMap.

orklah:
I would not expect it to be needed, otherwise we would have every standard exception in here too but I'm not certain. Could you take a look at CI failures? Looks like some casts became unnecessary after your changes

orklah:
Thanks!

PR #8150, opened 2022-06-23
Track taints in static properties

muglug:
I noticed that taints through static properties are not recognised — the following code does not currently trigger a security analysis issue:

```php
<?php
class A {
    private static string $last = "";

    public static function getLast(string $s): string {
        $last = self::$last;
        self::$last = $s;
        return $last;
    }
}

A::getLast($_GET["a"]);
echo A::getLast("foo");
```


orklah:
Thanks!

PR #8151, opened 2022-06-23
Fix array_* function signatures

othercorey:
`array_replace` historically could return `null` on a type error, but that didn't seem like a worthwhile edge case to add. It should always return an array and since 8.0 *will* always return an array.


othercorey:
Looks like a few other functions were changed to allow optional arguments, but I have to figure out when they changed since it's not documented.

AndrolGenhald:
> but I have to figure out when they changed since it's not documented.

You can go ahead an open an issue for https://github.com/php/doc-en for that.

othercorey:
> You can go ahead an open an issue for https://github.com/php/doc-en for that.

I submitted patches for missing documentation.

Looks like array_replace $replacements was always optional so that should be the last fix.

orklah:
Thanks for your work on that! That's great!

PR #8152, opened 2022-06-24
Fix filter_var and filter_var_array signatures

othercorey:
Although the parameter stubs and parsing code changed in 8.0, it doesn't look like the expected parameter types ever changed.


orklah:
Thanks!

PR #8153, opened 2022-06-24
Fix datefmt_* function signatures

othercorey:
The history of  `datefmt_create` is confusing, and I think I contributed to that years ago: 

* The `$dateType` and `$timeType` parameters do not accept  `null` and never did. What happened was `null` was coerced to `0` which was a valid formatting constant.

* In PHP 8.0, those parameters were given default values (actual constant values, not `null`).

* In PHP 8.0, the `$pattern` parameter was made nullable instead of defaulting to an empty string.

* Also, testing back to 7.0, `datefmt_create()` always returned NULL on error not false.

`datefmt_set_timezone_id` was removed in PHP 7.0 so removed from historical which seems to start at 7.0.


orklah:
Thanks!

PR #8155, opened 2022-06-24
Run method call prohibition analyzer when cloning

danog:
May or may not be used later in the immutable refactoring.


AndrolGenhald:
Hmm, I'm not too sure what the best way to handle this is. Usually deprecated methods are intended to be removed in the future, but if a deprecated `__clone` is removed that would mean Psalm would start to allow cloning again with no warning. I think a `@no-clone` tag on the class would be better, but I don't know how useful it would actually be in practice.

orklah:
Well if we want to deprecate cloning, we tag it as deprecated but instead of removing the method later, we will remove the deprecated tag, make the function private and throw an exception in it. That way the method won't be called again

AndrolGenhald:
I was wondering if adding something like `@no-clone` would be better so Psalm could check it instead of just failing at runtime, but maybe it's not a common enough use-case to be worth adding.

orklah:
Thanks!

Issue #8163, opened 2022-06-25
Infer non-empty assertion from count with variable


AndrolGenhald:
Comparing `count()` to integer literal and integer range variables should cause a non-empty assertion.

https://psalm.dev/r/ad7cd85058
https://psalm.dev/r/cd657ad3a6
https://psalm.dev/r/e706d0612e

Issue #7096, opened 2021-12-08
@psalm-assert with side effects


AndrolGenhald:
Sometimes I have a function that makes assertions about a type that also has other side effects, such as a function that takes a mixed argument and asserts it is a string is less than a maximum length.

If I annotate this with `@psalm-assert string $arg`, psalm will complain about `RedundantConditionGivenDocblockType` whenever I call it on something I already know is a string, but I still want to call it because it has the additional side effect of throwing an exception if the string is too large, as well is if it's not a string.

It would be nice if there were a way to tell psalm that such side effects exist so that it can ignore the function for the `RedundantConditionGivenDocblockType` check.

orklah:
Please provide an example of code

AndrolGenhald:
I guess my use case is more about having a function that is more strict than it's `@psalm-assert` annotation: https://psalm.dev/r/c0165e28b9

But the same issue exists with side-effects: https://psalm.dev/r/9dbaa9a230

AndrolGenhald:
@weirdan That's a much better explanation. Actually, I think if we treated assertion functions as having side effects unless the function is also `@psalm-pure` that would work perfectly.

Maybe `@psalm-assert` should imply `@psalm-pure`, but there needs to be some way of removing that implication, like having `@psalm-impure`? I'm not sure if it would be good or bad to have assertion functions default to warning about impurity.

AndrolGenhald:
In #8378 we have functions that are a good candidate for using `@psalm-assert` with a conditional type. The functions are pure, but the assertion should never be redundant, so I don't think `@psalm-assert-impure` is the way to go. Maybe `@psalm-assert-imprecise` since the assertion is less precise than what the function is actually checking (due to more specific types not actually being possible)?

orklah:
yeah, except that in the exemple, sometimes it also asserts that the type is more precise so I don't like `imprecise` terminology (and honestly, I also don't like we would end up with `@psalm-assert-imprecise-if-true`)

What we're dealing with in 8378, is pretty different to me. We're actually abusing the assertion module (that was used to express that the goal of the function was to refine the type of a param) to describe a mere side effect or an implied truth about the function

In that regard, most calls to str_starts_with or similar could end up Redundant from the point of view of the assertion module and this is the core of the issue.

So, there are two solutions to find:
[...]

PR #8164, opened 2022-06-25
Encapsed literal strings

AndrolGenhald:
- Infer literal string from encapsed string.
 - Allow casting literal float to literal string
 - Improve concatenation of literals to allow literal floats, avoid long loop if there are too many combinations instead of checking after loop
 - Improve `AssertionFinder::hasNonEmptyCountEqualityCheck()` and add tests for #8163


src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -1499,50 +1501,35 @@ protected static function hasNonEmptyCountEqualityCheck(
...
+            return false;
+        }
 
-        if ($right_count
-            && $conditional->left instanceof PhpParser\Node\Scalar\LNumber
-            && $operator_less_than_or_equal
-            && $conditional->left->value >= (
-                $conditional instanceof PhpParser\Node\Expr\BinaryOp\Smaller ? 0 : 1
-            )
+        // TODO get node type provider here somehow and check literal ints and int ranges
```
orklah, line 1526: you can pass $source from caller method and then do `$source->getNodeTypeProvider()` if you want

AndrolGenhald: I can't believe I missed that, I was looking for a `StatementsAnalyzer`... I'm about done for now, but I'll come fix that in the next day or so.

orklah: otherwise, the rest seems good :) Give me a sign when it's ready ;)

AndrolGenhald: Ok, well this is actually a bit complicated. I got it all working, but I got a couple `RedundantCondition`s... I have to figure out a way to make the assertion not redundant if `$foo` and `$bar` are both non-empty and there's an assertion for `count($foo) === count($bar)`.

Go ahead and merge this, I'll have to dig deeper later. I'll probably see if I can fix this on master along with #7096.

orklah:
Thanks!

PR #8165, opened 2022-06-25
Improve @psalm-internal and prevent usage of IssueBuffer::add().

AndrolGenhald:
I finally got around to investigating why I'm getting `ParseError`s from vendor files, so I decided to make sure stuff like that won't happen again.

 - Improve `@psalm-internal`
   - Allow multiple namespaces/identifiers
   - Allow specifying class names
   - Allow specifying functions and methods
 - Make `IssueBuffer::add()` internal so that it can't be used accidentally.

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -135,6 +135,7 @@ class ClassAnalyzer extends ClassLikeAnalyzer
         }
     }
 
+    /** @return non-empty-string */
     public static function getAnonymousClassName(PhpParser\Node\Stmt\Class_ $class, string $file_path): string
     {
         return preg_replace('/[^A-Za-z0-9]/', '_', $file_path)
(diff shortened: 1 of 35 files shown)


AndrolGenhald:
I suppose it might be somewhat debatable whether these ParseErrors should show for vendor files or not, it's a mystery to me why `composer` is keeping `symfony/process` installed when it requires PHP `>=8.1`, but if it were actually used it would crash and Psalm wouldn't notice. On the other hand, it's entirely possible to have some implementations depend on PHP version and conditionally require them, so parsing errors that can never actually show up should definitely not be shown for vendor files.

orklah:
Looks fine :)

About vendors, I'd say the ideal behaviour would be to crash if we encounter a docblock(or signature) that contains wrong elements (at method level or on class level for templates for example). The rest of the errors are debatable...

Let me know if it's ready to merge and I'll do it

AndrolGenhald:
Go ahead :+1:

orklah:
Thanks!

PR #8166, opened 2022-06-25
Allow testing expected CallMap return types

othercorey:
This test the CallMap return type against the reflected return type similar to the way parameter types are checked.

Because we already have an ignore list where users were only fixing parameters, this adds a separate ignore list that includes functions where only the return type is ignored. If a function is in the original `$ignoredFunctions` list, it skips all checks.

Cleaned up some of the debug testing code to handle unexpected types without killing the test script.

Return types that include `resource` fail when compared to the reflected type so they are ignored until that is resolved.


othercorey:
I added this because I didn't realized until after fixing the array_* function signatures that return types were ignored. I wanted a way to check those and also allow for other functions with parameter-only fixes.

othercorey:
The tests are failing because the ignore list was generated  with only some extensions enabled. Need to add the failing functions to the ignore list.

othercorey:
Looks like the `psalm shepherd` test needs to run with PHP 8.1 the same as the unit tests for this to pass.

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -446,36 +617,42 @@ private function isIgnored(string $functionName): bool
...
         }
 
-        $this->expectException(ExpectationFailedException::class);
+        $function = new ReflectionFunction($functionName);
+        /** @var string $entryReturnType */
+        $entryReturnType = array_shift($callMapEntry);
+
+        if ($functionIgnored) {
+            try {
+                /** @var array<string, string> $callMapEntry */
```
orklah, line 633: can you document this with an @param maybe?

othercorey: Added one for the CallMap entry. I don't think there's a way to define the shape so it's just `array<int|string, string>`

I am traveling until Friday so you can make changes or wait until I'm back for further issues.

othercorey:
@orklah Fixed the psalm errors and updated some parameter annotations. If I have time, I will come back for another round of clean up and try to get the class methods working as well. For now, this only adds return type testing for functions.

psalm.xml.dist
```diff
@@ -166,5 +166,12 @@
                 <directory name="vendor/nikic/php-parser" />
             </errorLevel>
         </MixedAssignment>
+
```
orklah, line 169: Could you suppress them with an @psalm-suppress instead please? Putting them here would mean any future usage of those outside of tests would not get flagged by Psalm so I'd like to avoid this

othercorey: done

AndrolGenhald:
> We need an equality check for the types instead of using IsContainedBy()

That's something I plan to do with longer term type system improvements. There's `Union::equals` right now, but it's probably not what you want.

othercorey:
> That's something I plan to do with longer term type system improvements. There's `Union::equals` right now, but it's probably not what you want.

The main issue this causes is nullable types aren't checked. If the PHP type is `int|null` and we just have `int` the test passes.

An example that is `easter_days` where `year` should be nullable:

`'easter_days' => ['int', 'year='=>'int', 'mode='=>'int'],`

https://www.php.net/manual/en/function.easter-days.php

AndrolGenhald:
> The main issue this causes is nullable types aren't checked. If the PHP type is int|null and we just have int the test passes.

Oh, I didn't notice that in the diff because it was already there. You can just do `UnionTypeComparator::isContainedBy($codebase, $a, $b) && UnionTypeComparator::isContainedBy($codebase, $b, $a)`. It's a bit verbose, but it's the best way to do it for now.

orklah:
Thanks!

othercorey:
> Oh, I didn't notice that in the diff because it was already there. You can just do `UnionTypeComparator::isContainedBy($codebase, $a, $b) && UnionTypeComparator::isContainedBy($codebase, $b, $a)`. It's a bit verbose, but it's the best way to do it for now.

FYI, a quick test shows this doesn't work with certain subsets such as for `strlen`. The reflection type is `int` which contains `0|positive-int` but `0|positive-int` does not contain `int`.

PR #8181, opened 2022-06-27
Avoid re-parsing template types

danog:
This caused the creation of a recursive type in the immutable refactoring, and a uselessly nested type in origin/master.


orklah:
Thanks!

PR #8188, opened 2022-06-27
Stop using deprecated string interpolation syntax in StatementsProvider.php

TimWolla:
This fixed the following PHP 8.2 deprecation:

> Uncaught RuntimeException: PHP Error: Using ${var} in strings is deprecated, use {$var} instead in …/vimeo/psalm/src/Psalm/Internal/Provider/StatementsProvider.php:140

see: https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation


weirdan (approved):

weirdan:
Thanks!

PR #8192, opened 2022-06-28
[5.x] Allow to specify an absolute URL for a custom Shepherd endpoint

lptn:
--shepherd CLI option has an undocumented functionality - you can specify a custom Shepherd domain (instead of default `shepherd.dev`). The problem is that down the road the plugin automatically appends `/hooks/psalm` to this domain, so a user doesn't have an option to use use another URL on the custom domain, what is critical in some cases.

In this PR I have changed `--shepherd` option to accept absolute URL instead of domain.

For users, who used `--shepherd` option without a variable (custom domain), these changes are not BC.

Related discussion: https://github.com/vimeo/psalm/discussions/8127

PS: I'll create a new PR for documentation update - to mention this hidden feature in new docs for v5

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -845,3 +845,4 @@
    - :information_source: migration possible using `Psalm\PluginFileExtensionsSocket`
  - [BC] Method `\Psalm\Plugin\EventHandler\Event\AfterFunctionLikeAnalysisEvent::getClasslikeStorage()` was removed,
    use correct `\Psalm\Plugin\EventHandler\Event\AfterFunctionLikeAnalysisEvent::getFunctionlikeStorage()` instead
+ - [BC] Property `\Psalm\Config::$shepherd_host` was removed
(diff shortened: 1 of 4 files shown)


orklah:
I'm worried about the upgrade path. People that don't read the README won't understand why Psalm suddenly stop sending result to their endpoint. Could we change the property name in order to clarify that something change? Something like --shepherd-url ?

lptn:
@orklah 

Thanks for your feedback, I understand and share your concern. 

Do you mean to support both -shepherd` and `--shepherd-url`, or support a new one only? If only a new one, then usage of `--shepherd-url` without a variable will look weird (bad DX). If support both - it will increase code complexity at the place where we can easily avoid it.

I think a new major version is a perfect chance to simplify codebase and have such minor breaking changes. Let me propose alternatives:
 1. write to stderr message that this option has a new value expectations
 2. support both domain and absolute URL options on 4.x branch and add a deprecation note (via `trigger_error`) for domain-only values. Support only URLs on 5.x

orklah:
Thanks for your input :)

I'd have indeed left support only for one, but yeah, that'd have been confusing

Option 2 seems ideal then !

Issue #8162, opened 2022-06-24
Paratest crash due to InternalCallMapHandlerTest


AndrolGenhald:
Paratest is now crashing for me on 4.x due to `InternalCallMapHandlerTest`, added in #8104. The error message "Command not found" is incredibly unhelpful, but the issue is due to the command line being too long.

The command is incredibly large, but it includes:

```
'--filter' '/testIgnoresAreSortedAndUnique(?:\s|$)|testGetcallmapReturnsAValidCallmap(?:\s|$)|testIgnoredFunctionsStillFail with data set "_\: \{"0"\:"string","message"\:"string"\}"$|testIgnoredFunctionsStillFail with data set "acos\: \{"0"\:"float","num"\:"float"\}"$|testIgnoredFunctionsStillFail with data set "acosh\: \{"0"\:"float","num"\:"float"\}"$|...'
```

so it looks like the command includes the entirety of the callmap...

According to https://github.com/paratestphp/paratest/pull/154 `--max-batch-size` is supposed to work with data providers, and I'm using `--max-batch-size=16`, so this might be a bug in paratest.

weirdan:
@AndrolGenhald does it still happen for you? I'm running paratest locally just fine (on current master).

AndrolGenhald:
Still fails partway through:

```
$ time vendor/bin/paratest -f --max-batch-size=16
ParaTest v6.6.5 upon PHPUnit 9.5.26 by Sebastian Bergmann and contributors.

....................................S......................    59 / 13480 (  0%)
...........................................................   118 / 13480 (  0%)
...........................................................   177 / 13480 (  1%)
...........................................................   236 / 13480 (  1%)
..................................S........................   295 / 13480 (  2%)
...........................................................   354 / 13480 (  2%)
...........................................................   413 / 13480 (  3%)
...........................................................   472 / 13480 (  3%)
...........................................................   531 / 13480 (  3%)
...........................................................   590 / 13480 (  4%)
...........................................................   649 / 13480 (  4%)
............................S..............................   708 / 13480 (  5%)
...........................................................   767 / 13480 (  5%)
..............SSS..........................................   826 / 13480 (  6%)
...........................................................   885 / 13480 (  6%)
.S.........................................................   944 / 13480 (  7%)
...........................................................  1003 / 13480 (  7%)
...........................................................  1062 / 13480 (  7%)
.......S...................................................  1121 / 13480 (  8%)
...........................................................  1180 / 13480 (  8%)
[...]

PR #8193, opened 2022-06-28
Various minor improvements to speed up by ~10%

kkmuffme:
Improve various minor things to speed up psalm performance by about 10%

No negative impact on memory consumption.

Will invalidate old caches (= most cache files created with any previous psalm versions will not be used anymore and new caches will be created on first run, since the cache hash changed for performance reasons)

### DIFF
--- a/src/Psalm/Internal/Clause.php
+++ b/src/Psalm/Internal/Clause.php
@@ -120,8 +120,14 @@ class Clause
             return false;
         }
 
+        foreach ($other_clause->possibilities as $var => $_) {
+            if (!isset($this->possibilities[$var])) {
+                return false;
+            }
+        }
+
         foreach ($other_clause->possibilities as $var => $possible_types) {
-            if (!isset($this->possibilities[$var]) || count(array_diff($possible_types, $this->possibilities[$var]))) {
+            if (count(array_diff($possible_types, $this->possibilities[$var]))) {
                 return false;
             }
         }
--- a/src/Psalm/Internal/Clause.php
+++ b/src/Psalm/Internal/Clause.php
@@ -10,14 +10,16 @@ use function array_map;
 use function array_unique;
 use function array_values;
 use function count;
+use function hash;
 use function implode;
 use function ksort;
-use function md5;
 use function reset;
 use function serialize;
 use function sort;
 use function strpos;
 use function substr;
+
+use const PHP_VERSION_ID;
 
 /**
  * @internal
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Clause.php
```diff
@@ -106,11 +108,15 @@ public function __construct(
         } else {
             ksort($possibilities);
 
-            foreach ($possibilities as $i => $_) {
+            foreach ($possibilities as $i => $v) {
+                if (count($v) < 2) {
```
orklah, line 112: Is this really useful? I'd expect sort to not execute anything if there's nothing to sort

kkmuffme: I'd expect that to happen too, but alas, this is not what PHP does. Calling sort conditionally is much faster, according to the profiler.

This is something that could be improved in PHP core btw

src/Psalm/Internal/Scanner/DocblockParser.php
```diff
@@ -87,15 +88,15 @@ public static function parse(string $docblock, int $offsetStart): ParsedDocblock
...
+            if (preg_match('/^[ \t]*\*?\s*@([\w\-\\\:]+)[\t ]*(.*)$/', $line, $matches, PREG_OFFSET_CAPTURE)) {
                 /** @var array<int, array{string, int}> $matches */
                 [, $type_info, $data_info] = $matches;
 
                 [$type] = $type_info;
                 [$data, $data_offset] = $data_info;
 
-                if (strpos($data, '*')) {
-                    $data = rtrim(preg_replace('/^[ \t]*\*\s*$/m', '', $data));
+                if (trim($data) == '*') {
```
orklah, line 98: please use === here

src/Psalm/Internal/Clause.php
orklah: Ok :) Thanks then :)

kkmuffme: https://github.com/php/php-src/issues/8887

src/Psalm/Internal/Scanner/DocblockParser.php
kkmuffme: There's some weird stuff going on in there, I just removed the commit for the docblock parser, as it's a time waste.

kkmuffme:
Branch is ready to be merged.

Sorry for many force push, but the tests don't work for me locally. Some step by step guidance/link on how to run the tests locally (or where the composer is installed, even when on a remote server) would be appreciated.

AndrolGenhald:
@kkmuffme I use `vendor/bin/paratest -f --max-batch-size=16` (tweak batch size to see what's fastest for your hardware), we probably should document that somewhere. Note that you may have to disable the callmap test due to #8162, I haven't had time to look into that yet.

orklah:
Thanks!

Issue #8187, opened 2022-06-27
The type 'object{string:'string'}' is more general than the declared return type 'stdClass'


BenMorel:
This false positive appeared in `4.24.0`:

```php
function test(): stdClass
{
    return (object) [
        'string' => 'string'
    ];
}
```

https://psalm.dev/r/22e82d7a19

> INFO: [LessSpecificReturnStatement](https://psalm.dev/129) - 5:12 - The type 'object{string:'string'}' is more general than the declared return type 'stdClass' for test

> INFO: [MoreSpecificReturnType](https://psalm.dev/070) - 3:18 - The declared return type 'stdClass' for test is more specific than the inferred return type 'object{string:'string'}'

Even though the cast from `array` to `object` yields an `stdClass`: https://3v4l.org/PkHjR

niconoe-:
This is caused in every direction, like this other snippet: https://psalm.dev/r/5dd2ab19c9

> ERROR: [ArgumentTypeCoercion](https://psalm.dev/193) - 9:6 - Argument 1 of test expects stdClass, parent type object{test:'string'} provided

orklah:
Yeah. The casting could be made more subtle in order to describe the type `stdClass&object{string:'string'}` which is a TNamedObject with extra_types. 

If someone wants to make a PR, it's handled here: https://github.com/vimeo/psalm/blob/e0acf22e402d45d1aebb75f4365ebf701fabce2f/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php#L209

edudobay:
@orklah I tried following the way you suggested and it worked — but I stumbled on another problem due to `stdClass` "testing positive" for existence of any property. I've opened a draft PR for suggestions: https://github.com/vimeo/psalm/pull/8194

sapphirecat:
I have just observed this in 5.0.0-rc1.  I have been lackadaisical about running Psalm, so I'm not sure if it appears earlier in the 5.x betas.

niconoe-:
Is there any news about this bug? If I understand well, @edudobay provided a PR on the 4.x version but due to the release of Psalm 5.x, the PR was closed and nothing have been tried again on the 5.x version.
It looks like this issue was almost closed but now it's back to its original state :disappointed: .

Do you need any help on reproducing the PR @edudobay provided but on 5.x?

orklah:
the PR was unfinished. Potentially, nothing would prevent submitting it again and taking review into account for 5.x

[...]

PR #8194, opened 2022-06-29
Draft: Fix stdClass cast

edudobay:
https://github.com/vimeo/psalm/issues/8187

I followed @orklah's suggestion of typing object casts as `stdClass&object{key:value}`, but my solution broke other tests. Turns out that `stdClass` seems to always "test positive" for existence of any property.

For starters, I added a silly check for the specific shape `stdClass & object{some properties here}`, but that doesn't seem right. I'd appreciate suggestions for improving that!

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/GetObjectVarsReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/GetObjectVarsReturnTypeProvider.php
@@ -16,6 +16,7 @@ use Psalm\Type\Atomic\TObjectWithProperties;
 use Psalm\Type\Union;
 use stdClass;
 
+use function count;
 use function reset;
 use function strtolower;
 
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Provider/ReturnTypeProvider/GetObjectVarsReturnTypeProvider.php
```diff
@@ -41,6 +42,14 @@ public static function getGetObjectVarsReturnType(
             $atomics = $first_arg_type->getAtomicTypes();
             $object_type = reset($atomics);
 
+            if ($object_type instanceof TNamedObject
+                && strtolower($object_type->value) === strtolower(stdClass::class)
+                && count($intersection_types = $object_type->getIntersectionTypes() ?: []) === 1
+                && ($inner_object_type = reset($intersection_types)) instanceof TObjectWithProperties
+            ) {
+                $object_type = $inner_object_type;
+            }
```
AndrolGenhald, line 51: This seems a bit hacky, I think what we should probably do is change how `Config::$universal_object_crates` works so that unknown properties aren't allowed to be read without checking if they exist, but writing them is fine (and should produce `&object{newProperty: [type]`, eg for `$obj = object{foo: string, bar: string}`, `$obj->foobar = 1` results in `object{foo: string, bar: string, foobar: 1}`).

Basically I'd like to see [this](https://github.com/vimeo/psalm/issues/7916#issuecomment-1115435524) fixed (so that [this](https://psalm.dev/r/75644eb2ad) can't happen), but maybe there's something I'm not considering here, @orklah do you know if there's a good reason it works like that?

AndrolGenhald: Also, since this would be a bigger change, if we go with it maybe this should target master?

orklah: I don't think there is a big design decision behind how it works right now. Universal object crates, intersection types (on named object) and the cast from array to object were just three separate modules that happen to work relatively well with each other

edudobay: Seems fair! I'll try to figure out something about this on the next few days :)

weirdan:

orklah:

PR #8214, opened 2022-07-03
Fix GEOSGeometry stubs with default values

BenMorel:
Reference:

- `GEOSGeometry::buffer()`: https://github.com/JetBrains/phpstorm-stubs/blob/v2022.1/geos/geos.php#L219
- `GEOSGeometry::simplify()`: https://github.com/JetBrains/phpstorm-stubs/blob/v2022.1/geos/geos.php#L331


orklah:
Thanks! Callmap_historical should be updated too (or whatever PHP version introduced this change)

BenMorel:
@orklah Done!

PR #8216, opened 2022-07-04
Add ReflectionFunctionAbstract tentative function and run shepherd with php 8.1

othercorey:
Trying to allow https://github.com/vimeo/psalm/pull/8166 to run which uses the tentative return types from PHP 8.1.

Shouldn't shepherd and nit tests run against the same version?


orklah:
Ideally, CI should run against every supported php version but that would take way too long. So having at least a version for unit tests and Psalm analysis is better than nothing. Could you revert that part?

othercorey:
> Ideally, CI should run against every supported php version but that would take way too long. So having at least a version for unit tests and Psalm analysis is better than nothing. Could you revert that part?

Do you mean revert the change in shepherd.yml? I couldn't get psalm to pass in the other PR. I assume it is using php 8.0 function signatures when you run shepherd with php 8.0?

orklah:
Yeah. For the other PR, just suppress the issue and explain that the signature doesn't exists before 8.1, it should be enough as long as it's only used in tests

othercorey:
Ok, I'll close this and update the other PR then.

PR #8217, opened 2022-07-05
Fix several CallMap function signatures

othercorey:
Fixed several standard functions.


dictionaries/CallMap_historical.php
```diff
@@ -10001,7 +10001,7 @@
     'copy' => ['bool', 'from'=>'string', 'to'=>'string', 'context='=>'resource'],
     'cos' => ['float', 'num'=>'float'],
     'cosh' => ['float', 'num'=>'float'],
-    'count' => ['int', 'value'=>'Countable|array|SimpleXMLElement|ResourceBundle', 'mode='=>'int'],
+    'count' => ['int', 'value'=>'mixed', 'mode='=>'int'],
```
AndrolGenhald, line 10004: I think this was intentionally more specific.

othercorey: SimpleXMLElement and ResourceBundle implement Countable so they aren't necessary. Should we just make it `Countable|array` for all versions?

orklah: Not for historical versions: https://3v4l.org/N8LQ5

orklah: Though I'm fine removing SimpleXMLElement from the signature starting from 7.4. Less complexity is always good. Not sure about ResourceBundle though

othercorey: I was going to check whether it always implemented that interface too. I'll adjust it back up based on that.

Some of these are tricky to follow the implementation.

AndrolGenhald: According to the [changelog](https://www.php.net/manual/en/class.resourcebundle.php) `ResourceBundle` started implementing `Countable` in 7.4.

othercorey: Restored the historical signature and moved the change to 8.0 where the TypeError was thrown.

orklah:
Thanks!

PR #8219, opened 2022-07-06
fix: ltrim may return class-string #8218

### LINKED ISSUE #8218: ltrim(class-string, '\\') should return a class-string
When using

```php
    /**
     * Removes the prepended backslash of a class string to conform with how php outputs class names
     *
     * @psalm-param class-string $className
     *
     * @psalm-return class-string
     */
    function normalizeClassName(string $className): string
    {
        return ltrim($className, '\\');
    }
```

2 errors will be thrown
```
ERROR: MoreSpecificReturnType - The declared return type 'class-string' for normalizeClassName is more specific than the inferred return type 'string' (see https://psalm.dev/070)
     * @psalm-return class-string

ERROR: LessSpecificReturnStatement - The type 'string' is more general than the declared return type 'class-string' for normalizeClassName (see https://psalm.dev/129)
        return ltrim($className, '\\');
```

It would be a nice addition to avoid this false negative as this specific ltrim of a class string will always result in the same class string being returned

This issue also exists for phpstan https://github.com/phpstan/phpstan/issues/7483

Tofandel:
Fixes #8218

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -557,6 +557,8 @@ function trim(string $string, string $characters = " \t\n\r\0\x0B") : string {}
 /**
  * @psalm-pure
  *
+ * @return ($string is class-string ? ($characters is '\\' ? class-string : string) : string)
+ *
  * @psalm-flow ($string) -> return
  */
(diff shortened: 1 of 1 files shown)


orklah:
Thanks! I'm not a big fan of string manipulation functions returning class-strings, but I guess I see why you need it

PR #8222, opened 2022-07-07
Merge 4.x to master.

AndrolGenhald:


AndrolGenhald:
Check PHP syntax is failing because it can't remove phpunit, running the suggested command gives:

```
$ composer why phpunit/phpunit
phpspec/prophecy-phpunit v2.0.1 requires phpunit/phpunit (^9.1)
```

I have no idea why this would be happening on this pull request and not on current master though.

orklah:
Oh, that's actually a change in your merge PR: https://github.com/vimeo/psalm/pull/8222/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR37

AndrolGenhald:
:facepalm: I should have checked that. Yeah, it's probably not relevant, If weirdan doesn't get back to us in a day or so I'll just remove it and see if the tests all pass.

weirdan:
That step was required to get CI (that particular job) working on PHP 7.1. It's not required to run the job on 7.4 (the minimum supported version for 5.x) and so could be removed in `master`.

AndrolGenhald:
Next time I'm checking the baseline before going and debugging the whole thing...

It's because the type changed from `Iterator` to `Traversable<mixed, mixed>&Iterator`, so it no longer matches the `ArgumentTypeCoercion`s [in the baseline](https://github.com/psalm/endtoend-test-phpunit/blob/master/.psalm/baseline.xml#L910-L919) (it _should_ still match them, but that's a separate bug that's probably been around for a while, lots of stuff doesn't bother checking all types in an intersection, and even if it only checks `Traversable` that should still count as coercion).

I'm going to keep looking into why the type changed, I would think that would still be `Iterator` since `Iterator<mixed, mixed>` is a subtype of `Traversable<mixed, mixed>`.

orklah:
Don't bother too much, the goal is just to make sure there isn't something fundamentally broken. There will always be fringe edge cases when doing massive changes. The master branch is already pretty heavy so it's bound to happen

AndrolGenhald:
Turns out it was actually one of my changes that made `GenericTypeComparator` get the actual templates (along with their bounds) when comparing instead of using `mixed`. I think the templates need to be replaced further up the chain somewhere, but I'm not sure exactly where, so I'll leave it for now.

PR #8228, opened 2022-07-08
Fix nullable return types for CallMap functions

othercorey:
This adds an assertion on matching nullability since this doesn't fail `isContainedBy()`. 

Because there are a lot of types that should be nullable, this only checks return types and fixes the return types for functions that are already "passing". When the parameters for already "passing" functions are fixed, the nullable check can be run unconditionally.

If we ever want to completely ignore a null return type in the future without ignoring the entire return type, the ignore logic could be more sophisticated. 

This type of special check can be done for `false`, but needed to limited the scope so can actually fix the signatures.


dictionaries/CallMap.php
```diff
@@ -1744,7 +1744,7 @@
 'datefmt_format' => ['string|false', 'formatter'=>'IntlDateFormatter', 'datetime'=>'DateTime|IntlCalendar|array|int'],
 'datefmt_format_object' => ['string|false', 'datetime'=>'object', 'format='=>'mixed', 'locale='=>'string'],
 'datefmt_get_calendar' => ['int', 'formatter'=>'IntlDateFormatter'],
-'datefmt_get_calendar_object' => ['IntlCalendar', 'formatter'=>'IntlDateFormatter'],
+'datefmt_get_calendar_object' => ['IntlCalendar|false|null', 'formatter'=>'IntlDateFormatter'],
```
othercorey, line 1747: This also adds missing `false`.

dictionaries/CallMap.php
```diff
@@ -6923,22 +6923,22 @@
...
-'locale_get_region' => ['string', 'locale'=>'string'],
-'locale_get_script' => ['string', 'locale'=>'string'],
-'locale_lookup' => ['string', 'languageTag'=>'array', 'locale'=>'string', 'canonicalize='=>'bool', 'defaultLocale='=>'string'],
-'locale_parse' => ['array', 'locale'=>'string'],
+'locale_get_keywords' => ['array|false|null', 'locale'=>'string'],
+'locale_get_primary_language' => ['?string', 'locale'=>'string'],
+'locale_get_region' => ['?string', 'locale'=>'string'],
+'locale_get_script' => ['?string', 'locale'=>'string'],
+'locale_lookup' => ['?string', 'languageTag'=>'array', 'locale'=>'string', 'canonicalize='=>'bool', 'defaultLocale='=>'string'],
+'locale_parse' => ['?array', 'locale'=>'string'],
```
othercorey, line 6941: These locale functions return null when locale string is too big. Do we want to ignore this possibility?

othercorey:
@orklah Would you like me to add something to finish this? I'll probably have to fix the nullable parameters in parts since it would be a large number.

AndrolGenhald:
Could you fix the [failing test](https://github.com/vimeo/psalm/runs/7246670369?check_suite_focus=true#step:9:110)? Looks like it's a Windows only function.

othercorey:
> Could you fix the [failing test](https://github.com/vimeo/psalm/runs/7246670369?check_suite_focus=true#step:9:110)? Looks like it's a Windows only function.

Fixed. I thought that was a timeout that kept happening.

Had to adjust a stub for sapi_windows_cp_conv.

PR #8240, opened 2022-07-10
use lock to fix race condition in cache

kkmuffme:
When running multiple psalm instances in the same directory (e.g. I want to analyze 2 distinct directories or X files only) at the same time (e.g. CI pipeline), I was getting random "Unexpected value when loading from file content hashes" errors due to race conditions

By using a read/write lock with wait this issue is resolved.
The "write" lock success is not checked, since this issue only happens with simultaneous runs which means the data of multiple writes at the same time the file is locked is the same that is currently being written.

### DIFF
--- a/src/Psalm/Internal/Provider/ParserCacheProvider.php
+++ b/src/Psalm/Internal/Provider/ParserCacheProvider.php
@@ -8,9 +8,14 @@ use Psalm\Config;
 use RuntimeException;
 
 use function error_log;
+use function fclose;
 use function file_get_contents;
 use function file_put_contents;
 use function filemtime;
+use function filesize;
+use function flock;
+use function fopen;
+use function fread;
 use function gettype;
 use function igbinary_serialize;
 use function igbinary_unserialize;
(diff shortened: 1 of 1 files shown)


kkmuffme:
This is ready to be reviewed & merged

src/Psalm/Internal/Provider/ParserCacheProvider.php
```diff
@@ -184,7 +190,29 @@ private function getExistingFileContentHashes(): array
...
+
+                if (!$has_lock) {
+                    fclose($fp);
+                    error_log('Could not acquire lock for content hashes file');
+                    $this->existing_file_content_hashes = [];
+
+                    return [];
+                }
+
+                // file_get_contents ignores any locks and gets the file, but we have the lock on $fp anyway
```
orklah, line 213: Why not use fgets then? file_get_contents would have to reopen a stream, no?

kkmuffme: Changed to fread now

orklah:
Thanks!

PR #8249, opened 2022-07-12
Improve inferring the "final" `static` type when calling static methods inside a different class

### LINKED ISSUE #8200: generics + static = invalid type inference
https://psalm.dev/r/5d79e9217a

someniatko:
fixes #8200

After some insane debugging and trying to wrap my head around it, I found the reason it behaves improperly.

When calling a static method of some class, there are two possibilities:
1. in the static method itself there might be something like `@return Collection<static>`, that means `static` is defined by the static method which we are calling.
2. this static method is parameterized with a template like `@return Collection<T>`, but the class which CALLS it can have something like `@return Collection<static>`. So `static` is defined by the CALLING code, not the CALLED. In this case, Psalm behaved incorrectly, because it always assumed only the 1st case.

I have to move the test case somewhere (for now it's a separate class), please give a hint.


src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/ExistingAtomicStaticCallAnalyzer.php
```diff
@@ -551,6 +551,18 @@ private static function getMethodReturnType(
...
+                    MethodIdentifier::fromMethodIdReference($context->calling_method_id),
+                    $context->self
+                );
+                // differentiate between these cases:
+                //   1. "static" in return type comes from return type of the
+                //   method CALLING the currently analyzed static method - use $context->self.
+                //   2. "static" comes from the CALLED static method - use $fq_class_name.
+                $static_type = $self_method_return !== null && self::hasStaticInType($self_method_return)
+                    ? $context->self
+                    : $fq_class_name;
```
AndrolGenhald, line 565: What is `$self_method_return` here? It's not completely clear to me what the condition is to decide if `static` should mean `$context->self` vs `$fq_class_name`.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/ExistingAtomicStaticCallAnalyzer.php
```diff
@@ -613,4 +625,26 @@ private static function getMethodReturnType(
...
+     */
+    private static function hasStaticInType(Union $union_type): bool
+    {
+        foreach ($union_type->getAtomicTypes() as $atomic_type) {
+            if ($atomic_type instanceof Atomic\TGenericObject) {
+                foreach ($atomic_type->type_params as $type_param) {
+                    if (self::hasStaticInType($type_param)) {
+                        return true;
+                    }
+                }
```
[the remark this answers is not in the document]
someniatko: Yeah, thank you. I'm completely new to both psalm and php-parser code, and just used whatever will make my test pass and not fail the others. Will add a new test and fix this code when I'm back to my machine.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/ExistingAtomicStaticCallAnalyzer.php
someniatko: Hmm, I already see where this logic is going to break. I think I have a better idea, which I'll implement a bit later.

someniatko: Also, both calling and called method could have their own `static`s, but I don't see any way to resolve it for now.

AndrolGenhald: Yeah, it would be good to make sure [this](https://psalm.dev/r/2cc402de71) works. It looks like `TypeExpander` allows `$static_class_type` to be a `TNamedObject`, so you could try turning the `"static"` string into a `TNamedObject` with `$is_static = true` somewhere and play around with it. Not sure exactly how to get that to work correctly though.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/ExistingAtomicStaticCallAnalyzer.php
```diff
@@ -613,4 +622,29 @@ private static function getMethodReturnType(
...
+
+            foreach ($atomic_type->getChildNodes() as $child_type) {
+                if (self::hasStaticInType($child_type)) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
```
AndrolGenhald, line 649: This could be simplified further to just the part inside the loop, something like:

```php
if ($type instanceof TNamedObject && $value === 'static') {
    return true;
}

foreach ($type->getChildNodes() as $child_type) {
    if (self::hasStaticInType($child_type)) {
        return true;
    }
}

return false;
```

That way there's also no need to assert `Union` or `Atomic`, which helps if I ever manage to add an `Intersection` type.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/ExistingAtomicStaticCallAnalyzer.php
someniatko: @AndrolGenhald let's extract handling two different `static`s into a different issue.

someniatko:
meh, when moving test cases I added them to "invalid code" section instead of "valid code"

AndrolGenhald:
Please indent the test code one level past the `<?php` tag to match the other tests.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/ExistingAtomicStaticCallAnalyzer.php
AndrolGenhald: Opened #8257 for this.

AndrolGenhald:
Thanks!

PR #8260, opened 2022-07-14
Fix comparison of named without templates specified to class with templates specified

AndrolGenhald:
This fixes the issue I noticed when merging 4.x to master.

### DIFF
--- a/src/Psalm/Internal/Type/Comparator/GenericTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/GenericTypeComparator.php
@@ -3,11 +3,15 @@
 namespace Psalm\Internal\Type\Comparator;
 
 use Psalm\Codebase;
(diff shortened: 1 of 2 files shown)


AndrolGenhald:
I got it mostly working, but it turns out the implementation of `AtomicTypeComparator::canBeIdentical` is fundamentally flawed, it only checks if `A` is a subtype of `B` or `B` is a subtype of `A`, so the corrected comparison breaks the fact that `Some` can intersect with `Maybe<int>` even though `Maybe<int>` isn't a subtype of `Some<mixed>` and `Some<mixed>` isn't a subtype of `Maybe<int>`.

This should be fixed by my type comparison rewrite, but that's probably also a long way off from being finished. The number of complicated issues like this is a bit discouraging sometimes :disappointed:

weirdan:

orklah:

PR #8273, opened 2022-07-17
Remove special handling for elseifs that breaks for else if

muglug:
This is a weird PR, because it breaks a niche existing analysis, but I think for the right reasons.

The test `twoVarLogicNotNestedWithElseifCorrectlyReinforcedInIf` breaks when the `elseif` is changed to `else if`, and the "elseif" version doesn't work online either: https://psalm.dev/r/a1f9a6439e for _unknown reasons_.

Generally I think it's bad when `elseif` and `else if` are treated differently (maybe I erred in having separate handling for `elseif`). Either way, I think this is a minimally-bad change, and it limits the use of the `Union::substitute` method, which I don't think should be used in if/else handling — if/else handling should just use regular type reconciliation everywhere.

I saw this mismatch when looking at Hack (which has no `elseif`)

### DIFF
--- a/src/Psalm/Context.php
+++ b/src/Psalm/Context.php
@@ -459,7 +459,7 @@ final class Context
         array $vars_to_update,
         array &$updated_vars
     ): void {
-        foreach ($start_context->vars_in_scope as $var_id => $old_type) {
+        foreach ($start_context->vars_in_scope as $var_id => $_) {
             // this is only true if there was some sort of type negation
             if (in_array($var_id, $vars_to_update, true)) {
                 // if we're leaving, we're effectively deleting the possibility of the if types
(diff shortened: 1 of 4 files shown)


orklah:
Feels weird to make a change that only worsen the analysis, even for the good reasons. I guess fixing it for all the cases would be hard, right?

Maybe we should leave a small comment to describe that a feature is missing there?

muglug:
I added a comment!

[The original test](https://github.com/vimeo/psalm/commit/a8220ba4f17bc517db70e64bbf58d2bd42272468) was added over 5 years ago, which likely means it was found in Vimeo code at some point. It's possible that Psalm's analysis will generate false-positives on other codebases due to this change, but I think that's pretty unlikely. And I think slightly-simplifying if/else analysis is worth that disruption.

muglug:
This matches a general limitation of Psalm when dealing with nested versions of things:

Consider https://psalm.dev/r/c8a6cfe83c vs https://psalm.dev/r/9a69e87e69

Both equivalent, but Psalm's special-sauce generation of logical clauses from these sorts of if statements only works at the level they're expressed on.

orklah:
Let's roll with it then! We'll see if we get feedback!

orklah:
Thanks!

Issue #5035, opened 2021-01-16
Detect flags in FilesystemIterator to get a more correct iterator type


xPaw:
https://psalm.dev/r/8930178275
https://psalm.dev/r/de250db5e3

See https://github.com/vimeo/psalm/pull/4725#issuecomment-761555513

Depending on the flag, it may return different things:
`FilesystemIterator::CURRENT_AS_PATHNAME`, `FilesystemIterator::CURRENT_AS_FILEINFO`, `FilesystemIterator::CURRENT_AS_SELF`.

Currently psalm will complain that methods are being called on strings, despite fileinfo being the default flag of `FilesystemIterator`.

cc @orklah

orklah:
I guess we could do something like that:
https://psalm.dev/r/ab96d2798b

It should work pretty well until a certain point: conditional returns don't seem to take masks for flags.

It means that if there is only 1 flag, (or even 0 for the default) it should work well, otherwise it won't.

It will still be an improvement compared to current situation but it's not optimal

@weirdan @muglug Do you have other ideas? Should psalm support something like `@return (T&1 is 1 ? type1 : type2)`

weirdan:
Do you propose to use flags value as template parameter? I don't think we should do that - just consider the inheritance and parameter types. Is FilesystemIterator<123> allowed where FilesystemIterator<456> expected? We shouldn't need to have to answer that.

orklah:
Mmh... didn't thought about that. It's unfortunate, it was nearly working.

I don't have any other idea...

weirdan:
I think we need to figure out what actual type parameter(s) should be. I'd say it probably has a single type parameter that matches the type of `current()`, giving us the following:
```php
/**
 * @template T of self|SplFileInfo|string
 * @template-implements SeekableIterator<string,T>
 * will probably need to adjust DirectoryIterator type and pass T to it as well
 */
class FilesystemIterator extends DirectoryIterator implements SeekableIterator {}
```
This way users will need to use explicit docblock types, turning OP's code into
```php
/** @var FilesystemIterator<SplFileInfo> $Images */
$Images = new FilesystemIterator( '/tmp' );

foreach( $Images as $Image )
{
    echo $Image->getRealPath();
}
```
which looks like acceptable interim solution to me until the time Psalm learns to infer T.

soyuka:
[...]

PR #8282, opened 2022-07-18
Update `CachingIterator::getCache` stub.

drupol:
This PR:

- [x] Fix `CachingIterator::getCache` stub

Examples: https://3v4l.org/i5faS#v8.1.8
False positive in PSalm: https://psalm.dev/r/15a83f948f

### DIFF
--- a/stubs/CoreGenericIterators.phpstub
+++ b/stubs/CoreGenericIterators.phpstub
@@ -358,7 +358,7 @@ class CachingIterator extends IteratorIterator implements OuterIterator , ArrayA
 
     /**
      * @param TIterator $iterator
-     * @param int-mask<1,2,4,8,16,256> $flags
+     * @param int-mask-of<self::*> $flags
      */
     public function __construct(Iterator $iterator, int $flags = self::CALL_TOSTRING) {}
 
--- a/stubs/CoreGenericIterators.phpstub
+++ b/stubs/CoreGenericIterators.phpstub
@@ -355,7 +355,7 @@ class CachingIterator extends IteratorIterator implements OuterIterator , ArrayA
     const TOSTRING_USE_CURRENT = 4 ;
     const TOSTRING_USE_INNER = 8 ;
     const FULL_CACHE = 256 ;
-    
+
     /**
      * @param TIterator $iterator
      * @param int-mask-of<self::*> $flags
@@ -376,6 +376,11 @@ class CachingIterator extends IteratorIterator implements OuterIterator , ArrayA
      * @psalm-ignore-nullable-return
      */
     public function key() {}
+
+    /**
+     * @return array<array-key, TValue>
+     */
+    public function getCache() {}
 }
 
 /**
--- a/stubs/CoreGenericIterators.phpstub
+++ b/stubs/CoreGenericIterators.phpstub
@@ -342,6 +342,7 @@ abstract class FilterIterator extends IteratorIterator {
  * @template-covariant TKey
  * @template-covariant TValue
  * @template-covariant TIterator as Iterator<TKey, TValue>
+ * @template TFlags of int-mask<1,2,4,8,16,256>
  *
  * @template-implements OuterIterator<TKey, TValue>
  * @template-implements ArrayAccess<TKey, TValue>
@@ -358,7 +359,7 @@ class CachingIterator extends IteratorIterator implements OuterIterator , ArrayA
 
     /**
      * @param TIterator $iterator
-     * @param int-mask-of<self::*> $flags
+     * @param TFlags $flags
      */
     public function __construct(Iterator $iterator, int $flags = self::CALL_TOSTRING) {}
 
@@ -378,6 +379,8 @@ class CachingIterator extends IteratorIterator implements OuterIterator , ArrayA
     public function key() {}
 
     /**
+     * @psalm-if-this-is self<mixed, mixed, 256>
+     *
      * @return array<array-key, TValue>
      */
     public function getCache() {}
--- a/stubs/CoreGenericIterators.phpstub
+++ b/stubs/CoreGenericIterators.phpstub
@@ -342,7 +342,6 @@ abstract class FilterIterator extends IteratorIterator {
  * @template-covariant TKey
  * @template-covariant TValue
  * @template-covariant TIterator as Iterator<TKey, TValue>
- * @template TFlags of int-mask<1,2,4,8,16,256>
  *
  * @template-implements OuterIterator<TKey, TValue>
  * @template-implements ArrayAccess<TKey, TValue>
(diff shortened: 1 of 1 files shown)


stubs/CoreGenericIterators.phpstub
```diff
@@ -376,6 +376,11 @@ class CachingIterator extends IteratorIterator implements OuterIterator , ArrayA
      * @psalm-ignore-nullable-return
      */
     public function key() {}
+
+    /**
+     * @return array<array-key, TValue>
```
AndrolGenhald, line 381: That's unfortunate, I would have expected `TKey`, but it looks like it takes the value of the key and [casts it](https://3v4l.org/lkgM0) to `array-key`. I don't think we have a way to annotate that more specifically :slightly_frowning_face:

drupol: That's indeed unfortunate and I couldn't find a way to type it better.

AndrolGenhald:
It'd be nice to show some error if `FULL_CACHE` isn't used: https://3v4l.org/G80RU.

I also expected `CALL_TOSTRING` to cause problems, but according to [this](https://www.php.net/manual/en/class.cachingiterator.php#124983) `FULL_CACHE` overrides it. What a lovely class full of awful edge cases...

Given that `FULL_CACHE` probably shouldn't be used with `CALL_TOSTRING` anyway, it might be ok to add a template for `$flags` and use `@psalm-if-this-is CachingIterator<256>`. We'll also probably want a couple of tests to make sure that works correctly.

AndrolGenhald:
I think the conservative route is to add a `TFlags` template for `$flags` and add `@psalm-if-this-is CachingIterator<mixed, mixed, mixed, 256>` sort of like [this](https://psalm.dev/r/4ea22ebdea). That will only allow calling `getCache()` if `FULL_CACHE` is used.

The downside is that it will result a false positive for `$flags = FULL_CACHE | CALL_TOSTRING`, but that's probably fine since according to the documentation that should also change the value type to `string` ([it actually doesn't](https://3v4l.org/Uiv7q) which confuses the heck out of me, as far as I can tell it calls `_toString`, but then discards the result and uses the element itself as the value...).

drupol:
Thanks, very clear. I will update the PR in a couple of hours.

orklah:
Unfortunately, templates are not very suited for handling flags. While it could work for simple cases, it creates a whole world of weirdness, as @weirdan explained here: https://github.com/vimeo/psalm/issues/5035#issuecomment-761621670

I don't know how other languages handle that kind of thing. But again, other languages may not have such a funky internal API...

AndrolGenhald:
I thought it might be ok in this case since it seems like it's probably fine to only care about the exact value and not the mask, would you be opposed to it? It would be nice to have a more general way to deal with bitmask templates eventually.

drupol:
I just updated the PR, feel free to give another review.

TIL: `@psalm-if-this-is` is a very nice feature !!!

orklah:
Hey @ondrejmirtes,

Do you have an opinion on adding templates to iterators (or possibly any class) to track flags in __construct in order to change return types in other methods?

If we started to do it, would PHPStan do the same in order to keep parity?

Right now, things like inheritance and covariance checks on templates makes me think it would not be such a great idea but I'd like to get more point of view

stubs/CoreGenericIterators.phpstub
```diff
@@ -376,6 +377,13 @@ class CachingIterator extends IteratorIterator implements OuterIterator , ArrayA
      * @psalm-ignore-nullable-return
      */
     public function key() {}
+
+    /**
+     * @psalm-if-this-is self<mixed, mixed, 256>
```
AndrolGenhald, line 382: This needs another `mixed` template. Should probably add a couple of tests somewhere to make sure it's working as intended. Might as well wait to see if @orklah wants to go this route though so you don't waste time only to undo it all.

drupol: Ok I'll wait a bit before revisiting this.

AndrolGenhald: Yep, missed that one, it would need the flags template added so it can extend with the template.

ondrejmirtes:
When adding a new `@template` it's always a struggle of "Is this useful enough to bother every user with this parameter when they're typehinthing this class?"

Apart from `@psalm-if-this-is self<mixed, mixed, 256>` (Not sure what it does), I can update `CachingIterator` stub if you decide to merge this.

AndrolGenhald:
> Apart from @psalm-if-this-is self<mixed, mixed, 256> (Not sure what it does)

It shows an error for [this issue](https://3v4l.org/a0Uc7) because it requires `$iterator` to be constructed with `$flags = 256`, otherwise `$iterator->getCache()` won't be allowed (showing an error like [this](https://psalm.dev/r/4ea22ebdea)). But @orklah might be right that it's not worth the trouble, especially since this class seems terribly designed and full of other problems that we won't be able to warn about anyway.

orklah:
Thanks for your input Ondrej!

Yeah, I don't think we should try to use templates for flags in iterator classes. They're really weird and will always have edge cases. I feel like every time we make a big change in those, it doesn't solve a lot of cases but it always bring back new unsolvable issues a few months later (users were used to the old wrong behaviour and the new one may not be better for their use case)

The rest of the PR seems good though

drupol:
I updated the PR, remove the new `@template` and updated `RecursiveCachingIterator` by removing some extended methods because they were the same as in `CachingIterator`.

stubs/CoreGenericIterators.phpstub
AndrolGenhald: Don't forget to remove this part (sorry for the extra work, I'm sometimes too much of a perfectionist :smile:).

drupol: No worries, I'm also perfectionist and I fully agree. Going to amend my last commit in a minute.

stubs/CoreGenericIterators.phpstub
```diff
@@ -355,10 +355,10 @@ class CachingIterator extends IteratorIterator implements OuterIterator , ArrayA
     const TOSTRING_USE_CURRENT = 4 ;
     const TOSTRING_USE_INNER = 8 ;
     const FULL_CACHE = 256 ;
-    
+
     /**
      * @param TIterator $iterator
-     * @param int-mask-of<self::*> $flags
+     * @param int-mask<1,2,4,8,16,256> $flags
```
orklah, line 361: Why change this? The previous type should work the same way but is more flexible to future changes, no?

drupol: But then we need to provide a template when creating the CachingIterator, and this is what we want to avoid based on what I read on previous comments, no ?

orklah: unless I'm mistaken, `int-mask-of<self::*>` should not need any template

drupol: I don't know, could you confirm?

AndrolGenhald: Well, that's what it had before this PR, so if it works on psalm.dev right now it should be fine: https://psalm.dev/r/ecacca8230

drupol: Fixed.

orklah:
Thanks!

drupol:
cool, merci aussi !

PR #8283, opened 2022-07-18
Allow `value-of` to work with backed enums (fixes #7874).

### LINKED ISSUE #7874: Support for `value-of<MyEnum>`
Psalm doesn't support backed enums as a type argument for `value-of`:

```php
enum ReservationState: string {
    case Pending = 'pending';
    case Confirmed = 'confirmed';
    case Cancelled = 'cancelled';
}

/**
 * @param value-of<ReservationState> $value
 */
function makeState(string $value): ReservationState {
    return ReservationState::from($value);
}

makeState(ReservationState::Pending->value); // Works
makeState('confirmed'); // Works
makeState('foo'); // Doesn't work
```

https://psalm.dev/r/1277567ae5

Expected:
No errors

Actual:
ERROR: [InvalidDocblock](https://psalm.dev/008) - 10:11 - Untemplated value-of param ReservationState should be an array in docblock for makeState

[PHPStan supports this use of `value-of`.](https://phpstan.org/writing-php-code/phpdoc-types#key-and-value-types-of-arrays-and-iterables)

(I've searched the existing issues for this, but surprisingly, I couldn't find one. I'm sorry if I've missed it.)

AndrolGenhald:


tests/ValueOfTest.php
```diff
@@ -212,6 +280,23 @@ function getValue() {
...
+                    {
+                        case Foo;
+                        case Bar;
+                    }
+
+                    /** @param value-of<Foo> $arg */
+                    function foobar(string $arg): void {}
+                ',
+                // TODO turn this into an InvalidDocblock with a better error message. This is difficult because it
+                // has to happen after scanning has finished, otherwise the class might not have been scanned yet.
```
AndrolGenhald, line 295: I'd like to get this to work, but unless there's a single point of analysis for docblocks after scanning has finished I'm not sure how, so I figure it can wait.

AndrolGenhald:
@orklah Do you think we should proactively change `TKeyOfArray` to `TKeyOf` even though it only works for arrays atm?

amberovsky:
Hi @orklah @AndrolGenhald ,
I can reproduce this on 4.27.0 - just wondering if this fix slipped off from the release or it is not ready be released yet?

orklah:
4.27 should contain the fix. Do you have an example snippet?

amberovsky:
It's the same as above with the same error message.
I can't the see fix in the change log and in [the commit diff?](https://github.com/vimeo/psalm/compare/4.26.0...4.27.0)
I also checked the code - it's not there e.g. in the 4.27 there are no changes in [this file](https://github.com/vimeo/psalm/commits/4.27.0/docs/running_psalm/plugins/plugins_type_system.md)

AndrolGenhald:
Actually this was targeted to the master branch, so the first version it's in is 5.0.

@orklah maybe we should release another beta soon?

orklah:
oh sorry, didn't check the target indeed.

We can release a beta if we want. I was hoping a final release in september after the immutable refactor but we may have to wait a bit more

PR #8287, opened 2022-07-19
Add CI test for minimum dependencies

AndrolGenhald:
Given how long it takes I don't think we want to have this run on every commit and pull request, so I think after I get all of the issues fixed we'll probably want to split this into a separate workflow that runs before releases, and perhaps daily or weekly (probably with an extra check like [this](https://github.community/t/trigger-action-on-schedule-only-if-there-are-changes-to-the-branch/17887/4)).


weirdan:

orklah:

Issue #8289, opened 2022-07-20
Error with array references in Reconciler 


ohader:
https://psalm.dev/r/03af5df16f

Check XHR response in browser's dev console (actually it's "just" a warning, however analyzer stops)

```
Warning: Undefined array key "$doesNotMatter[$a]" in /var/www/vhosts/psalm.dev/httpdocs/vendor/vimeo/psalm/src/Psalm/Type/Reconciler.php on line 128
{"error":{"message":"\/vendor\/vimeo\/psalm\/src\/Psalm\/Type\/Reconciler.php: __clone method called on non-object","line_from":128,"type":"psalm_error"}}
```

https://github.com/vimeo/psalm/blob/e268a053070ca1272df727673ea4f750ab3212b6/src/Psalm/Type/Reconciler.php#L128

ohader:
@AndrolGenhald Just create a test-case for this, feel free to reuse them from PR #8290

AndrolGenhald:
I'll try to get this fixed soon, but I've given up on further improving support for arrays and objects with references, they're (imo) incredibly buggy and unintuitive: https://3v4l.org/3RncH

ohader:
Incredible, I was not aware of that flaw in PHP! 😮 
For the time being I think, it would be enough to encapsulate those scenarios and just not process the buggy ones.

PR #8290, opened 2022-07-20
Address flaw in type reconciliation on array references

ohader:
Related: #8289

### DIFF
--- a/src/Psalm/Type/Reconciler.php
+++ b/src/Psalm/Type/Reconciler.php
@@ -124,7 +124,7 @@ class Reconciler
 
             $cloned_referenceds = [];
             foreach ($existing_references as $reference => $referenced) {
-                if (!isset($cloned_referenceds[$referenced])) {
+                if (!isset($cloned_referenceds[$referenced]) && isset($old_existing_types[$referenced])) {
                     $existing_types[$referenced] = clone $old_existing_types[$referenced];
                     $cloned_referenceds[$referenced] = true;
                 }
--- a/tests/TypeReconciliation/ReconcilerTest.php
+++ b/tests/TypeReconciliation/ReconcilerTest.php
@@ -18,6 +18,7 @@ use Psalm\Storage\Assertion\IsType;
 use Psalm\Storage\Assertion\NonEmpty;
 use Psalm\Storage\Assertion\Truthy;
 use Psalm\Tests\TestCase;
+use Psalm\Tests\TestConfig;
 use Psalm\Type;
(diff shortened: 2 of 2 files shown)


src/Psalm/Type/Reconciler.php
```diff
@@ -124,7 +124,7 @@ public static function reconcileKeyedTypes(
 
             $cloned_referenceds = [];
             foreach ($existing_references as $reference => $referenced) {
-                if (!isset($cloned_referenceds[$referenced])) {
+                if (!isset($cloned_referenceds[$referenced]) && isset($old_existing_types[$referenced])) {
```
ohader, line 127: @AndrolGenhald I'm not sure whether this is enough, or
* whether there should be a fallback union type assigned to `$existing_types[$referenced]`, or
* whether the `$referenced` value (e.g. `'$doesNotMatter[$a]'`) should to be split and processed

AndrolGenhald: I'll step through it with your test tonight or tomorrow and see what's going on. Thanks for the reproduction and possible fix!

ohader: Seems like it needs to have a type, I just ran into another error in a project having this patch applied:
```
Uncaught Error: Call to a member function equals() on null in vendor/vimeo/psalm/src/Psalm/Context.php:524
```
https://github.com/vimeo/psalm/blob/a32e63f131466a8965a47b58649bc72daeaf8b25/src/Psalm/Context.php#L521

AndrolGenhald: Oh, yeah, having nulls in there is not a good idea. I think I know what's going on though; when we do the check against `$a` it potentially changes `$a`'s type, which means we remove everything dependent on `$a` from `Context::$vars_in_scope` (somewhere, can't remember where right now) because `$doesNotMatter[$a]` doesn't have the same type if `$a` has changed. When that happens, we also need to do _something_ to any references that exist to `$doesNotMatter[$a]`, probably by changing whatever code removes `$doesNotMatter[$a]` to use the normal process by calling `Context::remove()` instead of whatever it's doing now.

Of course this means the fact that it's a reference will be lost, but like I said before, combining references with arrays and objects is just asking for trouble, and there's not much we can do about it.

AndrolGenhald:
I think I have a fix that should work, but I need to find a bit more time to finish it since it requires modifying callers of `Reconciler::reconcileKeyedTypes()`.

PR #8308, opened 2022-07-22
Fixed ini_set types for arg value

honca:
Hi guys,
in our codebase I have noticed that current version of Psalm has not updated `ini_set` types for the param `$value`. 
Since the 8.1 it is allowed to accept this set of types: `string|int|float|bool|null`. Not only `string`.

https://www.php.net/manual/en/function.ini-set

Best,
Honca


AndrolGenhald:
Thanks!

PR #8312, opened 2022-07-22
Add stub for DatePeriod

### LINKED ISSUE #5954: Iterating over \DatePeriod objects does not automatically yields \DateTimeInterface typed values
See: https://psalm.dev/r/e0a3a36944

Add `/** @var \DateTimeInterface $day */` right above the `foreach`  and the errors go away.

fluffycondor:
Fixes #5954

Actually I'd like to mark the returned Iterator as a non-empty one, it always yields at least one value (and 50 and 63 lines in the test actually should be non-nullable).
But I don't know how to do that :)


stubs/CoreImmutableClasses.phpstub
```diff
@@ -16,6 +16,27 @@ class DateTimeZone
...
+class DatePeriod implements IteratorAggregate
+{
+    const EXCLUDE_START_DATE = 1;
+    /**
+     * @param From $from
+     * @param (From is string ? 0|self::EXCLUDE_START_DATE : DateInterval) $interval_or_options
+     * @param (From is string ? never : DateTimeInterface|positive-int) $end_or_recurrences
+     * @param (From is string ? never : 0|self::EXCLUDE_START_DATE) $options
+     */
+    public function __construct($from, $interval_or_options = 0, $end_or_recurrences = 1, $options = 0) {}
```
AndrolGenhald, line 34: Let's use the actual parameter names that [PHP uses](https://github.com/php/php-src/blob/9dcb0bd1dc966a70cf6386ca9d49979dd417e0c5/ext/date/php_date_arginfo.h#L464-L469) instead of going off the documentation so that named parameters work: https://3v4l.org/3Rh8g

They like to make it difficult in these weird cases where they want to pretend it uses method overloading :roll_eyes:

AndrolGenhald:
> Actually I'd like to mark the returned Iterator as a non-empty one, it always yields at least one value (and 50 and 63 lines in the test actually should be non-nullable).

I don't think that's possible unfortunately, and the only way I can think of to do it would be to add a template, which would make all other iterators more annoying to deal with.

AndrolGenhald:
Per [this comment](https://github.com/vimeo/psalm/issues/5954#issuecomment-1048707556) it looks like we might need to have it implement `Traversable` instead and add the `IteratorAggregate` part in Php80.phpstub. It would be good to add a test for PHP 7.4 and 8.0 for that as well.

fluffycondor:
> it looks like we might need to have it implement `Traversable` instead and add the `IteratorAggregate` part in Php80.phpstub

Unfortunately, looks like I can't use a conditional `@implements` :(
It doesn't work whatever I try:
https://psalm.dev/r/f606edbe51
https://psalm.dev/r/46f7c9b9b4

stubs/CoreImmutableClasses.phpstub
fluffycondor: Done :)

AndrolGenhald:
@fluffycondor It looks like you must have branched from your fork's 4.x branch, which is quite a bit behind. Could you rebase on the current 4.x? I think that should fix the test-with-real-projects PHPUnit test.

AndrolGenhald:
Thanks!

Issue #8319, opened 2022-07-26
Broken array assignment


danog:
https://psalm.dev/r/337ce1e564, broken by 02d4f0c61467d28aba00a6d4c946d010bd3ba914

orklah:
this was fixed in master :)

PR #8324, opened 2022-07-27
Fix union assertions

### LINKED ISSUE #8322: Union assertions not detecting paradoxes
Possibly introduced by #8077: https://psalm.dev/r/49f98e503d

danog:
Fixes #8322, #8319 and a few other bugs introduced by #8077, by explicitly running the intersection logic only on IsIdentical assertions (due to the semantics of Type::intersectUnionTypes wasn't sure if we should run it on IsType assertions), and removing the custom intersection/equality logic which broke other stuff in the code.

### DIFF
--- a/src/Psalm/Type.php
+++ b/src/Psalm/Type.php
@@ -701,7 +701,7 @@ abstract class Type
                 && ($int_intersection->min_bound !== null || $int_intersection->max_bound !== null)
             ) {
                 $intersection_performed = true;
-                if ($int_intersection->min_bound === $int_intersection->max_bound) {
+                if ($int_intersection->min_bound !== null && $int_intersection->min_bound === $int_intersection->max_bound) {
                     return new TLiteralInt($int_intersection->min_bound);
                 }
                 return $int_intersection;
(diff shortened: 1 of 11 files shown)


danog:
The shepherd issue seems to be present upstream: https://psalm.dev/r/e0bee4534f

src/Psalm/Type.php
```diff
@@ -769,12 +787,19 @@ private static function intersectAtomicTypes(
     /**
      * @psalm-assert-if-true TIterable|TNamedObject|TTemplateParam|TObjectWithProperties $type
```
orklah, line 788: could you remove that assertion? Literal unions in `@psalm-assert` never worked correctly (there two incomplete PR open on that subject but none could get it to work properly yet. We probably should forbid this for the time being)

danog: I've half an idea on how to get unions working on all assertions, which would actually fit nicely with the changes of this MR which already deal with unions in assertions (even if just those generated from template types): this would require changing the Assertion API to work with Unions instead of atomics, and move the code I just wrote in this PR to the assertion reconcilers, might do in a future PRs.

danog:
Mm actually even with the additional int range handling in intersections we're still missing intersection logic for keyed arrays, I feel like I'm duplicating code already present in the assertion reconciler, once I fix the last unit case I'd still prefer to get this merged, but eventually we should be using the same code used in the assertion reconciler to perform intersections, differences&so on...

danog:
All done!

orklah:
Thanks!

PR #8326, opened 2022-07-28
Fix if propagation

### LINKED ISSUE #8317: Psalm remembers previous array key assertion even if array is replaced
Thought it was a bug in my immutable refactoring, turns out it's also present upstream: https://psalm.dev/r/b7f2600e09

danog:
Fixes #8317

### DIFF
--- a/src/Psalm/Context.php
+++ b/src/Psalm/Context.php
@@ -459,7 +459,7 @@ final class Context
         array $vars_to_update,
         array &$updated_vars
     ): void {
-        foreach ($start_context->vars_in_scope as $var_id => $_) {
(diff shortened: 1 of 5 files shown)


danog:
This reverts what is essentially a simplification of if handling logic, since the old logic allowed for patterns that aren't that uncommon in real code (an example is included in a new testcase).

orklah:
@muglug could you take a look at https://github.com/vimeo/psalm/issues/8317 please? Maybe there's an easier fix rather than reverting the whole thing?

muglug (approved):
Looks good! At some point I might try to figure out a solution that works with `else if`, and the nested of example I gave, but for now I’m happy to revert my change.

PR #8335, opened 2022-07-29
#8330 - take into account that `static` type may have been unwrapped in `hasStaticInType()`

### LINKED ISSUE #8330: [4.25.0] InvalidArgument for result of a method with a static return type
```php
<?php

declare(strict_types=1);

namespace HappyInc;

abstract class A
{
}

final class B extends A
{
    public static function create(): static
    {
        return new self();
    }
}

final class Service
{
    public function do(): void
    {
        $this->acceptA(B::create());
    }

    private function acceptA(A $_a): void
    {
    }
}
```

gives 

```
ERROR: InvalidArgument - src/A.php:23:24 - Argument 1 of HappyInc\Service::acceptA expects HappyInc\A, HappyInc\Service provided (see https://psalm.dev/004)
        $this->acceptA(B::create());
```

in 4.25.0.

By the way, for some reason https://psalm.dev/r/6081a71dec says "No issues!".

someniatko:
fixes #8330


someniatko:
I'm surprised this case haven't been covered by any tests :D

AndrolGenhald:
Aha, that explains it, thanks!

PR #8341, opened 2022-07-29
recognize `@psalm-allow-private-mutation` in PHP 8+ constructors

### LINKED ISSUE #7731: @psalm-readonly-allow-private-mutation on promoted property doesn't allow private mutation
It looks like the `@readonly` restriction is picked up, but not the `@psalm-allow-private-mutation` exception.

https://psalm.dev/r/8c55da797e

Changing the promoted property to a standard class property works as expected.

https://psalm.dev/r/da58a808ef

someniatko:
fixes #7731


AndrolGenhald:
Thanks!

PR #8344, opened 2022-07-29
ReflectionProperty::getValue $object is nullable since php 8.0

mathroc:
https://www.php.net/manual/en/reflectionproperty.getvalue.php


orklah:
Thanks!

PR #8345, opened 2022-07-29
Applied ArraySpreadInsteadOfArrayMergeRector (take 2)

weirdan:


AndrolGenhald:
Does this array spread syntax have the exact same result as `array_merge` in all cases, including sorting, with combined string and array keys? If so, do we want to add `array_merge` to our `ForbiddenFunctions` config to ensure it's used everywhere consistently? Maybe we should improve `ForbiddenFunctions` to allow configuring a custom issue description?

weirdan:
> Does this array spread syntax have the exact same result as array_merge in all cases, including sorting, with combined string and int keys? 

Not until PHP 8.1: https://wiki.php.net/rfc/array_unpacking_string_keys

orklah:
Thanks!

PR #8349, opened 2022-07-30
Fix formatCurrency return type

VincentLanglet:
Hi @orklah, I just got an error `[TypeDoesNotContainType](https://psalm.dev/056) - 6:9 - string does not contain false`.

but NumberFormatter::formatCurrency can return false
See https://www.php.net/manual/fr/numberformatter.formatcurrency.php

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -9028,7 +9028,7 @@ return [
 'NumberFormatter::__construct' => ['void', 'locale'=>'string', 'style'=>'int', 'pattern='=>'string'],
 'NumberFormatter::create' => ['NumberFormatter|false', 'locale'=>'string', 'style'=>'int', 'pattern='=>'string'],
 'NumberFormatter::format' => ['string|false', 'num'=>'', 'type='=>'int'],
-'NumberFormatter::formatCurrency' => ['string', 'num'=>'float', 'currency'=>'string'],
+'NumberFormatter::formatCurrency' => ['string|false', 'num'=>'float', 'currency'=>'string'],
 'NumberFormatter::getAttribute' => ['int|false', 'attr'=>'int'],
 'NumberFormatter::getErrorCode' => ['int'],
 'NumberFormatter::getErrorMessage' => ['string'],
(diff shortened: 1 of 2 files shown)


AndrolGenhald:
Do you know what cases this can return `false` for? If it only returns false when called with invalid arguments then we intentionally exclude it from the return type.

It looks like it happens [here](https://github.com/php/php-src/blob/625f1649639c2b9a9d76e4d42f88c264ddb8447d/ext/intl/formatter/formatter_format.c#L156-L163), but I don't know what the possibilities for `INTL_DATA_ERROR_CODE((nfo))` are or why they would happen. If you can find a case on 3v4l.org that returns `false` for valid arguments that would be excellent.

Further investigation leads [here](https://github.com/unicode-org/icu/blob/dcd19ae9bc0914f8be439a2de1a5fbc17e7b5447/icu4c/source/i18n/unum.cpp#L325-L360), which possibly sets it to `U_MEMORY_ALLOCATION_ERROR` which I'm not sure we care about.

orklah:
Usually, we would remove the false return when it was related to wrong param types. I'd say the cases are the same than function that now throws a TypeError in PHP8 instead of returning false.

So if a function still return false in PHP8, I think we should keep the falsable return

VincentLanglet:
You can see it's here for
```
'NumberFormatter::format' => ['string|false', 'num'=>'', 'type='=>'int'],
```
So it should be consistent and be here for formatCurrency.

It should allow to use a false check
```
$result = $formatter->formatCurrency(...);
if (false !== $result) {
     ...
}
```
which is valid php code.

> Further investigation leads [here](https://github.com/unicode-org/icu/blob/dcd19ae9bc0914f8be439a2de1a5fbc17e7b5447/icu4c/source/i18n/unum.cpp#L325-L360), which possibly sets it to U_MEMORY_ALLOCATION_ERROR which I'm not sure we care about.

As soon as it's possible, the false check should not be reported as an error.

> So if a function still return false in PHP8, I think we should keep the falsable return

100% agree.

Then if someone want to improve the return type, it should be done with a dynamic return type calculation

orklah:
Thanks!

PR #8350, opened 2022-07-31
Added better stubs for `DateTimeImmutable`, highlighting how the constructor is **NOT** immutable

Ocramius:
`DateTimeImmutable` is **almost** immutable: `DateTimeImmutable::__construct()` is in fact not a pure
method, since `new DateTimeImmutable('now')` produces a different value at each instantiation (by design).

This change makes sure that `DateTimeImmutable` loses its `@psalm-immutable` class-level marker,
preventing downstream misuse of the constructor inside otherwise referentially transparent code.

Note: only pure methods are stubbed here: all other methods declared by `DateTimeImmutable` (parent interface)
are NOT present here, and are either inferred from runtime reflection, or `CallMap*.php` definitions.

Methods are sorted in the order defined by reflection on PHP 8.1.8, at the time of writing this ( https://3v4l.org/3TGg8 ).

Following simplistic snippet was used to infer the current signature:

```php
<?php

$c = new \ReflectionClass(\DateTimeImmutable::class);

$methods = array_map(function ($m) {
    return $m->getName()
        . '(' . implode(',', array_map(function ($p) {
            return $p->getType()
                . ' $' . $p->getName()
                . ($p->isOptional() ? ' = ' . var_export($p->getDefaultValue(), true) : '');
        }, $m->getParameters())) . ')' . ($m->getReturnType() ? (': ' . $m->getReturnType()) : '');
}, $c->getMethods());

$properties = array_map(function ($m) {
    return $m->getName();
}, $c->getProperties());

var_dump($methods, $properties);
```

### DIFF
--- a/stubs/CoreImmutableClasses.phpstub
+++ b/stubs/CoreImmutableClasses.phpstub
@@ -91,7 +91,7 @@ class DateTimeImmutable implements DateTimeInterface
 
     /**
      * @psalm-mutation-free
-     * @return static|false
+     * @return static
      */
     public function setTimestamp(int $unixtimestamp) {}
 
--- a/stubs/CoreImmutableClasses.phpstub
+++ b/stubs/CoreImmutableClasses.phpstub
@@ -1,11 +1,106 @@
 <?php
 
-/**
- * @psalm-immutable
- */
 class DateTimeImmutable implements DateTimeInterface
 {
     public function __construct(string $datetime = "now", DateTimeZone $timezone = null) {}
+
+    /**
+     * @psalm-pure
+     * @return static|false
+     */
+    public static function createFromFormat(string $format, string $datetime, ?DateTimeZone $timezone = null) {}
+
+    /**
+     * @psalm-pure
+     * @param string $format
+     * @return string
+     */
+    public function format($format) {}
+
+    /**
+     * @psalm-pure
+     * @return DateTimeZone
+     */
+    public function getTimezone() {}
+
+    /**
+     * @psalm-pure
+     * @return int
+     */
+    public function getOffset() {}
+
+    /**
+     * @psalm-pure
+     * @return int
+     */
+    public function getTimestamp() {}
+
+    /**
+     * @psalm-pure
+     * @param bool $absolute
+     * @return DateInterval
+     */
+    public function diff(DateTimeInterface $targetObject, $absolute = false) {}
+
+    /**
+     * @psalm-pure
+     * @return static
+     */
+    public function modify(string $modifier) {}
+
+    /**
+     * @psalm-pure
+     * @return static
+     */
+    public function add(DateInterval $interval) {}
+
+    /**
+     * @psalm-pure
+     * @return static
+     */
+    public function sub(DateInterval $interval) {}
+
+    /**
+     * @psalm-pure
+     * @return static|false
+     */
+    public function setTimezone(DateTimeZone $timezone) {}
+
+    /**
+     * @psalm-pure
+     * @return static|false
+     */
+    public function setTime(int $hour, int $minute, int $second = 0, int $microsecond = 0) {}
+
+    /**
+     * @psalm-pure
+     * @return static|false
+     */
+    public function setDate(int $year, int $month, int $day) {}
+
+    /**
+     * @psalm-pure
+     * @return static|false
+     */
+    public function setISODate(int $year, int $week, int $dayOfWeek = 1) {}
+
+    /**
+     * @psalm-pure
+     * @return static|false
+     */
+    public function setTimestamp(int $unixtimestamp) {}
+
+    /**
+     * @psalm-pure
+     * @return static
+     */
+    public static function createFromMutable(DateTime $object) {}
+
+    /**
+     * @psalm-pure
+     * @return self
+     */
+    public static function createFromInterface(DateTimeInterface $object) {}
 }
 
 /**
--- a/stubs/CoreImmutableClasses.phpstub
+++ b/stubs/CoreImmutableClasses.phpstub
@@ -5,99 +5,99 @@ class DateTimeImmutable implements DateTimeInterface
     public function __construct(string $datetime = "now", DateTimeZone $timezone = null) {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @return static|false
      */
     public static function createFromFormat(string $format, string $datetime, ?DateTimeZone $timezone = null) {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @param string $format
      * @return string
      */
     public function format($format) {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @return DateTimeZone
      */
     public function getTimezone() {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @return int
      */
     public function getOffset() {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @return int
      */
     public function getTimestamp() {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @param bool $absolute
      * @return DateInterval
      */
     public function diff(DateTimeInterface $targetObject, $absolute = false) {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @return static
      */
     public function modify(string $modifier) {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @return static
      */
     public function add(DateInterval $interval) {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @return static
      */
     public function sub(DateInterval $interval) {}
 
     /**
-     * @psalm-pure
+     * @psalm-mutation-free
      * @return static|false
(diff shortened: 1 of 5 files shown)


Ocramius:
Note: this sparked from initial conversations on slack: I don't really know if I should change the existing `CallMap` files, after adjusting this stub.

/cc @guidobonuzzi, since you may be interested.

tests/MethodCallTest.php
```diff
@@ -264,15 +264,14 @@ public static function main(): void {
...
-                    /** @psalm-immutable */
                     final class MyDate extends DateTimeImmutable {}
 
                     $today = new MyDate();
                     $yesterday = $today->sub(new DateInterval("P1D"));
 
                     $b = (new DateTimeImmutable())->modify("+3 hours");',
                 'assertions' => [
-                    '$yesterday' => 'MyDate|false',
+                    '$yesterday' => 'MyDate',
```
Ocramius, line 274: Unsure about this change: is `MyDate|false` actually expected? From a type perspective, PHP 8 doesn't seem to have `false` as a possible return value anymore :thinking:

stubs/CoreImmutableClasses.phpstub
```diff
@@ -1,11 +1,106 @@
...
 
-/**
- * @psalm-immutable
- */
 class DateTimeImmutable implements DateTimeInterface
 {
     public function __construct(string $datetime = "now", DateTimeZone $timezone = null) {}
+
+    /**
+     * @psalm-pure
```
AndrolGenhald, line 8: I think these should all be `@psalm-mutation-free` instead of `@psalm-pure`. Compare: https://psalm.dev/r/39eb3a7c35 https://psalm.dev/r/7a678770a4.

tests/MethodCallTest.php
AndrolGenhald: `getTimestamp` was fixed by #7855, I was intending to go fix the others but I never got around to it. I think we might need to fix that in the CallMap for functions not defined in the stub, and for the stubbed functions we'll probably have to use the `|false` return and copy/paste it to Php80.phpstub without the `|false` return.

stubs/CoreImmutableClasses.phpstub
Ocramius: from a caller perspective, these should be usable in a pure context, and `$this` should still be immutable-ish?

AndrolGenhald: Mutation free methods are usable from a pure context: https://psalm.dev/r/1ac851b862

I'm not really sure if mutation-free vs pure is a useful distinction for us to have, but that's the way it is right now :shrug:

Ocramius: As long as it works, I'm fine with going either way.

Should I adjust it?

AndrolGenhald: I'd say so, I'm not sure if we have any logic that cares about the difference between mutation-free and pure, but better safe than sorry.

Ocramius: 68978b9e19c70036e6e5077a6234af513afdc77f

tests/MethodCallTest.php
Ocramius: The wording above confused me more :D 

If I understand it correctly:

 * the stub represents current (latest) PHP state
 * the `CallMap` includes PHP-version-specific overrides

If it's the opposite (stubs represent "oldest" state, CallMap represents "newest" state), then it's super-confusing overall :D

AndrolGenhald: The latter, CallMap represents newest state, CallMap_historical represents oldest state, and there are diffs for each PHP version to get from CallMap to CallMap_historical. For stubs, stubs represent oldest state except for PHPxx.phpstub, which is applied after the other stubs to add changes for newer versions. It's confusing.

Ocramius: > stubs represent oldest state

Argh @_@

AndrolGenhald: I think the reason it was done that way is because we don't actually have a way to "unapply" a stub, but for the callmap it's easy. It's unfortunate that they're opposite though...

orklah: stubs override Callmap though, so if you have a function declared in stubs, it doesn't really matter what you have in the Callmap (or Callmap_historical)

AndrolGenhald: Oh, yeah, sorry if that wasn't clear.

Ocramius: I'm a bit lost on what to do - anyone able to pick this up perhaps? :| 

If the stub takes precedence, is this patch acceptable?

orklah: Sorry for the confusion, here's a recap:
- stubs override callmap so your stub here is good
- you added new methods in the existing stub, so it overrided existing definition in callmap (that's why you saw types change in tests)
- there is a specific stub file for each minor version of php (it is mainly used for new classes and stuff in new version)
- this means your stub here applies to any php version, which is not great because types changed between versions
- however, we do have meta variables made to be used in cases like these: https://github.com/vimeo/psalm/blob/e0acf22e402d45d1aebb75f4365ebf701fabce2f/stubs/CoreGenericFunctions.phpstub#L1096

So, could you please adapt your stub to use conditional returns in order to reflect the correct return given a php version and it should be good :)

AndrolGenhald: Oh, I didn't even know about using `PHP_MAJOR_VERSION` in a conditional return, that's neat. I assumed it would have to be overridden in PHP80.phpstub.

orklah: Yeah, we also have this weird one too: https://github.com/vimeo/psalm/blob/b4fdc3e326fea865de49827f723a2703bb18617a/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockScanner.php#L504

But useful in some edge cases :)

AndrolGenhald: With your example though, won't that break when PHP 9.0 is released? Wouldn't you want to invert it and do `(PHP_MAJOR_VERSION is 7 ...)`? (assuming we don't care about 5.x at all anymore).

orklah: Some users, including some maintainers won't be happy if you break 5.x analysis :p. But yeah, we're bound to have some work when PHP 9 will come anyway. If it can be handled now it's better, otherwise we'll do it later

orklah: :o we can use this too: https://github.com/vimeo/psalm/blob/b49682a4853803f72996fbd6664b5e69a75472ae/tests/Template/ConditionalReturnTypeTest.php#L866

I didn't even know it worked with int ranges! That's better than the previous solution even!

AndrolGenhald: Oh, cool, that's a much better solution! I was thinking that wasn't possible because you have to use `is` and you can't say `PHP_MAJOR_VERSION < 8`.

Ocramius: Ok, I've documented all types:

 * looked at core PHP implementation
 * looked at core PHP stubs
 * implemented conditional types based on major PHP version (luckily, only one case of that)

Still one problem left, but this thread is resolved :-)

stubs/CoreImmutableClasses.phpstub
```diff
@@ -1,11 +1,111 @@
...
+     * @psalm-mutation-free
+     * @return static
+     */
+    public static function createFromMutable(DateTime $object) {}
+
+    /**
+     * @psalm-mutation-free
+     * @return self
+     */
+    public static function createFromInterface(DateTimeInterface $object) {}
```
Ocramius, line 108: This method is new, introduced in PHP 8.0.

Unsure how to declare this conditionally :D

tests/MethodCallTest.php
Ocramius: Note: each added commit tries to reference the sources of my findings.

Reads: avoid squashing, please :P

Ocramius:
Interesting finding: `DateTimeImmutable#format()` **always** produces a `string`, no matter what you pass to it.

I took the occasion to refine that `non-empty-string` input will lead to `non-empty-string` output :-)

stubs/CoreImmutableClasses.phpstub
Ocramius: I'll remove it from the patch: not so interested in declaring purity of this one anyway, since it may potentially cause mutation while reading a mutable `DateTimeInterface` (unlikely, but :shrug:).

That should bring the patch to completion.

Ocramius: 13828771c7c544aa61a3db1ef49b7aae9e0996fd

Ocramius:
@AndrolGenhald @orklah this is now ready, IMO: thanks for all the support! :+1:

orklah:
Could you please keep the content of callmaps please? I'm not entirely sure we can drop it, and even if we could, there's no harm to have a comprehensive list if needed some day

Ocramius:
I think it's duplicated at this point: shouldn't be kept, if what you said about priority is true.

What value does the added info provide, considering that each removal contains documentation/references?

orklah:
Fine then :p Thanks!

Ocramius:
Thanks @orklah!

PR #8360, opened 2022-08-03
Add BitwiseNot and BooleanNot operators to SimpleTypeInferer

### LINKED ISSUE #8307: Class constants where value starts with an operand are not recognized
If the value for a class constant starts with an operand, like `!` or `~`, Psalm doesn't register it, and throws an `UndefinedConstant` error.

https://psalm.dev/r/b100090d84

bitwise-operators:
`BitwiseNot` and `BooleanNot` operators were missing from the `SimpleTypeInferer`. This PR adds them.

This should fix #8307.

Thanks to @AndrolGenhald for pointing me in the right direction.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
@@ -249,7 +249,9 @@ class SimpleTypeInferer
                 $fq_classlike_name
             );
 
-            if ($stmt_expr_type === null || $stmt_expr_type->isAlwaysFalsy()) {
+            if ($stmt_expr_type === null) {
+                return null;
+            } elseif ($stmt_expr_type->isAlwaysFalsy()) {
                 return Type::getTrue();
             } elseif ($stmt_expr_type->isAlwaysTruthy()) {
                 return Type::getFalse();
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
@@ -203,6 +203,61 @@ class SimpleTypeInferer
             }
         }
 
+        if ($stmt instanceof PhpParser\Node\Expr\BitwiseNot) {
+            $stmt_expr_type = self::infer(
+                $codebase,
+                $nodes,
+                $stmt->expr,
+                $aliases,
+                $file_source,
+                $existing_class_constants,
+                $fq_classlike_name
+            );
+
+            if ($stmt_expr_type === null) {
+                return null;
+            }
+
+            $invalidTypes = clone $stmt_expr_type;
+            $invalidTypes->removeType('string');
+            $invalidTypes->removeType('int');
+            $invalidTypes->removeType('float');
+
+            if (!$invalidTypes->isUnionEmpty()) {
+                return null;
+            }
+
+            $types = [];
+            if ($stmt_expr_type->hasString()) {
+                $types[] = Type::getString();
+            }
+            if ($stmt_expr_type->hasInt() || $stmt_expr_type->hasFloat()) {
+                $types[] = Type::getInt();
+            }
+
+            return $types ? Type::combineUnionTypeArray($types, null) : null;
+        }
+
+        if ($stmt instanceof PhpParser\Node\Expr\BooleanNot) {
+            $stmt_expr_type = self::infer(
+                $codebase,
+                $nodes,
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -203,6 +203,61 @@ public static function infer(
...
+            );
+
+            if ($stmt_expr_type === null) {
+                return null;
+            }
+
+            $invalidTypes = clone $stmt_expr_type;
+            $invalidTypes->removeType('string');
+            $invalidTypes->removeType('int');
+            $invalidTypes->removeType('float');
```
AndrolGenhald, line 224: Is float actually a possible return type for the `~` operator?

I thought I actually added an issue about this, but apparently I only [considered it](https://github.com/vimeo/psalm/pull/7992#issuecomment-1133272149) then never did it (went ahead and added #8362). No need to worry about fixing that here though, just need to remove the `float` result type.

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -203,6 +203,61 @@ public static function infer(
...
+
+            if (!$invalidTypes->isUnionEmpty()) {
+                return null;
+            }
```
AndrolGenhald, line 228: I assume this causes stuff like `~[]` to infer as `mixed`? @orklah would we prefer to have it infer `int|string` since those are the only possible return values for `~`, even if it's an invalid operand?

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
AndrolGenhald: Nevermind, I see what's going on, this is just checking if the operand is valid.

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
orklah: I don't really care one way or another but I guess mixed has the advantage of returning at least a warning at level 1 that may help with debugging

orklah:
Wow, I can't believe how your username is fitting with this PR! I even went to check on your profile whether every previous issue you opened here was also related to bitwise operators or not 😄

src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -203,6 +203,61 @@ public static function infer(
...
+                $codebase,
+                $nodes,
+                $stmt->expr,
+                $aliases,
+                $file_source,
+                $existing_class_constants,
+                $fq_classlike_name
+            );
+
+            if ($stmt_expr_type === null || $stmt_expr_type->isAlwaysFalsy()) {
```
orklah, line 252: why does the `$stmt_expr_type === null` case returns a True result? Shouldn't it be Bool?

AndrolGenhald:
Could you add a couple of tests for these cases (probably in ConstantTest.php)? LGTM after that's done.

bitwise-operators:
> Could you add a couple of tests for these cases (probably in ConstantTest.php)? LGTM after that's done.

Done. I also added tests for two other missing bitwise operators.

tests/ConstantTest.php
```diff
@@ -998,6 +998,60 @@ class X {
...
+                'code' => '<?php
+                    class X {
+                        public const A = 1;
+                        public const B = 2;
+                        public const C = self::A & self::B;
+                    }
+
+                    $c = X::C;',
+                'assertions' => [
+                    '$c' => 'int',
```
AndrolGenhald, line 1011: Just one more minor nitpick, could you try using eg `'$c===' => '0'` for these assertions so we make sure it fully resolves the actual value? If it doesn't work that's fine and we can go ahead and merge it, but if it does work it seems better to make sure it continues working that way in the future.

AndrolGenhald: Actually looking at this again, I think this will require additional logic to the sections you added to check for literals and evaluate them, which we can leave for a future improvement.

orklah:
Thanks!

tests/ConstantTest.php
AndrolGenhald: Well, now that it's released it turns out this [does work](https://psalm.dev/r/b55d6ea10b). The other tests that use SimpleTypeInferer are probably enough to cover it though.

PR #8364, opened 2022-08-03
add ", but" for InvalidArgument error message where a type is provided

kkmuffme:
Added a "but" after the "," in InvalidArgument errors, so one can easily find where the provided type starts. If the function has a param with array{} with many parameters and nested array{}, it is often very hard to find where the expected type ends and the provided type starts.

This is a very simple example, in reality I often had cases with 25+ keys.
https://psalm.dev/r/2e6a506272


orklah:
Seems good! Can you fix remaining CI failures and it'll be good to go

orklah:
Thanks!

PR #8366, opened 2022-08-03
Fix invalid casts int and float

kkmuffme:
e.g. https://psalm.dev/r/0d284c6f48 gives `Warning: Object of class stdClass could not be converted to int`

Since by default there cannot be any "PossiblyInvalidCasts", as we always get 0/0.0 in error case, added mostly unwanted typecasts that are a leading cause of off-by-one errors as PossiblyInvalidCasts

Fix: https://github.com/vimeo/psalm/issues/6966 (the remaining half + cherry-pick the other half from master to 4.x)

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
@@ -23,6 +23,7 @@ use Psalm\Type;
 use Psalm\Type\Atomic\Scalar;
 use Psalm\Type\Atomic\TArray;
 use Psalm\Type\Atomic\TBool;
+use Psalm\Type\Atomic\TClosedResource;
 use Psalm\Type\Atomic\TFalse;
 use Psalm\Type\Atomic\TFloat;
 use Psalm\Type\Atomic\TInt;
@@ -33,6 +34,9 @@ use Psalm\Type\Atomic\TLiteralInt;
 use Psalm\Type\Atomic\TLiteralString;
 use Psalm\Type\Atomic\TMixed;
 use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Atomic\TNonEmptyArray;
+use Psalm\Type\Atomic\TNonEmptyList;
+use Psalm\Type\Atomic\TNonEmptyString;
 use Psalm\Type\Atomic\TNonspecificLiteralInt;
 use Psalm\Type\Atomic\TNonspecificLiteralString;
 use Psalm\Type\Atomic\TNull;
@@ -42,6 +46,7 @@ use Psalm\Type\Atomic\TObjectWithProperties;
 use Psalm\Type\Atomic\TResource;
 use Psalm\Type\Atomic\TString;
 use Psalm\Type\Atomic\TTemplateParam;
+use Psalm\Type\Atomic\TTrue;
 use Psalm\Type\Union;
 
 use function array_merge;
@@ -352,8 +357,28 @@ class CastAnalyzer
                 continue;
             }
 
+            if ($atomic_type instanceof TTrue
+            ) {
+                $valid_strings[] = new TLiteralString('1');
+                continue;
+            }
+
+            if ($atomic_type instanceof TBool
+            ) {
+                $valid_strings[] = new TLiteralString('1');
+                $valid_strings[] = new TLiteralString('');
+                continue;
+            }
+
+            if ($atomic_type instanceof TClosedResource
+               || $atomic_type instanceof TResource
+            ) {
+                $castable_types[] = new TNonEmptyString();
+
+                continue;
+            }
+
             if ($atomic_type instanceof TMixed
-                || $atomic_type instanceof TResource
                 || $atomic_type instanceof Scalar
             ) {
                 $castable_types[] = new TString();
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
@@ -17,6 +17,7 @@ use Psalm\Issue\InvalidCast;
 use Psalm\Issue\PossiblyInvalidCast;
 use Psalm\Issue\RedundantCast;
 use Psalm\Issue\RedundantCastGivenDocblockType;
+use Psalm\Issue\RiskyCast;
 use Psalm\Issue\UnrecognizedExpression;
 use Psalm\IssueBuffer;
 use Psalm\Type;
@@ -298,7 +299,7 @@ class CastAnalyzer
     ): Union {
         $codebase = $statements_analyzer->getCodebase();
 
-        $possibly_unwanted_cast = [];
+        $risky_cast = [];
         $invalid_casts = [];
         $valid_ints = [];
         $castable_types = [];
@@ -445,7 +446,7 @@ class CastAnalyzer
             if ($atomic_type instanceof TNonEmptyArray
                 || $atomic_type instanceof TNonEmptyList
             ) {
-                $possibly_unwanted_cast[] = $atomic_type->getId();
+                $risky_cast[] = $atomic_type->getId();
 
                 $valid_ints[] = new TLiteralInt(1);
 
@@ -458,7 +459,7 @@ class CastAnalyzer
             ) {
                 // if type is not specific, it can be both 0 or 1, depending on whether the array has data or not
                 // welcome to off-by-one hell if that happens :-)
-                $possibly_unwanted_cast[] = $atomic_type->getId();
+                $risky_cast[] = $atomic_type->getId();
 
                 $valid_ints[] = new TLiteralInt(0);
                 $valid_ints[] = new TLiteralInt(1);
@@ -486,10 +487,10 @@ class CastAnalyzer
                 ),
                 $statements_analyzer->getSuppressedIssues()
             );
-        } elseif (!empty($possibly_unwanted_cast)) {
+        } elseif ($risky_cast) {
             IssueBuffer::maybeAdd(
-                new PossiblyInvalidCast(
-                    'Casting ' . $possibly_unwanted_cast[0] . ' to int has possibly unintended value of 1',
+                new RiskyCast(
+                    'Casting ' . $risky_cast[0] . ' to int has possibly unintended value of 0/1',
                     new CodeLocation($statements_analyzer->getSource(), $stmt)
                 ),
                 $statements_analyzer->getSuppressedIssues()
@@ -525,7 +526,7 @@ class CastAnalyzer
     ): Union {
         $codebase = $statements_analyzer->getCodebase();
 
-        $possibly_unwanted_cast = [];
+        $risky_cast = [];
         $invalid_casts = [];
         $valid_floats = [];
         $castable_types = [];
@@ -671,7 +672,7 @@ class CastAnalyzer
             if ($atomic_type instanceof TNonEmptyArray
                 || $atomic_type instanceof TNonEmptyList
             ) {
-                $possibly_unwanted_cast[] = $atomic_type->getId();
+                $risky_cast[] = $atomic_type->getId();
 
                 $valid_floats[] = new TLiteralFloat(1.0);
 
@@ -684,7 +685,7 @@ class CastAnalyzer
             ) {
                 // if type is not specific, it can be both 0 or 1, depending on whether the array has data or not
                 // welcome to off-by-one hell if that happens :-)
-                $possibly_unwanted_cast[] = $atomic_type->getId();
+                $risky_cast[] = $atomic_type->getId();
 
                 $valid_floats[] = new TLiteralFloat(0.0);
                 $valid_floats[] = new TLiteralFloat(1.0);
@@ -712,10 +713,10 @@ class CastAnalyzer
                 ),
                 $statements_analyzer->getSuppressedIssues()
             );
-        } elseif (!empty($possibly_unwanted_cast)) {
+        } elseif ($risky_cast) {
             IssueBuffer::maybeAdd(
-                new PossiblyInvalidCast(
-                    'Casting ' . $possibly_unwanted_cast[0] . ' to float has possibly unintended value of 1.0',
+                new RiskyCast(
+                    'Casting ' . $risky_cast[0] . ' to float has possibly unintended value of 0.0/1.0',
                     new CodeLocation($statements_analyzer->getSource(), $stmt)
                 ),
                 $statements_analyzer->getSuppressedIssues()
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
@@ -54,9 +54,18 @@ use function array_merge;
 use function array_pop;
 use function array_values;
 use function get_class;
+use function strtolower;
 
 class CastAnalyzer
 {
+    /** @var string[] */
+    private const PSEUDO_CASTABLE_CLASSES = [
+        'SimpleXMLElement',
+        'DOMNode',
+        'GMP',
+        'Decimal\Decimal',
+    ];
+
     public static function analyze(
         StatementsAnalyzer $statements_analyzer,
         PhpParser\Node\Expr\Cast $stmt,
@@ -78,7 +87,6 @@ class CastAnalyzer
 
                 $type = self::castIntAttempt(
                     $statements_analyzer,
-                    $context,
                     $maybe_type,
                     $stmt->expr,
                     true
@@ -106,7 +114,6 @@ class CastAnalyzer
 
                 $type = self::castFloatAttempt(
                     $statements_analyzer,
-                    $context,
                     $maybe_type,
                     $stmt->expr,
                     true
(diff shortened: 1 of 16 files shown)


kkmuffme:
Feedback appreciated :-)

Could someone please help with the failing taint tests?

tests/ArrayFunctionCallTest.php
```diff
@@ -2163,7 +2163,7 @@ function($val) { return (int)$val; },
                     );
                 ',
                 'assertions' => [
-                    '$line===' => 'array{0: int}<array-key, int>',
+                    '$line===' => 'array{0: 0}<array-key, 0>',
```
AndrolGenhald, line 2166: This seems wrong. What if `$line` was `["123", "456"]`?

AndrolGenhald:
There's a lot going on here that I don't quite understand, could you add a description of the issue you're trying to solve?

tests/ArrayFunctionCallTest.php
kkmuffme: Then the inferred type will be correctly set to `array{0: 123, 1: 456}`. But the given array is `["a", "b"]` which, if you typecast will be 0 for both values.

kkmuffme:
>There's a lot going on here that I don't quite understand, could you add a description of the issue you're trying to solve?

Summary: this PR adds correct support for typecasting to int/float. 

- typecasting of objects/resources to int/float gives a PHP warning, but Psalm doesn't say anything about that. This PR adds InvalidCast error for this, https://psalm.dev/r/0d284c6f48

- typecasting arrays to int/float is always 0 or 1, not generic int https://psalm.dev/r/9e15ff2d1f

- due to this 0/1, typecasting of arrays often leads to off-by-one errors  in code, that are extremely hard to track down in some cases. This is now added as "PossiblyInvalidCast" for int/float casts of arrays.

- typecasting (string) true is literal string '1', not generic string.

>If taint tests (or unused variable tests) are failing, you probably messed up $parent_nodes somewhere.

Yes, thanks. That was it, the first commit that just unified the code messed it up. Reverted it now (will make it look nice later).
Now there are some other errors, e.g. for (array) casts, even though they weren't changed.
Some input on the failing tests would be great.

tests/ArrayFunctionCallTest.php
AndrolGenhald: But shouldn't `$line` be `array{0: string}<string>` when `array_map` is called? At the point the callback is evaluated it should only know that `$val` is a `string` shouldn't it?

AndrolGenhald:
> Now there are some other errors, e.g. for (array) casts, even though they weren't changed.
> Some input on the failing tests would be great.

I think the array one that's failing is the one I mentioned previously, looks like the rest of them are some taint tests that are still failing.

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -268,7 +254,7 @@ public static function analyze(
                 $type = Type::getArray();
             }
 
-            if ($statements_analyzer->data_flow_graph) {
+            if ($statements_analyzer->data_flow_graph instanceof VariableUseGraph) {
```
orklah, line 257: does reverting this change helps with taint errors?

AndrolGenhald:
> This is now added as "PossiblyInvalidCast" for int/float casts of arrays.

@orklah What do you think about adding `DangerousCast` for this instead? I think `PossiblyInvalidCast` should be used when casting a variable whose type is a union where the cast might be valid or might be invalid (although I think there are probably some issues where we use a `PossiblyInvalid...` issue when it should just be `Invalid...` and vice versa).

orklah:
Yeah, that makes sense, we should try to keep this rule consistent

src/Psalm/Issue/RiskyCast.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Psalm\Issue;
+
+class RiskyCast extends CodeIssue
+{
+    public const ERROR_LEVEL = 3;
+    public const SHORTCODE = 303;
```
AndrolGenhald, line 8: @orklah We need to think about conflicting short codes with Psalm 5...

src/Psalm/Internal/Cli/Refactor.php
```diff
@@ -284,7 +284,7 @@ function () use ($current_dir, $options, $vendor_dir): ?\Composer\Autoload\Class
             chdir($current_dir);
         }
 
-        $threads = isset($options['threads'])
+        $threads = isset($options['threads']) && is_numeric($options['threads'])
```
AndrolGenhald, line 287: Should we add some sort of notice if it's non-numeric?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -352,7 +778,14 @@ public static function castStringAttempt(
...
             }
 
+            if ($atomic_type instanceof TBool
+            ) {
+                $valid_strings[] = new TLiteralString('1');
+                continue;
+            }
+
             if ($atomic_type instanceof TMixed
+                || $atomic_type instanceof TClosedResource
```
AndrolGenhald, line 788: Does PHP ever give a warning for this? Do we need to add an issue in some cases?

AndrolGenhald: According to [this](https://www.php.net/manual/en/language.types.string.php#language.types.string.casting):

> Resources are always converted to strings with the structure "Resource id #​1", where 1 is the resource number assigned to the resource by PHP at runtime. While the exact structure of this string should not be relied on and is subject to change, it will always be unique for a given resource within the lifetime of a script being executed (ie a Web request or CLI process) and won't be reused.

so this could probably be non-empty-string.

src/Psalm/Internal/Cli/Refactor.php
AndrolGenhald: > However psalm's behavior has not changed

Actually it has a bit: https://3v4l.org/Xj0BS, don't you just love the way PHP handles casts?

I think this is probably fine though.

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
AndrolGenhald: > And resource/closed-resource TIntRange(1, null) for typecasts to (int) too

I see [this](https://www.php.net/manual/en/language.types.integer.php#language.types.integer.casting) which says:

> If a resource is converted to an int, then the result will be the unique resource number assigned to the resource by PHP at runtime.

I don't see anything guaranteeing it's positive, and I found [this](https://stackoverflow.com/a/6946116) saying non-positive numbers aren't expected, but they are possible. @orklah what do you think about that? Maybe `int` is safer here since I'm not sure if a non-positive value is considered a bug by PHP?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -300,6 +289,444 @@ public static function analyze(
...
+
+            if ($atomic_type instanceof TNull || $atomic_type instanceof TFalse) {
+                $valid_ints[] = new TLiteralInt(0);
+                continue;
+            }
+
+            if ($atomic_type instanceof TBool) {
+                $valid_ints[] = new TLiteralInt(1);
+                continue;
+            }
```
AndrolGenhald, line 355: Either you meant `TTrue` instead of `TBool`, or this should be `int<0, 1>`. Probably handle `TTrue` first and then `TBool`?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -300,6 +289,444 @@ public static function analyze(
...
+
+            // could be invalid, but allow it, as it is allowed for TString below too
+            if ($atomic_type instanceof TMixed
```
AndrolGenhald, line 358: @orklah What do you think about [this behavior](https://psalm.dev/r/170b48f07d)? Leave it as is to avoid a bunch of `mixed` issues?

It annoys me a bit because I sometimes have to use `mixed` to represent an unknown value that's able to be anything, but I still want to be warned if I do anything invalid with it. Unfortunately for me Psalm likes to ignore a lot of problems with `mixed` types because otherwise untyped codebases would have a ridiculous number of errors.

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -300,6 +289,444 @@ public static function analyze(
...
+                            continue 2;
+                        }
+                    }
+
+                    if ($intersection_type instanceof TObjectWithProperties
+                        && isset($intersection_type->methods['__toString'])
+                    ) {
+                        $castable_types[] = new TInt();
+
+                        continue 2;
```
AndrolGenhald, line 432: Same as above.

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -300,6 +289,444 @@ public static function analyze(
...
+            }
+
+            if ($atomic_type instanceof TTemplateParam) {
+                $atomic_types = array_merge($atomic_types, $atomic_type->as->getAtomicTypes());
+
+                continue;
+            }
+
+            // always 1 for "error" cases
+            $valid_ints[] = new TLiteralInt(1);
```
AndrolGenhald, line 468: Why `1` and not `int`?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -300,6 +289,444 @@ public static function analyze(
...
+        } elseif (!empty($risky_cast)) {
+            IssueBuffer::maybeAdd(
+                new RiskyCast(
+                    'Casting ' . $risky_cast[0] . ' to int has possibly unintended value of 0/1',
+                    new CodeLocation($statements_analyzer->getSource(), $stmt)
+                ),
+                $statements_analyzer->getSuppressedIssues()
+            );
+        } elseif ($explicit_cast && !$castable_types) {
+            // todo: emit error here
```
AndrolGenhald, line 490: Should this be `InvalidCast`?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -300,6 +289,444 @@ public static function analyze(
...
+        Union $stmt_type,
+        PhpParser\Node\Expr $stmt,
+        bool $explicit_cast = false
+    ): Union {
+        $codebase = $statements_analyzer->getCodebase();
+
+        $risky_cast = [];
+        $invalid_casts = [];
+        $valid_ints = [];
+        $castable_types = [];
```
AndrolGenhald, line 304: Not sure why there is both `$valid_ints` and `$castable_types`, is it just organizational because `$valid_ints` only holds literals? Mind adding a short comment?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
orklah: Well, in my mind at least, whenever I see an explicit cast, I tend to assume the developper knew what he was doing. I don't know if this hunch is correct and verified in the wild, but I would be personnally annoyed if Psalm were to report me such cases (If the projects I analyze were at level 1 at least!).

AndrolGenhald: I never like to assume the developer knows what they're doing! :stuck_out_tongue_closed_eyes:

But yeah, it fits with all the other special exceptions we have for `mixed`. Maybe I'll have to start using `object|resource|array|string|int|float|bool|null` instead, despite the verbosity.

orklah: I was going to say that Psalm would revert that to mixed [like we do with scalar](https://github.com/vimeo/psalm/blob/b4fdc3e326fea865de49827f723a2703bb18617a/src/Psalm/Internal/Type/TypeCombiner.php#L334) but it doesn't seem so!

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
AndrolGenhald: > No, this is correct. TFalse is already handled in the condition above (TLiteralInt(0)) therefore bool is only TTrue here.

Actually no. The type `false` is represented by `TFalse`, `true` is represented by `TTrue`, but bool isn't `Union([new TFalse(), new TTrue()])`, it's `TBool`, so you can absolutely have a `TBool` that's neither `TTrue` nor `TFalse`.

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
AndrolGenhald: None that I can think of, but I don't have a lot of time right now. @orklah might want to do a quick double check before merging.

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
AndrolGenhald: > But this can never be true, therefore I think we can remove this condition?

Probably, but if you like that can wait for #8486. It's probably easier to refactor these separate casts if they're currently as similar as possible.

AndrolGenhald:
> @AndrolGenhald rebased and issues fixed, except "ComplexMethod" in these 2 functions as it doesn't make sense to rework this now, when this has to be modified again in #8486 and [#8366 (review)](https://github.com/vimeo/psalm/pull/8366#pullrequestreview-1107780547)
> 
> I think fixing these 2 issues now, will just make it harder for the followup work.

Agreed, could you `@psalm-suppress` them though so they don't break other PRs once merged?

Also, it looks like there's now a regression in the bitwiseAssignment test.

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -300,6 +291,457 @@ public static function analyze(
...
+
+        if ($invalid_casts) {
+            IssueBuffer::maybeAdd(
+                new InvalidCast(
+                    $invalid_casts[0] . ' cannot be cast to int',
+                    new CodeLocation($statements_analyzer->getSource(), $stmt)
+                ),
+                $statements_analyzer->getSuppressedIssues()
+            );
+        } elseif (!empty($risky_cast)) {
```
AndrolGenhald, line 489: @orklah didn't we talk about banning `empty` in the Psalm config at some point? It does seem like a rather pointless function since it basically just casts everything to `bool`. Personally I prefer an explicit `count($arr) !== 0` or `$arr !== []`, but we should probably have some sort of consistency.

orklah: We still have a few to remove, but yeah, best not to add new ones

AndrolGenhald: I don't see it, did you forget to push that part?

src/Psalm/Internal/Analyzer/Statements/Expression/CastAnalyzer.php
```diff
@@ -300,6 +291,457 @@ public static function analyze(
...
+                if ($atomic_type->extra_types) {
+                    $intersection_types = array_merge($intersection_types, $atomic_type->extra_types);
+                }
+
+                foreach ($intersection_types as $intersection_type) {
+                    if ($intersection_type instanceof TNamedObject) {
+                        $intersection_method_id = new MethodIdentifier(
+                            $intersection_type->value,
+                            '__tostring'
+                        );
```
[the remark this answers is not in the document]
AndrolGenhald: Of course `DOMDocument` causes issues... Are you sure it uses `__toString` though? I sort of figured `DOMDocument` had its own weird cast handling? Maybe we just have to add a special case where it's a `TNamedObject` with `strtolower($...->value) === "domdocument"` (and maybe `"domnode"`, or maybe we should check if it extends `"domnode"`)?

@orklah do you have any idea how `DOM` casting works? I've never looked into it.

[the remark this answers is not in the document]
AndrolGenhald: I think the best way forward at this point is to just pretend all casts from objects extending `SimpleXMLElement`, `DOMNode`, `GMP`, or `Decimal\Decimal` are valid. We can try to get more specific later if it's possible.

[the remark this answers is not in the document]
AndrolGenhald: Try `$statements_analyzer->getCodebase()->classExtends(...)`, should also make the check a lot easier compared to what you have now. Also, maybe call it `$pseudo_castable_classes` since it's not really related to `__toString`, and (just to be _really_ nitpicky) make it a private const instead of a static var since it shouldn't change.

AndrolGenhald:
Out of curiosity, what was the issue with the broken bitwiseAssignment test?

AndrolGenhald:
Huh, weird. Guess there's other bugs around causing that then. @orklah everything look good to you?

orklah:
Code seems good, could you add some tests for RiskyCast please?

orklah:
Thanks!

PR #8370, opened 2022-08-04
Ensure we recognize inherited static methods for the first-class callables

### LINKED ISSUE #8363: first-class callables type is not inferred if it is an inherited static method
refs #6412

it seems the support for the first-class callables is not yet complete:
https://psalm.dev/r/bbc03df04c

someniatko:
fixes #8363

I have a strong feeling it's not the correct place where this functionality should belong.

For now first-class callables are implemented in the StaticCallAnalyzer just because they look like a static call syntax-wise (and are such from the perspective of `nikic/php-parser`). However, I am not sure where they should really belong: ClosureAnalyzer also seems odd here.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php
@@ -56,6 +56,7 @@ use Psalm\Type\Union;
 
 use function array_filter;
 use function array_map;
+use function array_values;
 use function assert;
 use function count;
 use function in_array;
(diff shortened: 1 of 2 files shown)


someniatko:
The [original example](https://psalm.dev/r/bbc03df04c) from the issue still doesn't work: it says
```
Psalm\Exception\CodeException : InvalidArgument - src/somefile.php:27:38 - Argument 1 of takesIntToHolder expects Closure(int):Holder<C>, but impure-Closure(int):Holder<static> provided
```

someniatko:
Okay, so I managed to fix my exact case in the last commit, however it still feels like a hack and duplicated efforts.

Needless to say it will suffer from other `static` problems which were fixed or maybe will be fixed in future for other features of the language.

Perhaps the proper way to implement the first-class callables is first to desugar them into a closure?

orklah:
It's one of the feature that was implemented after Matt departed from the project, so sadly, it's possible it was added without a "big picture" mindset at the time.
I didn't work very much on first class callable either so I wouldn't be a good adviser on that.

Frankly, if you feel refactoring it would be best and you're up for it, go ahead! Now is the best moment to do so, Psalm 5 was not yet released so implementation changes won't bother anyone depending on it

orklah:
> Each such fix for me is a debugging exercise

There's a few modules in Psalm I know and I can pinpoint some issues, but aside from that, I'm really the same. For example, I'm pretty sure you know more than me about static issues because I never went check how it works and it would take me a lot of time too :)

I'm up to merge this :)

someniatko:
> I'm up to merge this :)

@orklah pingie 👉👈

orklah:
Thanks!

PR #8372, opened 2022-08-05
always use lock when writing/reading cache data to/from file

kkmuffme:
ran into igbinary end-of-data race condition with PHP 8.1 in a multi-threaded environment in some cases where the file was modified during reading

I just used `UnexpectedValueException` as that is what was there already in one case. If you prefer another exception, please let me know.

### DIFF
--- a/src/Psalm/Internal/Provider/ClassLikeStorageCacheProvider.php
+++ b/src/Psalm/Internal/Provider/ClassLikeStorageCacheProvider.php
@@ -3,13 +3,13 @@
 namespace Psalm\Internal\Provider;
 
 use Psalm\Config;
+use Psalm\Internal\Provider\Providers;
 use Psalm\Storage\ClassLikeStorage;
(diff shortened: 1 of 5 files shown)


kkmuffme:
Code review please

orklah:
Could we move the method to somewhere common so we could have only one?

kkmuffme:
Yes, if you tell me where you want it :-)

Also which exception do you want then?

orklah:
Providers.php seems fine

For the exception, maybe RuntimeException?

kkmuffme:
Done, ready for review again.

There's an internal error with CircleCI that's not from this branch, which is why the test fails:
https://app.circleci.com/pipelines/github/vimeo/psalm/9325/workflows/3ee42730-0f0b-48f7-969a-8ef4a937b743/jobs/34190

orklah:
Thanks!

PR #8376, opened 2022-08-05
Use classlike_storage_provider only if it has the required data

### LINKED ISSUE #8373: Uncatched exception if psalm-assert references unknown class/interface
https://github.com/vimeo/psalm/blob/1ef3851580acef380d083d9b4044bd8c6204b780/src/Psalm/Internal/Type/TemplateStandinTypeReplacer.php#L1142

Might fail if the referenced class-like `$input_type_part->value` does not exists.

introduced in af5c191e7b446fb3f7f45a9d8f8c1df4e0ce9099

This can happen when a library has a `@psalm-assert` using a class/interface that it and the main application does not explicitly require.

An example is when using amphp/amp function `all` https://github.com/amphp/amp/blob/9d5100cebffa729aaffecd3ad25dc5aeea4f13bb/lib/functions.php#L384 which uses `React\Promise\PromiseInterface` in `@psalm-assert` but `React` is not hard requirement for `Amp` nor our application.

Wrapping the code inside a try-catch seems to fix the issue, but is this a correct way to fix this issue and is an empty array a good fallback for `$input_type_params`?
```
            try {
                $class_storage = $codebase->classlike_storage_provider->get($input_type_part->value);

                $container_class = $container_type_part->value;

                if (strtolower($input_type_part->value) === strtolower($container_type_part->value)) {
                    $input_type_params = $class_storage->getClassTemplateTypes();
                } elseif (!empty($class_storage->template_extended_params[$container_class])) {
                    $input_type_params = array_values($class_storage->template_extended_params[$container_class]);
                } else {
                    $input_type_params = array_fill(0, count($class_storage->template_types ?? []), Type::getMixed());
                }
            } catch (Throwable $e) {
                $input_type_params = [];
            }
```

Daeroni:
Fixes #8373

### DIFF
--- a/src/Psalm/Internal/Type/TemplateStandinTypeReplacer.php
+++ b/src/Psalm/Internal/Type/TemplateStandinTypeReplacer.php
@@ -1138,7 +1138,7 @@ class TemplateStandinTypeReplacer
     ): array {
         if ($input_type_part instanceof TGenericObject || $input_type_part instanceof TIterable) {
             $input_type_params = $input_type_part->type_params;
-        } else {
+        } elseif ($codebase->classlike_storage_provider->has($input_type_part->value)) {
             $class_storage = $codebase->classlike_storage_provider->get($input_type_part->value);
 
             $container_class = $container_type_part->value;
@@ -1150,6 +1150,8 @@ class TemplateStandinTypeReplacer
             } else {
                 $input_type_params = array_fill(0, count($class_storage->template_types ?? []), Type::getMixed());
             }
+        } else {
+            $input_type_params = [];
         }
 
         try {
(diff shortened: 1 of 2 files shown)


AndrolGenhald:
Could you add a test to ensure this continues working in the future? Probably add something in `AssertAnnotationTest`.

Daeroni:
It seems the bug requires the `@psalm-assert` annotation to be in a file that is not being analyzed (eg. library code from vendor-dir) but is included by the code being analyzed.

I tried to write a separate test method that adds to files to the scanner with `addFilesToShallowScan` and then just call `addFilesToAnalyze` to the file that is being tested (somewhat copied from the IncludeTest), but was unable to trigger the error.

I also tried to make a test project using the same package and similar code that our project has that triggers this error, but I couldn't reproduce it.

AndrolGenhald:
How about [this](https://psalm.dev/r/cce84834fa)?

Daeroni:
@AndrolGenhald Thanks, that seems to work.

AndrolGenhald:
Thanks!

PR #8384, opened 2022-08-08
fix: Return `TNever` for class intersections

### LINKED ISSUE #8333: Psalm crashing with intersection types in latest builds
https://psalm.dev/r/c189b6d2b0

petewalker:
Returns `TNever` for an intersection of 2 non-overlapping named classes instead of throwing an exception. Will raise a new `InvalidIntersectionType` in the case that this is detected.

I *think* this is all that's needed to fix this issue.

I've invented a new issue type - not sure whether there's a set process for this, but none of the existing types seemed to match.

Resolves #8333

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php
@@ -12,6 +12,7 @@ use Psalm\Aliases;
 use Psalm\CodeLocation;
 use Psalm\Codebase;
 use Psalm\Internal\Analyzer\ClassLikeAnalyzer;
+use Psalm\Issue\InvalidIntersectionType;
 use Psalm\Issue\ParseError;
 use Psalm\IssueBuffer;
 use Psalm\Storage\ClassLikeStorage;
@@ -113,8 +114,20 @@ class TypeHintResolver
                 $type = Type::intersectUnionTypes($resolved_type, $type, $codebase);
             }
 
-            if ($type === null) {
-                throw new UnexpectedValueException('Intersection type could not be resolved');
+            if ($type->isNever()) {
+                IssueBuffer::maybeAdd(
+                    new InvalidIntersectionType(
+                        'Intersection types can never be satisfied',
+                        $code_location
+                    )
+                );
+            } elseif ($type === null) {
+                IssueBuffer::maybeAdd(
+                    new ParseError(
+                        'Intersection type could not be resolved',
+                        $code_location
+                    )
+                );
             }
 
             return $type;
--- a/src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php
@@ -121,7 +121,7 @@ class TypeHintResolver
             if ($type->isNever()) {
                 IssueBuffer::maybeAdd(
                     new InvalidIntersectionType(
-                        'Intersection types can never be satisfied',
+                        'Intersection type can never be satisfied',
                         $code_location
                     )
                 );
--- a/src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php
@@ -114,17 +114,14 @@ class TypeHintResolver
                 $type = Type::intersectUnionTypes($resolved_type, $type, $codebase);
             }
 
+            if ($type === null) {
+                throw new UnexpectedValueException('Intersection type could not be resolved');
+            }
+
             if ($type->isNever()) {
                 IssueBuffer::maybeAdd(
                     new InvalidIntersectionType(
                         'Intersection types can never be satisfied',
-                        $code_location
-                    )
-                );
-            } elseif ($type === null) {
-                IssueBuffer::maybeAdd(
-                    new ParseError(
-                        'Intersection type could not be resolved',
                         $code_location
                     )
                 );
--- a/src/Psalm/Issue/InvalidIntersectionType.php
+++ b/src/Psalm/Issue/InvalidIntersectionType.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Psalm\Issue;
+
+final class InvalidIntersectionType extends CodeIssue
+{
+    public const ERROR_LEVEL = 6;
+    public const SHORTCODE = 283;
+}
--- a/src/Psalm/Issue/InvalidIntersectionType.php
+++ b/src/Psalm/Issue/InvalidIntersectionType.php
@@ -5,5 +5,5 @@ namespace Psalm\Issue;
 final class InvalidIntersectionType extends CodeIssue
 {
     public const ERROR_LEVEL = 6;
-    public const SHORTCODE = 283;
+    public const SHORTCODE = 313;
 }
--- a/src/Psalm/Type.php
+++ b/src/Psalm/Type.php
@@ -600,6 +600,9 @@ abstract class Type
                         );
 
                         if (null !== $intersection_atomic) {
+                            if ($intersection_atomic instanceof TNever) {
+                                return new Union([$intersection_atomic]);
+                            }
                             if (null === $combined_type) {
                                 $combined_type = new Union([$intersection_atomic]);
                             } else {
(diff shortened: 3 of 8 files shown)


src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php
```diff
@@ -117,6 +118,15 @@ public static function resolve(
                 throw new UnexpectedValueException('Intersection type could not be resolved');
             }
 
+            if ($type->isNever()) {
+                IssueBuffer::maybeAdd(
+                    new InvalidIntersectionType(
+                        'Intersection types can never be satisfied',
```
AndrolGenhald, line 124: Typo? Intersection **type** can never be satisfied?

src/Psalm/Issue/InvalidIntersectionType.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Psalm\Issue;
+
+final class InvalidIntersectionType extends CodeIssue
+{
+    public const ERROR_LEVEL = 6;
+    public const SHORTCODE = 283;
```
AndrolGenhald, line 8: @orklah do you know why we have a gap in shortcodes? Usually the think to do is run `bin/max_used_shortcode.php` then add 1 to it when adding a new issue.

AndrolGenhald:
I think sometimes for impossible types we return `mixed` instead of `never`, but I just tested it and [here's another case](https://psalm.dev/r/ea91f1cf0b) where we return `never`, so I guess it's fine. We should probably make sure we're consistent about that at some point.

AndrolGenhald:
> I've invented a new issue type - not sure whether there's a set process for this, but none of the existing types seemed to match.

Looks fine to me. I think I might be the one who has added the most issues besides Matt (who created Psalm), and my process has basically been create the issue, run the Documentation test, fix whatever fails :laughing: I've considered writing up documentation for adding issues, but I think it happens infrequently enough that it's not worth it, since the tests basically walk you through it anyway.

src/Psalm/Issue/InvalidIntersectionType.php
petewalker: I've bumped the shortcode using the command 👍

The `ERROR_LEVEL` was a copy-paste. Happy to change it, I'm not that familiar with the scale!

tests/NativeIntersectionsTest.php
```diff
@@ -83,6 +83,20 @@ function test(A&B $in): void {
...
+                    class Baz
+                    {
+                        public function __construct(private Foo&Bar $foobar) {}
+                    }
+                    new Baz(new Foo());
+                    ',
+                'error_message' => 'InvalidIntersectionType',
+                'ignored_issues' => [],
+                'php_version' => '8.1'
+            ],
```
AndrolGenhald, line 99: Can we add a test for docblock types as well? What happens [here](https://psalm.dev/r/fa2f58ec4d)? If it doesn't show the issue and you prefer to leave that for a future PR just put `SKIPPED-` in front of the test name and I'll make an issue to track it.

src/Psalm/Issue/InvalidIntersectionType.php
AndrolGenhald: > I'm not that familiar with the scale!

Neither am I! :laughing:

I mostly run Psalm at level 1 on newer projects, so I don't feel like I have a good idea on what should go where. There's some documentation [here](https://psalm.dev/docs/running_psalm/error_levels/) (that I've been meaning to write a test for, I'm sure a bunch of issues are missing from those lists).

tests/NativeIntersectionsTest.php
petewalker: It doesn't show the issue - and if we did that right now, I think it would probably surface a lot of issues for people who make use of `@var MockObject&SomeNonFinalClass`.

I know that's use in a few places in my company's source code, for example! I guess a decision would need to be made on the desired behaviour.

petewalker: Oh, wait - `MockObject` is an interface. Ignore that..!

src/Psalm/Type.php
```diff
@@ -600,6 +600,9 @@ public static function intersectUnionTypes(
                         );
 
                         if (null !== $intersection_atomic) {
+                            if ($intersection_atomic instanceof TNever) {
+                                return new Union([$intersection_atomic]);
+                            }
```
AndrolGenhald, line 605: I'm not quite sure this is correct. What if you have something like [this](https://psalm.dev/r/332b1a4764) and we end up calling `intersectUnionTypes` with `(Foo&A)|(Bar&B)` and `Bar&B`, I think this would call `intersectAcomicTypes` on `Foo&A` and `Bar&B`, get `TNever` and return `never`, when it should actually skip past the `never` result from the first iteration and return the `Bar&B` result when intersecting `Bar&B` and `Bar&B`.

petewalker: Ah yes, I see what you mean. I'll iterate.

petewalker: Okay, this doesn't seem to be an issue - but I'm not entirely sure why. I've added a few extra tests, and can't get it to fall over. I can only assume that types derived from docblocks don't call this code?

AndrolGenhald: Yeah, I'm having trouble actually finding a way to call `intersectUnionTypes` as well... It might be easiest to just add a normal unit test that creates the types and calls it directly.

AndrolGenhald: I think it's because `Type::intersectAtomicTypes` is slightly broken in that it's not checking for impossible intersections between two classes, so when you intersect `Foo&A` and `Bar&B` you actually get `Foo&A&Bar&B` instead of `never`. I'd say just fix it without having a test.

AndrolGenhald: Here's the test case I stepped through to figure it out:

```php
class Foo {}
class Bar {}
interface A {}
interface B {}
class FooA extends Foo implements A {}
class BarB extends Bar implements B {}

interface FooBarFactoryInterface
{
    /** @return (Foo&A)|(Bar&B) */
    public function create(): object;
}

interface FactoryInterface
{
    public function create(): object;
}

abstract class AbstractFooAFactory
{
    public function create(): Foo&A
    {
        return new FooA();
    }
}

class FooAFactory extends AbstractFooAFactory implements FooBarFactoryInterface, FactoryInterface {}

function getFooAFactory(): AbstractFooAFactory&FooBarFactoryInterface&FactoryInterface
{
    return new FooAFactory();
}

$factory = getFooAFactory();
$foobar = $factory->create();
```

It's pretty brittle though and there are a lot of unrelated things that could change and cause it to not test what it's supposed to be testing. If/when this is fixed I think it's worth having a separate test that just calls `intersectUnionTypes` directly.

src/Psalm/Issue/InvalidIntersectionType.php
orklah: Basically, the scale is 8 = super bad, will directly lead to fatal errors or will prevent Psalm from analyzing at a deep level (for example a missing class), 1= Mixed issues (can't say if bad or not)

In between you have the gradient. For example, `InvalidArgument` will be at level 6, `PossiblyInvalidArgument` will be at level 3.

For me, this particular issue is almost the same as `NoValue` (basically, every possible Type has been excluded so all that's left is `Never`: there is no type that will fit that particular variable/expression). Except that NoValue is emitted when using a `Never` expression, while this would be emitted while encountering one. However, NoValue is at `-1` (usually used when a particular issue is not dependent on level, for example when it's behind a config flag so it's always reported when triggered). NoValue used to be reported rarely (namely when using the return of a never-returning function) but it took a much bigger role recently.

Maybe we should decide a level and stick NoValue and this on the same level?

AndrolGenhald:
@petewalker Are you planning on coming back to this? Do you consider it done from your point of view? Should we go with #8385 instead? I think the issue I mentioned previously still needs fixed, I believe it's masked by a separate issue but I'd still rather have it fixed now instead of having it show up later when other parts of the code are improved.

petewalker:
@AndrolGenhald Apologies! I'd say it's unlikely at this stage - career changes are taking me away from PHP, so this project is unfortunately less on my radar. Happy for this PR to be built on by someone closer to the problem, as in the PR you referenced, or whatever, really..!

orklah:
I'll close this then :)

PR #8385, opened 2022-08-09
fix: Fix parsing of invalid intersection types

### LINKED ISSUE #8333: Psalm crashing with intersection types in latest builds

danog:
This fixes #8333 properly, replacing #8384.
I feel that we need to be consistent with the return type of `Type::intersectUnionTypes`: if there is no possible intersection between the types, return null, otherwise return a Union with the types.
The previous PR incorrectly returned TNever only for certain invalid intersections: if we really want to go with that approach, then let's always return either a Union or a Union(TNever), and never `null` (but I would personally leave it as it is right now, as it's easier to tell if there's no intersection between the types this way).

### DIFF
--- a/src/Psalm/Type.php
+++ b/src/Psalm/Type.php
@@ -600,9 +600,6 @@ abstract class Type
                         );
 
                         if (null !== $intersection_atomic) {
-                            if ($intersection_atomic instanceof TNever) {
-                                return new Union([$intersection_atomic]);
-                            }
                             if (null === $combined_type) {
                                 $combined_type = new Union([$intersection_atomic]);
                             } else {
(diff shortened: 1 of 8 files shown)


orklah:
Well, the question is, does `Type::intersectUnionTypes` sometimes return a `Never` anyway?

If it can happen, it would definitely mean a InvalidIntersectionType should be emitted and your change would not do that. If it can't happen then yeah, we should try to keep consistency.

danog:
All fixed!

danog:
Actually nevermind, it seems like there's a bug in the assertion reconciler

weirdan:
> Actually nevermind, it seems like there's a bug in the InArray assertion reconcile

Isn't it something you were fixing in #8773 (among other things)?

src/Psalm/Type.php
```diff
@@ -540,25 +540,19 @@ public static function combineUnionTypes(
...
      * Combines two union types into one via an intersection
      *
-     *
+     * Returns null (not TNever) if the intersection is empty
      */
     public static function intersectUnionTypes(
-        ?Union $type_1,
-        ?Union $type_2,
+        Union $type_1,
+        Union $type_2,
```
weirdan, line 547: Do we *really need* this BC break here? This is a part of our API surface, so the earliest it can change is in Psalm 6.

PR #8403, opened 2022-08-12
Fix `imageinterlace` function signature

paulfedorow:
This PR fixes the function signature of [`imageinterlace`](https://www.php.net/manual/en/function.imageinterlace.php):

* The return type is `bool` since PHP 8.0.5.
* The type of the parameter `enable` is `bool|null` since PHP 8.0.0.


orklah:
Thanks!

PR #8407, opened 2022-08-12
bump slevomat/coding-standard to fix CI

orklah:
This will fix the CI crash


orklah:
...except we can't run that on PHP 7.1.

I'll try to fix that another way. The core issue seems to come from phpstan/phpdoc-parser in version 1.7. We could pin the previous version on 4.x

weirdan:
@orklah there's already a number of packages removed on PHP 7.1, just add slevomat/coding-standard to that list. CS checks do not run on 7.1 anyway.

PR #8415, opened 2022-08-16
Consistent error logging for cache

kkmuffme:
Generally, psalm throws errors in cache.

However there were 2 places, where error_log was used/leftover from early versions. Changed now for consistency.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -35,6 +35,7 @@ use Psalm\Issue\VariableIssue;
 use Psalm\Plugin\PluginEntryPointInterface;
 use Psalm\Progress\Progress;
 use Psalm\Progress\VoidProgress;
+use RuntimeException;
 use SimpleXMLElement;
 use SimpleXMLIterator;
 use Throwable;
@@ -55,7 +56,6 @@ use function class_exists;
 use function clearstatcache;
 use function count;
 use function dirname;
-use function error_log;
 use function explode;
 use function extension_loaded;
 use function file_exists;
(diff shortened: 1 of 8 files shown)


kkmuffme:
@orklah there is a bug in psalm unrelated to this PR? Could you please quickly check

Test fails: https://github.com/vimeo/psalm/runs/7860346355?check_suite_focus=true (which will cause other tests to fail too), due to $config being null (how is this even possible when construct is called with Config, as it should already give an error then??)

When I add a check for that, it says config cannot be null:
https://github.com/vimeo/psalm/runs/7860822542?check_suite_focus=true

AndrolGenhald:
`Psalm\Tests\Internal\Provider\ParserInstanceCacheProvider` extends `ParserCacheProvider` and overrides its constructor to do nothing, so the parent constructor is never called.

kkmuffme:
Ok so I guess I will put back the call to static config as it was there before. (since properly fixing this would mean rewriting tons of code that is unrelated to this PR)

src/Psalm/Config.php
```diff
@@ -1013,8 +1013,16 @@ private static function fromXmlAndPaths(
             chdir($config->base_dir);
         }
 
-        if (is_dir($config->cache_directory) === false && @mkdir($config->cache_directory, 0777, true) === false) {
-            error_log('Could not create cache directory: ' . $config->cache_directory);
+        if (!is_dir($config->cache_directory)) {
+            try {
+                mkdir($config->cache_directory, 0777, true);
```
orklah, line 1018: seems weird to not check mkdir return. Does it throws on any error with some config?

orklah:
Thanks

Issue #8346, opened 2022-07-29
4.25.0 Regression: Could not get class storage for resourcebundle for PHP < 8.0


kkmuffme:
When I run --alter:

`php /opt/composer/vendor/bin/psalm --config=my-rules.xml --alter --plugin=vendor/orklah/psalm-strict-equality/src/Plugin.php --php-version=7.4 my-file.php` 

When I do `--php-version=8.0` or `--php-version=8.1` I don't get an error. For everythint <= 7.4 I get this error.

```
Uncaught InvalidArgumentException: Could not get class storage for resourcebundle in /path/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:46
Stack trace:
#0 /path/vendor/vimeo/psalm/src/Psalm/Internal/Type/TypeExpander.php(607): Psalm\Internal\Provider\ClassLikeStorageProvider->get('ResourceBundle')
#1 /path/vendor/vimeo/psalm/src/Psalm/Internal/Type/TypeExpander.php(178): Psalm\Internal\Type\TypeExpander::expandNamedObject(Object(Psalm\Codebase), Object(Psalm\Type\Atomic\TNamedObject), NULL, NULL, NULL, false, true)
#2 /path/vendor/vimeo/psalm/src/Psalm/Internal/Type/TypeExpander.php(82): Psalm\Internal\Type\TypeExpander::expandAtomic(Object(Psalm\Codebase), Object(Psalm\Type\Atomic\TNamedObject), NULL, NULL, NULL, true, false, false, true, false)
#3 /path/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php(284): Psalm\Internal\Type\TypeExpander::expandUnion(Object(Psalm\Codebase), Object(Psalm\Type\Union), NULL, NULL, NULL, true, false, false, true)
[...]

PR #8416, opened 2022-08-16
fix crash in <PHP8 with ResourceBundle

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/8346

partially reverts https://github.com/vimeo/psalm/pull/8217

### DIFF
--- a/tests/Internal/Codebase/InternalCallMapHandlerTest.php
+++ b/tests/Internal/Codebase/InternalCallMapHandlerTest.php
@@ -51,6 +51,7 @@ class InternalCallMapHandlerTest extends TestCase
     private static $ignoredFunctions = [
         'apcu_entry',
         'array_multisort',
+        'count', // #8346
         'bcdiv',
         'bcmod',
         'bcpowmod',
--- a/tests/Internal/Codebase/InternalCallMapHandlerTest.php
+++ b/tests/Internal/Codebase/InternalCallMapHandlerTest.php
@@ -51,7 +51,6 @@ class InternalCallMapHandlerTest extends TestCase
     private static $ignoredFunctions = [
         'apcu_entry',
         'array_multisort',
-        'count', // #8346
         'bcdiv',
         'bcmod',
         'bcpowmod',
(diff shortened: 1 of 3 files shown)


orklah:
count used to contain ResourceBundle, the previous PR removed it for newer version. I don't think dropping it altogether is the way to go here. We better keep it even on PHP >8

orklah:
count used to have ResourceBundle: https://github.com/vimeo/psalm/pull/8217/files#diff-5611cd00854405fe61befddd211dacd225d9c479b499ffd7e3c0c7d96eb568edR216

The change in https://github.com/vimeo/psalm/pull/8217 was to remove ResourceBundle in latest php versions definitions. So reverting that would mean still have ResourceBundle on every version like before, not removing it altogether

AndrolGenhald:
FWIW I do think `ResourceBundle` needs removed from the callmap (see [here](https://github.com/vimeo/psalm/issues/8346#issuecomment-1216940311)). Referencing extension-defined classes in the callmap in functions not defined by that extension is probably a bad idea. I'm not sure what the best fix is though, a provider for `count` seems like overkill in this case, but that might be the only way to correctly fix it.

orklah:
ouch, I didn't know it was an extension... Let's fix this the way it was and we'll look for a solution after I think

@kkmuffme can you check tests 1 and 4? There's callmap consistencies failures in tests

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -519,7 +519,13 @@ public function testIgnoresAreSortedAndUnique(): void
             /** @var string */
             $function = is_int($key) ? $value : $key;
 
-            $this->assertGreaterThan(0, strcmp($function, $previousFunction));
+            $diff = strcmp($function, $previousFunction);
```
orklah, line 522: Is this on purpose? Won't PHPUnit scream if there are outputs in a test?

orklah:
Thanks!

PR #8422, opened 2022-08-19
Fix DatePeriod tests for implementing IteratorAggregate

fluffycondor:
I wrote them in a wrong way in #8312, so they are rewritten now as @AndrolGenhald advised in #8327.
But actually I'm quite surprised Psalm allows me to call `getIterator()` on a pure Traversable. It actually results in Fatal Error: https://3v4l.org/MEpb3 (But it's the only one way I've found to test it.)
Looks like PHP 8 stubs indeed leaks to PHP 7 environment somehow.

### DIFF
--- a/tests/CoreStubsTest.php
+++ b/tests/CoreStubsTest.php
@@ -86,23 +86,21 @@ class CoreStubsTest extends TestCase
             'error_levels' => [],
             'php_version' => '8.0',
         ];
-        yield 'DatePeriod implements only Traversable on PHP 7' => [
-            '<?php
-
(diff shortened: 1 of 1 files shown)


AndrolGenhald:
I was thinking it should be in provider**Invalid**CodeParse to make sure it has an error, if that test is working fine then it does seem like something is broken.

fluffycondor:
> it should be in providerInvalidCodeParse
I've tried it like that, it yields no error. I mean the test fails with the reason "No CodeException was thrown" or something. So Psalm thinks this code is valid in PHP 7.3 environment.

weirdan:

orklah:

PR #8428, opened 2022-08-23
Allow any attribute for complex types in schema

den-kuz:
I think we can use full power of [XInclude](https://www.w3.org/TR/xinclude/) and allow any attribute with `lax` or `skip` value in `processContents` attribute. This gives more flexibility of building dynamic configs.

For example, with current schema i can't use XInclude for issueHandlers configuration, because it has "anyAttribute" configuration, but without `processContents="lax"` or `processContents="skip"` attribute. This PR solves this problem and other problems like this


den-kuz:
Ah, i see this bug already fixed in psalm 5 same way, can we just bring full support of xinclude in 4.x?

Cimus:
Really need

den-kuz:
@orklah can you review this pr?

orklah:
Thanks!

PR #8431, opened 2022-08-23
Allow *bin2hex and *bin2base64 functions to keep non-empty-string type

LeSuisse:
Those functions should not return a string when they receive a non-empty-string in input.

The following example is expected to work:
```php
<?php

/**
 * @param non-empty-string $i
 */
function takesNonEmptyString(string $i): void {
    echo $i;
}

takesNonEmptyString(bin2hex("a"));
takesNonEmptyString(base64_encode("a"));
```

https://psalm.dev/r/0d5b2b3965

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -335,6 +335,22 @@ function fclose(&$stream) : bool
 }
 
 /**
+ * @psalm-pure
+ * @template T as string
+ * @param T $string
+ * @return (T is non-empty-string ? non-empty-string : string)
+ */
+function sodium_bin2base64(string $string, int $id): string
+
+/**
+ * @psalm-pure
+ * @template T as string
+ * @param T $string
+ * @return (T is non-empty-string ? non-empty-string : string)
+ */
+function sodium_bin2hex(string $string): string {}
+
+/**
  * @param string $string
  * @param-out null $string
  */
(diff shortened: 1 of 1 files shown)


stubs/CoreGenericFunctions.phpstub
```diff
@@ -334,6 +334,22 @@ function fclose(&$stream) : bool
...
+ */
+function sodium_bin2base64(string $string, int $id): string
+
+/**
+ * @psalm-pure
+ * @template T as string
+ * @param T $string
+ * @return T
+ */
+function sodium_bin2hex(string $string): string {}
```
AndrolGenhald, line 351: This isn't quite right, passing a `numeric-string` shouldn't return a `numeric-string`. Please use a conditional return type instead.

AndrolGenhald:
I believe the `sodium_*` functions require an extension to be installed? Could we add a separate stub for that and conditionally include it like the other extension stubs?

LeSuisse:
> I believe the `sodium_*` functions require an extension to be installed? Could we add a separate stub for that and conditionally include it like the other extension stubs?

Hum maybe? I took the same approach than for the `sodium_*` functions that were already in the file.

AndrolGenhald:
> I took the same approach than for the sodium_* functions that were already in the file.

Well, if they're already there I guess this is fine then. Maybe I'll get around to moving them to a separate extension stub for Psalm 5.

AndrolGenhald:
Thanks!

stubs/CoreGenericFunctions.phpstub
```diff
@@ -334,6 +334,22 @@ function fclose(&$stream) : bool
...
+ * @psalm-pure
+ * @template T as string
+ * @param T $string
+ * @return (T is non-empty-string ? non-empty-string : string)
+ */
+function sodium_bin2base64(string $string, int $id): string
+
+/**
+ * @psalm-pure
+ * @template T as string
```
staabm, line 347: just a headsup from the review in phpstan. we don't need a template in phpstan for this case, see https://github.com/phpstan/phpstan-src/pull/1664

maybe thats also the case for psalm and this can be simplified

orklah: yeah, same on Psalm, but stubs are transparent for users so it doesn't really matters

PR #8432, opened 2022-08-24
Update call maps for MongoDB extension

alcaeus:
While adding psalm checks to the MongoDB driver, I noticed that a number of method signatures in the call map were either missing or out-of-date. This PR updates them to what we'll ship in 1.15. They have been automatically generated from the definitions in the driver and made to match the format required by Psalm.


orklah:
Thanks! Can you look at the mismatch reported in Unit Tests 4?

alcaeus:
Ah, there were some more (lowercased) entries in a completely separate portion of the file. These were now duplicated, but since they were originally lowercased they were treated as different. I removed the obsolete definitions.

alcaeus:
> I noticed that some of the interface declarations aren't correct. Please hold off on merging while I figure out where the script failed (probably has to do with tentative return types).

That's fixed now. I didn't consider tentative return types for interfaces.

dictionaries/CallMap.php
```diff
@@ -7693,256 +7693,314 @@
...
-'MongoDB\BSON\binaryinterface::getType' => ['int'],
-'MongoDB\BSON\dbpointer::__construct' => ['void'],
-'MongoDB\BSON\dbpointer::__toString' => ['string'],
-'MongoDB\BSON\dbpointer::jsonSerialize' => ['mixed'],
-'MongoDB\BSON\dbpointer::serialize' => ['string'],
-'MongoDB\BSON\dbpointer::unserialize' => ['void', 'serialized'=>'string'],
-'MongoDB\BSON\Decimal128::__construct' => ['void', 'value='=>'string'],
+'MongoDB\BSON\Binary::__toString' => ['string'],
+'MongoDB\BSON\Binary::serialize' => ['string'],
+'MongoDB\BSON\Binary::unserialize' => ['void', 'serialized' => ''],
```
orklah, line 7701: serialized should have a type here, either mixed or more narrow

dictionaries/CallMap.php
```diff
@@ -7693,256 +7693,314 @@
...
+'MongoDB\BSON\Binary::jsonSerialize' => ['mixed'],
+'MongoDB\BSON\BinaryInterface::getData' => ['string'],
+'MongoDB\BSON\BinaryInterface::getType' => ['int'],
+'MongoDB\BSON\BinaryInterface::__toString' => ['string'],
+'MongoDB\BSON\DBPointer::__toString' => ['string'],
+'MongoDB\BSON\DBPointer::serialize' => ['string'],
+'MongoDB\BSON\DBPointer::unserialize' => ['void', 'serialized' => ''],
```
orklah, line 7708: same

dictionaries/CallMap.php
```diff
@@ -7693,256 +7693,314 @@
...
-'MongoDB\BSON\binary::unserialize' => ['void', 'serialized'=>'string'],
-'MongoDB\BSON\binaryinterface::__toString' => ['string'],
-'MongoDB\BSON\binaryinterface::getData' => ['string'],
-'MongoDB\BSON\binaryinterface::getType' => ['int'],
-'MongoDB\BSON\dbpointer::__construct' => ['void'],
-'MongoDB\BSON\dbpointer::__toString' => ['string'],
-'MongoDB\BSON\dbpointer::jsonSerialize' => ['mixed'],
-'MongoDB\BSON\dbpointer::serialize' => ['string'],
-'MongoDB\BSON\dbpointer::unserialize' => ['void', 'serialized'=>'string'],
-'MongoDB\BSON\Decimal128::__construct' => ['void', 'value='=>'string'],
```
orklah, line 7711: In case you were wondering, the = sign after the parameter name is used to signal that the parameter is optional. Please make sure you actually wanted to remove it

orklah (changes requested):
Can you sort declaration in alphabetical order? It seems to be the previous order and it would help reviewing. Please see a few notes also

dictionaries/CallMap.php
alcaeus: Ah, thanks. Added the correct `string` type for all `unserialize` methods.

dictionaries/CallMap.php
alcaeus: I wasn't wondering - this was quite intentional: https://github.com/mongodb/mongo-php-driver/pull/1347/files#r948261399

dictionaries/CallMap.php
orklah: I don't see a new commit on the branch? Forgot to push?

alcaeus: Whopsie, yes. I pushed the commit now.

orklah: Well, I have bad news then, I still don't see anything

orklah: nvm, I was mistaken apparently

orklah:
Thanks!

PR #8435, opened 2022-08-24
Add support for callable in array_reduce

VincentLanglet:
Hi @orklah,

I applied the same fix than https://github.com/vimeo/psalm/pull/5373 but for array_reduce.

I assume another fix should be done to array_filter too.


AndrolGenhald:
Thanks!

PR #8442, opened 2022-08-30
Fix properties-of on generic objects&intersections + immutable atomic refactoring

danog:
Based on #8143, lays the groundwork for #8398.
Also fixes properties-of on generic objects and intersections.


danog:
Note on baselined issues: just a few unused method issues:)

Issue #8419, opened 2022-08-17
Type narrowing doesn't work after string to int conversion


fluffycondor:
https://psalm.dev/r/bf49f34907

A real-world example, quite easy, but it's a hard one for a static analyzer.
I'd be nice to see `int<2022, max>` there someday :)

danog:
You just need to change the assertion layout a bit: https://psalm.dev/r/e9953d77a7

orklah:
Yeah, a few things:
- substr on literal strings is not computed right now, so it result in a generic string by default. You could totally contribute to that change if that's your personal itch. 
- ctype_digit is not recognized by psalm currently. Likewise, it would be easy to add here: https://github.com/vimeo/psalm/blob/b69e22a33be8520bba9052d0bbabd5e6d525719c/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php#L1900
- Psalm is currently unable to detect assertions (understand type narrowing) when the variable is not directly checked (here you check the result of the cast and not the variable directly). This would be more difficult to implement I think (because it would mean retroactively changing the type of $a because a specific operation on $a gives you a specific result)

fluffycondor:
> Likewise, it would be easy to add here
Not so easy as it would be. `ctype_*` returns true not only for asserted strings but also for some ASCII codes passed as int. That "but" can be ignored though, because passing ASCII codes is deprecated in PHP 8.1.
But I'd like to stub it properly. Now looks like #8421 prevents me from doing it.

orklah:
> ctype_* functions return true not only for asserted strings but also for some ASCII codes passed as int.

doesn't matter a lot, all of that will be considered numeric. Psalm reconciliation should do the rest

PR #8450, opened 2022-09-02
Make ctype_digit and ctype_lower work as assertions

fluffycondor:
Make `ctype_digit` assert for `numeric-string`, and `ctype_lower` for `non-empty-lowercase-string`.
I've implemented it as @orklah advised it [here](https://github.com/vimeo/psalm/issues/8419#issuecomment-1218358112), but I've added two skipped tests that would fail with this naive implementation. It can be ignored though, because passing `int`s to `ctype_*` functions was deprecated in PHP 8.1.


fluffycondor:
Looks like that code style test fails isn't my fault :)

orklah:
Thanks!

> passing ints to ctype_* functions was deprecated in PHP 8.1.

Interesting, we should consider updating the callmap to reflect this from 8.1. It would prepare people to stop sending anything else than strings in there

PR #8462, opened 2022-09-06
Add dateTimeModify return type provider

### LINKED ISSUE #8454: \DatetimeImmutable::modify signature change (vs \Datetime::modify)
Currently https://psalm.dev/r/d94dac07ab is not reflecting the 4.27 release.
I'm running psalm on php 7.4 version

```
(new \DateTimeImmutable('2016-08-31'))->modify('+23 hours 59 minutes 59 seconds');
```
is reported as DateTimeImmutable&static|false
and I have an error reported when calling `getTimestamp` on it.
This is a new error related to the changes in https://github.com/vimeo/psalm/pull/8350 (@Ocramius, @orklah, @AndrolGenhald)

What is weird is the fact that
```
(new \DateTime('2016-08-31'))->modify('+23 hours 59 minutes 59 seconds');
```
is reported as DateTime&static|false
but not error is reported when calling `getTimestamp` on it.

If psalm is using a benevolent union, it seems like the recent DatetimeImmutable stub lost it.

VincentLanglet:
Close #8454

This is the same behavior than the Phpstan dynamicReturnTypeExtension
https://github.com/phpstan/phpstan-src/blob/f1fd385433480f87bdda1d7d8ff6887e25f003ba/src/Type/Php/DateTimeModifyReturnTypeExtension.php

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/DateTimeModifyReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/DateTimeModifyReturnTypeProvider.php
@@ -2,6 +2,7 @@
 
 namespace Psalm\Internal\Provider\ReturnTypeProvider;
 
+use DateTime;
 use Psalm\Internal\Analyzer\StatementsAnalyzer;
 use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
 use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
@@ -21,8 +22,7 @@ class DateTimeModifyReturnTypeProvider implements MethodReturnTypeProviderInterf
         $statements_source = $event->getSource();
         $call_args = $event->getCallArgs();
         $method_name_lowercase = $event->getMethodNameLowercase();
-        if (
-            !$statements_source instanceof StatementsAnalyzer
+        if (!$statements_source instanceof StatementsAnalyzer
             || $method_name_lowercase !== 'modify'
             || !isset($call_args[0])
         ) {
@@ -42,7 +42,7 @@ class DateTimeModifyReturnTypeProvider implements MethodReturnTypeProviderInterf
                 return null;
             }
 
-            if (@(new \DateTime())->modify($type_part->value) === false) {
+            if (@(new DateTime())->modify($type_part->value) === false) {
                 $has_false = true;
             } else {
                 $has_date_time = true;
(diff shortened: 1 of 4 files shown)


VincentLanglet:
I think Ci failure are not related

tests/DateTimeTest.php
```diff
@@ -0,0 +1,96 @@
...
+                    ',
+                'assertions' => [
+                    '$a' => 'DateTime',
+                    '$b' => 'DateTimeImmutable',
+                ],
+            ],
+            'modifyWithInvalidConstant' => [
+                '<?php
+                    /**
+                     * @return "kewk"|"koko"
```
AndrolGenhald, line 57: Can we stick to `"foo"`, `"bar"`, etc? I think using the widely known metasyntactic variables is a good idea in general so that it's more obvious that the value doesn't matter, or just use something like `"invalid modifier"|"also invalid"`. (Also, the first thing that turns up when you google "kewk" is urban dictionary...)

VincentLanglet: changed

AndrolGenhald (approved):

AndrolGenhald:
@orklah any idea what's up with the Shepherd issue? I thought I fixed it earlier so that we can't accidentally use `IssueBuffer::add` directly anymore, so I'm not sure how it's turning up vendor code.

orklah:
Most probably Psalm is retrieving the latest version of PHP-Parser that came out a few days ago and it broke something. We'll need to take a look at it

Otherwise I'm fine with the PR until we can have a working example of what we need to guard against

VincentLanglet:
> Most probably Psalm is retrieving the latest version of PHP-Parser that came out a few days ago and it broke something. We'll need to take a look at it
> 
> Otherwise I'm fine with the PR until we can have a working example of what we need to guard against

Tests are fixed @orklah :)

src/Psalm/Internal/Provider/ReturnTypeProvider/DateTimeModifyReturnTypeProvider.php
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        $has_date_time = false;
+        $has_false = false;
+        foreach ($first_arg_type->getAtomicTypes() as $type_part) {
+            if (!$type_part instanceof TLiteralString) {
+                return null;
+            }
+
+            if (@(new DateTime())->modify($type_part->value) === false) {
```
orklah, line 45: I have doubts if the @ here would work properly. Do you have a test that would produce a warning that is being silenced correctly?

AndrolGenhald: I believe `'modifyWithInvalidConstant'` should cover that: https://3v4l.org/UvjNN
I haven't actually stepped through it with a debugger to make sure it's actually running this with `->modify("foo")` though.

orklah: Great!

orklah:
Thanks!

Issue #8456, opened 2022-09-04
InvalidClass does not work PHP native classes and stubs


kkmuffme:
https://psalm.dev/r/a2caa7560a

When class Hello is not in the project files but provided via stubs, it doesn't report InvalidClass error.
This is true for global (psalm provided) and custom (user/config provided) stubs.

e.g. DateTime:
https://psalm.dev/r/78f1f31d1b

Should give an InvalidClass error too

AndrolGenhald:
Sounds good to me, go ahead and PR it and see if any tests fail!

orklah:
Beware, there is a lot of weirdness going on with stub dependencies.

Due to cache interaction and plugins, it is not clear whether the 'user-defined' part of a stub is defined.

PR #8465, opened 2022-09-07
report invalidCasing when using a class that is not user defined too

kkmuffme:
e.g. new DateTime

Fix https://github.com/vimeo/psalm/issues/8456

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -6,7 +6,7 @@ use Composer\Autoload\ClassLoader;
 use Composer\Semver\Constraint\Constraint;
 use Composer\Semver\VersionParser;
 use DOMDocument;
-use DomElement;
+use DOMElement;
 use InvalidArgumentException;
 use LogicException;
 use OutOfBoundsException;
@@ -752,7 +752,6 @@ class Config
 
         $psalm_nodes = $dom_document->getElementsByTagName('psalm');
 
-        /** @var DomElement|null */
         $psalm_node = $psalm_nodes->item(0);
 
         if (!$psalm_node) {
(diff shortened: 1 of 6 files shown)


AndrolGenhald:
> **Please do NOT merge this yet**

You can use drafts for that :slightly_smiling_face:

AndrolGenhald:
@orklah it looks fine to me, but I don't know why internal classes were excluded in the first place, so that makes me a bit nervous just merging this without investigating that first.

orklah:
> @orklah it looks fine to me, but I don't know why internal classes were excluded in the first place, so that makes me a bit nervous just merging this without investigating that first.

I'm not surprised given how many errors it raises in Psalm itself. I'm not even sure the documentation was correct on older versions of PHP for things like `DOMElement`.

@kkuffme can you fix the remaining casing errors and it should be good to go

AndrolGenhald:
> I'm not surprised given how many errors it raises in Psalm itself.

I didn't even notice those 😆 I pay less attention to that now because I know it's been failing. We should really get that fixed soon...

AndrolGenhald:
What about:

> src/Psalm/Config.php:755:18: UnnecessaryVarAnnotation: The @var DOMElement|null annotation for $psalm_node is unnecessary

Why is that showing up now? Did the incorrect casing cause the redundancy to be missed? If so maybe we should also open an issue to fix `UnnecessaryVarAnnotation` to ignore casing.

AndrolGenhald:
Hmm, that one already shows an error though. It's weird that fixing the casing on `DOMElement` would cause that `UnnecessaryVarAnnotation` to show up if it already shows up for other stuff with incorrect casing.

orklah:
Thanks! Let's fix the unnecessary var in another PR indeed

PR #8469, opened 2022-09-08
preg_replace with anchor will always only have 1 replacement

kkmuffme:
* add limit for clarity and performance
* adjust where sub-optimal


orklah:
Thanks!

PR #8473, opened 2022-09-09
make superglobals more specific

kkmuffme:
Make the types of superglobals like $_GET, $_POST,... more specific than "Mixed" to allow for simpler and stricter conditions and more effective psalm checks in many cases.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
@@ -541,18 +541,28 @@ class VariableFetchAnalyzer
 
         if ($var_id === '$argv') {
             // only in CLI, null otherwise
-            return new Union([
+            $argv_nullable = new Union([
                 new TNonEmptyList(Type::getString()),
                 new TNull()
             ]);
+            // use TNull explicitly instead of this
+            // as it will cause weird errors due to ignore_nullable_issues true
+            // e.g. InvalidPropertyAssignmentValue
+            // $this->argv 'list<string>' cannot be assigned type 'non-empty-list<string>'
+            // $argv_nullable->possibly_undefined = true;
+            $argv_nullable->ignore_nullable_issues = true;
+            return $argv_nullable;
         }
 
         if ($var_id === '$argc') {
             // only in CLI, null otherwise
-            return new Union([
+            $argc_nullable = new Union([
                 new TIntRange(1, null),
                 new TNull()
             ]);
+            // $argc_nullable->possibly_undefined = true;
+            $argc_nullable->ignore_nullable_issues = true;
+            return $argc_nullable;
         }
 
         if (!self::isSuperGlobal($var_id)) {
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
@@ -22,7 +22,6 @@ use Psalm\Issue\UndefinedVariable;
 use Psalm\IssueBuffer;
 use Psalm\Type;
 use Psalm\Type\Atomic\TArray;
-use Psalm\Type\Atomic\TFloat;
 use Psalm\Type\Atomic\TInt;
 use Psalm\Type\Atomic\TIntRange;
 use Psalm\Type\Atomic\TKeyedArray;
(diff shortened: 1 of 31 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -531,29 +540,111 @@ public static function getGlobalType(string $var_id): Union
         }
 
         if ($var_id === '$argv') {
+            // only in CLI, null otherwise
             return new Union([
-                new TArray([Type::getInt(), Type::getString()]),
+                new TNonEmptyList(Type::getNonEmptyString()),
```
AndrolGenhald, line 545:
```
$ echo -e '<?php\n\nvar_dump($argv);' >test.php && php test.php "" ""
array(3) {
  [0]=>
  string(8) "test.php"
  [1]=>
  string(0) ""
  [2]=>
  string(0) ""
}
```

AndrolGenhald:
We might have to think about this for the same reasons given in #1087. Mutating `$_GET` and other superglobals is a bad idea, but it's a bad idea that's still somewhat common unfortunately.

staabm:
Oke wa could be to make the more precise type inference a opt-in feature and also add a rule which reports errors when code modifies superglobals

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -531,26 +540,231 @@ public static function getGlobalType(string $var_id): Union
         }
 
         if ($var_id === '$argv') {
+            // only in CLI, null otherwise
```
orklah, line 543: You're right, but if we keep this as-is it will be a pain for users. Can you set `Union::ignore_nullable_issues` to true for both variables?

It will allow for nullable checks while still being lax enough to avoid reporting annoying `this may be null` errors

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -531,26 +540,231 @@ public static function getGlobalType(string $var_id): Union
...
+                    new Union([new TString(), $array]),
+                ]
+            );
+
+            return new Union([$type]);
+        }
+
+        if ($var_id === '$_SERVER' || $var_id === '$_ENV') {
+
+            $string_helper = Type::getNonEmptyString();
```
orklah, line 615: Should this be a Type::getString instead of NonEmpty?

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -531,26 +540,231 @@ public static function getGlobalType(string $var_id): Union
...
+            $string_helper->possibly_undefined = true;
+
+            $non_empty_string_helper = Type::getNonEmptyString();
+            $non_empty_string_helper->possibly_undefined = true;
+
+            $argv_helper = new Union([
```
orklah, line 621: You could reduce code here by just calling getGlobalType again for $argc/$argv

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -531,26 +540,231 @@ public static function getGlobalType(string $var_id): Union
...
+            ]);
+            $argv_helper->possibly_undefined = true;
+
+            $argc_helper = new Union([
+                new TIntRange(1, null)
+            ]);
+            $argc_helper->possibly_undefined = true;
+
+            $request_time_helper = new Union([
+                new TIntRange(time(), null)
```
orklah, line 632: I get the meaning here but I'm afraid users may find confusing the fact that they retrieve a seemingly random value like `1663105964` here, WDYT?

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -531,26 +540,231 @@ public static function getGlobalType(string $var_id): Union
...
+                'HTTP_SEC_FETCH_MODE'     => $non_empty_string_helper,
+                'HTTP_SEC_FETCH_SITE'     => $non_empty_string_helper,
+                'HTTP_SEC_CH_UA_PLATFORM' => $non_empty_string_helper,
+                'HTTP_SEC_CH_UA_MOBILE'   => $non_empty_string_helper,
+                'HTTP_SEC_CH_UA'          => $non_empty_string_helper,
             ]);
+
+            // generic case for all other elements
+            $detailed_type->previous_key_type = Type::getNonEmptyString();
+            $detailed_type->previous_value_type = Type::getString();
```
orklah, line 717: Are we sure it's always Strings? You documented a float above. Mixed could be more appropriate?

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -531,26 +540,231 @@ public static function getGlobalType(string $var_id): Union
...
+                    new TString(),
+                    new TNonEmptyList(Type::getString()),
+                ]),
+                'error' => new Union([
+                    new TInt(),
+                    new TNonEmptyList(Type::getInt()),
+                ]),
+            ];
+
+            $php_version = $config->getPhpVersionFromConfig();
```
orklah, line 746: Please use `$codebase->analysis_php_version_id` here. I believe the version in config can be overwritten by command line for example. The codebase holds the definite version

tests/Internal/CliUtilsTest.php
```diff
@@ -19,7 +19,7 @@ class CliUtilsTest extends TestCase
     protected function setUp(): void
     {
         global $argv;
-        $this->argv = $argv;
+        $this->argv = $argv ?? [];
```
orklah, line 22: This won't be necessary anymore once you set the flag in Union

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
AndrolGenhald: Maybe `scalar`? I haven't looked into how PHP populates `$_SERVER`, but I would _guess_ they would always be scalars.

orklah: note that scalar excludes null, so possibly scalar|null

AndrolGenhald:
[This](https://www.php.net/manual/en/reserved.variables.server.php#121008) might be something we'd want to consider.

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
AndrolGenhald: Many of these come from apache2/nginx (some of them from those servers themselves, some forwarded from the client), is it possible for those to set integer/float/whatever else values or is it forced to be a string?

AndrolGenhald:
> which is an issue with all globals anyway

If we don't have one already, we should open an issue to make superglobals immutable.

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
orklah: There's too much code where we can guarantee we're in CLI (I'm thinking about Commands in Symfony for example that are pieces of code only made to be ran on CLI). Adding CLI checks on that kind of code would be redundantly painful IMO.

I'm fine adding a config about this though in order to enable a stricter check for those

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
orklah: ok :)

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
orklah: Alright :)

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
orklah: We'll fix that if we get an example then, thanks for checking all that!

orklah:
> I think this is caused by having "possibly_undefined" and "ignore_nullable_issues" added as you suggested.

Yeah, it's probably due to the check for `InvalidPropertyAssignmentValue` considering that possibly_undefined could lead to null and not checking `ignore_nullable`. Can you coalesce with empty array here? It should solve the issue

orklah:
Thanks!

olleharstedt:
I think this PR broke our CI. See my latest bug here: https://github.com/vimeo/psalm/issues/8559

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -531,26 +540,239 @@ public static function getGlobalType(string $var_id): Union
...
+            ]);
+
+            // generic case for all other elements
+            $detailed_type->previous_key_type = Type::getNonEmptyString();
+            $detailed_type->previous_value_type = Type::getString();
+
+            return new Union([$detailed_type]);
+        }
+
+        if ($var_id === '$_FILES') {
```
olleharstedt, line 731: Probably incorrect, since first array access is form input name, not "name" or "type" etc.

olleharstedt: https://www.php.net/manual/en/features.file-upload.post-method.php

olleharstedt: You didn't add a unit test for this check?

PR #8478, opened 2022-09-12
`date_get_last_errors()`, `DateTime::getLastErrors()` may return false

gsteel:
Up to PHP 8.2, `date_get_last_errors()`, `DateTime::getLastErrors()` and `DateTimeImmutable::getLastErrors()` return false if no previous date operations have been performed. Currently, In PHP 8.2-RC, false is returned after a date operation that yields neither warnings nor errors:

https://3v4l.org/HBq0q
https://3v4l.org/3QsKY


boesing:
https://github.com/vimeo/psalm/pull/7758#issuecomment-1228577723

I guess thats related and thus this can be closed.

gsteel:
@boesing just educated me on `ignoreInternalFunctionFalseReturn` - closing!

AndrolGenhald:
Actually I think maybe you should reopen this. The example with my comment there showed that `@psalm-trace`ing the type shows the `false`, but errors involving that `false` type are ignored by default. If `ignoreInternalFunctionFalseReturn` is disabled though it should still give an error.

If we try the same thing with [`date_get_last_errors`](https://psalm.dev/r/393ea844d9), it doesn't show the `false` type, and if you try it locally with `ignoreInternalFunctionFalseReturn` disabled I bet it's not going to give an error like it should.

gsteel:
@AndrolGenhald - Re-opened. Happy to add some tests, but I could do with a little bit of guidance where these should go

orklah:
you're missing a change for date_get_last_errors is Callmap_historical it seems

gsteel:
Thanks @orklah

gsteel:
Rebased so cs tests pass 🤞

AndrolGenhald:
Thanks!

PR #8480, opened 2022-09-13
Make Psalter add `@throws` annotation with properly namespaced exception

### LINKED ISSUE #8467: Fixing `MissingThrowsDocblock` adds relative class names
The fix for `MissingThrowsDocblock` does not prepend the full class names of the exception(s) with a leading slash. The result of this is that **within a namespace, the class name in an added `@throws` annotation will be relative to the namespace**, probably resulting in an `UndefinedDocblockClass`. See https://psalm.dev/r/32c7b62a8d, which shows the result of fixing after initially no `@throws` annotation was available.

~~I haven’t seen a way to add an example of the fixer via psalm.dev~~ psalm.dev doesn't seem able to fix the `MissingThrowsDocblock` so I can't show an example of that, but this can be tested by adding an additional dataset to the data provider of `ThrowsBlockAdditionTest`. The dataset should look like:
```php
             'addThrowsAnnotationToFunctionInNamespace' => [
                '<?php
                    namespace Foo;
                    function foo(string $s): string {
                        if("" === $s) {
                            throw new \InvalidArgumentException();
                        }
                        return $s;
                    }',
                '<?php
                    namespace Foo;
                    /**
                     * @throws \InvalidArgumentException
                     */
                    function foo(string $s): string {
                        if("" === $s) {
                            throw new \InvalidArgumentException();
                        }
                        return $s;
                    }',
                '7.4',
                ['MissingThrowsDocblock'],
                true,
            ],
```

I fear there’s maybe an underlying challenge in regards to adding or detecting `use` statements. But I think it can be sufficient to simply add the fully qualified class name. Other tools like PHP_CodeSniffer can be used to further transform that into nice formatting, the goal here should be to ensure the `@throws` annotation is added with the right type.

d-claassen:
Fixes  #8467.

I've added the test provided in #8467, fixed that by prefixing the classname with a backslash and updated other tests.

I'm not certain if the fix for the exception class is now in the right place, and if there are possibly utilities to solve this. If there are, I'd be happy to apply them.

Running `$ composer tests` failed because of the same reasons for which the current main branch fails.

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
@@ -720,7 +720,14 @@ abstract class FunctionLikeAnalyzer extends SourceAnalyzer
             }
 
             if (!$is_expected) {
-                $missingThrowsDocblockErrors[] = $possibly_thrown_exception;
+                $missing_docblock_exception = new TNamedObject($possibly_thrown_exception);
+                $missingThrowsDocblockErrors[] = $missing_docblock_exception->toNamespacedString(
+                    $this->source->getNamespace(),
+                    $this->source->getAliasedClassesFlipped(),
+                    $this->source->getFQCLN(),
+                    true
+                );
+
                 foreach ($codelocations as $codelocation) {
                     // issues are suppressed in ThrowAnalyzer, CallAnalyzer, etc.
                     IssueBuffer::maybeAdd(
--- a/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
+++ b/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
@@ -405,6 +405,7 @@ class FunctionDocblockManipulator
             $inferredThrowsClause = array_reduce(
                 $this->throwsExceptions,
                 function (string $throwsClause, string $exception) {
+                    $exception = '\\' . $exception;
                     return $throwsClause === '' ? $exception : $throwsClause.'|'.$exception;
                 },
                 ''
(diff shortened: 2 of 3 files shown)


orklah:
I'm surprised it fixed the issue. Did you find what was the cause of the problem for it to be solved by adding a slash?

AndrolGenhald:
@orklah I haven't checked, but I'd assume it's currently adding `@throws` with the FQCN, but without the leading slash, so it's trying to resolve it from the current namespace instead of from the global namespace (could be wrong though). I'd prefer to fix it by using whatever namespace was used when `throw`ing instead of adding a slash, but I haven't looked too much into this, so I don't know how difficult it would be.

AndrolGenhald:
> This also adds the annotation for exceptions thrown by methods that are being called within the method and are not caught.

That makes sense, I hadn't considered that scenario.

> I can add a unit test to show this behavior!

Yeah, please do. That way if anyone ever tries to get it to use already-`use`d namespaces we'll have that test to make sure that part still works.

d-claassen:
The `FunctionDocblockManipulator` receives the missing `@throws` exception classes from the `FunctionLikeAnalyzer`. I looked higher up the original source, and the class names without a leading slash seem to come all the way up from `PhpParser\Node\Expr\New_`. Since that seems to be relied upon more, I concluded that the right place to create an exception class with leading slash is the `FunctionLikeAnalyzer` and provide the right class name to the `FunctionDocblockManipulator`.

I found a usage of the `TNamedObject::toNamespacedString` in the `FunctionLikeAnalyzer` that did what I also needed. It also takes into account the source, so I've added an additional test to show that it takes relevant `use` statement into account. After applying these changes and removing my original "fix", I could also revert some change to the other tests that didn't need the leading slash per se.

So the end result  that Psalter generates now is a lot nicer, which the tests show. 
I think that the code change that I added is on the right level now. 
I'm not sure if this the way to make that change, or if there are any common approaches I've missed.
I'm also not sure if I should use some property from the `TNamedObject` when adding the `MissingThrowsDocblock` to the `IssueBuffer`, instead of the full class name without the leading slash.

d-claassen:
@AndrolGenhald Done. Rebased & pushed

orklah:
Seems great! Thanks!

PR #8489, opened 2022-09-17
Fix array_key_exists first argument false positive

### LINKED ISSUE #7428: Weird property purity (?) issue
When using a property that is a union, `array_key_exists` assertions aren't parsed: https://psalm.dev/r/ceab5bdfcd
However, if that same property is first copied into a variable, the assertions are parsed correctly: https://psalm.dev/r/8e55452ecd

This somewhat reminds me of a purity issue, but all non-magic properties should be pure by default.

hirokinoue:
Fixes #7428


orklah:
Thanks!

PR #8490, opened 2022-09-17
Add null-type to several DOM-functions

tvdijen:
These functions accept `null` for their namespace-param to use the default namespace.


orklah:
Thanks for this! 

You must have corresponding changes in Callmap_historical too (or possibly a delta file if those signatures changed in time). The full explanation is here: https://psalm.dev/docs/contributing/editing_callmaps/

tvdijen:
Ah great, that was the info I was looking for! Fixed the historical CallMap as well!

orklah:
Thanks!

PR #8491, opened 2022-09-17
Fix array_column with object and column name null

### LINKED ISSUE #7929: New array_column inference doesnt handle indexing classes with index_key
The new array_column type inference introduced in https://github.com/vimeo/psalm/pull/7760 doesn't correctly handle passing an array of objects into it in combination with the index_key property.

When passing null to `column_key` the entire objects are returned again, but indexed by the `index_key` property. This case isn't currently handled property, instead the inferred value type is an array representation of the object, instead of the actual object.
See: https://psalm.dev/r/9f1e7ef328

The expected type is: `array<string, Fish>`

original issue: https://github.com/vimeo/psalm/issues/7750
introduced in: https://github.com/vimeo/psalm/pull/7760

HypeMC:
Fixes #7929


orklah:
Thanks!

PR #8492, opened 2022-09-17
Make template constraints examples in docs consistent

HypeMC:
Currently the examples in the documentation for template constraints vary between two syntaxes, the `as <type>` & the `of <type>`. Even though both syntaxes are valid & do the same thing, only the `of <type>` [one is documented](https://github.com/vimeo/psalm/blob/9ed9c4b64ca91654e39a2f00bd61002d5e7a506b/docs/annotating_code/templated_annotations.md?plain=1#L208) (and probably the preferred syntax), which can be very confusing for newcomers who are using psalm for the first time.

This PR aims to remove those inconsistencies by using only the `of <type>` syntax in the docs.

### DIFF
--- a/docs/annotating_code/templated_annotations.md
+++ b/docs/annotating_code/templated_annotations.md
@@ -229,7 +229,7 @@ Templated types aren't limited to key-value pairs, and you can re-use templates 
 ```php
 <?php
 /**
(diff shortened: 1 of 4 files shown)


orklah:
Thanks!

PR #8493, opened 2022-09-17
Document the object with properties syntax

HypeMC:
Currently the object with properties syntax is not documented anywhere.


orklah:
Thanks!

PR #8494, opened 2022-09-18
Fix type widening when `Any` assertion is used

### LINKED ISSUE #8084: False positives when using isset($array[$key]) === false instead of !isset($array[$key])
We have long ago adopted a code style that enforces an explicit `=== false` instead of a `!` negation within `if` conditions. I noticed that this sometimes leads to false positives when using an `isset()` check for a key within an array, while Psalm does not find issues on the equivalent version with a `!` instead of `=== false`.

One that I've been able to reproduce in isolation:
* https://psalm.dev/r/6ae508f44e (with `isset(...) === false`, gives issues)
* https://psalm.dev/r/5681b1b89c (with `!isset(...)`, no issues)

I also see some `RedundantPropertyInitializationCheck` issues on `if (isset($this->array[$key]) === false)` patterns on our code base (like Psalm does not understand that we're checking if a specific element is set rather than if the whole array is initialized), which disappear when I switch `=== false` to `!`. I haven't been able to reproduce these in an isolated case though, and hope these maybe also disappear if the above false positive is solved.

pvandommelen:
Fixes #8084

Also contains a related change to the reconcilers fast path where I believe an existing check is unnecessary.

PR targets the master branch. I have been unable to backport this to 4.x, the work done in the master branch makes this issue a lot easier to resolve.

### DIFF
--- a/src/Psalm/Internal/Type/AssertionReconciler.php
+++ b/src/Psalm/Internal/Type/AssertionReconciler.php
@@ -293,6 +293,10 @@ class AssertionReconciler extends Reconciler
 
         $old_var_type_string = $existing_var_type->getId();
 
+        if ($new_type_part instanceof TMixed) {
+            return $existing_var_type;
+        }
+
         $new_type_has_interface = false;
 
         if ($new_type_part->isObjectType()) {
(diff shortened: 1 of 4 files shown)


orklah:
Thanks! The rework of assertion Matt did is indeed helpful for this kind of changes :)

Issue #8457, opened 2022-09-04
InvalidReturnType are reported for stubs


kkmuffme:
I have a project where a function exists both in the stubs AND in the code to analyze (basically: overwriting a function/method, but also for cases where psalm is unable to infer/load files due to how paths are loaded)

In my file I have this (just a mock example):
https://psalm.dev/r/e7f20deb65

Problem:
as this function exists in my stubs too, Psalm now reports this error for the stubs file - which gets suppressed (since errors in stubs file paths are suppressed by psalm by default)
Therefore this error never shows up.

Why this is a problem?
Psalm always takes the phpdoc from the stubs as truth (this is fine/good). It then uses the code (not stubs) to check if the code matches the phpdoc.
In this case however, it would make much more sense to report the error not for the stubs but for the file where it checks the code in.

I remember I had a similar issue a long time ago, when I got started with psalm, so I think this is not an isolated issue with InvalidReturnType

EDIT: this affects all (but only) PHPDoc issues for functions. While most of them get reported for both the stubs + the file the issue is in, InvalidReturnType and MissingParamType get only reported for the stubs.
But e.g. InvalidDocblock is reported for both (stubs and the file)

AndrolGenhald:
Just to make sure I understand the issue, you're saying because `foo()` is stubbed with `@return never`, the actual `foo()` declaration in your code has `InvalidReturnType` suppressed?

AndrolGenhald:
I think the proper way to fix this would be to re-parse the docblock and return signature when analyzing a function, and use that instead of whatever exists in storage. That way if the function declaration conflicts with the stub, the function itself will still be analyzed correctly, but the rest of the codebase will assume the function works as declared in the stub.

AndrolGenhald:
:shrug: I just picked one because it's not clear to me if it was originally intentional behavior.

PR #8503, opened 2022-09-22
use cache for declared function when available before falling back to stubs

kkmuffme:
fixes return type issues reported for the wrong file

Fix https://github.com/vimeo/psalm/issues/8457


orklah:
Thanks!

AndrolGenhald:
Oh nice! I expected it to be a bit harder to fix than that.

PR #8513, opened 2022-09-26
Daft: Fix DateTime construct named argument call

magnetik:


dictionaries/CallMap.php
```diff
@@ -1772,8 +1772,8 @@
 'DatePeriod::getDateInterval' => ['DateInterval'],
 'DatePeriod::getEndDate' => ['?DateTimeInterface'],
 'DatePeriod::getStartDate' => ['DateTimeInterface'],
-'DateTime::__construct' => ['void', 'time='=>'string'],
-'DateTime::__construct\'1' => ['void', 'time'=>'?string', 'timezone'=>'?DateTimeZone'],
+'DateTime::__construct' => ['void', 'time'=>'?string', 'timezone'=>'?DateTimeZone'],
+'DateTime::__construct\'1' => ['void', 'time='=>'string'],
```
AndrolGenhald, line 1776: It looks to me like we should only have one definition with `['void', 'time='=>'string', 'timezone='=>'?DateTimeZone']`.

orklah: I think that at one point, it was considered ok to do https://3v4l.org/2a91R

but seems like it's now deprecated and it does the same as putting an empty string, I'm inclined to agree. It will simplify things while getting rid of this bug too

weirdan:

orklah:

PR #8530, opened 2022-10-04
Add int type aliases based on existing codes

hamburnyog:


orklah:
Hey! Thanks for this work!

Aliases are fine, what we're deprecating is not `positive-int` but the type `TPositiveInt`. It simply means that any `negative-int` will just be an alias for `int<min, -1>` so your `TNegativeInt` is not needed as it will be replaced with a `new TIntRange(null, -1);`.

So if you want to make the change, your PR is welcome :)

(Note: the removal of TPositiveInt and migration to TIntRange was done on the master branch but you can still add the aliases on 4.x to be able to use them before the Psalm 5 release)

hamburnyog:
I see 👍, I made some changes. Hopefully I get it this time. Thanks for your response!

orklah:
Almost perfect :)

There's a slight difference between `negative-int` and `non-positive-int` because 0 is neither positive or negative.

Which means negative-int won't contain 0 but non-positive-int does !

The same rule applies between `positive-int` and `non-negative-int`

hamburnyog:
I've made some updates, kindly check if you have free time. Thanks!

orklah:
Thanks!

cc @ondrejmirtes we added some aliases for parity with positive-int, not sure if those already exists in phpstan

PR #8541, opened 2022-10-06
Add composer lockfile to git for reproducible builds

kpcyrd:
hi!

I'm looking into packaging psalm for [Arch Linux](https://archlinux.org/) and attempting to build the phar from source. I noticed running `composer install` after unpacking the tarball prints this error:

```
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
```

This means the dependency gets resolved at build time, this means the phar is going to change over time, making verification of the compiled phar difficult. This patch resolves the current dependency graph and commits it to git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 /.phpunit.result.cache
 .*.swp
 .*.swo
-/composer.lock
 /phpunit.xml
 /vendor-bin/*/composer.lock
 /vendor-bin/*/vendor/
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,4432 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "This file is @generated automatically"
+    ],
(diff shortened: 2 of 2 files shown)


orklah:
Do you know Psalm actually produce a phar already?

https://github.com/psalm/phar

Or maybe there's something special that needs to be done for Arch Linux?

All things equal, I'd prefer to keep the current config, upgrading the lock file on each release is a pain

weirdan:
You can see how many issues this causes just by looking at CI results.

@kpcyrd are you building tagged releases?

weirdan:
Thinking further, a specific dependency graph is as much a function of the build environment (which should roughly match runtime environment) as it is of the source code. E.g. `psalm.phar` built on PHP 7.4 does necessarily have different dependency versions compared to the one built on 8.1. We as a vendor do our part by supplying `composer.json`, but the final dependency resolution happens during the build. 

Therefore if you need reproducible builds, you need to store the `composer.lock` file (generated during the build) as part of the package itself, alongside the build env definition. From quick google search it seems `composer.lock` needs to be referenced in `source` array of the `PKGBUILD` file, and the build env definition is stored in the `.BUILDINFO` file.

Issue #8294, opened 2022-07-20
Invalid class constant name


AndrolGenhald:
https://3v4l.org/T94Rc https://psalm.dev/r/6d865edf9e

We should show an issue for this.

PR #8542, opened 2022-10-07
Add check for class const with reserved word 'class'

gphargreaves:
First contribution so hopefully my changes are in sensible places 😁 Addresses issue #8294 

Had a read through the docs for the a structure of the analyzers and hopefully this change is fit for purpose, happy to tweak accordingly if you have any feedback.

### DIFF
--- a/tests/ClassTest.php
+++ b/tests/ClassTest.php
@@ -925,6 +925,18 @@ class ClassTest extends TestCase
                 ',
                 'error_message' => 'InvalidTraversableImplementation',
             ],
+            'cannotNameClassConstantClass' => [
+                '<?php
+                class Foo
+                {
+                    /** @var class-string<Bar> */
+                    protected const CLASS = Bar::class;
+                }
+                
+                class Bar {}
+                ',
+                'error_message' => 'ReservedWord',
+            ]
         ];
     }
 }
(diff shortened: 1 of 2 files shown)


tests/ClassTest.php
```diff
@@ -1004,6 +1004,18 @@ public function valid(): bool {
                 ',
                 'error_message' => 'MissingTemplateParam',
             ],
+            'cannotNameClassConstantClass' => [
+                '<?php
```
orklah, line 1008: this should have an array offset `'code' =>` to respect the format of tests if you're targeting master

orklah:
Looks good!

Note that you targeted master, which is the branch for Psalm 5 which is scheduled for release in a few weeks if all works out.

If you want to be sure to have this change on Psalm 4, you could target the 4.x branch, your choice :)

AndrolGenhald:
LGTM, not sure what's up with CI though. I think I'd prefer this go in `ClassConstAnalyzer`, but since it targets 4.x this spot is fine (that whole loop should probably be moved to `ClassConstAnalyzer` in 5.x just to keep things better organized).

AndrolGenhald (approved):

orklah:
Thanks!

CI often go nuts when retargeting PRs, I think it reverts back to some old version of CI from somewhere

PR #8553, opened 2022-10-09
Update dictionaries for hash ext functions

lptn:
from the 8.0, `hash()` and `hash_hmac()` never return `false`; only `hash_file()` and
`hash_hmac_file()` return `false` in case the data cannot be read.

see https://github.com/cmb69/php-src/commit/8982ef3321c20b532f7f45d031586a3615a7e4e8 and https://github.com/php/php-src/issues/7759

PHP 8.1 hash module function declarations: https://github.com/php/php-src/blob/php-8.1.11/ext/hash/hash.stub.php


orklah:
Nice! Thanks!

PR #8555, opened 2022-10-09
Update Phpredis stubs to return false on failure

kkmuffme:
Started in https://github.com/phpredis/phpredis/pull/2120#issuecomment-1166644919 - this PR adds false in additional methods

Makes mGet return type more detailed

Force values to be string, to avoid implicit cast creating hard to track bugs https://github.com/phpredis/phpredis/pull/2015#issuecomment-1139509679

### DIFF
--- a/stubs/phpredis.phpstub
+++ b/stubs/phpredis.phpstub
@@ -31,7 +31,7 @@ class Redis {
      */
     public function acl(string $subcmd, ...$args);
 
-	/** @return int|Redis */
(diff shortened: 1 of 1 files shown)


orklah:
Didn't we had the official stub since last PR? 

Is it incomplete still? Maybe it should be updated on their end first?

orklah:
ok fine but it's something we may have to revert if some user complain that some function can't return false and we can't find a case when it does

PR #8566, opened 2022-10-11
Fix callmap for DOMNode methods

tvdijen:


orklah:
Hey! Can you rebase? The CI crashed somehow.

Would you have an example for lookupNamespaceUri taking null and returning null on strict_types? Seems different than the official doc

tvdijen:
The docs are a total mess. For some reason they won't merge doc updates, so I gave up on that a long time ago.

Let me point you to the DOM specs:
https://www.w3.org/TR/DOM-Level-3-Core/core.html#Node3-lookupNamespaceURI

As well as the PHP source:
https://github.com/php/php-src/blob/master/ext/dom/node.c#L1513-L1518

Passing `null` to `lookupNamespaceURI` means to lookup the default namespace.
If no default namespace exists, this method will return `null`.

For `lookupPrefix`, if it cannot be found the method will return `null`.

orklah:
That explanation will do just fine :) Thanks for the fix!

PR #8567, opened 2022-10-11
Fix MinMaxReturnTypeProvider when handling TDependentListKeys

### LINKED ISSUE #8514: `TDependentListKey` in `MinMaxReturnTypeProvider`
https://psalm.dev/r/b1d48c0c66

`MinMaxReturnTypeProvider`, when checking indexes found in a `fgetcsv()` result, throws "Unexpected type"

https://github.com/vimeo/psalm/blob/36d5a2a83c1d7b3e2aca1cfe9e9dcc916619e962/src/Psalm/Internal/Provider/ReturnTypeProvider/MinMaxReturnTypeProvider.php#L75 

This test, after checking for some specific Int cases, should fallback back to `instanceof TInt`, rather than `get_class() === TInt::class`, eg:

```
} elseif ($atomic_type instanceof TInt) {
``` 
is desired.

Happy to submit a PR if this change makes sense.

nosnickid:
Closes #8514


orklah:
Thanks!

PR #8577, opened 2022-10-13
fix mapped property type for DOMNode childNodes

### LINKED ISSUE #8576: DOMNode childNodes expecting a DOMNodeList of DomNode instead of DOMNode
If you loop through each of the items in a `childNodes` list as `DOMNode`, like in the following code:

```php
$dom = new \DOMDocument();
$node = new \DOMNode();

echo implode('', array_map(
    fn (\DOMNode $node): string => trim($dom->saveHTML($node)),
    iterator_to_array($node->childNodes)
));
```

Psalm will return the following error:

```
InvalidClass - Class, interface or enum DomNode has wrong casing (see https://psalm.dev/007)
```

This is happening on `v4.8.1` due to [wrong casing on the PropertiesMap for `domnode`](https://github.com/vimeo/psalm/blob/d626d2424177591ba8fb173b2597ab47bdf2fa2d/dictionaries/PropertyMap.php#L376) that should be `DOMNodeList<DOMNode>` instead of `DomNodeList<DomNode>`.

This is the commit that introduced the bug: https://github.com/vimeo/psalm/pull/5403

I also tried to replicate the issue using psalm.dev but was unable to: https://psalm.dev/r/0957fc5f47
Maybe it depends on the case sensitivity of the system running psalm.

cristianobaptista:
Fixes #8576

### DIFF
--- a/dictionaries/PropertyMap.php
+++ b/dictionaries/PropertyMap.php
@@ -164,7 +164,7 @@ return [
     'domnode' => [
         'attributes' => 'DOMNamedNodeMap|null',
         'baseuri' => 'string|null',
-        'childnodes' => 'DomNodeList<DomNode>',
+        'childnodes' => 'DOMNodeList<DOMNode>',
         'firstchild' => 'DOMNode|null',
         'lastchild' => 'DOMNode|null',
         'localname' => 'string|null',


orklah:
Thanks!

PR #8579, opened 2022-10-13
Fix template param for SplDoublyLinkedList

ADmad:
It's key is always an integer.


orklah:
Hey! 

There's a few failures on CI because you're still using TKey in methods.

But aside from that, do you have any control over the keys in a SplDoublyLinkedList?

For example, can I decide to have one with only negative keys? Because if I can, then you don't want to remove TKey altogether, you just want to have an upper bound (like `@template Tkey as int`) instead of removing it altogether. Because if you leave it, someone could do a `SplDoublyLinkedList<int<-5, 5>, string>` for example

orklah:
Ok :)

It could be interesting to document `int<0, max>` then to describe that negative int can't be returned.

Could you target master? It will prevent people having to change their templates on a minor version

ADmad:
Re-targeted to master.

orklah:
Thanks!

PR #8583, opened 2022-10-14
Correct return type of DateTimeImmutable sub method stub

### LINKED ISSUE #8484: Fix DateTimeImmutable::sub() to not return false
@Ocramius The `->sub()` issue is separate from this, this issue is only for `->modify()`, but I do believe `->sub()` was incorrect in your PR, here's an example for that special case you found: https://3v4l.org/TIWRB

It indeed gives a warning and returns a clone, so there's no `false` return.

Also, go get some sleep :stuck_out_tongue:

_Originally posted by @AndrolGenhald in https://github.com/vimeo/psalm/issues/8454#issuecomment-1238386327_

gphargreaves:
Fixes #8484 

This tiny tweak corrects the return type of the DateTimeImmutable stub as it doesn't return false.

As seen [here](https://github.com/php/php-src/blob/534127d3b22b193ffb9511c4447584f0d2bd4e24/ext/date/php_date.c#L3157-L3160) sub never returns false and only a clone as mentioned by @AndrolGenhald

### DIFF
--- a/stubs/CoreImmutableClasses.phpstub
+++ b/stubs/CoreImmutableClasses.phpstub
@@ -58,8 +58,7 @@ class DateTimeImmutable implements DateTimeInterface
 
     /**
      * @psalm-mutation-free
-     * @return static|false this method can fail in case an {@see DateInterval} with relative
-     *                      week days is passed in.
+     * @return static
      *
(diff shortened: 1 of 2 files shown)


orklah:
Thanks!

PR #8584, opened 2022-10-14
Enhance type detection for internal php functions `key`, `current`, `end` and `reset`

### LINKED ISSUE #8521: Incorrect type inference when using `key`/`current`/`reset` on non-empty array
Hey there,

when both key and value types of an array are known, I'd expect psalm to be able to infer the return value of `key`/`current` in case of a `non-empty-array` or `non-empty-list`.

https://psalm.dev/r/bc365e8080

Not sure if this can already be done with some conditional magic (even tho, I did never understood how these internal functions are typed 😅).

But something like this might work, right?
```php
/**
  * @template T of array-key
  * @param array<T,mixed> $array
  * @return ($array is non-empty-list ? 0|positive-int : ($array is non-empty-array ? T : mixed))
  */
function key(array $array): mixed
{
}

/**
  * @template T
  * @param array<mixed,T> $array
  * @return ($array is non-empty-list ? T : ($array is non-empty-array ? T : T|false))
  */
function current(array $array): mixed
{}

/**
  * @template T
  * @param array<mixed,T> $array
  * @return ($array is non-empty-list ? T : ($array is non-empty-array ? T : T|false))
  */
function reset(array $array): mixed
{}
```

boesing:
Fixes #8521

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -136,11 +136,51 @@ function array_flip(array $array)
  *
  * @param TArray $array
  *
- * @return (TArray is array<empty, empty> ? null : TKey|null)
+ * @return (TArray is array<empty,empty> ? null : (TArray is non-empty-list ? int<0,max> : (TArray is non-empty-array ? TKey : TKey|null)))
  * @psalm-pure
  * @psalm-ignore-nullable-return
  */
 function key($array)
+{
+}
+
+/**
+ * @psalm-template TKey as array-key
+ * @psalm-template TValue
+ * @psalm-template TArray as array<TKey, TValue>
+ *
+ * @param TArray $array
+ *
+ * @return (TArray is array<empty,empty> ? false : (TArray is non-empty-array ? TValue : TValue|false))
+ * @psalm-pure
+ */
+function current($array)
+{
+}
+
+/**
+ * @psalm-template TKey as array-key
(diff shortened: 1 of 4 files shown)


tests/ArrayFunctionCallTest.php
```diff
@@ -9,7 +9,7 @@
 
 class ArrayFunctionCallTest extends TestCase
 {
-    use InvalidCodeAnalysisTestTrait;
+//    use InvalidCodeAnalysisTestTrait;
```
orklah, line 12: this should be restored :)

stubs/CoreGenericFunctions.phpstub
```diff
@@ -136,14 +136,54 @@ function array_flip(array $array)
  *
  * @param TArray $array
  *
- * @return (TArray is array<empty, empty> ? null : TKey|null)
+ * @return (TArray is array<empty,empty> ? null : (TArray is non-empty-list ? int<0,max> : (TArray is non-empty-array ? TKey : mixed)))
```
orklah, line 139: Mixed should be replaced by TKey|null

You excluded the empty array and the non-empty-array but it doesn't mean there's nothing left. You still have the simple array type where you don't know if there is content in the array. And in this case you still can't say if it will return null or TKey

boesing:
Fixed all issues.

orklah:
Seems good!

PR #8585, opened 2022-10-14
Add additional checks for concat of non-empty strings to return non-falsy

### LINKED ISSUE #8315: concat of 2 non-empty-strings should lead to non-falsy-string
https://psalm.dev/r/c77d73090f

Implemented in phpstan with https://github.com/phpstan/phpstan-src/pull/1542

gphargreaves:
Fixes #8315 

This change adds some additional checks so:

- Concat of two non-empty-string types returns a non-falsy-string
- Concat of a non-empty-string and a numeric-string returns a non-falsy-string


orklah:
Thanks!

PR #8591, opened 2022-10-16
Merge 4.x

danog:


danog:
It was precisely this commit from master: https://github.com/vimeo/psalm/commit/525a86e19d3f518bb4e964eb405eda106dfc5fa9
The problem is that the new logic from 4.x creates an array of atomic types (including the TBool/TTrue/TFalse logic introduced on master separately) which are then combined with the TypeCombiner.

The TypeCombiner has a very generic logic which joins literals with integer ranges by always extending the lower/higher bound, even for non-adjacent values (ie `0` and `int<3, max>` => `int<0, max>`).

This should be fixed separately by improving the literal+range and range+range combining logic directly inside TypeCombiner to generate multiple distinct disjoin ranges and literals, and I actually started doing that, but then noticed that some parts of psalm like offset detection misbehaved, so I postponed this for a later PR.

orklah:
Thanks a lot!

Issue #8593, opened 2022-10-17
InvalidArgument doesn't work for callable args count


kkmuffme:
https://psalm.dev/r/1b69254ca2

It correctly complains about InvalidArgument when the type is wrong.
But it should also give an error when I only pass 1 param, but the callable expects 2

vudaltsov:
I don't agree that this should have been fixed. It is valid to pass a callable with fewer parameters.

vudaltsov:
See https://3v4l.org/bCe3Q.

vudaltsov:
It's wrong only when passing a built-in function https://3v4l.org/GT86Q.

orklah:
I'm open to arguments on that, but saying that php doesn't crash is not the most convincing.

Do you have use cases that require you to call the callback with more param than it requires? 

Then it's a balance between:
- I'm a user that wants to call a callable with more params than it declares
- I'm a user that wants to be warned when I make a mistake by passing a callback that receives less than I passed

weirdan:
> Do you have use cases that require you to call the callback with more param than it requires?

Yes, it's very common. E.g. a lot of methods accepting `Closure(K,V)` here: https://github.com/doctrine/collections/blob/2.0.x/src/ArrayCollection.php can actually be called with closures accepting `K` only if we don't care about the values.

orklah:
Not sure how to proceed then. Should we keep it to help user find legit mistakes or should we remove the check for functions outside Callmap not to bother more advanced usages? Maybe a config?

danog:
+1 for keeping it as it is, you can always provide an alternative callable in a union if needed: `callable(A, B): int|callable(A): int`

vudaltsov:
> you can always provide an alternative callable

No, you can't, because it will cause a `TwoManyArguments` error: https://psalm.dev/r/b5116d9026.

orklah:
@vudaltsov Would a new CallableTooManyArguments emitted especially for those cases that you could suppress for your whole project work for you?

weirdan:
That would work I suppose. Or an alternative syntax (e.g. `strict-callable()`), but I would rather avoid that as it will result in a combinatorial explosion if we go that way (`strict-pure-callable` etc).

danog:
Ooo, I love the `strict-callable` idea :D

vudaltsov:
I don't think that `strict-callable` is a good idea, because it moves the responsibility towards the one who designs the contract, while the actual problem belongs to the user of the contract.

[...]

PR #8594, opened 2022-10-17
ensure callbacks have the required number of params

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/8593


orklah:
Can you add some tests to check how it reacts to variadics in callable's params and array destructuring on the call?

orklah:
Thanks!

Issue #8575, opened 2022-10-13
Implode does not complain about a list of non-stringable objects


vudaltsov:
https://psalm.dev/r/53090c73ee
https://3v4l.org/lQ7Zv

orklah:
yeah, clearly bugged: https://psalm.dev/r/8233ca5c0a

weirdan:
> could you PR this? Should be an easy fix, since you only need to adjust the callmap to be more strict than generic "array"

As mentioned in the linked PR, there's no problem with the `implode()` signature (which comes from stubs, not the callmap). It's Psalm's handling of `stringable-object` type (used by the stubs) that is buggy.

PR #8595, opened 2022-10-17
#8575 More strict signature for implode function

SLaVeZ:
Fix for issue #8575


vudaltsov (approved):

orklah:
Unfortunately, this is not correct.

Stringable is a symbol that was added on PHP 8. I'm not exactly sure how Psalm would react exactly on lower versions (most probably allow a Stringable class that would exist in user code or possibly crash outright) but it can't be good.

There is a `stringable-object` pseudo-type that is translated into an object that happens to have a __toString method in Psalm. It would be more correct in this situation but Psalm doesn't seem to enforce it: https://psalm.dev/r/8233ca5c0a

This is a bug that should be fixed first. 

Also, the actual fix should be located in https://github.com/vimeo/psalm/blob/285740a753cc040ad781e547a76c7f1be1262196/stubs/CoreGenericFunctions.phpstub#L607 instead of callmaps. When a stub is present, it overrides the callmap

weirdan:
This is fixed in #8688

Issue #8178, opened 2022-06-27
InvalidReturnType due to void and null not treated the same


kkmuffme:
https://psalm.dev/r/f4bb991d20

void and null must be treated as the same thing.

AndrolGenhald:
Hmmm, I don't think it actually makes sense to have a return type of `void|never`. Regardless though, that error is incredibly misleading.

orklah:
> Hmmm, I don't think it actually makes sense to have a return type of `void|never`. Regardless though, that error is incredibly misleading.

It can make sense for assertions methods where you either throw or do nothing. but the snippet in example doesn't do that. So yeah, it's just a wrong phrasing

AndrolGenhald:
> It can make sense for assertions methods where you either throw or do nothing.

Shouldn't that just combine to `void` though? `never` is the bottom type, so it's a subtype of all other types, so if you do `x|never` it should just simplify to `x`.

orklah:
True. The case I had some times ago was something like `($die_on_redirect is true ? never : void)`. So technically it's the same case but the combination doesn't happen

AndrolGenhald:
> The case I had some times ago was something like `($die_on_redirect is true ? never : void)`.

Ahh, yeah I could see that then. Interestingly though, [this](https://psalm.dev/r/8dad765588) actually works fine while [this](https://psalm.dev/r/d162606c86) doesn't.

PR #8597, opened 2022-10-17
treat void and null the same for return type

kkmuffme:
treat void and null the same for return type, since TypeCombiner treats them equally already (combines e.g. "void|never", "void|null",... to "null")

Fix: https://github.com/vimeo/psalm/issues/8178


orklah:
I'm confused about what you're trying to fix here

void and null are two different concepts that are not interchangeable, even in php core: https://3v4l.org/93Z5W

Psalm will see this function as returning void|null: https://3v4l.org/MfrIM but that doesn't makes sense as a whole so it's coherent that we're combining those two into null, but it's not something we can apply everywhere

Issue #8359, opened 2022-08-02
Global type not respected during taint analysis


mmcev106:
With the following `psalm.xml`:

```xml
<?xml version="1.0"?>
<psalm>
	<projectFiles>
		<directory name="." />
	</projectFiles>
	<globals>
		<var name="connection" type="mysqli" />
	</globals>
</psalm>
```

...and the following `file.php`:
```php
<?php
$connection->query($_GET['injection']);
// function a(mysqli $b){}
```

A taint is not reported, even though it should be.  Strangely, if you uncomment the `function a...` line, the taint is reported correctly.  It's like psalm doesn't become aware of the `mysqli` type unless it is referenced in an unrelated location.

This example was distilled from an actual use case in a much larger code base.  Let me know if there is any more information I can provide, or if you have a hunch that it's related to a particular part of the psalm source that I could look into.

orklah:
Good catch.

I have a small guess that there's a mistake here:
https://github.com/vimeo/psalm/blob/f9450656e1666fbf6ceb8b419b2d02d065fb76b5/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php#L155

This class is in charge of setting the type for a given variable we just read. For example when Psalm sees `$a = $b;` it will decompose that in a VariableFetch for $b and then an assignment in $a.

The specific line I linked is in charge of handling the case when we have a read on a global variable (and it taints it if necessary).
However, it only does that if the variable name is recognized by `self::isSuperGlobal()` which only checks hard coded global variables: https://github.com/vimeo/psalm/blob/f9450656e1666fbf6ceb8b419b2d02d065fb76b5/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php#L37

This is not optimal because the `<global>` tag in config extends this list in theory (you can see that on VariableFetchAnalyzer::getGlobalType(), it's much more extended)

So, I think we need a VariableFetchAnalyzer::isGlobalVariable() which checks if the name is a known global (the same way getGlobalTypes checks it) and use this function instead of isSuperGlobal.

I think isSuperGlobal would be ultimately useless after being replaced by isGlobalVariable but it would need a few tests to check that

Are you up to try? :)

mmcev106:
Thanks very much for the info.  It might be a while before I can prioritize it, but I would like to try!  It won't hurt my feelings if anyone beats me to it though.

PR #8598, opened 2022-10-17
Partial attempt to fix global type taint issue 8359

mmcev106:
This PR to fix issue #8359 is still in the works!

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
@@ -160,7 +160,7 @@ class VariableFetchAnalyzer
             return true;
         }
 
-        if (is_string($stmt->name) && self::isSuperGlobal('$' . $stmt->name)) {
+        if (is_string($stmt->name) && self::isGlobalVariable('$' . $stmt->name, $codebase->analysis_php_version_id)) {
             $var_name = '$' . $stmt->name;
 
             if (isset($context->vars_in_scope[$var_name])) {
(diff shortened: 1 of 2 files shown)


mmcev106:
@orklah, I believe this PR implements your recommendations from [here](https://github.com/vimeo/psalm/issues/8359#issuecomment-1203132090).  I've verified that `VariableFetchAnalyzer::taintVariable()` is getting called with a `$type` argument of `mysqli`, but the unit test is still failing to detect the taint.  Do you have any other recommendations?

orklah:
I don't have any more idea as to why the taint is not detected here. About failing tests, I suppose this broke this condition:
https://github.com/vimeo/psalm/blob/58e87fa5e0cb8c76d8dd8ae4423342a47b35d403/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php#L910
It was using isSuperGlobal as well and it seems to be the one that breaks.

weirdan:

mmcev106:
@weirdan, I'm happy to rewrite the PR for the master branch.  However, I haven't yet figured out what to do differently yet to make the PR work as @orklah originally envisioned.

To follow up on the test failure comments, I think @orklah & I were talking about different tests.  I have yet to make the test case I added for this PR pass (regardless of other tests/cases)

orklah:
Sorry for not getting back to you before. I'm not sure why everything broke, it would have to be debugged step by step to find out. It's even possible my solution can't work for some reason.

I'll close this for now given it targets a closed branch. The issues will be kept open for tracking the issue though

mmcev106:
@orklah, thanks very much for the input.  There's a chance I will have a bigger block of time to dig into https://github.com/vimeo/psalm/issues/8359 in the next few months.

Issue #8175, opened 2022-06-26
Missing InvalidReturnType check when exit;


kkmuffme:
https://psalm.dev/r/bd83421ff4

should have an `InvalidReturnType` error, just like when not calling a `never` function:
https://psalm.dev/r/5a71513323

Error message should be something like this I guess:
>The declared return type 'string' for a is incorrect, got never

kkmuffme:
It seems like this is a more general issue when a function calls exit/die - the return type never gets validated then:
https://psalm.dev/r/ffbb76add8

AndrolGenhald:
This is going to require some special casing. A function with `@return string` that always exits is technically correct because `never` is a subtype of `string` (`never` is a subtype of every type). I think if we add a warning for this it should only show up if the function can only return `never`, [this](https://psalm.dev/r/044884a3e1) should still be fine.

kkmuffme:
>never is a subtype of every type

Maybe add a option to make "never" it's own type?

>[this](https://psalm.dev/r/044884a3e1) should still be fine.

I'd expect to get an error here too, since @return should be string|never

>if we add a warning for this that'll also show up for [this](https://psalm.dev/r/1a2862d437), which I'm not sure is desirable.

Well, technically this would be correct, since this @return never

AndrolGenhald:
> Maybe add a option to make "never" it's own type?

That's not really possible, it would fundamentally break the type system.

I'm still not really convinced this is worth it, I can see that it would be helpful in certain instances, but the way it works right now is totally correct as well. A function that never returns can have any return type, because it's guaranteed that every value returned by the function will satisfy any type (because, obviously, it never returns). We'll have to think about it and see what's actually feasible here.

kkmuffme:
Issue is also with UnevaluatedCode error. Same code, just in one case I @return never explicitly:
No issues when @return string: https://psalm.dev/r/8a1cdb9164
UnevaluatedCode issue when @return never: https://psalm.dev/r/8a1cdb9164

AndrolGenhald:
@kkmuffme You did the same link twice there.

> Issue is also with UnevaluatedCode error.

You could compare that to stuff like [this](https://psalm.dev/r/98f994b9ea) where it's redundant but not detected unless you [narrow the return type](https://psalm.dev/r/ab652125d6). I think it's totally correct not to warn that a return type can be narrower, because often you want a wider return type to give yourself more freedom when changing the implementation in the future.

orklah:
[...]

PR #8599, opened 2022-10-17
check "never" return type more strictly

kkmuffme:
* require explicit "never" return type when function always exits
* error if function does not exit, but return type explicitly contains "never"
* Fix: https://github.com/vimeo/psalm/issues/8175
* Fix: https://github.com/vimeo/psalm/issues/8178


orklah:
I'm not sure I followed the actual changes yet. However, a few points first:
- I'll need tests for that
- This will have to be pushed to master for two reasons: we removed TEmpty and we made the Union immutable. This PR will break both. (I'll start to refuse almost any PR to 4.x soon anyway)

danog:
You can already retarget it on master right now :)

PR #8609, opened 2022-10-22
Add dirname return type provider

### LINKED ISSUE #7344: `UnresolvableInclude` when using variable
It's OK:

```php
require_once dirname(__DIR__) . '/autoload.php';
```

It's `ERROR: UnresolvableInclude`:

```php
$root = dirname(__DIR__);
require_once $root . '/autoload.php';
```

mcaskill:
Resolves #7344

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/DirnameReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/DirnameReturnTypeProvider.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Internal\Provider\ReturnTypeProvider;
+
+use PhpParser;
+use Psalm\Internal\Analyzer\Statements\Expression\IncludeAnalyzer;
+use Psalm\Plugin\EventHandler\Event\FunctionReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\FunctionReturnTypeProviderInterface;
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/Provider/ReturnTypeProvider/DirnameReturnTypeProvider.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $call_args = $event->getCallArgs();
+        if (count($call_args) === 0) {
+            return null;
+        }
+
+        $dir_level = 1;
+
+        if (isset($call_args[1])) {
+            if ($call_args[1]->value instanceof PhpParser\Node\Scalar\LNumber) {
```
danog, line 37: Please extract this literal value from the node type provider, not directly from the statement

danog (changes requested):

danog:
Also do you think you could target this on `master` and switching to psalm `dev-master` in your workflow?
v5 should be released soon after I submit my final immutability PR, and I wouldn't like to deal with merging too many commits from the deprecated `4.x` branch :)

src/Psalm/Internal/Provider/ReturnTypeProvider/DirnameReturnTypeProvider.php
orklah: True :) This will allow Psalm to infer the number even if it comes from a variable, a function call or any other expression

mcaskill:
Replaced by #8611

PR #8611, opened 2022-10-22
Add dirname return type provider

mcaskill:
Resolves #7344

Replaces #8609:

* [Targets Psalm v5](https://github.com/vimeo/psalm/pull/8609#issuecomment-1287668265) (`master`)
* [Fixes handling of call arguments](https://github.com/vimeo/psalm/pull/8609#discussion_r1002386040)
* [Fixes unit tests](https://github.com/vimeo/psalm/actions/runs/3301481330/jobs/5450204951)

PR #8619, opened 2022-10-26
keep literal string for simple str_replace

kkmuffme:
Just for the simplest case, as that is helpful to avoid tons of false positives, especially with callbacks that are created dynamically from paths

As discussed in https://github.com/vimeo/psalm/pull/8611#issuecomment-1289223380

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/StrReplaceReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/StrReplaceReturnTypeProvider.php
@@ -10,6 +10,7 @@ use Psalm\Type\Atomic\TNull;
 use Psalm\Type\Atomic\TString;
 use Psalm\Type\Union;
 
+use function call_user_func;
 use function count;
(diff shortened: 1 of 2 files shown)


mcaskill:
Thanks for the CS fix of `addslashes` on `DirnameTest`.
As for `str_replace` and `str_ireplace`, that's a nice and simple solution.

orklah:
yeah, circleCI is sometimes broken on PR, I'm not sure exactly why, but it happens without fault by the author so I'm merging anyway when I don't see obvious issues. It let some CS issues on master.

About that, you have one of your own right now :)

```
FILE: ...nal/Provider/ReturnTypeProvider/StrReplaceReturnTypeProvider.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 59 | WARNING | Line exceeds 120 characters; contains 127 characters
----------------------------------------------------------------------```

LGTM after that!

orklah:
Thanks!

PR #8620, opened 2022-10-26
add basename return type provider

kkmuffme:
Like dirname return type provider of https://github.com/vimeo/psalm/pull/8611


orklah:
Thanks!

PR #8624, opened 2022-10-27
check "never" return type more strictly

kkmuffme:
* require explicit "never" return type when function always exits
* error if function does not exit, but return type explicitly contains "never"
* Fix: https://github.com/vimeo/psalm/issues/8175
* Fix: https://github.com/vimeo/psalm/issues/8178

### DIFF
--- a/tests/FunctionCallTest.php
+++ b/tests/FunctionCallTest.php
@@ -1957,6 +1957,30 @@ class FunctionCallTest extends TestCase
                 'ignored_issues' => [],
                 'php_version' => '8.0',
             ],
+            'noNeverReturnError' => [
+                'code' => '<?php
+                    /**
+                     * @return string
+                     */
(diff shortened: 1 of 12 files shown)


tests/FunctionCallTest.php
```diff
@@ -2450,6 +2464,53 @@ function scope(int $mode){
...
+                    foo();',
+                'error_message' => 'InvalidReturnType'
+            ],
+            'shouldReturnNeverNotStringNoDocblockCaller' => [
+                'code' => '<?php
+                    /**
+                     * @return string
+                     */
+                    function foo() {
+                       finalFunc();
```
orklah, line 2504: This should return void, not never

orklah:
Yeah probably, did you check the snippet I posted above?

orklah:
Thanks!

PR #8627, opened 2022-10-27
Immutable unions

danog:
This PR introduces fully immutable union types and some work towards immutable storages (I realized that while that's a good thing to have, the only thing we really need to avoid mutability bugs are immutable unions).

Marking this as draft due to the ~30 psalm issues I have to fix, plus a very nasty (set of) side-effects related to parent_nodes, which I'm currently trying to fix.

Basically, the trick here is to fix https://psalm.dev/r/eb9860a6c5, which is caused by `src/Psalm/Internal/Analyzer/Statements/Block/IfElseAnalyzer.php:428` dropping all already-asserted info about `$function_call_arg->value->items[0]` and children, because `src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/InstancePropertyFetchAnalyzer.php:429` re-assigns some types after `AtomicPropertyFetchAnalyzer::processTaints`.

Technically, the behavior of `IfElseAnalyzer` is correct: when changing parent nodes ie due to a variable re-assignment (first part of the psalm.dev example), all known assertions about the variable and children should be dropped.
However, in this specific example in InstancePropertyFetchAnalyzer, originally the re-assignment is propagated to parent contexts, and I haven't figured out yet the proper way of replicating this.


danog:
BTW @orklah feel free to review this when you have time, I bet the fix for the remaining issue will be a two-liner as usual, so not much will change in this PR ;)

src/Psalm/CodeLocation.php
```diff
@@ -124,21 +133,38 @@ public function __construct(
         $this->previous_location = $previous_location;
         $this->text = $selected_text;
 
+        /** @psalm-suppress ImpureMethodCall Actually mutation-free just not marked */
```
orklah, line 136: Maybe adding this to PHPParser stub would be more practical?

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
```diff
@@ -765,7 +772,7 @@ protected function checkTemplateParams(
                     }
 
                     if (!$template_type->isMixed()) {
-                        $template_type_copy = clone $template_type;
+                        $template_type_copy = $template_type;
```
orklah, line 775: is the second variable even necessary now?

src/Psalm/Internal/Analyzer/Statements/Expression/EncapsulatedStringAnalyzer.php
```diff
@@ -33,7 +33,7 @@ public static function analyze(
         PhpParser\Node\Scalar\Encapsed $stmt,
         Context $context
     ): bool {
-        $stmt_type = Type::getString();
+        $stmt_type = new Union([new TString()]);
```
orklah, line 36: why changing this? Purity differences?

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
```diff
@@ -757,60 +788,35 @@ public static function getGlobalType(string $var_id, int $codebase_analysis_php_
...
-                ]),
-                'tmp_name' => new Union([
-                    new TString(),
-                    new TNonEmptyList(Type::getString()),
-                ]),
-                'error' => new Union([
-                    new TIntRange(0, 8),
-                    new TNonEmptyList(Type::getInt()),
-                ]),
+                'name' => $str,
```
orklah, line 793: should this change be included in this PR? Feels like it should be added for 4.x?

src/Psalm/CodeLocation.php
danog: Hmm, do we have a PHPParser stub?

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
danog: Fixed!

src/Psalm/Internal/Analyzer/Statements/Expression/EncapsulatedStringAnalyzer.php
danog: Refactored.

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
danog: I suppose this could be backported as well, just wanted to include it in this PR since I also introduced memoization for global types and wanted to avoid merge conflicts in this section of the VariableFetchAnalyzer.

src/Psalm/CodeLocation.php
orklah: We do :) It's pretty light for now though:
https://github.com/vimeo/psalm/blob/870ee4ff933cff92f5d34abb7ec66d908198316f/stubs/phpparser.phpstub#L10

src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
danog: Though it would have to be changed a bit, here I re-use the string type for example, that can cause nasty issues in 4.x where unions aren't immutable

src/Psalm/CodeLocation.php
danog: Whoops missed that, can I submit a second PR for this later? Since it seems I would basically have to copy over all of php-parser's classes, otherwise psalm complains about child methods not being immutable.
Or maybe I could instead submit a PR to PHP-Parser directly.

orklah: nikic didn't show a lot of enthusiasm when I tried to PR things there so stubbing is the best I think. I'm fine with pushing it later

orklah:
Seems like phpunit plugin broke :(

danog:
Yep, this PR requires the other PR I sent to the phpunit plugin

src/Psalm/CodeLocation.php
danog: OK then, will submit a PR with immutable stubs later for this then.

orklah:
Can you check there? We'll need to make a dedicated release for the phpunit plugin on version 5

danog:
Also fixed the parent_nodes issues.

orklah:
Not sure why, but composer install 0.16.1 of psalm's phpunit plugin, not 0.18.0...

danog:
That's because in semver `0.x` versions are treated as major, only `0.x.y` versions are treated as minor. 
I suggest pulling 0.18.0 and re-tagging 1.17.1, since there are no breaking changes.

danog:
Nevermind, already updated ;)

orklah:
0.18.1 doesn't seem to resolve the issue here

danog:
Yeah, maybe it'll work after merging on master?
Or maybe I should add a version constant somewhere...

danog:
Yeah, sorry about this, https://github.com/psalm/psalm-plugin-phpunit/pull/126 works for sure, just tested :

orklah:
I think this could be merged now. Can you cleanup some commits (or should I squish everything?). Please try to make sure every baseline addition is legit and it should be ok

danog:
You can squash-merge, afaik all commit messages are retained when squash-merging, and yes I made sure all the baselined issues are pre-existing psalm bugs :P

orklah:
Thanks a lot!

danog:
YAY :D 
Just fixed up https://github.com/vimeo/psalm/pull/8395, that should be good to merge too :)

PR #8629, opened 2022-10-29
Remove webmozart/pathUtil

VincentLanglet:
Backport of https://github.com/vimeo/psalm/pull/7292

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "sebastian/diff": "^3.0 || ^4.0",
         "symfony/console": "^3.4.17 || ^4.1.6 || ^5.0 || ^6.0",
         "symfony/polyfill-php80": "^1.25",
-        "webmozart/path-util": "^2.3"
+        "symfony/filesystem": "^4.4 || ^5.4 || ^6.0"
     },
     "provide": {
         "psalm/psalm": "self.version"
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -37,9 +37,9 @@ use Psalm\Progress\VoidProgress;
 use RuntimeException;
 use SimpleXMLElement;
 use SimpleXMLIterator;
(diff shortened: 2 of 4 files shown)


orklah:
In my memories, the suggestion was to use FileSystem from version 5.4. Unfortunately, version 5.4 does not support PHP 7.1 that we need to support on Psalm 4.

In this PR, you allowed FileSystem from 4.4 which supports PHP from 7.1.3.

So two things:
- Is FileSystem 4.4 actually OK for the purpose here? I don't think we have anything running on 7.1 on CI so it may not be checked.
- dependancies won't be able to be solved under 7.1.3, we should probably directly change that in composer.json for clarity

VincentLanglet:
> In my memories, the suggestion was to use FileSystem from version 5.4. Unfortunately, version 5.4 does not support PHP 7.1 that we need to support on Psalm 4.
> 
> In this PR, you allowed FileSystem from 4.4 which supports PHP from 7.1.3.
> 
> So two things:
> 
> * Is FileSystem 4.4 actually OK for the purpose here? I don't think we have anything running on 7.1 on CI so it may not be checked.
> * dependancies won't be able to be solved under 7.1.3, we should probably directly change that in composer.json for clarity

You're right, it was only introduce in SF 5.4, which only supports php >=7.2.5.

Should I close the PR or is it an ok thing to bump the dependencies/php version given the fact that
- Symfony 4 is EOL in november https://symfony.com/releases
- PHP 7.1 and 7.2 are EOL https://www.php.net/supported-versions.php

orklah:
Ideally, Psalm 4 should be EOL and replaced by Psalm 5. It should have already been released but my inexperience shows too much :p . Right now I'm just waiting for the end of the immutable PR and I'll ship it. Psalm 4 should stay as well in the meantime I think

PR #8633, opened 2022-10-30
Add stubs for ext-random (PHP 8.2)

lptn:
First, simple version of functionality introduced or moved to `ext-random`.

I'm new to make this sort of changes to Psalm and open for any suggestions (but prefer to spit changes into few PRs if it's possible).

At the first iteration it's almost a cope of https://github.com/php/php-src/blob/master/ext/random/random.stub.php


stubs/ext-random.phpstub
```diff
@@ -0,0 +1,131 @@
...
+ * PHP 8.2 introduces a new PHP extension named "random".
+ * @see https://github.com/php/php-src/blob/master/ext/random/random.stub.php
+ * @see https://php.watch/versions/8.2/ext-random
+ */
+
+define('MT_RAND_MT19937', 0);
+
+define('MT_RAND_PHP', 1);
+
+namespace {
```
lptn, line 13: these functions and consts from above are existing ones (in PHP 8.1 and older) and moved to ext-random. Not sure what to do with stubs in this case

stubs/ext-random.phpstub
```diff
@@ -0,0 +1,131 @@
...
+
+        public function __construct(?Engine $engine = null) {}
+
+        public function nextInt(): int {}
+
+        public function getInt(int $min, int $max): int {}
+
+        public function getBytes(int $length): string {}
+
+        public function shuffleArray(array $array): array {}
```
lptn, line 109: I would like to add more specific array info later as a separate PR (if it's possible)

orklah:
Thanks for this!

However, we kinda revamped our handling of extensions for Psalm 5. Would you mind targeting the master branch for this? It would be easier to merge later

stubs/ext-random.phpstub
orklah: Fine by me :)

stubs/ext-random.phpstub
orklah: Do we need those stubbed at all? Callmap should be enough for those. I skimmed the RFC, what I understood is that everything is in an extension but the extension is always installed and can't be disabled so it's basically part of the core.

lptn:
@orklah 
Oops. it seems like I produced some changed my solving conflicts after changing base branch. I'll take a loom to this PR in few days to fix it. Thanks for the review!

stubs/ext-random.phpstub
lptn: @orklah 
this is a good question. ext-random should be a part of the code, but you can potentially recompile PHP without it (as far as I understand it in case of this ext). What do you need in such cases - just add to the Callmap?

PR #8637, opened 2022-10-31
Passing class template params to MethodParamsProviderEvent

klimick:
Without templates in the `MethodParamsProvider` is difficult to do something helpful.

For example I want to support `mapN`, `flatMapN` and `etcN` methods in `Option` data class.
It's like `map`, `flatMap` but deconstruct input tuple.

```php
/** @var Option<array{int, int}> */
$option = ...;
$option->mapN(fn($a, $b) => $a + $b);
```

Without knowledge of `array{int, int}` inside `MethodParamsProvider` hook I can't do this.


klimick:
Hmmm. Why tests on Windows are success but on Linux not?)

orklah:
> Why tests on Windows are success but on Linux not?)

PHP 8.1.12 introduced something fishy but I can't pinpoint what exactly. Seem related to sql server

About the PR, would you mind rebasing into master? We plan to stop merging features in 4.x soon

klimick:
Ok. I will open a friend PR on master's later.

PR #8641, opened 2022-11-01
Fix XMLReader::expand() nullable parameter $baseNode

BenMorel:
See: https://www.php.net/manual/en/xmlreader.expand


BenMorel:
Not sure what to do with the [failing tests](https://github.com/vimeo/psalm/actions/runs/3369387343/jobs/5588988566), as they're all related to `sqlsrv`?

orklah:
Failures are unrelated. They came at the same time as PHP 8.1.12, don't know why

PR #8643, opened 2022-11-01
Allow to set PHP 8.2 version

franmomu:
To be able to run psalm with PHP 8.2


orklah:
Thanks! Please keep in mind that some changes of PHP8.2 are not yet implemented

PR #8644, opened 2022-11-01
Fix Spl file handling signatures

### LINKED ISSUE #8642: SplFileInfo::getSize returns int instead of int|false
At least, I think there's something wrong with the stub for this method. The documentation says int or null on failure but trying to handle the failure condition results in a warning from psalm.

 - https://psalm.dev/r/bdaae2f851
 - https://www.php.net/manual/en/splfileinfo.getsize.php

neclimdul:
Update SplFileInfo method signatures and related classes.

Fixes #8642


neclimdul:
After a little digging all these methods just wrap php_stat and the standard file methods which return false on failutre so this updates the signature to reflect that.

orklah:
Thanks!

Seems like SplTempFileObject inherits SplFileObject so those changes should apply to it too. Could you make those changes?

neclimdul:
@orklah think I got SplFileInfo, SplFileObject, and SplTempFileObject so should have covered everything.

orklah:
Oh right, I was confused because CI didn't run properly. Thanks!

Issue #8646, opened 2022-11-01
UnusedFunctionCall reported when using unserialize or serialize


VincentLanglet:
`serialize` and `unserialize` can throw exceptions.

I got the issue with a unit test
```
public function testCannotSerialize(): void
    {
        $this->expectException(BadMethodCallException::class);
        serialize(new Foo());
    }
```

but it can be reproduce this way https://psalm.dev/r/c501b39d6d
so psalm shouldn't report an error.

orklah:
serialize is considered pure. This is *very* questionnable, as serialize will possibly call a bunch of code that can't be determined.

Given it's pure and we don't have a specific stub that documents the function can throw, we consider it useless.

So 2 ways to fix: changing serialize to impure here: https://github.com/vimeo/psalm/blob/2ef993928c5d59a16c74de3841122ed30cd4e1ea/src/Psalm/Internal/Codebase/Functions.php#L406
or adding a stubs in CoreGenericFunctions to document the throw

VincentLanglet:
I tried https://github.com/vimeo/psalm/pull/8650 @orklah.
The idea was to still mark serialize as pure if we're not passing an object, because I have a psalm error with the code base (see first commit).

PR #8650, opened 2022-11-02
Do not report serialize as unused

VincentLanglet:
Closes https://github.com/vimeo/psalm/issues/8646

### DIFF
--- a/src/Psalm/Internal/Codebase/Functions.php
+++ b/src/Psalm/Internal/Codebase/Functions.php
@@ -519,12 +519,24 @@ class Functions
             //gettext
             'bindtextdomain',
 
-            // serialize
-            'serialize', 'unserialize'
+            // unserialize
+            'unserialize'
         ];
 
         if (in_array(strtolower($function_id), $impure_functions, true)) {
             return false;
+        }
+
+        if ($function_id === 'serialize' && isset($args[0]) && $type_provider) {
+            $serialize_type = $type_provider->getType($args[0]->value);
+
+            if ($serialize_type) {
+                foreach ($serialize_type->getAtomicTypes() as $atomic_serialize_type) {
+                    if ($atomic_serialize_type->isObjectType()) {
+                        return false;
+                    }
+                }
+            }
         }
 
         if (strpos($function_id, 'image') === 0) {
--- a/src/Psalm/Internal/Codebase/Functions.php
+++ b/src/Psalm/Internal/Codebase/Functions.php
@@ -518,6 +518,9 @@ class Functions
 
             //gettext
             'bindtextdomain',
+
+            // serialize
+            'serialize', 'unserialize'
         ];
 
         if (in_array(strtolower($function_id), $impure_functions, true)) {
--- a/tests/UnusedCodeTest.php
+++ b/tests/UnusedCodeTest.php
@@ -737,6 +737,31 @@ class UnusedCodeTest extends TestCase
 
                     new Foo();'
             ],
+            'ignoreSerializeAndUnserialize' => [
+                '<?php
+                    class Foo
+                    {
+                        public function __sleep(): array
+                        {
+                            throw new BadMethodCallException();
+                        }
+                        public function __wakeup(): void
+                        {
+                            throw new BadMethodCallException();
+                        }
+                    }
+
+                    function test(): bool {
+                        try {
+                            serialize(new Foo());
+                            unserialize("");
+                        } catch (\Throwable) {
+                            return false;
+                        }
+
+                        return true;
+                    }'
+            ],
             'useIteratorMethodsWhenCallingForeach' => [
                 '<?php
                     /** @psalm-suppress UnimplementedInterfaceMethod */


orklah:
Thanks!

src/Psalm/Internal/Codebase/Functions.php
```diff
@@ -518,12 +518,27 @@ public function isCallMapFunctionPure(
...
 
+        if ($function_id === 'serialize' && isset($args[0]) && $type_provider) {
+            $serialize_type = $type_provider->getType($args[0]->value);
+
+            if ($serialize_type) {
+                foreach ($serialize_type->getAtomicTypes() as $atomic_serialize_type) {
+                    if ($atomic_serialize_type->isObjectType()) {
+                        return false;
+                    }
+                }
```
weirdan, line 538: I don't think it would work for anything but the simplest cases. E.g. `serialize([new RuntimeException]);` would still be marked as unused, wouldn't it?

VincentLanglet: I wanted to avoid the error for
String[] but didnt think about object[] indeed.

Do you see an easy way to improve this or should we test for lot of atomic types ?

VincentLanglet: @orklah @weirdan Seems like this work https://github.com/vimeo/psalm/pull/8652

weirdan: In general, `serialize()` is impure when it operates on anything that *may* contain an object. Your PRs are focusing on the cases when `serialize()` is called on something that *does* contain an object. The distinction is important because there are supertypes to object types, such as `object|int` or `iterable` or `mixed`. In all of those cases `serialize()` should be considered impure:

```php
function a(mixed $p): void { serialize($p); } // impure as an object may be passed here
a(new stdClass); 

function b(iterable $p): void { serialize($p); } // ditto
b(new class implements IteratorAggregate { 
   public function __serialize() { throw new RuntimeException(); } 
   public function getIterator() { return new ArrayIterator([]); }
});
```

However, that's only a part of the problem. The other part is that the types that may contain objects could themselves be a part of an arbitrarily deep type. Say you fixed the case of `serialize([new stdClass]);`, but then you have to fix `serialize([[new stdClass]]);` (note the array depth) and so on.

VincentLanglet: I just try in https://github.com/vimeo/psalm/pull/8652/commits/b2f095f1b8da753f42284957d885c38c8691a3e9 and currently my PR works with
- nested arrays
- Union type like `Foo|int`.

But it doesn't work with mixed atm.
Should I just make it impure for iterable and mixed and it's ok ?

weirdan: In practice, though, limiting checks to a couple of levels (+ a check for subtype) should solve most false positives. The rest (which should be exceedingly rare) can be suppressed by the users.

VincentLanglet: Please review https://github.com/vimeo/psalm/pull/8652 and let me know the tests I need to add, then I'll update the support of 'serialize' method :)

weirdan: >  Should I just make it impure for iterable and mixed and it's ok ?

For a subtype check (e.g. `iterable`, `mixed`, etc) you can use `TypeComparator::canBeContainedBy()`:

https://github.com/vimeo/psalm/blob/8fc499e3416a0b867f61c5e44c86b4edc8ae4c91/src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php#L345

VincentLanglet: I tried and didn't succeed to use this function because it doesn't work for arrays.

weirdan: The type visitor approach is fine, you can just use `TypeComparator` (either `Union*` or `Atomic*`, depending on what the visitor gets) to compare individual type parts.

weirdan: This way, you'd be able to handle `iterable[][]`, for example.

VincentLanglet: I updated the PR, it works with the previous test but still doesn't work with mixed or iterable.

Did I mess up with the check:
```
$type instanceof Union
            && UnionTypeComparator::canBeContainedBy($this->codebase, $type, new Union([new TObject()]))
            ||
            $type instanceof Atomic
            && AtomicTypeComparator::isContainedBy($this->codebase, $type, new TObject())
```
?

weirdan: Yeah, I think the types should be the other way around. The container is the `$type` in this case, and the `$input_type` is `object`, because you want to check whether the `object` can be contained by the `$type`.

You may have to allow coercion (not sure if that method allows it, but there should be one that does) because `object` is not strictly contained by, e.g., `ClassName|int`, but it can be coerced into.

PR #8652, opened 2022-11-03
serialize is not pure for array of object

VincentLanglet:


orklah:
Seems good to me. Can you fix the CS errors?

orklah:
Thanks!

PR #8655, opened 2022-11-03
Fix return type of ReflectionExtension::getVersion

come-nc:
See https://psalm.dev/r/46f74f168a and https://www.php.net/manual/en/reflectionextension.getversion.php

Signed-off-by: Côme Chilliet <come.chilliet@nextcloud.com>


orklah:
Thanks for this!

You'll also need to change the Callmap_historical file that should match with Callmap (when there are no change of signature in recent PHP versions)

orklah:
Thanks!

PR #8658, opened 2022-11-04
Mark hash_update functions as impure

aboks:
This prevents Psalm from marking calls to them as UnusedFunctionCall.


orklah:
Thanks!

PR #8670, opened 2022-11-05
Fix #8669

### LINKED ISSUE #8669: Getting a PossiblyInvalidMethodCall error complaining about a method call on a possibly other variable
I have a value storing the following property

```php
/**
 * @var \Smpl\Utils\Contracts\Supplier<\Smpl\Collections\Contracts\Collection<array-key, E>>|iterable<E>
 * @noinspection PhpDocFieldTypeMismatchInspection
 */
private Supplier|iterable $elementSupplier;
```

Then later on I have the following method.

```php
/**
 * @return iterable<E>
 *
 * @throws \Smpl\Collections\Exceptions\InvalidArgumentException
 */
protected function getElements(): iterable
{
    $elements = $this->elementSupplier;

    if ($elements instanceof Supplier) {
        $elements = $elements->get();
    }

    if (! is_iterable($elements)) {
        throw InvalidArgumentException::invalidElementSupplier(static::class);
    }

    return $elements;
}
```

The `$elements = $elements->get()` line is causing the following error, even though it's wrapped in an `instanceof` check.

```
ERROR: PossiblyInvalidMethodCall - src/Concerns/SuppliesElementsForOperation.php:40:36 - Cannot call method on possible iterable<mixed, E:Smpl\Collections\Operations\DifferenceOperation as mixed> variable $elements (see https://psalm.dev/113)
            $elements = $elements->get();
```

danog:
This is a weird one

### DIFF
--- a/src/Psalm/Internal/Type/AssertionReconciler.php
+++ b/src/Psalm/Internal/Type/AssertionReconciler.php
@@ -601,7 +601,7 @@ class AssertionReconciler extends Reconciler
             $codebase,
             $type_1_atomic,
             $type_2_atomic,
-            !($type_1_atomic instanceof TNamedObject && $type_2_atomic instanceof TNamedObject),
+            $type_1_atomic instanceof TClassString && $type_2_atomic instanceof TClassString,
             false,
             $atomic_comparison_results
         );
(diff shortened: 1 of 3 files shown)


orklah:
Thanks! You were faster than me! I didn't even found a possible fix :p

PR #8691, opened 2022-11-10
Prevent array{a: Foo} going cleanly into array<Foo>

muglug:
With this PR the following unsound code:

```php
/**
 * @param array{a: float, b: float} $params
 */
function avg(array $params): void {
  takesArrayOfFloats($params);
}

/**
 * @param array<array-key, float> $arr
 */
function takesArrayOfFloats(array $arr): void {
    foreach ($arr as $a) {
        echo $a;
    }
}
```

now generates the error:

```
Argument 1 of takesArrayOfFloats expects array<array-key, float>, but parent type array{a: float, b: float} provided
```

## Internal changes

`TKeyedArray` gets a simplified signature since `$sealed` always implied a null key/value param.

## Future work

We need to decide if we want to change the behaviour of `array{a: float, b: float}` to be closed by default.


src/Psalm/Type/Atomic/TKeyedArray.php
```diff
@@ -45,27 +45,19 @@ class TKeyedArray extends Atomic
      */
     public $class_strings;
 
-    /**
-     * True if the objectlike has been created from an explicit array,
-     * or if we're sure that this objectlike has only these keys.
-     *
-     * @var bool
-     */
-    public $sealed = false;
```
weirdan, line 54: Can't we keep the attribute and maintain invariant `$this->sealed = $this->fallback_key_type === null && $this->fallback_value_type === null;` instead? Shouldn't be too much of a hassle with an immutable object, I imagine.

I find it much more readable to check `if ($type->sealed) ...` instead of `if ($type->fallback_value_type === null) ...`.

weirdan: Or replace it with a method like this:
```php
public function isSealed(): bool {
    return $this->fallback_key_type === null && $this->fallback_value_type === null;
}
```

muglug: I'll replace `$this->fallback_key_type` and `$this->fallback_value_type` with `$this->fallback_params`, which will at least make it a slightly prettier check (and it'll match up with my Hack equivalent data structure)

othercorey:
@muglug Could you explain why a string key isn't valid for array-key?

weirdan:
@othercorey it's not about the key type (in this example). It's about unknown values that may be present in the unsealed shape. E.g. Psalm currently sees nothing wrong with https://psalm.dev/r/fb73a41c80, but it actually produces a fatal error: https://3v4l.org/b7XcPK

othercorey:
@weirdan ah, so `array{}` doesn't exclude keys not specified? I thought it was always exclusive.

src/Psalm/Type/Union.php
```diff
@@ -41,7 +40,6 @@
  */
 final class Union implements TypeNode, Stringable
 {
-    use ImmutableNonCloneableTrait;
```
danog, line 44: Why this change?
Full Union immutability was recently introduced by one of my PRs, since the class is marked as `psalm-immutable` it makes no sense to clone it (and after I removed all clones in a later-merged PR performance improved by some 15%)

muglug: That's great! Sadly that PR failed our end-to-end tests, because PSL apparently has a Psalm plugin that calls `Union::__clone`. [Not sure how to work around that](https://github.com/php-standard-library/psalm-plugin/blob/af8580e603772c9aad47ca6272d33ae15fc0502a/src/EventHandler/Type/Optional/FunctionReturnTypeProvider.php#L31) — @azjezz do you have ideas?

muglug: In the meantime I'm going to merge this — we can always add the trait back once the end-to-end test has been fixed.

danog: @azjezz Fixing this is actually pretty simple, just replacing the last three lines with `return $argument_type->setPossiblyUndefined(true)` :)

azjezz: @danog Thanks! will do so :)

danog: @azjezz You should generally also remove all `clone`s of Unions and atomics from your plugin code :)

danog: (Please note that this is only valid for Psalm v5, Psalm v4 still requires clones and your legacy logic)

azjezz: ops! didn't think of that :D not released broken 2.0.3 :D

danog: I actually already did this in https://github.com/vimeo/psalm/pull/8697 (basically provided a custom patch file for our version), this was just a heads-up of the required changes for v5 :)

Issue #8267, opened 2022-07-15
Invalid enum backing type causes crash


AndrolGenhald:
https://psalm.dev/r/89eb27ab21?php=8.1

_Originally posted by @weirdan in https://github.com/vimeo/psalm/issues/6428#issuecomment-913163809_

PR #8693, opened 2022-11-10
Flag invalid enum case value types

weirdan:
Fixes vimeo/psalm#8267

### DIFF
--- a/docs/running_psalm/issues/InvalidEnumCaseValue.md
+++ b/docs/running_psalm/issues/InvalidEnumCaseValue.md
@@ -75,3 +75,28 @@ enum Status: string
     case Open = "open";
 }
 ```
+
+## Case with a type that back an enum
+
+Case type should be either `int` or `string`.
+
+```php
+<?php
+
+enum Status: int {
+    case Open = [];
+}
+```
+
+### How to fix
+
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
```diff
@@ -325,17 +337,6 @@ public function start(PhpParser\Node\Stmt\ClassLike $node): ?bool
                     $this->file_storage->has_visitor_issues = true;
                     $storage->has_visitor_issues = true;
                 }
-                $storage->class_implements['backedenum'] = 'BackedEnum';
```
weirdan, line 328: This means we no longer consider enums backed with an invalid type as actually backed.

docs/running_psalm/issues/InvalidEnumCaseValue.md
```diff
@@ -75,3 +75,28 @@ enum Status: string
     case Open = "open";
 }
 ```
+
+## Case with a type that back an enum
```
weirdan, line 79:
```suggestion
## Case with a type that cannot back an enum
```

PR #8700, opened 2022-11-11
Add extra types to Memcache::getExtendedStats and Memcached::getStats

bram123:
Tightened the return types of Memcached::getStats and Memcache::getExtendedStats, following the examples from:
https://www.php.net/manual/en/memcached.getstats.php
https://www.php.net/manual/en/memcache.getextendedstats.php


orklah:
Thanks!

danog:
Hmm, wouldn't it be better to add shaped arrays instead of generic `array<...>` arrays?

bram123:
> Hmm, wouldn't it be better to add shaped arrays instead of generic `array<...>` arrays?

Ye probably, but afaik this depends on the $type input. And I'd have no idea on how to do this.

PR #8702, opened 2022-11-12
Cleanup documentation

danog:


muglug (approved):

muglug:
I've separately removed a few unnecesary navigation sections in Markdown, since the psalm.dev/docs generator provides those:
 
Screenshot 2022-11-12 at 2 59 30 PM

danog:
>I've separately removed a few unnecesary navigation sections in Markdown

Looks OK for the subtype pages, but could you revert this for https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/?

The link color, while being IMO very pleasant and noticeable when used next to normal text, is less obvious in the main atomic type list, and I think it does not make it clear that the links are actually links.

The first times I came across the type reference page, I thought there was no detailed documentation about the types themselves, and stumbled across more detailed documentation only via the search bar.

Also the autogenerated heading navigation section is not visible on mobile, and is not generated at all for https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/

PR #8705, opened 2022-11-14
Add support for type aliases in @psalm-assert(-*) annotations

### LINKED ISSUE #8263: @psalm-assert-if-true cannot be used with imported types.
https://psalm.dev/r/a9678dc68b

In the same file, it doesn't work if the imported type is also aliased to something else. Probably because the original type is available to the calling code anyway. When the class is used from a different file, it just doesn't work, even if the type is imported in the current scope and it has the correct name.

annervisser:
Fixes #8263

### DIFF
--- a/tests/TypeAnnotationTest.php
+++ b/tests/TypeAnnotationTest.php
@@ -595,12 +595,8 @@ class TypeAnnotationTest extends TestCase
             ],
             'importedTypeUsedInAssertion' => [
                 'code' => '<?php
-                    /**
-                     * @psalm-type Foo = string
-                     */
-                    class A
-                    {
-                    }
+                    /** @psalm-type Foo = string */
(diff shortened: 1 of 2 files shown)


tests/TypeAnnotationTest.php
```diff
@@ -595,12 +595,8 @@ class C extends A {}
...
             'importedTypeUsedInAssertion' => [
                 'code' => '<?php
-                    /**
-                     * @psalm-type Foo = string
-                     */
-                    class A
-                    {
-                    }
+                    /** @psalm-type Foo = string */
+                    class A {}
```
weirdan, line 599: There was nothing wrong with the formatting, but I needed to add some change to re-trigger CircleCI build.

weirdan:
Thanks!

PR #8707, opened 2022-11-15
improve cache flush debug info and code if cache disabled

kkmuffme:
When debugging the message didn't make sense when no cache directory exists yet, wasting time debugging why a config file changed (even though it didn't change, the cache just wasn't populated yet)

Also remove some duplicate code and bail earlier if cache directory is null, since all subsequent calls need it to not be null


orklah:
Thanks!

PR #8708, opened 2022-11-15
Allow using imported types in other types within the same file

### LINKED ISSUE #7116: Use imported type type cannot be used in template
Hey,

If a type is imported into a class, that type cannot be used as a template argument.

See: https://psalm.dev/r/56e1dfad6e

annervisser:
In the test case (from issue #7116) both the type definition and the import of the typed lived in the same file. This caused `OpeningTypes` to be an `InlineTypeAlias` instead of a `LinkableTypeAlias`, in turn causing an 'Invalid type alias' exception[^1]

By replacing the array union (+) with an array_merge, the import of the type overrides the initial type declaration within the `Main` class. This means type imports within one file act more like they would when in separate files.

Fixes #7116

[^1]: [src/Psalm/Type/Atomic.php:407](https://github.com/vimeo/psalm/blob/1986c8b4a8018b1819bc4b83b0f7e69c2c936792/src/Psalm/Type/Atomic.php#L407)

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
+++ b/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
@@ -32,6 +32,7 @@ use Psalm\Storage\MethodStorage;
 use Psalm\Type;
 use UnexpectedValueException;
 
+use function array_merge;
 use function array_pop;
 use function end;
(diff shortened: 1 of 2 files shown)


orklah:
Thanks!

PR #8709, opened 2022-11-15
add --no-progress to psalter

kkmuffme:
like exists in Psalm.php

### DIFF
--- a/src/Psalm/Internal/Cli/Psalter.php
+++ b/src/Psalm/Internal/Cli/Psalter.php
@@ -86,7 +86,8 @@ final class Psalter
         'find-unused-code', 'threads:', 'codeowner:',
         'allow-backwards-incompatible-changes:',
         'add-newline-between-docblock-annotations:',
-        'no-cache'
+        'no-cache',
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Cli/Psalter.php
```diff
@@ -257,9 +261,13 @@ public static function run(array $argv): void
         }
 
         $debug = array_key_exists('debug', $options) || array_key_exists('debug-by-line', $options);
-        $progress = $debug
-            ? new DebugProgress()
-            : new DefaultProgress();
+        if ($debug) {
+            $progress = new DebugProgress();
+        } elseif (isset($options['no-progress'])) {
+            $progress = new VoidProgress();
```
orklah, line 267: you seem to be missing an import for VoidProgress here

orklah: I'm not sure it's classified as Code Style issue given it was an outright bug (Can php cs even know this class does not exists in this namespace??)

orklah:
Thanks!

PR #8710, opened 2022-11-15
safe delete cache directory files

kkmuffme:
since we safeFileGetContents, it makes sense to also acquire lock before deleting, to avoid reading partial data

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -63,9 +63,12 @@ use function count;
 use function dirname;
 use function explode;
 use function extension_loaded;
+use function fclose;
 use function file_exists;
 use function file_get_contents;
 use function filetype;
+use function flock;
+use function fopen;
 use function get_class;
 use function get_defined_constants;
 use function get_defined_functions;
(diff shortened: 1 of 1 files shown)


orklah:
Thanks!

weirdan:
Are there any filesystems where `unlink()` is not atomic?

weirdan:
> RHEL 8/9 it seems.

I find it unlikely, as POSIX specifically mandates that data should not be removed until there are no open file descriptors referencing the file:

> The `unlink()` function shall remove a link to a file. If path names a symbolic link, `unlink()` shall remove the symbolic link named by path and shall not affect any file or directory named by the contents of the symbolic link. Otherwise, `unlink()` shall remove the link named by the pathname pointed to by path and shall *decrement the link count of the file referenced by the link*.
>
> *When the file's link count becomes `0` **and no process has the file open**, the space occupied by the file shall be freed* and the file shall no longer be accessible. If one or more processes have the file open when the last link is removed, the link shall be removed before `unlink()` returns, but the *removal of the file contents shall be postponed until all references to the file are closed*.

*[POSIX.1-2017: unlink()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/unlink.html)*

In my experience, partial reads were pretty much always caused by *partial writes*, so to avoid them, all one needed was to write data into a temporary file (on the target filesystem) and then rename them to the desired file name.

PR #8717, opened 2022-11-18
Fix DateTimeZone::listAbbreviations return type

franmomu:
Based on https://www.php.net/manual/en/datetimezone.listabbreviations.php and https://www.php.net/manual/en/datetimezone.getoffset.php#refsect1-datetimezone.getoffset-changelog


weirdan:
@franmomu please target `master`, we don't plan to cut any more releases from `4.x` branch.

orklah:
Thanks!

PR #8722, opened 2022-11-19
Declaring more precise types and purity boundaries on `ext-reflection` symbols in `.phpstub` files

### LINKED ISSUE #6378: ReflectionClass and impurity of methods
Found a strange issue: https://psalm.dev/r/770fadfec3

I believe that this is a bug, once initiated \ReflectionClass should have immutable state, right? 

Didn't checked for other reflective classes....

### LINKED ISSUE #8720: `ReflectionIntersectionType#getType()` is not `@psalm-pure`
Just reporting this, may get to fix it myself, but for now I found it :) 

https://psalm.dev/r/bb142a0346

```php
<?php

/** @psalm-pure */
function gimmeType(ReflectionIntersectionType $type): array
{
    return $type->getTypes();
}
```
```
Psalm output (using commit 9d4c718): 

ERROR: [ImpureMethodCall](https://psalm.dev/203) - 6:19 - Cannot call a non-mutation-free method ReflectionIntersectionType::getTypes from a pure context
```

Some stub adjustments needed

Ocramius:
Also:

 * added PHP 8.2 stubs
 * refined types to make impossible scenarios more clear (like `ReflectionIntersectionType#allowsNull()`)

This is a first attempt at refining these types: the structure of these stubs is quite confusing to me, so I don't know if this approach is correct, and if the stubs are merged together, or if entire symbols need to be completely re-declared for each PHP version.

Fixes #8720
Fixes #6378

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -2085,6 +2085,16 @@ class Config
             $core_generic_files[] = $stringable_path;
         }
 
+        if (PHP_VERSION_ID < 8_02_00 && $codebase->analysis_php_version_id >= 8_02_00) {
+            $stringable_path = dirname(__DIR__, 2) . '/stubs/Php82.phpstub';
+
+            if (!file_exists($stringable_path)) {
+                throw new UnexpectedValueException('Cannot locate PHP 8.2 classes');
+            }
+
+            $core_generic_files[] = $stringable_path;
+        }
+
         $stub_files = array_merge($core_generic_files, $this->preloaded_stub_files);
 
         if (!$stub_files) {
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -2065,7 +2065,7 @@ class Config
 
         $core_generic_files = [];
 
-        if (PHP_VERSION_ID < 8_00_00 && $codebase->analysis_php_version_id >= 8_00_00) {
+        if ($codebase->analysis_php_version_id >= 8_00_00) {
             $stringable_path = dirname(__DIR__, 2) . '/stubs/Php80.phpstub';
 
             if (!file_exists($stringable_path)) {
@@ -2075,7 +2075,7 @@ class Config
             $core_generic_files[] = $stringable_path;
         }
 
-        if (PHP_VERSION_ID < 8_01_00 && $codebase->analysis_php_version_id >= 8_01_00) {
+        if ($codebase->analysis_php_version_id >= 8_01_00) {
             $stringable_path = dirname(__DIR__, 2) . '/stubs/Php81.phpstub';
 
             if (!file_exists($stringable_path)) {
@@ -2085,7 +2085,7 @@ class Config
             $core_generic_files[] = $stringable_path;
         }
 
-        if (PHP_VERSION_ID < 8_02_00 && $codebase->analysis_php_version_id >= 8_02_00) {
+        if ($codebase->analysis_php_version_id >= 8_02_00) {
             $stringable_path = dirname(__DIR__, 2) . '/stubs/Php82.phpstub';
 
             if (!file_exists($stringable_path)) {
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -2065,7 +2065,7 @@ class Config
 
         $core_generic_files = [];
 
-        if ($codebase->analysis_php_version_id >= 8_00_00) {
+        if (PHP_VERSION_ID < 8_00_00 && $codebase->analysis_php_version_id >= 8_00_00) {
             $stringable_path = dirname(__DIR__, 2) . '/stubs/Php80.phpstub';
 
             if (!file_exists($stringable_path)) {
@@ -2075,7 +2075,7 @@ class Config
             $core_generic_files[] = $stringable_path;
         }
 
-        if ($codebase->analysis_php_version_id >= 8_01_00) {
+        if (PHP_VERSION_ID < 8_01_00 && $codebase->analysis_php_version_id >= 8_01_00) {
             $stringable_path = dirname(__DIR__, 2) . '/stubs/Php81.phpstub';
 
             if (!file_exists($stringable_path)) {
@@ -2085,7 +2085,7 @@ class Config
             $core_generic_files[] = $stringable_path;
         }
 
-        if ($codebase->analysis_php_version_id >= 8_02_00) {
+        if (PHP_VERSION_ID < 8_02_00 && $codebase->analysis_php_version_id >= 8_02_00) {
             $stringable_path = dirname(__DIR__, 2) . '/stubs/Php82.phpstub';
 
             if (!file_exists($stringable_path)) {
@@ -2135,13 +2135,18 @@ class Config
             $dir_lvl_2 . DIRECTORY_SEPARATOR . 'stubs' . DIRECTORY_SEPARATOR . 'SPL.phpstub',
         ];
 
-        if (PHP_VERSION_ID >= 8_00_00 && $codebase->analysis_php_version_id >= 8_00_00) {
+        if ($codebase->analysis_php_version_id >= 8_00_00) {
             $stringable_path = $dir_lvl_2 . DIRECTORY_SEPARATOR . 'stubs' . DIRECTORY_SEPARATOR . 'Php80.phpstub';
             $this->internal_stubs[] = $stringable_path;
         }
 
-        if (PHP_VERSION_ID >= 8_01_00 && $codebase->analysis_php_version_id >= 8_01_00) {
+        if ($codebase->analysis_php_version_id >= 8_01_00) {
             $stringable_path = $dir_lvl_2 . DIRECTORY_SEPARATOR . 'stubs' . DIRECTORY_SEPARATOR . 'Php81.phpstub';
+            $this->internal_stubs[] = $stringable_path;
+        }
+
+        if ($codebase->analysis_php_version_id >= 8_02_00) {
+            $stringable_path = $dir_lvl_2 . DIRECTORY_SEPARATOR . 'stubs' . DIRECTORY_SEPARATOR . 'Php82.phpstub';
             $this->internal_stubs[] = $stringable_path;
         }
 
--- a/stubs/Php80.phpstub
+++ b/stubs/Php80.phpstub
@@ -17,14 +17,24 @@ class ReflectionAttribute
     {
     }
 
+    /**
+     * @return non-empty-string
+     *
+     * @psalm-pure
+     */
     public function getName() : string
     {
     }
 
+    /**
+     * @psalm-pure
+     * @return int-mask-of<Attribute::TARGET_*>
+     */
     public function getTarget() : int
     {
     }
 
+    /** @psalm-pure */
     public function isRepeated() : bool
     {
     }
@@ -48,13 +58,101 @@ class ReflectionAttribute
     }
 }
 
+/**
+ * @template-covariant T as object
+ *
+ * @property-read class-string<T> $name
+ */
+class ReflectionClass implements Reflector {
+    /**
+     * @return non-empty-string|false
+     * @psalm-pure
+     */
+    public function getFileName(): string|false {}
+
+    /**
+     * @return positive-int|false
+     * @psalm-pure
+     */
+    public function getStartLine(): int|false {}
+
+    /**
+     * @return positive-int|false
+     * @psalm-pure
+     */
+    public function getEndLine(): int|false {}
+
+    /**
+     * @return non-empty-string|false
+     * @psalm-pure
+     */
+    public function getDocComment(): string|false {}
+
+    /**
+     * @param ReflectionClass|class-string $class
+     *
+     * @psalm-pure
+     */
+    public function isSubclassOf(self|string $class): bool {}
+
+    /**
+     * @param self|class-string $interface
+     *
+     * @psalm-pure
+     */
+    public function implementsInterface(self|string $interface): bool {}
+
+    /**
+     * @return non-empty-string|false
+     *
+     * @psalm-pure
+     */
+    public function getExtensionName(): string|false {}
+}
+
+/** @psalm-immutable */
 class ReflectionClassConstant
 {
     public const IS_PUBLIC = 1;
     public const IS_PROTECTED = 2;
     public const IS_PRIVATE = 4;
+
+    /** @return non-empty-string|false */
+    public function getDocComment(): string|false {}
 }
 
+abstract class ReflectionFunctionAbstract implements Reflector
+{
+    /**
+     * @return non-empty-string|false
+     *
+     * @psalm-pure
+     */
+    public function getDocComment(): string|false {}
+
+    /**
+     * @return positive-int|false
+     *
+     * @psalm-pure
+     */
+    public function getStartLine(): int|false {}
+
+    /**
+     * @return positive-int|false
+     *
+     * @psalm-pure
+     */
+    public function getEndLine(): int|false {}
+
+    /**
+     * @return non-empty-string|false
+     *
+     * @psalm-pure
+     */
+    public function getFileName(): string|false {}
+}
+
+/** @psalm-immutable */
 class Attribute
 {
     public int $flags;
@@ -76,11 +174,20 @@ class Attribute
     }
 }
 
-class ReflectionUnionType extends ReflectionType {
+class ReflectionProperty implements Reflector
+{
     /**
-     * @return non-empty-list<ReflectionNamedType>
+     * @return non-empty-string|false
+     *
+     * @psalm-pure
      */
-    public function getTypes() {}
+    public function getDocComment(): string|false {}
+}
+
+/** @psalm-immutable */
+class ReflectionUnionType extends ReflectionType {
+    /** @return non-empty-list<ReflectionNamedType> */
+    public function getTypes(): array {}
 }
 
 class UnhandledMatchError extends Error {}
--- a/stubs/Reflection.phpstub
+++ b/stubs/Reflection.phpstub
@@ -432,7 +432,11 @@ abstract class ReflectionFunctionAbstract implements Reflector
 
 class ReflectionFunction extends ReflectionFunctionAbstract
 {
-    /** @psalm-pure */
+    /**
+     * @param callable-string|Closure $function
+     *
+     * @psalm-pure
+     */
     public function __construct(callable $function) {}
 
     /**
--- a/stubs/Reflection.phpstub
+++ b/stubs/Reflection.phpstub
@@ -446,7 +446,7 @@ class ReflectionFunction extends ReflectionFunctionAbstract
 class ReflectionProperty implements Reflector
 {
     /**
-     * @var non-empty-string
+     * @var string
      * @readonly
      */
     public $name;
--- a/stubs/Reflection.phpstub
+++ b/stubs/Reflection.phpstub
@@ -14,13 +14,194 @@ class ReflectionClass implements Reflector {
 
     /**
      * @param T|class-string<T>|interface-string<T>|trait-string|enum-string<T> $argument
+     * @psalm-pure
      */
     public function __construct($argument) {}
 
     /**
      * @return class-string<T>
+     * @psalm-pure
      */
     public function getName(): string {}
+    
+    /** @psalm-pure */
+    public function isInternal(): bool {}
+
+    /** @psalm-pure */
+    public function isUserDefined(): bool {}
+
+    /** @psalm-pure */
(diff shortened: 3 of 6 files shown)


stubs/Reflection.phpstub
```diff
@@ -41,23 +222,205 @@ class ReflectionClass implements Reflector {
...
+     * @psalm-pure
      */
     public function getTraitNames(): array {}
 
     /**
      * @since 8.0
      * @template TClass as object
      * @param class-string<TClass>|null $name
-     * @return ($name is null ? array<ReflectionAttribute<object>> : array<ReflectionAttribute<TClass>>)
+     * @return ($name is null ? list<ReflectionAttribute<object>> : list<ReflectionAttribute<TClass>>)
```
Ocramius, line 292: This is based off the work done by me and @kukulich on [roave/better-reflection](https://github.com/Roave/BetterReflection), btw.

Most of these types were refined after careful testing of feature parity with upstream PHP.

stubs/Reflection.phpstub
```diff
@@ -41,23 +222,205 @@ class ReflectionClass implements Reflector {
...
+
+    /** @psalm-pure */
+    public function getClosureCalledClass(): ?ReflectionClass {}
+
+    /**
+     * @return non-empty-string|false
+     *
+     * @psalm-pure
+     */
+    public function getDocComment() {}
```
Ocramius, line 328: I experimented a bit with this: the minimal doc comment that can be written is `/** */`, and it will produce the string `'/** */'`.

It is not possible to get a `string&empty` doc comment from reflection

stubs/Reflection.phpstub
```diff
@@ -92,32 +475,74 @@ class ReflectionProperty implements Reflector
...
-    * @since 8.1
-    */
+     * @since 8.1
+     * @psalm-pure
+     */
     public function isReadOnly(): bool {}
 }
 
-class ReflectionMethod implements Reflector
+class ReflectionMethod extends ReflectionFunctionAbstract
```
Ocramius, line 545: Inheritance was completely broken here :|

stubs/Reflection.phpstub
```diff
@@ -67,10 +430,23 @@ class ReflectionFunction implements Reflector
     public function getAttributes(?string $name = null, int $flags = 0): array {}
 }
 
+class ReflectionFunction extends ReflectionFunctionAbstract
+{
+    /** @psalm-pure */
+    public function __construct(callable $function) {}
```
Ocramius, line 436: I'm a bit on the fence on this one: not all `callable` types are valid types here.

```php
<?php

class Foo
{
    public function bar() {}
    private function baz() {}
}

var_dump(new ReflectionFunction('Foo::bar'));
var_dump(new ReflectionFunction('Foo::baz'));
```

https://3v4l.org/Fsh3P

It is true that `callable` is a much better type than `non-empty-string|Closure` here, but not all `non-empty-string` types are callable, and not all `callable` are `non-empty-string` (nor valid in this context).

Anyway, this is perhaps just overthinking, and `callable` may be an improvement over the past.

src/Psalm/Config.php
```diff
@@ -2107,6 +2107,16 @@ public function visitPreloadedStubFiles(Codebase $codebase, ?Progress $progress
...
 
+        if (PHP_VERSION_ID < 8_02_00 && $codebase->analysis_php_version_id >= 8_02_00) {
+            $stringable_path = dirname(__DIR__, 2) . '/stubs/Php82.phpstub';
+
+            if (!file_exists($stringable_path)) {
+                throw new UnexpectedValueException('Cannot locate PHP 8.2 classes');
+            }
+
+            $core_generic_files[] = $stringable_path;
+        }
```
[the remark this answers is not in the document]
weirdan: and `$stringable_path` is a misleading name (was used to import `Stringable` stub once and then was just copy-pasted around a lot).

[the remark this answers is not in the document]
weirdan: @orklah do you think we should drop `PHP_VERSION_ID` checks here? Stubs usually contain more detailed type info compared to reflection, so why don't we load them regardless of the runtime PHP version?

stubs/Php80.phpstub
```diff
@@ -76,11 +174,20 @@ class Attribute
...
+     *
+     * @psalm-pure
      */
-    public function getTypes() {}
+    public function getDocComment(): string|false {}
+}
+
+/** @psalm-immutable */
+class ReflectionUnionType extends ReflectionType {
+    /** @return non-empty-list<ReflectionNamedType> */
```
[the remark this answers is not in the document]
weirdan: An unsealed tuple may allow specifying that two first elements are `ReflectionNamedType`s, but it would allow passing other types with offsets 2+. Psalm currently doesn't support array intersections. If it did, it would be possible with `array{ReflectionNamedType,ReflectionNamedType} & list<ReflectionNamedType>`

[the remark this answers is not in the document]
Ocramius: Was thinking the same, but then a tuple is too restrictive anyway

stubs/Reflection.phpstub
weirdan: However I think we should exclude callable arrays and objects here. `callable-string|Closure` (in docblock) should be better in this context.

stubs/Reflection.phpstub
```diff
@@ -67,10 +430,23 @@ class ReflectionFunction implements Reflector
...
+     * @psalm-pure
+     */
+    public function __toString(): string {}
+}
+
 class ReflectionProperty implements Reflector
 {
     /**
-     * @var string
+     * @var non-empty-string
```
weirdan, line 449: Property name can actually be an empty string: https://3v4l.org/K3kpQ

Ocramius: I didn't see this: I will keep doing as if I didn't 😱

src/Psalm/Config.php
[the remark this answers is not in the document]
orklah: I'd be very cautious here. The Reflection library change frequently and breaks often. For example, you don't want to say that some method can return `ReflectionIntersectionType` before PHP 8.2 because it will break all kind of assumptions. This specific case can be covered with conditional return, but the existence or not of some methods/class is more tricky I think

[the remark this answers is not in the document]
Ocramius: Just be aware that I only replicated the 2 code blocks above this one

stubs/Php80.phpstub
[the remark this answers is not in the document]
Ocramius: Skipping on this, for now: maybe one day :)

src/Psalm/Config.php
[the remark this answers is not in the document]
weirdan: To clarify, I don't want to drop `$codebase->analysis_php_version_id` check.

Now we have the following:
* `visitPreloadedStubFiles()` does this: 
  ```php
        if (PHP_VERSION_ID < 8_00_00 && $codebase->analysis_php_version_id >= 8_00_00) {
            // load Php80.phpstub
        }

        if (PHP_VERSION_ID < 8_01_00 && $codebase->analysis_php_version_id >= 8_01_00) {
            // load Php81.phpstub
        }
  ```
* `visitStubFiles()` does this:
  ```php
        if (PHP_VERSION_ID >= 8_00_00 && $codebase->analysis_php_version_id >= 8_00_00) {
             // load Php80.phpstub
        }

        if (PHP_VERSION_ID >= 8_01_00 && $codebase->analysis_php_version_id >= 8_01_00) {
             // load Php81.phpstub
        }
  ```
* Usually both are called, except when in the `diff` mode. When both are called, `PHP_VERSION_ID` check is useless (it will load stubs either in `visitPreloaded...` or in `visitStub...`.

stubs/Reflection.phpstub
```diff
@@ -14,13 +14,194 @@ class ReflectionClass implements Reflector {
...
+    /** @psalm-pure */
+    public function hasMethod(string $name): bool {}
+
+    /**
+     * @param non-empty-string $name
+     *
+     * @psalm-pure
+     * @throws ReflectionException
+     */
+    public function getMethod(string $name): ReflectionMethod {}
```
kukulich, line 75: `non-empty-string` is not right: https://github.com/Roave/BetterReflection/issues/1278

Ocramius: This is about preventing nonsensical code though, not a runtime check

Ocramius:
Besides feedback provided so far, is this the right way of implementing this patch, with multiple stubs for the same class, each stub "decorating" previous PHP version stubs?

If that's the case, then I can gladly apply the provided feedback, get the build green, and then we can move this forward.

orklah:
To be honest, this is pretty much uncharted territories, The versioned stubs are somewhat recent (we never needed those before PHP 8) and we mainly built them empirically when issues are raised.

So yeah, if it works as is, I, for one, will gladly merge this but I'm not perfectly sure it will. For example, I'm not completely sure we can have two loaded stubs of the same class (especially if there are conflicting methods). If it's causing an issue, we may have to resort to having exclusive stubs loaded for only one PHP version.

By the way, sorry to bring this only now, we have an extension folder in which the reflection stubs should probably be extracted (I kinda forgot this since it's something new from Psalm 5 that was added some time ago). If we have to have versioned stubs, this will be easier to have a Reflection71, Reflection72... in there

stubs/Reflection.phpstub
Ocramius: d5cccbade209819c3a350ec96d89150bcb7bc7bb

Ocramius: (commit message humorous, please do tell if you dislike :D )

stubs/Reflection.phpstub
Ocramius: d9a0cc531113c75ea98dba04cbeccc01f4ae7ba4

stubs/Reflection.phpstub
```diff
@@ -14,13 +14,194 @@ class ReflectionClass implements Reflector {
...
+    /**
+     * @template TCheckedObject of object
+     *
+     * @param TCheckedObject $object
+     *
+     * @return (TCheckedObject is T ? true : bool)
+     *
+     * @psalm-pure
+     */
+    public function isInstance(object $object): bool {}
```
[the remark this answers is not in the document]
Ocramius: 79a1a8b26ce5f8fc9589fd779a3ec1da939b07c8

Ocramius:
I did some more local testing: PHP 8.2 stubs are somehow not picked up correctly.

That will take some time to debug :D

src/Psalm/Config.php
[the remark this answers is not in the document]
Ocramius: Based on local testing, I will indeed need to change this a bit.

Currently, the logic does:

```php
if (PHP_VERSION_ID < 8_00_00 && $codebase->analysis_php_version_id >= 8_00_00) {
```

Practically, it should just do this everywhere.

```php
if (PHP_VERSION_ID >= 8_00_00 && $codebase->analysis_php_version_id >= 8_00_00) {
```

Unsure about implications of changing this :thinking:

stubs/Reflection.phpstub
```diff
@@ -92,32 +471,74 @@ class ReflectionProperty implements Reflector
...
-    * @since 8.1
-    */
+     * @since 8.1
+     * @psalm-pure
+     */
     public function isReadOnly(): bool {}
 }
 
-class ReflectionMethod implements Reflector
+class ReflectionMethod extends ReflectionFunctionAbstract
```
Ocramius, line 541: Possibly some trouble with this? This is the analysis of PHPUnit after my changes:

```
ERROR: MissingThrowsDocblock - src/Framework/Constraint/ObjectEquals.php:64:28 - ReflectionException is thrown but not caught - please either catch or add a @throws annotation (see https://psalm.dev/169)
        $method = $object->getMethod($this->method);

ERROR: ArgumentTypeCoercion - src/Framework/Constraint/ObjectEquals.php:64:38 - Argument 1 of ReflectionObject::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
        $method = $object->getMethod($this->method);

ERROR: MissingThrowsDocblock - src/Framework/MockObject/Generator.php:741:40 - ReflectionException is thrown but not caught - please either catch or add a @throws annotation (see https://psalm.dev/169)
                $cloneMethod = $class->getMethod('__clone');

ERROR: ArgumentTypeCoercion - src/Framework/MockObject/Generator.php:770:49 - Argument 1 of ReflectionClass::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
                    $method = $class->getMethod($methodName);

ERROR: MissingThrowsDocblock - src/Framework/TestCase.php:1839:21 - ReflectionException is thrown but not caught - please either catch or add a @throws annotation (see https://psalm.dev/169)
            $class->getMethod($methodName)

ERROR: ArgumentTypeCoercion - src/Framework/TestCase.php:1839:31 - Argument 1 of ReflectionClass::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
            $class->getMethod($methodName)

ERROR: ArgumentTypeCoercion - src/Framework/TestCase.php:1992:38 - Argument 1 of ReflectionObject::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
               $reflector->getMethod($methodName)->getDeclaringClass()->getName() === self::class;

ERROR: ArgumentTypeCoercion - src/Framework/TestSuite.php:615:38 - Argument 1 of ReflectionClass::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
               $reflector->getMethod($methodName)->getDeclaringClass()->getName() === TestCase::class;

ERROR: ArgumentTypeCoercion - src/Metadata/Api/DataProvider.php:96:45 - Argument 1 of ReflectionClass::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
                $method = $class->getMethod($_dataProvider->methodName());

ERROR: UndefinedConstant - src/Util/Reflection.php:27:45 - Constant ReflectionMethod::IS_PUBLIC is not defined (see https://psalm.dev/020)
        return $this->filterMethods($class, ReflectionMethod::IS_PUBLIC);
```

Most notably: 

```
ERROR: UndefinedConstant - src/Util/Reflection.php:27:45 - Constant ReflectionMethod::IS_PUBLIC is not defined (see https://psalm.dev/020)
        return $this->filterMethods($class, ReflectionMethod::IS_PUBLIC);
```

Unsure about this one - this has to do with the stub loading mechanism, I think :thinking:

Ocramius: Same in my local test, so more work needed here :cold_sweat: 

```
ERROR: UndefinedConstant - reflection-test.php:41:53 - Constant ReflectionMethod::IS_PUBLIC is not defined (see https://psalm.dev/020)
$results = [$name, $unionTypes, $intersectionTypes, \ReflectionMethod::IS_PUBLIC];
```

I wonder if constants in declared reflection symbols are not merged with stubs :thinking:

orklah:
Thanks for your work on this! I'll find some time this week end to look at what's going on to see if I can help :)

stubs/Reflection.phpstub
Ocramius: As per 68d88c546b286a5fbdabe4728f6be9dd1db0dbe0, I migrated everything to `visitStubFiles`, which fixed most issues :)

src/Psalm/Config.php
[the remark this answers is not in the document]
Ocramius: Moved everything to `visitStubFiles`, and it looks fine :+1:

src/Psalm/Config.php
```diff
@@ -2085,6 +2085,16 @@ public function visitPreloadedStubFiles(Codebase $codebase, ?Progress $progress
...
+        if ($codebase->analysis_php_version_id >= 8_02_00) {
+            $stringable_path = dirname(__DIR__, 2) . '/stubs/Php82.phpstub';
+
+            if (!file_exists($stringable_path)) {
+                throw new UnexpectedValueException('Cannot locate PHP 8.2 classes');
+            }
+
+            $core_generic_files[] = $stringable_path;
+        }
+
```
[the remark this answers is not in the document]
Ocramius: I moved everything to `visitStubFiles`, and that's fine

stubs/Reflection.phpstub
Ocramius: PHPUnit real-world test looks much better now:

```
ERROR: MissingThrowsDocblock - src/Framework/Constraint/ObjectEquals.php:64:28 - ReflectionException is thrown but not caught - please either catch or add a @throws annotation (see https://psalm.dev/169)
        $method = $object->getMethod($this->method);

ERROR: ArgumentTypeCoercion - src/Framework/Constraint/ObjectEquals.php:64:38 - Argument 1 of ReflectionObject::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
        $method = $object->getMethod($this->method);

ERROR: MissingThrowsDocblock - src/Framework/MockObject/Generator.php:741:40 - ReflectionException is thrown but not caught - please either catch or add a @throws annotation (see https://psalm.dev/169)
                $cloneMethod = $class->getMethod('__clone');

ERROR: ArgumentTypeCoercion - src/Framework/MockObject/Generator.php:770:49 - Argument 1 of ReflectionClass::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
                    $method = $class->getMethod($methodName);

ERROR: MissingThrowsDocblock - src/Framework/TestCase.php:1839:21 - ReflectionException is thrown but not caught - please either catch or add a @throws annotation (see https://psalm.dev/169)
            $class->getMethod($methodName)

ERROR: ArgumentTypeCoercion - src/Framework/TestCase.php:1839:31 - Argument 1 of ReflectionClass::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
            $class->getMethod($methodName)

ERROR: ArgumentTypeCoercion - src/Framework/TestCase.php:1992:38 - Argument 1 of ReflectionObject::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
               $reflector->getMethod($methodName)->getDeclaringClass()->getName() === self::class;

ERROR: ArgumentTypeCoercion - src/Framework/TestSuite.php:615:38 - Argument 1 of ReflectionClass::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
               $reflector->getMethod($methodName)->getDeclaringClass()->getName() === TestCase::class;

ERROR: ArgumentTypeCoercion - src/Metadata/Api/DataProvider.php:96:45 - Argument 1 of ReflectionClass::getMethod expects non-empty-string, but parent type string provided (see https://psalm.dev/193)
                $method = $class->getMethod($_dataProvider->methodName());

ERROR: ArgumentTypeCoercion - src/Util/Reflection.php:45:37 - Argument 1 of ReflectionClass::getMethods expects 0|1|2|3|4|5|6|7|16|17|18|19|20|21|22|23|32|33|34|35|36|37|38|39|48|49|50|51|52|53|54|55|64|65|66|67|68|69|70|71|80|81|82|83|84|85|86|87|96|97|98|99|100|101|102|103|112|113|114|115|116|117|118|119|null, but parent type int|null provided (see https://psalm.dev/193)
        foreach ($class->getMethods($filter) as $method) {

```

Ocramius:
@orklah I think actual coding work on this is complete: what is missing is validation / decision making, but otherwise we are green :) 

Hopefully that means not cutting too much into your time anymore  :+1:

orklah:
Thanks Marco!

I checked that and I was surprised that Stringable was not flagged as an error here: https://psalm.dev/r/3d60a7e0ea?phpVersion=7.1

So we already have global namespace pollution and it didn't cause much issue, I'm not worried about Reflection classes leaking then.

I'd be interested in the result of running Psalm against BetterReflection in different php versions for verification but otherwise, I'm good! We can merge when ready :)

Ocramius:
The two most aggressive reflection codebases I know of are:

 * `laminas/laminas-code` (very inheritance-heavy)
 * `roave/better-reflection`, which is providing full polyfills for core reflection 

I will check both against this patch and report back 👍

Ocramius:
I ran a report over `roave/better-reflection` and `laminas/laminas-code` (attached):


[laminas-laminas-code-report.txt](https://github.com/vimeo/psalm/files/10206665/laminas-laminas-code-report.txt)

[roave-betterreflection-report.txt](https://github.com/vimeo/psalm/files/10206667/roave-betterreflection-report.txt)


Overall just `MissingImmutableAnnotation` and a few less specific types downstream :)

orklah:
Seems great! Thanks a lot Marco!

Ocramius:
🎉

PR #8724, opened 2022-11-20
include CLI args in PHP errors to more quickly identify run issues in CI

kkmuffme:
When errors are collected in a general log, identifying the origin Psalm command requires multiple additional steps if Psalm encounters an PHP error. Including the CLI args in the error makes this much quicker.


weirdan:
Thanks!

PR #8727, opened 2022-11-21
validate cache writes are successful

kkmuffme:


weirdan:
It would be nice to hide all this heavy FS wizardry behind a simpler interface. E.g.
```php
interface CacheBlob {
   public function read(): string;
   public function write(string): bool;
}

final class CacheFile implements CacheBlob {
   private string $filename;
   public function read(): string {
      // lock the file
      // read
      // validate that it's fully read
      // return 
   }

   public function write(string $data): bool {
      // write to a temporary file
      // validate it's fully written
      // move it to $this->filename
   }
}
```

PR #8728, opened 2022-11-21
Add note for 5.1 BC break

danog:


weirdan:
Shouldn't we mark `TList` and `TNonEmptyList` as deprecated?

danog:
Ah yes good idea

orklah:
Shouldn't this be included in Psalm 5? Seems like a big BC break for plugins, no?

danog:
Yes but it's a lot of work and my https://github.com/danog/psalm/tree/tnon_empty_list_refactoring branch is not ready yet, and I really don't want to hold back the release any longer for a last-minute idea :P 
I feel like forewarning about an near-future BC break during a major release ought to be enough.

danog:
I've already created Type::getListAtomic and Type::getNonEmptyListAtomic methods in https://github.com/danog/psalm/tree/tnon_empty_list_refactoring and I've already replaced them across the entire codebase, I could copy in a placeholder version in this PR as well.

danog:
All done!

PR #8729, opened 2022-11-21
Fix dictionary for strip_tags

lptn:
Changes:
 - allow null for 2nd parameter `$allowed_tags` from PHP 8.0 (see [docs](https://www.php.net/manual/en/function.strip-tags.php#refsect1-function.strip-tags-changelog))
 - allow array for 2nd parameter `$allowed_tags` from PHP 7.4 (see [docs](https://www.php.net/manual/en/function.strip-tags.php#refsect1-function.strip-tags-changelog))

Proofs:
 - https://3v4l.org/VjbXO
 - https://3v4l.org/VWsPr

### DIFF
--- a/dictionaries/CallMap_74_delta.php
+++ b/dictionaries/CallMap_74_delta.php
@@ -45,6 +45,10 @@ return [
       'old' => ['resource|false', 'command'=>'string', 'descriptor_spec'=>'array', '&pipes'=>'resource[]', 'cwd='=>'?string', 'env_vars='=>'?array', 'options='=>'?array'],
       'new' => ['resource|false', 'command'=>'string|array', 'descriptor_spec'=>'array', '&pipes'=>'resource[]', 'cwd='=>'?string', 'env_vars='=>'?array', 'options='=>'?array'],
     ],
+    'strip_tags' => [
+      'old' => ['string', 'string'=>'string', 'allowed_tags='=>'?string'],
+      'new' => ['string', 'string'=>'string', 'allowed_tags='=>'string|null|list<non-empty-string>'],
+    ],
   ],
   'removed' => [
   ],
--- a/dictionaries/CallMap_74_delta.php
+++ b/dictionaries/CallMap_74_delta.php
@@ -46,8 +46,8 @@ return [
       'new' => ['resource|false', 'command'=>'string|array', 'descriptor_spec'=>'array', '&pipes'=>'resource[]', 'cwd='=>'?string', 'env_vars='=>'?array', 'options='=>'?array'],
     ],
     'strip_tags' => [
-      'old' => ['string', 'string'=>'string', 'allowed_tags='=>'?string'],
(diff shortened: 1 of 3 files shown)


dictionaries/CallMap_74_delta.php
```diff
@@ -45,6 +45,10 @@
       'old' => ['resource|false', 'command'=>'string', 'descriptor_spec'=>'array', '&pipes'=>'resource[]', 'cwd='=>'?string', 'env_vars='=>'?array', 'options='=>'?array'],
       'new' => ['resource|false', 'command'=>'string|array', 'descriptor_spec'=>'array', '&pipes'=>'resource[]', 'cwd='=>'?string', 'env_vars='=>'?array', 'options='=>'?array'],
     ],
+    'strip_tags' => [
+      'old' => ['string', 'string'=>'string', 'allowed_tags='=>'?string'],
```
kamil-tekiela, line 49: As I understand it, null was not accepted prior to PHP 8.0

lptn: I also found this information in documentation. But I also checked on PHP 7.2 - it also works with `null`: https://3v4l.org/nkHYY#v7.2.0

kamil-tekiela: Yeah, but that's because of undefined behaviour. Just because it used to work doesn't mean that it was designed like that or that it should be used this way. 
Arrays worked too https://3v4l.org/PE5hP#v7.2.0 and so did booleans https://3v4l.org/qoDFT#v7.2.0
Even with strict types, you could pass anything you'd like https://3v4l.org/mJvL3#v7.2.0 That's why in PHP 8.0 and 8.1 this behaviour was fixed.

lptn: @kamil-tekiela 
thanks a lot, should be fixed by d33c338

orklah:
Thanks for that! Two changes:
- Could you target master branch please? We're very close to releasing the next Psalm version and it will be a part of it
- Callmap.php should also be changed to reflect the most up to date signature

lptn:
@orklah 
>We're very close to releasing the next Psalm version and it will be a part of it

that's great news! Are you going to merge changed from 4.x to 5.x (made after 5.x branch off)?



> Could you target master branch please? We're very close to releasing the next Psalm version and it will be a part of it

done

> Callmap.php should also be changed to reflect the most up to date signature

done

orklah:
> that's great news! Are you going to merge changed from 4.x to 5.x (made after 5.x branch off)?

If I'm not mistaken, we did already :) The only thing left to do would be to make a (last?) release from 4.x because some PR were merged after the latest release. We can do that even after a Psalm 5 release though so no emergency!

orklah:
Thanks!

lptn:
@orklah 
How about a new tag on master branch then? A new beta or even RC will be great before a major release.

orklah:
I made one. We should have released a few days ago already so I didn't bother but as we took some more time, We might as well make one

PR #8731, opened 2022-11-22
Allow parameter types to be contained by a class template type 

danog:
... in template function calls.

Not sure why the original logic checked for equality (in a weird way by actually checking A⊆B and B⊆A, not by actually running B==A with equals).

Fixes https://github.com/vimeo/psalm/issues/8732


orklah:
I never understood what's going on here. @muglug could you review this one?

muglug (approved):

PR #8736, opened 2022-11-23
Simplify issue sorting

weirdan:
As we're on 7.4 now, we can use more concise constructs

### DIFF
--- a/src/Psalm/IssueBuffer.php
+++ b/src/Psalm/IssueBuffer.php
@@ -574,21 +574,10 @@ final class IssueBuffer
             foreach (self::$issues_data as $file_path => $file_issues) {
                 usort(
                     $file_issues,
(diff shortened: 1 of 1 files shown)


src/Psalm/IssueBuffer.php
```diff
@@ -574,21 +574,10 @@ public static function finish(
...
-
-                                return $d1->column_from > $d2->column_from ? 1 : -1;
-                            }
-
-                            return $d1->line_from > $d2->line_from ? 1 : -1;
-                        }
-
-                        return $d1->file_path > $d2->file_path ? 1 : -1;
-                    }
+                    static fn (IssueData $d1, IssueData $d2): int =>
```
theodorejb, line 577: For consistency the `fn` keyword should not have a space after it.

orklah:
Thanks!

PR #8749, opened 2022-11-24
Fix #8748

### LINKED ISSUE #8748: Magic property is inferred as mixed after method call
https://psalm.dev/r/10eec1df24

This really surprised me, because it looks like Psalm is automatically creating some mistaken assertion on the property assigned in the `updateErrors` method, because if the assignment is removed from `updateErrors` the error is fixed: https://psalm.dev/r/4e00eed4c5

danog:


weirdan:
A test case for this would be nice to have.

orklah:
Seems great! Thanks!

Issue #8741, opened 2022-11-23
Simple example takes over 64 seconds to scan


danog:
https://psalm.dev/r/159cebbc4b

Most of the time is taken up by the `simplifyCNF` function, specifically in the array_diff call issued by `Clause::contains` which calls `__toString` on Assertions (which causes calls to __toString on Atomics) multiple times without memoizing the result.

Will investigate a bit more tomorrow to see if I can come up with an easy performance fix.

danog:
Seems like we're hitting the worst case 2^n expression explosion when converting DNF->CNF during negation.
Using [Tseytin](https://en.wikipedia.org/wiki/Tseytin_transformation) to convert to CNF should mitigate exponential explosion of terms, switching to isset also helps: https://psalm.dev/r/9daf0ab071

danog:
In this specific case, there's also the problem is that Psalm is useless doing a useless double-negation of `\array_key_exists($key, self::CRITICAL_ERRORS)`, which can be easily avoided if the original expression is known.

This is a performance issue that can be fixed by keeping references to the original expressions (including outer CNF expressions) during negation, will submit a PR.

PR #8750, opened 2022-11-24
Algebra refactoring

danog:
This fixes long scan times for problematic expressions like https://github.com/vimeo/psalm/issues/8741 by optimizing double negations by caching all negations of a CNF boolean expression.

This brings down the scan times in https://github.com/vimeo/psalm/issues/8741 from 64 to 2 seconds.

There's one place where I made a performance improvement by removing a loose comparison, but I'm not sure if this was intended (@muglug is this OK? :) https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/AlgebraAnalyzer.php#L99


danog:
Hmm, now that's a weird test to fail...

danog:
Actually, it's from 64 to 2 seconds :D

src/Psalm/Internal/ClauseConjunction.php
```diff
@@ -0,0 +1,730 @@
...
+    public array $clauses;
+
+    /**
+     * Negation of the current clause.
+     */
+    private ?self $negation = null;
+
+    public bool $isSimplified = false;
+
+    private static ?self $empty = null;
```
weirdan, line 42:
```suggestion
    private ?ClauseConjunction $negation = null;

    public bool $isSimplified = false;

    private static ?ClauseConjunction $empty = null;
```

I have a hunch the above may resolve phar build issue

danog: huh, let's give it a try!

weirdan: And it seems it did.

muglug:
Mind rebasing this PR and including performance differences when running `tests/TypeReconciliation/TypeAlgebraTest.php` and `tests/SwitchTypeTest.php`? Feel free to just compare cumulative analysis time for both (vs scanning & analysis time)

danog:
2.14s on this branch vs 2.34s on master (cumulative time for both tests), with the biggest difference in `tests/SwitchTypeTest.php` (1.19s vs 1.36s).

src/Psalm/Internal/ClauseConjunction.php
weirdan: The master now has `humbug/box` updated, which does not exhibit this issue.

muglug:
Mind merging master and updating the description now that https://github.com/vimeo/psalm/issues/8741 has been fixed?

danog:
Done!

psalm-baseline.xml
```diff
@@ -30,6 +30,11 @@
       <code>$matches[3]</code>
     </PossiblyUndefinedIntArrayOffset>
   </file>
+  <file src="src/Psalm/Internal/Analyzer/AlgebraAnalyzer.php">
```
muglug, line 33: It seems like these new things are because we're replacing `count($foo->clauses)` with `$foo->count()`, and I wonder whether the former is better? It's a bit more verbose, but it allows the typechecker to do its thing.

danog: Yeah makes sense, reverted.

muglug:
Run on the entire Psalm codebase with `--no-diff --threads=1` I see a 2.5% performance bump with this branch, but a 40% increase in memory. I'm not sure that 2.5% improvement is worth the large memory increase.

orklah:
I confirm that increasing the memory by 40% will probably be enough to make Psalm impossible to run for my work's development server! It's not worth 2.5% in speedup for now. If you can get those numbers to improve, it would be cool. I'm marking this as draft for now

PR #8752, opened 2022-11-24
Test CI on 8.2

orklah:
This is a first test for the CI in 8.2

@weirdan I expected to find 8.1 running for at least shepherd or some unit test(unix or windows) but apparently not. Do you remember if there's a reason why we wouldn't have tried running those in 8.1?


weirdan:
>  Do you remember if there's a reason

Not really. Perhaps we just never set that up.

orklah:
Probably then :)

Seems 8.2 won't be as easy as I thought. There are dependencies conflict for now so we'll have to wait

weirdan:
> Seems 8.2 won't be as easy as I thought.

Was more tedious than hard. On a positive note, we now have a Psalm 5-compatible plugin for Mockery.

PR #8754, opened 2022-11-24
improve docs and phrasing about NoValue

### LINKED ISSUE #8041: Misleading error after not empty assertion on empty array
https://psalm.dev/r/125683552a

The NoValue error complains that `This function or method call never returns output` for `$foo`, which isn't a function or method.

In my actual code the `TypeDoesNotContainType` didn't show up for some reason, which made it much more confusing. I think in this case it would probably be better if after the paradoxical `!empty` condition `$foo` were changed to `array<array-key, mixed>` instead of `never`.

orklah:
Feedback very welcome on this one, English phrasing is not my best skill!

This will fix #8041

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php
@@ -530,7 +530,7 @@ class AssignmentAnalyzer
             if ($context->vars_in_scope[$var_id]->isNever()) {
                 if (IssueBuffer::accepts(
                     new NoValue(
-                        'All possible type for this assignment were invalidated - This may be dead code',
+                        'All possible types for this assignment were invalidated - This may be dead code',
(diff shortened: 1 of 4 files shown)


orklah:
This will fix #8041

src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php
```diff
@@ -530,7 +530,7 @@ public static function analyze(
             if ($context->vars_in_scope[$var_id]->isNever()) {
                 if (IssueBuffer::accepts(
                     new NoValue(
-                        'This function or method call never returns output',
+                        'All possible type for this assignment were invalidated - This may be dead code',
```
weirdan, line 533:
```suggestion
                        'All possible types for this assignment were invalidated - This may be dead code',
```

weirdan (changes requested):

Issue #7387, opened 2022-01-13
A string after strlen > 0 check should be a non-empty-string


MauricioFauth:
https://psalm.dev/r/ca291dd273

orklah:
Good idea!

This is handled in AssertionFinder. The role of this class is to find any hint it can use to refine a variable type. For example `if(!empty($a))` will generate an assertion `!empty`. That way, once we enter the AssertionReconciler, we'll be able to refine the type of $a (previously `stdClass|null`) into `stdClass`.

AssertionFinder has 4 main methods:
getGreaterAssertions (for cases like $a > 5)
getLesserAssertions (for $a < 5)
scrapeInequalityAssertions (for $a !== 5)
scrapeEqualityAssertions (for $a === 5)
So every case must be taken into account for (mb_?)strlen

What you want to do is actually closer to what we already have with count (for example in `if(count($arr) > 0)`, so you need something similar to that.

the count() case is handled here for getGreaterAssertions:
https://github.com/vimeo/psalm/blob/e7bc6f273abef924b177b6b326fbf2565c096da6/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php#L3736

The method we call must detect when count() > 5 is used, and then extract an assertion from that

In your case, you'll want to detect any strlen used on a string and then deduct if you can infer something from it.
Here are the assertions you should use:
- strlen($a) > 5 should create an assertion `!empty` (to transform `string` into `non-empty-string`
- strlen($a) == 1 should create an assertion `!empty`
- strlen($a) != 0 should create an assertion `!empty`

So this means you don't have to change anything in getLesserAssertions.

Don't hesitate to ask questions or to ping me for more help or explanations!

weirdan:
The PR implementing this has been reverted, so I'm reopening the issue.

PR #8761, opened 2022-11-25
#7387 Add asserting non-empty-string by strlen

LeoVie:
This solves #7387.

I'm not quite sure, if I catched all cases... What do you think?


weirdan (commented):
All tests show `strlen()` call on the left-hand side of the expression; it would be great to have at least some of them checking the other direction, e.g. `0 < strlen($str)`.

Does it handle (or ignore) expressions as `strlen()` arguments? E.g. `strlen("a" . $str . "b") > 2`?

LeoVie:
Indeed, having `strlen` on the right-hand side or calling `strlen` with a concatenated string like in your example does not work, yet. I will take a look at that. :)

LeoVie:
I implemented both cases. There is some code redundancy between `strlen` and `count` assertions, now. That could be removed by extracting generic methods. But I am not sure, if this would lead to any benefit here, as the current implementation is very verbose and therefore quite easy to unterstand.

tests/TypeReconciliation/EmptyTest.php
```diff
@@ -417,6 +417,83 @@ function foo(array $arr): void {
...
+                        return 1 === strlen($str) ? $str : "string";
+                    }'
+            ],
+            'strlenOnRightHandSide' => [
+                'code' => '<?php
+                    /** @return non-empty-string */
+                    function nonEmptyString(string $str): string {
+                        return 0 < strlen($str) ? $str : "string";
+                    }'
+            ],
```
weirdan, line 482: Isn't it identical to `strlenRighthandWithGreaterZero`?

tests/TypeReconciliation/EmptyTest.php
```diff
@@ -417,6 +417,83 @@ function foo(array $arr): void {
...
+            'strlenWithConcatenatedString' => [
+                'code' => '<?php
+                    /** @return non-empty-string */
+                    function nonEmptyString(string $str): string {
+                        return strlen("a" . $str . "b") >= 2 ? "foo" : "string";
```
weirdan, line 487: I think you can remove this case and keep `preventStrlenWithConcatenatedString` only.

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
```diff
@@ -3836,6 +4035,39 @@ private static function getGreaterAssertions(
             return $if_types ? [$if_types] : [];
         }
 
+        if ($strlen_equality_position) {
+            if ($strlen_equality_position === self::ASSIGNMENT_TO_RIGHT) {
+                $strlened_expr = $conditional->left;
+            } else {
+                throw new UnexpectedValueException('$count_equality_position value');
```
weirdan, line 4042: Did you mean
```suggestion
                throw new UnexpectedValueException('$strlen_equality_position value');
```
?

weirdan (changes requested):

weirdan:
Also, can you check if Shepherd failures are actual code issues? 

https://github.com/vimeo/psalm/actions/runs/3559331115/jobs/5980058302

tests/TypeReconciliation/EmptyTest.php
LeoVie: In fact, it is. I remove that case.

tests/TypeReconciliation/EmptyTest.php
LeoVie: Removed that case, too.

src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
LeoVie: Yes, thank you. :)

LeoVie:
> Also, can you check if Shepherd failures are actual code issues?
> 
> https://github.com/vimeo/psalm/actions/runs/3559331115/jobs/5980058302

What's your suggestion on these issues?
They say, the call to `getArgs` and then accessing the first array item is not safe. That's right, in general. But here in this case, we know, that the expression is a `strlen` call, which always has exactly one argument.
```php
/** @var PhpParser\Node\Expr\FuncCall $strlen_expr */
$var_name = ExpressionIdentifier::getExtendedVarId(
    $strlen_expr->getArgs()[0]->value,
    $this_class_name,
    $source
);
```

I've seen similar errors in `psalm-baseline.xml`. Should I just add the false-positives there?

orklah:
> the expression is a strlen call, which always has exactly one argument

It will have exactly one parameter on valid code yes, but Psalm is used to detect invalid code :)

It's also used to analyse code being typed which may miss parts to be correct. Trying to access an undefined offset would be risky.

We should probably try to fix the ones in the baseline and avoid adding new ones :) Just check isset($strlen_expr->getArgs()[0]) and skip the assertion if it returns false

LeoVie:
I've added the proposed checks.

LeoVie:
Ah, missed one check. Psalm now reports no PR related errors locally, anymore.
Please give the workflow another try.

orklah:
There you go :)

CI is failing, not because of your change directly, but because there is an assertion in a loop that messes with other condition in that same loop.

@muglug if you could enlighten us on what can be done to avoid this kind of error, it would be great. I had the same kind of errors when working on integer ranges and I never could fix them correctly

LeoVie:
Hi, is there anything I can do to make this PR mergable? :)

orklah:
Yeah, put the error in the baseline, the behaviour is not related to this PR and is already broken anyway, it may be fixed later

LeoVie:
Hi, @orklah, I've added the two errors to the baseline. CI hopefully gets green, now :)

orklah:
Thanks!

weirdan:
This is causing a lot of issues, so I'm going to revert it and reopen the original ticket.

PR #8762, opened 2022-11-25
Add diagnostic message when shape fields are missing

muglug:
This is a small DX improvement so that developers who run into issues with Psalm 5 sealed array shapes have a bit more information about why the shapes were incompatible.

Before:

> Argument 1 of a expects array{test: string}, but array{somethingElse: 'test', test: 'str'} provided

After:

> Argument 1 of a expects array{test: string}, but array{somethingElse: 'test', test: 'str'} with additional array shape fields (somethingElse) was provided

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
@@ -56,6 +56,7 @@ use function array_diff;
 use function array_filter;
 use function array_values;
 use function count;
+use function implode;
 use function in_array;
 use function strpos;
 use function strtolower;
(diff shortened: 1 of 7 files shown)


weirdan:
How does that look with nested shapes? E.g. container type `array{data:array{id:int}}` with input type `array{data:array{di:int}}`.

muglug:
It should give the missing shapes I think, since they're just propagated up the `TypeComparisonResult`

weirdan:
What I was asking was whether Psalm would help me identify the specific nested key that was missing. By, e.g., printing the full path like `data.some.deeply.nested.array.shape.path.id` for
```
array{data:array{some:array{deeply:array{nested:array{shape:array{path:array{id:int}}}}}}}
vs 
array{data:array{some:array{deeply:array{nested:array{shape:array{path:array{di:int}}}}}}}
```

orklah:
Thanks for this! It also unlocks the possibilty of a specific psalter handling, when `missing_shape_fields` is not empty, then adding `fallback_params` to the keyed array should fix the issue

muglug:
@weirdan it doesn't show the nesting (but no reason it couldn't)

@orklah to unlock Psalter handling would require checking the positions of the types. That would require us to start adding more positional information to the `Atomic`.

It would be a non-trivial task, and would mean we'd probably also want to break this out into a separate issue (which we could do anyway to appease some folks)

PR #8766, opened 2022-11-25
Suppress errors when validation regexps

### LINKED ISSUE #8759: PHP Warning when configuring referencedMethod
### Problem
I have the following in my psalm.xml file.
```xml
<issueHandlers>
    <UndefinedInterfaceMethod>
        <errorLevel type="suppress">
            <referencedMethod name="ask"/>
        </errorLevel>
    </UndefinedInterfaceMethod>
</issueHandlers>
```

When running `psalm` the following error shows up.

> PHP Warning:  preg_match(): Delimiter must not be alphanumeric or backslash in /var/www/project/vendor/vimeo/psalm/src/Psalm/Config/FileFilter.php on line 440

After checking the [corresponding code](https://github.com/vimeo/psalm/blob/5.0.0-rc1/src/Psalm/Config/FileFilter.php#L434)  I think the idea was to suppress the warnings while validating the regexp. However, it doesn't work because the temporary error handler returns `false` which causes the default error handler to continue.

### Proposed solution
Return `true` from the error handler or use `@` to suppress any errors from `preg_match()`.

Chi-teck:
Fixes #8759

### DIFF
--- a/src/Psalm/Config/FileFilter.php
+++ b/src/Psalm/Config/FileFilter.php
@@ -434,7 +434,7 @@ class FileFilter
     private static function isRegularExpression(string $string): bool
     {
         set_error_handler(
-            static fn(): bool => false,
+            static fn(): bool => true,
             E_WARNING
         );
         $is_regexp = preg_match($string, '') !== false;


Chi-teck:
We might also consider using `@` to suppress the errors.

PR #8773, opened 2022-11-26
List refactoring (v6)

danog:
Fixes a bunch of bugs all over the codebase.


UPGRADING.md
```diff
@@ -1,6 +1,12 @@
 # Upgrading from Psalm 4 to Psalm 5
 ## Changed
 
+- [BC] Psalm 5.1 will switch its internal representation of `list<T>` and `non-empty-list<T>` from the TList and TNonEmptyList classes to an unsealed list shape: the TList, TNonEmptyList and TCallableList classes will be removed.
```
weirdan, line 4: I'd suggest to follow semver and name that version 6.0

danog: I actually was hoping to get this merged by Psalm 5.0 given https://github.com/psalm/psalm.dev/commit/dbe30c025aa59ad602b1b3a1ab7792da021679e4, that description is outdated

danog: Probably not gonna make it though, I'd still love to get at least https://github.com/vimeo/psalm/pull/8728 merged so people already stop using TList and switch over to keyed lists in time for 5.1

tests/Template/FunctionTemplateTest.php
```diff
@@ -291,7 +291,7 @@ function splat_proof(array $arr, array $arr2) {
 
                     $a = splat_proof(...$foo);',
                 'assertions' => [
-                    '$a' => 'array<int, int>',
+                    '$a' => 'array<int<0, 2>, int>',
```
muglug, line 294: I feel like int ranges where the `max` is 2 should render as `0|1|2`

orklah: Well, that's not just a display preference, it will have different effects on things afterwards (union of 3 literal ints are not always handled the same way as a range)

muglug: yeah, I get it, it was a bad feeling!

danog:
Wew, this PR triggered a huge amount of previously uncovered and broken edge cases, I fixed a bunch of bugs also present in master.
I've been thinking of merging this into master for v6, and then backporting a TList-compatible version (that removes TList-specific logic but keeps compatibility by converting plugin-originated TLists into the equivalent TKeyedArrays before handling, possibly directly inside the constructor of Union).

danog:
The HasArrayKey assertion turned out to be completely broken and actually useless, for example when asserting `array_key_exists($key, $arr)` it asserted the existance of `$arr['$key']` (not `$arr[$key]`), only to bypass a bug in class constant key assertions (which I also fixed in this PR).

danog:
Closing until the master branch becomes the v6

Issue #8746, opened 2022-11-23
Feature: Output issues sorted and grouped by severity level


bdsl:
When there are many issues, such as in a baseline or when adding Psalm to an existing project, it might be useful to see the issues in severity order and grouped by severity (i.e. which [error level](https://psalm.dev/docs/running_psalm/error_levels/) they would be displayed at).

This may help to get an understanding of the state of the project as well as to decide which error level to run Psalm at in future.

I might be able to find time to implement this if there's interest in it.

bdsl:
Done in https://github.com/vimeo/psalm/releases/tag/5.1.0

PR #8774, opened 2022-11-26
Report by issues grouped by level and type

bdsl:
#8746 

This adds a new output format, `--output-format=by-issue-level`, which is almost exactly the same as the default console output, except that:

- Issues are sorted by error level then issue type, rather than by file and position in file
- Error level number is included in output, e.g. `ERROR (3): PossiblyNullArgument `

This is intended to be useful for getting an overview of a project, particularly when onboarding Psalm to a new codebase, and for choosing what error level to set in psalm config. The most severe errors (plus any reported due to feature specific config) appear first in the output, with the issues that would only be reported at error level 1 appearing at the end.

I wasn't sure where to document this, so for now I've made the report include a block of explanation as a heading. Happy to move that text to a more appropriate place.

### DIFF
--- a/src/Psalm/Report/ByIssueLevelAndTypeReport.php
+++ b/src/Psalm/Report/ByIssueLevelAndTypeReport.php
@@ -180,8 +180,11 @@ HEADING;
     {
         usort($this->issues_data, function (IssueData $left, IssueData $right): int {
             // negative error levels go to the top, followed by large positive levels, with level 1 at the bottom.
-            return [$left->error_level > 0, -$left->error_level, $left->type, $left->file_path, $left->file_name, $left->line_from] <=>
-                [$right->error_level > 0, -$right->error_level, $right->type, $right->file_path, $right->file_name, $right->line_from];
+            return [$left->error_level > 0, -$left->error_level, $left->type,
+                    $left->file_path, $left->file_name, $left->line_from]
+                <=>
+                [$right->error_level > 0, -$right->error_level, $right->type,
+                    $right->file_path, $right->file_name, $right->line_from];
         });
     }
 }
--- a/src/Psalm/Report/ByIssueLevelAndTypeReport.php
+++ b/src/Psalm/Report/ByIssueLevelAndTypeReport.php
@@ -44,8 +44,6 @@ final class ByIssueLevelAndTypeReport extends Report
 
 HEADING;
 
-        ;
-
         foreach ($this->issues_data as $issue_data) {
             $output .= $this->format($issue_data) . "\n" . "\n";
         }
@@ -54,7 +52,7 @@ HEADING;
     }
 
     /**
-     * Copied from ConsoleReport with only very minor changes. Todo consider reducing code duplication
+     * Copied from ConsoleReport with only very minor changes.
      */
     private function format(IssueData $issue_data): string
     {
@@ -106,7 +104,7 @@ HEADING;
     }
 
     /**
-     * Copied from ConsoleReport unchanged. Todo consider reducing code duplication
+     * Copied from ConsoleReport unchanged. We could consider moving to another class to reduce duplication.
      * @param non-empty-list<DataFlowNodeData|array{label: string, entry_path_type: string}> $taint_trace
      */
     private function getTaintSnippets(array $taint_trace): string
@@ -141,6 +139,7 @@ HEADING;
     }
 
     /**
+     * Copied from ConsoleReport unchanged. We could consider moving to another class to reduce duplication.
      * @param IssueData|DataFlowNodeData $data
      */
     private function getFileReference($data): string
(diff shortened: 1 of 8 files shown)


src/Psalm/Report/ByIssueSeverityReport.php
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        $link = strtr($this->link_format, ['%f' => $data->file_path, '%l' => $data->line_from]);
+        // $reference = $data->file_name . ':' . $data->line_from . ':' . $data->column_from;
+
+
+        return "\033]8;;" . $link . "\033\\" . $reference . "\033]8;;\033\\";
+    }
+
+    public function sortIssuesByLevelAndType(): void
```
weirdan, line 159:
```suggestion
    private function sortIssuesByLevelAndType(): void
```

src/Psalm/Report/ByIssueSeverityReport.php
```diff
@@ -0,0 +1,180 @@
...
+                }
+
+                if ($rightLevel > 0) {
+                    return -1;
+                }
+
+                return $leftLevel <=> $rightLevel;
+            }
+
+            return $left->type <=> $right->type;
```
weirdan, line 177: isn't it equivalent to `return [$left->error_level, $left->type] <=> [$right->error_level, $right->type]`?

bdsl: I think you're right, I didn't realise you can use comparison operators with arrays.

bdsl: Turns out this was almost right, but not quite. complication is that negative levels sort to the top, small positive levels sort to the bottom.

src/Psalm/Report/ByIssueLevelAndTypeReport.php
```diff
@@ -0,0 +1,188 @@
...
+|----------------------------------------------------------------------------------------|
+|    Issues have been sorted by level and type. Feature-specific issues and the          |
+|    most serious issues that will always be reported are listed first, with             |
+|    remaining issues in level order. Issues near the top are usually the most serious.  |
+|    Reducing the errorLevel in psalm.xml will suppress output of issues further down    |
+|    this report.                                                                        |
+|                                                                                        |
+|    The level at which issue is reported as an error is given in brackets - e.g.        |
+|    `ERROR (2): MissingReturnType` indicates that MissingReturnType is only reported    |
+|    as an error when Psalm's level is set to 4 or below.                                |
```
weirdan, line 37: Is it level 2 or level 4?

bdsl:
I'm not sure why the composer update inside the Code Style Analysis job is failing - I don't think I've changed anything that should affect that.

tests/ByIssueLevelAndTypeReportTest.php
```diff
@@ -0,0 +1,97 @@
...
+ERROR (4): SomeLevel4IssueType - file.php:1:1 - message
+
+
+ERROR (4): SomeLevel4IssueType - file.php:1:1 - message
+
+
+ERROR (2): SomeLevel2IssueType - file.php:1:1 - message
+
+
+ERROR (1): SomeIssueType - file.php:1:1 - message
```
weirdan, line 68: This looks a little sparse. Wouldn't it be more readable if we had 1 or two empty lines between levels and none between the issues of the same level?

```
ERROR: SomeAlwaysReportedIssueType - file.php:1:1 - message
ERROR: SomeFeatureSpecificIssueType - file.php:1:1 - message

ERROR (7): SomeLevel7IssueType - file.php:1:1 - message

ERROR (4): AnotherLevel4IssueType - file.php:1:1 - message
ERROR (4): SomeLevel4IssueType - file.php:1:1 - message
ERROR (4): SomeLevel4IssueType - file.php:1:1 - message

ERROR (2): SomeLevel2IssueType - file.php:1:1 - message

ERROR (1): SomeIssueType - file.php:1:1 - message
```

tests/ByIssueLevelAndTypeReportTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+
+
+EXPECTED, $report->create());
```
weirdan, line 72: BTW now that we require PHP >7.3 you can indent the heredoc terminator.

weirdan:
> I'm not sure why the composer update inside the Code Style Analysis job is failing

I think we need to add an explicit `COMPOSER_ROOT_VERSION=dev-master` env variable to composer install / update. Otherwise composer won't be able to guess the root package version and fail the install where there circular dependencies are involved.

tests/ByIssueLevelAndTypeReportTest.php
bdsl: It's basically the same format as the console report, it just looks sparse in the test because the code snippets are empty.

tests/ByIssueLevelAndTypeReportTest.php
bdsl: Done

src/Psalm/Report/ByIssueLevelAndTypeReport.php
bdsl: It's level 2, thanks for catching that. Fixed in 54239838f.

weirdan:
@bdsl the `docs/running_psalm/error_levels.md` file is generated by `bin/generate_levels_doc.php` script, so perhaps you need to update that too.

Issue #8775, opened 2022-11-26
Add deprecation mechanism for issues


weirdan:
There could be cases when we want to get rid of a specific issue completely (e.g. `MixedAssignment` because it outlived its usefulness).

We need to make sure that:
1. The issue class is marked as `@deprecated` (this is a starting point for issue deprecation process)
2. For all issues marked as `@deprecated` we have a corresponding marker in `config.xsd` (our test suite should verify it)
3. Whenever a deprecated issue is referenced in `psalm.xml` we should emit `ConfigIssue` (and we need a test for that)
4. Whenever a deprecated issue is used in `@psalm-suppress` we need to emit `DeprecatedSuppression` (and we need a test for that too)
5. Whenever a deprecated issue is used in the baseline, we should ignore it.

PR #8776, opened 2022-11-26
Remove MixedAssignment

muglug:
`MixedAssignment` is largely unnecessary now that we report where [mixed data comes from](https://psalm.dev/r/fa55323580).

This PR removes the issue `MixedAssignment` and its corresponding class `Psalm\Issue\MixedAssignment`.

If this PR passes, `@psalm-suppress MixedAssignment` annotations will now generate an `UnusedPsalmSuppress` issue if the `findUnusedPsalmSuppress` option is true or `--find-unused-psalm-suppress` is passed as a command-line flag.

This would be a breaking change for a major version. If it's too late for Psalm 5, then I'll modify this PR to instead just deprecate the class and associated methods (while keeping the `UnusedPsalmSuppress` issue since it will only affect a small number of people who track unused suppressions).


config.xsd
```diff
@@ -320,13 +320,13 @@
             <xs:element name="MixedArrayAssignment" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="MixedArrayOffset" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="MixedArrayTypeCoercion" type="IssueHandlerType" minOccurs="0" />
-            <xs:element name="MixedAssignment" type="IssueHandlerType" minOccurs="0" />
```
weirdan, line 323: It should remain here, but be marked as deprecated. See https://github.com/orklah/psalm/blob/24ae96b373cb44186c58360543dcf0987436df0c/config.xsd#L20-L25 for example

src/Psalm/Issue/MixedAssignment.php
```diff
@@ -1,11 +0,0 @@
-<?php
-
-namespace Psalm\Issue;
-
-final class MixedAssignment extends CodeIssue implements MixedIssue
```
weirdan, line 5: The class has to remain here, but be marked as `@deprecated`. It's part of our API surface.

src/Psalm/IssueBuffer.php
```diff
@@ -341,50 +341,6 @@ public static function add(CodeIssue $e, bool $is_fixable = false): bool
...
-            }
-        }
-
-        self::$recorded_issues[self::$recording_level] = $filtered_issues;
-    }
-
-    /**
-     * This will try to remove an issue that has been added for emission
-     */
-    public static function remove(string $file_path, string $issue_type, int $file_offset): void
```
weirdan, line 363: Public functions of non-internal classes have to be marked `@deprecated` and can only be removed in a major release.

tests/CallableTest.php
```diff
@@ -1327,6 +1327,11 @@ abstract class TestClass {
                         use TestTrait;
                     }'
             ],
+            'stringFunctionCall' => [
+                'code' => '<?php
+                    $bad_one = "hello";
+                    $a = $bad_one(1);',
```
weirdan, line 1333: Shouldn't it complain about the function being unknown? I wouldn't keep this as a valid test.

tests/CloneTest.php
```diff
@@ -45,6 +45,19 @@ public function foo(): self {
...
+                    class A {}
+                    /**
+                     * @param A|string $a
+                     */
+                    function foo($a): void {
+                        /**
+                         * @psalm-suppress PossiblyInvalidClone
+                         */
+                        $cloned = clone $a;
+                    }',
```
weirdan, line 59: I think this test needs to return `$cloned`. and make sure MixedReturn is reported.

tests/ErrorBaselineTest.php
```diff
@@ -67,34 +62,6 @@ public function testLoadShouldParseXmlBaselineToPhpArray(): void
         );
     }
 
-    public function testLoadShouldIgnoreLineEndingsInIssueSnippet(): void
```
weirdan, line 70: This test is still valid, we just need to use a different issue.

weirdan (changes requested):

weirdan:
Overall I'm not sure whether you wanted to remove the issue or deprecate it. This PR seems to do a little of both, but none completely.

muglug:
I want to remove the issue, but not break psalm.xml or `@psalm-suppress ...` uses of it.

weirdan:
I'm a little worried about tests that may start passing even with the broken code because they previously relied on MixedAssignment to detect problematic code. I've already found one by accident (fixed in #8781), but there's no telling how many more there are.

weirdan:
> I want to remove the issue, but not break psalm.xml or @psalm-suppress ... uses of it.

I don't mind keeping the issue class, XSD entry and docs for 5.x release cycle, as long as we stop emitting this issue. Could actually use it as a testing ground for #8775

src/Psalm/Issue/MixedAssignment.php
muglug: I think it should be safe to remove the class between major versions — I don't expect plugins to rely on its existence.

config.xsd
muglug: As discussed, it doesn't seem like anyone references it in `psalm.xml`

weirdan: Github search shows 97 entries, including some popular repos like `doctrine/collections`: https://github.com/search?q=path%3Apsalm.xml+language%3AXML+MixedAssignment&type=code

src/Psalm/Issue/MixedAssignment.php
weirdan: I can confirm none of the plugins available on Packagist use `MixedAssignment` in the code:
```console
> fetch https://packagist.org/packages/list.json?type=psalm-plugin | get packageNames | each {|it| fetch $"https://packagist.org/packages/($it).json"} | get package.repository | each { str replace 'https://github.com/' ''} | save plugin_repos.json

> open plugin_repos.json | each {|it| {repo: $it dir: ($it | str replace "/" "---")} } | each {|it| gh repo clone $it.repo $it.dir }

> rg --pcre2 '(?<!-suppress )(?<!\<)MixedAssignment'
psalm---psalm-plugin-symfony/tests/acceptance/acceptance/Response.feature
7:  Scenario: MixedAssignment error about $statusTexts is not raised

psalm---psalm-plugin-symfony/tests/acceptance/acceptance/DenormalizerInterface.feature
41:      | MixedAssignment        | Unable to determine the type that $result is being assigned to |

psalm---psalm-plugin-symfony/tests/acceptance/acceptance/serializer/SerializerInterface6.feature
41:      | MixedAssignment        | Unable to determine the type that $result is being assigned to |

psalm---psalm-plugin-symfony/tests/acceptance/acceptance/console/ConsoleOption.feature
297:      | MixedAssignment | Unable to determine the type that $option is being assigned to |

psalm---psalm-plugin-symfony/tests/acceptance/acceptance/serializer/SerializerInterface.feature
41:      | MixedAssignment        | Unable to determine the type that $result is being assigned to |

psalm---psalm-plugin-symfony/tests/acceptance/acceptance/forms/FormGetErrors.feature
70:      | MixedAssignment         | Unable to determine the type of this assignment                                     |

psalm---symfony-psalm-plugin/tests/acceptance/acceptance/console/ConsoleOption.feature
297:      | MixedAssignment | Unable to determine the type that $option is being assigned to |

psalm---symfony-psalm-plugin/tests/acceptance/acceptance/forms/FormGetErrors.feature
70:      | MixedAssignment         | Unable to determine the type of this assignment                                     |

psalm---symfony-psalm-plugin/tests/acceptance/acceptance/DenormalizerInterface.feature
41:      | MixedAssignment        | Unable to determine the type that $result is being assigned to |

psalm---symfony-psalm-plugin/tests/acceptance/acceptance/Response.feature
7:  Scenario: MixedAssignment error about $statusTexts is not raised

psalm---symfony-psalm-plugin/tests/acceptance/acceptance/serializer/SerializerInterface6.feature
41:      | MixedAssignment        | Unable to determine the type that $result is being assigned to |

psalm---symfony-psalm-plugin/tests/acceptance/acceptance/serializer/SerializerInterface.feature
41:      | MixedAssignment        | Unable to determine the type that $result is being assigned to |

Practically---PsalmPluginYii2/tests/acceptance/PropertyAccessor.feature
136:      | MixedAssignment | Unable to determine the type that $contentType is being assigned to |
```

Two use it in their tests, as shown above.

Issue #8778, opened 2022-11-27
Variables should not be namespaced


weirdan:
https://psalm.dev/r/0cbef292cd

https://3v4l.org/2qvrC

PR #8779, opened 2022-11-27
Variables should outlive namespaces

weirdan:
Variables in PHP are not namespaced. In other words, namespaces share the context with the file they are located in.

See https://3v4l.org/2qvrC

Fixes vimeo/psalm#8778

Depends on #8780


muglug:
Thanks!

PR #8780, opened 2022-11-27
Replace all references to static variables when moving class

weirdan:
Previously `psalter` would only replace the first occurrence of the property access. Now it replaces all occurrences (by keeping the old name unknown in the context).


muglug:
Thanks!

PR #8792, opened 2022-11-30
Fixes vimeo#8112

emmanuelGuiton:
Added 2 cases in the nested template tests.
Fixed the issue for these 2 cases by adding 2 short-circuit operations in the `ObjectComparator` class when we compare identical template parameters of type `TNamedObject` or `TMixed`.
Not sure that's the way it should be implemented, but at least the tests pass.


orklah:
Sorry for not getting back to you earlier.

Seems like some tests are failing. Can you take a look?

emmanuelGuiton:
@orklah Indeed, I had not run all the tests locally (due to some laziness, the process is quite long) and I forgot to check back the results of the CI. Thanks for the reminder. Should be OK now.

emmanuelGuiton:
Looks like the label is the last required piece.
Can't edit that unless I missed something.

orklah:

PR #8798, opened 2022-11-30
fix sealed combination

### LINKED ISSUE #8719: Psalm incorrectly identifying array as non-empty-array
https://psalm.dev/r/d2a5d7fe21

I don't know why it's doing this, since trace shows the condition (`$v`) is `null|object`

orklah:
This will fix #8719

We were creating an non-empty-array based on the fact that there were sealed properties without actually checking if they could be undefined

@danog could you review that? Not sure if it was the best move here


danog:
LGTM!
The logic seems correct to me, even if I would've used a simple ! instead of !== true :)

PR #8820, opened 2022-12-03
List refactoring v5

danog:
Fixes a huge number of list-related and non-list-related bugs.


danog:
Whoops sorry muglug wanted to tag just orklah :)

orklah:
I didn't start reviewing this yet.

I'm a little worried about possible BC breaks. Wouldn't it be better to make a new sweep of changes and ship it as Psalm 6? 

We're not obligated to any kind of waiting period between major and it would probably be cleaner, no?

muglug:
@danog I'd strongly encourage you to close this PR and reopen one with a single commit (or, if it makes sense, a small number of curated commits).

When I review a PR, I generally look at commit messages to figure out what's going on. This PR has 134 commits, which is incredibly noisy (I had to unsubscribe from this PR because it was spamming my inbox).

tests/ArrayFunctionCallTest.php
```diff
@@ -651,7 +667,7 @@ function foo(array $arr) {
                     '$b' => 'int',
                 ],
             ],
-            'arrayPopNonEmptyAfterMixedArrayAddition' => [
+            'SKIPPED-arrayPopNonEmptyAfterMixedArrayAddition' => [
```
muglug, line 670: Why is this skipped?

danog:
Sorry about the noise, I just prefer to use GHA to run unit tests, will submit PRs with fewer and more relevant squashed commits in the future.

Squashed everything, the main changes are documented in UPGRADING.md, the rest is just a lot of bugfixing for bugs detected using the existing unit tests.

tests/ListTest.php
```diff
@@ -191,7 +191,7 @@ function getKey() {
 
                     $a = [getKey() => 1];
                     takesList($a);',
-                'error_message' => 'MixedArgumentTypeCoercion',
+                'error_message' => 'InvalidArgument',
```
muglug, line 194: This *could* be wrong — some form of argument type coercion, since it's coercing a list with a typed array with a key that could be 0. But OTOH if you wanted a simple list you'd remove the `getKey()` call altogether. So I'll allow it :)

tests/ReturnTypeTest.php
```diff
@@ -1298,7 +1298,7 @@ function foo(): array {
                       }
                       return $arr;
                     }',
-                'error_message' => 'LessSpecificReturnStatement',
+                'error_message' => 'InvalidReturnStatement',
```
muglug, line 1301: This feels like a change that should be in latest Psalm 5.x — can you extract whatever changes you had to make in a separate PR please?

tests/ThisOutTest.php
```diff
@@ -85,6 +83,10 @@ public function getData(): array { return $this->data; }
                     '$data1===' => 'list<1>',
                     '$data2===' => 'list<2>',
                     '$data3===' => 'list<2|3>',
+                ],
+                'ignored_issues' => [
```
muglug, line 87: as a general rule I like to not use `ignored_issues` for new issues, explicitly adding `@psalm-suppress` in the exact places where issues crop up

tests/TypeReconciliation/TypeTest.php
```diff
@@ -1710,7 +1710,7 @@ function returnsFalse() { return rand() % 2 > 0; }
 
                     if (count($a) < 1) {
                     }',
-                'error_message' => 'RedundantConditionGivenDocblockType'
+                'error_message' => 'DocblockTypeContradiction'
```
muglug, line 1713: This also feels like something that should be in Psalm 5.x

tests/ArrayAssignmentTest.php
```diff
@@ -2335,9 +2350,10 @@ function getList(array $list): array {
 
                         return $list;
                     }',
-                'error_message' => 'LessSpecificReturnStatement',
+                'error_message' => 'InvalidReturnStatement',
```
muglug, line 2353: I think this should still be a `LessSpecificReturnStatement` — we're returning an `array<int, string>`, which is a supertype of `list<string>`.

tests/ArrayAccessTest.php
```diff
@@ -1355,7 +1355,7 @@ function takesList(array $arr) : array {
 
                         return $arr;
                     }',
-                'error_message' => 'LessSpecificReturnStatement',
+                'error_message' => 'InvalidReturnStatement',
```
muglug, line 1358: This is correct IMO, but I _only_ think this works in Psalm's type system if we're willing to go further here and say that calling `unset` for any list element would turns it into a (new-in-Psalm) `map-array<int, string>`, which is a subtype of `array<int, string>` and disjoint from `list<string>`

danog:
About BC breaks, I actually wanted to ask about the timeline for the Psalm v6 release/Psalm v5 feature freeze, I really think this should be planned out a bit (like PHP itself does) to avoid an excessively postponed release like for v5.

Obviously personally I would love to merge in all BC breaks right away and release v6 (Phpstan has it easier at least visually with 0.x semver), but if that can't be done I can just revert the remaining getGenericArray BC break.

tests/ArgTest.php
```diff
@@ -630,7 +630,7 @@ function foo(int ...$values): array
                         return $values;
                     }
                 ',
-                'error_message' => 'MixedReturnTypeCoercion',
+                'error_message' => 'InvalidReturnStatement',
```
muglug, line 633: This again is incorrect — `array<array-key, int>` is a supertype of `list<int>`

muglug (commented):
Just reviewing the test changes (always the easiest part of a large PR to review) saw some things to improve/break out

tests/ReturnTypeTest.php
danog: I removed the BC break from this PR, this PR can now be directly merged in 5.x :)

tests/TypeReconciliation/TypeTest.php
danog: I removed the BC break from this PR, this PR can now be directly merged in 5.x :)

tests/ArrayFunctionCallTest.php
danog: With this PR `$a[] = 'hello';` still generates an `array<array-key, mixed>` as result, fixing this issue caused other actual bugs so I just skipped this testcase, given that `$a[] = 'hello';` on an `array<array-key, mixed>` still technically gives an `array<array-key, mixed>` as result.

tests/ArrayAssignmentTest.php
danog: Actually this test was skipped because it previously relied on a incorrect ref-out type for `array_pop` which was fixed in this PR.
Now `array_pop` on a list correctly keeps the listness of the array (when previously it was turned into an array), but I couldn't find an easy way to make this complex example work with the new technically correct behavior of array_pop:

```
[daniil@daniil-arch psalm]$ php a.php ; cat a.php
array(3) {
  [0]=>
  string(1) "0"
  [1]=>
  string(1) "1"
  [2]=>
  string(1) "2"
}
array(2) {
  [0]=>
  string(2) "0!"
  [1]=>
  string(1) "1"
}
array(2) {
  [0]=>
  string(2) "0!"
  [1]=>
  string(2) "1!"
}
array(2) {
  [0]=>
  string(2) "0!"
  [2]=>
  string(2) "2!"
}
<?php
                    /**
                     * @param list<string> $list
                     * @return list<string>
                     */
                    function getList(array $list): array {
                        foreach ($list as $key => $value) {
                            var_dump($list);
                            array_pop($list);

                            $list[$key] = $value . "!";
                        }

                        return $list;
                    }
                    var_dump(getList(["0", "1", "2"]));
```

danog: Here array_pop pops the 1! value at key 1, still keeping listness, but the next line sets a 2! value at key 2, breaking listness: I couldn't think of an easy way to make Psalm correctly infer the non-listness of the return type.

tests/ArrayAccessTest.php
danog: Hmm, with this PR psalm correctly infers the type of $arr to be `array<int<1, max>, string>`, which is not a list, how is an InvalidReturnStatement wrong here?

danog: Fixed!

tests/ArrayAssignmentTest.php
danog: Will try fixing this really nasty edge case in a future PR, if at all possible: right now the dependent list key logic works exactly the same as before, the only reason why this test is broken is due to the new (correct) behavior of array_pop.

danog:
Do you think it would be OK to merge this? :)

muglug:
Mind looking at the `test-with-real-projects` job to see what's up there? it's been failing for over a week

danog:
I already checked them and prepared https://github.com/psalm/endtoend-test-phpunit/pull/7 and https://github.com/psalm/endtoend-test-psl/pull/3 to fix

muglug:
Thanks! I've just merged those — two remaining errors in the PHPUnit run (btw, don't feel shy about merging PRs yourself in the endtoendtest- repos)

danog:
Hehe I'd love to, but I don't have permission to merge PRs in the e2e or plugin repos in the psalm org.
Come to think of it, I don't have merge perms here, but that's OK I guess, since most of my PRs are quite big refactorings that require review from the other maintainers as well :)

muglug:
I've added you!

orklah:
Thanks!

PR #8823, opened 2022-12-03
Update dictionaries: use more specific boolean types

lptn:
functions updated:
 - imagecolorset https://www.php.net/manual/en/function.imagecolorset.php
 - mysqli_ssl_set https://www.php.net/manual/en/mysqli.ssl-set.php
 - restore_exception_handler https://www.php.net/manual/en/function.restore-exception-handler.php


orklah:
Thanks!

PR #8826, opened 2022-12-04
Add validation for Class::Method names in CallMap

othercorey:
We only validate classes that exist as the original provider only validate functions that exist.

Unreflectable methods are handled and validated. Sometimes these are methods that were removed in PHP and should be removed from the CallMap, but there are some like '__invoke' that are unreflectable.

Some of the methods in CallMap are missing return types, but this just ignores them for now.


dictionaries/CallMap.php
```diff
@@ -3211,6 +3211,18 @@
...
+'Fiber::resume' => ['mixed', 'value='=>'null|mixed'],
+'Fiber::throw' => ['mixed', 'exception'=>'Throwable'],
+'Fiber::isStarted' => ['bool'],
+'Fiber::isSuspended' => ['bool'],
+'Fiber::isRunning' => ['bool'],
+'Fiber::isTerminated' => ['bool'],
+'Fiber::getReturn' => ['mixed'],
+'Fiber::getCurrent' => ['?self'],
+'Fiber::suspend' => ['mixed', 'value='=>'null|mixed'],
+'FiberError::__construct' => ['void'],
```
othercorey, line 3225: Moved to match alphabetical order.

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -49,63 +52,162 @@ class InternalCallMapHandlerTest extends TestCase
      * @var array<int|string, string|list<string>>
      */
     private static $ignoredFunctions = [
```
othercorey, line 54: These lists were generated so might need to update local environment to catch extensions that were added/supported.

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -125,113 +227,319 @@ class InternalCallMapHandlerTest extends TestCase
...
         'mysqli_poll',
         'mysqli_real_connect',
+        'mysqli_stmt::__construct',
+        'mysqli_stmt::bind_param',
         'mysqli_stmt_bind_param',
+        'normalizer::getrawdecomposition',
+        'normalizer::isnormalized',
+        'normalizer::normalize',
         'normalizer_get_raw_decomposition',
-        'oauth_get_sbs',
```
othercorey, line 180: These should be re-added. I will work on an updated list.

orklah: do you want to do that now or in a next PR?

orklah:
Seems great! A few errors to solve in CI and it can be merged!

tests/Internal/Codebase/InternalCallMapHandlerTest.php
othercorey: Let me check this list before merging.

othercorey: The arrays should show only additions now.

We have problems where certain functions aren't ignored for specific PHP versions. If you run the test with PHP 8.1 it will tell you to remove some functions from the ignore list even though they will fail with PHP 8.0. This can be addressed separately.

orklah: Yeah, that's an issue, CI is failing because of that. Maybe we should build the list with the version psalm uses for tests now and enforce that version in tests? The day where we want to change it, we'll have to update the list with the correct version?

othercorey: Yea, I'll look into it some more. 

I'm pretty sure this tells us what functions/methods changed  in 8.1 though :).

othercorey: The php 8.0 installed in github actions on windows is returning reflection different from all the other tests environments I can find.

For example `phardata::compress` returns `null` for `getReturnType()` everywhere else except here.

othercorey: @orklah Updated the ignore arrays to support php versions like the original $ignoredFunctions does. This should work for php 8.0 - 8.2, but there are probably a couple missing from php 8.2.

Looks like your github actions config doesn't enable most of the extensions anymore, but I generated the lists with as as many extensions as I could load.

othercorey: Most of the methods not ignored for PHP 8.0 are missing reflection information. It doesn't exactly mean that CallMap is right or wrong.

orklah:
Thanks!

PR #8827, opened 2022-12-04
Improve source autodiscovery failure message

weirdan:
The message we emitted previously did not provide any instructions on
how to resolve the issue, leaving users to guess what they need to do.


orklah:
Thanks!

Issue #8797, opened 2022-11-30
Psalm 5: Unable to determine the type for ZipArchive property


kamil-tekiela:
I don't know if this is a more general bug, but I was able to reproduce it with `ZipArchive`.  The type should obviously be int as specified in the PHP manual. 

https://psalm.dev/r/57dc6bd858

kamil-tekiela:
I believe it's a regression as I only observed it today after upgrading to Psalm 5.

danog:
The problem is just an incorrect casing of the property in Psalm's stubs: https://psalm.dev/r/542226d4a0

orklah:
I was pretty sure those were case insensitive (and actually all lowercased in the file). Maybe something changed...

It should be case sensitive however, as property are actually case sensitive in PHP itself...

orklah:
Something did change. I actually forgot about that: https://github.com/vimeo/psalm/pull/8000/files#diff-05767d078a39dade308500b084250222a67729fbf3db16ffb1aaf0247724755dL34

We're bound to have some more property that were lowercased by error in there

PR #8831, opened 2022-12-04
capitalize properties

orklah:
Properties became case sensitive in propertyMap (because they are in PHP) but we forgot to fix the casing for existing properties. I probably missed a few, but this should solve the majority of them

fix https://github.com/vimeo/psalm/issues/8797


orklah:
fix https://github.com/vimeo/psalm/issues/8797

PR #8833, opened 2022-12-04
Integrate FidryCpuCoreCounter

theofidry:
See https://github.com/theofidry/cpu-core-counter/issues/11 for the motivations.

The code should be sensibly the same, the notable differences are:

- a runtime exception `NumberOfCpuCoreNotFound` is thrown instead of `LogicException` (can easily be changed)
- fixed the deprecated usage of `hw.ncpu`

I would also wait on 1.x before merging this if you're happy with it. Technically the current version is fine, I'm just looking for a round of feedback before publishing 1.0.


orklah:
Yeah, I'm pretty happy with that, I'd hate to have to tweak this somehow if something were to change on that front. Please ping us when 1.0 is released so we can merge this :)

Thanks for the comments on previous checks before. Shouldn't those checks be included in your library too?

theofidry:
>Thanks for the comments on previous checks before. Shouldn't those checks be included in your library too?

Not really because this is really specific to Psalm, although I admit I don't know the _why_. For any of those items the library does work though and so would spawning several processes

orklah:
ok! Looking at the comments, I thought this was a common known issue with multi threading

src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
```diff
@@ -1468,23 +1470,17 @@ public function getFunctionLikeAnalyzer(
      * Adapted from https://gist.github.com/divinity76/01ef9ca99c111565a72d3a8a6e42f7fb
      * returns number of cpu cores
      * Copyleft 2018, license: WTFPL
-     * @throws RuntimeException
-     * @throws LogicException
+     * @throws NumberOfCpuCoreNotFound
      * @psalm-suppress ForbiddenCode
```
sanmai, line 1474:
```suggestion
```

theofidry:
Should be good; there is now a distinction possible between physical and logical cores. The default picked is logical (as PHP source does and as was done previously here).

If you fancy checking out what finder finds what, to find a bug or play around with the order or something, I can recommend [those commands](https://github.com/theofidry/cpu-core-counter#checks-what-finders-find-what-on-your-system)

orklah:
Thanks!

PR #8835, opened 2022-12-04
Extend impure_functions list by socket functions

### LINKED ISSUE #8818: UnusedFunctionCall on socket* functions
Hello everyone,

I noticed that Psalm complains about my use of socket functions:

for example:
`socket_set_block($socket);`
results in:
`The call to socket_set_block is not used (see https://psalm.dev/206)`

Based on the PHP documentation, the functions socket_bind, socket_set_block and socket_set_nonblock, can be used like that when the return code is irrelevant.
https://www.php.net/manual/en/function.socket-set-nonblock.php

Relevant snippet: https://psalm.dev/r/43f28bcd5d

Thank you.

lptn:
This prevents UnusedFunctionCall for some socket ext functions.

Result of all of these functions is boolean and can be ignored.

Note, socket errors can be fetched by:
 - `socket_strerror`
 - `socket_last_error`

Closes #8818

### DIFF
--- a/src/Psalm/Internal/Codebase/Functions.php
+++ b/src/Psalm/Internal/Codebase/Functions.php
@@ -429,7 +429,8 @@ class Functions
             // stream/socket io
             'stream_context_set_option', 'socket_write', 'stream_set_blocking', 'socket_close',
             'socket_set_option', 'stream_set_write_buffer', 'stream_socket_enable_crypto', 'stream_copy_to_stream',
-            'stream_wrapper_register',
+            'stream_wrapper_register', 'socket_connect', 'socket_bind', 'socket_set_block', 'socket_set_nonblock',
+            'socket_listen',
 
             // meta calls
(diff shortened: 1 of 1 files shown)


orklah:
Thanks!

danog:
Hmm this does not feel like a good idea, generally you should always check if a network operation failed or not, are we sure this is a good idea?

lptn:
@danog 
This is opinionated, agree. 

In this PR I simply continues on what we already do here. Is it good a bad - it's a separate topic.
IMO ideally it should throw a new type of error, e.g. `UnusedOptionalFunctionCallResult`, so a user can easily suppress any such error.

danog:
I see now that a bunch of other socket functions were already included in the list, and that the list is actually indicating impure functions, so this PR actually makes sense, as it is (correctly) indicating that these functions have side effects.

However, ignoring the result of at least the write, read, connect and bind functions is still an antipattern, it might be worth introducing an UnusedImpureFunctionCallResult issue for this case.

PR #8841, opened 2022-12-05
Improve parsing of list shapes

danog:
Fixes parsing of list shape phpdocs with optional keys: `list{0: int, 1?: int, 2?: string}`


danog:
This is actually a valid usecase that's [already documented](https://psalm.dev/docs/annotating_code/type_syntax/array_types/#list-shapes) and works properly in assertions, but not in docblocks.

weirdan:
Would you mind adding a test for `list{2: string, 1: string, 0: string}`? I'm not even sure whether that's valid (if order is not preserved) or invalid (if it is) list. That's the case that would have been prevented if we did not support keys in `list{}`.

danog:
Done!

tests/TypeParseTest.php
```diff
@@ -427,6 +427,100 @@ public function testTKeyedArrayNotSealed(): void
...
+            'list{int, int, string}',
+            (string)Type::parseString('list{int, int, string}')
+        );
+    }
+
+    public function testTKeyedListOptional(): void
+    {
+        $this->assertSame(
+            'list{0: int, 1?: int, 2?: string}',
+            (string)Type::parseString('list{0: int, 1?: int, 2?: string}')
```
orklah, line 442: Just wondering here, do we have sufficient checks to allow this kind of list? Is it even possible to enforce? If 1 can be undefined, then there is no list anymore

danog: As I understand it, something like `list{0: 0, 1?: 1, 2?: 2}` only makes sense for arrays whose real runtime values are always either `[0]`, `[0, 1]`, `[0, 1, 2]`.

Now this was in my mind, and I assumed Psalm already enforced this when creating an alias for the old list-like `array{}` syntax.

However it seems like the reality is a bit different, this isn't implemented properly in Psalm: https://psalm.dev/r/4eefeac2a2

Note that I didn't actually change if reconciliation logic for list-like arrays when introducing the list{} alias, and the same broken behavior is also present in v4, though it is hidden away by the `array{}` syntax:

```
[daniil@daniil-arch a]$ composer require psalm/phar:^4
[daniil@daniil-arch a]$ vendor/bin/psalm.phar --no-cache --show-info=true a.php
Target PHP version: 8.1 (inferred from current PHP version)
Scanning files...
Analyzing files...

I

INFO: Trace - a.php:9:23 - $a: array{0, 1, 2} (see https://psalm.dev/224)
```

danog: I'll submit a separate PR to fix this, after the list refactoring PR is merged :)

orklah:
Thanks!

PR #8848, opened 2022-12-05
Allowed taints to pass through urlencode()

mmcev106:
The `urlencode()` function currently escapes all taints.  I suspect it should only escape html & has_quotes taints.  Here's an example where a taint should likely be reported, but is not:

https://psalm.dev/r/4c72d132a4

This PR attempts to fix this.  Don't hesitate to let me know if I'm missing anything, or if you have any concerns.


weirdan (approved):

weirdan:
Thanks!

PR #8861, opened 2022-12-08
Precise count range

VincentLanglet:
According to https://github.com/JetBrains/phpstorm-stubs/blob/ccff41f1a6b40ff4ecfaa4964fa5007fa78f6225/Core/Core_c.php#L701

Countable should return int<0, max>. cc @orklah

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -1543,7 +1543,7 @@ return [
 'Couchbase\WildcardSearchQuery::jsonSerialize' => ['array'],
 'Couchbase\zlibCompress' => ['string', 'data'=>'string'],
 'Couchbase\zlibDecompress' => ['string', 'data'=>'string'],
-'count' => ['int', 'value'=>'Countable|array', 'mode='=>'int'],
+'count' => ['int<0, max>', 'value'=>'Countable|array', 'mode='=>'int'],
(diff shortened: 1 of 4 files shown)


orklah:
Well, technically, they can return negative, even if it doesn't makes much sense.

What's PHPStan's stance on this?

VincentLanglet:
> What's PHPStan's stance on this?

PHPStan rely on PHPStorm stubs.
So it returns errors like
```
Return type (int) of method Foo::count() should be covariant with return type (int<0, max>) of method  
         Countable::count()
```
But might be worth asking to @ondrejmirtes if he's ok with `int<0, max>` from PHPStorms stubs or if he want to override it.

Currently if I change to `int<0, max>`, I get an error with psalm, that's why I opened this PR.
(And I will still get the error since I currently use Psalm 4 since all plugins are not compatible with v5 ^^')

ondrejmirtes:
I tried it but it broke too much stuff, so currently it's only in bleedingEdge and will be enabled for everyone in PHPStan 2.0: https://github.com/phpstan/phpstan-src/blob/1.9.x/stubs/bleedingEdge/Countable.stub

Current non-bleedingEdge users get `int` only.

orklah:
I'm fine with that then :) Thanks!

PR #8870, opened 2022-12-09
Add some missing sodium functions to CallMap

lptn:
1. Add some new sodium functions
2. Use more specific return types for sodium keygen functions

## PHP extensions
I have a question: What is the policy for PHP extensions? There are [few types of ext](https://www.php.net/manual/en/extensions.membership.php):
1. **core** (These are not actual extensions. They are part of the PHP core and cannot be left out of a PHP binary with compilation options.)
2. **bundled** (These extensions are bundled with PHP.)
3. **external** (These extensions are bundled with PHP but in order to compile them, external libraries will be needed.)
4. **PECL** (These extensions are available from [PECL](https://pecl.php.net/). They may require external libraries.)

There is `\Psalm\Config::$php_extensions`, but it's not complete. Also, it doesn't take into account PHP version (different versions have different bundled extensions, e.g. [sodium](https://www.php.net/manual/en/sodium.installation.php) were PECL and became bundled as of PHP 7.2.0). So, Psalm's message like:
```
Target PHP version: 8.1 (inferred from composer.json) Extensions enabled: dom, pdo (unsupported extensions: bcmath, curl, filter, gd, hash, iconv, intl, json, mbstring, openssl, pcntl, pdo_mysql, redis, tokenizer, zip)
```

looks confusing (Psalm actually supports at lest some of them). Do you have any ideas, plans to improve it? I can try to improve it if it doesn't require deep understanding how Psalm works, but I will appreciate any input before I'll start to work on it. Thanks!


weirdan:
@lptn please open a new discussion (https://github.com/vimeo/psalm/discussions) for better visibility.

weirdan:
Thanks!

PR #8871, opened 2022-12-09
Add libxml_get_external_entity_loader to CallMap (PHP 8.2)

lptn:
Docs: https://www.php.net/manual/en/function.libxml-get-external-entity-loader.php

PS: not sure how `'(callable(string,string,array):(resource|string|null))|null'` syntax readable by psalm. I can even specify array shape here, but the risk to make it unreadable will be even higher. So I've decided to start from a simpler solution.


danog:
It's actually OK and I strongly recommend you provide an array shape for the third callable argument of libxml_set_external_entity_loader and libxml_get_external_entity_loader

lptn:
Array shape added by aced003

weirdan (approved):

weirdan:
Thanks!

PR #8881, opened 2022-12-10
Unify a way how to load stubs for extensions

lptn:
Idea: do not use `extension_loaded()` function to determinate loaded extensions. Instead, Psalm already has 2 ways of doing this:
 1. composer.json (declare ext in `require` block)
 1. using config: enableExtentions: https://psalm.dev/docs/running_psalm/configuration/#enableextensions

Note, it may potentially change live output from code snippets on psalm.dev. To fix it, please use one of 2 methods from a list above

## Why?
1. Reduce entopy
2. Provide potential contributors a clear way how to support extensions vis stubs

## Linked discussion

https://github.com/vimeo/psalm/discussions/8879


lptn:
replaced by https://github.com/vimeo/psalm/pull/8885

Issue #8874, opened 2022-12-09
Psalm 5 cache is unstable


jack-worman:
The follow has been frustrating:
1. Clear psalm cache
2. Run psalm and get no errors
3. Do any change (even adding a blank line)
4. Run psalm and get errors in totally unrelated files. Seems like mostly unused errors: UnusedProperty, PossiblyUnusedMethod, PossiblyUnusedProperty

weirdan:
@kkmuffme you've done a lot of work on caching lately, perhaps this may interest you.

jack-worman:
Trying to work on a reproduction repo, but seems hard to reproduce.

But something I just noticed, all the issues it is reporting are in the baseline. So something about using the cache causes the baselined issues to not be ignored.

weirdan:
I did see a similar issue in Psalm's own codebase.

weirdan:
```console
> ./psalm  # run with a cache, no files changed
Target PHP version: 7.4 (inferred from composer.json) Extensions enabled: dom, simplexml (unsupported extensions: ctype, json, libxml, mbstring, tokenizer)
Scanning files...
Analyzing files...



------------------------------

       No errors found!

------------------------------
217 other issues found.
You can display them with --show-info=true
------------------------------

Checks took 0.13 seconds and used 68.385MB of memory
No files analyzed
Psalm was able to infer types for 99.8722% of the codebase

> vim src/Psalm/Config.php # edit the file - I just added a bunch of empty lines in `use` block

> ./psalm  # run Psalm again
Target PHP version: 7.4 (inferred from composer.json) Extensions enabled: dom, simplexml (unsupported extensions: ctype, json, libxml, mbstring, tokenizer)
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  60 / 263 (22%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 120 / 263 (45%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 180 / 263 (68%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 240 / 263 (91%)
[...]

PR #8883, opened 2022-12-11
Prevent duplicate (Possibly)UnusedMethod/(Possibly)UnusedProperty

weirdan:
This also allows (Possibly)UnusedProperty to be suppressed with
`referencedProperty` attribute in psalm.xml

Fixes vimeo/psalm#8874


weirdan:
None of the 75 plugins I found on Packagist are instantiating (Possibly)UnusedProperty, so I think the BC break is acceptable.

PR #8899, opened 2022-12-14
Fix DOM and Reflection unreflectable methods

othercorey:
Dropped the methods that were actually properties  or misspelled duplicates.


orklah:
Thanks!

PR #8902, opened 2022-12-16
fix return type never for static function calls

kkmuffme:


orklah:
Please add a test so we don't break this in the future

orklah:
Thanks, but please still add a test next time. It really helps us with the context of a given change. You know perfectly what you changed and you understand why. Looking at your PR, I don't even have an example of what kind of code you're trying to fix and you expect me to review how you fixed it, that's really hard.

PR #8913, opened 2022-12-17
Recognize casts from object-with-properties to array

weirdan:
Previously `(array)(object)['a' => 1, 'b' => 'foo']` would result in
`array<array-key, mixed>`. Now it would be inferred as
`array{a:1,b:foo,...<array-key,mixed>}`


orklah:
Thanks!

PR #8918, opened 2022-12-17
Fix DOMDocument's propertery name preserveWhiteSpace

### LINKED ISSUE #8915: `DOMDocument::$preserveWhitespace` should be reported as dynamic property. Not `$preserveWhiteSpace`
Hi. I noticed that psalm would not report `DOMDocument::$preserveWhitespace` is dynamic property.
( When after I saw this commit  https://github.com/php/doc-base/commit/9965c2b16fbeba6d2451bd300d6da85260856de9 )

php-src declares as `preserveWhiteSpace` .
https://github.com/php/php-src/blob/027add9e1bf935c2d874f736105cced87fdeb226/ext/dom/php_dom.stub.php#L679

```php
<?php

$dom = new DOMDocument;
$dom->preserveWhiteSpace = false;
$dom->preserveWhitespace = false; // should be reported as dynamic property

$dom->PreserveWhitespace = false; // ERROR: UndefinedPropertyAssignment
$dom->preservewhitespace = false; // ERROR: UndefinedPropertyAssignment

// PropertyMap  `'strictErrorChecking' => 'bool',` 
$dom->strictErrorChecking = false;
$dom->strictErrorchecking = false;
```
https://psalm.dev/r/a2760ed2e5

It seems current PropertyMap is mistaken.
> 'preserveWhitespace' => 'bool',
https://github.com/vimeo/psalm/blob/5afc3282d41c382e3803a971aaff2800aa7adc3e/dictionaries/PropertyMap.php#L116

But I don't know why currently `$dom->preserveWhiteSpace` is not reported as ERROR against case sensitive...

sasezaki:
Fixes #8915

### DIFF
--- a/dictionaries/PropertyMap.php
+++ b/dictionaries/PropertyMap.php
@@ -113,7 +113,7 @@ return [
         'formatOutput' => 'bool',
         'implementation' => 'DOMImplementation',
         'lastElementChild' => 'DOMElement|null',
-        'preserveWhitespace' => 'bool',
+        'preserveWhiteSpace' => 'bool',
         'recover' => 'bool',
         'resolveExternals' => 'bool',
         'standalone' => 'bool',


orklah:
Thanks!

PR #8921, opened 2022-12-17
Json baseline formatter

jack-worman:
I just think JSON is much easier to read and work with.
And not including `occurrences` will lead to smaller diffs, which is important for my team, as our baseline is constantly being updated.

TODO: document new cli option

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -107,12 +107,13 @@
     ],
     "scripts": {
         "cs": "phpcs -ps",
-        "cs-fix": "phpcbf -ps",
+        "cs-fix": "phpcbf -p",
         "lint": "parallel-lint ./src ./tests",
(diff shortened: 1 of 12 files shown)


orklah:
> And not including occurrences will lead to smaller diffs, which is important for my team, as our baseline is constantly being updated.

What do you mean by that? Number of occurences is an important part of the baseline when there is no snippet to register, you can't just drop those?

jack-worman:
> What do you mean by that? Number of occurences is an important part of the baseline when there is no snippet to register, you can't just drop those?

I removed the part of the code that didn't include the selection if it had a line break in it. Are there other parts reasons for a selection to not be included?
`ErrorBaseline::countIssueTypesByFile()` is the relevant method.

jack-worman:
Closing in favor of #8967

PR #8922, opened 2022-12-18
Make `sprintf` return `non-empty-string` when possible

fluffycondor:


stubs/CoreGenericFunctions.phpstub
```diff
@@ -1116,7 +1116,13 @@ function preg_quote(string $str, ?string $delimiter = null) : string {}
 /**
  * @psalm-pure
  *
- * @param string|int|float $values
+ * @template TFormat of string
```
orklah, line 1119: FYI, You're not forced to use templates for a simple case like this.

Psalm would have understood 
```
 * @return ($format is non-empty-string
 *      ? ($values is non-empty-string|int|float ? non-empty-string : string)
 *      : string)
```
as well

fluffycondor: Thanks!

orklah:
Thanks!

PR #8927, opened 2022-12-18
Trailing commas

jack-worman:
We can now use trailing commas in many more places.
This will marginally decrease noise in diffs.


weirdan (approved):

orklah:
Thanks!

PR #8928, opened 2022-12-18
Make md5 return type more precise

fluffycondor:


jack-worman:
Is it impossible for md5() to return all zeros?

`php -r "var_export((bool) "00000000000000000000000000000000");"`
false

orklah:
Are you sure?
https://3v4l.org/6cLB2

jack-worman:
That is very strange. Why is the inline cli version returning the wrong answer...?
Screenshot 2022-12-18 at 11 22 58 AM

orklah:
I was starting to get really confused at that too :p

orklah:
Thanks!

PR #8931, opened 2022-12-18
Fix getParentClass stub

VincentLanglet:
I found another error in the stub @orklah, cf 
https://www.php.net/manual/en/reflectionclass.getparentclass.php


orklah:
Thanks!

PR #8934, opened 2022-12-18
Allow conditions inside loops to preserve or narrow int range

### LINKED ISSUE #7555: Incorrect type checking (0|positive-int)
https://psalm.dev/r/0b7a15278a

### LINKED ISSUE #8020: Int range not updated when in loop
https://psalm.dev/r/65c57ad265

I figured this would work since the variable is redefined on each loop iteration.

### LINKED ISSUE #8865: Psalm fails to narrow type to positive int inside loop
Psalm fails to infer that a variable is a positive int when the condition occurs inside a loop.

https://psalm.dev/r/155977bb20

theodorejb:
Fixes #7555, fixes #8020, fixes #8865

I lack familiarity with the assertion reconciler code, so I'm not sure if there is something the continue if `$inside_loop` checks were necessary for. But at least the current tests still seem to pass without them, and this fixes issues where conditions inside a loop failed to preserve or narrow the range of an int as expected.

### DIFF
--- a/src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php
@@ -123,10 +123,6 @@ class ArrayTypeComparator
         }
 
         foreach ($input_type_part->type_params as $i => $input_param) {
-            if ($i > 1) {
-                break;
-            }
-
             $container_param = $container_type_part->type_params[$i];
 
             if ($i === 0
--- a/tests/IntRangeTest.php
+++ b/tests/IntRangeTest.php
@@ -443,9 +443,8 @@ class IntRangeTest extends TestCase
                     '$remainder===' => 'int<min, 1>',
                 ],
             ],
-            'SKIPPED-IntRangeRestrictWhenUntouched' => [
+            'IntRangeRestrictWhenUntouched' => [
                 'code' => '<?php
-                    //skipped, int range in loops not supported yet
                     foreach ([1, 2, 3] as $i) {
                         if ($i > 1) {
                             takesInt($i);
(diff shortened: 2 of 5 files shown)


tests/IntRangeTest.php
```diff
@@ -457,6 +456,35 @@ function takesInt(int $i): void{
...
+                        }
+                    }
+                    takesNonNegativeInt($sum);
+
+                    /** @psalm-param int<0, max> $i */
+                    function takesNonNegativeInt(int $i): void{
+                        return;
+                    }',
+            ],
             'SKIPPED-wrongLoopAssertion' => [
```
orklah, line 488: Can you try to unskip that one too?

src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php
```diff
@@ -123,10 +123,6 @@ public static function isContainedBy(
         }
 
         foreach ($input_type_part->type_params as $i => $input_param) {
-            if ($i > 1) {
```
orklah, line 126: why did you remove that one?

theodorejb: It causes the following error:
> TypeDoesNotContainType - src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php:126:17 - Type int<0, 1> for $i is never >1

tests/IntRangeTest.php
theodorejb: I'm not sure what that one is trying to test, but it causes the following error:
> Psalm\Exception\CodeException: Trace - src\somefile.php:8:51 - $i: int<0, max>

src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php
orklah: yep but that's wrong. This loop can go past one

tests/IntRangeTest.php
orklah: Strange that traces were left in there. Can you get rid of the trace comments to see if there is other errors?

src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php
orklah: That must be one of the reasons I skipped tests and did not try analyzing assertions inside loops. Unfortunately, this is not ready

tests/IntRangeTest.php
theodorejb: There aren't any errors after removing the trace comments.

src/Psalm/Internal/Type/Comparator/ArrayTypeComparator.php
theodorejb: What would cause the loop to go past one since the `type_params` property has a type of `array{Union, Union}`? And what is the `break` needed for that isn't covered by a test?

orklah: Oh yeah, I guess you're right. I thought it was just a performance thing but the loop can't go over 1 indeed.

Array count has improved a lot recently, maybe it has fixed things that were broken when I decided to skip processing assertions in loop

orklah:
Thanks!

PR #8943, opened 2022-12-19
More array fixes

### LINKED ISSUE #8940: Multiple array/list are incorrectly merged as return type
Symfony has a fairly complex return type specified on their [EventSubscriberInterface](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/EventDispatcher/EventSubscriberInterface.php).

`@return array<string, string|array{0: string, 1: int}|list<array{0: string, 1?: int}>>`

I copied both their `@return` and the phpdoc (which includes examples), so you can just read the psalm.dev-snippet:

https://psalm.dev/r/49064bc34c

When using the most complex variant (a string-key => list of array<string, ?int>), Psalm fails to recognize that.

```
does not match the declared return type array<string, array{0?: array{0: string, 1?: int}|string, 1?: int, ...<int<0, max>, array{0: string, 1?: int}>}|string>
```

So according to Psalm the array-value should be either of these:
- `string`
- `array{0?: array{0: string, 1?: int}|string, 1?: int, ...<int<0, max>, array{0: string, 1?: int}>}`

That second option seems to be a mix of the array and list options from the phpdoc, but merged such that this bit doesn't match anymore, while it should match the list-option:
```
'b' => [
                ['methodB1', 0],
                ['methodB2', 10],
            ],
```

The snippet also fails with only the above key 'b'.

danog:
We really should add support for disjoint arrays...

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -7,6 +7,8 @@
 
 - [BC] The only optional boolean parameter of `TKeyedArray::getGenericArrayType` was removed, and was replaced with a string parameter with a different meaning.
 
+- [BC] The `TDependentListKey` type was removed, replacing it with an optional property of the `TIntRange` type.
+
 # Upgrading from Psalm 4 to Psalm 5
 ## Changed
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -7,7 +7,7 @@
 
 - [BC] The only optional boolean parameter of `TKeyedArray::getGenericArrayType` was removed, and was replaced with a string parameter with a different meaning.
 
-- [BC] The `TDependentListKey` type was removed, replacing it with an optional property of the `TIntRange` type.
+- [BC] The `TDependentListKey` type was removed and replaced with an optional property of the `TIntRange` type.
 
 # Upgrading from Psalm 4 to Psalm 5
 ## Changed
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="dev-master@6eb37b9dc2321e4eaade9d3d2dca1aff6f2c0a8f">
+<files psalm-version="dev-master@d90a9a28a53176b4eb329d4c062d37516d3227f3">
   <file src="examples/TemplateChecker.php">
     <PossiblyUndefinedIntArrayOffset occurrences="2">
       <code>$comment_block-&gt;tags['variablesfrom'][0]</code>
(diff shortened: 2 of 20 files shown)


danog:
Whoops, the logic is actually incorrect, hang on...

danog:
One very nasty remaining baseline issue is caused by https://github.com/vimeo/psalm/issues/8949, the rest is caused by false positives

orklah:
Thanks!

Ocramius:
Not sure if worth investigating, but consider checking the baseline additions @ https://github.com/laminas/laminas-validator/pull/166, @danog

I think what's happening here is that Psalm cannot determine the type of `$decoded` after this code block:

```php

        $decoded   = [];
        $separator = strrpos($encoded, '-');
        if ($separator > 0) {
            for ($x = 0; $x < $separator; ++$x) {
                // prepare decoding matrix
                $decoded[] = ord($encoded[$x]);
            }
        }

        $lengthd = count($decoded);
        $lengthe = strlen($encoded);

        // decoding
        $init  = true;
        $base  = 72;
        $index = 0;
        $char  = 0x80;

        for ($indexe = $separator ? $separator + 1 : 0; $indexe < $lengthe; ++$lengthd) {
            for ($oldIndex = $index, $pos = 1, $key = 36; 1; $key += 36) {
                $hex   = ord($encoded[$indexe++]);
                $digit = $hex - 48 < 10 ? $hex - 22
                       : ($hex - 65 < 26 ? $hex - 65
                       : ($hex - 97 < 26 ? $hex - 97
                       : 36));

                $index += $digit * $pos;
                $tag    = $key <= $base ? 1 : ($key >= $base + 26 ? 26 : $key - $base);
                if ($digit < $tag) {
                    break;
                }

                $pos = (int) ($pos * (36 - $tag));
            }

            $delta  = intval($init ? ($index - $oldIndex) / 700 : ($index - $oldIndex) / 2);
            $delta += intval($delta / ($lengthd + 1));
            for ($key = 0; $delta > 910 / 2; $key += 36) {
                $delta = intval($delta / 35);
            }

            $base   = intval($key + 36 * $delta / ($delta + 38));
            $init   = false;
            $char  += (int) ($index / ($lengthd + 1));
            $index %= $lengthd + 1;
            if ($lengthd > 0) {
                for ($i = $lengthd; $i > $index; $i--) {
                    $decoded[$i] = $decoded[$i - 1];
                }
            }

            $decoded[$index++] = $char;
        }
```

For now, I'm OK with adding it to the baseline, but I think this patch may have led to this tiny regression :-)

danog:
@Ocramius Hmm, mind opening a new issue for this one?
At a quick glance, the inferred type seems correct: https://psalm.dev/r/8adf377b9f

PR #8945, opened 2022-12-19
getopt() return type fix

jack-worman:
You can have false, string, and array values (with false and string values) at the same time:

`% php tester.php -a -b=value -c -c -c -c -c=value`
```php
<?php
var_export(getopt('ab:c::'));
```
```txt
array (
  'a' => false,
  'b' => 'value',
  'c' => 
  array (
    0 => false,
    1 => false,
    2 => false,
    3 => false,
    4 => 'value',
  ),
)
```


orklah:
Thanks!

This function would probably benefit from having a returnTypeProvider that would return an array shape according to the parameter

PR #8956, opened 2022-12-20
Preserve case-sensitive class name in method issues

jack-worman:
Currently, a plugin like this is impossible in a PSR-4 environment because autoloading is case-sensitive:
```php
<?php

declare(strict_types=1);

namespace App\Psalm\Hook;

use App\Controller\AbstractLegacyController;
use Psalm\Issue\PossiblyUnusedMethod;
use Psalm\Plugin\EventHandler\BeforeAddIssueInterface;
use Psalm\Plugin\EventHandler\Event\BeforeAddIssueEvent;

class ZendControllerActionSuppressUnusedHook implements BeforeAddIssueInterface
{
    public static function beforeAddIssue(BeforeAddIssueEvent $event): ?bool
    {
        $codeIssue = $event->getIssue();
        if ($codeIssue instanceof PossiblyUnusedMethod) {
            [$class, $method] = \explode('::', $codeIssue->method_id);
            if (
                \is_subclass_of($class, AbstractLegacyController::class)
                && \str_ends_with($method, 'action')
            ) {
                return false;
            }
        }
        return null;
    }
}
```
`\is_subclass_of($class, AbstractLegacyController::class)` is returning false.

### DIFF
--- a/src/Psalm/Issue/MethodIssue.php
+++ b/src/Psalm/Issue/MethodIssue.php
@@ -10,7 +10,10 @@ abstract class MethodIssue extends CodeIssue
 {
     /**
      * @var string
+     * @psalm-suppress PossiblyUnusedProperty Allows plugins to autoload the class.
      */
+    public string $fq_class_name;
+    /** @var lowercase-string */
     public $method_id;
 
     public function __construct(
@@ -19,6 +22,7 @@ abstract class MethodIssue extends CodeIssue
         string $method_id
     ) {
         parent::__construct($message, $code_location);
+        [$this->fq_class_name] = \explode('::', $method_id);
         $this->method_id = strtolower($method_id);
     }
 }


jack-worman:
Another solution is just give plugins access to Codebase and I don't have to use the autoloader at all.

orklah:
> give plugins access to Codebase

This is the way. In normal conditions, Psalm will already have mapped the entire codebase and you should not need autoloading from your plugin.

I think we were too restrictive in what we marked as internal given we have no proper plugin API so feel free to use internal symbol in Psalm, just know that we don't offer guarantees in terms of BC breaks here, so we may need to break those in the future (without warning...)

jack-worman:
@orklah How do I get access to the `Codebase` instance that Psalm uses?

jack-worman:
> feel free to use internal symbol in Psalm

I am going to retrieve the instance via `ProjectAnalyzer::getInstance()->getCodebase()`.

weirdan:
`Codebase` is usually supplied in the event itself, and you can retrieve it with `$event->getCodebase()`. It doesn't seem to be the case for `BeforeAddIssueEvent`, though. Feel free to propose a PR adding it.

weirdan:
There's only two events that don't provide access to the `Codebase`:

```console
> rg -l --files-without-match 'function getCodebase\(' | rg -v Provider | lines
╭───┬────────────────────────────╮
│ 0 │ StringInterpreterEvent.php │
│ 1 │ BeforeAddIssueEvent.php    │
╰───┴────────────────────────────╯
```

orklah:
Note that in BeforeAddIssue, we're in IssueBuffer::add() which does not have Codebase at hand.

Codebase being an almost Singleton, not sure if it's worth passing it to every issue creation just for that plugin when it can be retrieved otherwise

weirdan:
> we're in `IssueBuffer::add()` which does not have Codebase at hand.

It's a call away, and is used in that very method: https://github.com/vimeo/psalm/blob/c52970958dea2292d12ae4a4b5fc9bf72434d908/src/Psalm/IssueBuffer.php#L274

so there shouldn't be any problems passing it to the event constructor.

orklah:
Right, I didn't look after the call to the plugin.

Issue #4139, opened 2020-09-06
LSP completion returns snippets when client does not support them


matthijskooijman:
I'm using [YouCompleteMe](https://github.com/ycm-core/YouCompleteMe) as an LSP client, and noticed an assertion error in that client. I asked they YCM developers, and they said:

> That assertion means that the server is sending snippets instead of plaintext completion, despite what YCM declares as its capabilities. In other words the server violates the protocol.

I'm not familiar with the protocol, so I'm not sure what this means exactly, but I hope it is clear to you and helpful.

To reproduce it, I've used this code snippet:

```
<?php

class bar {
        /** @var int */
        var $fooz = 1;

        function foo() {
                $var = new bar();
                $var->;
        }
}
```

Triggering a completion after the unfinished `$var->`.

The YCM log, which includes all JSONRPC messages sent, is shown below. Psalm itself does not produce any output.

```
2020-09-06 12:17:58,990 - DEBUG - No global extra conf, not calling method YcmCorePreload
2020-09-06 12:17:59,059 - INFO - Received ready request
2020-09-06 12:17:59,067 - INFO - Received signature help available request
2020-09-06 12:17:59,149 - INFO - Received event notification
2020-09-06 12:17:59,149 - DEBUG - Event name: BufferVisit
2020-09-06 12:17:59,154 - INFO - Received event notification
2020-09-06 12:17:59,154 - DEBUG - Event name: FileReadyToParse
2020-09-06 12:17:59,154 - INFO - Adding buffer identifiers for file: /var/www/html/hypha/system/datatypes/festival.php
2020-09-06 12:17:59,155 - INFO - Starting phpCompleter: ['/home/matthijs/bin/psalm', '--language-server']
2020-09-06 12:17:59,163 - INFO - phpCompleter started with PID 113403
[...]

PR #8960, opened 2022-12-20
WIP: Performance/Feature Improvements to the Language Server

tm1000:
This is something I've been working on for the past year with my fellow coworkers @bsandstrom @berrugo @GameGamer43 and @xrobau in v4 but just brought it to v5. Keeping it as a draft here against v5 while I work through tests and other issues though it currently works as is

*Status*: Updated to https://github.com/vimeo/psalm/releases/tag/5.6.0

### TODO:
- [ ] Add more test cases for the LSP
- [ ] Fixed errored test case
- [ ] Add back internal functions that were removed 
- [X] Fix dynamic configuration changes

### Fixed:
* #4139 
* #9180

### Added:
* `ClientConfiguration` class to store client configuration for the LSP
* Dynamically request configuration from client. 
  * This means that clients can now set updates and not have to restart Psalm
* `PHPMarkdownContent` class that extends `LanguageServerProtocol\MarkupContent`.
  * This helps to standardize how hovers look in clients across the usage in Psalm
* Provide hover analysis results for any context in php
  * This actually ends up providing decent generic support (At least in my usage)
* `--clear-cache-on-boot` option to clear the cache folder for the LSP on start
  * This is mainly a development setting as I needed to delete the cache folder while working on the LSP itself (Default is false)
  * This does not exit on the cache being cleared (which `--clear-cache` does)
* Provide more granular control over what the LSP does:
  * `--enable-code-actions` Enables or disables code actions. Default is true.
  * `--enable-provide-diagnostics` Enables or disables providing diagnostics. Default is true.
  * `--enable-provide-hover` Enables or disables providing hover. Default is true.
  * `--enable-provide-signature-help` Enables or disables providing signature help. Default is true.
  * `--enable-provide-definition` Enables or disables providing definition. Default is true.
  * `--show-diagnostic-warnings` Enables or disables showing diagnostic warnings. Default is true. (continuation of #7379)
* Added `--on-change-debounce-ms` which helps to control CPU when working with large codebases . Default is 500ms
* Created `Reference` class to pass file reference information around instead of passing arrays (which are hard to type)
* `AsyncTestCase` base Test for testing anything in the LSP, this extends `Amp\PHPUnit\AsyncTestCase` and is now being used in a few of the test cases
* Added `--use-baseline=` to use a baseline file that was generated from Psalm CLI. This will turn any errors into warnings. The warnings can then be hidden with `--show-diagnostic-warnings=false`

### Changed:
* Created a `Progress` helper class that extends `Psalm\Progress\Progress` to help forward Psalm messages back to the LSP. This means Psalm's LSP is much more verbose than before
* Temp file array now contains the version of the file that was changed/opened as per the LSP spec, this value is then sent back to the client so that it knows when to ignore old updates.
* Because of the addition of `--on-change-debounce-ms` updates to the client are not always instant (comparatively to v4 which was 0ms) (default value for this is 500ms which is 1/2 second). This can be changed but it should be noted (I do not see a big difference in large code bases myself but it helps to prevent Psalm from backing up with change events and has actually helped the overall dev experience in my IDE)
* No longer run `doAnalysis` when event `readMessageGroup` is received. The reasoning for this is because when #7255 was done it caused a huge influx of `readMessageGroup` events

### Testing:
composer.json:
```
    "require-dev": {
        "vimeo/psalm": "dev-feature/upgrade-lsp-v5 as 5.6.0"
    },
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/tm1000/psalm"
        }
    ]
```

vscode settings:
```
    "psalm.psalmVersion": "5.6.0",
    "psalm.psalmScriptArgs": [
        "--clear-cache-on-boot",
        "--use-baseline=your-baseline.xml",
        "--on-change-debounce-ms=500",
        "--show-diagnostic-warnings=false"
    ],
    "psalm.psalmScriptPath": "app/vendor/bin/psalm-language-server",
```

Note that `psalmScriptPath` is specific to my setup. The point is you want to use `psalm-language-server` not `psalm` (even though psalm does a pass-thru to `psalm-language-server`)


tm1000:
@weirdan @orklah 

Is it important to fix: https://github.com/vimeo/psalm/actions/runs/3744291311/jobs/6357489048

These seem like they are internal files and the mentioned methods are only used in the LSP part of Psalm

weirdan:
> These seem like they are internal files and the mentioned methods are only used in the LSP part of Psalm

All of those can also be used by plugins. Isn't there any way to keep them as they were until the next major version?

tm1000:
@weirdan thanks for the reply. Yes it's easy to add them back

ging-dev:
the language server uses `pcntl_fork()` to multiprocess, but the processes use the same cache. Not sure if it works or not but has the potential to cause race condition. Should consider deleting it and using In-memory cache?
In-memory cache in tests:
https://github.com/vimeo/psalm/tree/master/tests/Internal/Provider?

tm1000:
@ging-dev It actually shouldn't use any cache but when I did that it totally broke it. Caching is useless in an LSP.

ging-dev:
> @ging-dev It actually shouldn't use any cache but when I did that it totally broke it. Caching is useless in an LSP.

Yes i know psalm needs to be cached to check the difference of files. You can use Providers like in tests https://github.com/vimeo/psalm/blob/c69ebc6e0047988574e55d5839c5b21409e33baf/tests/Internal/Provider/ParserInstanceCacheProvider.php#L15 and and dont use `pcntl_fork`. It can work without reading or writing anything to memory.

tm1000:
Updated to 5.6.0

tm1000:
The CI is strangely bugged on this PR. It refuses to run any tests except for two

tm1000:
@ging-dev in regards to the usage of `pcntl_fork` thats used when its initiated as a TCP server (which I don't use).

In vscode I can see how only one server services one connection but I dont know how this applies to people not using vscode, in those cases they technically could service n+1 connections even though I think thats wrong (one server should be used for one IDE and so forth)

weirdan:
@tm1000 I added the `Meta` section to the PR description so that issues this pull request fixes are properly linked.

tm1000:
Technically this is still WIP and draft but I undrafted it to fix the non running checks.

tm1000:
> the language server uses `pcntl_fork()` to multiprocess, but the processes use the same cache. Not sure if it works or not but has the potential to cause race condition. Should consider deleting it and using In-memory cache?
> In-memory cache in tests:
> [`master`/tests/Internal/Provider](https://github.com/vimeo/psalm/tree/master/tests/Internal/Provider?rgh-link-date=2022-12-21T09%3A16%3A46Z)?

This was the magic to get the LSP to work without a cache directory.

Additionally I needed to then set `$config->cache_directory = null;` so that it wouldnt create any directories.

After the above changes everything is now in memory and as a result Psalm's LSP doesnt have to wait on disk I/O so it's faster as a result

tm1000:
It seems the in memory cache still has issues vs the file base cache. I'll have to investigate further. Maybe add a switch as well

tm1000:
added experimental `--in-memory` option. For some reason when using this option the reflected in file changes are very delayed or not reflected at all. I suspect this is happening because of a race condition where the disk IO normally blocks but because the in memory is so fast things are getting missed

tm1000:
Technically this could be merged as-is right now and it would function without any breaking changes. There is nothing holding this up anymore. All checks now pass. 

When I get time I will add additional tests. Either on this PR or a new one.

orklah:
Thanks!

PR #8961, opened 2022-12-20
Add Codebase to remaining events

jack-worman:
As discussed in #8956.


weirdan:
BCC failures are false positives, reported upstream: https://github.com/Roave/BackwardCompatibilityCheck/issues/723

weirdan:
Thanks!

PR #8965, opened 2022-12-21
Modernize strpos() calls

jack-worman:
Ran php-cs-fixer rule [modernize_strpos](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/alias/modernize_strpos.rst)

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -82,6 +82,8 @@ use function krsort;
 use function ksort;
 use function preg_match;
 use function preg_replace;
+use function str_contains;
+use function str_starts_with;
 use function strlen;
 use function strpos;
 use function strrpos;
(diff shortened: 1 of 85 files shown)


jack-worman:
Not sure why the phar build is failing since `symfony/polyfill-php80` is a require dependency.

weirdan:
> Not sure why the phar build is failing

The code in question is being run before any autoloader is registered. 

> Also, does not happen when I build the phar locally.

You need to run `psalm.phar --version` on PHP 7.4 to reproduce it.

Overall I don't think we need this PR just yet. Possibly we can drop the polyfill too, as it was introduced in a 'revert' commit, so it's probably not really used.

jack-worman:
> Possibly we can drop the polyfill too, as it was introduced in a 'revert' commit, so it's probably not really used.

I see the `\Stringable` interface being used.

weirdan:
> I see the \Stringable interface being used.

Indeed, missed that.

orklah:
This would be easy to remove though, as Stringable is implied when __toString is there

weirdan:
@danog is there any reason you added explicit `implements Stringable`?

danog:
Naah, no specific reason, just wanted to make it explicit and at the time I was hoping to raise the minimum psalm php version to 8.0 :)

Issue #1062, opened 2018-11-05
Improve messaging when baseline for a given file is exceeded


muglug:
Causing some confusion locally.

We could store the snippet selection text for each error in the baseline - it would make the file larger, but it would make false-positives much less likely.

cc @ErikBooijCB

ErikBooijCB:
Interesting idea indeed.

The file would look something like this?

```xml
<?xml version="1.0" encoding="UTF-8"?>
<files>
  <file src="src/folder/File1.php">
    <MoreSpecificImplementedParamType>
       ... code snippet
    </MoreSpecificImplementedParamType>
  </file>
  <file src="src/folder/File1.php">
    <PossiblyInvalidMethodCall>
       ... code snippet
    </PossiblyInvalidMethodCall>
  </file>
</files
```

And when checking, compare the type *and* snippet?

I can look into that, sometime soon.

muglug:
Maybe more like 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<files>
  <file src="src/folder/File1.php">
    <MoreSpecificImplementedParamType occurrences="2">
       <code><!-- function fooBar() --></code>
    </MoreSpecificImplementedParamType>
  </file>
  <file src="src/folder/File1.php">
    <PossiblyInvalidMethodCall>
       ... code snippet
    </PossiblyInvalidMethodCall>
  </file>
</files>
```

Any code with a line break would be omitted (most don't have line breaks)

weirdan:
> Any code with a line break would be omitted (most don't have line breaks)

@muglug I wonder why would we need to omit those? I'm observing funny behaviour here, which I suppose is what this ticket was about, for multiline calls to a deprecated function like this:
```php
<?php
/** @deprecated */
function send_mail(string $_subject, string $_body): void {}

send_mail( // newly added, but not flagged
    'Another subj',
    'Another message'
);

send_mail( // grandfathered in in baseline, but is flagged 
    'Subj', 
    'Message'
);
```

PR #8967, opened 2022-12-21
Remove occurrences from error baseline

jack-worman:
Occurrences are redundant information in the baseline and are noise in diffs.

The only time (as far as I can tell) that a code snippet was not included is if it had a "\n". I just removed this restriction. Not 100% the reason for it in the first place.

### DIFF
--- a/src/Psalm/ErrorBaseline.php
+++ b/src/Psalm/ErrorBaseline.php
@@ -22,8 +22,8 @@ use function min;
 use function phpversion;
 use function preg_replace_callback;
 use function sort;
+use function sprintf;
 use function str_replace;
-use function strpos;
 use function trim;
 use function usort;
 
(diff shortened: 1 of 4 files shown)


weirdan:
> The only time (as far as I can tell) that a code snippet was not included is if it had a "\n". I just removed this restriction. Not 100% the reason for it in the first place.

I wondered about that myself: https://github.com/vimeo/psalm/issues/1062#issuecomment-861650525

src/Psalm/ErrorBaseline.php
```diff
@@ -112,15 +112,19 @@ public static function read(FileProvider $fileProvider, string $baselineFile): a
...
-                ];
+                $files[$fileName][$issueType] = ['o' => 0, 's' => []];
                 $codeSamples = $issue->getElementsByTagName('code');
 
                 foreach ($codeSamples as $codeSample) {
+                    $files[$fileName][$issueType]['o'] += 1;
                     $files[$fileName][$issueType]['s'][] = trim($codeSample->textContent);
                 }
+
+                // TODO: Remove in v6.0.0
```
weirdan, line 123:
```suggestion
                // TODO: Remove in Psalm 6
```

This is easier to grep for.

weirdan (changes requested):

weirdan (approved):

orklah:
Does this keeps the old behaviour (counting occurences) but only generates new format (without occurences) or does this risk breaking existing baselines?

jack-worman:
> Does this keeps the old behaviour (counting occurences) but only generates new format (without occurences) or does this risk breaking existing baselines?

It detects if the "occurrences" attribute exists, if it does, it uses it. So, existing baselines should be unaffected.

orklah:
Thanks!

PR #8971, opened 2022-12-21
Extend XML enum of supported PHP ext

lptn:
This is just a quick-fix. We probably should allow to use any value in this list OR add a unit test to update this list when add a new stub for PHP ext.


weirdan:
Thanks!

kenjis:
Thank you!

PR #8972, opened 2022-12-21
Revert #8955

danog:
My PR completely broke iterators in our codebase in a way that is hard to reproduce in standalone example, and I realized the approach was actually wrong, because the issue only occurs when *not* using `@return` phpdocs but only native typehints.

I'll have to dig elsewhere to fix this :)


weirdan:
Thanks!

Issue #8963, opened 2022-12-21
RedundantFunctionCall problem


matiux:
This is my  enum
https://psalm.dev/r/b147e05cf3

As you can see array_values is marked as RedundantFunctionCall but I'm using it to reset array indexes as the call to array_diff above produces an array with non continuous indexes. Is there a way to handle this case?

jack-worman:
The problem is that psalm does not like that array_column call because `value` is a property not a key: https://psalm.dev/r/ebd8c22399
You should use array_map instead: https://psalm.dev/r/cf56254b4e

jack-worman:
This works: https://psalm.dev/r/7faba56988
So Psalm just does not know about enums having properties. `\BackedEnum` should be thought of as `object{name: string, value: int|string}`

weirdan:
> Probably a bug in `ArrayColumnReturnTypeProvider`

Most likely.

matiux:
Thanks!
For now I try the solution of not using array_column, so doing:
```
$allLangValues = array_map(fn (Lang $lang): string => $lang->value, Lang::cases());
```

danog:
Actually, it's a bug in the get_object_vars return type provider, which is invoked by the array_column type provider: https://psalm.dev/r/22dca38fd8

jack-worman:
I am working on a fix for this.

orklah:
This was fixed :)

PR #8973, opened 2022-12-21
Fix get_object_vars on enums

jack-worman:
Fix for #8963

I am starting to understand this codebase :)

### DIFF
--- a/src/Psalm/Internal/Codebase/Populator.php
+++ b/src/Psalm/Internal/Codebase/Populator.php
@@ -2,6 +2,7 @@
 
 namespace Psalm\Internal\Codebase;
 
+use BackedEnum;
 use InvalidArgumentException;
 use Psalm\Internal\Analyzer\ClassLikeAnalyzer;
 use Psalm\Internal\MethodIdentifier;
@@ -14,8 +15,13 @@ use Psalm\Progress\Progress;
 use Psalm\Storage\ClassConstantStorage;
 use Psalm\Storage\ClassLikeStorage;
 use Psalm\Storage\FileStorage;
+use Psalm\Storage\PropertyStorage;
+use Psalm\Type\Atomic\TInt;
+use Psalm\Type\Atomic\TNonEmptyString;
+use Psalm\Type\Atomic\TString;
 use Psalm\Type\Atomic\TTemplateParam;
 use Psalm\Type\Union;
+use UnitEnum;
 
 use function array_filter;
 use function array_intersect_key;
@@ -688,6 +694,19 @@ class Populator
             $parent_interface_storage->parent_interfaces,
             $storage->parent_interfaces,
         );
+
+        if (isset($storage->parent_interfaces[strtolower(UnitEnum::class)])) {
+            $storage->declaring_property_ids['name'] = $storage->name;
+            $storage->appearing_property_ids['name'] = "{$storage->name}::\$name";
+            $storage->properties['name'] = new PropertyStorage();
+            $storage->properties['name']->type = new Union([new TNonEmptyString()]);
+        }
+        if (isset($storage->parent_interfaces[strtolower(BackedEnum::class)])) {
+            $storage->declaring_property_ids['value'] = $storage->name;
+            $storage->appearing_property_ids['value'] = "{$storage->name}::\$value";
+            $storage->properties['value'] = new PropertyStorage();
+            $storage->properties['value']->type = new Union([new TInt(), new TString()]);
+        }
     }
 
     private function populateDataFromImplementedInterface(
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
@@ -73,6 +73,8 @@ use function assert;
 use function count;
 use function get_class;
 use function implode;
+use function is_int;
+use function is_string;
 use function preg_match;
 use function preg_replace;
 use function preg_split;
@@ -286,6 +288,7 @@ class ClassLikeNodeScanner
             $this->codebase->classlikes->addFullyQualifiedTraitName($fq_classlike_name, $this->file_path);
         } elseif ($node instanceof PhpParser\Node\Stmt\Enum_) {
             $storage->is_enum = true;
+            $storage->final = true;
 
             if ($node->scalarType) {
                 if ($node->scalarType->name === 'string' || $node->scalarType->name === 'int') {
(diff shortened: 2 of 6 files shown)


stubs/Php81.phpstub
```diff
@@ -13,6 +13,8 @@ namespace {
 
     interface BackedEnum extends UnitEnum
     {
+        /** @var non-empty-string $name */
+        public readonly string $name;
```
jack-worman, line 17: For some reason, it was not picking up this property from `UnitEnum`

tests/ReturnTypeProvider/GetObjectVarsTest.php
```diff
@@ -125,6 +125,29 @@ public function __construct(public mixed $t) {}
...
+            'code' => '<?php
+                enum A: int { case One = 1; case Two = 2; }
+                function getBackedEnum(): BackedEnum { return A::One; }
+                $b = get_object_vars(getBackedEnum());',
+            'assertions' => [
+                '$b===' => 'array{name: non-empty-string, value: int|string}',
+            ],
+            'ignored_issues' => [],
+            'php_version' => '8.1',
+        ];
```
jack-worman, line 149: Test with `StringBackedEnum`, `IntBackedEnum`, and with literals of each type

stubs/Php81.phpstub
weirdan: I bet the reason is that interfaces in PHP cannot have properties.

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
```diff
@@ -1580,9 +1580,8 @@ private function visitPropertyDeclaration(
                 }
 
                 if ($doc_var_group_type) {
-                    $property_storage->type = count($stmt->props) === 1
-                        ? $doc_var_group_type
-                        : $doc_var_group_type;
+                    // TODO: What was up with this ternary?
```
danog, line 1583: Probably a leftover from the `clone` removal refactoring, it's OK to remove this comment :)

src/Psalm/Internal/Provider/ReturnTypeProvider/GetObjectVarsReturnTypeProvider.php
```diff
@@ -126,7 +126,9 @@ public static function getGetObjectVarsReturnType(
                 return new TKeyedArray(
                     $properties,
                     null,
-                    $class_storage->final ? null : [Type::getString(), Type::getMixed()],
+                    $class_storage->final || \is_a($class_storage->name, \UnitEnum::class, true)
```
danog, line 129: This should use Codebase::classImplements

src/Psalm/Internal/Provider/ReturnTypeProvider/GetObjectVarsReturnTypeProvider.php
```diff
@@ -126,7 +156,12 @@ public static function getGetObjectVarsReturnType(
                 return new TKeyedArray(
                     $properties,
                     null,
-                    $class_storage->final ? null : [Type::getString(), Type::getMixed()],
+                    $class_storage->final
+                        || $class_storage->name === UnitEnum::class
+                        || $class_storage->name === BackedEnum::class
```
jack-worman, line 161: Perhaps update `classImplements()` to work with interfaces?

weirdan: I'd suggest using `classImplements() || interfaceExtends()`. Extension of enum interfaces is valid in PHP: https://3v4l.org/eKJ49

This will also cover the case of `BackedEnum` (as it extends `UnitEnum`). And to save some CPU cycles you can check `$class_storage->is_enum` before checking the inheritance tree.

weirdan: The test case for that may look something like this;
```php
interface I extends BackedEnum {}

/** @return array{name: non-empty-string, value: int|string} */
function f(I $e): void {
   return get_object_vars($e);
}
```

jack-worman: Another thought, maybe we should mark all enums as final? It is practically true, since enums cannot have properties.

weirdan: It would match the way enums are represented in PHP internally:
image

so it sounds good to me. We'll still need to check `interfaceExtends()`, as we may be dealing with an interface extending one of the enum interfaces, and that interface won't be final itself.

src/Psalm/Internal/Codebase/Populator.php
```diff
@@ -688,6 +694,19 @@ private function populateInterfaceDataFromParentInterface(
...
+            $storage->appearing_property_ids['name'] = "{$storage->name}::\$name";
+            $storage->properties['name'] = new PropertyStorage();
+            $storage->properties['name']->type = new Union([new TNonEmptyString()]);
+        }
+        if (isset($storage->parent_interfaces[strtolower(BackedEnum::class)])) {
+            $storage->declaring_property_ids['value'] = $storage->name;
+            $storage->appearing_property_ids['value'] = "{$storage->name}::\$value";
+            $storage->properties['value'] = new PropertyStorage();
+            $storage->properties['value']->type = new Union([new TInt(), new TString()]);
+        }
```
jack-worman, line 709: This makes the properties in the stubs not needed? Testing that now.

src/Psalm/Internal/Provider/ReturnTypeProvider/GetObjectVarsReturnTypeProvider.php
```diff
@@ -55,6 +55,22 @@ public static function getGetObjectVarsReturnType(
...
+                if ($enum_classlike_storage->enum_type === null) {
+                    return new TKeyedArray($properties);
+                }
+                $enum_case_storage = $enum_classlike_storage->enum_cases[$object_type->case_name];
+                if (is_int($enum_case_storage->value)) {
+                    $properties['value'] = new Union([new Atomic\TLiteralInt($enum_case_storage->value)]);
+                } elseif (is_string($enum_case_storage->value)) {
+                    $properties['value'] = new Union([new Atomic\TLiteralString($enum_case_storage->value)]);
+                }
+                return new TKeyedArray($properties);
```
jack-worman, line 71: This property logic should be moved into TEnumCase.

src/Psalm/Internal/Provider/ReturnTypeProvider/GetObjectVarsReturnTypeProvider.php
```diff
@@ -126,7 +142,11 @@ public static function getGetObjectVarsReturnType(
                 return new TKeyedArray(
                     $properties,
                     null,
-                    $class_storage->final ? null : [Type::getString(), Type::getMixed()],
+                    $class_storage->final
+                        || $class_storage->name === UnitEnum::class
```
jack-worman, line 146: Annoying this is not covered by `interfaceExtends()` but I guess it makes sense.

weirdan (approved):
LGTM!

orklah:
Thanks!

Issue #8957, opened 2022-12-20
MethodSignatureMismatch - Argument 1 of DateTime::__unserialize has wrong type 'mixed', expecting 'array<array-key, mixed>' as defined by DateTimeInterface::__unserialize 


michnovka:
Running on PHP 8.2, I have this code:

I cannot reproduce it on psalm.dev, but I assume it has to do with stubs.

```php
<?php

declare(strict_types=1);

namespace App\Extension\Type;

use DateTime as PHPGlobalDateTime;

class DateTime extends PHPGlobalDateTime
{
    public function __toString(): string
    {
        return $this->format('U');
    }
}
```
Getting:
> ERROR: MethodSignatureMismatch - src/Extension/Type/DateTime.php:10:7 - Argument 1 of DateTime::__unserialize has wrong type 'mixed', expecting 'array<array-key, mixed>' as defined by DateTimeInterface::__unserialize (see https://psalm.dev/042)
> class DateTime extends PHPGlobalDateTime implements StringableDateTimeInterface

orklah:
Can you check if it reproduces if you don't name your class `DateTime` ?

Also, it seems that your snippet is not complete, it seems it should implement `StringableDateTimeInterface` too. Can we have the code for that?

michnovka:
@orklah the `StringableDateTimeInterface` is irrelevant, I didnt include it in the snippet as I was reproducing the issue without it also.

I renamed the class from `DateTime` and the issue persist. This is the minimum reproducible example for me on PHP 8.2 and psalm 5.4

```php
<?php

declare(strict_types=1);

namespace App\Extension\Type;

use DateTime;

class DateTime2 extends DateTime
{
    public function __toString(): string
    {
        return $this->format('U');
    }
}
```

orklah:
Oh, great, in 8.2, both DateTime and DateTimeInterface have a __unserialize method: https://3v4l.org/RtT39#v8.2.0 contrary to what we can find in documentation: https://www.php.net/manual/en/class.datetimeinterface.php

So we're actually missing definitions for it in https://github.com/vimeo/psalm/blob/master/dictionaries/CallMap_82_delta.php

We have an article about editing those files: https://psalm.dev/docs/contributing/editing_callmaps/ Are you up for a PR? :)

michnovka:
@orklah have a look

KevinMarques:
[...]

PR #8976, opened 2022-12-21
Add DateTimeInterface::__serialize and __unserialize

michnovka:
Fixes https://github.com/vimeo/psalm/issues/8957

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -1803,6 +1803,8 @@ return [
 'DateTimeInterface::getOffset' => ['int'],
 'DateTimeInterface::getTimestamp' => ['int'],
 'DateTimeInterface::getTimezone' => ['DateTimeZone|false'],
+'DateTimeInterface::__serialize' => ['array'],
+'DateTimeInterface::__unserialize' => ['void', 'data'=>'array'],
 'DateTimeZone::__construct' => ['void', 'timezone'=>'string'],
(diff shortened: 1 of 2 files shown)


orklah:
Seems good! Can you also add it for DateTime too?

orklah:
I'm actually not sure. It should be ok indeed.

Thanks!

michnovka:
btw the callmaps are somewhat strange.

There are call maps that repeat for `DateTime` and `DateTimeInterface` but not for `DateTimeImmutable`.

So if inheritance works as expected, why not implement properly the call maps for the `DateTimeInterface` and only add whatever (if something) is missing in implementations `DateTimeImmutable` and `DateTime`?

michnovka:
e.g.

```php
'DateTime::diff' => ['DateInterval', 'datetime2'=>'DateTimeInterface', 'absolute='=>'bool'],
'DateTimeInterface::diff' => ['DateInterval', 'datetime2'=>'DateTimeInterface', 'absolute='=>'bool'],
//but not 'DateTimeImmutable::diff'
```

orklah:
Those callmaps existed before the DateTimeInterface was even a thing, DateTimeImmutable even. They have their own lives now 😄 . So when things evolve, for example, a new interface, old signatures are not always updated (technically, we still need them for older php version, so they would be removed only from recent deltas)

Feel free to improve on what's there if that's your itch :)

PR #8985, opened 2022-12-22
callback param can be a list for preg_replace_callback

kkmuffme:


jack-worman:
Unfortunately, this is incorrect:
```php
\preg_replace_callback('~(?<foo>foo)(?<bar>bar)~', function (array $array) {
    var_export($array);
}, 'foobar');
```
```php
array (
  0 => 'foobar',
  'foo' => 'foo',
  1 => 'foo',
  'bar' => 'bar',
  2 => 'bar',
)
```

orklah:
Not sure I understood your question here

orklah:
You'll need to do a FunctionParamTypeProvider. There's a dummy example here:
https://github.com/vimeo/psalm/blob/c2ce7e63daa97a1f31a6f1baa08204cfa2ce0d89/tests/Config/Plugin/Hook/MagicFunctionProvider.php#L15

Then it should just be a matter of building a correct TCallable type

PR #8987, opened 2022-12-22
Add @psalm-api annotation

jack-worman:
- Added new `@psalm-api`
  - Added documentation
- Method param usages are checked regardless of if the class or method has references

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -117,7 +117,7 @@
             "phpunit"
         ],
         "verify-callmap": "phpunit tests/Internal/Codebase/InternalCallMapHandlerTest.php",
-        "psalm": "@php ./psalm --find-dead-code",
+        "psalm": "@php ./psalm",
         "tests": [
             "@lint",
             "@cs",
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -48,4 +48,8 @@
             <directory>tests</directory>
         </testsuite>
     </testsuites>
+
+    <php>
+        <const name="__IS_TEST_ENV__" value="1" />
+    </php>
 </phpunit>
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -60,6 +60,7 @@ use function chdir;
 use function class_exists;
 use function clearstatcache;
 use function count;
+use function defined;
 use function dirname;
 use function explode;
 use function extension_loaded;
@@ -69,6 +70,7 @@ use function file_get_contents;
 use function filetype;
 use function flock;
 use function fopen;
+use function fwrite;
 use function get_class;
 use function get_defined_constants;
 use function get_defined_functions;
@@ -122,6 +124,7 @@ use const PHP_EOL;
 use const PHP_VERSION_ID;
 use const PSALM_VERSION;
 use const SCANDIR_SORT_NONE;
+use const STDERR;
 
 /**
  * @psalm-suppress PropertyNotSetInConstructor
@@ -441,6 +444,8 @@ class Config
     public $forbidden_functions = [];
 
     /**
+     * TODO: Psalm 6: Update default to be true and remove warning.
+     *
      * @var bool
      */
     public $find_unused_code = false;
@@ -1090,6 +1095,9 @@ class Config
             $attribute_text = (string) $config_xml['findUnusedCode'];
             $config->find_unused_code = $attribute_text === 'true' || $attribute_text === '1';
             $config->find_unused_variables = $config->find_unused_code;
+        } elseif (!defined('__IS_TEST_ENV__')) {
+            fwrite(STDERR, 'Warning: "findUnusedCode" will be defaulted to "true" in Psalm 6. You should explicitly'
+                . ' enable or disable this setting.' . PHP_EOL);
         }
 
         if (isset($config_xml['findUnusedVariablesAndParams'])) {
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -630,7 +630,7 @@ class Success implements Promise {
  * @return Promise<string>
  */
 function fetch(): Promise {
-	return new Success('{"data":[]}');
+    return new Success('{"data":[]}');
 }
 
 function (): Generator {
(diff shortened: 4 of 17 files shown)


danog:
I think a better way to implement this is by adding a `@psalm-public-api` annotation that can be used on classes and methods to signal that they're part of a public API, and thus cannot be marked as unused even if no references can be found within the current codebase.

jack-worman:
> I think a better way to implement this is by adding a `@psalm-public-api` annotation

I don't think this PR and a `@psalm-public-api` annotation are mutually exclusive.
This PR still has value because Psalm is making an unsafe assumption that just because it can't find usages of a class it is not worthy of further analysis.


Wouldn't `@psalm-public-api` just be equal to `@psalm-suppress UnusedClass`?

orklah:
A test would be welcome here, I was quite confused at first.

So if I get it, tomorrow, an unused class with 50 methods will emit 51 errors because the class and each methods are unused?

I get the point for properties (that can be used inside methods that are seen as unused) but for methods, I'm not sure, seems like unnecessary noise

jack-worman:
> but for methods, I'm not sure

In my code base, I found unused method arguments being masked because because the class was unused.

> seems like unnecessary noise

So, you would be in favor of adding a mew option/config? This will allow end users to opt-in.

danog:
I don't think adding a new config is a good idea, because it's the wrong way of solving the problem.

The problem is that code that *will* get used by users of a public API is incorrectly marked as unused.  
To solve this issue, a new attribute signaling to Psalm that a class or method is public API should be created.

>Wouldn't @psalm-public-api just be equal to @psalm-suppress UnusedClass?

This is not the same thing, because if for example the `isClassReferenced` is made always to return true if a @psalm-public-api annotation will be present, `checkPropertyReferences` and `checkMethodReferences` will be invoked; `@psalm-suppress UnusedClass` does not have the same effect.

jack-worman:
> I don't think adding a new config is a good idea, because it's the wrong way of solving the problem.

I was suggesting the config as a BC layer. I would eventually want (in Psalm 6) to remove the config option.


But anyways, I'll experiment with trying to implement your `@psalm-public-api` annotation.

danog:
Thanks!
Could you also add some docs? :)

jack-worman:
> Could you also add some docs? :)

Done.

But, this solution still doesn't solve my original concern. Unused arguments in a classes marked with `@psalm-public-api` will still mask this error.

I still want a way to say, I don't care if the class is not used, still check my properties, methods, and method args.

weirdan:
>  Classes like Symfony controllers and event subscribers are not referenced directly in code and are still used

This sounds like the problem the corresponding plugin should be solving. It's not that hard either; it was one of the first things I did in the PHPUnit plugin.

`@api` tag would still be helpful for libraries, though.

> `UnusedParam` issues in classes marked with `@psalm-public-api` are still suppressed.

Then it's something we need to improve.

weirdan:
Prior art:
* https://docs.phpdoc.org/guide/references/phpdoc/tags/api.html
* https://github.com/TomasVotruba/unused-public#skip-false-positives

So we may want to use `@api` and `@psalm-api` instead.

danog:
++, and now that I think of it, it would be cool to change the default of `findUnusedCode` to `true`, now that we have this `@api` annotation that can be used to cleanup false positives (possibly mentioning the `@api` annotation in the docs for the unusedmethod/unusedclass issues).

danog:
LGTM!

src/Psalm/Internal/LanguageServer/LanguageServer.php
```diff
@@ -188,18 +188,14 @@ function (): void {
     /**
      * The initialize request is sent as the first request from the client to the server.
      *
-     * @param ClientCapabilities $capabilities The capabilities provided by the client (editor)
```
orklah, line 191: Please revert changes here and just baseline issues. https://github.com/vimeo/psalm/pull/8960 plans to use those and make changes everywhere here, it will be easier if we don't change that before

composer.json
```diff
@@ -117,7 +117,7 @@
             "phpunit"
         ],
         "verify-callmap": "phpunit tests/Internal/Codebase/InternalCallMapHandlerTest.php",
-        "psalm": "@php ./psalm --find-dead-code",
+        "psalm": "@php ./psalm --no-cache -m",
```
orklah, line 120: what was the reason for this change?

phpunit.xml.dist
```diff
@@ -48,4 +48,8 @@
             <directory>tests</directory>
         </testsuite>
     </testsuites>
+
```
orklah, line 51: I never used this config, does it makes sense to keep it?

src/Psalm/Config.php
```diff
@@ -1090,6 +1095,9 @@ private static function fromXmlAndPaths(
             $attribute_text = (string) $config_xml['findUnusedCode'];
             $config->find_unused_code = $attribute_text === 'true' || $attribute_text === '1';
             $config->find_unused_variables = $config->find_unused_code;
+        } elseif (!defined('__TESTING__')) {
```
orklah, line 1098: Seems a weird name

phpunit.xml.dist
jack-worman: This config is used by default.

src/Psalm/Config.php
jack-worman: How about `__IS_TEST_ENV__`?

composer.json
jack-worman: `--find-dead-code` is on already in the config

`--no-cache` I was getting some weird errors appears, cache probably got into some weird state

`-m` phpstorm's run composer script terminal does not support ansi :(

I will revert the two additions

src/Psalm/Config.php
orklah: Oh, I misread what it was supposed to do. I thought it displayed only in test env (but it's the opposite)

Why did you go to the trouble of excluding that for tests? Was it breaking tests?

jack-worman: Tests are set to fail if there was any data sent to output streams.

orklah:
Thanks!

PR #8988, opened 2022-12-22
Drop symfony/polyfill-php80 dependency

weirdan:


weirdan:
BCC failure is technically correct. If someone passed `Union` to a method in an argument that was typehinted as `Stringable` it would break on PHP 7.4 (as `Stringable` is not implicitly implemented in that version). However, I feel it's very unlikely.

PR #8990, opened 2022-12-22
Verify nullable callmap parameters

othercorey:
Only return type nullability was being tested before.

Fixed all previously valid callmap parameters. Updated assert message to be the same for param and return type.

Tested with PHP 8.0 and 8.1.


othercorey:
The grapheme functions are reporting `haystack` string parameter as nullable although I don't see nullable support in code or docs.

orklah:
Wow, that looks like massive improvements to callmaps! Thanks!

othercorey:
> Wow, that looks like massive improvements to callmaps! Thanks!

Did not know it was going to be so many but I finally finished it lol.

PR #8993, opened 2022-12-23
Always declare strict types

jack-worman:
First, commit is adding the phpcs rule.
Additional commits are for fixing CI.


orklah:
I added those a while ago but this was rejected by Matt: https://github.com/vimeo/psalm/pull/4229

I see both pros and cons to not have them in the codebase

Pros of not adding them:
- The codebase allows more weird things that should be caught by Psalm. Adding strict_types means we will use a limited subset of PHP in which some errors we want to catch can't exists. As we know, Psalm's codebase is often the first thing in the CI that breaks (and sometimes the only thing that breaks) and tell us the change we made was wrong. We could loose some of that

Cons of not adding them:
- When I tried adding them, it allowed me to find issues Psalm didn't catch. For example, implicit cast from object with __toString to string is not possible in strict types(or the case you had were you added a int cast). Psalm didn't catch those because it wasn't confronted to that constraint

I'm pretty neutral on this

@danog @weirdan what's your take on this?

danog:
Personally I would greatly appreciate any change that has potential for performance improvements.
However, I also think that if we enable strict types, we should also bring in directly into psalm the rules present in your repo, as they can be quite useful when working with strict types: https://github.com/orklah/psalm-strict-types

lptn:
If we decide to NOT use strict_types, another idea to consider can be declare it explicitly:

```php
declare(strict_types=0);
``

So, other devs will not create such PR again and can use git blame to see reasons why we don't want to use strict_types

weirdan:
>  what's your take on this?

Pretty ambivalent as well (read: pro-status quo).

orklah:
`declare(strict_types=0);` feels very wrong to me for some reason. Thanks for your PR but we'll just leave it like this for now at least

Issue #7606, opened 2022-02-08
Psalm 4.20.0 PHAR does not report correction version


sebastianbergmann:
```
$ wget https://github.com/vimeo/psalm/releases/download/4.20.0/psalm.phar
--2022-02-08 15:13:46--  https://github.com/vimeo/psalm/releases/download/4.20.0/psalm.phar
.
.
.
2022-02-08 15:13:48 (11.2 MB/s) - ‘psalm.phar’ saved [11749531/11749531]
```
#### Actual
```
$ php psalm.phar --version     
Psalm 4.x-dev@
```
#### Expected
```
$ php psalm.phar --version     
Psalm 4.20.0
```

orklah:
Interesting. I don't have much experience using phar, would you mind providing a PR for that? It would be cool to have this fixed indeed

mind-bending-forks:
I don't have experience of building phars (only using them!) and I don't have any experience/knowledge of the inner workings of psalm either. So, I'm starting from nothing here. I've had a dig around the code, and as far as I can tell, the building of the phar is outsourced to a dependency called box:

- Script that calls box to build the phar: https://github.com/vimeo/psalm/blob/4.x/bin/build-phar.sh
- Box: https://github.com/box-project/box
- Psalm's box config file: https://github.com/vimeo/psalm/blob/4.x/box.json.dist

The box documentation says that, by default, composer.json and composer.lock files are excluded: https://github.com/box-project/box/blob/4.0.2/doc/configuration.md#excluding-the-composer-files-exclude-composer-files They can be included by setting `exclude-composer-files` to `false`.

So, a possible solution would be to add  `"exclude-composer-files": false` to `box.json.dist`. This might fix the problem, but I couldn't say for sure without testing.

LauraTaylorUK:
Some months ago I opened issue #7921 and I was told its a duplicate of this issue, which appears "closed".

But even today with version 5.4.0 the `--version` parameter is broken:

```
$ psalm --version
Psalm dev-master@
```

Still no real version, I expected to see 5.4.0 and not dev-master@.

Thank you.

PS:
This means I can never be sure which version I have running on my system, I need to manually write down the version I've downloaded.

PR #8997, opened 2022-12-23
Fix missing version in PHARs build on GA

weirdan:
We were overriding the root version with COMPOSER_ROOT_VERSION, so all
PHARs had `dev-master` as the version for `vimeo/psalm` baked in.

Fixed vimeo/psalm#7606


orklah:
cool!

Issue #8932, opened 2022-12-18
Psalm 5: TypeDoesNotContainType is thrown instead of DocblockTypeContradiction


VincentLanglet:
The following example is throwing a TypeDoesNotContainType with Psalm 5 but not with 4
https://psalm.dev/r/9950912ce2

This seems related to the union type since
https://psalm.dev/r/66b46ed7a0
works fine.

pilif:
given the type of the `$value` parameter (array or null), after the null check, I expert $value to always be array and thus the check to be redundant, so I don't feel like this is a bug

VincentLanglet:
> given the type of the `$value` parameter (array or null), after the null check, I expert $value to always be array and thus the check to be redundant, so I don't feel like this is a bug

I agree it's redundant but i expect it to be redundant with the PHPDoc, and not a "native type".
That's why I expect a different type of error.

This should be:
```
    public function reverseTransform(array|null $value): null
    {
        if (null === $value) {
            return null;
        }

        if (!\is_array($value)) { // TypeDoesNotContainType
```
```
/**
     * @param array|null $value
     */
    public function reverseTransform($value): null
    {
        if (null === $value) {
            return null;
        }

        if (!\is_array($value)) { // DocblockTypeContradiction
```

That's why I gave the second example (without the union).
The `is_array` check is redundant, but is reported as a DocblockTypeContradiction ; the behavior I want.

This is something important for open-source library.
When type cannot be added for some reason, we cannot consider the PHPdoc as enough and need to add extra-safety check with specific message. So I generally use `treatPhpdocTypeAsCertain: false` on phpstan and ignore `DocblockTypeContradiction` on psalm.

orklah:
WIP: https://github.com/vimeo/psalm/compare/master...orklah:psalm:from_docblock?expand=1

VincentLanglet:
> WIP: https://github.com/vimeo/psalm/compare/master...orklah:psalm:from_docblock?expand=1

I added a test about this change.
Seems like psalm consider now it's a `RedundantConditionGivenDocblockType`, I would say it's right indeed.
And especially it fix my issue since I ignore both
```
DocblockTypeContradiction
RedundantConditionGivenDocblockType
```
to reproduce the behavior of `treatPhpdocTypeAsCertain: false` on PHPStan.

PR is https://github.com/vimeo/psalm/pull/8999

VincentLanglet:
[...]

PR #8999, opened 2022-12-24
Preserve from_docblock in TypeCombiner

VincentLanglet:
Closes https://github.com/vimeo/psalm/issues/8932

### DIFF
--- a/tests/TypeReconciliation/RedundantConditionTest.php
+++ b/tests/TypeReconciliation/RedundantConditionTest.php
@@ -1085,7 +1085,7 @@ class RedundantConditionTest extends TestCase
                     /** @psalm-suppress PossiblyNullArgument */
                     takesA($a);
                     if ($a instanceof A) {}',
-                'error_message' => 'RedundantConditionGivenDocblockType - src' . DIRECTORY_SEPARATOR . 'somefile.php:15',
+                'error_message' => 'RedundantCondition - src' . DIRECTORY_SEPARATOR . 'somefile.php:15',
             ],
             'replaceFalseType' => [
                 'code' => '<?php
--- a/tests/TypeReconciliation/RedundantConditionTest.php
+++ b/tests/TypeReconciliation/RedundantConditionTest.php
@@ -1085,7 +1085,7 @@ class RedundantConditionTest extends TestCase
                     /** @psalm-suppress PossiblyNullArgument */
                     takesA($a);
                     if ($a instanceof A) {}',
-                'error_message' => 'RedundantCondition - src' . DIRECTORY_SEPARATOR . 'somefile.php:15',
+                'error_message' => 'RedundantConditionGivenDocblockType - src' . DIRECTORY_SEPARATOR . 'somefile.php:15',
             ],
             'replaceFalseType' => [
                 'code' => '<?php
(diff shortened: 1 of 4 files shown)


orklah:
Thanks for pushin that. Unfortunately, there was a reason why I didn't merged it yet.

I *think* this test: https://github.com/vimeo/psalm/pull/8999/files#diff-5ad90d98b0768c1fdb94a7e73224d8709dd256fec470744ea1785b61ab72b39fL1088

is supposed to check that calling a methode that takes only `A` refines the original type from `A|null` (from docblock) to `A` (not from docblock)

The fact that you needed to go from `RedundantCondition` to `RedundantConditionGivenDocblockType` seems to indicate this is not working anymore. That's why I paused and I planned to check some things:
- check if the test was really supposed to test what I think
- check if it was already changed recently (to go from `RedundantConditionGivenDocblockType` to `RedundantCondition`)
- check if the thing that restrict the parameter after a method call did consider removing the from_docblock property

In order to know if it's a regression or just something that was not implemented and the test was changed when the from_docblock preservation broke

If you have time to look at this, I'd be glad :) Otherwise, I'll try to check this next week

tests/TypeReconciliation/RedundantConditionTest.php
```diff
@@ -1546,6 +1546,27 @@ function f(array $p) : void {
                     }',
                 'error_message' => 'RedundantCondition',
             ],
+            'https://github.com/vimeo/psalm/issues/8932' => [
```
orklah, line 1549: Having a reference is a great idea but I still would like to have some kind of title Maybe a comment in the code itself? If someday, we were to migrate from github, we could loose the meaning behind the tests completely

VincentLanglet:
> I _think_ this test: https://github.com/vimeo/psalm/pull/8999/files#diff-5ad90d98b0768c1fdb94a7e73224d8709dd256fec470744ea1785b61ab72b39fL1088
> 
> is supposed to check that calling a methode that takes only `A` refines the original type from `A|null` (from docblock) to `A` (not from docblock)
> 
> The fact that you needed to go from `RedundantCondition` to `RedundantConditionGivenDocblockType` seems to indicate this is not working anymore.

Indeed you're right.

> * check if the test was really supposed to test what I think

When reading again the test, it seems so.

> * check if it was already changed recently (to go from `RedundantConditionGivenDocblockType` to `RedundantCondition`)

No, this wasn't changed recently.

> * check if the thing that restrict the parameter after a method call did consider removing the from_docblock property

Seems like it need to be improve

VincentLanglet:
> If you have time to look at this, I'd be glad :) Otherwise, I'll try to check this next week

To me, it's an issue when creating the TArray type https://github.com/vimeo/psalm/issues/8932#issuecomment-1364664519

VincentLanglet:
The issue seems to be inside the
```
TypeCombiner::combine
```
method.

The input is types:
```
[Array(from_docblock = true), Null(from_docblock = true)]
```
but when `getArrayTypeFromGenericParams` is called there is 
```
$array_type = new TArray($generic_type_params);
```
which doesn't care about the previous `from_docblock` part.

I tried adding `from_docblock` everywhere in the TypeCombiner and it seems to solve everything.
I dunno if I missed some or if we have to remove some.

orklah:
Seems like CI is failing, are you stuck?

VincentLanglet:
> Seems like CI is failing, are you stuck?

Didn't saw the last commit introduced an error. We could have wait for a long time ^^'.
This is fixed now @orklah.

orklah:
Thanks!

PR #9000, opened 2022-12-24
restrictReturnTypes configuration documentation

jack-worman:
Also removed redundant default configuration in psalm.xml.dist


orklah:
Cool! Can you please add a warning to say that forcing the tighter type is not always the best course of action and this may unexpected results?

This is valid: 
https://psalm.dev/r/565ff971d6
but this is not:
https://psalm.dev/r/e85aeea6cf

That's the reason this config will probably never be true by default. It's just a nice tool to check once in a while

jack-worman:
> Can you please add a warning

Done

orklah:
Thanks!

PR #9001, opened 2022-12-24
Make `$http_response_header` a non-empty-list

fluffycondor:
This list always contains at least one element: http status code. So I think it's safe to assume it's a non-empty-list.
https://www.php.net/manual/en/reserved.variables.httpresponseheader.php

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php
@@ -612,7 +612,12 @@ class VariableFetchAnalyzer
         }
 
         if ($var_id === '$http_response_header') {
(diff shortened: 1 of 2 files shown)


weirdan:
It's undefined in CLI though:

image

fluffycondor:
> It's undefined in CLI though

Yep. the current inmplementation is quite naive, because $http_response_header exists only after performing a successful network request.
PHPStan assumes it exists only after calling `fopen` or `file_get_contents`: https://github.com/phpstan/phpstan-src/blob/1.9.x/src/Analyser/NodeScopeResolver.php#L1993

weirdan:
It probably makes sense to mark it as possibly undefined, similar to `$_SERVER['argv']`:
```php
            $argv_helper = new Union([
                Type::getNonEmptyListAtomic(Type::getString())
            ], ['possibly_undefined' => true]);
```

orklah:
Thanks!

PR #9002, opened 2022-12-24
forbidden function bug and better get_defined_functions() signature

jack-worman:
* The full namespace of the used function was not being considered when checking if it is forbidden.
  * I am surprised that names are not fully qualified before analysis
* Gave `get_defined_functions()` its proper array shape.
  * Sad I can't do `callable-string&lowercase-string`


weirdan:
>  I am surprised that names are not fully qualified before analysis

I think they are:

https://github.com/vimeo/psalm/blob/8b05f2ead810d535adbfba7dbc9d4fdeef18b5d6/src/Psalm/Internal/PhpVisitor/SimpleNameResolver.php#L116-L119

weirdan:
Thanks!

PR #9003, opened 2022-12-25
Use TString properties instead of inheritance

jack-worman:
Goals:
- Use properties instead of inheritance to more strictly define `TString`
  - Once fully implemented, I believe this will simplify working with string types
- Allow intersection types between strings
- Easily allow the negation of any property

TODO (in this PR):
- [ ] Update `TNonEmptyLowercaseString` to use `TString::$lowercase`
- [ ] Support `TString::$lowercase = false` (non-lowercase-string)

Future scope:
- Deprecate all `TString` child classes and convert to use a property

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="dev-master@91081f77fdd47a35d12bd87b31291c95f98be8ae">
+<files psalm-version="dev-master@8b05f2ead810d535adbfba7dbc9d4fdeef18b5d6">
   <file src="examples/TemplateChecker.php">
     <PossiblyUndefinedIntArrayOffset>
       <code>$comment_block-&gt;tags['variablesfrom'][0]</code>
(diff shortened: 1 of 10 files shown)


jack-worman:
And notes/comments/opinions would be much appreciated!

orklah:
A few things:
- You introduce a new notion of `non-lowercase-string`. Any function that alters strings should be checked for correct stubbing, return type providers also. I'd advise first dealing with `lowercase-string` and introduce this new type in another PR
- You made the lowercase property public. While you don't do it yet, if the property were to change (for example to null), you would get to a weird state where you have a `TLowercaseString` that behaves like a `TString`. However, `TString::isPlain` is actually checking that the class must be a `TString` for the string to be plain. I'd say you need to choose between having a public property and keeping `TLowercaseString` at all (but removing it would mean BC break so Psalm 6)
- If we get to the point where we have properties for all childs of `TString`, we'll probably have to find a new way to express those types. We can't continue having a special keyword for each combination (non-falsy-non-lowercase-callable-string anyone? 😄). I believe PHPStan have a notion of accessory types: https://github.com/phpstan/phpstan-src/tree/1.9.x/src/Type/Accessory They are sometimes displayed as intersection with the main type (which means you can have a `array{a?: string, b?:string}&non-empty-array` that describe that while every key is optional, the array is not empty). We could use a similar notation for non obvious cases (for example `string&callable&non-lowercase&non-falsy`)

PR #9007, opened 2022-12-26
Add imap_is_open() to PHP 8.2 dictionary

lptn:
New function added to PHP 8.2.1: https://github.com/php/php-src/pull/10094

Also, it can be potentially added to PHP 8.1.15 (I'll update dictionaries in this case)


weirdan:
Thanks!

Issue #9008, opened 2022-12-26
__destruct on an interface throws MissingReturnType


JoshuaBehrens:
I assume [MissingReturnType](https://psalm.dev/050) on this comes originally from https://github.com/vimeo/psalm/issues/1846 . I also assume, that __construct and __destruct to not need a return type. My assumptions can be wrong.

What I noticed when I run psalm on a project that I get a MissingReturnType in an interface expecting __destruct:
https://psalm.dev/r/fdd8a87e28

Is this a bug? On classes it does not break.

JoshuaBehrens:
Thank you for fixing this so quickly :O

PR #9009, opened 2022-12-26
Do not require return type on destructors in interfaces

weirdan:
Fixes vimeo/psalm#9008


JoshuaBehrens:
Thank you for fixing this so quickly :O

Issue #3886, opened 2020-07-24
Psalm should not report MissingFile if file is checked if it exists before include/require


filips123:
Psalm reports MissingFile error when using `include` or `require` on a file that does not exist. This is normal, because you shouldn't be including unexisting file.

However, it also reports the same error, even if you used `is_file` to check if file exists and skip `include`/`require` if it doesn't. I think that this shouldn't cause error, because it also won't cause error when actually running script in PHP. Checking if file exists is useful if you want to include some user settings, which are not guaranteed and required to exist, and most likely won't exist when you are running Psalm in CI.

Example snippets which fail:

```php
<?php

$settings = is_file(__DIR__ . '/settings.php') ? require __DIR__ . '/settings.php' : [];
```

```php
<?php

if (is_file(__DIR__ . '/settings.php')) {
    $settings = require __DIR__ . '/settings.php';
} else {
    $settings = [];
}
```

This causes:

```
ERROR: MissingFile - test1.php:3:50 - Cannot find file /tmp/psalm-test/settings.php to include (see https://psalm.dev/107)
$settings = is_file(__DIR__ . '/settings.php') ? require __DIR__ . '/settings.php' : [];


ERROR: MissingFile - test2.php:4:17 - Cannot find file /tmp/psalm-test/settings.php to include (see https://psalm.dev/107)
    $settings = require __DIR__ . '/settings.php';
```

I can't show example on https://psalm.dev because file includes are not allowed there.

muglug:
The issue here is that Psalm can't figure out what you're trying to include (because it wants to analyse the contents). Feel free to ignore the issue.

PR #9012, opened 2022-12-28
dont report MissingFile error when file_exists is validated beforehand

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/3886


orklah:
The issue you're linking explicitly tell the issue with that. If you're including a file and then using symbols from the file, Psalm won't be able to analyze properly and just silently fail

orklah:
@weirdan what do you think?

orklah:
Thanks!

PR #9013, opened 2022-12-28
Partial ScalarTypeComparator rewrite

jack-worman:
Scalar family tree for reference:
Scalar

Add support and tests for the following container types:
- [x] TArrayKey
- [x] TNumeric
- [x] TScalar
- [x] TBool
- [x] TFloat
- TInt (will do in a future PR)
- TString (will do in a future PR)


jack-worman:
Ready for review

src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
```diff
@@ -366,15 +357,6 @@ public static function isContainedBy(
             return false;
         }
 
-        if (get_class($input_type_part) === TFloat::class && $container_type_part instanceof TLiteralFloat) {
-            if ($atomic_comparison_result) {
-                $atomic_comparison_result->type_coerced = true;
```
orklah, line 371: This is dropped from here but I don't see where it was moved

src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
```diff
@@ -550,4 +525,115 @@ public static function isContainedBy(
...
+            if ($key_types) {
+                foreach ($key_types->getAtomicTypes() as $key_type) {
+                    if (AtomicTypeComparator::isContainedBy($codebase, $input_type, $key_type)) {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+        throw new Exception('Unexpected container type class: ' . $container_type_class);
```
orklah, line 552: This needs to be less cryptic. If this is reported to us in the future, we need to know in which method this container was found to be able to fix

src/Psalm/Internal/Type/Comparator/ScalarTypeComparator.php
```diff
@@ -55,6 +61,44 @@ public static function isContainedBy(
...
+            } elseif ($container_type_part instanceof TBool) {
+                $result = self::isContainedByBool($input_type_part, $container_type_part);
+            } else {
+                $result = self::isContainedByFloat($input_type_part, $container_type_part, $allow_float_int_equality);
+            }
+            if ($result) {
+                return true;
+            } else {
+                if ($atomic_comparison_result) {
+                    // The type is coerced if the container type is contained by the input type (the opposite).
```
orklah, line 86: Are you confident about this rule working in every case? I would have like to have TString and TInt comparison in this PR also because those are the most complex and I fear some generalization here could be a problem

jack-worman: I cannot think of another way of defining it. It is when you try to put a bigger type into a smaller type.

In the mean time, I'll work on adding TInt support.

orklah: > It is when you try to put a bigger type into a smaller type.

I agree in the general case but I wonder if there are some edge cases. Think like having a `numeric-string` being pushed where `lowercase-string` is expected. There's not a big or small type in this case, but it should maybe be a scalar coercion still?

Also pushing `int` where `positive-int` is expected would be a scalar coercion, but based on the same definition, pushing a `int<-1, 3>` into a `int<0, 4>` would not even though it's clearly the same kind of operation with restricted range

I just want to be sure we won't merge a partial work for which we will find fatal flaws later

PR #9016, opened 2022-12-28
Refined `explode()` types

### LINKED ISSUE #5039: Better types for the `explode()` function?
I was trying to get more precise types for `explode($delimiter, $input, $limit = PHP_MAX_INT)`, which is used quite a lot.

Specifically:

 * `explode()` returns `array<empty, empty>|non-empty-list<string>|false` on PHP 7.4
 * `explode()` is `array<empty, empty>|non-empty-list<string>|never-return` on PHP 8.x, with `@throws ValueError` (https://3v4l.org/rmaRS)

I tried applying some improvements to the stubs, but am a bit puzzled about what needs to be done ( https://psalm.dev/r/b52283d0f5 ).

In the case of `$limit` being negative (`!positive-int && !0`, not sure if there's a way to verify this), the return type is `array<empty, empty>`.
In the case of `$limit` being `positive-int|0`, the return type is restricted to `non-empty-list<string>|false` or `non-empty-list<string>|never-return`.

## The Question

The big question is:

 * do we treat a `ValueError` as a type error, as in require `explode()` to have a `non-empty-string` for `$delimiter`?
 * do we treat a `ValueError` as an assertion (`@psalm-assert`)?
 * do we ignore the `ValueError` and instead map the `never-return`?

Overall, I tried a bit of everything, and am tempted to say (my opinion!) that people should never expect a `ValueError` nor an `array<empty, empty>` here, and we should map the function as `explode(non-empty-string $delimiter, string $input, positive-int|0 $limit)`

Thoughts?

### LINKED ISSUE #7831: Constant as separator in explode() produces false-positives
Errors with const: https://psalm.dev/r/d5a6c69fdc
No errors without const: https://psalm.dev/r/76b2904ea2

Some relevant comments from Slack:

Error comes from here: https://github.com/vimeo/psalm/blob/b924032850df54dd05ffae409a709820fd6e688e/src/Psalm/Internal/Provider/ReturnTypeProvider/ExplodeReturnTypeProvider.php#L56

> this condition is suboptimal because it checks that the parameter is a literal string, not a variable nor a a constant

> Yes with SPLIT_SEPARATOR psalm thinks, the separator could be empty. which would result in returning false before php 8.0.
So I think there are two bugs. psalm should detect that the constant is not empty. and psalm should know, that explode can not return false anymore since php 8.0.

Ocramius:
Fixes #5039
Fixes #7831

This patch removes the need for a custom function return type
provider for `explode()`, and instead replaces all that with a single
stub for the `explode()` function, which provides types for some of
the most common `$limit` input values.

With this change, the `$delimiter` is enforced to be a `non-empty-string`,
which will lead to downstream consumers having to adjust some code accordingly,
but that shouldn't affect the most common scenario of exploding a string
based with a constant `literal-string` delimiter, which most PHP devs tend to do.

This change didn't come with an accompanying test, since that would be a bit
wasteful, but it was verified locally with following script:

```php
<?php

$possible0  = explode(',', 'hello, world', -100);
$possible1  = explode(',', 'hello, world', -1);
$possible2  = explode(',', 'hello, world', 0);
$possible3  = explode(',', 'hello, world', 1);
$possible4  = explode(',', 'hello, world', 2);
$possible5  = explode(',', 'hello, world', 3);
$possible6  = explode(',', 'hello, world', 4);
try {
    $impossible1 = explode('', '', -1);
} catch (Throwable $impossible1) {}

$traced = [$possible0, $possible1, $possible2, $possible3, $possible4, $possible5, $possible6, $impossible1];

/** @psalm-trace $traced */

var_dump($traced);

return $traced;
```

Running psalm locally, this produces:

```
psalm on  feature/#5039-more-refined-types-for-explode-core-function [?] via 🐘 v8.1.13 via ❄️  impure (nix-shell) 
❯ ./psalm --no-cache explode.php 
Target PHP version: 7.4 (inferred from composer.json) Extensions enabled: dom, simplexml (unsupported extensions: ctype, json, libxml, mbstring, tokenizer)
Scanning files...
Analyzing files...

░

ERROR: InvalidArgument - explode.php:11:28 - Argument 1 of explode expects non-empty-string, but '' provided (see https://psalm.dev/004)
    $impossible1 = explode('', '', -1);


ERROR: PossiblyUndefinedGlobalVariable - explode.php:14:96 - Possibly undefined global variable $impossible1 defined in try block (see https://psalm.dev/126)
$traced = [$possible0, $possible1, $possible2, $possible3, $possible4, $possible5, $possible6, $impossible1];


ERROR: ForbiddenCode - explode.php:18:1 - Unsafe var_dump (see https://psalm.dev/002)
/** @psalm-trace $traced */

var_dump($traced);


ERROR: Trace - explode.php:18:1 - $traced: list{0: array<never, never>, 1: non-empty-list<string>, 2: list{string}, 3: list{string}, 4: array{0: string, 1?: string}, 5: array{0: string, 1?: string, 2?: string}, 6: non-empty-list<string>, 7?: Throwable|non-empty-list<string>} (see https://psalm.dev/224)
/** @psalm-trace $traced */

var_dump($traced);


4 errors found

Checks took 6.31 seconds and used 265.386MB of memory
```

The actual runtime behavior on PHP 8.x: https://3v4l.org/0NKlW

```
array(8) {
  [0]=>
  array(0) {
  }
  [1]=>
  array(1) {
    [0]=>
    string(5) "hello"
  }
  [2]=>
  array(1) {
    [0]=>
    string(12) "hello, world"
  }
  [3]=>
  array(1) {
    [0]=>
    string(12) "hello, world"
  }
  [4]=>
  array(2) {
    [0]=>
    string(5) "hello"
    [1]=>
    string(6) " world"
  }
  [5]=>
  array(2) {
    [0]=>
    string(5) "hello"
    [1]=>
    string(6) " world"
  }
  [6]=>
  array(2) {
    [0]=>
    string(5) "hello"
    [1]=>
    string(6) " world"
  }
  [7]=>
  object(ValueError)#1 (7) {
    ["message":protected]=>
    string(51) "explode(): Argument #1 ($separator) cannot be empty"
    ["string":"Error":private]=>
    string(0) ""
    ["code":protected]=>
    int(0)
    ["file":protected]=>
    string(9) "/in/0NKlW"
    ["line":protected]=>
    int(11)
    ["trace":"Error":private]=>
    array(1) {
      [0]=>
      array(4) {
        ["file"]=>
        string(9) "/in/0NKlW"
        ["line"]=>
        int(11)
        ["function"]=>
        string(7) "explode"
        ["args"]=>
        array(3) {
          [0]=>
          string(0) ""
          [1]=>
          string(0) ""
          [2]=>
          int(-1)
        }
      }
    }
    ["previous":"Error":private]=>
    NULL
  }
}
```

On PHP 7:

```
Warning: explode(): Empty delimiter in /in/0NKlW on line 11
array(8) {
  [0]=>
  array(0) {
  }
  [1]=>
  array(1) {
    [0]=>
    string(5) "hello"
  }
  [2]=>
  array(1) {
    [0]=>
    string(12) "hello, world"
  }
  [3]=>
  array(1) {
    [0]=>
    string(12) "hello, world"
  }
  [4]=>
  array(2) {
    [0]=>
    string(5) "hello"
    [1]=>
    string(6) " world"
  }
  [5]=>
  array(2) {
    [0]=>
    string(5) "hello"
    [1]=>
    string(6) " world"
  }
  [6]=>
  array(2) {
    [0]=>
    string(5) "hello"
    [1]=>
    string(6) " world"
  }
  [7]=>
  bool(false)
}
```


src/Psalm/Internal/Provider/ReturnTypeProvider/ExplodeReturnTypeProvider.php
```diff
@@ -1,103 +0,0 @@
...
-use Psalm\Type\Atomic\TNonEmptyString;
-use Psalm\Type\Atomic\TString;
-use Psalm\Type\Union;
-
-use function count;
-
-/**
- * @internal
- */
-class ExplodeReturnTypeProvider implements FunctionReturnTypeProviderInterface
```
Ocramius, line 21: Haven't run tests for this: waiting for CI to tell me how bad this change will affect this patch :D

stubs/CoreGenericFunctions.phpstub
```diff
@@ -725,6 +725,26 @@ function join($separator, array $array = []): string
...
+ *         : (
+ *             $limit is int<-1, 1>
+ *                 ? array{string}
+ *                 : (
+ *                     $limit is 2
+ *                         ? array{0: string, 1?: string}
+ *                         : (
+ *                             $limit is 3
+ *                                 ? array{0: string, 1?: string, 2?: string}
+ *                                 : non-empty-list<string>
```
Ocramius, line 742: Note: I stopped at `3` array keys, but this could be made bigger and uglier in future.

For now, I think this is a decent trade-off to help most developers (most `explode()` calls provide a limit of `2`, and you should check for array key existence  regardless)

stubs/CoreGenericFunctions.phpstub
```diff
@@ -725,6 +725,26 @@ function join($separator, array $array = []): string
...
  * @psalm-pure
  *
+ * @param non-empty-string $separator
+ *
+ * @return (
+ *     $limit is int<min, -2>
+ *         ? array<empty, empty>
+ *         : (
+ *             $limit is int<-1, 1>
+ *                 ? array{string}
```
Ocramius, line 735: Never mind: this is wrong, since `-1` has the special meaning of "no limit". Will push and amend.

stubs/CoreGenericFunctions.phpstub
orklah: `list{0: string, 1?: string, 2?: string}` would be better here (it would carry the meaning that the `2` offset can't exists if `1` does not

psalm-baseline.xml
```diff
@@ -695,6 +701,8 @@
     <PossiblyUnusedMethod>
       <code>allFloatLiterals</code>
       <code>allFloatLiterals</code>
+      <code>hasLowercaseString</code>
+      <code>hasLowercaseString</code>
```
Ocramius, line 705: This code is not marked `@internal`, so I didn't drop it.

tests/FunctionCallTest.php
```diff
@@ -433,52 +433,63 @@ public static function Baz($mixed) : string {
                      */
                     $elements = explode(" ", $string, $limit);',
                 'assertions' => [
-                    '$elements' => 'list<string>',
+                    '$elements' => 'array{0?: string, 1?: string, 2?: string, ...<int<0, max>, string>}',
```
Ocramius, line 436: This is kinda ugly: the array type changed from `list<string>` to a weird union type here, possibly due to the fact that psalm performs a union type between `array<never, never>|array{string}|array{0: string, 1?: string}|non-empty-list<string>`.

This is a bit of a weird one, and unsure if it indicates a problem with this approach

tests/FunctionCallTest.php
```diff
@@ -433,52 +433,63 @@ public static function Baz($mixed) : string {
...
             'explodeWithDynamicDelimiterAndNegativeLimit' => [
                 'code' => '<?php
                     /**
-                     * @var string $delim
+                     * @var non-empty-string $delim
                      * @var string $string
                      */
                     $elements = explode($delim, $string, -5);',
                 'assertions' => [
-                    '$elements' => 'false|list<string>',
```
Ocramius, line 469: The `false` scenario is gone gone gone

stubs/CoreGenericFunctions.phpstub
Ocramius: Oooooh, I didn't know about this syntax, thanks!

tests/FunctionCallTest.php
Ocramius: Related: https://github.com/vimeo/psalm/pull/9016#discussion_r1058458616

stubs/CoreGenericFunctions.phpstub
Ocramius: c0c01168091ef8a2eadf8b06532f303b480d6686

tests/FunctionCallTest.php
Ocramius: c0c01168091ef8a2eadf8b06532f303b480d6686

Ocramius:
The CS failure seems to be unrelated to this patch.

tests/FunctionCallTest.php
```diff
@@ -433,52 +433,63 @@ public static function Baz($mixed) : string {
...
                     /**
-                     * @var string $delim
+                     * @var non-empty-string $delim
                      * @var string $string
                      * @var int $limit
                      */
                     $elements = explode($delim, $string, $limit);',
                 'assertions' => [
-                    '$elements' => 'false|list<string>',
+                    '$elements' => 'list{0?: string, 1?: string, 2?: string, ...<int<0, max>, string>}',
```
orklah, line 492: I wonder if adding an explicit  `$limit is int` case after the `$limit is 3` would prevent this smashing of type and still convey the info that `0` is definitely defined?

orklah:
> The CS failure seems to be unrelated to this patch.

this was fixed since ;)

tests/FunctionCallTest.php
Ocramius: I can try adding such a case, although tomorrow 😁

orklah:
Todo: check if this fixes https://github.com/vimeo/psalm/issues/7831

tests/FunctionCallTest.php
Ocramius: I checked this: doesn't seem to have any effect (which is kinda expected)

Ocramius:
I checked the code examples from #7831: they are fixed by this patch :+1:

tests/FunctionCallTest.php
orklah: why is this expected? I understand that when the type is mixed, Psalm just merge all types together, but when it's a standard int, it should take the correct type, no?

Ocramius: If I add an explicit `$limit is int` at the end of the conditional, it is equivalent to having the default case.

orklah:
Thanks!

PR #9019, opened 2022-12-28
Prevent DB escaping functions from affecting non-sql taints

mmcev106:
I found a developer who was using `mysql_escape_string()` to prevent Psalm from reporting non-sql taint types.  This PR aims to prevent any DB escape method from being misused in this way.  I'm happy to do anything a different way if you'd prefer.


orklah:
Thanks!

PR #9023, opened 2022-12-29
Fix never return union type wrong condition

kkmuffme:
Fix never return union type wrong condition

Related https://github.com/vimeo/psalm/pull/8624


orklah:
Thanks!

PR #9025, opened 2022-12-29
fixes return type issues of methods reported for the wrong file

### LINKED ISSUE #8457: InvalidReturnType are reported for stubs

kkmuffme:
fixes return type issues of methods reported for the wrong file

Fix #8457

See https://github.com/vimeo/psalm/pull/8503 which fixed this issue for functions

### DIFF
--- a/src/Psalm/Internal/Codebase/Methods.php
+++ b/src/Psalm/Internal/Codebase/Methods.php
@@ -920,6 +920,12 @@ class Methods
 
         $storage = $this->getStorage($method_id);
 
+        // if function exists in stubs and in analyzed code
+        // use the return type location of the analyzed code instead of the stubbed location
+        if ($storage->stubbed) {
+            return null;
+        }
+
         if (!$storage->return_type_location) {
             $overridden_method_ids = $this->getOverriddenMethodIds($method_id);
 


orklah:
Thanks!

Issue #8912, opened 2022-12-17
Psalter wants to add "@result never" in void methods


UlrichBoehm:
Running the following code with psalter (dry-run, all issues) leads to the suggestion to add a doc block with "(at)return never" which is wrong.
```
function checkForFoo(string $test): void
{
    if (str_contains($test, 'foo')) {
        throw new \Exception('No Foo here!');
    }
}
```
FYI @kkmuffme

UlrichBoehm:
It's reproducible as soon as you enable the setting "Force return types to be as tight as possible".
https://psalm.dev/r/40196637cc

Issue #8862, opened 2022-12-08
Incorrect inferred return type 'never'


jack-worman:
https://psalm.dev/r/4d5cfb806a

`void|never` is being incorrectly reduced to `never` instead of `void`.

weirdan:
Currently it's ef02ded - a couple of commits behind the HEAD of master.

weirdan:
@kkmuffme it requires (undocumented) `restrictReturnTypes="true"` attribute (`Force return types to be as tight as possible` on psalm.dev).

PR #9026, opened 2022-12-29
fix implicit return type

kkmuffme:
* fix implicit void return type errors https://psalm.dev/r/b0e397c14e (this issue existed since forever in psalm)
* fix void + never = null return type error https://psalm.dev/r/6f227ae9b9 (partially existed since forever, partially caused by https://github.com/vimeo/psalm/pull/8624)

Fix https://github.com/vimeo/psalm/issues/8912 
Fix https://github.com/vimeo/psalm/issues/8862

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
@@ -215,7 +215,6 @@ class ReturnTypeAnalyzer
             return null;
         }
 
-
         if ($return_type
             && $return_type->isNever()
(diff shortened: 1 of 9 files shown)


orklah:
- About the first one, you can suppress that (please add a comment to say there's a bug here)
- For the Unit test 4, I don't agree this is an existing bug. The test works fine right now. The issue is that the option `allow_backwards_incompatible_changes` is to false, so Psalm should not risk changing signature type (the option does not exists on psalm.dev so it's probably why it makes the change in the sandbox)
- For the other test, you should be able to fix that by having a `return null;` in each method. It should make Psalm happy

orklah:
Thanks!

PR #9027, opened 2022-12-29
Assign TMixed when we find an irreconciliable assertion from docblock

orklah:
This is to keep consistency with things like https://github.com/vimeo/psalm/blob/0f4db694df248fdd2064b5f6e0be0d6cd4fbbebe/src/Psalm/Internal/Type/SimpleAssertionReconciler.php#L2314

When we get to a state where an assertion removed every possible type, we should return never only if the type was from a signature. Otherwise, it could be a wrong docblock (or a user checking an impossible type given the docblock)


orklah:
Can't be done, switch completion depend on getting never apparently

PR #9030, opened 2022-12-30
fix @psalm-scope-this not working with FQCN with leading backslash

kkmuffme:


orklah:
Can you add a test for that please?

orklah:
Thanks!

PR #9035, opened 2022-12-31
Fix iterator and exception callmaps

othercorey:
Did not change getInnerIterator() because I'm not sure if it will ever return null.


orklah:
Looks great! Seems like there's a few mismatch on delta callmap left to fix and it should be good to merge :)

othercorey:
> Looks great! Seems like there's a few mismatch on delta callmap left to fix and it should be good to merge :)

Always something with these changes.

orklah:
Thanks!

PR #9036, opened 2022-12-31
Fix 9034

### LINKED ISSUE #9034: Deprecation message breaks JSON output
https://github.com/psalm/codeception-psalm-module/actions/runs/3810538569/jobs/6482681536#step:10:283

```
Failed to parse output: Warning: "findUnusedCode" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.
```

This has the potential to break tests for most of the plugins. The Codeception module used for testing redirects STDERR to STDOUT, so the JSON it then tries to read won't be valid.

/cc: @jack-worman

jack-worman:
Fixes #9034

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -9,6 +9,8 @@
 
 - [BC] The `TDependentListKey` type was removed and replaced with an optional property of the `TIntRange` type.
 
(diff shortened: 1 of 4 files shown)


jack-worman:
Making a draft because @danog does not think this is the correct solution:
https://github.com/vimeo/psalm/issues/9034#issuecomment-1368224450

PR #9053, opened 2023-01-03
Fix `ReflectionFunctionAbstract::getExtensionName()` stub

kukulich:
https://3v4l.org/XaplB

I'm not able to find a situation when the extension name can be empty string.


orklah:
Thanks!

PR #9061, opened 2023-01-05
Make it more clear what do we test from Psalm output

lptn:
`tests/ProjectCheckerTest.php` test is too dependent from any changes in Psalm output. I believe that the main assertion is `$this->assertSame(0, IssueBuffer::getErrorCount());` and it perfectly works.

I propose instead of asserting for the full output to check some important substrings. This is not allow to catch some unintended output, but I believe this is now the responsibility of these existing tests.

This is first PR, in the [next one](https://github.com/vimeo/psalm/compare/master...lptn:psalm:cleaunup-unsupported-extentions-message?expand=1) I'm going to change Psalm output (cleanup `unsupported extensions` message from core and bundled extensions), so these changes will simplify my next PR

### DIFF
--- a/tests/ProjectCheckerTest.php
+++ b/tests/ProjectCheckerTest.php
@@ -38,13 +38,6 @@ class ProjectCheckerTest extends TestCase
 
     protected ProjectAnalyzer $project_analyzer;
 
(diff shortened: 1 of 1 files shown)


orklah:
Thanks!

PR #9063, opened 2023-01-05
Internal: use more specific types

lptn:
Use `non-empty-list<Union>` instead of `Union[]`

I'm working on updating psalm/laravel-plugin  and got some issues caused by Union[] `non-empty-list<Union>` instead of `Union[]`.


weirdan:
Thanks!

Issue #9065, opened 2023-01-06
False positive for missing BackedEnum::cases()


BenMorel:
Not sure why, but the following code fails both on `5.4.0` and on the latest commit, 8329309:

```php
<?php

interface FeatureFlagInterface extends BackedEnum
{
}

function test(FeatureFlagInterface $featureFlag): void
{
    $featureFlag::cases();
}
```

> ERROR: UndefinedMethod - src/test.php:9:5 - Method FeatureFlagInterface::cases does not exist (see https://psalm.dev/022)
    $featureFlag::cases();

Even though the Psalm website does not report any error, using the same commit:

https://psalm.dev/r/8e66ae9cd8

---

Reproduce:

```sh
mkdir -p bug/src && cd bug
cat > src/test.php <<'EOF'
<?php

interface FeatureFlagInterface extends BackedEnum
{
}

function test(FeatureFlagInterface $featureFlag): void
{
    $featureFlag::cases();
}
EOF
composer require vimeo/psalm:dev-master#8329309
vendor/bin/psalm --init src 1
vendor/bin/psalm
```

BenMorel:
Thanks for the fix! Do you know why `psalm.dev` did not report any issue, using the same commit?

PR #9067, opened 2023-01-06
fix `I::cases()` where `interface I extends BackedEnum`

weirdan:
Fixes vimeo/psalm#9065

Inheritance in stubs seems to be broken


orklah:
Thanks!

BenMorel:
Thank you!

PR #9072, opened 2023-01-06
Narrow down the return type of DOMXPath::query()

MidnightDesign:
See https://www.php.net/manual/en/domxpath.query.php#refsect1-domxpath.query-returnvalues


orklah:
Thanks!

PR #9073, opened 2023-01-07
Update Call Maps and Signature for get_headers

### LINKED ISSUE #9066: Signature for get_headers() is incorrect for PHP 8+
In PHP 8.0.0 the  associative parameter to get_headers was changed from an integer to a boolean. https://www.php.net/manual/en/function.get-headers.php

In psalm if you set the associative parameter as a boolean you get this error "Argument 2 of get_headers expects int, but true provided"

Example. 
https://psalm.dev/r/d4db103791

Incorrect Stub https://github.com/vimeo/psalm/blob/master/stubs/CoreGenericFunctions.phpstub#L1489

therealryanbonham:
Fixes #9066 

Marks the associative paramater as bool in the CallMap and CallMap_80. Also fixes context in CallMap and CallMap_71 to get ?resource.  Updates stubs/CoreGenericFunctions.phpstub to match as well. I was not sure if stub should also be updated or just callmaps. If stub change is not needed, i will be happy to adjust the PR.

Thanks

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1482,11 +1482,11 @@ function bin2hex(string $string): string {}
  *
  * @param resource|null $context
  *
- * @return ($associative is false ? list<string> : array<string, string|list<string>>)|false
+ * @return ($associative is false|0 ? list<string> : array<string, string|list<string>>)|false
  *
  * @psalm-taint-sink ssrf $url
  */
-function get_headers(string $url, bool $associative = false, $context = null) : array|false {}
+function get_headers(string $url, int $associative = 0, $context = null) : array|false {}
 
 /**
  * @psalm-pure
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1482,11 +1482,11 @@ function bin2hex(string $string): string {}
  *
  * @param resource|null $context
  *
- * @return ($associative is 0 ? list<string> : array<string, string|list<string>>)|false
+ * @return ($associative is false ? list<string> : array<string, string|list<string>>)|false
  *
  * @psalm-taint-sink ssrf $url
  */
-function get_headers(string $url, int $associative = 0, $context = null) : array|false {}
+function get_headers(string $url, bool $associative = false, $context = null) : array|false {}
 
 /**
  * @psalm-pure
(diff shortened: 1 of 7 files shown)


stubs/CoreGenericFunctions.phpstub
```diff
@@ -1482,11 +1482,11 @@ function bin2hex(string $string): string {}
  *
  * @param resource|null $context
  *
- * @return ($associative is 0 ? list<string> : array<string, string|list<string>>)|false
+ * @return ($associative is false ? list<string> : array<string, string|list<string>>)|false
```
orklah, line 1485: The change is needed, but the stubs are valid for each php version so this need to stay compatible with old versions.

Can you add handling for both false and 0?

stubs/CoreGenericFunctions.phpstub
```diff
@@ -1482,11 +1482,11 @@ function bin2hex(string $string): string {}
...
  *
  * @psalm-taint-sink ssrf $url
  */
-function get_headers(string $url, int $associative = 0, $context = null) : array|false {}
+function get_headers(string $url, bool $associative = false, $context = null) : array|false {}
```
orklah, line 1489: This will be harder to handle because we don't have conditional params so you can't really have different types for different versions. Could you try to leave this stub as it is here and try to duplicate it in PHP80.phpstub ?
The hope is that the second stub will override the first but I'm not sure it will.

therealryanbonham: Seems to work with my local testing...  I test my example file in the bug report with   php ../psalm --php-version=7.4.0 --no-cache and  php ../psalm --php-version=8.0.0--no-cache with both int and bool and correct correct results for each version. 

I am happy to also add a test case for this, i am uncertain where in the tests that should live... If you can provide some guidance I am to add that.

orklah: You can add your tests on FunctionCallTest::providerValidCodeParse :)

therealryanbonham: Thanks for the pointer. Test are added.

stubs/CoreGenericFunctions.phpstub
therealryanbonham: This is completed.

stubs/CoreGenericFunctions.phpstub
orklah: Wan you add the expected return type in the assertion array part of the test?

In PHP 7 and 8, you should get slightly different types, this should help us see if the return behaves correctly too

therealryanbonham: Assertions are added.

orklah:
Thanks!

PR #9084, opened 2023-01-08
Fix CallMap for Reflection(Union|Intersection)Type::getTypes()

### LINKED ISSUE #9079: Allow DNF types in ReflectionUnionType::getTypes()
Since [DNF types](https://wiki.php.net/rfc/dnf_types) have been introduced in PHP 8.2, types returned by `ReflectionUnionType::getTypes()` include `ReflectionIntersectionType` in addition to `ReflectionNamedType`:

https://3v4l.org/uOIBj

But Psalm is not aware of this:

```php
function test(): (iterable&Throwable)|null
{
    return null;
}

$r = new ReflectionFunction(test(...));
$type = $r->getReturnType();

if ($type instanceof ReflectionUnionType) {
    foreach ($type->getTypes() as $unionType) {
        if ($unionType instanceof ReflectionIntersectionType) {
            
        }
    }
}
```

https://psalm.dev/r/945e0213b2

> ERROR: [DocblockTypeContradiction](https://psalm.dev/155) - 13:13 - Cannot resolve types for $unionType - docblock-defined type ReflectionNamedType does not contain ReflectionIntersectionType

If you point me to how the CallMap hierarchy works (`CallMap.php` vs `CallMap_historical.php` vs `CallMap_82_delta.php`), I can handle it.

BenMorel:
Trying to fix #9079, I looked at stubs as advised in https://github.com/vimeo/psalm/issues/9079#issuecomment-1374617156:

> Reflection classes were added in stubs (which overrides callmap) recently, so you may need to dig into that instead

But the correct type was already there:

```php
    class ReflectionUnionType extends ReflectionType {
        /** @return non-empty-list<ReflectionNamedType|ReflectionIntersectionType> */
        public function getTypes(): array {}
    }
```

https://github.com/vimeo/psalm/blob/5.4.0/stubs/Php82.phpstub

So just in case, I tried to fix the CallMaps as well (this PR), but Psalm still reports the same false positive with the code in the original issue:

> docblock-defined type ReflectionNamedType does not contain ReflectionIntersectionType

Any idea?


orklah:
The issue seems to come from a wrong order when Psalm loads stubs. On my computer, the PHP80 stub is loaded last. We had a discussion about the way those stubs are loaded here: https://github.com/vimeo/psalm/pull/8722#discussion_r1027156087

I think we introduced (or revealed) an issue here.

Could you try to take a look at how those stubs are loaded to see if we can restore the correct order?

/cc @weirdan @Ocramius

PR #9085, opened 2023-01-08
Fix ReflectionClass::newInstanceArgs() with named parameters

### LINKED ISSUE #9077: ReflectionClass::newInstanceArgs() accepts associative arrays
Since PHP 8, `ReflectionClass::newInstanceArgs()` accepts an associative array of parameter names.

This code works (https://3v4l.org/L0kj4):

```php
class X {
    public function __construct(
        string $a,
        string $b,
    ) {
        echo "\$a = $a\n";
        echo "\$b = $b\n";
    }
}

$r = new ReflectionClass(X::class);
$r->newInstanceArgs([
    'b' => 1,
    'a' => 2,
]);
```

But Psalm fails with:

> ERROR: [InvalidArgument](https://psalm.dev/004) - 14:21 - Argument 1 of ReflectionClass::newInstanceArgs expects array<int, mixed>, but array{a: 2, b: 1} provided

https://psalm.dev/r/835445aea1

BenMorel:
Fixes #9077.

A couple questions:

- should the method be declared in `ReflectionObject` as well, or does it automatically inherit from `ReflectionClass`? (it was already declared twice)
- should CallMaps still be maintained, when the method is present in stubs? Shouldn't we remove these methods from CallMaps, and add a CI check to ensure that there are no conflicting definitions?


othercorey:
One benefit of keeping callmaps for methods defined in stubs, is it tell us if something changed in a future version. The annoying part is callmaps do not support inheritance.

othercorey:
@orklah Any issue with this one? Looks ok to me. I skipped this function to avoid any conflicts.

orklah:
Thanks @BenMorel ! Thanks @othercorey for the remainder!

PR #9088, opened 2023-01-09
Internal cleanup: Unify syntax for redis.phpstub

lptn:
A simple PR to unify syntax for stubs: only redis stub uses interface-like function/method sigratures, other stubs uses empty body


orklah:
Thanks!

PR #9089, opened 2023-01-09
Cleanup unsupported extensions message

lptn:
In this PR I'm trying to get rig of annoying "(unsupported extensions: ..." message. The issue is that some of extensions Psalm supports using CallMap/dictionaries,  some of them - using optional stubs that loaded only by condition, and some extensions are present in both sources (e.g. Redis).

I'm not sure we need to make the code more complex just for this `unsupported extensions` message, alternative solution can be just remove this message.


src/Psalm/Config.php
```diff
@@ -612,6 +612,73 @@ class Config
...
 
+    /**
+     * A list of php extensions described in CallMap Psalm files
+     * as opposite to stub files loaded by condition (see stubs/extensions dir).
+     * @see https://www.php.net/manual/en/extensions.membership.php
+     *
+     * @var list<non-empty-string>
+     * @readonly
+     */
+    public $php_extensions_supported_by_psalm_callmaps = [
```
lptn, line 623: probably not a full list, but a good start :)

orklah:
Seems good! Can you fix CS issues remaining?

lptn:
@orklah 
>Can you fix CS issues remaining?

done

orklah:
Still an issue in ProjectAnalyzer :)

lptn:
@orklah 
sorry, done (2)

orklah:
Thanks!

Issue #8951, opened 2022-12-19
Using a @var annotation on promoted property breaks with generic


alex-dev:
With `@var`: https://psalm.dev/r/88aad69808
With `@param`: https://psalm.dev/r/f4a2e8938a

klimick:
Issue fixed in #9092. It should be closed.

PR #9092, opened 2023-01-10
Handle template type on promoted property

vincent4vx:
See #8951 
This PR simply merge class templates with function ones to resolve docblock type.


orklah:
Thanks!

PR #9095, opened 2023-01-11
Additional test and fix when comparing nested templates

### LINKED ISSUE #9094: Incorrect inference of template definition using nested templates in function parameters
#8112, does not fix all type inference issues, in particular when nested templates are used to define function parameters, as illustrated in this case :
https://psalm.dev/r/0c33f892f1

emmanuelGuiton:
Fixes #9094


orklah:
Thanks!

Issue #8958, opened 2022-12-20
Deprecation: Psalm stubs for ext-apcu loaded using legacy way. 


michnovka:
I have lot of these warnings in my psalm output:

> Target PHP version: 8.2 (inferred from current PHP version) Extensions enabled: decimal (unsupported extensions: ctype, curl, gd, iconv, intl, json, mbstring)
> Scanning files...
> Deprecation: Psalm stubs for ext-apcu loaded using legacy way. Instead, please declare ext-apcu as dependency in composer.json or use <enableExtensions> directive in Psalm config.
> Deprecation: Psalm stubs for ext-random loaded using legacy way. Instead, please declare ext-random as dependency in composer.json or use <enableExtensions> directive in Psalm config.
> Deprecation: Psalm stubs for ext-redis loaded using legacy way. Instead, please declare ext-redis as dependency in composer.json or use <enableExtensions> directive in Psalm config.

My project does not use apcu, nor redis. It is not inside psalm.xml nor inside composer.json.

Why are these extensions even loaded?

orklah:
Those are extensions that are loaded in your current PHP installation (we retrieve those through [extension_loaded](https://www.php.net/manual/fr/function.extension-loaded.php): https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Config.php#L2142

Historically, Psalm used to have every extension symbol automatically known. This was having weird effects when users declared those symbol but did not have that specific extension. We're changing things in order to have a better workflow here

We created separate stubs for extensions and we only load them when we detect an extension;
- Either because it is loaded at runtime
- Or because it is in composer.json
- Or because it is in psalm.xml

We wish to deprecate the first possibility, but for better migration, we're warning users that they will have to migrate extensions to one of the other possibility. If you don't wish to enable the extensions we warn about, just ignore the deprecation, we'll disable those when we finish the migration. Otherwise, please add them in composer or in config.

Ref: https://github.com/vimeo/psalm/pull/8885
(ping @lptn )

michnovka:
@orklah maybe we can introduce some psalm.xml config option to suppress these messages? I keep all my requirements correctly included in composer.json. Still, my PHP installation has maybe few extra extensions. But this does not mean I need to load their stubs for my project.

So while I get the reason for the warning, if a user sees it, they decide its OK, adding an option into psalm.xml to suppress them would be welcome. WDYT?

lptn:
[...]

PR #9099, opened 2023-01-12
Automatically load stubs for random ext on PHP 8.2

lptn:
This solves issues raised here: https://github.com/vimeo/psalm/issues/8958#issuecomment-1377604449


orklah:
Thanks!

Issue #9070, opened 2023-01-06
Getting `InvalidDocblock` when using the `non-negative-int` scalar type on a class with namespace


eopnetto:
I'm trying to validate that the index of an array is a non-negative integer, which works fine in the example below:
https://psalm.dev/r/9113171040

However, I'm getting an `InvalidDocblock` error when using `non-negative-int` on a class with a namespace:
https://psalm.dev/r/21478194fc

```php
namespace App;

class Entity {
    /**
    * @psalm-var array<non-negative-int, string>
    */
    private $actions = [];
    
    public function __construct() {
        $this->actions[-1]= 'test'; // I would expect an `InvalidPropertyAssignmentValue` error here, but I get an `InvalidDocblock` instead
    }
}
```

Note: it works if add a backslash before: `\non-negative-int`
https://psalm.dev/r/33769b88c8

I'm assuming this is a bug, right?! Or is it the expected behavior?

orklah:
I answered here:
https://github.com/vimeo/psalm/issues/9069#issuecomment-1374305274

There's a few keywords missing, are you up for a PR? :)

PR #9107, opened 2023-01-13
Add missing `int-range` aliases into the list of psalm reserved words

eopnetto:
This PR fixes the the issue raised on https://github.com/vimeo/psalm/issues/9070.

After these changes, it should be possible to use the `non-negative-int`, `negative-int` and `non-positive-int` scalar types in classes with a namespace.


orklah:
Thanks!

eopnetto:
closes https://github.com/vimeo/psalm/issues/9070

PR #9112, opened 2023-01-14
Allow to specify a custom shepherd endpoint

lptn:
Currently it's possible to specify only a custom domain, so you have to use `hooks/psalm/` route. In this PR I made it more flexible, so a user can specify any URL.

Before this PR, it's possible to specify shepherd host using 2 ways (by priority):
 1. `--shepherd` cli option. Without value it just a flag to send report to http://shepherd.dev. It's also possible to use it with value: `--shepherd=custom.domain`
 2. Using a combination of 2 env vars: `PSALM_SHEPHERD` and `PSALM_SHEPHERD_HOST`. Where `PSALM_SHEPHERD` is a flag, using `PSALM_SHEPHERD_HOST` you can specify a domain (introduced by https://github.com/vimeo/psalm/pull/5512, shipped within [4.9.0](https://github.com/vimeo/psalm/releases/tag/4.9.0))

I think it will be nice to use a single env variable instead 2, as we do for cli option (for consistency). Also, any van names that includes "host" are not valid anymore as we can specify full URL/endpoint.


lptn:
ping @ngmy also. What do you think about these changes?

lptn:
Hey @orklah 

Do you know why circleci almost always fails with `Uncaught InvalidArgumentException: Could not get class storage for sebastianbergmann\objectenumerator\invalidargumentexception`?  Can it be a memory issue? If so - can we use GitHub actions instead?

src/Psalm/Plugin/Shepherd.php
```diff
@@ -140,82 +140,89 @@ public static function getCurlErrorMessage($ch): string
...
-                case 31:
-                    return 'authority and issuer serial number mismatch';
-                case 32:
-                    return 'key usage does not include certificate signing';
-                case 50:
-                    return 'application verification failure';
-            }
-
-            return '';
+        /** @psalm-suppress MixedAssignment */
```
lptn, line 143: this method exist just to minimise possible BC

orklah:
Can we make it so the current property doesn't change value and behaves like before until it's removed? I advised at least one person to change the value here and I wouldn't want to break that if it's not needed

> Do you know why circleci almost always fails

Something is broken, it's not just a memory thing. As far as I can tell, it did not begin on a specific commit, rather at a specific time. Also the code we're analyzing did not change (we analyze a fork of phpunit). My best guess is that a compatibility layer was installed on a dependency of phpunit that makes Psalm think a class has been analyzed when it didn't.

I can't find the time to debug that unfortunately

lptn:
@orklah 
>Can we make it so the current property doesn't change value and behaves like before until it's removed? I advised at least one person to change the value here and I wouldn't want to break that if it's not needed

Yes, we can, but it requires to write high complexity code, but I can do that. Then, for Psalm 6, we need to cleanup this complexity and it's better to do it with this (create a new PR for Psalm 6, but I don't see a branch for it)

orklah:
Seems good to merge, can you rebase to solve the conflict please?

lptn:
@orklah 
sorry, there is some noise on this PR (rebased changes are also tracked here).  I have just created a clean version of this R with the same code: https://github.com/vimeo/psalm/pull/9133

PR #9113, opened 2023-01-15
Clean up more callmap return types

othercorey:
This is just a general pass at cleaning up the ignored callmap return types.

I patched return types in the php docs and left some return types ignored because of incorrect stubs in php.


othercorey:
I seem to have fixed a function used in a test that expects a `mixed` return type. What should I replace this with?

```php
            'nullCoalescePossibleMixed' => [
                'code' => '<?php
                    /**
                     * @psalm-suppress MixedReturnStatement
                     * @psalm-suppress MixedInferredReturnType
                     */
                    function foo() : array {
                        return filter_input_array(INPUT_POST) ?? [];
                    }',
            ],
```

orklah:
Maybe replace that with filter_input, that way we still have the mixed part

othercorey:
> Maybe replace that with filter_input, that way we still have the mixed part

done.

orklah:
Thanks! That's great!

PR #9118, opened 2023-01-16
Fix libxml_get_errors() return type

jack-worman:


orklah:
You'll have to update the last delta file that contained this function too (or if it never changed, update the callmap_historical) . For more explanations: https://psalm.dev/docs/contributing/editing_callmaps/

orklah:
Thanks!

PR #9122, opened 2023-01-17
Fix remaining ignored CallMap return types that can be updated

othercorey:
The remaining functions are either ignored on purpose, have incorrect types in PHP or are just hard to figure out and therefore not sure if should be changed.


orklah:
Seems like a inconsistency is still there according to CI, but it looks great!

othercorey:
> Seems like a inconsistency is still there according to CI, but it looks great!
 
Re-added the xmlrpc functions for this PR. We can drop this entire set of functions for php 8.0 since the extension was unbundled and is abandoned.

orklah:
good to know! A new failure appeared in the meantime :p

othercorey:
> good to know! A new failure appeared in the meantime :p

Should be good. I only run certain checks locally when doing these.

orklah:
Thanks a lot!

PR #9126, opened 2023-01-17
Support constants in traits

### LINKED ISSUE #9123: [PHP8.2] Support constants in traits
Since PHP 8.2 it's possible to use constants in traits, and use them in other classes: https://php.watch/versions/8.2/constants-in-traits.

However this is not yet working in Psalm, see https://psalm.dev/r/d4b2946941

```<?php

trait FooBar {
    const BAZ = 'baz';
}

class Test {
    use FooBar;
}

echo Test::BAZ; // 'bar'
```

This triggers the following error: `ERROR: [UndefinedConstant](https://psalm.dev/020) - 11:6 - Constant Test::BAZ is not defined`

jack-worman:
Fixes #9123

### DIFF
--- a/src/Psalm/Storage/ClassLikeStorage.php
+++ b/src/Psalm/Storage/ClassLikeStorage.php
@@ -4,6 +4,7 @@ namespace Psalm\Storage;
 
 use Psalm\Aliases;
 use Psalm\CodeLocation;
+use Psalm\Internal\Analyzer\ClassLikeAnalyzer;
 use Psalm\Internal\MethodIdentifier;
 use Psalm\Internal\Type\TypeAlias\ClassTypeAlias;
 use Psalm\Issue\CodeIssue;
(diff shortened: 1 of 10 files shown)


src/Psalm/Storage/ClassLikeStorage.php
```diff
@@ -183,13 +183,18 @@ final class ClassLikeStorage implements HasAttributesInterface
...
+    public array $trait_alias_map_cased = [];
+
     /**
      * @var array<lowercase-string, bool>
      */
     public $trait_final_map = [];
 
     /**
-     * @var array<string, int>
+     * @var array<string, 1|2|3>
```
orklah, line 197: ClassLikeAnalyzer::VISIBILITY_* would fit better here

weirdan:
Constants are not supported in traits before PHP 8.2, but I don't see any check for the target PHP version here.

src/Psalm/Storage/ClassLikeStorage.php
jack-worman: Psalm says that is an invalid doc-block, so I resorted to this instead

jack-worman:
> Constants are not supported in traits before PHP 8.2, but I don't see any check for the target PHP version here.

Added a check and a new issue type

orklah:
Thanks!

PR #9130, opened 2023-01-18
Report unused baseline entries

### LINKED ISSUE #9102: [Feature request] Report unmatched baseline rules
It would be great IMO to force contributors to show the positive effect of their changes by erroring out if there are unused baseline elements or [issue handlers](https://github.com/vimeo/psalm/issues/3262):
1. It validates further that what they are doing is right.
2. It allows the baseline to be kept in sync avoiding people to get confused when `--update-baseline` results in changes unrelated to their own changes.

There already is `--find-unused-psalm-suppress` for `@psalm-suppress`, but there does not seem to be something for other ignore mechanisms such as the baseline or issue handlers.

jack-worman:
Closes #9102


jack-worman:
> Oh wow thanks!!!

I have been wanting this for awhile too!

PR #9134, opened 2023-01-18
Enhance error output for shepherd

lptn:
I just tested shepherd with invalid URL and found a small issue in output, probably caused by parallel writing to STDERR.

Note lines 98 and 105 - they should be one right after another

image


orklah:
Thanks!

PR #9135, opened 2023-01-18
Unify warning messages for deprecated config variables

lptn:
Improve output:
 1. Unify deprecation messages style
 2. Do not display "Enabled extensions: " for empty ext list

new output:
image

before the fix:
image


lptn:
I've also considered to add a new `deprecation` method to Progress, but Progress instance is not available when we display "Warning: "findUnusedCode" will be defaulted to "true" in Psalm 6. You should explicitly..." message.

weirdan:
> but Progress instance is not available when we display "Warning: "findUnusedCode"

Can we delay the output until `Progress` is available? I'd very much like to be able to suppress deprecation messages with `--no-progress` CLI switch, as that would also fix #9034

lptn:
@weirdan 

I think we can, but I don't see a clean code solution ATM. We probably need to make Progress a singleton (what is fast and dirty) or use DI (this is BC change and probably a lot of work).

weirdan:
We could store a list of config warnings on the config instance itself, and then report it after Progress has been constructed.

lptn:
@weirdan 
>We could store a list of config warnings on the config instance itself, and then report it after Progress has been constructed.

Right, we can use it, but on other hand `Config` created for other reasons. In any case, I think this is out of the scope of this PR. There is another issue where I found we can improve output: https://github.com/vimeo/psalm/pull/9159 . Let's collect more requirements for a new output and then develop it in a separate PR (probably for Psalm major release)

src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
```diff
@@ -564,8 +564,8 @@ private function generatePHPVersionMessage(): string
             $source,
         );
 
-        if (count($codebase->config->php_extensions) > 0) {
-            $enabled_extensions_names = array_keys(array_filter($codebase->config->php_extensions));
+        $enabled_extensions_names = array_keys(array_filter($codebase->config->php_extensions));
+        if (count($enabled_extensions_names) > 0) {
```
lptn, line 568: Fixes empty extensions list output like:
```
Target PHP version: 7.4 (inferred from composer.json) Enabled extensions: .
```

new output:
```
Target PHP version: 7.4 (inferred from composer.json).
```

orklah:
Seems to have a CI error left, otherwise I'm happy to merge :)

lptn:
>Seems to have a CI error left, otherwise I'm happy to merge :)


I see now, sorry, I thought this error is not related to my changes :) . Fixed

orklah:
That created a code style failure :p

lptn:
done :)

How about automatically fixing CS issues?

orklah:
I'd be fine with adding that :)

PR #9156, opened 2023-01-20
Indent docblocks in DocComment test

weirdan:
Since heredocs in PHP 7.3+ automatically strip leading whitespace we can
fix the formatting for this test.


orklah:
Thanks!

PR #9157, opened 2023-01-21
Narrow ->value of enum case(s) to only the possible values

### LINKED ISSUE #8568: "Hard coded" values defined using a backed enum are wider than they could be
https://psalm.dev/r/ce8f48f2b1

```php
enum Allowed: string {
    case A = 'A';
    case B = 'B';
}

// This is hard-coded so should always be 'A'
$alwaysA = Allowed::A->value;
```

In this example, the value `$alwaysA` could be `'A'|'B'`, which is technically correct, but since it is defined literally as `Allowed::A->value`, the actual type is really just `'A'`.

I'm not sure if it's possible to narrow the type of `$alwaysA` here, since it's hard-coded?

Note, that using the string literal assignment, e.g. `$alwaysA = 'A';` instead of the backed enum value, will eliminate the issue, so is a potential workaround.

### LINKED ISSUE #8812: [5.0] false positive ParadoxicalCondition with match() and enum
https://psalm.dev/r/74b7b0ba5a

Code:
```php
enum SomeType: string
{
    case FOO = 'FOO';
    case BAR = 'BAR';
}

function getSomething(string $moduleString): int
{
    return match ($moduleString) {
        SomeType::FOO->value => 1,
        SomeType::BAR->value => 2,
    };
}
```

Output:
```
ERROR: [ParadoxicalCondition](https://psalm.dev/089) - 13:9 - Condition ($moduleString is string(FOO)) contradicts a previously-established condition ($moduleString is not string(FOO))
```

This did not happen with Psalm 4.x

annervisser:
Using $stmt_var_type to determine if we're dealing with a subset of the enum cases. If so, we only consider those cases for the possible values.

Fixes #8568
Fixes #8812

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/AtomicPropertyFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/AtomicPropertyFetchAnalyzer.php
@@ -62,13 +62,17 @@ use Psalm\Type\Atomic\TString;
 use Psalm\Type\Atomic\TTemplateParam;
 use Psalm\Type\Union;
 
+use function array_filter;
 use function array_keys;
+use function array_map;
 use function array_search;
 use function count;
 use function in_array;
 use function is_int;
 use function is_string;
 use function strtolower;
+
+use const ARRAY_FILTER_USE_KEY;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


annervisser:
`ci/circleci: test-with-real-projects` for PHPUnit seems to be broken on multiple recent pull requests, so I'm assuming that to be unrelated

orklah:
Yeah, that's unrelated :)

Thanks!

PR #9159, opened 2023-01-22
PoC: Add an example of command to output Psalm config

lptn:
This is a prove of concept of a command to output "compiled" Psalm config. Output example:
image

## Goals
 - simplify Psalm issue debugging
 - simplify debugging when develop psalm plugins

## Expected feedback
 1. do we need such functionality in the core?
 2. any ideas about the output formatting? 
 3. any ideas what info to output?

### DIFF
--- a/src/Psalm/Internal/Cli/Commands/DisplayConfig.php
+++ b/src/Psalm/Internal/Cli/Commands/DisplayConfig.php
@@ -1,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Internal\Cli\Commands;
+
+use Psalm\Config;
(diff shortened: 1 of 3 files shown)


lptn:
@weirdan 
@orklah 
I would love to hear some feedback from you guys

weirdan:
Looks nice.

> do we need such functionality in the core?

Personally, I don't believe I would need this. This may even give people an idea that this should be included in bug reports, but most of the time it would be superfluous.

> any ideas about the output formatting?

If we're looking for human-friendly output, the one you have here is good.

> any ideas what info to output?

Assuming we do integrate this, `issueHandler` section could potentially be useful.

lptn:
@weirdan 

>Personally, I don't believe I would need this.


ok, thanks, closing for now

Issue #9148, opened 2023-01-20
static closures with arrow functions  and normal functions are not detected


torian257x:
```
class MyClass
{


    private $mything = ['gaga'];
    /** @test */
    public function mytest(){
        $rv = array_map(static function($record ) { return $this->mything;}, ['a', 'b']);
    }
```

```
class MyClass
{
    private $mything = ['gaga'];
public function mytest(){
   $rv = array_map(static fn($record ) => $this->mything, ['a', 'b']);
}
```

both is wrong php code since you have no access to `$this` inside of static closures (static closures are, supposedly, a bit faster https://github.com/Ocramius/GeneratedHydrator/releases/tag/3.0.0 )

https://psalm.dev/r/6414bd28cf

Issue #2813, opened 2020-02-13
Check $this in static functions


vudaltsov:
Using `$this` in a `static` closure should  add an error.
https://psalm.dev/r/670e9410f3

zerkms:
I discovered a different but relevant use-case: https://psalm.dev/r/748ec7f1c9

PR #9164, opened 2023-01-23
Check `$this` in static closures

weirdan:
Fixes vimeo/psalm#9148
Fixes vimeo/psalm#2813

### DIFF
--- a/src/Psalm/Internal/LanguageServer/Client/TextDocument.php
+++ b/src/Psalm/Internal/LanguageServer/Client/TextDocument.php
@@ -58,7 +58,7 @@ class TextDocument
             /**
              * @return Generator<int, Promise<object>, object, TextDocumentItem>
              */
-            static function () use ($textDocument) {
+            function () use ($textDocument) {
                 /** @var Promise<object> */
                 $promise = $this->handler->request(
                     'textDocument/xcontent',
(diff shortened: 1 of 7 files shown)


src/Psalm/Internal/LanguageServer/Client/TextDocument.php
```diff
@@ -58,7 +58,7 @@ public function xcontent(TextDocumentIdentifier $textDocument): Promise
             /**
              * @return Generator<int, Promise<object>, object, TextDocumentItem>
              */
-            static function () use ($textDocument) {
+            function () use ($textDocument) {
```
weirdan, line 61: @tm1000 it looks like a number of closures used by the language server were invalid. This means the LS behaviour may change when this PR is merged, so I'd like to request your review here.

tm1000: @weirdan thanks for the heads up. No idea why these were static I don't think it'll affect anything too bad. Once you merge this I'll update my WIP pr and see how it runs for a bit

tm1000 (approved):

src/Psalm/Internal/LanguageServer/Client/TextDocument.php
weirdan: It looks like all of those problematic closures were used in the code path handling `textDocument/xcontent` request. `textDocument/xcontent` is not a part of a standard protocol, it's a Sourcegraph extension, so chances are VSCode never used it.

tm1000: xcontent is also removed in my pr. It existed when Felix was adding features that Microsoft hadn't yet. 

Those features are now in Microsoft's lsp spec.

PR #9170, opened 2023-01-23
Added @psalm-pure to filter_var()

mmcev106:
I believe this is correct.  I struggled to come up with a unit test for it (suggestions welcome).  The only side effect I encountered was an odd taint trace like the following (though a taint was still correctly detected).

https://psalm.dev/r/e019ea8049


orklah:
what would happen if we were to push a Stringable object in filter_var with a string filter? Would __toString would be called within filter_var?

mmcev106:
@orklah,  great question.  You're correct.  The following example does indeed execute `__toString()`:

```php
class A{
    function __toString(){
        return "foo\n";
    }
}

echo filter_var(new A());
```

This is true for many other methods already in `CoreGenericFunctions.phpstub` as well (htmlspecialchars(), strip_tags(), etc.).  Does this mean `@psalm-pure` should be removed from all of them, and perhaps `@psalm-taint-specialize` added it its place?

orklah:
I'm not sure. A lot of function calls __toString as a result of coalescing the object into a string just before the call. What make me think of that is that filter_var accepts mixed, not string, so it's likely the __toString call is made inside of the function.

I'm not sure how big the difference is or how Psalm would handle that though...

mmcev106:
Adding `@psalm-pure` is consistent with other functions in `CoreGenericFunctions.phpstub` (htmlspecialchars(), strip_tags(), etc.).  However, it seems like there is a conflict between `@psalm-pure` and `@psalm-taint-escape` that is causing test failures.  I have yet to determine exactly what is going on.

mmcev106:
@orklah, I believe I’ve found & fixed the place where `removed_taints` were getting lost.  The tests also pointed out that `filter_var()` for ints, floats, & booleans should also escape all other taint types (not just `html`).  The behavior should now be the same as casting to int/float/boolean types.  Don’t hesitate to ask if you have any questions/concerns.

orklah:
Thanks!

Issue #9042, opened 2023-01-03
DateTimeImutable::modify() should return `static`


pilif:
DateTime type issues strike back...

https://psalm.dev/r/09e238bf00

I expect `$foo2` to be of type `Subclass` and indeed, according to PHP it is:

https://3v4l.org/j9ZSo

I agree that the PHP manual [disagrees](https://www.php.net/manual/en/datetimeimmutable.modify.php), so that should probably be raised on the php internals list too, but seeing how big a BC break it would be for PHP to change this behavior, I assume that this is a doc bug, not a language issue.

pilif:
if you agree that Psalm should probably represent what PHP currently does, I'll try to submit a PR to fix the return type, though I'm a bit reluctant to start working on this right now as there were so many DateTime related issues and PRs raised in the past which kind of scares me to work on those parts at this point 😛

weirdan:
Actual behaviour is the ground truth, so PR is welcome.

DaDeather:
Didn't this break up a few things?

If you now call the given snippet it causes a `ERROR: [UndefinedClass](https://psalm.dev/019) - 10:6 - Class or interface <empty string> does not exist` which is given on any method call after a modify being made.

PR #9172, opened 2023-01-24
fix (DateTime|DateTimeImmutable)::modify() return types

DaDeather:
vimeo#9042 caused issues to any called method on an instance of DateTime|DateTimeImmutable after calling the modify method.

This fixes vimeo#9171

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/DateTimeModifyReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/DateTimeModifyReturnTypeProvider.php
@@ -8,7 +8,6 @@ use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
 use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
 use Psalm\Type;
 use Psalm\Type\Atomic\TLiteralString;
-use Psalm\Type\Atomic\TNamedObject;
 use Psalm\Type\Union;
 
 /**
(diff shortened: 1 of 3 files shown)


pilif:
This will cause #9042 to be re-opened though. ˋmodify()ˋ returns something of the same type that it was called on which is consistent to the ˋstaticˋ return type

DaDeather:
Do you got any better idea how to solve this correctly? TBH i'm not really into the codebase of psalm yet. Which is why this is the only thing I came up with up till now :shrug:.

DaDeather:
FYI: another approach I found would be to set:
`$event->getContext()->self = $event->getCalledFqClasslikeName() ?? $event->getFqClasslikeName();`
right before the return.

This would also fix this issue. But as I said before I'm not well into the ecosystem of psalm :shrug:.

DaDeather:
Is there anything I can do to make this one happen soon?

weirdan:
I see where the confusion is. Signature *return* types and return *value* types, while being related concepts, are not the same thing. `static` is a relative type which is only valid in the context of the original class. When it's leaked outside, it may make Psalm assume funny things happen, like this: https://psalm.dev/r/c23709c02e

In the snippet above, Psalm assumes you're trying to call `Unrelated::format()` because `static` is interpreted in the context of `class Unrelated`. 

@pilif with this PR, your snippet reports `$foo2` from #9042 to be `Subclass`, as it should. In fact, it's included in a test.

weirdan:
@DaDeather would you mind rebasing this on top of master? I think it should clear up BCC failures.

weirdan:
Thanks!

PR #9173, opened 2023-01-24
Resolve class constants within PhpStorm metadata

shvlv:
The PHPStorm allows using class constants for the `override` method - https://www.jetbrains.com/help/phpstorm/ide-advanced-metadata.html#passing-a-class-constant. The Psalm supports only strings for now.

The current PR replicates PHPStorm functionality. i.e. for PHPStorm metadata
```php
namespace PHPSTORM_META {
    override(
        \Psr\Container\ContainerInterface::get(0),
        map([
            \Ns\MyClass::PROPERTIES => \Ns\MyClass\Properties::class,
            '' => '@',
        ])
    );
}
```

the following code will be parsed and analyzed correctly
```php
$properties = $container->get(\Ns\MyClass::PROPERTIES); // $properties instance of  \Ns\MyClass\Properties::class
```

In my tests, the PHPStorm functionality is a bit broken because the IDE doesn't recognize ` $container->get(self::PROPERTIES);` or `$container->get('properties');`, FQCN only (or vice versa if the `override` definition made with a string literal the PHPStorm expects exactly string literal). But I think it should be considered a bug so the current PR resolves every class constant to string so every combination works.

### DIFF
--- a/src/Psalm/Internal/Scanner/PhpStormMetaScanner.php
+++ b/src/Psalm/Internal/Scanner/PhpStormMetaScanner.php
@@ -76,7 +76,7 @@ class PhpStormMetaScanner
                     $constant_type = $codebase->classlikes->getClassConstantType(
                         $resolved_name,
                         $array_item->key->name->name,
-                        ReflectionProperty::IS_PUBLIC,
+                        ReflectionProperty::IS_PRIVATE,
                     );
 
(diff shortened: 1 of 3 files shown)


weirdan (approved):

shvlv:
Hi, @weirdan I fixed Psalm errors. Let me know if something else should be improved.

weirdan (approved):

weirdan:
Thanks!

PR #9177, opened 2023-01-25
Detect duplicate keys in array shapes

danog:
Valinor throws an exception on those, and it makes sense to statically detect them.


orklah:
Thanks!

Issue #9038, opened 2023-01-01
DocblockTypeContradiction error creates incorrect type and wrong error


kkmuffme:
https://psalm.dev/r/abc33fab15

The first trace should be int - not int|float

PR #9209, opened 2023-02-02
Remove some logic that didn't need to be there

muglug:
Fixes https://github.com/vimeo/psalm/issues/9038


muglug:
I need to do more a bit more work on this — I've added a newly-failing test

PR #9212, opened 2023-02-02
Fix xmlrpc callmap functions that were dropped in PHP 8.0

othercorey:
This extension was unbundled and now reverts to an abandoned version of the extension.


orklah:
Thanks!

Issue #9147, opened 2023-01-20
PHP 8.2 - support DatePeriod::INCLUDE_END_DATE


wooky:
PHP 8.2, the constant DatePeriod::INCLUDE_END_DATE (literal value 2) was introduced, and can be passed in the 4th parameter of the DatePeriod constructor. Psalm does not allow anything except 0 or 1 to be passed in as the 4th parameter, as seen in this snippet: https://psalm.dev/r/8760140a05

PR #9220, opened 2023-02-03
Add DatePeriod::INCLUDE_END_DATE option

othercorey:
closes https://github.com/vimeo/psalm/issues/9147

I'm not sure how the anonymous namespace in php82 stubs works, but added DatePeriod to match other classes.

### DIFF
--- a/stubs/Php82.phpstub
+++ b/stubs/Php82.phpstub
@@ -10,4 +10,27 @@ namespace {
         /** @return non-empty-list<ReflectionNamedType|ReflectionIntersectionType> */
         public function getTypes(): array {}
     }
+
+    /**
+    * @psalm-immutable
+    *
+    * @template-covariant Start of string|DateTimeInterface
+    * @implements IteratorAggregate<int, DateTimeInterface>
+    */
+    class DatePeriod implements IteratorAggregate
+    {
+        const EXCLUDE_START_DATE = 1;
+        const INCLUDE_END_DATE = 2;
+
+        /**
+        * @param Start $start
+        * @param (Start is string ? int-mask<self::EXCLUDE_START_DATE, self::INCLUDE_END_DATE> : DateInterval) $interval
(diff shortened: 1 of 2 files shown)


stubs/Php82.phpstub
```diff
@@ -10,4 +10,27 @@ namespace {
...
+    */
+    class DatePeriod implements IteratorAggregate
+    {
+        const EXCLUDE_START_DATE = 1;
+        const INCLUDE_END_DATE = 2;
+
+        /**
+        * @param Start $start
+        * @param (Start is string ? int-mask<self::EXCLUDE_START_DATE, self::INCLUDE_END_DATE> : DateInterval) $interval
+        * @param (Start is string ? never : DateTimeInterface|positive-int) $end
```
weirdan, line 28:
```suggestion
        * @param (Start is string ? never : (DateTimeInterface|positive-int)) $end
```

stubs/Php82.phpstub
```diff
@@ -10,4 +10,27 @@ namespace {
...
+        * @param (Start is string ? never : int-mask<self::EXCLUDE_START_DATE, self::INCLUDE_END_DATE>) $options
+        */
+        public function __construct($start, $interval = 0, $end = 1, $options = 0) {}
+
+        /** @psalm-return (Start is string ? (Traversable<int, DateTime>&Iterator) : (Traversable<int, Start>&Iterator)) */
```
weirdan, line 33: Wouldn't `(Start is string ? Iterator<int, DateTime> : Iterator<int, Start>)` be easier to read here?

othercorey: Up to you. Everything but the new constant was copied from the PHP80 stub. We can change both places if you want.

weirdan: Yes, please do.

othercorey: done.

weirdan (approved):

Issue #3358, opened 2020-05-14
DateInterval::format() considered "possibly-mutating"


sebastianbergmann:
```php
<?php declare(strict_types=1);
/**
 * @psalm-immutable
 */
final class C
{
	private DateTimeImmutable $a;
    
    private DateTimeImmutable $b;
    
    public function __construct(DateTimeImmutable $a, DateTimeImmutable $b)
    {
        $this->a = $a;
        $this->b = $b;
    }
    
    public function numberOfDays(): int
    {
        return (int) $this->b->diff($this->a)->format('%a') + 1;
    }
}
```
https://psalm.dev/r/332288b1ea
```
Psalm output (using commit 48504aa): 

ERROR: ImpureMethodCall - 19:48 - Cannot call an possibly-mutating method DateInterval::format from a mutation-free context
```
* `$this->a` is immutable
* `$this->b` is immutable
* `$this->b->diff($this->a)` returns a `DateInterval` instance
* The creation of a `DateInterval` instance does not have a side effect
* `DateInterval::format()` generates a string representation of a `DateInterval` instance

gharlan:
some more methods/functions: https://psalm.dev/r/373cc81b14
I think they should be all marked as mutation-free.

gharlan:
Hmm, I think my code above is not correct. The function should be mutation-free instead of pure:
https://psalm.dev/r/0d9ac83849

But the issue is the same, these methods/functions should be mutation-free.
I would like to help. But where is the correct place to mark core functions/methods as mutation-free?

weirdan:
Seems to be fixed.

weirdan:
https://psalm.dev/r/332288b1ea

PR #9225, opened 2023-02-06
Add DateTime stub

othercorey:
closes https://github.com/vimeo/psalm/issues/3358

This is a copy of DateTimeImmutable with the psalm-mutation-free tags removed where appropriate.

Moved the attributes the top of the file for better organization

### DIFF
--- a/stubs/CoreGenericClasses.phpstub
+++ b/stubs/CoreGenericClasses.phpstub
@@ -1,5 +1,23 @@
 <?php
 
+#[Attribute(Attribute::TARGET_METHOD)]
+final class ReturnTypeWillChange
+{
+    public function __construct() {}
+}
+
+#[Attribute(Attribute::TARGET_PARAMETER)]
+final class SensitiveParameter
+{
+    public function __construct() {}
+}
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final class AllowDynamicProperties
+{
+    public function __construct() {}
+}
+
 /**
  * Interface to detect if a class is traversable using &foreach;.
  * @link http://php.net/manual/en/class.traversable.php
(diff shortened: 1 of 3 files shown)


othercorey:
I'm not sure I follow what this failing method call test is trying to check:

```php
'datetimeformatNotFalse' => [
    'code' => '<?php
        $format = random_bytes(10);
        $dt = new DateTime;
        $formatted = $dt->format($format);
        if (false !== $formatted) {}
        takesString($formatted);',
    'assertions' => [],
    'ignored_issues' => [],
    'php_version' =>  '7.4',
],
```

weirdan:
> I'm not sure I follow what this failing method call test is trying to check

Apparently, it's a test ensuring there's `@psalm-ignore-falsable-return` on that method.

`if (false !== $formatted) {}` ensures that checks against `false` are allowed and not considered redundant.

`takesString($formatted);` validates that checks for `false` are not required.

othercorey:
Ok, that's what I thought, but since DateTime::format() was just in the callmaps before, wasn't sure if it was specific to that method since all callmap falsable returns should be ignored. 

Maybe we can just use a different method?

othercorey:
> Maybe we can just use a different method?

I changed it to use another method that returns false in the callmap. I had to use one that isn't defined in stubs.

These tests aren't labeled very well, but I think it was meant to test the ignore return false of ANY callmap method (not just a function) and isn't specific to Date Time.

othercorey:
@weirdan I now don't understand this failure:

```php
        $this->expectExceptionMessage('LessSpecificImplementedReturnType - somefile.php:5:');

        $this->addFile(
            'somefile.php',
            '<?php

                /**
                 * @method int test()
                 * @method \DateTime modify($modify)
                 */
                class WTF extends \DateTime { }',
        );
```

Why does `static` behave differently from a stub than the callmap? The return type is the same.

PR #9226, opened 2023-02-06
Use "numeric-string" type for some MySQLi properties

phansys:
>If the number of rows is greater than PHP_INT_MAX, the number will be returned as a string.

* https://www.php.net/manual/en/mysqli-result.num-rows.php
* https://www.php.net/manual/en/mysqli-stmt.num-rows.php

>If the number of affected rows is greater than maximum PHP int value, the number of affected rows will be returned as a string value.

* https://www.php.net/manual/en/mysqli.affected-rows.php
* https://www.php.net/manual/en/mysqli-stmt.affected-rows.php


orklah:
Thanks for this!

CI is failing because the signature in Callmap.php should always match with the most recent signature in a delta file (for 8.2, 8.1, 8.0... up until Callmap_historical for signatures that represents any version before 7.1)

There's a full explanation here if you need it: https://psalm.dev/docs/contributing/editing_callmaps/

phansys:
Thank you for the hint!
I found no deltas for the changed methods between the supported PHP versions, so I only updated `CallMap_historical.php`.

orklah:
Thanks!

PR #9237, opened 2023-02-07
Always enable JIT

danog:
This allows for 20%+ performance improvements in our codebase.

I wonder how much more can be squeezed out of opcache if we also enable preloading (should reduce lock contention when initially opcaching files in multithreaded mode)

PR #9238, opened 2023-02-07
Add ext-zend-opcache as composer requirement

danog:
Ref https://github.com/vimeo/psalm/pull/9237, it's worth it, given the noticeable (20%+) performance improvements.


danog:
Actually nevermind that, psalm works just fine even without the opcache extension installed, the only side effect is a warning.

PR #9243, opened 2023-02-08
Do not override DOMElement::$attributes in stubs

### LINKED ISSUE #9242: DOMElement::$attributes is declared not null, PHP documents otherwise
Reproducer: https://psalm.dev/r/10d681a370

Event the stubs report that property as non-null: https://github.com/vimeo/psalm/blob/4eacb2f234867d34d8489f904c2445aecb9158d0/stubs/extensions/dom.phpstub#L752-L758

But PHP docs (and PHPStorm stubs) report the opposite: https://www.php.net/manual/en/class.domnode.php#domnode.props.attributes

> A [DOMNamedNodeMap](https://www.php.net/manual/en/class.domnamednodemap.php) containing the attributes of this node (if it is a [DOMElement](https://www.php.net/manual/en/class.domelement.php)) or null otherwise.

Jean85:
Fixes #9242

### DIFF
--- a/stubs/extensions/dom.phpstub
+++ b/stubs/extensions/dom.phpstub
@@ -750,14 +750,6 @@ class DOMElement extends DOMNode implements DOMParentNode, DOMChildNode
     public ?DOMElement $nextElementSibling;
 
     /**
-     * Inherited from DOMNode, but always non-null.
-     *
-     * @readonly
-     * @var DOMNamedNodeMap<DOMAttr>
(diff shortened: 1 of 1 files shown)


Jean85:
Should I add any regression test? How can I do that?

Jean85:
This is wrong, closing in favor of https://github.com/JetBrains/phpstorm-stubs/pull/1527

Issue #9263, opened 2023-02-11
 Cannot load Zend OPcache - it was already loaded


weirdan:
https://github.com/psalm/codeception-psalm-module/actions/runs/4149651001/jobs/7178758106#step:10:143

I'm seeing this locally too.

/cc: @danog

PR #9240, opened 2023-02-07
Emit warning if opcache cannot be enabled

danog:

PR #9265, opened 2023-02-11
Fix opcache being loaded twice

weirdan:
Fixes vimeo/psalm#9263

### DIFF
--- a/src/Psalm/Internal/Fork/PsalmRestarter.php
+++ b/src/Psalm/Internal/Fork/PsalmRestarter.php
@@ -14,6 +14,8 @@ use function in_array;
 use function ini_get;
 use function preg_replace;
 
+use const PHP_VERSION_ID;
+
 /**
  * @internal
  */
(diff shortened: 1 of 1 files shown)


src/Psalm/Internal/Fork/PsalmRestarter.php
```diff
@@ -74,16 +74,25 @@ protected function restart(array $command): void
 
             file_put_contents($this->tmpIni, $content);
         }
+
+        $additional_options = [];
+
+        // executed in the parent process (before restart)
+        // if it wasn't loaded then we apparently don't have opcache installed and there's no point trying
+        // to tweak it
+        if (extension_loaded('opcache') || extension_loaded('Zend OPcache')) {
```
weirdan, line 83: `Zend OPcache` is what the extension called on my system:

```console
$ php -r 'var_dump(["opcache" => extension_loaded("opcache"), "Zend OPcache" => extension_loaded("Zend OPcache")]);'
array(2) {
  'opcache' =>
  bool(false)
  'Zend OPcache' =>
  bool(true)
}
```

danog:
BTW, my PR @ https://github.com/vimeo/psalm/pull/9240 also fixes the issue, and imo loading opcache even on 7.4 should improve performance.

src/Psalm/Internal/Fork/PsalmRestarter.php
danog: I think using the `opcache_get_status` function is more reliable, as I did in https://github.com/vimeo/psalm/pull/9240/files

weirdan:
> imo loading opcache even on 7.4 should improve performance.

My stance is that the decision to load the extension or not should be left to the user (with an appropriate suppressible suggestion, as in your PR, when it's not loaded). There could be various reasons why it's not available.

src/Psalm/Internal/Fork/PsalmRestarter.php
weirdan: Actually it may be less reliable as it can throw warnings depending on the configuration: https://www.php.net/opcache_get_status#refsect1-function.opcache-get-status-errors

PR #9266, opened 2023-02-11
Drop abandoned mcve m_ functions

othercorey:
All the `m_` functions appear to be part of a MCVE credit card processing extension that was dropped after PHP 5 and is abandoned.


weirdan (approved):

weirdan:
Thanks!

PR #9269, opened 2023-02-11
Drop removed xsl callmap functions

othercorey:
These functions no longer exist. Users should be using [XSLTProcessor](https://www.php.net/manual/en/class.xsltprocessor.php).


weirdan:
I think this one was removed in PHP 5.0

weirdan:
Thanks!

Issue #7722, opened 2022-02-22
Forbid implementation of UnitEnum & BackedEnum by classes


weirdan:
https://psalm.dev/r/b7187df53e
https://3v4l.org/iL7gQ

**Expected**: Psalm should warn about implementation of `Z` by `C`

AndrolGenhald:
Also forbid implementing methods from `UnitEnum` and `BackedEnum`
https://3v4l.org/k5i42
https://psalm.dev/r/d0aece15a6

[This](https://3v4l.org/2bfve) is technically allowed since it's a `UnitEnum` and not a `BackedEnum`.

PR #9272, opened 2023-02-12
Forbid implementing some interfaces

weirdan:
- `Throwable` can only be implemented when classes extend one of
  `Exception` or `Error`
- `UnitEnum` and `BackedEnum` cannot be implemented by user-defined
  classes

Refs vimeo/psalm#7722


orklah:
Thanks!

PR #9273, opened 2023-02-12
Update callmap types for php 8.2

othercorey:
This updates the types, that already pass tests, for PHP 8.2. Most of the changes are using `true` instead of `bool` for functions that always returned true.


orklah:
Thanks!

PR #9278, opened 2023-02-12
Fix remaining callmap functions

othercorey:
This fixes all of the non-method callmap entries that I could. The remaining functions are either extensions that don't reflect properly or php functions that have special issues.

I'm sure I missed something so it might not be "everything" fixable, but it's most.

Checked with php 8.0 and 8.2.


tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -349,20 +349,6 @@ class InternalCallMapHandlerTest extends TestCase
         'intltz_get_display_name',
         'iteratoriterator::__construct',
         'jsonexception::__construct',
-        'ldap_compare' => ['8.0'],
```
othercorey, line 352: Many of these had changes from php 7.2 and 8.0 combined into a single php 8.1 delta for some reason.

orklah:
Awesome! Seems like there's some error in the callmap according to CI

othercorey:
> Awesome! Seems like there's some error in the callmap according to CI

Fixed.

orklah:
Thanks!

Issue #9258, opened 2023-02-10
Uncaught InvalidArgumentException: Could not get class storage for dokuwiki_auth_plugin


kamil-tekiela:
I don't know if it can be analyzed from stack trace. I can't run the debugger now but if this is not enough I can debug it later. 

```
Uncaught InvalidArgumentException: Could not get class storage for dokuwiki_auth_plugin in C:\Users\tekie\AppData\Roaming\Composer\vendor\vimeo\psalm\src\Psalm\Internal\Provider\ClassLikeStorageProvider.php:45
Stack trace:
#0 C:\Users\tekie\AppData\Roaming\Composer\vendor\vimeo\psalm\src\Psalm\Internal\Analyzer\ClassAnalyzer.php(783): Psalm\Internal\Provider\ClassLikeStorageProvider->get('DokuWiki_Auth_P...')
#1 C:\Users\tekie\AppData\Roaming\Composer\vendor\vimeo\psalm\src\Psalm\Internal\Analyzer\ClassAnalyzer.php(416): Psalm\Internal\Analyzer\ClassAnalyzer::addContextProperties(Object(Psalm\Internal\Analyzer\ClassAnalyzer), Object(Psalm\Storage\ClassLikeStorage), Object(Psalm\Context), 'testable_auth_p...', 'auth_plugin_aut...', Array)
#2 C:\Users\tekie\AppData\Roaming\Composer\vendor\vimeo\psalm\src\Psalm\Internal\Analyzer\FileAnalyzer.php(203): Psalm\Internal\Analyzer\ClassAnalyzer->analyze(Object(Psalm\Context), Object(Psalm\Context))
#3 C:\Users\tekie\AppData\Roaming\Composer\vendor\vimeo\psalm\src\Psalm\Internal\Codebase\Analyzer.php(1581): Psalm\Internal\Analyzer\FileAnalyzer->analyze()
#4 C:\Users\tekie\AppData\Roaming\Composer\vendor\vimeo\psalm\src\Psalm\Internal\Codebase\Analyzer.php(522): Psalm\Internal\Codebase\Analyzer->analysisWorker(608, 'D:\\projects\\dok...')
#5 C:\Users\tekie\AppData\Roaming\Composer\vendor\vimeo\psalm\src\Psalm\Internal\Codebase\Analyzer.php(272): Psalm\Internal\Codebase\Analyzer->doAnalysis(Object(Psalm\Internal\Analyzer\ProjectAnalyzer), 1)
[...]

PR #9283, opened 2023-02-13
Add curl stubs to prevent crashes

kamil-tekiela:
This is related to https://github.com/vimeo/psalm/issues/9258 but not the fix.


weirdan:
Shouldn't we add private final constructors to those classes to prevent direct instantiations?

weirdan (approved):

weirdan:
Thanks!

Issue #4092, opened 2020-08-31
TypeDoesNotContainType with is_a (strict type)


andreev1024:
Hi. Please, check my examples:
https://psalm.dev/r/0a1bc59985
https://psalm.dev/r/c552b7bfe4
https://psalm.dev/r/957ebfd724
https://psalm.dev/r/957ebfd724 + custom psalm stub = also error
```php
/**
 * @param mixed $object
 * @param string $class_name
 * @param bool $allow_string
 *
 * @return bool
 */
function is_a($object, $class_name, $allow_string = false)
{
}
```

Why does it work like this? I had no issues. Then I added stub for `is_a` and ran the analyzer. Psalm showed me a new issue. Yeah, I made a mistake - forgot the third argument. Thus, Psalm helped me to discover a mistake and it is very useful. But why it discovered the issue only when I added stub?

PR #9287, opened 2023-02-14
Emit issues for calls to `is_a(string, class-string, false)`

weirdan:
Fixes vimeo/psalm#4092


weirdan:
I don't think you can: https://psalm.dev/r/f531eb1775

PR #9291, opened 2023-02-14
Do not display ext deprecation note when `disableExtensions` used

### LINKED ISSUE #8958: Deprecation: Psalm stubs for ext-apcu loaded using legacy way.

lptn:
Fixes #8958

To make it works, I needed to know xml config in `\Psalm\Config::visitStubFiles` or whether an ext disabled by config. But because public values of `\Psalm\Config::$php_extensions` are `false` by default, I can't say  - whether they falsy these ext disabled in xml config or this is just a default value. In order to fix it, I used a new default value - `null`, what is, in theory, a breaking change (`\Psalm\Config::$php_extensions`, Config class is not final). Also, I'm not sure that we should add more complexity just to hide deprecation message, so I'm open for new ideas how to solve it.


orklah:
Sorry for not merging earlier, it makes total sense. I kinda missed that one

Could you resolve the conflict and we'll be able to merge

lptn:
>Could you resolve the conflict and we'll be able to merge

done!

>EDIT: I don't think anyone would have used that array of extensions, and especially depend on the constant false value in there so I think we can ignore the BC break

Agree, for my understanding, this is BC formally only.

orklah:
Thanks!

weirdan:
>  this is BC formally only.

Maybe it's just me, but I find it super confusing when people use `BC` when they mean `BC break`. They are actually opposite things.

PR #9296, opened 2023-02-14
[READY] Simplify and fix Shepherd to support custom endpoints for reporting

lptn:
This PR focuses on 2 things:
 1. make Shepherd more readable (basically split into smaller methods)
 2. fix an issue with the custom endpoint via a small hack. As I wea asked in https://github.com/vimeo/psalm/pull/9112#issuecomment-1383239821, we needed to support deprecated behaviour of the Shepherd plugin, for this reason I made this change https://github.com/vimeo/psalm/pull/9112/commits/9271946dd6929c8a40c87d216120c034d5bca922#diff-af6fe3b522bfdc9c9a5b9059e58def7e6557ab86c81e7860adce3978d0e3bbbcL84 that broke support of custom endpoints.  The problem is that it's not possible to fully support both old behaviour and custom endpoints at the same time (in short - because we do `$base_address . '/hooks/psalm'` in Shepherd class), without hacks, and I found a smallest evil - do not append base_address by `/hooks/psalm` if the endpoint/address string ended by `#` (what is totally valid URL and server-side, it should normally have no difference).

Unfortunately, this code is not very testable (I meant automated tests), that's why I tested manually on a separate repository, using different combinations:
https://github.com/lptn/shepherd-test/actions/runs/4232108681/jobs/7351438882 

_______
Next steps: I would like to add more data to the report and start from Psalm version, number of found issues


orklah:
Thanks!

PR #9300, opened 2023-02-15
Add option to skip callmap test

weirdan:
With this change, callmap tests can be skipped with
`--exclude-group=callmap`:

```console
$ php-noxdebug vendor/bin/paratest --passthru-php="-dmemory_limit=-1" --exclude-group=callmap
```


weirdan:
@othercorey are tons of failures expected for these tests? I'm getting failures with `redis`, `imagick` and `memcached` locally.

othercorey:
No. It could be your local environment has extensions that GA does not. I will make sure those are enabled and check he list.  Easy to disable method checks if those are the problem.

PR #9310, opened 2023-02-15
[ibm_db2] Improve return type inference for `db2_autocommit()`

phansys:
Prior to version ibm_db2 2.1, the `db2_autocommit()` function was expecting a boolean as argument 2.

See:
* https://www.php.net/manual/en/function.db2-autocommit.php
* https://bugs.php.net/bug.php?id=77591
* php/pecl-database-ibm_db2@e9fb4b173046ab4fa1bb69125adf3de8d3818b97
* php/doc-en#2300

Given the release date of the version 2.1.0 in the extension (2020-11-30 / https://pecl.php.net/package/ibm_db2/2.1.0), I guess  8.1 is the lowest PHP release capable to ensure the new (integer) type.

Additionally, the return type of this function can be `int|bool` depending on the received arguments:
>When db2_autocommit() receives only the connection parameter, it returns the current state of AUTOCOMMIT for the requested connection as an integer value. A value of DB2_AUTOCOMMIT_OFF indicates that AUTOCOMMIT is off, while a value of DB2_AUTOCOMMIT_ON indicates that AUTOCOMMIT is on.
>
>When db2_autocommit() receives both the connection parameter and autocommit parameter, it attempts to set the AUTOCOMMIT state of the requested connection to the corresponding state. Returns true on success or false on failure.

Please, let me know if should I perform any extra change.


weirdan:
> Additionally, the return type of this function can be int|bool depending on the received arguments:

If you want Psalm to be smart about this, you will need to add this function to stubs (https://github.com/vimeo/psalm/tree/master/stubs) and use conditional return types / templates to specify the logic. E.g. something like this:

```php
/**
 * @return (func_num_args() > 1 ? bool : (1|2))
 */
function db2_autocommit($connection, $value = null): int|bool {}
```

Unfortunately, this is not possible in Callmap format.

othercorey:
Does the db2 extension have a PHP 8.1 minimum requirement?

phansys:
>Does the db2 extension have a PHP 8.1 minimum requirement?

No, I'was using that version because the date inference I mentioned, as in the extension docs I was not able to identify the requirements. But checking the repository, version 2.1.0 seems to declare [PHP 7.0.0](https://github.com/php/pecl-database-ibm_db2/blob/4eebfe146a873e25af26fbca30f480747cd237af/package.xml#L327) as minimum requirement.

I've pushed a new commit removing the deltas.
Thank you.

weirdan:
Thanks!

Issue #3166, opened 2020-04-17
Non-existent arguments should probably give an error


vudaltsov:
Over the time I find a lot of useless `@param` phpdocs in my project. They are for arguments that no longer exist. It would be nice if psalm could point them out.

https://psalm.dev/r/ca86c581f6

muglug:
I've always thought of this as more of a code style issue – if someone created a PR, and made it something that only appeared with `--find-unused-code` I'd happily accept it

weirdan:
Original snippet now reports a couple of `UnusedDocblockParam`s: https://psalm.dev/r/ca86c581f6

PR #9312, opened 2023-02-16
Deprecate `FunctionLikeStorage::$unused_docblock_params`

weirdan:
Refs vimeo/psalm#3166


weirdan:
To implement #3166 I need the storage always to contain the list of unused docblock params. However, the property being deprecated is populated only when there's at least one signature parameter with an underspecified type (like, no type or just `array`).

Instead, two new properties are added. One always contains a list of unused docblock params, and another tells whether there were any underspecified signature params, so that the original logic for emitting `InvalidDocblockParamName` can be moved to the read side.

PR #9313, opened 2023-02-16
Skip verifying callmaps based on name pattern

othercorey:
@weirdan refs https://github.com/vimeo/psalm/pull/9300#issuecomment-1430744026


othercorey:
Can add any extension pattern to the list if it is enabled and failing (and we can't fix it). For example, I did not have imagick and redis enabled when generating the ignore list, but they contain several hundred entries.

weirdan (approved):
Looks great

weirdan:
Thanks!

othercorey:
@weirdan Are you able to run verify-callmap without errors now?

PR #9323, opened 2023-02-17
Fix array_unique callmap

othercorey:
refs https://github.com/vimeo/psalm/issues/7956

The array element must be stringable regardless of sort type.


weirdan:
> The array element must be stringable regardless of sort type.

Shouldn't we accept non-stringable objects when `SORT_REGULAR` is used? It seems to work: https://3v4l.org/BLlbI

othercorey:
I was getting string conversion errors when testing. Will take a closer look.

othercorey:
Guess I had that test disabled when doing some of this. I'll move this to this stubs as was going to originally then.

othercorey:
@weirdan This does not seem to throw errors when using an array like `[(object)[], 1]`, any ideas? Does psalm not determine a union type for array values?

```php
/**
 * @psalm-pure
 *
 * @psalm-template TValue
 *
 * @param array<TValue> $array
 * @param (TValue is (\Stringable|string|int|float|bool|null) ? (0|1|2|5) : 0) $flags
 *
 * @return array<TValue>
 */
function array_unique(array $array, int $flags = SORT_STRING): array
{
}
```

weirdan:
>  Does psalm not determine a union type for array values?

It does. The problem, I suspect, is in the handling of conditional types in `@param` - it's not the most polished part of Psalm code.

othercorey:
Can you point me to that area of the code? Maybe there's some clear logic errors.

weirdan:
It should be happening somewhere around `src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php`, I suppose. Fair warning: template handling in Psalm is not for the faint of heart.

othercorey:
It seems like the array element type just gets merged into `mixed`, but don't know if that's just at the point of tracing:

https://psalm.dev/r/c2dbbe4982

othercorey:
@weirdan For now, reduced this to just fixing the array_unique() callmap and will have to tackle the stubs issue separate. Unmarked this PR as fixing the original issue.

weirdan:
Note that there is a return type provider for `array_unique` which may override both callmap and stubs.

https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayUniqueReturnTypeProvider.php

othercorey:
@weirdan I went back to the original commit for the array_unique return provider. It wants to preserve a non-empty array types (there were fixes to it in the history as well). This should be do-able with templates and re-added the stub along with missing tests. What do you think?

https://github.com/vimeo/psalm/commit/38de68f5079ac45a3a6f17d63091a4271c0819af#diff-0782323bf242be8068800ab7ffdd467b06492b5e60ff4ef2094673de4e9bcff9

othercorey:
@weirdan If you don't want to go this route with array_unique, we can close it.

weirdan:
> If you don't want to go this route with array_unique, we can close it.

I just didn't have the time and energy to review it. All maintainers are volunteers here, and we do it in our spare time. I want to keep this project rush-free, or it would feel like an unpaid job.

othercorey:
> > If you don't want to go this route with array_unique, we can close it.
> 
> I just didn't have the time and energy to review it. All maintainers are volunteers here, and we do it in our spare time. I want to keep this project rush-free, or it would feel like an unpaid job.

Well then I apologize for making it feel rushed. In large projects, sometimes reviews are passed on by simply not merging so I was going to free up your time to not have to keep analyzing.

You are all doing an excellent job with a difficult project especially with reviews and feedback.

orklah:
Thanks!

Issue #2206, opened 2019-10-04
NoInterfaceProperties triggers when checking if property exists on an object


pamil:
Runnable example: https://psalm.dev/r/de34c21744

Code from the example:
```php
<?php
  
interface SomeInterface
{
  public function doSomething(): void;  
}

final class FirstImplementation implements SomeInterface
{
  /** @var int */
  public $someField = 42;
  
  public function doSomething(): void {}
}

function magic(SomeInterface $x): void {
  if (!property_exists(get_class($x), 'someField')) {
    return;  
  }
  
  if (!isset($x->someField)) {
    return;
  }
  
  $x->someField = 0;
};

$x = new FirstImplementation();
magic($x);
assert($x->someField === 0);
```

It should be possible to access the property after checking that it exists.

Issue #4182, opened 2020-09-14
UndefinedThisPropertyFetch despite property_exists() check


caugner:
https://psalm.dev/r/6cfebb5320

Shouldn't the `property_exists()` check have the same effect as `isset()`?

https://psalm.dev/r/58bb4ad896

caugner:
Sorry, it should be `!isset` instead of `isset`:

https://psalm.dev/r/1b76e227ca

Weirdly Psalm does NOT emit the `UndefinedThisPropertyFetch` after the `isset`, even though the property `id` most likely does NOT exist after the `isset` check.

gndk:
https://psalm.dev/r/4121eceef8

gndk:
https://psalm.dev/r/d206ded407

weirdan:
This now works: https://psalm.dev/r/58bb4ad896

orklah:
Closing this then. #2206 will track the property_exists issue

AndrolGenhald:
Actually still broken. It should be a `!isset()` not `isset()`, and Psalm doesn't catch it.

https://psalm.dev/r/58bb4ad896

PR #9324, opened 2023-02-17
Add `has-property-` assertion

weirdan:
On itself it doesn't do much good as property fetch analyzer doesn't
support object intersections yet. But it's a groundwork for future
enhancements.


orklah:
Seems good. Calling isset on a property should probably create that assertion too (though it can't be reverted because the property could just be null or un-initialized)

orklah:
Does this fix https://github.com/vimeo/psalm/issues/2206 and maybe https://github.com/vimeo/psalm/issues/4182?

weirdan:
> though it can't be reverted 

By the way, how do we make assertions that can't be inverted?

weirdan:
> Does this fix #2206 and maybe #4182 

Not entirely. To fix those, we also need to fix property fetch analysis on intersections.

orklah:
> By the way, how do we make assertions that can't be inverted?

In theory, it's mainly this: https://psalm.dev/docs/annotating_code/assertion_syntax/#equality-assertions

IsIdentical and IsLooselyEqual can't be meaningfully inverted whereas IsType can

But those only work with a specific type that must be given to the assertion. However, I found that we already have an isset assertion (but it doesn't seem to do much with properties). In fact, we have two: IsIsset and IsEqualIsset. The former can be inverted into IsNotIsset but the latter can't (trying to invert it will result in the "useless" assertion: https://github.com/vimeo/psalm/blob/f11ed46f26325a5d8c04769e525336e7fadd51c5/src/Psalm/Storage/Assertion/IsEqualIsset.php#L14)

PR #9325, opened 2023-02-17
Skip verifying memcache extension callmap

othercorey:
The documentation does not match reflection. Users will have to maintain it manually.


weirdan:
Thanks!

PR #9326, opened 2023-02-17
Disable JIT-incompatible extensions to prevent warnings

weirdan:
This also enables JIT for `psalter` and syncs the list of disabled
extensions between `psalm`, `psalter` and `psalm-language-server`

/cc: @danog


danog:
LGTM!

PR #9331, opened 2023-02-18
Allow spatie/array-to-xml v3

gharlan:
https://github.com/spatie/array-to-xml/releases/tag/3.0.0


weirdan (approved):

weirdan:
v2 is supported as well (by me), but this can be useful to prevent conflicts with projects that use v3 only.

weirdan:
Thanks!

Issue #9335, opened 2023-02-19
First-class callable new causes internal error


edsrzf:
https://psalm.dev/r/e69536471b

This appears to have been overlooked when support was added for other types of first-class callable syntax. There are no test cases for it included in `ClosureTest`.

weirdan:
Note that this is an invalid construct not supported by PHP itself: https://3v4l.org/40D9J

Of course, Psalm should report it, rather than crashing.

PR #9336, opened 2023-02-19
Fix language server first-class callable crashes

edsrzf:
The language server crashes whenever it encounters any kind of first-class callable construct. This fixes the problem for functions, method, and static methods. (`new` looks like it requires some more work; see #9335.)

I'm unfamiliar with this code base, so am not sure if this is a _good_ fix, or if I've written the test in the best way. Very open to suggestions!

### DIFF
--- a/tests/ClosureTest.php
+++ b/tests/ClosureTest.php
@@ -2,6 +2,7 @@
 
 namespace Psalm\Tests;
 
+use Psalm\Context;
 use Psalm\Tests\Traits\InvalidCodeAnalysisTestTrait;
 use Psalm\Tests\Traits\ValidCodeAnalysisTestTrait;
 
@@ -11,6 +12,36 @@ class ClosureTest extends TestCase
 {
     use InvalidCodeAnalysisTestTrait;
     use ValidCodeAnalysisTestTrait;
+
+    public function testLanguageServerFirstClassCallable(): void
+    {
+        $code = '<?php
+            function foo(): void {}
+
+            class A {
+                public static function publicStatic(): void {}
+
+                public function publicMethod(): void {}
+            }
+
+            foo(...);
+            A::publicStatic(...);
+            (new A())->publicMethod(...);
+        ';
+
+        $context = new Context();
+
+        $this->project_analyzer->setPhpVersion('8.1.0', 'tests');
+
+        $codebase = $this->project_analyzer->getCodebase();
+        $codebase->enterServerMode();
+        $codebase->config->visitPreloadedStubFiles($codebase);
+
+        $file_path = self::$src_dir_path . 'somefile.php';
+
+        $this->addFile($file_path, $code);
+        $this->analyzeFile($file_path, $context);
+    }
 
     public function providerValidCodeParse(): iterable
     {
--- a/tests/ClosureTest.php
+++ b/tests/ClosureTest.php
@@ -2,7 +2,6 @@
 
 namespace Psalm\Tests;
 
-use Psalm\Context;
 use Psalm\Tests\Traits\InvalidCodeAnalysisTestTrait;
 use Psalm\Tests\Traits\ValidCodeAnalysisTestTrait;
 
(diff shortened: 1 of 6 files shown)


tests/ClosureTest.php
```diff
@@ -12,6 +13,36 @@ class ClosureTest extends TestCase
...
+            A::publicStatic(...);
+            (new A())->publicMethod(...);
+        ';
+
+        $context = new Context();
+
+        $this->project_analyzer->setPhpVersion('8.1.0', 'tests');
+
+        $codebase = $this->project_analyzer->getCodebase();
+        $codebase->enterServerMode();
```
edsrzf, line 37: I couldn't see a way to add this to `providerValidCodeParse` because this `enterServerMode` call was necessary to reproduce the issue.

weirdan:
I don't see why this would be specific to the language server mode.

weirdan:
Ah, I see it now. That's because those parts are only executed when `$codebase->store_node_types` is `true`, and Psalm normally doesn't enable it, but language server does.

tests/ClosureTest.php
weirdan: Looking at how `Codebase::$store_node_types` is used, I think we should rather enable it for all tests.

edsrzf:
I've added the `$codebase->enterServerMode()` line to the two traits that power many of the tests. I hope that's what you meant.

I also removed the test case I originally added, since after this change the existing tests can reproduce the crashes.

weirdan:
Actually, I can't reproduce those crashes locally, either with or without `Codebase::$store_node_types = true` (and the other flag that `enterServerMode()` sets also makes no difference). That's puzzling.

Don't get me wrong, the changes to `src` make total sense, but I would also like to ensure we have the correct tests for this problem.

edsrzf:
Were you able to reproduce the crash with my original test case?

For reference, the failures look like:

```
13) Psalm\Tests\ClosureTest::testInvalidCode with data set "FirstClassCallable:UndefinedMagicStaticMethod" ('<?php\n                    cl...      ', 'MixedAssignment', array(), '8.1')
Failed asserting that exception of type "AssertionError" matches expected exception "Psalm\Exception\CodeException". Message was: "assert(!$this->isFirstClassCallable())" at
psalm/vendor/nikic/php-parser/lib/PhpParser/Node/Expr/CallLike.php:36
psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentMapPopulator.php:38
psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php:332
psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php:207
psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticCallAnalyzer.php:208
psalm/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php:190
psalm/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php:86
psalm/src/Psalm/Internal/Analyzer/Statements/Expression/AssignmentAnalyzer.php:242
psalm/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php:544
psalm/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php:178
psalm/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php:86
psalm/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php:566
psalm/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php:195
psalm/src/Psalm/Internal/Analyzer/FileAnalyzer.php:179
psalm/tests/TestCase.php:128
psalm/tests/Traits/InvalidCodeAnalysisTestTrait.php:80
```

PHP version: 8.1.15
`nikic/php-parser` version: v4.15.3

Happy to share any other details that might be relevant.

weirdan:
> Were you able to reproduce the crash with my original test case?

Not initially; that's what got me to investigate this. It turned out I had `zend.assertions` ini setting set to `-1`.

weirdan (approved):

weirdan:
Thanks!

PR #9337, opened 2023-02-19
Added the test to validate test environment itself

weirdan:
We should make sure that the environment the tests are run is as strict
as possible. For example, zend.assertions should be enabled.


weirdan:
This would have prevented confusion like this: https://github.com/vimeo/psalm/pull/9336#issuecomment-1435916244

PR #9339, opened 2023-02-19
Negated is_numeric narrows scalar type

### LINKED ISSUE #8703: PossiblyInvalidArgument
I can't see how line 14 could be reached with `$literal` still being a float or an integer. IMO Psalm should consider it to be a string at that point.

https://psalm.dev/r/034888a55f

edsrzf:
This narrows `scalar` to `string|bool` when `is_numeric` returns false. (At least, if I've done it right! It seems to work.) Fixes #8703.

### DIFF
--- a/tests/TypeReconciliation/ConditionalTest.php
+++ b/tests/TypeReconciliation/ConditionalTest.php
@@ -203,7 +203,7 @@ class ConditionalTest extends TestCase
                      * @param scalar $v
                      * @return bool|string
                      */
-                    function toString($)
+                    function toString($v)
                     {
                         if (is_numeric($v)) {
                             return false;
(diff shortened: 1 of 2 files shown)


tests/TypeReconciliation/ConditionalTest.php
```diff
@@ -197,6 +197,20 @@ function foo ($a) {
                     '$a' => 'string',
                 ],
             ],
+            'typeRefinementonWithNegatedIsNumeric' => [
+                'code' => '<?php
+                    /**
+                     * @param scalar $v
+                     * @return bool|string
+                     */
+                    function toString($)
```
weirdan, line 206:
```suggestion
                    function toString($v)
```

weirdan (changes requested):

weirdan (approved):

weirdan:
Thanks!

PR #9353, opened 2023-02-20
Load opcache extension if it is loadable

danog:
This avoids warnings if the opcache extension isn't installed at all, and loads it if it isn't already loaded.
I've removed some PHP 7.4 exclusion code, IMO it doesn't hurt to load opcache on PHP 7.4, it's bound to yield at least some performance improvements even without JIT, at least due to opcode optimizations.

### DIFF
--- a/src/Psalm/Internal/Fork/PsalmRestarter.php
+++ b/src/Psalm/Internal/Fork/PsalmRestarter.php
@@ -63,7 +63,7 @@ class PsalmRestarter extends XdebugHandler
         }
         $cmd = [PHP_BINARY, '-n', '-dzend_extension=opcache', '-r', 'var_dump(function_exists("opcache_get_status"));'];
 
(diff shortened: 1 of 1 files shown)


weirdan:
Ain't this a bit too pushy, though? If we merge this as is, how do I, as an end user, would opt out of using OPcache? If I understand this correctly, to disable opcache (e.g. to workaround bugs like we had recently) I would need to *rename* the so file, is that right?

weirdan:
Moving `opcache.enable_cli` and `opcache.optimization_level` outside of the version-specific `if` should be ok.

danog:
You're right, this can wait a bit, at least until JIT-related issues are fixed upstream and we stop getting JIT-related reports.
Either way the user is always notified of the benefits of enabling JIT when running psalm, so it's OK anyway.

Issue #9352, opened 2023-02-20
php 8 interface method used in ErrorBaseline


SignpostMarv:
#9184 introduced a call to `DOMParentNode::append()`, which doesn't exist on 7.4

#8140 added the method to the stub, but didn't flag it as only being present from 8.x onwards, so it's falsely flagged as being fine

expected behaviour: set baseline should behave on php 7.4
actual behaviour: set baseline fails due to undefined method on `DOMElement`

SignpostMarv:
this impacts psalm 5.7.0 and 5.7.1

othercorey:
@SignpostMarv Did the patch fix your crash?

SignpostMarv:
@othercorey that appears to fix the crash, yes- so that covers the bug introduced by #9184, but the issue introduced by #8140 that you noted in your comment on #9354 (re: the php-versioned stubs) hasn't been fixed so the bug can't be closed yet.

othercorey:
I've created a separate issue to trach the architectural issue vs this crash from a BC break.

weirdan:
The crash is fixed and the rest is tracked in #9379

PR #9354, opened 2023-02-20
Replace use of DOMParentNode::append() with appendChild()

othercorey:
Possible fix for https://github.com/vimeo/psalm/issues/9352. The php-versioned stubs don't seem to throw the expected error here.


weirdan (approved):

weirdan:
Thanks!

PR #9356, opened 2023-02-21
Include setrawcookie as an impure function

edsrzf:
Without this, a call to `setrawcookie` when the return value isn't used causes an `UnusedFunctionCall` error.

https://psalm.dev/r/7764222d83


weirdan (approved):

weirdan:
Thanks!

PR #9360, opened 2023-02-21
Fix several spl callmap methods

othercorey:
Eventually, we'll be able to verify non-extension stubs against callmaps.


weirdan (approved):

weirdan:
Thanks!

PR #9362, opened 2023-02-21
Convert callmap associative-array types into proper array types

othercorey:
`associative-array` is just an alias for `array` and does not actually represent arrays with string keys. It was actually not used for literal associative arrays either so not sure why it was added.

Fixed a couple places where the type was incorrect.


dictionaries/CallMap.php
```diff
@@ -359,29 +359,29 @@
...
-'array_count_values' => ['associative-array<mixed,int>', 'array'=>'array'],
-'array_diff' => ['associative-array', 'array'=>'array', '...arrays='=>'array'],
-'array_diff_assoc' => ['associative-array', 'array'=>'array', '...arrays='=>'array'],
-'array_diff_key' => ['associative-array', 'array'=>'array', '...arrays='=>'array'],
-'array_diff_uassoc' => ['associative-array', 'array'=>'array', 'rest'=>'array', 'data_comp_func'=>'callable(mixed,mixed):int'],
-'array_diff_uassoc\'1' => ['associative-array', 'array'=>'array', 'rest'=>'array', 'arr3'=>'array', 'arg4'=>'array|callable(mixed,mixed):int', '...rest='=>'array|callable(mixed,mixed):int'],
-'array_diff_ukey' => ['associative-array', 'array'=>'array', 'rest'=>'array', 'key_comp_func'=>'callable(mixed,mixed):int'],
-'array_diff_ukey\'1' => ['associative-array', 'array'=>'array', 'rest'=>'array', 'arr3'=>'array', 'arg4'=>'array|callable(mixed,mixed):int', '...rest='=>'array|callable(mixed,mixed):int'],
+'array_combine' => ['array', 'keys'=>'string[]|int[]', 'values'=>'array'],
+'array_count_values' => ['array<int>', 'array'=>'array<string|int>'],
```
othercorey, line 366: Can be more specific since the values must be string|int.

dictionaries/CallMap.php
```diff
@@ -359,29 +359,29 @@
...
-'array_intersect' => ['associative-array', 'array'=>'array', '...arrays='=>'array'],
-'array_intersect_assoc' => ['associative-array', 'array'=>'array', '...arrays='=>'array'],
-'array_intersect_key' => ['associative-array', 'array'=>'array', '...arrays='=>'array'],
-'array_intersect_uassoc' => ['associative-array', 'array'=>'array', 'rest'=>'array', 'key_compare_func'=>'callable(mixed,mixed):int'],
-'array_intersect_uassoc\'1' => ['associative-array', 'array'=>'array', 'rest'=>'array', 'arr3'=>'array', 'arg4'=>'array|callable(mixed,mixed):int', '...rest'=>'array|callable(mixed,mixed):int'],
-'array_intersect_ukey' => ['associative-array', 'array'=>'array', 'rest'=>'array', 'key_compare_func'=>'callable(mixed,mixed):int'],
-'array_intersect_ukey\'1' => ['associative-array', 'array'=>'array', 'rest'=>'array', 'arr3'=>'array', 'arg4'=>'array|callable(mixed,mixed):int', '...rest'=>'array|callable(mixed,mixed):int'],
+'array_filter' => ['array', 'array'=>'array', 'callback='=>'callable(mixed,mixed=):scalar|null', 'mode='=>'int'],
+'array_flip' => ['array<string|int>', 'array'=>'array<string|int>'],
```
othercorey, line 377: The value must be string or int not the entire array. The input array has the same restriction.

orklah:
Thanks!

I think associative-array was a thing in Phan (from which the callmaps were used in the beginning)

othercorey:
@orklah I did not remove associative-array from the type parser in case it was a backwards compatibility break. Did not know if users were using this type.

weirdan:
> I think associative-array was a thing in Phan

Still is:
https://github.com/phan/phan/blob/93c1c287d646d08249b4f5539bfe147550b163d7/src/Phan/Language/Type/AssociativeArrayType.php#L96-L100

>  I did not remove associative-array from the type parser in case it was a backwards compatibility break.

It certainly is. I don't see a practical need to remove it from the parser anytime soon.

PR #9375, opened 2023-02-22
Support sizeof alias of count

### LINKED ISSUE #9372: Support for aliases?
Psalm does not seem to take PHP aliases into account. Examples:

- `count`: https://psalm.dev/r/9d70fdf31d
- `sizeof`: https://psalm.dev/r/8a26a0a0f9

Is this the expected behavior?

jsanahuja:
Closes #9372.

### DIFF
--- a/src/Psalm/Internal/Codebase/Functions.php
+++ b/src/Psalm/Internal/Codebase/Functions.php
@@ -555,7 +555,7 @@ class Functions
             return true;
         }
 
-        if ($function_id === 'count' && isset($args[0]) && $type_provider) {
+        if (in_array($function_id, ['count', 'sizeof']) && isset($args[0]) && $type_provider) {
             $count_type = $type_provider->getType($args[0]->value);
 
             if ($count_type) {
@@ -563,7 +563,7 @@ class Functions
                     if ($atomic_count_type instanceof TNamedObject) {
                         $count_method_id = new MethodIdentifier(
                             $atomic_count_type->value,
-                            'count',
+                            $function_id,
                         );
 
                         try {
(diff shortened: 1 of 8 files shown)


weirdan (changes requested):
There are also several places where `'count'` is used in the source code.

src/Psalm/Internal/Analyzer/Statements/Expression/MatchAnalyzer.php
src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
src/Psalm/Internal/Codebase/Functions.php

All of those looks like they need to be updated too.

jsanahuja:
I have found it too in `dictionaries` folder but I have no idea what is that. Should I update it there too?

weirdan:
> I have found it too in dictionaries folder but I have no idea what is that. Should I update it there too?

I'll fix them myself.

> PS: Same for src/Psalm/Report.php

That's unrelated, it's just a report type and has nothing to do with `count`/`sizeof` constructs

src/Psalm/Internal/Codebase/Functions.php
```diff
@@ -555,15 +555,15 @@ public function isCallMapFunctionPure(
...
+        if (in_array($function_id, ['count', 'sizeof']) && isset($args[0]) && $type_provider) {
             $count_type = $type_provider->getType($args[0]->value);
 
             if ($count_type) {
                 foreach ($count_type->getAtomicTypes() as $atomic_count_type) {
                     if ($atomic_count_type instanceof TNamedObject) {
                         $count_method_id = new MethodIdentifier(
                             $atomic_count_type->value,
-                            'count',
+                            $function_id,
```
weirdan, line 566: Here we're checking if `ClassName::count()` (an implementation of `Countable`) has side-effects. Since both `count` and `sizeof` will still call `ClassName::count()`, this should stay constant

```suggestion
                            'count',
```

weirdan (changes requested):

src/Psalm/Internal/Codebase/Functions.php
jsanahuja: Got it, changed :)

weirdan (approved):

weirdan:
Thanks!

Issue #9376, opened 2023-02-22
InvalidConstantAssignmentValue with no docblock type


edsrzf:
When a constant contains a long string, it's possible to receive an `InvalidConstantAssignmentValue` even though there's no docblock-defined type for the constant:

https://psalm.dev/r/8f0c1f98d6

I believe there are a few things happening here:

* The long string constant hits the max string length (#8075), so the assigned type of the constant is `non-empty-string` rather than `literal-string`.
* However, for some reason, the stored constant type is still `literal-string`. Maybe it's constructing its type through a different code path that doesn't respect the max string length?
* `ClassConstAnalyzer` is type-checking the constant, which kinda seems like a waste of effort since a constant without a docblock type can never result in a type error. (I think?)

PR #9381, opened 2023-02-23
Enforce literal string length limit

weirdan:
Fixes vimeo/psalm#9376

Ensures:
* that we never have a literal string exceeding the length limit
* that we call string interpreter for all literal strings

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php
@@ -170,7 +170,10 @@ class ArrayAssignmentAnalyzer
         $key_values = [];
 
         if ($current_dim instanceof PhpParser\Node\Scalar\String_) {
-            $key_values[] = new TLiteralString($current_dim->value);
+            $value_type = Type::getAtomicStringFromLiteral($current_dim->value);
+            if ($value_type instanceof TLiteralString) {
+                $key_values[] = $value_type;
+            }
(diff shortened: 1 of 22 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php
```diff
@@ -170,7 +170,10 @@ public static function updateArrayType(
         $key_values = [];
 
         if ($current_dim instanceof PhpParser\Node\Scalar\String_) {
-            $key_values[] = new TLiteralString($current_dim->value);
+            $value_type = Type::getAtomicStringFromLiteral($current_dim->value);
+            if ($value_type instanceof TLiteralString) {
```
orklah, line 174: Seems weird to ignore the key when it's not a literal, no?

orklah (approved):

src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php
weirdan: It kinda does, but then the same happens below if some of the keys are not literal (`if (count($string_literals) + count($int_literals) === count($all_atomic_types))`).

The only difference is that in that branch, `$key_values` is either empty or contains all keys, and here we can have partial results. On the other hand, no tests broke, so I'm not really sure if it's that critical.

weirdan: It could be an interesting exercise to run the entire test suite with `max_str_length = 2`

PR #9385, opened 2023-02-23
Fixes #9384 (Wrong type for variable checked with is_long)

### LINKED ISSUE #9384: Wrong type for variable checked with is_long
Documentation lists `is_long()` as an alias of `is_int`. Checking the type of a variable with `is_long` gives `float` instead of `int`.
See: https://psalm.dev/r/a362f0e861
Documentation: https://www.php.net/manual/en/function.is-long (and https://www.php.net/manual/en/aliases.php)

ygottschalk:
Not sure if I've put the test in the right Test Class...
Fixes #9384


weirdan (approved):

weirdan:
Thanks!

Issue #9382, opened 2023-02-23
Another preload issue


SCIF:
I have preload script configured. As of 5.7+ Psalm versions the preloading is forced on analysis start, I've got next errors appeared:

```
ERROR: UndefinedClass - app/File.php:36:24 - Class, interface or enum named Psl\Type\TypeInterface does not exist (see https://psalm.dev/019)
            $country = Type\non_empty_string()->coerce($country);
```

By adding `return;` on top of my preload files, the behavior of Psalm becomes fine (no new issues).

weirdan:
Can you provide a minimal reproducer in a separate repo?

SCIF:
> Can you provide a minimal reproducer in a separate repo?

Yep, trying to do that right now.

SCIF:
Managed to reproduce. Final docker grooming and will push soon.

SCIF:
Hi guys, I managed to reproduce the issue finally: https://github.com/SCIF/repro9382

weirdan:
Thanks, I've been able to reproduce using the linked repo.

weirdan:
Ok, the quick fix would be to disable preload here: https://github.com/vimeo/psalm/blob/2315c98a73f18244b9818fbbc72d6f0c26b47894/src/Psalm/Internal/Fork/PsalmRestarter.php#L103-L106

That's, I think, the best we can do at the moment. @danog would you agree?

danog:
Sure, adding a forced `opcache.preload=` config to the xdebug restarter (along with a config check) should fix the issue.

PR #9388, opened 2023-02-24
Disable `opcache.preload` if it's enabled

weirdan:
Fixes vimeo/psalm#9382

Also reorganized the opcache settings check a bit and enforced
`opcache.jit_buffer_size`

### DIFF
--- a/src/Psalm/Internal/Fork/PsalmRestarter.php
+++ b/src/Psalm/Internal/Fork/PsalmRestarter.php
@@ -14,6 +14,8 @@ use function implode;
 use function in_array;
 use function ini_get;
 use function preg_replace;
+use function strlen;
+use function strtolower;
 
 use const PHP_VERSION_ID;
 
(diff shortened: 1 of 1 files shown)


weirdan:
I've created 5.7.x branch so that we can merge this PR without necessarily releasing 5.8

src/Psalm/Internal/Fork/PsalmRestarter.php
```diff
@@ -53,28 +63,54 @@ protected function requiresRestart($default): bool
...
+                'enable_cli' => in_array(ini_get('opcache.enable_cli'), ['1', 'true', true, 1]),
+                'jit' => (int) ini_get('opcache.jit'),
+                'optimization_level' => (string) ini_get('opcache.optimization_level'),
+                'preload' => (string) ini_get('opcache.preload'),
+                'jit_buffer_size' => self::toBytes(ini_get('opcache.jit_buffer_size')),
+            ];
 
-            if (((int) ini_get('opcache.jit')) !== 1205) {
-                return true;
+            foreach (self::REQUIRED_OPCACHE_SETTINGS as $ini_name => $required_value) {
```
danog, line 78: Couldn't this be just an `if (self::REQUIRED_OPCACHE_SETTINGS !== $opcache_settings) { return true; }`?

danog (changes requested):
LGTM!

src/Psalm/Internal/Fork/PsalmRestarter.php
weirdan: Perhaps, but that would require the keys in both arrays to be in the same order, which could easily be overlooked when adding/removing items.

danog: Makes sense!

PR #9389, opened 2023-02-24
Update Phar callmaps

othercorey:
Verified against php 8.0 and php 8.2.


composer.json
```diff
@@ -117,7 +117,7 @@
             "Composer\\Config::disableProcessTimeout",
             "phpunit"
         ],
-        "verify-callmap": "phpunit tests/Internal/Codebase/InternalCallMapHandlerTest.php",
+        "verify-callmap": "@php phpunit tests/Internal/Codebase/InternalCallMapHandlerTest.php",
```
othercorey, line 120: This allows me to test multiple php versions. Thought I'd try to commit it instead of changing it every time.

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -61,8 +61,7 @@ class InternalCallMapHandlerTest extends TestCase
         '/^redis/', // redis extension
         '/^imagick/', // imagick extension
         '/^uopz/', // uopz extension
-        '/^memcache_/', // memcache extension
-        '/^memcache::/', // memcache extension
+        '/^memcache[_:]/', // memcache extension
```
othercorey, line 64: This is unrelated but not worth a separate PR. Simplified pattern to reduce number of patterns applied to every callmap (we'll add more in the future).

dictionaries/CallMap_80_delta.php
```diff
@@ -209,21 +209,17 @@
       'old' => ['bool', 'mode'=>'int'],
       'new' => ['bool', 'mode'=>'int', '...args='=>'mixed'],
     ],
-    'PharData::compress' => [
```
othercorey, line 212: Some of these entries were out of order so the diff is weird.

weirdan (approved):

weirdan:
Thanks!

Issue #9373, opened 2023-02-22
UndefinedConstant when using enum property as const array key


Brajk19:
When using `Enum::CASE->value` or `Enum::CASE->name` as key in constant, psalm throws UndefinedConstant error.
No error is thrown when enum property is used as key in regular array (not const).

https://psalm.dev/r/3307971eb7

weirdan:
The linked commit looks like it should fix it.

ygottschalk:
Still working on it, did not really fix the issue. PR should be incoming today

PR #9393, opened 2023-02-24
Fix/9373 using enum property as const array key

ygottschalk:
Please review, I'm not too fimiliar with the codebase

#9373


ygottschalk:
Ready to merge.

Note that the syntax is only valid in php >= 8.2. I did not include any checks for that... 
If desired, the check should be added to this [`if`](https://github.com/vimeo/psalm/pull/9393/files#diff-6836f9419b6ecd7cbb0eac91e803a29f835331049d3da5875a9422dfecf426d1R303) (something like `$codebase->config->php_version >= 8.2`)

orklah:
Thanks! Sorry, this PR flew under my radar!

Issue #6796, opened 2021-11-02
psalm doesn't recognize some grpc classes & constants


nocive:
psalm seems to have some trouble recognizing some grpc classes and constants.

```
$ php --version
PHP 8.0.12 (cli) (built: Oct 21 2021 16:45:17) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.12, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.12, Copyright (c), by Zend Technologies

$ php -i | grep grpc
/usr/local/etc/php/conf.d/docker-php-ext-grpc.ini,
grpc
grpc support => enabled
grpc module version => 1.41.0
grpc.enable_fork_support => 0 => 0
grpc.grpc_trace => no value => no value
grpc.grpc_verbosity => no value => no value
grpc.log_filename => no value => no value
grpc.poll_strategy => no value => no value

$ psalm --version
Psalm 4.x-dev@
```

Despite having the extension enabled, the following code fails https://psalm.dev/r/60285c8421.
Additionally, all the constants from the `Grpc` namespace are also not recognized, eg: https://psalm.dev/r/1a904c1046.

I noticed that there are some stubs defined in the `dictionaries/CallMap.php` and also that some internal psalm code disables the `grpc` extension under some circumstances, so I'm wondering if that could have anything to do with it.

Any help or hints would be appreciated. Thanks!

weirdan:
Does it work with `--threads=1`?

nocive:
> Does it work with `--threads=1`?

@weirdan yes!

I guess it's indeed because of https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Cli/Psalm.php#L895?
I can easily force `--threads=1` for my runs to work around the issue for now, thanks!

Out of curiosity, what's the background here for having to disable the extension if `threads > 1`?

weirdan:
I'm not sure. The commit is here: https://github.com/vimeo/psalm/commit/73aa0df08a580ddfad512b35586b7f698624fc3a, but it doesn't reference any issue and does not provide any explanation. Not sure if @muglug remembers that either.

orklah:
I'm starting to wonder if we should really require having enabled extensions in order to use the signatures in callmaps. There's a few other issues that are related where people are suprised the signature they documented are not used because they lack the extension. Here it's even worse because we disable the extension ourselves

muglug:
[...]

PR #9398, opened 2023-02-25
Only disable grpc extension if it is not properly configured

gndk:
Related to #6796, but I'm not sure if it fixes it.

I ran into a conflict between my composer platform requirements and hardcoded disabling of this extension in psalm after installing `grpc` extension as a requirement of a google library.

This PR combines muglug's and weirdan's suggestions from the issue, to disable the extension only if it is not configured correctly (see https://github.com/grpc/grpc/issues/20250#issuecomment-531321945). It now also displays a warning message, instead of silently disabling it.

> I added this code because of https://github.com/grpc/grpc/issues/13412 (the same was true I think of the APC extension). Since that issue has been fixed for a few years, and users can selectively disable extensions anyway, it should be good to remove this exception.

> So I suspect we would still have to have some special treatment for grpc ext, e.g. check if the environment var is set and warn + downgrade to single thread otherwise.

I'd be happy to add tests, however I don't know if/what makes sense for something like this.

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -44,6 +44,7 @@ use function array_sum;
 use function array_values;
 use function chdir;
 use function count;
+use function extension_loaded;
 use function file_exists;
 use function file_put_contents;
 use function function_exists;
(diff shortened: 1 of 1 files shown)


weirdan (approved):

weirdan:
Thanks!

PR #9415, opened 2023-02-26
An attempt to fix #4460

### LINKED ISSUE #4460: Using setters with class-string-map
I've been experimenting with undocumented `class-string-map` type, and I think the following (slight modification of the passing test case) should be type-safe:  https://psalm.dev/r/53e6c832e1

Related: #4458 (using a `class-string-map`  there could be a valid solution).

ptomulik:
A quick fix made just using my intuition. I guess, the ``$offset_already_existed`` should only be applied to list, to avoid introducing discontinuities in lists. This is not the case for class-string-map however, where only the type of key and value matters.


orklah:
I can't say I'm certain about this, but it breaks no tests, it solves the issue and makes some sense so let's try it :)

Issue #8626, opened 2022-10-27
Uncaught Exception: InvalidArgumentException Could not get class storage for class-string-map


yunosh:
Emitted in /Users/janschneider/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:45

This is a strange class name it is looking for, obviously not from my code base.
Running psalm with --debug-by-line reveals that it stops at line 111 of a compiled Symfony container:

```
110:        $this->privates['service_container'] = function () {
111:            include_once \dirname(__DIR__, 4).'/vendor/league/openapi-psr7-validator/src/PSR7/ValidatorBuilder.php';
112:            include_once \dirname(__DIR__, 4).'/src/EventListener/RequestLanguageListener.php';
```

orklah:
class-string-map is a Psalm's pseudo type that describe an array which has class-string offsets and for each offset, contains an instance of that class-string.

I'm not sure why Psalm would fail to recognize it though. Do you have the latest Psalm version? Do you know what `/src/EventListener/RequestLanguageListener.php` contains?

yunosh:
Yes, this is both with Psalm 4.15.0 (locally) as with the latest 4.29.0 (inside Docker container). RequestLanguageListener ist harmless:

```php
<?php
declare(strict_types = 1);

namespace XXX\EventListener;

use Symfony\Component\HttpKernel\Event\RequestEvent;

final class RequestLanguageListener
{

    public function onKernelRequest(RequestEvent $event): void
    {
        if (!$event->isMainRequest()) {
            return;
        }

        $request = $event->getRequest();
        $request->setLocale($request->headers->get('Accept-Language', 'de-DE'));
    }

}
```

weirdan:
`class-string-map` is likely something used (or inferred) in the container, as it will likely contain a map of class/interface names to instances of those classes/interfaces. Can you upload your compiled container somewhere for us to see?

yunosh:
If you are talking about the Symfony container of the app that's being analyzed, this won't help, because the error even occurs if there is no container at all.
Or do you mean the container of Psalm (does it have one?)? That's running off a separate Docker instance.

weirdan:
I mean the DI container used by Symfony.

Also, it would be useful to see the backtrace for this exception.

yunosh:
[...]

PR #9418, opened 2023-02-26
Fix #9401

### LINKED ISSUE #9401: Internal Psalm error - Could not get class storage for class-string-map
https://psalm.dev/r/dd5e1b7f63

ptomulik:
An attempt to fix #9401 and #8626.


orklah:
Thanks!

PR #9420, opened 2023-02-27
Fix remaining spl and iterator callmaps

othercorey:


dictionaries/CallMap.php
```diff
@@ -604,7 +604,7 @@
 'call_user_func_array' => ['mixed|false', 'callback'=>'callable', 'args'=>'list<mixed>'],
 'call_user_method' => ['mixed', 'method_name'=>'string', 'object'=>'object', 'parameter='=>'mixed', '...args='=>'mixed'],
 'call_user_method_array' => ['mixed', 'method_name'=>'string', 'object'=>'object', 'params'=>'list<mixed>'],
-'CallbackFilterIterator::__construct' => ['void', 'iterator'=>'Iterator', 'func'=>'callable(mixed):bool|callable(mixed,mixed):bool|callable(mixed,mixed,mixed):bool'],
+'CallbackFilterIterator::__construct' => ['void', 'iterator'=>'Iterator', 'callback'=>'callable(mixed):bool|callable(mixed,mixed):bool|callable(mixed,mixed,mixed):bool'],
```
othercorey, line 607: Is this the correct/only way to have a callable with optional arguments?

orklah: `callable(mixed,mixed=): bool` means "a callable with a mandatory first param then an optional second param"

othercorey: I changed that callback to use 2 optional params instead.

orklah:
Thanks!

Issue #9411, opened 2023-02-26
Concat should never remove "non-empty-"/"non-falsy-" from string


kkmuffme:
https://psalm.dev/r/0baa3234b5

When concatenating a non-empty/non-falsy string with something else that is unknown (constants), the type should not "fall back", since it's wrong. And instead keep the type it already has (e.g. non-falsy-string in my example)

orklah:
It's handled here: https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ConcatAnalyzer.php#L255

It should be a pretty simple change

EgorBakulin:
> It's handled here:
> 
> https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ConcatAnalyzer.php#L255
> 
> It should be a pretty simple change

The problem is more complex than it might seem.

https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ConcatAnalyzer.php#L65

This variable is null for this snippets https://psalm.dev/r/f264a4732f https://psalm.dev/r/9ef452ff58

orklah:
yeah, I guess it makes sense given FOO was not defined. Then https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ConcatAnalyzer.php#L69 will return false, so adding an else and setting $result_type accordingly with the known other operand should still be easy

EgorBakulin:
#9422 I created pool request with fixes. But I newer worked with gihub pool requests. And I don't know how to link pool request to this issue.

PR #9422, opened 2023-02-27
concat should never remove non empty non falsy from string #9411

EgorBakulin:


src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ConcatAnalyzer.php
```diff
@@ -277,6 +278,41 @@ public static function analyze(
                     }
                 }
             }
+        } elseif ($left_type || $right_type) {
+            /**
+             * @var Union $known_operand
+             */
+            $known_operand = $right_type ?? $left_type;
+
+            $numeric_type = new Union([
```
orklah, line 287: I think this is a wrong copy/paste because you can't assume a numeric type concatenated with anything will still produce a numeric type.

You can make this assumption with non-empty-string and non-falsy-string but it stops there. So instead of building an Union here, you should start one from scratch with something like Type::getNonEmptyString()

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ConcatAnalyzer.php
```diff
@@ -277,6 +278,41 @@ public static function analyze(
...
+                new TNumericString,
+                new TInt,
+                new TFloat,
+            ]);
+
+            $non_empty_string = $numeric_type->getBuilder()->addType(new TNonEmptyString())->freeze();
+
+            $non_falsy_string = $numeric_type->getBuilder()->addType(new TNonFalsyString())->freeze();
+
+            $known_operand_not_empty = UnionTypeComparator::isContainedBy(
```
orklah, line 297: You can use something like 
`if($known_operand->isSingle() && $known_operand->getSingleAtomic() instanceof TNonEmptyString)` to check for the type in the Union, it will be quicker

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ConcatAnalyzer.php
```diff
@@ -277,6 +278,23 @@ public static function analyze(
...
+            if ($known_operand->isSingle()) {
+                $known_operands_atomic = $known_operand->getSingleAtomic();
+
+                if ($known_operands_atomic instanceof TNonEmptyString) {
+                    $result_type = Type::getNonEmptyString();
+                }
+
+                if ($known_operands_atomic instanceof TNonFalsyString) {
+                    $result_type = Type::getNonFalsyString();
+                }
```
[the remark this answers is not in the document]
EgorBakulin: Ok i will

orklah:

Issue #9429, opened 2023-02-28
Properties passed as arg 1 to `array_pop` are mutated, but checks for pure methods do not see the property as mutated


DaveLiddament:
See example: https://psalm.dev/r/b0916c1ed2

I'd expect line 16 to be reported as an error. Something along the lines of:

[ImpurePropertyAssignment](https://psalm.dev/204) - 16:9 - Cannot assign to a property from a mutation-free context

weirdan:
Naive approach doesn't work

PR #9434, opened 2023-02-28
`array_pop` is impure

weirdan:
Fixes vimeo/psalm#9429


weirdan:
Oops, I thought we had mandatory checks.

PR #9435, opened 2023-02-28
Revert "`array_pop` is impure"

weirdan:
Reverts vimeo/psalm#9434

It's more complex than I thought.


orklah:
I'm not sure array_pop is impure on it's own. The issue is that it changes whatever you give it because of the reference.

I think what needs to change is "any property pushed into a reference make the call impure"  (note: it still may be external-mutation-free)

PR #9436, opened 2023-03-01
Issue 9433

ptomulik:
An attempt to fix #9433. Some errors appeared, reported by "Run Shepherd" action. Need a little guidance to fix this, but it looks like the immediate effect of bug fixing (i.e. the code seemed to work thanks to false-negative behavior, that has just been fixed here).

PR #9439, opened 2023-03-01
Fix #9433

### LINKED ISSUE #9433: IfAnalyzer incorrectly updates outer scope
Consider this example, which correctly reports ReferenceConstraintViolation (dumb inserts into ``list`` may introduce discontinuites in indexint turning ``list`` into ``array``):

https://psalm.dev/r/461cbad1b2

Now, adding dummy ``if(isset(...)){}``, which has no chance to change flow of control, results with false negative (error disappears):

https://psalm.dev/r/dd0484a239

ptomulik:
A follow up of accidentally closed #9436. As it was said, shepherd tests started reporting errors related to list. This needs closer look.

### DIFF
--- a/src/Psalm/Internal/Type/TypeCombiner.php
+++ b/src/Psalm/Internal/Type/TypeCombiner.php
@@ -551,6 +551,8 @@ class TypeCombiner
             }
 
             foreach ($type->type_params as $i => $type_param) {
+                // See https://github.com/vimeo/psalm/pull/9439#issuecomment-1464563015
+                /** @psalm-suppress PropertyTypeCoercion */
                 $combination->array_type_params[$i] = Type::combineUnionTypes(
                     $combination->array_type_params[$i] ?? null,
                     $type_param,
@@ -599,6 +601,8 @@ class TypeCombiner
 
         if ($type instanceof TClassStringMap) {
             foreach ([$type->getStandinKeyParam(), $type->value_param] as $i => $type_param) {
+                // See https://github.com/vimeo/psalm/pull/9439#issuecomment-1464563015
+                /** @psalm-suppress PropertyTypeCoercion */
                 $combination->array_type_params[$i] = Type::combineUnionTypes(
                     $combination->array_type_params[$i] ?? null,
                     $type_param,
--- a/src/Psalm/Type/Reconciler.php
+++ b/src/Psalm/Type/Reconciler.php
@@ -761,7 +761,7 @@ class Reconciler
                                 return null;
                             }
 
-                            $new_base_type_candidate = $existing_key_type_part->getGenericValueType();
+                            $new_base_type_candidate = $existing_key_type_part->getGenericValueType(true);
                         } else {
                             $array_properties = $existing_key_type_part->properties;
 
--- a/tests/TypeReconciliation/IssetTest.php
+++ b/tests/TypeReconciliation/IssetTest.php
@@ -1153,6 +1153,25 @@ class IssetTest extends TestCase
                     }',
                 'error_message' => 'RedundantPropertyInitializationCheck',
             ],
+            'setArbitraryListElementAfterIsset' => [
+                'code' => '<?php
+                    /** @param list<string> $list */
+                    function foo(array &$list, int $offset): void {
+                        if (isset($list[$offset])) {}
+                        $list[$offset] = "";
+                    }',
+                'error_message' => 'ReferenceConstraintViolation'
+            ],
+            'setArbitraryListWithinNotIsset' => [
+                'code' => '<?php
+                    /** @param list<string> $list */
+                    function foo(array &$list, int $offset): void {
+                        if (!isset($list[$offset])) {
+                            $list[$offset] = "";
+                        }
+                    }',
+                'error_message' => 'ReferenceConstraintViolation'
+            ],
         ];
     }
 }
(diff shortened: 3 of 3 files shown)


orklah:
I'm not familiar with the piece of code you changed. Can you explain what you found during your debug and if you know the reason why it fixes your issue?

It seems to me the new errors raised in Shepherd are linked to another issue in Psalm (that already pops up on other occasions) and your fix just revealed them but I'm not sure of that for now

ptomulik:
> I'm not familiar with the piece of code you changed. Can you explain what you found during your debug and if you know the reason why it fixes your issue?

The very first thing I found, was the line where the outer scope of the ``if(isset($list[$offset])) { }`` gets incorrectly updated. This was the following line:

https://github.com/vimeo/psalm/blob/381a23bd4eb9207c0e4b0fc8cfdc5365db6bf268/src/Psalm/Internal/Analyzer/Statements/Block/IfElse/IfAnalyzer.php#L289

After that line, the ``$outer_context->vars_in_scope`` contained extra variable: ``$list[$offset]``. This only happened for ``$list`` declares as ``list<>``. When declared as  ``array<>``, the ``$outer_context`` was not updated (this is what we expect also for ``list<>``). The difference cause was ``$pre_assignment_else_redefined_vars`` provided as argument to ``IfAnalyzer::analyze()``. Searching for place, where that argument is prepared, led me to the piece of code starting at the following line

https://github.com/vimeo/psalm/blob/06266c220b53c1ce0ccd0303e1e44402ea3af4f0/src/Psalm/Type/Reconciler.php#L710

What I found was, that for ``array<>``  the code was setting a ``$new_base_type_candidate->possibly_undefined`` flag, whereas for ``list<>`` the flag was not set.  For ``list<>`` (which is handled by TKeyedArray case here) the ``$new_base_type_candidate`` is prepared entirely within the following call

https://github.com/vimeo/psalm/blob/06266c220b53c1ce0ccd0303e1e44402ea3af4f0/src/Psalm/Type/Reconciler.php#L764

The method ``$existing_key_type_part->getGenericValueType()`` used here  accepts boolean argument ``$possibly_undefined`` which defaults to false. So, I just passed ``true`` here.

ptomulik:
To be honest, I'm not very sure about that hard-coded ``true`` passed to ``getGenericValueType()``. It's just a first try. Perhaps some additional logic should be implemented to chose between ``true`` and ``false``.  The piece of code handling ``array<>`` doesn't seem so straightforward w.r.t to

orklah:
Well, that does seem consistent and I'm pretty sure the bug that appeared exists on other places too. I'd advise suppressing the new issues and go with it. If we somehow find an adverse effect, we'll at least add a new test and we'll be able to revert this if needed.

Can you fix the CI? I'll proceed to merge afterwards

ptomulik:
Well, my nose says me it is the right place, except we shouldn't pass ``true`` (possiblyUndefined) unconditionally, the same way as we shouldn't pass ``false`` (default value for the parameter). I believe, the solution should be based on what is implemented for ``TArray``:

```
if (($has_isset || $has_inverted_isset) && isset($new_assertions[$new_base_key])) {
                                if ($has_inverted_isset && $new_base_key === $key) {
                                    $new_base_type_candidate = $new_base_type_candidate->getBuilder();
                                    $new_base_type_candidate->addType(new TNull);
                                    $new_base_type_candidate->possibly_undefined = true;
                                    $new_base_type_candidate = $new_base_type_candidate->freeze();
                                } else {
                                    $new_base_type_candidate = $new_base_type_candidate->setPossiblyUndefined(true);
                                }
```

orklah:
Well, the way I see it:
we're in getValueForKey, a method used to get a potential value from an array for a given offset of said array.
We get in this condition:
https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Type/Reconciler.php#L759
because we have a TKeyedArray (we had a TList but is was converted to TKeyedArray at line 702

This condition in particular checks what the offset looks like. We enter here if we have an offset that is a variable (first operand) or if we have an offset that is not a literal string or a literal number (second operand).

So yeah, we could be more precise in some case. For example, if the offset is a non-numeric string, then it will never be defined because we have a list. If we can get the value for the variable in the offset, we could try to see if it's a known offset for the current array and keep the value defined.

But in other cases, if we have no way to know what the offset is, we're forced to return a possibly undefined type.

The thing is, we never do that kind of checks with TArray. Both condition will return possibly_undefined to true so I may be missing something here

ptomulik:
> Can you fix the CI? I'll proceed to merge afterwards

I see that CI fails same way for this PR and for master? Shouldn't we investigate this?

weirdan:
> I see that CI fails same way for this PR and for master? 

Does it? I see one job failed on master, due to the composer timing out, but here you have an actual error reported by Psalm.

ptomulik:
I think, I found explanation for the failing shepherd test. It's just revealed another variant of the bug we attempt to fix in this PR. To this end, we were concerned on this variant:

```php
if (isset($list[$i])) { }
$list[$i] = something... // false negative (psalm assumes falsely that $list[$i] exists)
```

The other variant, that is (in a simplified form):

```php
$list[$i] = expression (... $list[$i] ?? something... );
```

namely,  it's this fragment of code from ``src/Psalm/Internal/Type/TypeCombiner.php`` (starting from line 554)
```php
$combination->array_type_params[$i] = Type::combineUnionTypes(
  $combination->array_type_params[$i] ?? null,
  $type_param,
  $codebase,
  $overwrite_empty_array,
);
```

I think the best we can do at this moment is to just add @psalm-suppress here. The proper solution seems to be impossible now, because of incomplete implementation of list indexing in psalm (consider the following example, which is fine, but psalm is not smart enough to recognize it):

https://psalm.dev/r/8972d76050

orklah:
Thanks!

Issue #4192, opened 2020-09-16
psalter --issues=MissingParamType adds duplicate @param annotations


caugner:
I have observed several occurrences where `psalter` added duplicate `@param` annotations when there has already been an (invalid) `@param` annotation without a type. It would be nice if Psalm could reuse those annotations.

Before:
```php
/**
 * @param $x
 */
function foo($x) {}

foo(true);
foo(false);
```

After:
```php
/**
 * @param $x
 * @param bool $x
 */
function foo($x) {}

foo(true);
foo(false);
```

PS: This example is abstracted from what I observed.

caugner:
Alternatively, `@param $x` should cause an `InvalidDocblock` issue:

https://psalm.dev/r/d560c42190

orklah:
Fixed by #9441

PR #9441, opened 2023-03-02
Merge param descriptions when adding or updating types

### LINKED ISSUE #9440: Psalter: Docblock param descriptions duplicated or removed
I found two issues when getting Psalter to add/fix parameters into doc blocks (either `MissingParamType` or `MismatchingDocblockParamType`, both with `safe_types`) where the existing param descriptions are either ignored or removed.

I made one issue as these seem related. Happy to split if you prefer.

### Issue 1: Param is duplicated

If I have the following code and I run Psalter to add a type in then it duplicates the param in the doc block rather than merging it.

```php
class C {
    /**
     * @param $ab the string you pass in
     */
    public function fooFoo($ab): void {}
}

/**
 * @param string $ab
 */
function callsWithString($ab): void {
    (new C)->fooFoo($ab);
}
```
is turned into
```php
class C {
    /**
     * @param $ab the string you pass in
     * @param string $ab
     */
    public function fooFoo($ab): void {}
}

/**
 * @param string $ab
 */
function callsWithString($ab): void {
    (new C)->fooFoo($ab);
}
```

Where as I would hope would be that the docblock would become
```php
/**
 * @param string $ab the string you pass in
 */
```

### Issue 2: Param description removed when fixing incorrect doc block types

Worse, and somewhat relatedly, if Psalter is fixing a return type, it drops the param description entirely.

```php
/**
 * @param int $s the string
 */
function foo(string $s): string {
    return "hello";
}
```
becomes
```php
/**
 * @param string $s
 */
function foo(string $s): string {
    return "hello";
}
```

aboyton:
Fixes #9440

### DIFF
--- a/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
+++ b/src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
@@ -17,7 +17,9 @@ use Psalm\Internal\Scanner\ParsedDocblock;
 use function array_key_exists;
 use function array_merge;
 use function array_reduce;
+use function array_slice;
 use function count;
+use function implode;
 use function is_string;
 use function ltrim;
 use function preg_match;
@@ -332,7 +334,29 @@ class FunctionDocblockManipulator
                 foreach ($parsed_docblock->tags['param'] as &$param_block) {
                     $doc_parts = CommentAnalyzer::splitDocLine($param_block);
 
+                    // If there's no type
+                    if (($doc_parts[0] ?? null) === '$' . $param_name) {
+                        // If the parameter has a description add that back
+                        if (count($doc_parts) > 1) {
+                            $new_param_block .= " ". implode(" ", array_slice($doc_parts, 1));
+                        }
+
+                        if ($param_block !== $new_param_block) {
+                            $modified_docblock = true;
+                        }
+
+                        $param_block = $new_param_block;
+                        $found_in_params = true;
+                        break;
+                    }
+
+                    // If there is a type
                     if (($doc_parts[1] ?? null) === '$' . $param_name) {
+                        // If the parameter has a description add that back
+                        if (count($doc_parts) > 2) {
+                            $new_param_block .= " ". implode(" ", array_slice($doc_parts, 2));
+                        }
+
                         if ($param_block !== $new_param_block) {
                             $modified_docblock = true;
                         }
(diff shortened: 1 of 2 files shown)


orklah:
Seems great! I'm surprised about the error on CircleCI, I think this was fixed some times ago. Can you rebase on up to date master to see if it gets fixed?

src/Psalm/Internal/FileManipulation/FunctionDocblockManipulator.php
```diff
@@ -332,7 +334,29 @@ private function getDocblock(): string
...
+                            $modified_docblock = true;
+                        }
+
+                        $param_block = $new_param_block;
+                        $found_in_params = true;
+                        break;
+                    }
+
+                    // If there is a type
                     if (($doc_parts[1] ?? null) === '$' . $param_name) {
```
weirdan, line 354: Doesn't it actually handle the case when there's *no* type?

weirdan: Ah, ignore me. `$doc_parts[0]` is (normally) a type, and `$doc_parts[1]` is the param name.

aboyton:
> Seems great! I'm surprised about the error on CircleCI, I think this was fixed some times ago. Can you rebase on up to date master to see if it gets fixed?

I force pushed an update, not sure if that's correct. Turns out I'm accidentally still on the branch I made from my last PR. Let me know if this hasn't worked properly.

weirdan:
> I force pushed an update, not sure if that's correct

It's fine.

> Turns out I'm accidentally still on the branch I made from my last PR

Not a problem either.

All is good now.

weirdan (approved):

aboyton:
Thanks for the quick merge. How often do you cut a new release?

orklah:
Note: https://github.com/vimeo/psalm/issues/4192 should be fixed or almost fixed:
@aboyton if you still feel like working on it, you could try to check if this snippet https://psalm.dev/r/1d19c39522 with --safe-types=true fixes the partial @param annotation correctly :)

aboyton:
I somehow completely missed #4192. Yes, the code from https://psalm.dev/r/1d19c39522 with `--safe-types=true` used to return

```php
/**
 * @param $x
 * @param bool $x
 */
function foo($x) {}
```
and with this change it now returns
```php
/**
 * @param bool $x
 */
function foo($x) {}
```
as expected, so I think #4192 can be closed now too.

orklah:
Cool! Thanks!

weirdan:
> How often do you cut a new release?

There's no fixed schedule, as @orklah said, but lately it's been once in 2-4 weeks unless some critical issue appears.

Issue #6782, opened 2021-10-29
@psalm-type does not support parentheses


sctice-ifixit:
https://psalm.dev/r/a55241a2df

The error message implies a syntax error, but it seems to be the case that parentheses are just unsupported in `@psalm-type`. Perhaps that's by design (last I checked, the docs have very little to say about parentheses). If it is intentional or not worth supporting, perhaps the error message could be a little more straightforward? "Unterminated parentheses" seems incorrect; perhaps "Parentheses are not supported in @psalm-type"?

The parentheses in the example are obviously superfluous. It's less obvious (though probably still superfluous?) if you want to define a large union over multiple lines.

danielbeardsley:
Hi Shawn! I ran into the same thing, but I think it's even more odd.

`@psalm-type, anything after the `>` greater-than symbol is just straight-up ignored by the parser:

https://psalm.dev/r/67abe4ae45

PR #9458, opened 2023-03-05
Do not strip the last `)` bracket when parsing phpdoc type aliases.

pvandommelen:
Fixes https://github.com/vimeo/psalm/issues/6782

The example in the tests is a bit simplistic. The type which was causing problems for us can be found at https://github.com/webonyx/graphql-php/blob/5ad059b288f5e21e77cc9dcc45a525bb929e77bc/src/Type/SchemaConfig.php#L29.

Reworked the regex logic to not check each grouping character separately. Removed the `^` character from the character list, which was likely unintended.


weirdan (approved):

weirdan:
Thanks!

PR #9459, opened 2023-03-06
Replace package-versions with Composer 2 APIs

Jean85:
`composer/package-versions-deprecated` is currently required by Psalm. It's a fork of the original `ocramius/package-versions` package, and there's currently a circular "replace" directive between the two packages, which may create some issues for end users:
* https://github.com/Ocramius/PackageVersions/blob/fa375ea469c9d6a5b30184cc2aa1e2cb0250598d/composer.json#L24-L26
* https://github.com/composer/package-versions-deprecated/blob/b4f54f74ef3453349c24a845d22392cd31e65f1d/composer.json#L20-L22

The package itself is used only once in Psalm, and it also has a `@deprecation` notice, advising to migrate toward Composer 2 APIs: https://github.com/composer/package-versions-deprecated/blob/b4f54f74ef3453349c24a845d22392cd31e65f1d/src/PackageVersions/Versions.php#L21

This PR does exactly this, replacing the requirement for the package for a requirement against Composer 2, and using its direct APIs.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "ext-tokenizer": "*",
         "amphp/amp": "^2.4.2",
         "amphp/byte-stream": "^1.5",
-        "composer/package-versions-deprecated": "^1.10.0",
+        "composer-runtime-api": "^2",
         "composer/semver": "^1.4 || ^2.0 || ^3.0",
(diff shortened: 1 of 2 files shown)


weirdan:
Thanks!

PR #9460, opened 2023-03-06
Fix DOM callmaps

othercorey:
This also updates and comments a few remaining DateTime methods.


dictionaries/CallMap_80_delta.php
```diff
@@ -53,10 +53,6 @@
       'old' => ['int|false'],
       'new' => ['int'],
     ],
-    'DateTime::setTime' => [
```
othercorey, line 56: These never officially returned false. The return from an invalid parameter was always undefined behavior and PHP does not document these changes when parameter parsing throws instead of returning null/false.

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -476,12 +404,17 @@ class InternalCallMapHandlerTest extends TestCase
...
         'datetime::createfrominterface',
-        'datetime::setdate' => ['8.1', '8.2'],
-        'datetime::settimezone' => ['8.1', '8.2'],
-        'datetime::sub' => ['8.1', '8.2'],
+        'datetime::setdate' => ['8.1', '8.2'], // DateTime does not contain static
+        'datetime::setisodate' => ['8.1', '8.2'], // DateTime does not contain static
+        'datetime::settime' => ['8.1', '8.2'], // DateTime does not contain static
+        'datetime::settimestamp' => ['8.1', '8.2'], // DateTime does not contain static
+        'datetime::settimezone' => ['8.1', '8.2'], // DateTime does not contain static
+        'datetime::sub' => ['8.1', '8.2'], // DateTime does not contain static
```
othercorey, line 417: Hopefully can add a simple comment for each return type error to know why it's blocked.

orklah:
Thanks!

Issue #9348, opened 2023-02-20
assert(!array_is_list()) changes type to be maybe list


othercorey:
This is a strange issue that seems straight-forward to reproduce in psalm.dev, but does not trigger there.

```
ERROR: InvalidArgument - src/ORM/Table.php:2590:30 - Argument 1 of Cake\ORM\Query\SelectQuery::applyOptions expects array<string, mixed>, but array<int<0, max>|string, mixed> provided (see https://psalm.dev/004)
        $query->applyOptions($options);
```

A parameter of type `array<string, mixed` is passed to another parameter of the same type `array<string, mixed>`.

However, when introducing an assert that it is associative, it seems to think the array could now be a list.

Same code in psalm.dev does not throw an error: https://psalm.dev/r/c7b80db58a

The function starts with this code so not sure what else could be affecting it:
```php
    public function callFinder(string $type, SelectQuery $query, array $options = []): SelectQuery
    {
        assert(empty($options) || !array_is_list($options), 'Finder options should be an associative array not a list');

        $query->applyOptions($options);
```

weirdan:
@othercorey you omitted `@` in `@param`, that's why you couldn't reproduce it on psalm.dev. Here's the fixed reproducer: https://psalm.dev/r/715994eafd

othercorey:
Thanks. I updated the description with your example so it's not confusing to read.

PR #9468, opened 2023-03-07
Fix added null type after if

### LINKED ISSUE #9395: Null wrongly added as possible variable value
Psalm adds null as a possible value to a variable that never gets assigned null.
After the comparison `$a`s value according to psalm is `-1|null`
```php
<?php

$a = -1;
if ($a !== null) {
}
echo $a;
```
Obviously the  if block serves no purpose here. However the same problem occurred in a more complex function where an `int` variable was compared to an `int|null` variable.

https://psalm.dev/r/1502d48747

MoonE:
Fixes #9395, but *not* #9348

Maybe the condition needs to be refined according to this:
https://github.com/vimeo/psalm/blob/222887acfefe72f8cc212392ebef2b58f23f5d9e/src/Psalm/Internal/Type/AssertionReconciler.php#L478-L481

With a simple `else` instead of `elseif ($key !== '$this')` these three tests started to fail:
- ClassTemplateExtendsTest => inferPropertyTypeOnThisInstanceofExtended
- TraitTest => staticClassTraitUser
- TraitTest => getClassTraitUser

Some helpful comments for debugging: https://github.com/vimeo/psalm/commit/6638deb2536ee3f53397e8fcc717dba3e7b15c6d

As noted above, I'm very unsure about the correct condition needed here, but this seems to work with the current test base.

### DIFF
--- a/src/Psalm/Internal/Type/AssertionReconciler.php
+++ b/src/Psalm/Internal/Type/AssertionReconciler.php
@@ -47,6 +47,7 @@ use Psalm\Type\Atomic\TLiteralString;
 use Psalm\Type\Atomic\TLowercaseString;
 use Psalm\Type\Atomic\TMixed;
 use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Atomic\TNever;
 use Psalm\Type\Atomic\TNonEmptyLowercaseString;
 use Psalm\Type\Atomic\TNonEmptyString;
 use Psalm\Type\Atomic\TNull;
@@ -508,6 +509,8 @@ class AssertionReconciler extends Reconciler
 
             if ($intersection_type) {
                 $new_type = $intersection_type;
+            } elseif ($key !== '$this') {
+                $new_type = new Union([new TNever()]);
             }
         }
 
--- a/tests/TypeReconciliation/TypeTest.php
+++ b/tests/TypeReconciliation/TypeTest.php
@@ -16,6 +16,17 @@ class TypeTest extends TestCase
     public function providerValidCodeParse(): iterable
     {
         return [
+            'strictNotNullCheck' => [
+                'code' => '<?php
+                    $a = 1;
+
+                    /** @psalm-suppress RedundantCondition */
+                    if ($a !== null) {
+                    }',
+                'assertions' => [
+                    '$a===' => '1',
+                ],
+            ],
             'sealArray' => [
                 'code' => '<?php
                     /** @var array */


weirdan:
:thinking: 

At first glance, this fix looks dubious to me, especially the fact that it touches the `refine()` method. I suspect the correct fix would have something to do with either branch interruption logic or the distinction between *asserted* and *assigned* variables.

orklah:
It makes some sense though. The issue seems definitely linked to reconciliation to me, not control flow.

See snippet: https://psalm.dev/r/9432deb60a
Psalms sees the redundant condition but correctly infers that the type inside the if is -1. However, it still tries to reconcile -1 and null and somehow it results in null in the else branch. That's the error, this should be never because we can't enter in the else (at least when we have a type that's not from_docblock). Psalm then proceeds to reconcile the if and the else branch and at that point, it's bound to make -1|null incorrectly whereas -1|never would have found the correct -1 as a result

I'm not perfectly sure about the fix, especially the part where it exclude $this seems suspicious, but I think there's something correct in this

weirdan:
If no assignments have been made to the variable in the `if`/`else` block, then that `if`/`else` block couldn't have affected the variable type in the enclosing scope, and there's no point even trying to reconcile it. Even though assertions made in the `if` clause do affect the type in the `if` body and inverse assertions affect the type in the `else` branch. That's the difference between the asserted and assigned types I was talking about.

You can also say that assertions for `if` and `else` cancel each other (`null & !null => never`, and then reconciliation would result in the same type the variable had outside the `if`/`else`), but it doesn't seem to work all too good in practice, as not every assertion can be negated. Also, why reconcile assertions that we know should have cancelled each other?

weirdan:
> assertions for if and else cancel each other (`null & !null => never`)

On second thought, it should be `null | !null => mixed`.

orklah:
I don't know enough of condition reconciliation then. I thought we made a union between the branches in every case.

weirdan:
That's why we track assigned/redefined vars in `IfScope`. The distinction eludes me, though.

weirdan:
Psalm indeed seems to rely on the mutual cancellation because my 'fix' resulted in over 3.5K errors reported on Psalm's own codebase. 

But there's definitely also something wrong with the scope manipulation logic as well

```php
$a = 1;
if ($a !== null) {}
```
evaluates a hypothetical `else` with a negated set of assertions introduced by `if`. And then it treats the types that were generated by applying those assertions as if they were generated by an assignment in that hypothetical `else`, resulting in the type ending up in `$if_scope->possibly_redefined_variables`, even though there was no such `else` (and no assignments) at all. It does work fine for cases when the assertions in `if` and (hypothetical) `else` had opposite effects on the original type, but it breaks badly when one of the sides had no effect, like in these cases we have in the original issue and here.

Psalm also doesn't seem to treat hypothetical and real `else` constructs any differently; thus, the errors are similar.

weirdan:
I believe #9538 is a better fix for that specific issue.

PR #9472, opened 2023-03-08
fix: check for php v8.1.0 in ReturnTypeAnalyzer (rather than v8.10.0)

pereorga:


orklah:
seems like it makes this code fails:
```php
$generator = static function (): \Generator
{
    yield from [];
};
```

we should probably add a test with this

cc @kkmuffme , do you have an idea why this fails?

weirdan:
Can we declare `$analysis_php_version_id` as `int<5_06_00, 8_03_00>` to prevent such typos?

pereorga:
> Can we declare `$analysis_php_version_id` as `int<5_06_00, 8_03_00>` to prevent such typos?

Should it be 7_04_00? I don't see 5_06_00 mentions in the codebase. And composer.json says minimum version is 7.4.

(By the way, I'm seeing that in ProjectAnalyzer.php's `setPhpVersion()` there is a pattern allowing for 5.4 and 5.5 PHP versions. Mentioning it in case these can be removed.)

Ideally, I guess versions should be written in a single place, so they do not need to be updated on every PHP release, on dozens of places. Or maybe a test could be written to enforce these versions.

Going to the point, if I try to def try to declare this in Codebase.php:

```
    /** @var int<5_06_00, 8_03_00> */
    public $analysis_php_version_id = PHP_VERSION_ID;
```

I get:

```
int<5_06_00, 8_03_00> is not a valid type (Invalid type '5_06_00')
```

If I put:

```
    /** @var int<50600, 80300> */
    public $analysis_php_version_id = PHP_VERSION_ID;
```

I get:

```
ERROR: InvalidPropertyAssignmentValue - src/Psalm/Codebase.php:303:39 - $this->analysis_php_version_id with declared type 'int<50600, 80300>' cannot be assigned type 'int<1, max>' (see https://psalm.dev/145)
    public $analysis_php_version_id = PHP_VERSION_ID;
```

If I try to suppress InvalidPropertyAssignmentValue here, I get around 80 errors, like:

```
ERROR: MissingPropertyType - src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayCombineReturnTypeProvider.php:114:20 - Property Psalm\Codebase::$analysis_php_version_id does not have a declared type (see https://psalm.dev/045)
            return $statements_source->getCodebase()->analysis_php_version_id >= 8_00_00
```

I'm not familiar with the psalm codebase (nor in using psalm a lot, yet, to be honest), and I'm a bit unsure how to best address this (without specifying a declaration that include PHP versions, in many places)

And the same regarding the impact of this change, as unfortunately I don't fully follow the code.

weirdan:
BCC failure is ok I think.

src/Psalm/Codebase.php
```diff
@@ -299,8 +300,8 @@ final class Codebase
      */
     public $allow_backwards_incompatible_changes = true;
 
-    /** @var int */
-    public $analysis_php_version_id = PHP_VERSION_ID;
+    /** @var int<50400,80299> */
+    public $analysis_php_version_id;
```
pereorga, line 304: Off-topic, but should it be a space after the comma? Not sure phpcs supports that

PR #9474, opened 2023-03-09
Add missing array, float, int and bool types to callmaps

othercorey:
Although a callmap type might be "contained" by the reflected type, that does not mean it includes all of the expected types.

I'm sure there is a more comprehensive way to test this that includes other types like objects, but I haven't thought it through yet.

Disabled bool check because there are a lot of functions that return false we might not want. However, I added it to several functions that seemed useful.


tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -764,7 +764,7 @@ private function assertTypeValidity(ReflectionType $reflected, string $specified
         $callMapType = Type::parseString($specified);
 
         try {
-            $this->assertTrue(UnionTypeComparator::isContainedBy(self::$codebase, $callMapType, $expectedType), "{$msgPrefix} type '{$specified}' should be contained by reflected type '{$reflected}'");
+            $this->assertTrue(UnionTypeComparator::isContainedBy(self::$codebase, $callMapType, $expectedType, allow_float_int_equality: false), "{$msgPrefix} type '{$specified}' is not contained by reflected type '{$reflected}'");
```
othercorey, line 767: This only affected one function. We want an explicit check for callmap validity.

othercorey:
> Shepherd seems to be failing now that `get_cfg_var()` can return ~false on error~ an array.

I removed array from the return type and added to ignore return list with a comment.

weirdan (approved):

orklah:
Thanks!

PR #9479, opened 2023-03-10
`json_encode()` returns a non-empty-string with `JSON_THROW_ON_ERROR | JSON_UNESCAPED_UNICODE`

fluffycondor:
It's a common use case in our codebase, and I encounter it often in the wild when there's a lot of non-ASCII data is encoded.


fluffycondor:
Looks like the failed test is unrelated to the PR.

weirdan:
Thanks!

Issue #9485, opened 2023-03-10
Only first array item is used to validate list type


MoonE:
`list` (not ok, no warning):
https://psalm.dev/r/00e15a381c

`array` (ok, warning generated)
https://psalm.dev/r/6206fb1d17

Psalm should warn in both cases.

weirdan:
The key problem here is in `KeyedArrayComparator::isContainedBy()`; it ignores `$fallback_params` of the `$container_type_part`.

Another thing I'm not sure about is `list<false>` being represented as `list{false?,...<int<0,max>, false>}`. @danog is this expected?

PR #9487, opened 2023-03-11
Compare list generic params

weirdan:
So that `list<int>` does not accept `[1, new stdClass]`

Fixes vimeo/psalm#9485


weirdan:
PHPUnit failure is legit, `realpath()` can return `false` indeed.

PR #9491, opened 2023-03-12
Support int separators in docblocks

weirdan:
PHP understands `1_000` to be `1000`, so there's no reason why it shouldn't be recognized in docblocks as well.

Refs #9472

### DIFF
--- a/src/Psalm/Internal/Type/TypeParser.php
+++ b/src/Psalm/Internal/Type/TypeParser.php
@@ -91,6 +91,7 @@ use function stripslashes;
 use function strlen;
 use function strpos;
 use function strtolower;
+use function strtr;
 use function substr;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Type/TypeParser.php
```diff
@@ -420,8 +421,8 @@ public static function getTypeFromTree(
             return new TLiteralFloat((float) $parse_tree->value, $from_docblock);
         }
 
-        if (preg_match('/^\-?(0|[1-9][0-9]*)$/', $parse_tree->value)) {
-            return new TLiteralInt((int) $parse_tree->value, $from_docblock);
+        if (preg_match('/^\-?(0|[1-9]([0-9_]*[0-9])?)$/', $parse_tree->value)) {
```
orklah, line 424: do we want `int<0_000, 1_000>` to work? PHP allows it but I think this regex won't

weirdan: We did not allow leading zeroes, and PHP interprets them as octal numbers, which can get pretty confusing: https://3v4l.org/hprRC, so I don't think we want that.

orklah: Right, forgot about octals. Fine by me then!

PR #9492, opened 2023-03-12
Refuse to use PHP versions we did not request

weirdan:
There were a number of test failures caused by setup-php failing to
install the required PHP version. While we certainly should fix our
failing tests, we also need to be sure which PHP version we're using.

This PR does exactly that.


weirdan:
Example of a failed job can be found here: https://github.com/vimeo/psalm/actions/runs/4377883211/jobs/7661927364#step:3:20

PR #9497, opened 2023-03-12
Fixed #9496

### LINKED ISSUE #9496: @psalm-type  loses return type when declaring callable
https://psalm.dev/r/4a64c9c24d

This changed somewhere between 5.7.7 and 5.8.0. For 5.7.7, both, ``$callback1`` and ``$callback2`` are of type ``callable():int``.

ptomulik:
Work in progress (just created a unit test). No idea how to fix the bug. Can anyone help?

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
@@ -1905,9 +1905,12 @@ class ClassLikeNodeScanner
             }
 
             $type_string = str_replace("\n", '', implode('', $var_line_parts));
-
-            // Strip any remaining characters after the last grouping character >, } or )
-            $type_string = preg_replace('/(?<=[>})])[^>})]*$/', '', $type_string, 1);
+            try {
+                $type_string = CommentAnalyzer::splitDocLine($type_string)[0];
+            } catch (DocblockParseException $e) {
+                throw new DocblockParseException($type_string . ' is not a valid type: '.$e->getMessage());
+            }
+            $type_string = CommentAnalyzer::sanitizeDocblockType($type_string);
 
             try {
                 $type_tokens = TypeTokenizer::getFullyQualifiedTokens(
--- a/tests/TypeAnnotationTest.php
+++ b/tests/TypeAnnotationTest.php
@@ -743,6 +743,34 @@ class TypeAnnotationTest extends TestCase
                     class A {}
                     ',
             ],
+            'commentAfterType' => [
+                'code' => '<?php
+                    /**
+                     * @psalm-type TTest string
+                     *
+                     * This is a test class.
+                     */
+                    class Test {}',
+            ],
+            'multilineTypeWithComment' => [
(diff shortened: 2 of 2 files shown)


orklah:
This should be a regex change somewhere around https://github.com/vimeo/psalm/blob/40a05d574ae246a1a3da6de581566fff990d817c/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php#L1848

The previous PR added tests so as long as the CI pass, the fix should be correct

ptomulik:
> This should be a regex change somewhere around
> 
> https://github.com/vimeo/psalm/blob/40a05d574ae246a1a3da6de581566fff990d817c/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php#L1848
> 
> The previous PR added tests so as long as the CI pass, the fix should be correct

Sure? This was parsed correctly until https://github.com/vimeo/psalm/commit/c72edefc9e019e05f24894b1b5fa3ec0f822af32

orklah:
What I mean is that their fix broke something but their tests are legit. So as long as the next fix pass both tests, it will be an improvement :)

ptomulik:
> What I mean is that their fix broke something but their tests are legit. So as long as the next fix pass both tests, it will be an improvement :)

Only both? What about this one?: 

https://psalm.dev/r/dfc3177cdf

Looks like I just opened another Pandora's box :)

ptomulik:
Well, this is interesting. The current way of parsing type in @psalm-type is quite buggy. A bunch of examples below:

https://psalm.dev/r/e259fad4ec
https://psalm.dev/r/84df87cc84
https://psalm.dev/r/4379a14de4
https://psalm.dev/r/517fe131ac
https://psalm.dev/r/54ab72acce

pvandommelen:
https://github.com/vimeo/psalm/commit/6bbcfb242928b83e2bcf832e27f8cc82e447b966

This commit reuses the normal type parser logic (which is also used for `@param` etc) to find the part of the string that belongs to the type. It is more invasive, but it's much better than trying to improve on the regex solution.

It also fixes the issue with comments as found in https://github.com/vimeo/psalm/pull/9497#issuecomment-1465285981 :)

ptomulik:
> [6bbcfb2](https://github.com/vimeo/psalm/commit/6bbcfb242928b83e2bcf832e27f8cc82e447b966)
> 
> This commit reuses the normal type parser logic (which is also used for `@param` etc) to find the part of the string that belongs to the type. It is more invasive, but it's much better than trying to improve on the regex solution.
> 
> It also fixes the issue with comments as found in [#9497 (comment)](https://github.com/vimeo/psalm/pull/9497#issuecomment-1465285981) :)

@pvandommelen Thanks. I let myself to merge https://github.com/vimeo/psalm/commit/6bbcfb242928b83e2bcf832e27f8cc82e447b966 here. Looks like it fixed all the mentioned issues (I've also created additional unit tests).

PS. I have seen similar issues with `@psalm-template T of ....`, so probably similar solution there will be necessary.

orklah:
Thanks! that's great!

Issue #9498, opened 2023-03-12
Crash when '9223372036854775807' literal is used as an array key


MoonE:
https://psalm.dev/r/cbd73b78e5

MoonE:
> PHP Fatal error:  Uncaught Error: Cannot add element to the array as the next element is already occupied
thrown from php when executing this code
```php
<?php
$x = [PHP_INT_MAX => ''];
$x[] = '';
```

When `int_offset` would store the current offset (initialized with -1) instead of the next offset it would be possible to check if the array is already full and add a message.

```diff
diff --git a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
index 0c0e30fcc..b2a64ef6d 100644
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -333,11 +333,11 @@ class ArrayAnalyzer
                 } elseif ($key_type->isSingleIntLiteral()) {
                     $item_key_value = $key_type->getSingleIntLiteral()->value;
 
-                    if ($item_key_value >= $array_creation_info->int_offset) {
+                    if ($item_key_value > $array_creation_info->int_offset) {
                         if ($item_key_value === $array_creation_info->int_offset) {
                             $item_is_list_item = true;
                         }
-                        $array_creation_info->int_offset = $item_key_value + 1;
+                        $array_creation_info->int_offset = $item_key_value;
                     }
                 }
             } else {
@@ -345,7 +345,11 @@ class ArrayAnalyzer
             }
         } else {
             $item_is_list_item = true;
-            $item_key_value = $array_creation_info->int_offset++;
+            if ($array_creation_info->int_offset < PHP_INT_MAX) {
+                $item_key_value = ++$array_creation_info->int_offset;
+            } else {
+                // FIXME: Cannot insert array element, php does not allow it.
+            }
             $key_atomic_type = new TLiteralInt($item_key_value);
             $array_creation_info->item_key_atomic_types[] = $key_atomic_type;
             $key_type = new Union([$key_atomic_type]);
```

Code is duplicated in SimpleTypeInferer and ArrayAnalyzer and should be fixed there, too.

weirdan:
[...]

PR #9499, opened 2023-03-12
Fix integer overflow in array keys

weirdan:
Fixes vimeo/psalm#9498

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -333,7 +333,9 @@ class ArrayAnalyzer
                 } elseif ($key_type->isSingleIntLiteral()) {
                     $item_key_value = $key_type->getSingleIntLiteral()->value;
 
-                    if ($item_key_value >= $array_creation_info->int_offset) {
+                    if ($item_key_value < PHP_INT_MAX
+                        && $item_key_value >= $array_creation_info->int_offset
+                    ) {
                         if ($item_key_value === $array_creation_info->int_offset) {
                             $item_is_list_item = true;
                         }
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -333,21 +333,32 @@ class ArrayAnalyzer
                 } elseif ($key_type->isSingleIntLiteral()) {
                     $item_key_value = $key_type->getSingleIntLiteral()->value;
 
-                    if ($item_key_value < PHP_INT_MAX
-                        && $item_key_value >= $array_creation_info->int_offset
+                    if ($item_key_value <= PHP_INT_MAX
+                        && $item_key_value > $array_creation_info->int_offset
                     ) {
-                        if ($item_key_value === $array_creation_info->int_offset) {
+                        if ($item_key_value - 1 === $array_creation_info->int_offset) {
                             $item_is_list_item = true;
                         }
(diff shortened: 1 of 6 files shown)


tests/ConstantTest.php
```diff
@@ -1539,6 +1539,27 @@ class B
...
+                        public const SO = ['9223372036854775808' => 1];
+                    }
+                    $s = A::S;
+                    $i = A::I;
+                    $so = A::SO;
+                    PHP,
+                'assertions' => [
+                    '$s===' => 'array{9223372036854775807: 1}',
+                    '$i===' => 'array{9223372036854775807: 1}',
+                    '$so===' => "array{'9223372036854775808': 1}",
```
weirdan, line 1560: This is an example of a numeric that PHP keeps as a string.

src/Psalm/Type/Atomic/TKeyedArray.php
```diff
@@ -663,8 +663,22 @@ public function getList(): TList
...
+
+            if ($name === '' || preg_match('/[^a-zA-Z0-9_]/', $name)) {
+                $quote = true;
+            }
+
+            if (preg_match('/^[1-9][0-9]*$/', $name)
+                && (string)(int) $name !== $name // overflow occured
+            ) {
+                $quote = true;
+            }
```
weirdan, line 677: This is to distinguish strings that look like integers, but won't be converted to integers by PHP (e.g. any that is greater than `PHP_INT_MAX`);

MoonE: What about numbers smaller than PHP_INT_MIN?
```suggestion
            if (preg_match('/^-?[1-9][0-9]*$/', $name)
                && (string)(int) $name !== $name // overflow occured
            ) {
                $quote = true;
            }
```

Am I understanding this correctly that quoting done here is only for users to distinguish numeric-string keys from ints?

MoonE:
I don't think this will work with 
```php
$a = [
  9223372036854775807 => 0,
  1,
];
```
With my suggestion to change the stored value in int_offset from the next  possible to the current value it should work.

weirdan:
> I don't think this will work with

It doesn't crash with that input, but it still crashes with 
```php
<?php
$a = [
  9223372036854775806 => 0,
  1,
];
```

weirdan:
> With my suggestion to change the stored value in int_offset from the next possible to the current value it should work.

That was a good suggestion, thank you.

src/Psalm/Type/Atomic/TKeyedArray.php
weirdan: > Am I understanding this correctly that quoting done here is only for users to distinguish numeric-string keys from ints?

Correct.

src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
```diff
@@ -333,19 +333,32 @@ private static function analyzeArrayItem(
                 } elseif ($key_type->isSingleIntLiteral()) {
                     $item_key_value = $key_type->getSingleIntLiteral()->value;
 
-                    if ($item_key_value >= $array_creation_info->int_offset) {
-                        if ($item_key_value === $array_creation_info->int_offset) {
+                    if ($item_key_value <= PHP_INT_MAX
```
orklah, line 336: Ugh, I hate to think this could ever be false. Technically, TLiteralInt::$value is public and not typed so it could be a float higher than Max Int

I feel we should do something about that so this case would be handled upstream instead...

weirdan: I guess we can make it typed (in a separate PR) and see what happens.

PR #9504, opened 2023-03-13
Fix return type of str_split()

othercorey:
The stubs do not include the `false` return type from PHP 7. Instead, it defines the $length parameter as `positive-int`. If that parameter is not `positive-int`, then `false` would be return. It's hard to define both conditions in stubs, so aligning with the callmap.

In PHP 8.0, the false return type was dropped for a ValueError.

In PHP 8.2, the function was changed to return an empty array if the string is empty so the return type is no longer non-empty, but the strings are now non-empty.

https://3v4l.org/mHMvb

https://github.com/php/php-src/blob/PHP-8.2.0/UPGRADING#L51

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -12898,7 +12898,7 @@ return [
 'str_replace' => ['string|string[]', 'search'=>'string|array', 'replace'=>'string|array', 'subject'=>'string|array', '&w_count='=>'int'],
 'str_rot13' => ['string', 'string'=>'string'],
 'str_shuffle' => ['string', 'string'=>'string'],
-'str_split' => ['list<string>', 'string'=>'string', 'length='=>'positive-int'],
+'str_split' => ['list<non-empty-string>', 'string'=>'string', 'length='=>'positive-int'],
 'str_starts_with' => ['bool', 'haystack'=>'string', 'needle'=>'string'],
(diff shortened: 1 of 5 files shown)


othercorey:
Does psalm prefer to keep return type edge cases or parameter edge cases?

weirdan:
> Does psalm prefer to keep return type edge cases or parameter edge cases where one controls the other?

IIRC, we approach it like this: if a function only returns a particular type (usually `null` or `false`) when invalid arguments are passed (and we do flag those invalid arguments), then checking for that type should not be required. In practical terms, we keep `false` and `null` in callmaps because all functions from callmaps (with some deliberate exceptions) are automatically flagged as `ignore-falsable-return` and `ignore-nullable-return`. For stubs, we also keep `null` and/or `false` return type, and add `@psalm-ignore-falsable-return`/`@psalm-ignore-nullable-return` annotations (https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-ignore-nullable-return).

othercorey:
I think that works. 

The reason I wasn't sure about the stub is if we assert the parameter should be `positive-int` then you can't do a conditional return since the parameter should always be valid. 

I wasn't sure about the callmap since it also requires a `positive-int` parameter.

othercorey:
Removed the warning suppression to fix shepherd.

weirdan (approved):

weirdan:
Thanks!

PR #9513, opened 2023-03-14
Fix mysqli fetch call map for associative arrays

MoonE:
these functions will return integer keys when the column name is an integer.


dictionaries/CallMap.php
```diff
@@ -7878,12 +7878,12 @@
 'mysqli_execute' => ['bool', 'statement'=>'mysqli_stmt', 'params='=>'list<mixed>|null'],
 'mysqli_execute_query' => ['mysqli_result|bool', 'mysql'=>'mysqli', 'query'=>'non-empty-string', 'params='=>'list<mixed>|null'],
 'mysqli_fetch_all' => ['list<array<array-key,null|int|float|string>>', 'result'=>'mysqli_result', 'mode='=>'3'],
-'mysqli_fetch_all\'1' => ['list<array<string,null|int|float|string>>', 'result'=>'mysqli_result', 'mode='=>'1'],
```
othercorey, line 7881: I think this can just be `array<...>` instead of having to use array-key.

MoonE: Wasn't sure if its better to be explicit here, and there were a handful of other entries already using array-key.

Both should work.

othercorey: Up to the maintainers, but I think the distinction is use array-key outside of an array but it's not necessary in an array.

weirdan:
> when the column name is an integer.

How do you create such a column?

MoonE:
You can have pretty much anything as a column name when you quote it.
```sql
CREATE TABLE test_numeric_column2 (`1234` INT);
```

weirdan:
So it's more about PHP's inability to properly store `["1234" => whatever]`. We ignore this issue in other places (https://psalm.dev/r/33bc693967) so it would be surprising to account for this in the callmap.

PR #9522, opened 2023-03-15
Remove always-empty ClassAnalyzer leftover_stmts array

edsrzf:
Happened to notice this as I was reading through the code.


orklah:
Good find!

PR #9525, opened 2023-03-16
Improve json_encode/json_decode functions signatures

fluffycondor:
I just realized there's an easy way to add support for every flags combination with JSON_THROW_ON_ERROR.
Also it adds a constrant for `$depth` according to [specs](https://www.php.net/manual/en/function.json-decode.php#:~:text=The%20value%20must%20be%20greater%20than%200%2C%20and%20less%20than%20or%20equal%20to%202147483647).

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1468,11 +1468,7 @@ function json_decode(string $json, ?bool $associative = null, int $depth = 512, 
 /**
  * @psalm-pure
  *
- * @return (
- *      $flags is 4194304
- *          ? non-empty-string
- *          : ($flags is 4194560 ? non-empty-string : non-empty-string|false)
- *  )
+ * @return ($flags is int<4194304, 8388607> ? non-empty-string : non-empty-string|false)
  *
  * @psalm-flow ($value) -> return
  * @psalm-ignore-falsable-return
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1460,6 +1460,7 @@ function ldap_escape(string $value, string $ignore = "", int $flags = 0) : strin
 /**
  * @psalm-pure
  *
+ * @param int<1, 2147483647> $depth
  * @return mixed
  * @psalm-flow ($json) -> return
  */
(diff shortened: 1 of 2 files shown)


tests/CoreStubsTest.php
```diff
@@ -131,5 +133,35 @@ function foo(string $foo): string
...
+                $a = json_encode([], JSON_THROW_ON_ERROR | JSON_HEX_TAG);
+                $b = json_encode([], JSON_THROW_ON_ERROR | JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP | JSON_UNESCAPED_UNICODE);
+                $c = json_encode([], JSON_THROW_ON_ERROR | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
+                $d = json_encode([], JSON_THROW_ON_ERROR | JSON_PRESERVE_ZERO_FRACTION);
+            ',
+            'assertions' => [
+                '$a===' => 'non-empty-string',
+                '$b===' => 'non-empty-string',
+                '$c===' => 'non-empty-string',
+                '$d===' => 'non-empty-string',
```
weirdan, line 147: Do we have any tests for `non-empty-string|false` variant?

stubs/CoreGenericFunctions.phpstub
```diff
@@ -1468,11 +1469,8 @@ function json_decode(string $json, ?bool $associative = null, int $depth = 512,
 /**
```
weirdan, line 1469:
```suggestion
/**
 * The conditional return type below relies on the fact that JSON_THROW_ON_ERROR is 
 * the highest-valued of JSON constants
```

weirdan (changes requested):

tests/CoreStubsTest.php
fluffycondor: Nice catch. Will add some.
Actually there is also `JSON_PARTIAL_OUTPUT_ON_ERROR` that narrows the type to `non-empty-string`, but adding support for it requires adding 10 condition branches for `int<512,1023>`, `int<1536,2047>` and so on 😅, so I'm not really sure is it worth it.

weirdan (approved):

orklah:
Thanks!

PR #9526, opened 2023-03-16
Clean up unreflectable callmap methods

othercorey:
This removes methods that were removed before PHP 7 and fixes some missing methods.


tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -378,31 +365,17 @@ class InternalCallMapHandlerTest extends TestCase
         'pdo::sqlitecreateaggregate',
         'pdo::sqlitecreatecollation',
         'pdo::sqlitecreatefunction',
-        'pdostatement::__sleep',
-        'pdostatement::__wakeup',
         'simplexmlelement::__get',
         'simplexmlelement::offsetexists',
         'simplexmlelement::offsetget',
         'simplexmlelement::offsetset',
         'simplexmlelement::offsetunset',
```
othercorey, line 372: SimpleXmlElement implements array access even though it's not documented. If we keep offset functions will that help psalm accept array access on the object?

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -378,31 +365,17 @@ class InternalCallMapHandlerTest extends TestCase
         'pdo::sqlitecreateaggregate',
         'pdo::sqlitecreatecollation',
         'pdo::sqlitecreatefunction',
-        'pdostatement::__sleep',
-        'pdostatement::__wakeup',
         'simplexmlelement::__get',
```
othercorey, line 368: SimpleXmlElement also implement property access without defining a __get() function. I don't know if psalm needs this to allow arbitrary property access.

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -378,31 +365,17 @@ class InternalCallMapHandlerTest extends TestCase
...
         'simplexmlelement::offsetget',
         'simplexmlelement::offsetset',
         'simplexmlelement::offsetunset',
         'spldoublylinkedlist::__construct',
-        'splfileinfo::__wakeup',
-        'splfixedarray::current',
-        'splfixedarray::key',
-        'splfixedarray::next',
-        'splfixedarray::rewind',
-        'splfixedarray::valid',
```
othercorey, line 394: SplFixedArray now implements IteratorAggregate.

tests/Internal/Codebase/InternalCallMapHandlerTest.php
```diff
@@ -378,31 +365,17 @@ class InternalCallMapHandlerTest extends TestCase
...
         'splmaxheap::__construct',
         'splobjectstorage::__construct',
         'splpriorityqueue::__construct',
         'splstack::__construct',
-        'weakmap::__construct',
-        'weakmap::current',
-        'weakmap::key',
-        'weakmap::next',
-        'weakmap::rewind',
-        'weakmap::valid',
```
othercorey, line 405: No idea where these came from. WeakMap was added in PHP 8. I added the correct methods.

othercorey:
@orklah I think this clean up is fairly simple. Do you see any issues?

tests/Internal/Codebase/InternalCallMapHandlerTest.php
orklah: Yeah, though it would be great to also update the stub

tests/Internal/Codebase/InternalCallMapHandlerTest.php
orklah: Psalm does indeed checks for __get/__set method existance at some point

orklah:
Thanks!

tests/Internal/Codebase/InternalCallMapHandlerTest.php
othercorey: Looks like SimpleXmlElement now implements RecursiveIterator in PHP 8. I think need to fix how stubs are loaded for target php version first.

https://3v4l.org/bNNkh#v8.0.28
https://3v4l.org/bNNkh#v7.4.33

PR #9528, opened 2023-03-16
Fix MissingThrowsDocblock when documented interface is extended by thrown exception interface

paoloconi96:
When an interface exception is documented via the `@throws` tag and a called method has a documented interface exception which extends the parent's one Psalm reports a `MissingThrowsDocblock`.

Example: https://psalm.dev/r/db257bfea4

With this PR I'm adding a check that ensures a better validation for interfaces extensions.


orklah:
Thanks!

orklah:
Congrats for your first PR!

Issue #9196, opened 2023-01-29
Incorrect `InvalidLiteralArgument` when usage `dirname()`


vjik:
```php
function test(): bool {
    $path = dirname(__DIR__, 4);
    return strlen($path) > 6 && str_contains($path, 'vendor');
}
```

For no duplicate `dirname(__DIR__, 4)` I use variable `$path`, but psalm throw `InvalidLiteralArgument`. Seems, it's incorrect.

https://psalm.dev/r/3603178ff0

weirdan:
Simplified: https://psalm.dev/r/760dffc723

PR #9534, opened 2023-03-17
Improve PHP8 str_* function signatures

### LINKED ISSUE #8378: str_starts_with(), str_ends_with() don't assert string as non-empty-string
https://psalm.dev/r/1c49ff9da6

True result on non-empty $needle should lead to asserting the $haystack as non-empty-string.

Probably `str_*_with()` functions should be extracted from ClassMap and introduced into stub files. That's just my assumption though.

fluffycondor:
Closes #8378, #9196
It helps detecting some redundant conditions and also asserts $haystack to `non-empty-string`.

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1059,18 +1059,27 @@ function str_split(string $string, int $length = 1) {}
 
 /**
  * @psalm-pure
+ * @template T as string
+ * @param T $needle
+ * @psalm-assert-if-true =(T is '' ? string : non-empty-string) $haystack
  * @return ($needle is '' ? true : ($haystack is '' ? false : bool))
  */
 function str_starts_with(string $haystack, string $needle): bool {}
 
 /**
  * @psalm-pure
+ * @template T as string
+ * @param T $needle
+ * @psalm-assert-if-true =(T is '' ? string : non-empty-string) $haystack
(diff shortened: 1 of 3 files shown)


stubs/CoreGenericFunctions.phpstub
```diff
@@ -1057,6 +1057,33 @@ function str_shuffle(string $string): string {}
  */
 function str_split(string $string, int $length = 1) {}
 
+/**
+ * @psalm-pure
+ * @template T as string
+ * @param T $needle
+ * @psalm-assert-if-true =(T is '' ? string : non-empty-string) $haystack
```
fluffycondor, line 1064: Actually this should be written as
`@psalm-assert-if-true (T is '' ? string : =non-empty-string) $haystack`
but it explodes psalm parser.
But it's a really edge case when this difference matters. I don't think there would be any issues against any real-world code.

fluffycondor:
Looks like the failed unit test is unrelated to the PR. But some weird stuff happened there 🤔

fluffycondor:
Looks like I'm missing something.
It was intended this PR makes `non-empty-string` out of `$string` here:
````php
  $string = file_get_contents("");
  if (!str_ends_with($string, ".html")) {
      throw new RuntimeException();
  }
 /** @psalm-trace $string */; // non-empty-string expected, but got string|false
````
Should I remove `=` from asserts? 🤔

orklah:
> Is it worth to open an issue?..

yeah probably!

Thanks for the PR!

PR #9537, opened 2023-03-17
Fix long2ip return type

kamil-tekiela:


dictionaries/CallMap.php
```diff
@@ -6513,7 +6513,7 @@
 'LogicException::getPrevious' => ['?Throwable'],
 'LogicException::getTrace' => ['list<array{file?:string,line?:int,function:string,class?:class-string,type?:\'::\'|\'->\',args?:array<mixed>}>'],
 'LogicException::getTraceAsString' => ['string'],
-'long2ip' => ['string', 'ip'=>'int'],
+'long2ip' => ['string|false', 'ip'=>'int'],
```
weirdan, line 6516: I'm not sure if this can actually ever happen. 

Looking at the source for `long2ip()`[1] it may appear as if it would return `false` if `inet_ntop` failed. But the only documented failure conditions[2] seem to be either wrong address family (can't happen as it's hardcoded) or insufficient memory (also can't happen (?) as it's stack-allocated, and has enough memory to hold the longest string representation of an IPv6 address, even though only IPv4 addresses are supported by this function). 

[1]: https://github.com/php/php-src/blob/60fffd296abce5fc071f3c173c25a2696cf683c6/ext/standard/basic_functions.c#L3974-L4003
[2]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/inet_ntop.html

jorgsowa: I checked the code and IMO you are right. I have created PR for this change: https://github.com/php/php-src/pull/13395

jorgsowa: The change has been merged. Will be available in the next PHP version.

weirdan:
Formally, the return type will change in PHP.next (?), but since it doesn't seem `long2ip()` could ever return `false` I'm going to close this.

PR #9547, opened 2023-03-21
[mysql] Add stubs for `mysqli`, `mysqli_result` and `mysqli_stmt` classes

phansys:


stubs/extensions/mysqli.phpstub
```diff
@@ -18,6 +31,44 @@ class mysqli_result implements Traversable
     function fetch_object(string $class = stdClass::class, array $constructor_args = []): object|false|null {}
 }
 
+lass mysqli_stmt
```
weirdan, line 34: Should this have been `class`?

stubs/extensions/mysqli.phpstub
```diff
@@ -1,12 +1,25 @@
...
  * @template TValue
  *
  * @template-implements Traversable<int, TValue>
  */
 class mysqli_result implements Traversable
 {
+    /**
+     * @var int<0, max>|numeric-string
+     */
+    public int|string $num_rows;
```
weirdan, line 21: Is this to account for the limited int range on 32-bit platforms? While the 64-bit integer range is still limited, it's large enough to ignore possible overflow, in my opinion.

phansys: I don't know if in practice the limit from [`PHP_MAX_INT`](https://www.php.net/manual/en/reserved.constants.php#constant.php-int-max) in 64-bit platforms (9223372036854775807) can be reached, but docs don't make any distinction in this matter:
>If the number of affected rows is greater than maximum PHP int value, the number of affected rows will be returned as a string value.

stubs/extensions/mysqli.phpstub
phansys: :man_facepalming:

Fixed, thank you.

PR #9555, opened 2023-03-22
Add better return type to mysqli fetch_field*

MoonE:
These methods all return an object with the same properties. Some as a list, some just a single object.

https://www.php.net/manual/en/mysqli-result.fetch-field.php


orklah:
Thanks!

PR #9556, opened 2023-03-22
Do not produce invalid array-key types with array_column

MoonE:
When using the third parameter of array_column, only scalar types and null are valid, and those types need to be converted to string/int.

https://psalm.dev/r/5a2ca24ef5


MoonE:
I'm not sure why foreach and array_keys would be relevant here. They don't convert an arbitrary value to an array key like array_column does.

Maybe you are right and it makes more sense to solve it at array creation. I did find these two places that do something similar already:
https://github.com/vimeo/psalm/blob/b908f4491f32b28002e8b84ce61093735b031905/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ArrayFetchAnalyzer.php#L817-L863
https://github.com/vimeo/psalm/blob/b908f4491f32b28002e8b84ce61093735b031905/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php#L179-L220

orklah:
What I mean is foreach will also take whatever type is inside the array offset and put it in a variable. If the offset is a float, you'll end up with a $key with a float value

yeah, I had the example you posted in mind, there must be an other place where it should be executed (and probably refactored in a single place...)

Issue #9305, opened 2023-02-15
Imported type alias as template parameter inside union type of the same classes does not work


provokateurin:
This is a really hard edge case.
The main problem seems to be that the type alias is imported, since it works fine if it is not imported. It also works when two different wrapper classes are used.
There could be more issues though, because only this specific constellation is broken.

https://psalm.dev/r/cf65a18261

weirdan:
The effects are easier to see with `@param`: https://psalm.dev/r/5e863cdb32

provokateurin:
wow thank you so much for fixing this bug so fast!

PR #9562, opened 2023-03-23
always combine the result of expansions

orklah:
This fixes https://github.com/vimeo/psalm/issues/9305

What happened is that we had two array types in $new_return_type_parts at the end of the loop. However, an Union can only take a single array type without having to combine them

There was a specific handling for `split_values` but I couldn't get why Psalm thinks TTypeAlias or a single type that is extended into two need to be combined but two types would not need to be combined.

This may result in cases where performance are degraded but I'm not sure how to check the need to make a combination without reimplementing half the checks made in TypeCombiner...

### DIFF
--- a/src/Psalm/Internal/Type/TypeExpander.php
+++ b/src/Psalm/Internal/Type/TypeExpander.php
@@ -77,8 +77,6 @@ class TypeExpander
     ): Union {
         $new_return_type_parts = [];
 
-        $had_split_values = false;
-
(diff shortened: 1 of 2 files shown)


weirdan (approved):

provokateurin:
May I ask when this will be in a release? I don't need it right now, just some rough timeline when the next release is expected would be nice.

orklah:
released :)

provokateurin:
Thanks so much! :heart:

PR #9566, opened 2023-03-24
Handle different line endings from baseline.

### LINKED ISSUE #9239: Harmonise line endings in baseline for multi-line code snippets
Hello,
This is bug very similar to #5979, but contrary to this initial issue, my problems comes when the `code` element of the baseline contains a multi-line snippet. The fix was to `trim` the code content (that was relocated as part of the 5.x upgrade, but the idea remains the same), but that would remove only the characters at the beginning or the end of the string, but do not cater for potential carriage returns and line feeds inside the string.
So, depending on whether CRLF or LF mode is used, when the baseline is generated and/or compared, there may be respectively additional or missing `&#13;` characters in the baseline file, that means that errors are not recognised as part of the baseline if the environment do not use the same line ending. In my case, my dev environment is Windows, using CRLF, where the baseline is produced and I can get the "No errors found!" result, whilst the GitHub actions use a Linux environment, and LF mode, and fail (on a message not clear at all, as it seems identical to the baseline).

I feel that the in the line below, there should be an harmonisation of the line endings, by maybe dropping the `&#13;` entity if present, so that CRLF will be harmonised to LF. Of course, this is not the only change that would be required, and that may 
need to wait Psalm 6, or have an alternative solution, as this is maybe a breaking change.
https://github.com/vimeo/psalm/blob/4eacb2f234867d34d8489f904c2445aecb9158d0/src/Psalm/ErrorBaseline.php#L121

I have tried to showcase it in the [small example attached.](https://github.com/vimeo/psalm/files/10678807/test-bug.zip) (hopefully, it will make sense).

Using Psalm 5.6.0, I have created the same code twice, once in a file with CRLF endings, one with LF endings, to simulate the two environments. The anonymous function will of course raise an error, as an `int` is expected rather than `bool`.

```php
<?php
$test = ['b', 'a'];
usort($test, function (mixed $a, mixed $b): bool {
    return $a <=> $b;
});
print_r($test);
```
Taking the CRLF as reference for the baseline, it generates the below (which I manually duplicate for `test.LF.php`). Please note the `&#13;` entities to encode the carriage return; the line feed is not encoded, and is a line break in the file.

```xml
<files psalm-version="5.6.0@e784128902dfe01d489c4123d69918a9f3c1eac5">
  <file src="test.CRLF.php">
    <InvalidArgument>
      <code>function (mixed $a, mixed $b): bool {&#13;
    return $a &lt;=&gt; $b;&#13;
}</code>
    </InvalidArgument>
    ...
  </file>
</files>
```

Then, I run for the CRLF environment:

<details>
 <summary>Running on CRLF environment</summary>

```
$ psalm --version
Psalm 5.6.0@e784128902dfe01d489c4123d69918a9f3c1eac5

$ psalm --config=psalm.CRLF.xml
Warning: "findUnusedCode" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.
Target PHP version: 8.1 (inferred from current PHP version) Enabled extensions: .
Scanning files...
Analyzing files...

░

       No errors found!

3 other issues found.
You can display them with --show-info=true
Psalm can automatically fix 1 of these issues.
Run Psalm again with
--alter --issues=InvalidReturnType --dry-run
to see what it can fix.

Checks took 1.63 seconds and used 111.493MB of memory
Psalm was able to infer types for 100% of the codebase
```
</details>

And for the LF environment, using the same baseline file:

<details>
 <summary>Running on LF environment</summary>

```

$ psalm --config=psalm.LF.xml
Warning: "findUnusedCode" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.
Target PHP version: 8.1 (inferred from current PHP version) Enabled extensions: .
Scanning files...
Analyzing files...

░

ERROR: InvalidArgument - test.LF.php:5:14 - Argument 2 of usort expects callable(mixed, mixed):int, but pure-Closure(mixed, mixed):bool provided (see https://psalm.dev/004)
usort($test, function (mixed $a, mixed $b): bool {
    return $a <=> $b;
});


1 errors found
2 other issues found.
You can display them with --show-info=true
Psalm can automatically fix 1 of these issues.
Run Psalm again with
--alter --issues=InvalidReturnType --dry-run
to see what it can fix.

Checks took 1.64 seconds and used 111.494MB of memory
Psalm was able to infer types for 100% of the codebase
```
</details>

The LR environment throws an error (with a "displayed" code identical to the one in the baseline, but the line endings are actually different), whereas the CRLF completes fine. I would expect both environments to complete successfully with the same baseline, as the code is seemingly the same.

DnwK98:
This change allows to have different file encoding in baseline and PHP file. This is issue when working in multi-platform team, generating baseline on Linux, and trying to run code analysis on Windows.

Fixes #9239

### DIFF
--- a/src/Psalm/ErrorBaseline.php
+++ b/src/Psalm/ErrorBaseline.php
@@ -118,7 +118,7 @@ final class ErrorBaseline
 
                 foreach ($codeSamples as $codeSample) {
                     $files[$fileName][$issueType]['o'] += 1;
-                    $files[$fileName][$issueType]['s'][] = trim($codeSample->textContent);
+                    $files[$fileName][$issueType]['s'][] = str_replace("\r\n", "\n", trim($codeSample->textContent));
                 }
 
                 // TODO: Remove in Psalm 6
--- a/src/Psalm/Internal/LanguageServer/LanguageServer.php
+++ b/src/Psalm/Internal/LanguageServer/LanguageServer.php
@@ -782,7 +782,7 @@ class LanguageServer extends Dispatcher
                             if ($issue_baseline[$file][$type]['o'] === count($issue_baseline[$file][$type]['s'])) {
                                 /** @psalm-suppress MixedArrayAccess, MixedAssignment */
                                 $position = array_search(
-                                    trim($issue_data->selected_text),
+                                    str_replace("\r\n", "\n", trim($issue_data->selected_text)),
                                     $issue_baseline[$file][$type]['s'],
                                     true,
                                 );
--- a/src/Psalm/IssueBuffer.php
+++ b/src/Psalm/IssueBuffer.php
@@ -592,7 +592,7 @@ final class IssueBuffer
                         if (isset($issue_baseline[$file][$type]) && $issue_baseline[$file][$type]['o'] > 0) {
                             if ($issue_baseline[$file][$type]['o'] === count($issue_baseline[$file][$type]['s'])) {
                                 $position = array_search(
-                                    trim($issue_data->selected_text),
+                                    str_replace("\r\n", "\n", trim($issue_data->selected_text)),
                                     $issue_baseline[$file][$type]['s'],
                                     true,
                                 );
--- a/tests/ErrorBaselineTest.php
+++ b/tests/ErrorBaselineTest.php
@@ -98,6 +98,37 @@ class ErrorBaselineTest extends TestCase
         );
     }
 
+    public function testShouldIgnoreCarriageReturnInMultilineSnippets(): void
+    {
+        $baselineFilePath = 'baseline.xml';
+
+        $this->fileProvider->allows()->fileExists($baselineFilePath)->andReturns(true);
+        $this->fileProvider->allows()->getContents($baselineFilePath)->andReturns(
+            "<?xml version=\"1.0\" encoding=\"UTF-8\"?>
+            <files>
+              <file src=\"sample/sample-file.php\">
+                <MixedAssignment>
+                  <code>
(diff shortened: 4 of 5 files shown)


weirdan:
Do you think this would fix #9239?

DnwK98:
Yes, I reproduced it.

baseline:
```
  <file src="src/Psalm/Config/ProjectFileFilter.php">
    <DocblockTypeContradiction>
      <code>-1</code>
    </DocblockTypeContradiction>
    <InvalidReturnStatement>
      <code>$filter&#13;
            ?? -1&#13;</code>
    </InvalidReturnStatement>
```

code (LF encoded):
```
    public static function loadFromXMLElement(
        SimpleXMLElement $e,
        string $base_dir,
        bool $inclusive
    ): ProjectFileFilter {
        ...

        return $filter
            ?? -1;
    }
```

Before change:
```
ERROR: InvalidReturnStatement - src/Psalm/Config/ProjectFileFilter.php:35:16 - The inferred type '-1|Psalm\Config\ProjectFileFilter' does not match the declared return type 'Psalm\Config\ProjectFileFilter' for Psalm\Config\ProjectFileFilter::loadFromXMLElement (see https://psalm.dev/128)
        return $filter
            ?? -1;


1 errors found
```

After change:
```
                              
       No errors found!       
                              
```

weirdan (changes requested):
Please add a test showing that baselines generated with older Psalm versions (so, including `\r\n`) would still work once this gets merged.

orklah:
Thanks!

PR #9568, opened 2023-03-25
@property annotations: allow *not* implying @psalm-seal-properties

danielbeardsley:
I was going to open an issue for something that felt like a bug, but I figured a pull request adding a config option may be better received.

---

Add a setting that allows usage of `@property` to *augment* classes that use `__get()` and `__set()`. Previously, using `@property` once would force you to exhaustively list all possible properties. This didn't use to be the case, but was changed in df33405635cbdb6d41474503ca20c8adda55d438

This was really unexpected for our team and for a while we thought it was a psalm bug until I found the above commit.

We are using `__get()` for column access on ORM objects and we want to use `@property` to explicitly document the types of some of the columns without being forced to document every column.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -195,7 +195,7 @@ class Config
     public $use_docblock_property_types = false;
 
     /**
-     * Whether using @property in docblocks should implicitly seal properties
+     * Whether using property annotations in docblocks should implicitly seal properties
      *
      * @var bool
      */
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -195,6 +195,13 @@ class Config
     public $use_docblock_property_types = false;
 
     /**
+     * Whether using @property in docblocks should implicitly seal properties
+     *
+     * @var bool
+     */
+    public $docblock_property_types_seal_properties = true;
+
+    /**
      * Whether or not to throw an exception on first error
      *
      * @var bool
@@ -1049,6 +1056,7 @@ class Config
         $booleanAttributes = [
             'useDocblockTypes' => 'use_docblock_types',
             'useDocblockPropertyTypes' => 'use_docblock_property_types',
+            'docblockPropertyTypesSealProperties' => 'docblock_property_types_seal_properties',
             'throwExceptionOnError' => 'throw_exception',
             'hideExternalErrors' => 'hide_external_errors',
             'hideAllErrorsExceptPassedFiles' => 'hide_all_errors_except_passed_files',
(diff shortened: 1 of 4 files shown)


src/Psalm/Config.php
```diff
@@ -359,6 +366,11 @@ class Config
      */
     public $seal_all_properties = false;
 
+    /**
+     * @var bool
+     */
+    public $seal_all_properties = false;
+
```
weirdan, line 373: It's already declared few lines above

weirdan (changes requested):

src/Psalm/Config.php
danielbeardsley: oops, I accidentally duplicated this when I was trying to decide where to put the new property. Fixing.

danielbeardsley: Fixed. Thanks!

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
```diff
@@ -568,7 +568,9 @@ public function start(PhpParser\Node\Stmt\ClassLike $node): ?bool
                     }
                 }
 
-                $storage->sealed_properties = true;
+                if ($config->docblock_property_types_seal_properties) {
```
weirdan, line 571:
```suggestion
                if ($this->config->docblock_property_types_seal_properties) {
```

weirdan (changes requested):

src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
danielbeardsley: Wow...  Typos when moving fast I guess :-) Thanks.

src/Psalm/Config.php
```diff
@@ -194,6 +194,13 @@ class Config
      */
     public $use_docblock_property_types = false;
 
+    /**
+     * Whether using @property in docblocks should implicitly seal properties
```
orklah, line 198: Psalm seem to parse `@property` in your docblock as a property. Maybe replace that with `using the property annotation should...`

danielbeardsley: Haha... Ok, makes sense.

```suggestion
     * Whether using property annotations in docblocks should implicitly seal properties
```

orklah:
Sorry, didn't see it before, there's two things left to do:
- add that config in config.xsd so the schema can be validated
- document the default value in Psalm's documentation
Otherwise, ready to merge :)

danielbeardsley:
> there's two things left to do:

Derp. Sorry to drag this out. Fixed.

weirdan (approved):

orklah:
Thanks!

PR #9586, opened 2023-03-31
Introduce `value-of` with backed enum cases in assertions

boesing:
Fixes assertion inconsistencies as described in here:

https://psalm.dev/r/8e09e2bfb4

### DIFF
--- a/src/Psalm/Internal/Codebase/ClassConstantByWildcardResolver.php
+++ b/src/Psalm/Internal/Codebase/ClassConstantByWildcardResolver.php
@@ -9,25 +9,23 @@ use Psalm\Type\Atomic;
 use Psalm\Type\Atomic\TMixed;
 
 use function array_merge;
-use function array_values;
-use function preg_match;
-use function sprintf;
-use function str_replace;
 
 /**
  * @internal
  */
 final class ClassConstantByWildcardResolver
 {
+    private StorageByPatternResolver $resolver;
     private Codebase $codebase;
 
     public function __construct(Codebase $codebase)
     {
+        $this->resolver = new StorageByPatternResolver();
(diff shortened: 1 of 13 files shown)


boesing:
PR is ready for review, GHA just failed installing composer in windows unit test 1 chunk.

orklah:
Thanks <3!

weirdan:
@boesing it looks like the feature wasn't implemented as it was intended. Can you confirm if the following were your intentions:
1. `@psalm-assert AEnum::A|AEnum::B $param` should **not** be the same as `@psalm-assert value-of<AEnum::A|AEnum::B> $param` (because otherwise `value-of` would be useless here)
2. `@psalm-assert value-of<AEnum::A|AEnum::B> $param` should be the same as `@psalm-assert value-of<AEnum::A>|value-of<AEnum::B> $param`
3. `value-of<AEnum::A>` should produce the backing type (e.g. `string` or `int`, or a literal string or int) and not the case type (essentially an object of type `AEnum`)

weirdan:
Snippet for the above list: https://psalm.dev/r/0eb3427541

boesing:
@weirdan As already mentioned in #10082, that was ofc not intended and I am aware that value-of should result in a different result than the values without.

Seems that there were no test to verify the behavior and thus, it was dropped. I can see when I can find some time to fix that, but earliest would be this weekend.

boesing:
Yes, thats how it is tested in the tests provided by this PR.
Seems there weren't any tests for the initial enum implementation and thus, that got most probably lost.
The only thing I did not tested explicitly was `@psalm-assert value-of<AEnum::A>|value-of<AEnum::B> $param` but I guess that would work the same way.

PR #9599, opened 2023-04-03
Introduce callable object intersection

boesing:


src/Psalm/Internal/Type/Comparator/ObjectComparator.php
```diff
@@ -134,7 +137,7 @@ public static function isShallowlyContainedBy(
 
     /**
      * @param  TNamedObject|TTemplateParam|TIterable  $type_part
-     * @return array<string, TNamedObject|TTemplateParam|TIterable|TObjectWithProperties>
+     * @return array<string, TNamedObject|TTemplateParam|TIterable|TObjectWithProperties|TCallableObject>
```
boesing, line 140: Not 100% sure about this, as this might be a BC break.

weirdan: Changes in internal classes (like this) are not considered BC breaks.

tests/Traits/InvalidCodeAnalysisTestTrait.php
```diff
@@ -16,17 +16,26 @@
...
 use const PHP_VERSION;
 
+/**
+ * @psalm-type DeprecatedDataProviderArrayNotation = array{
+ *     code: string,
+ *     error_message: string,
+ *     ignored_issues?: list<string>,
+ *     php_version?: string
+ * }
+ * @psalm-type NamedArgumentsDataProviderArrayNotation = array{
```
boesing, line 26: not 100% sure why, but I had issues with psalm not properly suppressing my stuff. Might be due to the fact that somewhere named arguments kicked in (due to the fact that I might've executed tests with PHP 8+ and thus had issues).
So for now, I've allowed both here, might be refactored when tests are executed on PHP 8 at some point.

boesing:
Ready for review, all tests finally green.

orklah:
Could you explain why TUnknownClassString is needed?

boesing:
Mainly for BC reason. The class is not final and used in other cases which might be more tough. But I am open for additions

boesing:
I'd like to modify this PR a bit due to ondrejs Feedback: https://github.com/phpstan/phpstan/issues/9148

src/Psalm/Internal/Type/Comparator/ObjectComparator.php
```diff
@@ -268,6 +271,8 @@ private static function isIntersectionShallowlyContainedBy(
             $intersection_input_type_lower = 'iterable';
         } elseif ($intersection_input_type instanceof TObjectWithProperties) {
             $intersection_input_type_lower = 'object';
+        } elseif ($intersection_input_type instanceof TCallableOBject) {
```
orklah, line 274: Oh, I'm surprised Psalm didn't complain about a case mismatch on TCallableOBject

orklah:
Seems good! Can you fix the case issue and this can be merged

src/Psalm/Internal/Type/Comparator/ObjectComparator.php
boesing: Fixed.

orklah:
Thanks!

PR #9608, opened 2023-04-04
fix: treat includes starting with '.' and '..' correct

### LINKED ISSUE #9605: Include path starting with '.' or '..'
Problem: 
PSALM gives false positive errors of "cannot find file" if an include/require starts with '.' or '..'.

PSALM obviously uses the base path to resolve this as an relative path, but indeed PHP handles those paths in a special way, see documentation snippet below.

this was also stated in https://github.com/vimeo/psalm/issues/268#issuecomment-342651032, but was not handled further.

from https://www.php.net/manual/en/function.include.php
> If a path is defined — whether absolute (starting with a drive letter or \ on Windows, or / on Unix/Linux systems) or relative to the current directory (starting with . or ..) — the [include_path](https://www.php.net/manual/en/ini.core.php#ini.include-path) will be ignored altogether. For example, if a filename begins with ../, the parser will look in the parent directory to find the requested file.

jjjb03:
this fixes #9605 for includes with non compound strings, eg.:
```php
include './path/to/file.php';
```

but it does not work for compound expressions, eg: 
```php
include '.' . '/path/to/file.php';
```

Compound expressions are evaluated within `IncludeAnalyzer::getPathTo(...)` and are directly returned as absolute path, so there is no chance to detect the need of this special treatment.

The function is used in several different places, and I am not familar enough with this code to do the refactoring, sorry.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/IncludeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/IncludeAnalyzer.php
@@ -395,11 +395,9 @@ class IncludeAnalyzer
             return $file_name;
         }
 
-        if (
-            (substr($file_name, 0, 2) === '.' . DIRECTORY_SEPARATOR)
+        if ((substr($file_name, 0, 2) === '.' . DIRECTORY_SEPARATOR)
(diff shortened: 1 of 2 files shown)


orklah:
Thanks!

PR #9623, opened 2023-04-07
Contextual inference for const callable

klimick:
Same as #9570 but for `const`.
Also I've extracted previously added functionality to separate class.


klimick:
I don't know what to do with failed test `templateStaticWithParam` in the `ClassTemplateTest`.

Looks whats Psalm currently infers: https://psalm.dev/r/b926080ed9
I don't think that is what should be.

klimick:
I tried to fix test but now it fails with another error:
```
Argument 1 of takesInts expects ArrayCollection<int>, but ArrayCollection<int<0, max>>&static provided
```
Similar error on psalm.dev: https://psalm.dev/r/7a6c14737b
I think it's a old variance issue. My PR did not add this issue.

klimick:
I've changed `ArrayCollection<int>` to `ArrayCollection<int<0, max>>` and it fixed test.
CI fails at `test-with-real-projects`. But it's unrelated troubles with deprecations.

orklah:
Thanks! Sorry for not helping you more, I'm a little out of my league on PR you submit :D

PR #9629, opened 2023-04-09
Do not report required closure arguments as unused

### LINKED ISSUE #8095: `UnusedClosureParam` for required closure arguments
When I have a method which declares a callback to consume 2 arguments but my defined callback does not care about the first argument, psalm is issuing `UnusedClosureParam` even tho, its not like I could omit it.

I stated this question already a few months back and the suggestion was to use something like `$_`. 
Sadly, from my perspective, this is:
1. less descriptive of what the value would contain (e.g. in case it has to be used in a future feature)
2. mostly violates styleguides (i.e. when camel cased variables should be used)
3. can not be used more than once (i.e. when one has 1 or more required arguments which are not used)

---

I would like to propose a last-minute feature to the next major which allows closures to consume unused parameters when they are required due to its order.

## Disallowed
https://psalm.dev/r/c2b33c5cae

## Allowed
https://psalm.dev/r/a6201b4d4a

---

Is there a strong veto against this behavior?

boesing:
Fixes #8095

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
@@ -72,6 +72,7 @@ use function count;
 use function end;
 use function in_array;
 use function is_string;
+use function krsort;
 use function mb_strpos;
 use function md5;
 use function microtime;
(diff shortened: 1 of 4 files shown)


orklah:
Thanks!

PR #9630, opened 2023-04-10
Preserve `non-empty-string` type when combining literal and numeric strings

### LINKED ISSUE #8787: `numeric-string` in return statement is not always considered as `non-empty-string`
Hey there,

I would expect that this is valid.
https://psalm.dev/r/f3c31d26b4

I played a bit and it seems that `numeric-string` is internally considered as `non-empty-string` in some cases.

Works: https://psalm.dev/r/0b89d54eda
Does not work: https://psalm.dev/r/9af2a31713
But this works: https://psalm.dev/r/d2c8893799

So it seems that somewhere is an inconsistency for `numeric-string` being interpreted as `non-empty-string`.
Any ideas where to tackle this?

boesing:
Fixes #8787


orklah:
Thanks!

PR #9632, opened 2023-04-10
Fix various db callmaps

othercorey:
Not sure if we should ignore oci callmaps. I'm not setup to test them and not sure if the subs are reflectable.


orklah:
Thanks! Ignoring oci is fine, if someone is interested, they'll contribute and remove it from the ignore list :)

Issue #4193, opened 2020-09-16
[Question] How to annotate "insert element" operation on object-like array


doggor:
I have a function that accept an object-like array and insert a new element into it. The keys of the input array are varied, and the name of inserted key and type of the inserted value are known. So the type of the output should be intersection of something like `TemplateOfTheObjectLikeArray & array{'known': string}`.

https://psalm.dev/r/ef60918b5a

I found a similar issue https://github.com/vimeo/psalm/issues/3339 that describes a more general situation, but don't know if there is a solution for this case.

weirdan:
At least the message is wrong, as `TKeyedArray` is what previously was called object-like array.

doggor:
> At least the message is wrong, as `TKeyedArray` is what previously was called object-like array.

Message updated for readability.

weirdan:
I was referring to the
> ERROR: InvalidDocblock - 10:1 - Intersection types must be all objects or all **object-like arrays**, **Psalm\Type\Atomic\TKeyedArray** provided in docblock for addBar

It's self-contradicting, as `TKeyedArray` is object-like array.

orklah:
The message was changed, so this is no longer a bug, rather a enhancement

PR #9638, opened 2023-04-11
Allow to intersect type alias with non-type-aliases

boesing:


boesing:
hm, @orklah @weirdan I've modified the code so that the `TTypeAlias` is now resolved early on. Due to that, the intersections in the `TTypeAlias` are not used anymore and thus psalm detects that as unused code.
Any suggestions on how to proceed?

orklah:
Suppress the error and add an @deprecated tag explaining this will be removed in Psalm 6 :)

boesing:
I am unsure if merging all keyed arrays into one dedicated one is a good solution, any feedback on this?

Is it necessary to provide the information that the keyed array consists of 1 or more array shapes and how their shapes were before getting merged?

boesing:
Circle CI real projects check might need some love, phpunit has deprecated methods.

orklah:
Seems good!

Yeah, CircleCI is pretty fragile and can break when there are changes upstream (which is often!). I need to look into it when I find some time!

orklah:
Note: I wouldn't have expected that error on the snippet shown in https://github.com/vimeo/psalm/issues/4193

boesing:
Created a fix for #4193 in #9667

PR #9646, opened 2023-04-12
fix: add missing constructor parameter to IteratorIterator

SenseException:
The second parameter exist since PHP 5.1 and wasn't present in the docs before. Even reflection wasn't showing this parameter until 8.0.

I was wondering about the reason for the existence of this parameter and with the help of @choult and [an old blog post](https://blog.ircmaxell.com/2011/10/iteratoriterator-php-inconsistencies.html) the reason for the `$class` parameter was made clear.


orklah:
Thanks!

PR #9647, opened 2023-04-12
Flatten `match` arm conditions to check conditions independently

### LINKED ISSUE #9633: ArgumentTypeCoercion with multiple conditional expressions in `match`
Should be no error here:

https://psalm.dev/r/e6cd62e225

tuqqu:
Flatten match statement arms to one condition per arm to check conditions independently.

Fixes #9633


orklah:
Thanks!

PR #9655, opened 2023-04-14
Provide support for templated `value-of` enum values

### LINKED ISSUE #9145: `value-of<MyEnum>` does not work when used as template param in `use` statement
Hey there, 

I wanted to implement a trait which provides a `values` method to `BackedEnum` implementations without our project.

https://psalm.dev/r/86d5f5d471

But sadly, it is not possible to provide `value-of<self>` within a `use` annotation. 
Am I missing something or is that a bug?
If the latter, having a point to start would be awesome and I might find some time to fix it.

boesing:
Fixes #9145


orklah:
Thanks!

PR #9656, opened 2023-04-15
(re-)implement object-shape assertions

### LINKED ISSUE #8981: Object shapes in assertions do not work anymore
Hey there,

I've recently upgrade psalm to v5 in another project where we use object shapes to determine properties from `json_decode` objects. So we have an assertion method which then should tell psalm that there are properties with specific types.

This does not properly seem to work, at least in my example here:
https://psalm.dev/r/eb1bed0bf9

Is that something which is related to psalm v5 or am I oversee something?

boesing:
Fixes #8981


orklah:
Thanks a lot for your PRs! This is great!

PR #9658, opened 2023-04-15
Fix list<T> template replacement

klimick:
Currently `T` from `list<T>` incorrectly replaces from literal lists: https://psalm.dev/r/310b315ccc
Psalm infers `list<1>` instead `list<1|2|3>`. i.e. it takes first literal type and drops rest.
This PR fixes it.


orklah:
Thanks!

PR #9663, opened 2023-04-17
Specify array share for `stream_context_get_params` return value

lptn:
Based on https://www.php.net/manual/en/function.stream-context-get-params.php docs, `#[ArrayShape(["notification" => "string", "options" => "array"])]` PHPStorm's attribute


orklah:
Thanks!

PR #9667, opened 2023-04-18
Recursively resolve template parameters

### LINKED ISSUE #4193: [Question] How to annotate "insert element" operation on object-like array

boesing:
Fixes #4193

### DIFF
--- a/src/Psalm/Internal/Type/TypeParser.php
+++ b/src/Psalm/Internal/Type/TypeParser.php
@@ -1116,23 +1116,11 @@ class TypeParser
             return new TMixed();
         }
 
+        $intersection_types = self::resolveTemplatedTypes($intersection_types);
+        $onlyTKeyedArray = self::validateAllKeyedArray($intersection_types);
+
         $first_type = reset($intersection_types);
         $last_type = end($intersection_types);
-
-        $onlyTKeyedArray = $first_type instanceof TKeyedArray
-            || $last_type instanceof TKeyedArray;
-
-        foreach ($intersection_types as $intersection_type) {
-            if (!$intersection_type instanceof TKeyedArray
-                && ($intersection_type !== $first_type
-                    || !$first_type instanceof TArray)
-                && ($intersection_type !== $last_type
-                    || !$last_type instanceof TArray)
-            ) {
-                $onlyTKeyedArray = false;
-                break;
-            }
-        }
 
         if ($onlyTKeyedArray) {
             /**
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Type/TypeParser.php
```diff
@@ -1692,4 +1691,50 @@ private static function getTypeFromKeyedArrays(
...
+            $resolved_types[] = [$type];
+        }
+
+        $resolved_types = array_merge(...$resolved_types);
+        assert($resolved_types !== []);
+        return $resolved_types;
+    }
+
+    /**
+     * @param non-empty-array<Atomic> $types
```
boesing, line 1718: I've tried to add `@psalm-assert-if-true` so that we can provide `non-empty-array<TKeyedArray>` to get rid of the inline `@var` annotations above but that sadly didn't work but I dont have time to improve this for now.

src/Psalm/Internal/Type/TypeParser.php
```diff
@@ -1566,14 +1560,19 @@ private static function extractKeyedIntersectionTypes(
...
-            );
+        if ($any_object_type_found) {
+            if ($any_array_found) {
+                throw new TypeParseTreeException(
+                    'Intersection types must be all objects or all keyed array.',
+                );
+            }
+
+            if ($keyed_intersection_types === []) {
+                return ['object' => new TObjectWithProperties([], [], [], $object_from_docblock)];
```
boesing, line 1571: - [ ] Seems that this actually leads to the issuebuffer test issue.
will see how to improve that later this week

tests/Template/ClassTemplateTest.php
```diff
@@ -10,11 +10,47 @@
 
 class ClassTemplateTest extends TestCase
 {
-    use InvalidCodeAnalysisTestTrait;
+//    use InvalidCodeAnalysisTestTrait;
```
boesing, line 13: - [ ] fix me

tests/Template/ClassTemplateTest.php
```diff
@@ -10,11 +10,47 @@
...
     use ValidCodeAnalysisTestTrait;
 
     public function providerValidCodeParse(): iterable
     {
+        return [
```
boesing, line 18: - [ ] fix me

PR #9670, opened 2023-04-18
Avoid verbose opcache logging

danog:
Just a small fix to avoid cluttering the opcache log while running psalm if verbose logging is enabled.


orklah:
Thanks, can you fix the CS issues and this should be good to go

danog:
Fixed!

orklah:
Thanks!

Issue #6482, opened 2021-09-14
Psalm does not know $nonEmptyString[0] is also a non-empty-string


simPod:
https://psalm.dev/r/8ef12885ae

Accessing first character of `non-empty-string` should yield a `non-empty-string` as well.

orklah:
You can fix that in ArrayFetchAnalyzer::handleArrayAccessOnString::2043

We just confirmed that the array offset used is valid given the variable type (which, for a non-empty-string, would be potentially any integer) and we're trying to see what type should be returned. The current type returned is a TSingleLetter (which behaves pretty much exactly like TString). In your case, you should check if the "array" type is a non-empty-string an return that instead.

orklah:
This was fixed, probably by an extended stub for explode

PR #9675, opened 2023-04-19
Valid array access on a non-empty-string yields a non-empty-string

robchett:
Fix for https://github.com/vimeo/psalm/issues/6482

Unsure if TSingleLetter provides more info than TNonEmptyString as it is technically a TNonEmptyString of size 1.
 TSingleLetter isn't exposed to userland (I don't think) so should be providing more information than is available currently.

### DIFF
--- a/src/Psalm/Type/Atomic/TSingleLetter.php
+++ b/src/Psalm/Type/Atomic/TSingleLetter.php
@@ -7,6 +7,6 @@ namespace Psalm\Type\Atomic;
  *
  * @psalm-immutable
  */
-final class TSingleLetter extends TString
+final class TSingleLetter extends TNonEmptyString
 {
 }
--- a/tests/ArrayAccessTest.php
+++ b/tests/ArrayAccessTest.php
@@ -524,6 +524,15 @@ class ArrayAccessTest extends TestCase
                         }
                     }',
             ],
+            'nonEmptyStringAccess' => [
+                'code' => '<?php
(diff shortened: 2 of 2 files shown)


orklah:
Thanks for the PR!

We don't have size tracking on TNonEmptyString so TSingleLetter could still provide more information. 

The snippet you added in tests is not currently failing: https://psalm.dev/r/7b2a4579da 

So I'm not sure what you tried to fix. In fact, it seems #6482 was fixed in an other way. The original issue was that explode could still return false with the snippet, but it doesn't anymore: https://psalm.dev/r/1ff0613ad1?phpVersion=7.1

Can you give more information about what issue you were facing?

robchett:
Honestly, I was just trying to help with an issue marked as `Good first issue` + `Help wanted` because I love Psalm and wanted to pay a little back.

When I ran the original snippet from #6482 and took the issue to be that `$delimiter` was a string not a non-empty-string in the trace. You are right however that the return type is correctly picked up as a non-empty-string. Because TSingleLetter is by definition a TNonEmptyString? 

Here `baz` is reporting what looks like an error in the trace (string doesn't fulfil non-empty-string)  but is passing correctly? https://psalm.dev/r/8d8391db91

Is this because TSingleLetter is traced as a string when it could be better traced as non-empty-string?

orklah:
Can you fix the code style issue? Otherwise, seems good like that :)

orklah:
Thanks!

PR #9676, opened 2023-04-19
Fix comparison of unions of multiple callables

danog:


tests/TypeComparatorTest.php
```diff
@@ -159,6 +159,10 @@ public function getSuccessfulComparisons(): array
                 'lowercase-string',
                 'callable-string',
             ],
+            'callableUnionAcceptsCallableUnion' => [
+                'callable(int,string[]): void|callable(int): void',
```
orklah, line 163: can you add parenthesis please? It breaks my internal parser :D

tests/ArgTest.php
```diff
@@ -805,7 +805,7 @@ function caller($callback) {}
                     function foo($a) {}
 
                     caller("foo");',
-                'error_message' => 'InvalidScalarArgument',
+                'error_message' => 'PossiblyInvalidArgument',
```
orklah, line 808: Not sure why this could be considered possibly invalid

src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php
```diff
@@ -175,7 +175,7 @@ public static function isContainedBy(
                     if ($container_required_param_count > $input_all_param_count
                         || $container_all_param_count < $input_required_param_count
                     ) {
-                        return false;
+                        continue;
```
orklah, line 178: Seems fundamentally wrong to me. You basically removed the only thing in the loop that made it not useless (if I'm not clear, I think you could remove the loop altogether and get the same effect)

I think you need to inverse the logic of the loop. Instead of failing (returning false) at the first issue, you need to return false if and only if every part of the loop is wrong

orklah (changes requested):

src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php
danog: Hmmm, this logic seems correct to me, because we're basically doing

```
foreach ($input_type as $input_type_part) {
    $type_match_found = false;
    foreach ($container_type as $container_type_part) {
        if (<container_type_part contains input_type_part>) {
            $type_match_found = true;
        }
    }
    if (!$type_match_found) {
        return false;
    }
}

return true;
```

There's no place for a `return false` in the second foreach, since we're only acting on the `$type_match_found` variable to signal whether a match was found or not.

tests/ArgTest.php
danog: This seems to be caused by a call to `UnionTypeComparator::canBeContainedBy`, which doesn't have the same callable parameter comparison logic that `UnionTypeComparator::isContainedBy` has.

Actually, that logic should probably be moved to the `CallableTypeComparator`...

danog: Ah I see now, it's to allow things like https://psalm.dev/r/3cc38f5a0c

src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php
orklah: Sorry, I misread where the loop ended. My previous remark don't make any sense

danog: Yeah the loop is huge, it was a bit hard for me too :P

orklah:
Thanks!

Issue #6531, opened 2021-09-24
mb_strtolower has wrong return type


snapshotpl:
https://psalm.dev/r/042270ea20

snapshotpl:
When changed execution order in `strtolower` also infern wrong type https://psalm.dev/r/7a469a04d7

orklah:
The first part should be as easy as adding a stub for mb_strtolower, like there is already for strtolower: https://github.com/vimeo/psalm/blob/master/stubs/CoreGenericFunctions.phpstub#L508

The second part is a little more complex. It involve adding a method in AssertionFinder::scrapeEqualityAssertions named getEmptyEqualityAssertions (which would be the opposite of getEmptyInequalityAssertions) to find cases where we do `=== ''`. When we find them, we should add an `empty` assertion to the variable so that in SimpleAssertionReconciler::reconcile we could add a reconcileEmpty that would check for existing known TString or TLowercaseString to transform them into TNonEmptyString and TNonEmptyLowercaseString (that would also solve this: https://psalm.dev/r/d9332678bc)

fluffycondor:
> like there is already for strtolower: https://github.com/vimeo/psalm/blob/master/stubs/CoreGenericFunctions.phpstub#L508

I'm not quite sure it's correct to assume strtolower returns lowercase-string. It's a lowercase-ascii-string if to be exact. `strtolower('École');` will return `École`. It can be confusing to have something like that in a var docblocked as lowercase-string.

orklah:
Well, Psalm seems to considers `École` to already be in lowercase so it's coherent: https://psalm.dev/r/649f4508ac

weirdan:
Fixed in #9678

PR #9678, opened 2023-04-19
Return a non-empty-(lowercase-)string from mb_strtolower

robchett:
Fixes the remaining issue on #6531

I'm unsure if adding isNonEmptyString to src/Psalm/Type/UnionTrait.php was thecorrect way to get the non-empty state.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
@@ -482,15 +482,25 @@ class FunctionCallReturnTypeFetcher
 
                     return $call_map_return_type;
                 case 'mb_strtolower':
+                    $string_arg_type = $statements_analyzer->node_data->getType($call_args[0]->value);
+                    if ($string_arg_type !== null && $string_arg_type->isNonEmptyString()) {
+                        $returnType = Type::getNonEmptyLowercaseString();
+                    } else {
+                        $returnType = Type::getLowercaseString();
+                    }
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php
```diff
@@ -482,15 +483,25 @@ private static function getReturnTypeFromCallMapWithArgs(
 
                     return $call_map_return_type;
                 case 'mb_strtolower':
+                    $string_arg_type = $statements_analyzer->node_data->getType($call_args[0]->value);
+                    if ($string_arg_type->isNonEmptyString()) {
```
orklah, line 487: You need to check that the type was correctly inferred (!== null)

src/Psalm/Type/UnionTrait.php
```diff
@@ -1016,6 +1017,25 @@ public function isString(bool $check_templates = false): bool
...
+     * @psalm-mutation-free
+     * @return bool true if this is a string
+     */
+    public function isNonEmptyString(bool $check_templates = false): bool
+    {
+        return count(
+            array_filter(
+                $this->types,
+                static fn($type): bool => $type instanceof TNonEmptyString
+                    || ($type instanceof TLiteralString && $type->value)
```
orklah, line 1030: The way you checked, you inferred a non-falsy-string: '0' is non empty but it will return false

You need to explicitly check `!== ''` to check for non-emptyness

robchett: D'oh, well spotted. Now updated.

orklah:
Thanks!

Issue #6752, opened 2021-10-27
Unable to determine the type that $this->value is being assigned to


KevinVanSonsbeek:
I ran into this problem when working with an abstract class.
https://psalm.dev/r/7c3665dc1e

The constructor parameter needs to be mixed, as the extending classes will need to decide (through the template), what type it's going to be. But when i try to add a var annotation, i get an UnnecessaryVarAnnotation.
https://psalm.dev/r/8b76b19f48

PR #9679, opened 2023-04-19
Don't throw UnnecesseryVarAnnotation when hinting a mixed template var

robchett:
Fix for #6752, the snippet https://psalm.dev/r/8b76b19f48 no longer throws a warning.

Feels wrong to require the annotation but I guess there is no way of specifying `@tempate T of not-mixed`

### DIFF
--- a/src/Psalm/Type/UnionTrait.php
+++ b/src/Psalm/Type/UnionTrait.php
@@ -52,6 +52,8 @@ use function reset;
 use function sort;
 use function strpos;
 
+use const ARRAY_FILTER_USE_BOTH;
+
 /**
  * @psalm-immutable
  * @psalm-import-type TProperties from Union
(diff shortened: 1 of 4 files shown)


src/Psalm/Type/UnionTrait.php
```diff
@@ -795,9 +797,20 @@ public function hasMixed(): bool
...
      * @psalm-mutation-free
      */
-    public function isMixed(): bool
+    public function isMixed(bool $check_templates = false): bool
     {
-        return isset($this->types['mixed']) && count($this->types) === 1;
+        return count(
+            array_filter(
+                $this->types,
+                static fn($type, $key): bool => $key == 'mixed'
```
orklah, line 805: Can you use === please? I see it's not used everywhere, but IMHO it should be :p

orklah:
nicely fixed :)

orklah:
Thanks!

Issue #2588, opened 2020-01-10
Support no-seal-properties/no-seal-methods on classes to avoid emitting UndefinedMagicMethod 


TysonAndre:
Some use cases, such as test frameworks, may rely on supporting unknown magic methods, while still having a few of their own magic methods.

It'd be useful if those classes could indicate psalm shouldn't emit UndefinedMagicMethod for any of their use cases.

`@psalm-no-seal-methods` would be one way to do this. (or `@no-seal-methods`

https://github.com/ifwe/phockito/blob/bd1849468daf9539e052a3af9d11d981008397a8/Phockito.php#L617-L626 is one example of this - it supports both of the following syntaxes for the returned value of Phockito::when(...) (return() has a known type, someMethod() doesn't).

```php
Phockito::when($mockObj)->someMethod()->return($value);
Phockito::when($mockObj->someMethod())->return($value);
```

jarstelfox:
This just came up for us using [Faker](https://github.com/fzaninotto/Faker)

Faker allows for a plugin-based approach by:

* [Allowing plugins to be added](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Generator.php#L203-L206)
* [Plugins fall through __get](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Generator.php#L272-L275)
* [Eventually asking the plugins](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Generator.php#L226-L249)

Note: The library has [@method annotations](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Generator.php#L5-L197) to define the behaviour of the "built-in" plugins

Due to https://github.com/vimeo/psalm/commit/df33405635cbdb6d41474503ca20c8adda55d438 this third party library is becoming sealed.

Which means psalm complains about the methods defined in any plugins.

We cannot unseal the library without editing the source.

edudobay:
I have the same problem with queue-interop, or specifically [amqp-interop](https://github.com/queue-interop/amqp-interop/blob/master/src/AmqpContext.php). In the base Context, **createMessage** returns a Message, but AmqpContext narrows the return type to AmqpMessage using a `@method` annotation. Psalm doesn't take that into account and I need to add a `@var` annotation if I need AmqpMessage-specific methods.

Are those kinds of overrides by design not meant to work with Psalm?

[...]

PR #9681, opened 2023-04-20
Add support for @psalm-no-seal-properties and @psalm-no-seal-methods

robchett:
Implements #2588, 

I don't like using `public ?bool $sealed_properties = null;` where

- `true` is explicitly or inherited @seal-properties
- `false` is explicitly or inherited @no-seal-properties
- `null` is unspecified so use the configuration setting

Also unsure what the best practice is for accessing `Config::getInstance()`


orklah:
Thanks!

PR #9682, opened 2023-04-20
Add stub for DateInterval

robchett:
Unsure if a stub must define all functions for that class or only the ones with additional annotations?
Also are properties read from the stubs or should I remove them? Thought it would be useful to mark them as @readonly

### DIFF
--- a/stubs/CoreGenericClasses.phpstub
+++ b/stubs/CoreGenericClasses.phpstub
@@ -531,3 +531,93 @@ final class ReturnTypeWillChange
     public function __construct() {}
 }
 
+class DateInterval
+{
+    /**
+     * Number of years
+     * @var int
+     * @readonly
+     */
+    public $y;
+
+    /**
+     * Number of months
+     * @var int
+     * @readonly
+     */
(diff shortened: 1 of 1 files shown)


robchett:
@orklah can you advise why adding the stub would have changed the return of `DateInterval::createFromDateString(...)` to `DateInterval|false` when it would seem from the tests it was `DateInterval` before? `createFromDateString` isn't included in the stub. 
https://www.php.net/manual/en/dateinterval.createfromdatestring.php indicates that `false` has always been an option if I'm reading it correctly.

orklah:
Yeah :)

Before adding your stub, the best definition we had was in the Callmaps. Every definition in the Callmaps is assumed to be internal to PHP and we automatically flag each method that can returns false as "@psalm-ignore-falsable-return" and each method that can return null as `@psalm-ignore-nullable-return`, that way, Psalm will only raise an issue when the user explicitly said he wants to handle those. However, it's not done automatically in stubs so you need to do it yourself.

orklah:
Thanks!

PR #9685, opened 2023-04-20
Avoid wrong assertions when working with objects returned by methods

danog:
The added unit test fails otherwise.


orklah:
Wow, that was a weird bug

I'm not completely sure I understood your fix but it seems to me you're checking that the property that's asserted is on $this->. Wouldn't it be simpler to check that the property and the method must share a hierarchy before applying the assertion?

danog:
Mmm, the issue here is that if:
1) A method has an assertion on a property of the current object (so `$this->something`)
2) and that method is called on an object returned by another method, or by something else we can't make assertions about due to impurity, instead of using the usual logic of replacing the `$this->` part of the assertion with `$var->` , since there is no `$var->` but rather a `$this->someOtherMethod()->`, we do nothing (see the logic in `getUntemplatedCopy`).

This caused the assertion to be applied to the `$this` of the *caller* context, which is absolutely incorrect.

This patch essentially skips all method assertions if the method was called on an object we can't make assertions about, instead of wrongly applying the assertions to the caller context.

orklah:
Thanks!

PR #9686, opened 2023-04-20
Fix is_object assertions on final classes

danog:
They were broken by https://github.com/vimeo/psalm/pull/9656/files, this should fix them.

Even with this fix, not too sure the type reconciliation logic added in #9656 is entirely correct...


orklah:
Let's see if we find more issue in the future.

Anyway, we should probably follow phpstan and handle object shapes the same way as array shapes in the long run(note that they consider their object shape sealed unless there is an intersection with stdClass or another universal object crate)

orklah:
Thanks!

Issue #1450, opened 2019-03-11
Allow limiting who can inherit/implement a class


muglug:
```php
/**
 * @psalm-inheritors FooClass|BarClass
 */
class BaseClass {}

class FooClass extends BaseClass {
  public function thing(string $s) : void { return $s . "hello"; }
}

class BarClass extends BaseClass {
  public function thing(int $i) : string { return $i . "hello"; }
}

class BazClass extends BaseClass {} // this is an error

function f1(BaseClass $c) : string {
  switch (get_class($c)) {
    case FooClass:class:
      return $c->thing("me");
    case BarClass:class:
      return $c->thing(5);
  }
}

function f2(BaseClass $c) : string { // this is an error as not all paths are met
  switch (get_class($c)) {
    case FooClass:class:
      return $c->thing("me");
  }
}
```

orklah:
This would be a pain to implement for a very niche usage I think.

I'll close this btw, it has been implemented

PR #9687, opened 2023-04-20
WIP - Limit inheritance to a subset of classes #1450

robchett:
Going a little out of my depth with this one but giving it a shot. 

Only covering the first section, not the match cases.

### DIFF
--- a/docs/running_psalm/issues/InheritorViolation.md
+++ b/docs/running_psalm/issues/InheritorViolation.md
@@ -1,0 +1,17 @@
+# InheritorViolation
+
+Emitted when a class/interface using `@psalm-inheritors` is extended/implemented
+by a class that does not fulfil it's requirements.
+
+```php
+<?php
+
+/**
+ * @psalm-inheritors FooClass|BarClass
+ */
+class BaseClass {}
+class BazClass extends BaseClass {}
+// InheritorViolation is emitted, as BaseClass can only be extended
+// by FooClass|BarClass, which is not the case
+$a = new BazClass();
+```
--- a/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
@@ -335,7 +335,7 @@ abstract class ClassLikeAnalyzer extends SourceAnalyzer
 
 
         $classUnion = new Union([new TNamedObject($fq_class_name)]);
-        foreach ($class_storage->parent_classes as $parent_class) {
+        foreach ($class_storage->parent_classes + $class_storage->direct_class_interfaces as $parent_class) {
             $parent_storage = $codebase->classlikes->getStorageFor($parent_class);
             if ($parent_storage && $parent_storage->inheritors) {
                 if (!UnionTypeComparator::isContainedBy($codebase, $classUnion, $parent_storage->inheritors)) {
--- a/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
@@ -14,8 +14,8 @@ use Psalm\Internal\Type\Comparator\UnionTypeComparator;
 use Psalm\Internal\Type\TemplateResult;
 use Psalm\Internal\Type\TemplateStandinTypeReplacer;
 use Psalm\Issue\InaccessibleProperty;
+use Psalm\Issue\InheritorViolation;
 use Psalm\Issue\InvalidClass;
-use Psalm\Issue\InvalidExtendClass;
 use Psalm\Issue\InvalidTemplateParam;
 use Psalm\Issue\MissingDependency;
 use Psalm\Issue\MissingTemplateParam;
@@ -340,10 +340,9 @@ abstract class ClassLikeAnalyzer extends SourceAnalyzer
             if ($parent_storage && $parent_storage->inheritors) {
                 if (!UnionTypeComparator::isContainedBy($codebase, $classUnion, $parent_storage->inheritors)) {
                     IssueBuffer::maybeAdd(
-                        new InvalidExtendClass(
+                        new InheritorViolation(
                             'Class ' . $fq_class_name . ' is not an allowed inheritor of parent class ' . $parent_class,
                             $code_location,
-                            $fq_class_name,
                         ),
                         $suppressed_issues,
                     );
--- a/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
@@ -15,6 +15,7 @@ use Psalm\Internal\Type\TemplateResult;
 use Psalm\Internal\Type\TemplateStandinTypeReplacer;
 use Psalm\Issue\InaccessibleProperty;
 use Psalm\Issue\InvalidClass;
+use Psalm\Issue\InvalidExtendClass;
 use Psalm\Issue\InvalidTemplateParam;
 use Psalm\Issue\MissingDependency;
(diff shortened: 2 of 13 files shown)


src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
```diff
@@ -331,6 +332,28 @@ public static function checkFullyQualifiedClassLikeName(
             return null;
         }
 
+        foreach ($class_storage->invalid_inheritors as $dependency_class_name => $_) {
+            IssueBuffer::maybeAdd(
+                new InvalidExtendClass(
```
robchett, line 337: Unsure on the process for using/reusing Issue classes. Might need a new one for this?

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
```diff
@@ -331,6 +332,28 @@ public static function checkFullyQualifiedClassLikeName(
...
+                    'Class ' . $fq_class_name . ' can be inherited by ' . implode('|', $_) . ' but parent class ' . $dependency_class_name . ' cannot',
+                    $code_location,
+                    $fq_class_name,
+                ),
+                $suppressed_issues,
+            );
+        }
+
+        if ($class_storage->active_inheritors && !in_array($fq_class_name, $class_storage->active_inheritors)) {
```
robchett, line 346: Unsure right now on how to verify that a FQCN is a subclass of a class-string

tests/ClassTest.php
```diff
@@ -847,6 +847,58 @@ private final function __construct() {}
...
+                    class BaseClass {}
+                    class FooClass extends BaseClass {}
+                    $a = new FooClass();
+                    PHP,
+            ],
+            'unionInheritorIsAllowed' => [
+                'code' => <<<'PHP'
+                    <?php
+                    /**
+                     * @psalm-inheritors FooClass|BarClass
```
robchett, line 865: This test is failing because I don't know how to deal with the union yet

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
orklah: The issue is that this specific error is used to prevent inheritance of final classes. This is a big issue so it will be reported at a very high level (in fact, it will be reported on any level). Your issue however will not provoke any crash so you don't want to report it that soon. So another issue would be better here

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
orklah: The easiest way is to get two Union where you have your current class in an Union as a TNamedObject and the allowed classes in the other one with other TNamedObject. Then you just have to use `UnionTypeComparator::isContainedBy` with both Union and Psalm will tell if you if the first is in the other

src/Psalm/Internal/Codebase/Populator.php
```diff
@@ -509,6 +509,14 @@ private function populateDataFromParentClass(
             return;
         }
 
+        if ($parent_storage->inheritors) {
```
orklah, line 512: I'm not sure what invalid_inheritors and active_inheritors means here.

Can't you just have each storage know what allowed_inheritors it declared and check, for each class if it respects inheritors declared by every parent_class/parent_interface?

orklah (commented):
Sorry for the slow response :(

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
robchett: Lovely, that and `TypeParser::parseTokens(TypeTokenizer::getFullyQualifiedTokens(...), ...)` got me where I needed to be.

src/Psalm/Internal/Codebase/Populator.php
robchett: This was my naïve approach to dealing with with parents of parents, but I realise now that the `class_storage->parent_classes` deals with this for me. Much simpler now.

robchett:
I've been thinking about the second part of #1450 

```php
function f1(BaseClass $c) : string {
  switch (get_class($c)) {
    case FooClass:class:
      return $c->thing("me");
    case BarClass:class:
      return $c->thing(5);
  }
}

function f2(BaseClass $c) : string { // this is an error as not all paths are met
  switch (get_class($c)) {
    case FooClass:class:
      return $c->thing("me");
  }
}
```

For `get_class` to work for this then @psalm-inheritors would have to be limited to direct decedents? 

```php
/**
 * @psalm-inheritors FooClass
 */
class BaseClass {}
class FooClass extends BaseClass {}
class BarClass extends FooClass {} // Is this an error?
```

Here `BarClass` fullfils the requirement of being a `FooClass` so unless `FooClass` was final there could be any number of outputs of `get_class()`. also unless `BaseClass` was abstract or an interface then `BaseClass::class` would also need to be in the cases.

We'd still be able to do this:

```php
/**
 * @psalm-inheritors FooClass
 */
abstract class BaseClass {}

function f1(BaseClass $c) : string {
    return match (true) {
        $c instanceof FooClass => $c->thing("me"),
        $c instanceof BarClass => $c->thing(5),
    };
}

function f2(BaseClass $c) : string {
    return match (true) {
        $c instanceof FooClass => $c->thing("me"),
        // error, missing branch BarClass 
    };
}
```

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
orklah: Yeah :)

src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
```diff
@@ -331,6 +333,23 @@ public static function checkFullyQualifiedClassLikeName(
...
 
+
+        $classUnion = new Union([new TNamedObject($fq_class_name)]);
+        foreach ($class_storage->parent_classes + $class_storage->direct_class_interfaces as $parent_class) {
+            $parent_storage = $codebase->classlikes->getStorageFor($parent_class);
+            if ($parent_storage && $parent_storage->inheritors) {
+                if (!UnionTypeComparator::isContainedBy($codebase, $classUnion, $parent_storage->inheritors)) {
+                    IssueBuffer::maybeAdd(
+                        new InheritorViolation(
+                            'Class ' . $fq_class_name . ' in not an allowed inheritor of parent class ' . $parent_class,
```
orklah, line 344: in -> is

orklah (commented):
Seems ready to merge apart from the typo! Again, sorry for the delay, reviewing new features always takes me more time so I need to have some free time dedicated to that :) Your PR are very much appreciated :)

About the part with the switch, I agree, it seems dangerous to assume we know all the classes (when they're not all final) and even then it's not that much interesting to check for switch/match completeness

robchett:
Typo fixed.
No worries on the "delay", a new feature from a 4y old issue is hardly a priority.
Thank you for all your help with this, I understand the internals much better now.

docs/running_psalm/issues/InheritorViolation.md
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * @psalm-inheritors FooClass|BarClass
+ */
+class BaseClass {}
+class BazClass extends BaseClass {}
+// InheritorViolation is emitted, as BaseClass can only be extended
+// by FooClass|BarClass, which is not the case
+$a = new BazClass();
+  
+}
```
orklah, line 18: This brace is the reason behind the last failure it seems

orklah:
Thanks!

vudaltsov:
This is cool, thank you!

PR #9690, opened 2023-04-20
Replace `@template T as X` with `@template T of X` in documentation

janopae:
According to https://github.com/vimeo/psalm/commit/1986c8b4a8018b1819bc4b83b0f7e69c2c936792#r108961921, `as` is an alias to `of`. However, only `of` is documented (in docs/annotating_code/templated_annotations.md). That caused me confusion reading this.

I think as long as the alias is not documented, it's better to not use it in the docs. Even if it was documented, it would probably be better to not use aliases to avoid confusion.


orklah:
When I talked about aliases, I wanted to say those are equivalent. I'm not actually sure one is the alias of the other

@weirdan do you have more history on that? Is there a preferred way to annotate a template or are those both for legacy reason?

weirdan:
I think `as` appeared first, but I prefer `of` as it reads better

janopae:
@orklah rebased it!

orklah:
Thanks!

Issue #9634, opened 2023-04-10
date return type numeric-string for specfic formats


kkmuffme:
https://psalm.dev/r/0264caa556

If the first argument of (gm)date matches `^[djNwzWmntLoyBgGhHisuvIZU]+$` then the return type should be numeric-string instead of string

orklah:
This may be too complicated to put into a stub. A return type provider would do the trick though

PR #9691, opened 2023-04-20
Add return type provider for date/gmdate

robchett:
Solution for #9634


debug.log
```diff
@@ -0,0 +1,5 @@
+
+Psalm\Type\Atomic\TLiteralInt
```
orklah, line 2: This file should be removed :)

PR #9701, opened 2023-04-23
Handle included directory paths without crashing

### LINKED ISSUE #9639: Empty require statement generates an error that is difficult to track down
Hi,
I cannot provide a test since `allowFileIncludes` flag is rightfully turned off on the test site, but basically,

```php
<?php

require ''; 

// ...
``` 

With `--debug-by-line` this will give you something like,

```sh
Finished registering autoloaded files
Visiting autoload files took 1.402s
Checking [...]
Target PHP version: [...]
Scanning files...
Initializing plugins...
Uncaught UnexpectedValueException: File [...]
Stack trace:
[...]
(Psalm 5.9.0@8b9ad1eb9e8b7d3101f949291da2b9f7767cd163 crashed due to an uncaught Throwable)
```

I had to manually pinpoint the exact file causing an issue and this is when I saw the empty `require` statement. Removing it fixed the issue.

I know this is VERY niche but it would have been nice to see the problematic file being reported right away.

Cheers,

edsrzf:
If an include or require resolves to a directory, we now raise a `MissingFile` error rather than throwing an unhandled `UnexpectedValueException`.

Fixes #9639.

### DIFF
--- a/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
@@ -1139,6 +1139,11 @@ class ProjectAnalyzer
         return $this->file_provider->fileExists($file_path);
     }
 
+    public function isDirectory(string $file_path): bool
+    {
+        return $this->file_provider->isDirectory($file_path);
+    }
+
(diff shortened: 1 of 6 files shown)


orklah:
Thanks!

PR #9704, opened 2023-04-24
Allow running new on template objects

danog:


tests/ReturnTypeTest.php
```diff
@@ -1264,6 +1264,29 @@ function produceParameters(array $configuration): array
...
+                     */
+                    function aggregate($type) {
+                        $t = new $type;
+                        return $t;
+                    }
+
+
+                    class b extends AggregateResult {}
+
+                    $b = aggregate(new b);',
```
orklah, line 1285: For clarity, the new b here (and even the whole function call) should not be part of the unit test given it already works as expected: https://psalm.dev/r/33cf9009d7

Stopping [InvalidReturnType](https://psalm.dev/011) is the sole purpose of the test I think

orklah:
you need to remove the assertion too

orklah:
Thanks!

Issue #3626, opened 2020-06-21
Add @mixin documentation


githoober:
@mixin was added to psalm in https://github.com/vimeo/psalm/releases/tag/3.8.2. The information about what kind of problems it solves is missing there.

klimick:
Seems like the issue can be closed.

PR #9706, opened 2023-04-25
Document @mixin

robchett:
As requested in #3626 

Phpdocumentor doesn't have a page for @mixin so I've used the one from phpstan.org.


orklah:
I'm somewhat afraid that using PHPStan as a documentation will leave us in a weird situation if they ever decide to implement something differently.

It's not so much an issue with phpdocumentor and php.net because the pace of changes is way lower than the static analysis world
 I'd rather have an internal

orklah:
Thanks!

Issue #8892, opened 2022-12-12
UnevaluatedCode for `declare` with `return`


ElisDN:
It works in global namespace:

https://psalm.dev/r/d4d61a5673

but it throws error with namespace:

https://psalm.dev/r/49ce4900b9

orklah:
There's a fault in Psalm handling of statements in and out namespaces

What happens here is:
- Psalm sees a new file
- It first process the classes, interfaces, traits, and namespaces
- For each namespace, it also extracts classes, interfaces and traits
- It puts every leftover statement in the namespace in a list
- It analyze the list of leftover statement for each namespace (so here, the return) because return could contain an expression that needs to be analyzed from within the namespace
- Then Psalm analyze statements that were out of the namespaces (here, it's the declare). However, the return already flagged the whole file as having returned so it fails.

There's a reason to that though, because any statement in the file could use any symbol also declared in the file, no matter the order of the statements.

So I guess, we would need to first extract symbols, collect leftover statement for each namespace and then process statements in the correct context and in order(so declare first, then leftovers by namespace). This may be a big change to support a very unusual pattern though...

What I'm more worried about is the fact that the declare is analyzed at then end (after namespaces) so I'm not sure it's correctly processed when analyzing other statements

weirdan:
> What I'm more worried about is the fact that the declare is analyzed at then end (after namespaces) so I'm not sure it's correctly processed when analyzing other statements

Indeed: https://psalm.dev/r/3a0287fa0f

vudaltsov:
This issue can be now closed!

orklah:
Yes, this specific issue has been fixed. For future reference, https://github.com/vimeo/psalm/issues/8892#issuecomment-1347412144 still holds and the way Psalm process this is not optimal (and lead to very marginal edge cases for files that would contain a strict type declaration and namespace that have expressions that need to be evaluated)

PR #9719, opened 2023-04-27
Fix UnevaluatedCode false positive at declare(strict_types=1)

klimick:
I can't reproduce it at psalm.dev and can't write useful test. But locally I see `UnevaluatedCode` on namespaced files without classes:

image

https://psalm.dev/r/59e04972a3 (no errors, but actually there is)


orklah:
reproducer: https://psalm.dev/r/59e04972a3

with the correct settings :)

orklah:
Note: this is not an optimal fix

I explained that somewhere else but not sure where. The underlying issue has to do with how Psalm process files. It processes classes first, then it goes to read statements, however it ends up reading the exit() before the declare and then fails because it thinks the declare came after

But that fix is simple enough and works for this simple case so we might as well go with it for now

orklah:
Found the explanation: https://github.com/vimeo/psalm/issues/8892#issuecomment-1347412144

PR #9720, opened 2023-04-27
Fix ambiguous constant inheritance false positive

### LINKED ISSUE #7818: AmbiguousConstantInheritance does not handle multiple interhitances of same interface
https://psalm.dev/r/b1dfb3d8c0

We are using a similar construct. A main interface for all Services, and several interfaces for what the Service supports. The supporting interface must implement ServiceInterface, so we are clear that it is a service interface.

ServiceWithPermissionsInterface extends ServiceInterface
DistributeableServiceInterface extends ServiceInterface

etc.

Tried to add `final` to solve the problem, but it just became worse. PHP itself handles the `final` very well.
https://psalm.dev/r/99f7bf3fc4

Jean85:
Fixes #7818

Thanks to @orklah for [his comment](https://github.com/vimeo/psalm/issues/7818#issuecomment-1371420642), that allowed me to fix this even without knowing anything about the Psalm codebase!

I've used https://psalm.dev/r/99f7bf3fc4 as a regression test.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php
@@ -844,6 +844,7 @@ class ClassConstAnalyzer
                     assert($parent_classlike_storage !== null);
                     if (!isset($parent_classlike_storage->parent_interfaces[strtolower($interface)])
                         && !isset($interface_storage->parent_interfaces[strtolower($parent_classlike_storage->name)])
+                        && $interface_const_storage !== $parent_const_storage
                     ) {
                         IssueBuffer::maybeAdd(
                             new AmbiguousConstantInheritance(
--- a/tests/ConstantTest.php
+++ b/tests/ConstantTest.php
@@ -1892,6 +1892,19 @@ class ConstantTest extends TestCase
                     '$s===' => 'array{9223372036854775806: 0, 9223372036854775807: 1}',
                 ],
             ],
+            'inheritedConstantIsNotAmbiguous' => [
+                'code' => <<<'PHP'
+                    <?php
+                    interface MainInterface {
+                        public const TEST = 'test';
+                    }
+
+                    interface FooInterface extends MainInterface {}
+                    interface BarInterface extends MainInterface {}
+
+                    class FooBar implements FooInterface, BarInterface {}
+                    PHP,
+            ],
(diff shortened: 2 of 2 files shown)


Jean85:
Ouch, the failing test is highlighting that my fix is too shallow, it swallows the error from the docs snippet. Any pointers?

orklah:
Thanks for this!

There's 3 variables with a constant in that code

$const_storage: it's received as a param. It's the storage of the constant retrieved from the class storage of Baz (it's arbitrarily chosen from the 2 constant storage that exists for Baz::CONSTANT)
$parent_const_storage: it's the constant storage that is retrieved from the interface that we're looking for in the current iteration of the loop
$interface_const_storage: it's the constant storage that was retrieved from the interface we were looking on the previous iteration of the loop

So when everything is ok, we iterate for each interface, we check for the existence of that constant. If it doesn't exists, we continue looping until we found that constant in one interface. When that happens we store the storage for comparison with the next iteration:
https://github.com/vimeo/psalm/blob/8c491e93e8a4bc0934f07fa232ae4bd8124b58aa/src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php#L859 

So your code compares the arbitrary constant storage with the current constant storage so the result is arbitrary. What you want to do is compare the storage between two interfaces, so between two iterations of the loop (so between  $parent_const_storage and $interface_const_storage)

As a note: you should be able to directly compare storages instead of comparing the hash of their location given each instance is should be unique

src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php
```diff
@@ -844,6 +844,7 @@ private static function getOverriddenConstant(
                     assert($parent_classlike_storage !== null);
                     if (!isset($parent_classlike_storage->parent_interfaces[strtolower($interface)])
                         && !isset($interface_storage->parent_interfaces[strtolower($parent_classlike_storage->name)])
+                        && $interface_const_storage->stmt_location !== $parent_const_storage->stmt_location
```
orklah, line 847: $interface_const_storage !== $parent_const_storage should be good too :)

Jean85: Done in e8bc9a620

orklah:
Thanks!

Issue #9677, opened 2023-04-19
Way to widen/narrow inference


klimick:
Currently, Psalm aggressively narrows all inferred types:

https://psalm.dev/r/c97bc21c8e
```php
<?php

$a = 42; // inferred as 42
/** @psalm-trace $a */;

$b = [40, 41, 42]; // inferred as list{40, 41, 42}
/** @psalm-trace $b */;

$c = new ArrayObject(['str' => 42]); // inferred as ArrayObject<'str', 42>
/** @psalm-trace $c */;
```

It would be nice to be able to manage it somehow. For example:

```php
<?php

/** @widen */
$a = 42; // inferred as int
/** @psalm-trace $a */;

/** @widen */
$b = [40, 41, 42]; // inferred as non-empty-list<int>
/** @psalm-trace $b */;

/** @widen */
$c = new ArrayObject(['str' => 42]); // inferred as ArrayObject<non-empty-string, int>
/** @psalm-trace $c */;
```

Or class templates: 
https://psalm.dev/r/bacd1cb8e7
```php
<?php

/**
 * @template A widen
 */
final class Container
{
    /**
     * @param A $value
     */
    public function __construct(
        public readonly string $value,
    ) {}
}

// Currently infers as Container<42>
// but with the "widen" modifier it will be Container<int>
$container = new Container(42);
/** @psalm-trace $container */;
```

Or even with functions:
https://psalm.dev/r/ec18b1dd81
```php
<?php

/**
 * @template A widen
 * @param A $value
 * @return A
 */
function id(mixed $value): mixed
{
    return $value;
}

// Currently infers as 42
// but with the "widen" modifier it will be int
$value = id(42);
/** @psalm-trace $value */;
```

What do you think?
I could try to implement this.

danog:
To be frank, I don't see any sane usecase for such a feature.

klimick:
> To be frank, I don't see any sane usecase for such a feature.

What I want: https://psalm.dev/r/a8805000d4
What I do actually: https://psalm.dev/r/3ebeacd959

What I want: https://psalm.dev/r/a9500084ee
What I do actually: https://psalm.dev/r/a541f502a1

I often have to write dock blocks in places where this could have been avoided.

klimick:
> While I see your issue in this case, personally I would simply use a single template parameter, because IMO there is no real use in specifying a separate template type just for the accumulator type, which in 99% of cases (including this one) is of the same overall type as the input type: ...

[...]

PR #9729, opened 2023-05-02
Opt-in unconstrained templates widening

### LINKED ISSUE #4564: Problem with literals
I'm trying to write stubs for a liibrary and I have some problems with literal strings. I just need to compare the type and not the value.

How can I do something like that?

This is a simple example where I would like that the callable should return the same type of the input param, but not the same.
There is a way to do it? Maybe would be nice some magic type like `non-literal<A>`?

```php
/**
 * @template A
 * @param A $a
 * @return callable(A): A
 */
function foo($a): callable
{
    return fn ($b) => $b;
}

foo('a')('b');
```

https://psalm.dev/r/915c2c72eb

klimick:
Closes #4564

1) I prefer leave old behavior by default.
Even It aligns Psalm with PHPStan It still is BC change.

2) Widening/narrowing is controllable both with config and annotations. 

3) I relly don't know how to widen string and array/list.
Should `'some-literal'` widens to `non-empty-string` or `string`?
Should `[1, 2, 3]` widens to `non-empty-list<int>` or `list<int>`?

I chose non-empty side, but it can't fix original snippet from #4564.
```php
<?php

/**
 * @template A
 * @param A $a
 * @return Closure(A): bool
 */
function eq($a): Closure
{
    return fn($b) => $a === $b;
}

function isEqTo123(string $value): bool
{
    $equals = eq('123');

    // ERROR: ArgumentTypeCoercion
    // Argument 1 expects non-empty-string, but parent type string provided
    return $equals($value);
}
```

It seems to PHPStan widens non-empty: https://phpstan.org/r/707c06d9-e28b-4209-a9af-9c5602190d27
Should we do the same?

### DIFF
--- a/src/Psalm/Internal/Type/Comparator/GenericTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/GenericTypeComparator.php
@@ -136,36 +136,33 @@ class GenericTypeComparator
                 && !$container_type_part instanceof TIterable
                 && !$container_param->hasTemplate()
                 && !$input_param->hasTemplate()
+                && !($container_type_params_covariant[$i] ?? false)
             ) {
-                if ($input_param->containsAnyLiteral()) {
+                if ($input_param->containsAnyLiteral() || $input_param->containsAnyNonEmptiness()) {
                     if ($atomic_comparison_result_type_params !== null) {
                         $atomic_comparison_result_type_params[$i] = $container_param;
                     }
-                } else {
-                    if (!($container_type_params_covariant[$i] ?? false)
-                        && !$container_param->had_template
+                } elseif (!$container_param->had_template) {
+                    // Make sure types are basically the same
+                    if (!UnionTypeComparator::isContainedBy(
+                        $codebase,
+                        $container_param,
+                        $input_param,
+                        $container_param->ignore_nullable_issues,
+                        $container_param->ignore_falsable_issues,
+                        $param_comparison_result,
+                        $allow_interface_equality,
+                    ) || $param_comparison_result->type_coerced
                     ) {
-                        // Make sure types are basically the same
-                        if (!UnionTypeComparator::isContainedBy(
-                            $codebase,
-                            $container_param,
-                            $input_param,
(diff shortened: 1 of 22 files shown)


src/Psalm/Internal/Type/Comparator/GenericTypeComparator.php
```diff
@@ -136,36 +136,33 @@ public static function isContainedBy(
                 && !$container_type_part instanceof TIterable
                 && !$container_param->hasTemplate()
                 && !$input_param->hasTemplate()
+                && !($container_type_params_covariant[$i] ?? false)
```
klimick, line 139: It's here because I think there is no sense do specialization for covariant parameter.

danog:
I really don't think it's a good idea to offer this as a configuration parameter, or even as a function-wide `@psalm-widen-unconstrained-templates`, since it can potentially break a lot of stuff.

A `@template-widen` annotation to use instead of `@template` would be much better,  since it provides a lot more granularity.

src/Psalm/Internal/Type/Comparator/GenericTypeComparator.php
```diff
@@ -136,36 +136,33 @@ public static function isContainedBy(
...
             ) {
-                if ($input_param->containsAnyLiteral()) {
+                if ($input_param->containsAnyLiteral() || $input_param->containsAnyNonEmptiness()) {
```
klimick, line 141: In the past, I wanted to remove it, but changed my mind. Without this, it can be really hard to write tests or fixtures where the values are hardcoded.

Moreover, I have added here `Box<non-empty-list<T>>` to `Box<list<T>>` specialization.

klimick:
> A @template-widen annotation to use instead of @template would be much better, since it provides a lot more granularity.

And It easy to implement. Similar I previously suggested in the #9677.

But what about non-emptiness? How to widen it?
Literal string should be widen to `string` or `non-empty-string`.
And the same question about `non-empty-list`/`non-empty-array`.

klimick:
I need to collect more information before returning to this issue again.

PR #9735, opened 2023-05-03
Fix missing global class type when scanning cached files

tscni:
Once the project files are cached, `ProjectAnalyzer` will no longer scan them again if they haven't changed.
This means that a global class type defined in the configuration ([`globals`](https://psalm.dev/docs/running_psalm/configuration/#globals)) must be scanned in another way. Otherwise analyzing a file using such a global would raise an "UndefinedClass" issue.

To me it seems to be the best solution to consider the configured global types for scanning in `ReflectorVistor`, similar to the way an `@var` annotation would be handled.

I haven't found any tests for file scanning and wasn't sure if a more involved end-to-end test case would be desired, so I just created a test case for `FileScanner`.

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
+++ b/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php
@@ -238,6 +238,12 @@ class ReflectorVisitor extends PhpParser\NodeVisitorAbstract implements FileSour
                             $var_id = '$' . $var->name;
 
                             $functionlike_node_scanner->storage->global_variables[$var_id] = true;
+
+                            if (isset($this->codebase->config->globals[$var_id])) {
+                                $var_type = Type::parseString($this->codebase->config->globals[$var_id]);
+                                /** @psalm-suppress UnusedMethodCall */
+                                $var_type->queueClassLikesForScanning($this->codebase, $this->file_storage);
+                            }
                         }
                     }
                 }
--- a/tests/Internal/Scanner/FileScannerTest.php
+++ b/tests/Internal/Scanner/FileScannerTest.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Psalm\Tests\Internal\Scanner;
+
+use Psalm\Aliases;
+use Psalm\Codebase;
+use Psalm\Config;
+use Psalm\Internal\Provider\FakeFileProvider;
+use Psalm\Internal\Provider\Providers;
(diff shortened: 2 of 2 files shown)


orklah:
I'm not sure I understood how the cache would result in UndefinedClass, could you rephrase or maybe show Psalm's output with/without caching?

tscni:
> I'm not sure I understood how the cache would result in UndefinedClass, could you rephrase or maybe show Psalm's output with/without caching?

Makes sense, I already found it quite difficult to phrase :P
So here's an example to demonstrate the issue (I probably should've lead with that or created an actual issue for it first? It wasn't quite clear to me what your preferences are.)

#### Setup
```xml
<!-- psalm.xml -->
<?xml version="1.0"?>
<psalm errorLevel="1" xmlns="https://getpsalm.org/schema/config" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd">
    <projectFiles>
        <file name="GlobalClass.php"/>
        <file name="test.php"/>
    </projectFiles>

    <globals>
        <var name="global" type="GlobalClass"/>
    </globals>
</psalm>
```
```php
// GlobalClass.php
<?php

class GlobalClass {
    public function run(): void { echo 'run'; }
}

// test.php
<?php

function begin_transaction(): void
{
    global $global;
    $global->run();
}
```

#### Reproducible demonstration

1. Use PHP 8.1.17
1. `composer init -n --name test/test --require vimeo/psalm:5.10.0 && composer install`
1. Run `vendor/bin/psalm-v`
   ```
   Psalm 5.10.0@a5effd2d2dddd1a7ea7a0f6a051ce63ff979e356
   ```
2. Run `vendor/bin/psalm`
   ```
   No errors found! 
   ```
3. Change `test.php`
   ```diff
    <?php
   +// random comment
   ```
4. Run `vendor/bin/psalm`
   ```
   ERROR: UndefinedClass - test.php:6:5 - Class, interface or enum named GlobalClass does not exist (see https://psalm.dev/019)
       $global->run();
   ```

Effectively there's no reference to `GlobalClass.php` that'd trigger Psalm to scan it again, meaning in the context of step 4, the class doesn't exist.

orklah:
I'd have expected GlobalClass to be a part of the cache so a trigger would not have been needed but it doesn't appear to be the case!

PR #9736, opened 2023-05-03
SplDoublyLinkedList's index is always int

kamil-tekiela:
Hopefully it will fix "offsetSet cannot be null, null value provided to parameter with type int" when using `$dll[] = 42` syntax.


orklah:
Thanks for the fix! I'm doubtful about the resolution of the error you posted though

PR #9738, opened 2023-05-04
Fix invalid cast for templated class-string

klimick:
`TClassString` atomic is accidentally replaced with `TGenericObject`:
https://psalm.dev/r/831ddeffde

There is nothing wrong with that, because it is necessary for things like that:
https://psalm.dev/r/6b6ea39278

But this case is obviously missed in the `ClassLikeStringComparator`.


orklah:
Thanks for this! I tried to fix it before but I failed!

PR #9739, opened 2023-05-04
Fix ParadoxicalCondition with complex match expression

klimick:
Works as expected: https://psalm.dev/r/5252b45cae
But it raise an unexpected issue: https://psalm.dev/r/c0a5164e67

I hope this can be fixed in this way)


orklah:
Thanks!

PR #9740, opened 2023-05-04
Fix OverriddenPropertyAccess error message

### LINKED ISSUE #9727: Confusing error message for OverriddenPropertyAccess
The error message here is confusing: https://psalm.dev/r/f066e6d131

I would expect it to mention the base class `A`, like:

```
Property B::$foo has different access level than A::$foo
```

edsrzf:
The message previously didn't mention the base class, which made it confusing.

Fixes #9727.


orklah:
Thanks!

PR #9743, opened 2023-05-04
Fix string negation to never

klimick:
In else branch `$str` should be `string` but actually `never`:
https://psalm.dev/r/9a6d40e0f6


orklah:
Thanks!

PR #9745, opened 2023-05-04
Ignore nullable return form SimpleXMLElement::addChild()

greg0ire:
It seems unrealistic to require users to perform a null check every time they call this method.


orklah:
Thanks!

Issue #9684, opened 2023-04-20
Whats happen with invariant template?


klimick:
I expect errors in the both cases. But case with scalars silently hides mistakes.
```php
// No error here. But Wrapper<42> is not Wrapper<int>
acceptStringWrapper(new Wrapper(42));

// Eroror here as expected. Because Wrapper<MoreSpecificObj> is not Wrapper<Obj>
acceptObjWrapper(new Wrapper(new MoreSpecificObj()));
```
https://psalm.dev/r/ded731f803

klimick:
Sorry. Previous example may be confusing.
https://psalm.dev/r/50bdd20919

orklah:
Seems like it boils down to this code:
https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Type/Comparator/GenericTypeComparator.php#L140-L172

Both checks end up here, but when the input type contains literal, it goes in the first if and don't say the comparison is false. When it doesn't contains literals, we check if the param is contained in the input (basically the two types are the same) and if they're not, we flag an error because of covariance

So there seems to be no check for covariance when dealing with literals. I'm not sure if it was on purpose, but I can definitely say we would probably get a lot more support issue if it was the case (for exactly the reason why you opened the other ticket: Psalm tries to infer the lowest possible template type for your input)

klimick:
> but I can definitely say we would probably get a lot more support issue if it was the case

So you suppose to leave it as is?

orklah:
I'd be inclined to yeah

Is there a practical case you have in mind that justifies that, for a template with literals, you want to inforce covariance?

klimick:
> Is there a practical case you have in mind that justifies that, for a template with literals, you want to inforce covariance?

It's more about the type system consistency than the practical cases.
It's hard to explain some things that happen in Psalm to other folks.
[...]

Issue #9203, opened 2023-01-31
Urgent: templates lost when calling a function!


thomasvargiu:
I think this is a big bug: whenever I call a function using a class with a template, after the function call the template values are lost!

https://psalm.dev/r/5a9138cf56

thomasvargiu:
I was thinking that probably Psalm doesn't know if someone inside the function can change the property, but this happens even with readonly properties and/or immutable classes.
https://psalm.dev/r/ee0b12f00c

thomasvargiu:
Ok, I don't know why I notice it just right now, probably because I worked to a project going to level 1.

> This will have to be investigated to see if it's really a bug or some kind of safety for whatever purpose.

Probably it's better to have a `mixed` than a wrong value when the class is mutable, or the template is for a mutable property. A good first improvement should be to don't loose the template when the class is immutable.
It will also be safe for sub-templates like `ImmutableContainer<MutableContainer<B>>` because we can't call mutable methods (of the mutable) from an immutable class.

orklah:
My current hypothesis is that it's a normal behaviour. I'm pretty sure this is somewhat related to covariance: https://psalm.dev/docs/annotating_code/templated_annotations/#template-covariance

For example, if you create a class that has a template with `new Collection([new Dog()]);` the intent is not perfectly clear. Do you want to have a Collection that can contain only Dogs, any Animal or maybe anything is welcome?

When you pass the Collection to a function that can handle a `Collection<mixed>`, Psalm seems okay with that. The function could do anything (what if add() inserts a `bar` inside your `Collection<'foo'>` ?). So after calling add(), Psalm is forced to trust the upper bound used by add() (in your example it's mixed, but you can try with string, it works too)

So either:
- Psalm should forbid you to call add() 
- This behaviour is perfectly sane but has the side effect of widening your inferred template in order to the upper_bound used by add()

PR #9746, opened 2023-05-04
Fix keyed array generalization

klimick:
`$container` should be generalized after `test` call https://psalm.dev/r/6239d321f0 like https://psalm.dev/r/23e5380fa9 isn't it?

### DIFF
--- a/src/Psalm/Internal/TypeVisitor/ContainsLiteralVisitor.php
+++ b/src/Psalm/Internal/TypeVisitor/ContainsLiteralVisitor.php
@@ -4,6 +4,7 @@ namespace Psalm\Internal\TypeVisitor;
 
 use Psalm\Type\Atomic\TArray;
 use Psalm\Type\Atomic\TFalse;
+use Psalm\Type\Atomic\TKeyedArray;
 use Psalm\Type\Atomic\TLiteralFloat;
 use Psalm\Type\Atomic\TLiteralInt;
 use Psalm\Type\Atomic\TLiteralString;
@@ -26,6 +27,11 @@ class ContainsLiteralVisitor extends TypeVisitor
             || $type instanceof TTrue
             || $type instanceof TFalse
         ) {
+            $this->contains_literal = true;
+            return self::STOP_TRAVERSAL;
+        }
+
+        if ($type instanceof TKeyedArray && !$type->isGenericList()) {
             $this->contains_literal = true;
             return self::STOP_TRAVERSAL;
         }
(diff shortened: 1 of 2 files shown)


orklah:
This is very related to https://github.com/vimeo/psalm/issues/9203 and https://github.com/vimeo/psalm/issues/9684#issuecomment-1516147457. In those issue, I said that I was not sure if it was a bug or a feature. 

I'm still very not sure. Your PR however goes many miles further. The Visitor is used to check if a type contains Literals recursively. It currently considers that scalar are literals and an empty array is a literal (not even null is considered a literal here).

What you added here says "as long as we have a keyed array, at any depth, we have a literal". In your example, you go from `Container<list{42}>` to `Container<non-empty-list{int}>`. Most likely the important part here was to go from `42` to `int` to allow widening a template that would have been inferred too specific. The `list` to `non-empty-list` goes in the other way, it make the type more narrow.

So it's already a weird edge case. We somehow ends up with a Container that has been inferred to being non-empty just because we passed it to a function that expected it to be non-empty (that should have been an error if there was no literal in there)

That's actually the part you extended by allowing any template containing a TKeyedArray to be widened (and narrowed) to match the called function

So I don't think I agree with the spirit of the PR. On top of that, the ContainsLiteralVisitor now detects as a literal something that is very much not a literal so I don't think I agree with the technical side either

klimick:
> In your example, you go from Container<list{42}> to Container<non-empty-list{int}>.

Did you misspelled? My example shows generalization from `Container<list{42}>` to `Container<non-empty-list<int>>`

For me strange that `Container<list{42}>` can be generalized  to `Container<non-empty-list<int>`
but `Container<list{AnyNonLiteralType}>` cannot be generalized  to `Container<non-empty-list<AnyNonLiteralType>`.

For me strange at all that generalization works for invariant templates.
It is normal for covariant templates, but not for invariant.

I guess it done because we can't explicitly instantiate template during initialization.
I.e `new Container<string>('foo')`. It would be `Container<string>`, but we lack syntax like that.
I don't consider `@var` because it have another aims (It lies to Psalm sometime for example :D)
And because we have that omission by design.

Maybe we can ask Matthew Brown? 
I understand that he is tired of Psalm, but apparently he did this behavior with literals and who can better know why this happens?)

orklah:
> Did you misspelled

I did (though it does not change the point too much)

But yeah, we can try to ask Matt about this if we want to get the reason for this whole thing

klimick:
I need to collect more information before returning to this issue again.

Issue #9748, opened 2023-05-05
glob stubs - non-empty-string


kkmuffme:
https://psalm.dev/r/0c83fa73ac

Afaik results in glob cannot be empty string, should be `false|list<non-empty-string>`

orklah:
You can directly make the PR for small things like this :) don't bother creating issues!

PR #9752, opened 2023-05-06
improve glob stubs

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/9748

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -3329,7 +3329,7 @@ return [
 'gettimeofday' => ['array<string, int>'],
 'gettimeofday\'1' => ['float', 'as_float='=>'true'],
 'gettype' => ['string', 'value'=>'mixed'],
-'glob' => ['list<string>|false', 'pattern'=>'string', 'flags='=>'int'],
+'glob' => ['list<non-empty-string>|false', 'pattern'=>'non-empty-string', 'flags='=>'int<1, max>'],
 'GlobIterator::__construct' => ['void', 'pattern'=>'string', 'flags='=>'int'],
 'GlobIterator::count' => ['int'],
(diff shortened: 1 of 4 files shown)


orklah:
Thanks!

Hanmac:
I think glob `pattern` should be allowed to be empty string. That is total valid code.

currently it causes problems with linked project above because we use glob with sprintf:
 `$files = glob(sprintf('%s/*.xliff', $path));`

psalm says that sprintf just returns empty string, but glob now wants non-empty-string and that causes an error.

VincentLanglet:
The pattern can be an empty string cf https://onlinephp.io/c/002b0, it's valid code with valid response. cc @orklah 

You can do a PR to change this @Hanmac

Hanmac:
@kkmuffme no the problem is YOU broke it, so YOU should fix it.

>@Hanmac The solution is to create a return type provider for sprintf to return non-empty-string/non-falsy-string if the pattern contains anything but placeholders (or if only placeholders but they are non-empty, then non-empty too)

example `sprintf('%s', $var)` does this return empty or non empty string? you can't predict it

VincentLanglet:
> @VincentLanglet it's valid, but it doesn't make any sense, since you always get an empty array and in 99.99% of cases an empty string in glob is uninteded behavior. The solution is to create a return type provider for sprintf

Still, it's valid. It's not because it returns an empty array that you have to forbid passing an empty string.
You should only restrict param when it create exceptions.

Because if I follow your strategy I could forbid this code https://psalm.dev/r/fe8f6e0100
strlen should only accept non empty string, since you always get 0 when you pass an empty string.

ygottschalk:
Actually the assumption of non-empty-list 's in my stubs might be wrong considering `GLOB_ERR`.

PR #9761, opened 2023-05-09
Removed return type false from date and gmdate

### LINKED ISSUE #9756: False positives: PossiblyFalseOperand and PossiblyFalsePropertyAssignmentValue raised when using date functions
It seems like `PossiblyFalseOperand` and `PossiblyFalsePropertyAssignmentValue` are raised when using [date](https://www.php.net/manual/en/function.date.php) and [gmdate](https://www.php.net/manual/en/function.gmdate.php) functions, as shown in this snippet: https://psalm.dev/r/b56da40fc6 .

The PHP documentation specifies that the return type of these 2 functions is `string`, so I wouldn't expect `false` to be returned.

Alkisum:
Closes #9756


orklah:
Thanks!

Issue #9697, opened 2023-04-21
Writes to references of readonly properties not detected


jack-worman:
https://psalm.dev/r/c663093cf6

jack-worman:
Here is a simplified example: https://psalm.dev/r/eb2781fdef

weirdan:
A sensible approach would be to entirely forbid obtaining references to read-only properties

orklah:
There are cases where this could be legit though. Memory management would probably be one use case where you want to pass references around, even for reading

PR #9769, opened 2023-05-10
UnsupportedPropertyReferenceUsage

jack-worman:
Psalm should, at a minimum, give errors when it knows it cannot guarantee the soundness of a piece of code.

Closes: https://github.com/vimeo/psalm/issues/9697


orklah:
Thanks!

PR #9779, opened 2023-05-11
Enable opcache if it is installed

danog:
I realized the message is a bit misleading, it asks the user to install the opcache extension, but it also has to be enabled via `opcache.enable=1` for psalm to pick it up, but `opcache.enable=1` only affects php-fpm, while Psalm is only run via cli, so it makes little sense for us to ask to change that setting.

Users can still disable opcache for Psalm by not loading the extension at all, which generally is the standard situation on dev machines anyway.


orklah:
Can you fix the CS check?

danog:
Done!

orklah:
Thanks!

PR #9780, opened 2023-05-12
Documented suggestions for HTML user input

mmcev106:
Some of our developers have been struggling with how to handle user input from rich text editors like [TinyMCE](https://www.tiny.cloud/).  I think these suggestions are a good start.  What do y'all think?


orklah:
Thanks!

Issue #9707, opened 2023-04-25
Document new generic array-shape syntax


ygottschalk:
#9694 implements a great addition to the array syntax.
We should let useres know about this!

robchett:
Can be closed now?

PR #9784, opened 2023-05-13
Document array shape unsealing syntax

robchett:
Requested in issue https://github.com/vimeo/psalm/issues/9707


orklah:
Cool! Can you also explain that `...` is an alias for `...<array-key, mixed>`?

robchett:
Yeah, sure. I also noticed there was a section on "Unsealed array and list shapes" so I moved this as an extension of that.

orklah:
Thanks!

PR #9787, opened 2023-05-14
Incorrectly asserting non-empty-list after combining with empty list

### LINKED ISSUE #9594: No error when adding list to array of non-empty-list
There should be a warning when adding a list to an array of non-empty-lists
https://psalm.dev/r/7ddbfc99b9

robchett:
Fixes #9594

### DIFF
--- a/src/Psalm/Internal/Type/TypeCombiner.php
+++ b/src/Psalm/Internal/Type/TypeCombiner.php
@@ -1516,7 +1516,7 @@ final class TypeCombiner
                     $generic_type_params[1],
                     $objectlike_generic_type,
                     $codebase,
-                    $overwrite_empty_array,
+                    false,
                     $allow_mixed_union,
                 );
             }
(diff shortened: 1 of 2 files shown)


robchett:
Tracked this down to here: https://github.com/vimeo/psalm/blob/5370492f971941f6e9d6f23f75321eeecfba3fc9/src/Psalm/Internal/Type/TypeCombiner.php#L1364-L1448

Need to get my head around the interactions between `$overwrite_empty_array`, `$objectlike_entry->setPossiblyUndefined` and `$combination->array_type_params = [];` 

Current best effort is switching `$overwrite_empty_array` at the callsite

https://github.com/vimeo/psalm/blob/5370492f971941f6e9d6f23f75321eeecfba3fc9/src/Psalm/Internal/Analyzer/Statements/Expression/Assignment/ArrayAssignmentAnalyzer.php#L692-L704

but it needs to be smarter than that as then we loose `[] + T' = non-empty-list<T>`

orklah:
Yeah, this breaks things pretty dramatically :D 

Not sure if I'll be able to find time to dig into it to help in the foreseeable future

robchett:
Finally found the time to crack this, nested array type combiners were inheriting overwrite_empty_array so I've made it only apply to root level array type modifications and not any value types.

orklah:
Thanks!

Issue #9753, opened 2023-05-06
strpos/stripos should assert non-empty-string like str_contains


kkmuffme:
https://psalm.dev/r/2f458027ac

orklah:
seems easy to implement: https://github.com/vimeo/psalm/blob/1a1d9c9bf5b48f5e71bdc0a1f9dd19ac20052ea8/stubs/CoreGenericFunctions.phpstub#L1085

robchett:
I'm not sure this is a simple fix as the return type of str(i)pos is false|int, there would need to be an assert-if-int annotation?

orklah:
Some tests would be needed but it could still work with `assert-if-true` because Psalm does not check for true but check if you will enter the condition or not, so it's more a assert-if-truthy

robchett:
`int` is not truthy because of `0` and `0` is a valid return when the haystack starts with the needle.

Is it currently possible to do input type narrowing in a ReturnTypeProvider?

orklah:
0 is not truthy, but for things like 

`if(strpos(...)){ }`
or 
`if(strpos(...) !== false){ }`
it should still work as intended, but yeah, it's a dirty hack.

It could probably be possible to narrow the type from a ReturnTypeProvider but it may be hard. You do have the context so you know the type of each variable in the caller context and you can change them

robchett:
assert-if-false would require changing the type to non-empty-string in the default case then widening when false which would require multiple annotations if it worked at all
``` 
/** 
 * @psalm-assert non-empty-string $haystack
 * @psalm-assert-if-false string $haystack
 */
```
Orklah's "hack" did indeed work he said it would so I've opened https://github.com/vimeo/psalm/pull/9788

orklah:
That was fixed :)

PR #9788, opened 2023-05-14
Narrow the type of $haystack when strpos(...) !== false

robchett:
Solves #9753

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -627,12 +627,23 @@ function substr_replace($string, $replace, $offset, $length = null) {}
 
 /**
  * @psalm-pure
- *
  * @param string $haystack
- *
+ * @param string $needle
+ * @param int $offset
+ * @psalm-assert-if-true =non-empty-string $haystack
  * @psalm-return positive-int|0|false
  */
-function strpos($haystack, $needle, int $offset = 0) : int|false {}
+function strpos(string $haystack, string $needle, int $offset = 0) {}
+
+/**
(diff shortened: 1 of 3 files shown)


stubs/CoreGenericFunctions.phpstub
```diff
@@ -627,12 +627,23 @@ function substr_replace($string, $replace, $offset, $length = null) {}
 
 /**
  * @psalm-pure
- *
  * @param string $haystack
- *
+ * @param string|int $needle
```
orklah, line 631: Seems to contradict the type on the param below

robchett: I've removed the int type now. It was for this

> needle
Prior to PHP 8.0.0, if needle is not a string, it is converted to an integer and applied as the ordinal value of a character. This behavior is deprecated as of PHP 7.3.0, and relying on it is highly discouraged. Depending on the intended behavior, the needle should either be explicitly cast to string, or an explicit call to chr() should be performed.

Seeing as Psalm doesn't support 7.3 I think it's safe to ignore it?

orklah: Psalm doesn't run on PHP 7.3 but it can still analyze 7.1+ code anyway (it can work on older code too but there may be method signatures mismatches)

However, given the discouragement/deprecation/change for this, I think we should consider this as an error even for previous versions and only support string

orklah: You'll have to move the test that fails because of this to invalid tests case to ensure this throws an error now

orklah:
Thanks!

PR #9794, opened 2023-05-16
Improved type of callable-array

### LINKED ISSUE #7778: Stricter typing of callable-array
It seems like psalm is understanding a callable array as the equivalent of `array<array-key, mixed>` instead of `array{0: class-string|object, 1: string}`.

https://psalm.dev/r/95bfee547d

https://psalm.dev/r/0bdc7fda82

Is it possible to help psalm understand callable arrays a little better? Or am I missing something obvious/making an incorrect assumption about what makes a valid callable array?

robchett:
BC: Removes TCallableArray & TCallableList in favour of TCallableKeyedArray
Closes #7778

Also improves the type of callable-array so that it is interpreted as a `list{class-string|object, string}` instead of `non-empty-array<array-key, mixed>`


robchett:
@orklah what would be preferred here

  -'callable-array{0: class-string, 1: string}'
  +'callable-array{class-string, string}'

As this is now a list of size 2 I'm not sure which better defines the shape.

orklah:
Semantically, the second syntax implies that the order is locked while the first doesn't (meaning `[1=>'method', 0=>stdClass::class]` would be valid, but it shouldn't). So the second syntax is best

orklah:
I see TCallableArray being removed but I don't see TCallableList being removed as well, did I miss something?

robchett (commented):
TCallableList was removed upstream - remove BC note about it

robchett:
I rebased this branch last week as it was a long way behind. Looks like TCallableList was removed elsewhere upstream. Tracked that down to cca276768ef2f3e91e3763afdc32d16cc2b4da03, removed the note about that in the BC note

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2651,10 +2650,6 @@ private static function reconcileCallable(
             ) {
                 $callable_types[] = $type;
                 $redundant = false;
-            } elseif ($type instanceof TArray) {
```
orklah, line 2654: Shouldn't this create a TCallableKeyedArray now that TCallableArray doesn't exists?

src/Psalm/Internal/Type/TypeParser.php
```diff
@@ -1411,7 +1411,6 @@ private static function getTypeFromKeyedArrayTree(
                         !is_numeric($property_key)
                         || ($had_optional && !$property_maybe_undefined)
                         || $type === 'array'
-                        || $type === 'callable-array'
```
orklah, line 1414: why remove this? TCallableKeyedArray still have a `callable-array` key, no?

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
robchett: Yes, that would make more sense and I don't remember why I made that change to begin with. It doesn't appear to effect any of the tests either way which

src/Psalm/Internal/Type/TypeParser.php
robchett: I have no good reason why it was removed. Now back and tests are unaffected,

orklah:
Thanks!

PR #9795, opened 2023-05-16
Improve return types of pow()

### LINKED ISSUE #7218: `pow` written as `**`: FP
I think `**` should be the same as `pow` but I see different results:

https://psalm.dev/r/4c9f5290b6

https://3v4l.org/9MfYo

PS: Happy Christmas :christmas_tree:

robchett:
Fixes #7218


orklah:
Please keep the signatures in the Callmap files (it won't have any effect but for completeness purpose, it's good to keep those)

There's a few CI errors because you put a Union inside another Union `new Union([Type::getFloat()])` is redundant, `Type::getFloat()` is enough

orklah:
Great, there's just a few CS error to fix and good to go :)

robchett:
They are in bin/update-property-map.php which is not a file I've changed, I'm also up-to-date with master so not sure where these have come from.

orklah:
Yeah, probably a change in the tool upstream. That particular file was supposed to be ignored: https://github.com/vimeo/psalm/blob/master/phpcs.xml#L50

orklah:
Thanks!

orklah:
note: it's actually a legit fix upstream: https://github.com/slevomat/coding-standard/commit/019d946f1babedb9cfa8eaddec29cddea371e579 It was just never detected before

Issue #9804, opened 2023-05-21
True comparison on method of nullable object asserts object as non-nullable in elseif


tscni:
https://psalm.dev/r/b2ef61192d

Adding a null-safe operator to the first check doesn't help much: https://psalm.dev/r/ff20aa26bc
The issue disappears when the result of the method is compared against a non-true value with a null-safe operator: https://psalm.dev/r/2c16762b74

kylekatarnls:
Likely the same root cause, with this cause:
https://psalm.dev/r/fa5d15a8ee

`PossiblyNullReference` false positive is triggered (impossible because `$a` cannot be truthy if `$b === null`

Issue #9454, opened 2023-03-05
Incorrect type recognition


viktorprogger:
It's really nullable, while psalm says it's not.

https://psalm.dev/r/9a23dd53a0

PR #9806, opened 2023-05-22
Fix nullsafe method call assertions being reduced to !null in if conditions

tscni:
Fixes https://github.com/vimeo/psalm/issues/9804
The issue is more extensive. Any other assertion a method call would guarantee is lost with a nullsafe method call.

Still a draft, but feedback would be appreciated.
I'd like to add some more tests and I haven't yet verified whether the changes have unintended side-effects or if there's a better way. E.g. shouldn't this somehow be handled / helped by the `NullsafeAnalyzer` instead?


ygottschalk:
Have a look at https://github.com/vimeo/psalm/commit/fbc2b48f7a330c7d92826ce50d9885855822f48b

I was trying to tackle #9454 and my first idea was to rewrite nullsafe-access to virtual ternarys, but that had some problems with the short circuit functionality, if I remember correctly.

My next idea was to mark the first / last member of a nullsafe-access chain during the scanning phase and use that in the analyzing phase.

Dont remember how much of the work is done and what is left. It's been a while.

Let me know if I can help

Issue #5721, opened 2021-05-05
Support path wildcards for paths


samdark:
I have a PHP templates in the project. The project files are grouped by use-case:

```php
src
  ContactForm
    views
      form.php
    Form.php
    FormAction.php
   Blog
     Tag
       views
         tags.php
       TagAction.php
     Archive
       views
         archive.php
       ArchiveAction.php
```

Use-cases aren't fixed. They're added/deleted as development goes.

Currently I can not exclude `views` directories via `ignoreFiles` since wildcard pattern doesn't work for paths, see #4853. I'd like to be able to.

PR #9813, opened 2023-05-24
Add a support for multilevel glob wildcards

2e3s:
This can work like `glob.glob` in Python with `**` in the path, e.g. `src/**/BinaryOp*` would match
- src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
- src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp
- src/Psalm/Node/Expr/BinaryOp

If there are no `/**/` in the path then it works as before.

There is some backward incompatibility: PHP's `glob` treats `path/**/path` and `path/*/path` and even `path/***/path` as the same pattern, so if someone used `/**/` in their config file, they need to adjust it to avoid an excessively wide match.

Closes https://github.com/vimeo/psalm/issues/5721

### DIFF
--- a/src/Psalm/Config/FileFilter.php
+++ b/src/Psalm/Config/FileFilter.php
@@ -452,7 +452,13 @@ class FileFilter
             if ($onlyDir) {
                 return array_map('realpath', glob($parts[0], GLOB_ONLYDIR | GLOB_NOSORT));
             } else {
-                return array_map('realpath', glob($parts[0], GLOB_NOSORT));
+                return array_map(
+                    'realpath',
+                    array_filter(
+                        glob($parts[0], GLOB_NOSORT),
+                        'file_exists',
+                    ),
+                );
             }
         }
         $firstDir = self::slashify($parts[0]);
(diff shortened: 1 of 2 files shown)


orklah:
Cool! Can you take a look at the failing tests and the few issues in Shepherd and it should be good to go!

orklah:
I can reproduce on my windows env but I'm not sure what I should look at.

The error raised is this one: https://github.com/2e3s/psalm/blob/116795abc4ec7330728920d1c8161e70f44aace0/src/Psalm/Config/FileFilter.php#L268

Tell me if you want to know the state of anything in there

2e3s:
Thanks a lot for a helping hand! I don't have Windows anywhere, but I've just realized that `D:\a\psalm\psalm\**/**/**/**/*Analyzer.php` would be transformed into `D:\a\psalm\psalm\**/*Analyzer.php` which has no `/**/`.

I'll split the path differently.

orklah:
Great! Thanks!

PR #9814, opened 2023-05-24
glob: allow empty string for pattern

Hanmac:
Fixed problem with #9752

glob should be able to allow empty string for pattern

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -3329,7 +3329,7 @@ return [
 'gettimeofday' => ['array<string, int>'],
 'gettimeofday\'1' => ['float', 'as_float='=>'true'],
 'gettype' => ['string', 'value'=>'mixed'],
-'glob' => ['list<non-empty-string>|false', 'pattern'=>'non-empty-string', 'flags='=>'int<1, max>'],
+'glob' => ['list<non-empty-string>|false', 'pattern'=>'string', 'flags='=>'int<1, max>'],
 'GlobIterator::__construct' => ['void', 'pattern'=>'string', 'flags='=>'int'],
 'GlobIterator::count' => ['int'],
 'GlobIterator::current' => ['FilesystemIterator|SplFileInfo|string'],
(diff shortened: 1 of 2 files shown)


orklah:
Thanks!

Hanmac:
@kkmuffme that doesn't work
You can't predict if sprintf would return an empty or non empty string with the tools psalm provides (as far as I saw that parts with glob)

It's either your MR was merged too early

VincentLanglet:
> @orklah that merge was too early, bc this will hide tons of unwanted behavior (that's why I PRed the fix initially) Can you please "unmerge" (revert) it?

It's your PR which was merged to early. 

Your unwanted behavior are perfectly valid since the glob method does accept an empty string.
I have use case with empty string, which are not bugs. If you consider empty-string is an unwanted value, you can create your own glob method to restrict the param, or a more specific stub in your project.

> @Hanmac The solution is to create a return type provider for sprintf to return non-empty-string/non-falsy-string if the pattern contains anything but placeholders (or if only placeholders but they are non-empty, then non-empty too) [#9752 (comment)](https://github.com/vimeo/psalm/pull/9752#issuecomment-1562356244)

The sprintf example was only one example.

I can just write a method which does
```
function searchFirst(string $userInput): ?string
{
     return glob($userInput)[0] ?? null;
}
```
And there is no valid argument to force myself writing
```
function searchFirst(string $userInput): ?string
{
     if ('' === $userInput) {
        return null;
     }

     return glob($userInput)[0] ?? null;
}
```
Because I could use the same to ask you changing every `strlen` call from 
```
function computeLength(string $userInput): int
{
     return strlen($userInput);
}
```
to
function computeLength(string $userInput): int
{
     if ('' === $userInput) {
          return 0;
     }

     return strlen($userInput);
}
```

And there is tons of more complex example.

As soon as `glob('')` doesn't throw an exception, it's valid. There is nothing more to add.

Hanmac:
> The current PR has a bug though, see #9820 - @VincentLanglet or @Hanmac please add a PR with that.

No. your PR broke it first, YOU try to fix it without causing `glob(sprintf("%s", $val))` to fail

PR #9815, opened 2023-05-24
Fix redis stub for methods set, setnx, setex, sismember

nexor:
Error example: https://psalm.dev/r/a59428a767

Documentation: https://github.com/phpredis/phpredis/blob/552f41a44f0687c221e5d4adf9ee387479bb78a1/redis.stub.php#L382

### DIFF
--- a/stubs/extensions/redis.phpstub
+++ b/stubs/extensions/redis.phpstub
@@ -390,7 +390,7 @@ class Redis {
     public function select(int $db): bool {}
 
     /** @return bool|Redis */
-    public function set(string $key, string $value, mixed $opt = NULL) {}
+    public function set(string $key, mixed $value, mixed $opt = NULL) {}
 
 	/** @return false|int|Redis */
     public function setBit(string $key, int $idx, bool $value) {}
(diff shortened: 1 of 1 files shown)


orklah:
Could you check this comment https://github.com/phpredis/phpredis/pull/2015#issuecomment-1139509679 and tell us what you think? 

Should Psalm allow you to input an int if it's handled as a string anyway?

cc @kkmuffme

nexor:
> Should Psalm allow you to input an int if it's handled as a string anyway?

Yes I think(at least thought until now) it should, because official signature is `mixed $value`, not `string $value`
https://github.com/phpredis/phpredis/blob/develop/redis.stub.php#L2988

So the code:
```php
$redis->set('key', 1);
```
is absolutely correct from all points of view and psalm error looks here like "false positive".

Actually you guys made me doubt :)

Considering the fact that "official" `get()` signature assumes a return of mixed, 
https://github.com/phpredis/phpredis/blob/develop/redis.stub.php#L1427
but in fact it always returns string, I think it is a good idea to force using string as `set()`s input.

I'll close this PR and update my code. Thank you!

PR #9829, opened 2023-05-27
Mapping closed inheritance to union during assertion

### LINKED ISSUE #5248: False positive with assert-if-true/if-false

klimick:
I expected to see this in #9687, but unfortunately it was missed.
It's also closes #5248.

### DIFF
--- a/src/Psalm/Internal/Type/ClosedInheritanceToUnion.php
+++ b/src/Psalm/Internal/Type/ClosedInheritanceToUnion.php
@@ -1,0 +1,81 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Internal\Type;
+
+use Psalm\Codebase;
+use Psalm\Type\Atomic\TGenericObject;
+use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Union;
+
+use function array_keys;
+
+/**
+ * @internal
+ */
+final class ClosedInheritanceToUnion
+{
+    public static function map(Union $input, Codebase $codebase): Union
+    {
+        $new_types = [];
+        $meet_inheritors = false;
+
+        foreach ($input->getAtomicTypes() as $atomic_type) {
+            if ($atomic_type instanceof TNamedObject) {
+                $storage = $codebase->classlikes->getStorageFor($atomic_type->value);
+
+                if (null === $storage || null === $storage->inheritors) {
+                    $new_types[] = $atomic_type;
+                    continue;
+                }
+
+                $template_result = self::getTemplateResult($atomic_type, $codebase);
+
+                $replaced_inheritors = TemplateInferredTypeReplacer::replace(
+                    $storage->inheritors,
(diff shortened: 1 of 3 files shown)


orklah:
Thanks!

klimick:
I described the original problem rather badly.
Added a lot of unnecessary examples that poorly describe the essence of the problem.

The problem was that the psalm could not certain that the Either (or any other abstract class/interface) in the if branch is one value and the else branch is another. And this behavior of the psalm was quite safe.

Look at this example:

```php
/**
 * @template L
 * @template R
 */
interface Either {}

/**
 * @template R
 * @implements Either<empty, R>
 */
final class Right implements Either {}

/**
 * @template L
 * @implements Either<L, empty>
 */
final class Left implements Either {}

/**
 * @template L
 * @template R
 *
 * @param Either<L, R> $either
 * @return bool
 *
 * @psalm-assert-if-true Left<L> $either
 */
function isLeft(Either $either): bool
{
     throw new RuntimeException('???');
}

/**
 * @param Either<string, int>
 */
function test(Either $e): void {
    if (isLeft($e)) {
         // exactly Left<string> 
    } else {
        // Either<string, int> here.
        // Because we can extend/implement the Either at any point of the program (even at runtime)
    }
}
```

In the `else` branch should be `Right<int>`. But how achieve it? Needs somehow to tell Psalm that the `Either` can be extended/implemented by closed list of types. And `@psalm-inheritors` can do it: https://psalm.dev/r/408248f46f.

Thanks to `@psalm-inheritors` psalm can negates complex types same as for simple scalars: https://psalm.dev/r/c848406bc5

Regarding the multiple definition of the `psalm-assert-if-true`/`false`:
I don't know why psalm maintainers did it. I think this is bad design.
I always use only one `@psalm-assert-if-true` annotation. Due to the psalm assertion system it should be enough (I'm referring to the examples above).

@orklah what do you think about multiple assertion annotation?
I can't remember any compiler where it would be.

klimick:
@kkmuffme all you examples works If you leave only one assertion. Why is that? I don't know)

klimick:
> Multiple assertions can make sense, e.g. if you have array/list or multiple variables.

For example type `array<string, string> | list<string>`.
Currently, Psalm can't assert that some times a variable is `array<string, string>` and some times is `list<string>`, because types will be collapsed to `array<int | string, string>`: https://psalm.dev/r/d23a4022d9
For me it's looks like a bug. It's completely unrelated types. If the types would not collapse into one, then one assertion would be enough.

Can you provide your examples? I am curious.

PR #9832, opened 2023-05-27
base64_decode returns false only in strict mode

kamil-tekiela:
I am not sure if this is the right way to fix this.


orklah:
That could have been a way but stubbed functions take priority over the callmaps:
https://github.com/vimeo/psalm/blob/bd0ba6ee957a1789cf5c5fe647893bfe3bda2425/stubs/CoreGenericFunctions.phpstub#L1567

So you need to add a conditional return here instead

kamil-tekiela:
@orklah  Is this what you meant?

orklah:
yeah! Just FYI, you can do `@return ($strict is false ? string : false|string)` also. The template is implied

orklah:
Thanks!

PR #9833, opened 2023-05-28
Fix #9824 const enum self reference

### LINKED ISSUE #9824: Regression in psalm 5.10.0 on enum types
This example worked with 5.9.0 but does not starting with 5.10.0: https://psalm.dev/r/5fe87722dd

ygottschalk:
#9824


orklah:
Thanks!

Issue #9819, opened 2023-05-25
sprintf return type inconsistent


kkmuffme:
https://psalm.dev/r/439491af3d

There should be an error for `non-empty-string` in both cases, not only in the case with %d.

It seems there is some return type provider in psalm for this already, but I couldn't find it. @orklah could you point me to the file that has it?

orklah:
It's handled by the conditional return here: https://github.com/vimeo/psalm/blob/bd0ba6ee957a1789cf5c5fe647893bfe3bda2425/stubs/CoreGenericFunctions.phpstub#L1282

Issue #9818, opened 2023-05-25
(s)printf basic param type validation


kkmuffme:
When the first param of (s)printf is a literal string (otherwise not possible to validate in psalm):

https://psalm.dev/r/7749cadcab
- for %d (or %1$d) the corresponding param must be type int (as string will be returned as 0 and float as int)
- this applies not only to "d" but to `[bcdouxX]`
- if placeholder is any of `[eEfFgGhH]` type should be float
- if placeholder is `s` should be type string (if we want to be less strict allow int|float|string like now)

Would be the next step after https://github.com/vimeo/psalm/issues/9817

Issue #9817, opened 2023-05-25
(s)printf fatal error not reported when params don't match placeholders


kkmuffme:
When the first param of (s)printf is a literal string (otherwise not possible to validate in psalm):

https://psalm.dev/r/4be0b0cec6
The number of params - 1 must match the number of unique placeholders. Atm this reports as "no errors" while in reality this is a fatal error (ArgumentCountError)

PR #9841, opened 2023-05-30
sprintf basic non-empty-string return type provider

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/9819

This PR is a starting point for improving the sprintf return type and eventually validate the format, param types and param count. (see https://github.com/vimeo/psalm/issues/9817, https://github.com/vimeo/psalm/issues/9818)


orklah:
Thanks!

Issue #9830, opened 2023-05-27
PropertyNotSetInConstructor with abstract constructor


kkmuffme:
https://psalm.dev/r/f0411b4fa2

An abstract constructor cannot contain any code, therefore reporting `PropertyNotSetInConstructor` on the class does not make sense (line 7)
Btw without the abstract constructor it works correctly (no false positive error): https://psalm.dev/r/39534b5051

ygottschalk:
Consider this:
https://psalm.dev/r/0dec2365d4
I think the abstract constructor does set the property this way.

orklah:
@kkmuffme 
> what is the suggested/easiest way to check if the constructor is abstract or not (aka get the MethodStorage for the construct function)? Then I'll quickly PR this.

Yeah, that's the easiest way

PR #9844, opened 2023-05-30
PropertyNotSetInConstructor should not report for abstract constructors

kkmuffme:
since they do not have any code

Fix https://github.com/vimeo/psalm/issues/9830


orklah:
ignored_issues are for issues that are unrelated to the test (so you can ignore noise).

If you want to test that a specific issue IS NOT emitted, you put your test in the valid part and you make sure your test does not emit any non-ignored issue. If you want to test that a specific issue IS emitted, you put your test in the invalid part and you check that the error emitted is the one you want

orklah:
Thanks!

PR #9845, opened 2023-05-31
Shepherd: send a list of issues (instead of array with int keys)

lptn:
As result Shepherd will send an array is issues instead of object with "random" numeric keys

Right now shepherd sends cURL request with payload like:
```json
    "issues": {
        "1558": {
            "severity": "error",
            "line_from": 43,
            "line_to": 46,
            "type": "PossiblyUndefinedIntArrayOffset",
           ...
```

There is no sense of keeping these numeric keys like "1558". So, it should be like:

```json
    "issues": [
        {
            "severity": "error",
            "line_from": 43,
            "line_to": 46,
            "type": "PossiblyUndefinedIntArrayOffset",
           ...
```

### DIFF
--- a/src/Psalm/Plugin/EventHandler/Event/AfterAnalysisEvent.php
+++ b/src/Psalm/Plugin/EventHandler/Event/AfterAnalysisEvent.php
@@ -10,7 +10,7 @@ final class AfterAnalysisEvent
 {
     private Codebase $codebase;
     /**
-     * @var IssueData[][]
+     * @var array<string, list<IssueData>> where string key is a filepath
      */
     private array $issues;
     private array $build_info;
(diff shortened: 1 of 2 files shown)


orklah:
Seems like it created a conflict with another PR of yours.

Just to make sure, could you check that the random ID is not the offset of first char of the issue in the file? Psalm uses that now and then and I wouldn't want to break something that rely on that

lptn:
@orklah 
Thanks! In this particular case this key is just a regular array index:
image


This is how new elements pushed to this this array:

https://github.com/vimeo/psalm/blob/67648ac248e791d97e61ed756a0b726d015587a5/src/Psalm/IssueBuffer.php#L620-L621


In this particular case variable `$key` is not required as far as I understand and can be removed in order to not create confusing.
https://github.com/vimeo/psalm/blob/67648ac248e791d97e61ed756a0b726d015587a5/src/Psalm/IssueBuffer.php#L612

orklah:
Thanks!

PR #9848, opened 2023-05-31
Use PHPStorm output format by default on JetBrains terminals

lptn:
Goal: improve developer experience by using better defaults.

about `TERMINAL_EMULATOR` env var - it's tested on macOS and Ubuntu PHPStorm versions - works as expected.

We can also auto-detect other output formats, but firstly I would like to check this idea on this PR


orklah:
Thanks!

PR #9862, opened 2023-06-02
Mention --find-unused-code instead of the old option

HenkPoley:
Let's finish what was started.

Some people don't like being remembered about death.

And `psalm --help` calls it `--find-unused-code` anyways, since a while already.

### DIFF
--- a/docs/running_psalm/issues/PossiblyUnusedMethod.md
+++ b/docs/running_psalm/issues/PossiblyUnusedMethod.md
@@ -1,6 +1,6 @@
 # PossiblyUnusedMethod
 
-Emitted when `--find-dead-code` is turned on and Psalm cannot find any calls to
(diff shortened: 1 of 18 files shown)


HenkPoley:
I have no clue what a failure "Pull Request Labels / label (pull_request)" means.

Please consider helping, if that's a blocker somehow.

orklah:
Thanks!

It seems we only use the old option here: https://github.com/vimeo/psalm/blob/3271b1b8f37ca070f7ec6c9aa82f5845d2a1f58a/src/Psalm/Internal/Cli/LanguageServer.php#L380

Could you adapt to take both into account for now? And check if there's other places like that before we replace the documentation

(Note: the pull request check is for us, to remember us to classify the PR)

PR #9868, opened 2023-06-02
Update the signature of debug_backtrace

stof:
The file and line keys might be missing for frames. This happens when the frame are calls originating from native code.


stof:
See https://github.com/doctrine/deprecations/issues/44 for a case where this happened

orklah:
Seems like a lot changed in the Callmap file. Did you change the formatting maybe? (couldn't see through github as the diff is too big)

stof:
Apparently, the github editor decided to change all line endings to CRLF instead of using LF...

orklah:
Thanks! Tests are failing because you must make the same change in Callmap_historical (so we have a full verification across all maintained php versions)

stof:
@orklah updated with the historical Callmap

orklah:
Thanks!

Issue #9874, opened 2023-06-05
SprintfReturnTypeProvider crashes with unknown sprintf format


fancyweb:
Hello, SprintfReturnTypeProvider crashes when sprintf is used with an invalid format, eg:
```php
$a = 'a';
sprintf('foo "%" bar', $a);
```

https://psalm.dev/r/eef72f64c3

PR #9877, opened 2023-06-06
Sprintf improve return param type validation

kkmuffme:
Implement https://github.com/vimeo/psalm/issues/9817
Fix https://github.com/vimeo/psalm/issues/9874


orklah:
It is unrelated indeed. Mockery/mockery released a 1.6.1 tag yesterday that Psalm doesn't like, not sure why

orklah:
Seems like Mockery introduced Psalm in their own repo in that release. Either there's a weird interaction or something changed that I'm not getting.

Anyway, it does seems to only break tests cases that redefined the function `mock`. Could you change the name of that specific function in those tests? It should resolve the issue the easy way

orklah:
Order of priority goes like:
ReturnTypeProvider > Stubs > Callmap

ReturnTypeProvider only take priority if it doesn't return null. Stubs always take priority over Callmap

So when you removed the string return type from the stub, if you happen to return null in the ReturnTypeProvider, it just fails to infer

orklah:
> Isn't this a problem with consistency between PHP versions? e.g. PHP 7 can return false from sprintf, but PHP 8 cannot.

We have different options for that in stubs. We can use conditionals to check for php version in stubs or we can create a whole stub file that is conditionally loaded depending on php version

orklah:
Thanks!

PR #9878, opened 2023-06-06
Glob return type stub

### LINKED ISSUE #9820: glob wrong return type
https://psalm.dev/r/726b2ba5b5

Result is not `list<non-empty-string>` when the argument is empty string and no flags are present. (but should be `array<never, never>`)
Glob needs a return type provider added to ensure the "non-empty-string" is not lost when a valid pattern is passed.

Bug added in https://github.com/vimeo/psalm/pull/9814 @Hanmac

ygottschalk:
Fixes #9820
Discussion is in #9820


ygottschalk:
IMO Unit test have nothing to do with this PR
There might be style issues introduced though...

PR #9865, opened 2023-06-02
Intersect template types during inheritance check

klimick:
Fix regression after #8926
Closes #8998
Closes #9245

Issue #6316, opened 2021-08-16
Strange InvalidReturnType/InvalidReturnStatement since 4.9.3


vudaltsov:
https://psalm.dev/r/30e1b04cd7

orklah:
Adding phpdoc manually solve the issue: https://psalm.dev/r/31dad97d4d

WossomGonzo:
I have a similar code that in Psalm 4.9.3 it runs ok, but in Psalm 4.10 it returns:

`ERROR: InvalidReturnType - index.php:23:30 - The declared return type 'array<array-key, mixed>' for multiply is incorrect, got 'array<array-key, mixed>|object' (see https://psalm.dev/011)
function multiply($numbers): array {`

`ERROR: InvalidReturnStatement - index.php:27:12 - The inferred type 'array<array-key, mixed>|object' does not match the declared return type 'array<array-key, mixed>' for multiply (see https://psalm.dev/128)
    return $numbers;
`



```php
<?php
require 'vendor/autoload.php';

class Num {
    public int $value;

    public function __construct(int $value) {
        $this->value = $value;
    }
}

$numbers = [
    new Num(1),
    new Num(2),
    new Num(3),
    new Num(4),
    new Num(5),
];

$numbers = multiply($numbers);
printThem($numbers);

function multiply($numbers): array {
    array_walk($numbers, function($number) {
        $number->value = $number->value * $number->value;
    });
    return $numbers;
}

function printThem($numbers): void {
    print_r($numbers);
}
```

orklah:
@WossomGonzo your issue is related to #6441 and has been fixed. The original issue here is not yet identified

klimick:
Issue resolved. Can be closed.

vudaltsov:
Thank you, @klimick !

PR #9883, opened 2023-06-07
Better intersection of template types during inheritance check

klimick:
Looking back to #9865, it seems to me that I did some stupid thing.
Issue #6316 should be fixed too but my previous PR didn't it.

This PR reverts changes from #9865 and indroduce better fix (as it seems to me)


orklah:
Thanks!

Issue #9831, opened 2023-05-27
NonInvariantDocblockPropertyType when type extends declared class


kkmuffme:
https://psalm.dev/r/7d9f795223

Maybe I am misunderstanding, but why does this give an error?
Since `Foo extends Abstract_Foo` this shouldn't be an issue.

ygottschalk:
https://psalm.dev/docs/running_psalm/issues/NonInvariantDocblockPropertyType/

Add ing to your example:
```php
class NotFoo extends Abstract_Foo {}
function takesA (A $a) { $a->s = new NotFoo(); }

$b = new B( new Foo () );
takesA($b);
```
Now `$b->s` is typed `Foo` but is actually `NotFoo`

orklah:
> So technically doing an intersection of generic object&Abstract_Foo should fix the issue though?

I don't understand what you meant there

Normally, having a different type on a property in a child that in the parent make no sense because you have read AND write access on it (so depending on the context of what you know of the object, you'd be allowed to write a more general value and retrieve a more specific one)

I guess that if the property is readonly, it could change things but it seems weird to me

PR #9887, opened 2023-06-09
readonly does not have write access, therefore is safe as long as the…

kkmuffme:
… type is contained in the parent type

Fix https://github.com/vimeo/psalm/issues/9831#issuecomment-1569022701


orklah:
Thanks!

PR #9889, opened 2023-06-09
feat: cache.gz

### LINKED ISSUE #8495: Compress cache using gzip
Using gzip, cache could be stored more efficiently space wise. For reference, my PHPStan cache folder is 16MB, my Psalm folder is 385MB on a project, they run in about the same time.

[Symfony uses gzinflate / gzdeflate to decorate the cache marshaler](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Cache/Marshaller/DeflateMarshaller.php), Psalm could do the same. It provides quite an improvement over nothing or igbinary, gzip features are also bundles in PHP while igbinary might require additional extensions.

cc @bendavies @orklah

### LINKED ISSUE #8679: Corrupted cache files shouldn't interrupt execution
For some "out of disk space" circumstances my Psalm wrote to cache but couldn't do it fully.

When I cleared up the storage, it still failed:
```
Uncaught Exception: RuntimeException PHP Error: unserialize(): Error at offset 262140 of 262144 bytes in /tools/.composer/vendor-bin/psalm/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageCacheProvider.php:145
```

The cache file itself was corrupted, but this shouldn't interrupt execution, it can just be ignored like it doesn't exits.

dkarlovi:
Closes #8679
Closes #8495

# Benchmark

## Gzip off

Cache folder size: 378M

```
# cache empty run
Checks took 20.86 seconds and used 808.750MB of memory

# cache primed run
Checks took 0.17 seconds and used 66.564MB of memory
```

## Gzip on

Cache folder size: 47M

```
# cache empty run
Checks took 22.24 seconds and used 808.750MB of memory

# cache primed run
Checks took 0.66 seconds and used 74.394MB of memory
```


dkarlovi:
@orklah @weirdan could you add a label to this PR so it's not spamming me with failures?

dkarlovi:
Ready for review, the build passes but the label is missing and I don't know what Circle is on about, it's finding errors in code which doesn't exist, I can't reproduce it locally, `composer cs` passed cleanly for me.

orklah:
Added the label :)

Please make sure to update your dependancies before running the CS fix, they improved detection a few versions ago, you may still be stuck with an old one

(anyway, the error is legit, you're missing a use function)

dkarlovi:
I'm updated fully, this is a fresh checkout from today and I just tried to `composer update -W`, it's all up to date and `cs-fix` misses this. I've added it manually.

dkarlovi:
@orklah green \o/

src/Psalm/Internal/Cache.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->config = $config;
+        $this->use_igbinary = $config->use_igbinary;
+    }
+
+    /**
+     * @return array|object|string|null
+     */
+    public function getItem(string $path)
+    {
+        if (!file_exists($path)) {
```
faizanakram99, line 40: `is_file` is faster than `file_exists`, PHP does some opcache optimizations with former

orklah:
Should we put a config to opt in/out of gzip?

I get that cache size can be an issue for some situations, but runs on CI are likely more limited on CPU usage or memory usage rather than disk space, no?

dkarlovi:
@orklah IMO the performance impact is negligible enough to allow this to be the default, you could opt out via config, but at 5% difference is it worth it?

orklah:
Well, my work codebase make Psalm run with 8.5Gb of memory. Even 5% of that is not really negligible when I have to run it on my workstation

My specific example is not relevant because I can't even run cache at all (I have to disable it, otherwise Psalm never ends) but I think others may be in similar cases

dkarlovi:
IMO it would make more sense to actually fix memory usage there instead of adding dozens of tiny tweaks in config.

dkarlovi:
Quickly checked, the majority of memory spent when cache is not in use is by `Psalm\Internal\Provider\StatementsProvider::parseStatements` (65%) which in turn runs PHP Parser which itself eats most of the memory (50%). It's quite possible the memory doesn't get cleared properly here and usage always just increments.

The analyze step takes about 25% of memory. This is with cache disabled and a single thread.

dkarlovi:
@orklah can you try what happens on your codebase if you reduce `\Psalm\Internal\Provider\StatementsVolatileCache::__construct` max size to 1?

src/Psalm/Internal/Cache.php
dkarlovi: I can't see any difference one way or another.

orklah:
Discussion about memory is continuing on #9899.

If this change is causing performance issues in the future, we'll always be able to add a config then, I'll merge it as is for now

PR #9903, opened 2023-06-12
Fixes #9827

### LINKED ISSUE #9827: [Bug] Union of empty-array and false is false
https://psalm.dev/r/101d23c155

ygottschalk:
Fixes #9827

### DIFF
--- a/src/Psalm/Internal/Type/TypeCombiner.php
+++ b/src/Psalm/Internal/Type/TypeCombiner.php
@@ -130,7 +130,9 @@ class TypeCombiner
         if (count($combination->value_types) === 1
             && !count($combination->objectlike_entries)
             && (!$combination->array_type_params
-                || $combination->array_type_params[1]->isNever()
+                || ( $overwrite_empty_array
+                    && $combination->array_type_params[1]->isNever()
+                )
             )
             && !$combination->builtin_type_params
             && !$combination->object_type_params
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Type/TypeCombiner.php
```diff
@@ -130,7 +130,9 @@ public static function combine(
         if (count($combination->value_types) === 1
             && !count($combination->objectlike_entries)
             && (!$combination->array_type_params
-                || $combination->array_type_params[1]->isNever()
+                || ( $overwrite_empty_array
```
orklah, line 133: I feel like $overwrite_empty_array should also apply to `!$combination->array_type_params`

So it would end up as
```
&& ($overwrite_empty_array && (!$combination->array_type_params || $combination->array_type_params[1]->isNever()))
```
or something

ygottschalk: I thought `((bool) $combination->array_type_params) === false` means, there was no (unsealed/generic) `array` type given to combine with. So that case was for me "no array" instead of "empty array". Thats why I decided to not include that in the check...

ygottschalk: I put another PR with the version you suggested (#9909) but (for reasons given above) I feel like this is the right version.

orklah:
Thanks!

PR #9916, opened 2023-06-13
docblock parser minor performance gain

kkmuffme:
Profiled psalm and reduced preg_ calls in DocblockParser to save between 0-5% runtime


orklah:
Thanks!

PR #9920, opened 2023-06-17
Add a stub file for Attributes defined by PHP

robchett:
Adding a stub so that running Psalm in a PHP environment that doesn't support a built-in Attribute eg 8.0 with #[AllowDynamicProperties]

Should be fine until php-src introduces a new attribute the conflicts with a userland class...

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -2236,6 +2236,8 @@ class Config
         ];
 
         if ($codebase->analysis_php_version_id >= 8_00_00) {
+            $stringable_path = $dir_lvl_2 . DIRECTORY_SEPARATOR . 'stubs' . DIRECTORY_SEPARATOR . 'CoreGenericAttributes.phpstub';
+            $this->internal_stubs[] = $stringable_path;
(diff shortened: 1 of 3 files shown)


stubs/CoreGenericAttributes.phpstub
```diff
@@ -0,0 +1,19 @@
+<?php
+
+#[Attribute(Attribute::TARGET_METHOD)]
+final class ReturnTypeWillChange
```
orklah, line 4: https://github.com/vimeo/psalm/blob/eb64a471c4eba7886d59703695853591335aee2f/stubs/CoreGenericClasses.phpstub#L529

ReturnTypeWillChange is already here. Probably because we need to use it before php 8

src/Psalm/Config.php
```diff
@@ -2236,6 +2236,7 @@ public function visitStubFiles(Codebase $codebase, ?Progress $progress = null):
         ];
 
         if ($codebase->analysis_php_version_id >= 8_00_00) {
+            $this->internal_stubs[] = $dir_lvl_2 . DIRECTORY_SEPARATOR . 'stubs' . DIRECTORY_SEPARATOR . 'CoreGenericAttributes.phpstub';
```
orklah, line 2239: Can you split the line in two here? CS check doesn't like that!

orklah:
Thanks!

PR #9921, opened 2023-06-17
set type of  __LINE__ to positive int

robchett:
Also adds the helper `Type::getIntRange($min, $max);`

### DIFF
--- a/tests/ConstantTest.php
+++ b/tests/ConstantTest.php
@@ -122,7 +122,7 @@ class ConstantTest extends TestCase
                     $a = __LINE__;
                     $b = __file__;',
                 'assertions' => [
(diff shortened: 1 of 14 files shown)


tests/ConstantTest.php
```diff
@@ -122,7 +122,7 @@ function defineConstant() {
                     $a = __LINE__;
                     $b = __file__;',
                 'assertions' => [
-                    '$a' => 'int',
+                    '$a' => 'positive-int',
```
orklah, line 125: you'll have to use `int<1, max>` here. positive-int is only used as a read alias, it's not a type Psalm can output

orklah:
Thanks!

Issue #9894, opened 2023-06-11
Wrong TypeDoesNotContainNull determination


chriskapp:
Hi, I have found also another case where Psalm thinks that in the example the `$existing` variable does not contain `null` s.

https://psalm.dev/r/74b5ec5ba5

But it could be `null` in case `$foo` is `1` and `$bar` is `null`. It looks like Psalm does not recognize the `elseif` and thinks that there is a standalone `if` which would indeed remove the `null` type.

Maybe this is also related to #7099

chriskapp:
If the `elseif` block is moved into a separate `if` inside an `else` it works like expected s.
https://psalm.dev/r/d274855806

theodorejb:
Is this issue still relevant? It looks like it was fixed.

weirdan:
I believe this is fixed in #9922

PR #9922, opened 2023-06-17
Don't update types of the outer scope when in an elseif conditional

robchett:
This feels extreme for https://github.com/vimeo/psalm/issues/9894 but I'm struggling to come up with a situation where the condition in an elseif can safely modify the outer scope unless the `elseif` could actually just be an `if` (the if branch was an early return).

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Block/IfElse/ElseIfAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Block/IfElse/ElseIfAnalyzer.php
@@ -358,16 +358,6 @@ class ElseIfAnalyzer
                         $elseif_context->inside_loop,
                         new CodeLocation($statements_analyzer->getSource(), $elseif, $outer_context->include_location),
                     );
-
-                $updated_vars = [];
-
-                $outer_context->update(
-                    $elseif_context,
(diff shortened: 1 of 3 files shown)


orklah:
Could you investigate the CI failure in Shepherd?

The fact that the only failure is in a test that is unrelated to if/else is both concerning and encouraging. Concerning because you'd expect that kind of feature to be covered by at least one test, but if it's not, it may means it's useless !

robchett:
Looks like the Shepard failure is surfacing an error as the result of the bug fix

``` 
            $classNameEnd = strpos($function, '::');
            if ($classNameEnd !== false) {
                $className = substr($function, 0, $classNameEnd);
                if (!class_exists($className, false)) {
                    continue;
                }
            } elseif (!function_exists($function)) {
                continue;
            }

            yield "$function: " . json_encode($entry) => [$function, $entry];
```
The `if` doesn't verify that $function is a callable-string, and I'm not convinced they are callable-strings as the methods are not all static. 
The usages of this method don't appear to require callable-strings anyway so I've corrected the return to simply `string`

orklah:
Let's keep that one in mind, it's rare that removing code just works, but given we have no tests, any feedback will have the advantage of giving a useful test case

PR #9924, opened 2023-06-18
make compressor configurable v2

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/9905
Suppress throws for igbinary_unserialize like it was done for @unserialize

Cherry picked from https://github.com/vimeo/psalm/pull/9914

PR #9925, opened 2023-06-18
add back invalid cache reporting

kkmuffme:
which was removed in #9889 - now we won't report for every error but if there are at least 10 errors (chosen arbitrarily)

Cherry picked from https://github.com/vimeo/psalm/pull/9914

### DIFF
--- a/src/Psalm/Internal/Cache.php
+++ b/src/Psalm/Internal/Cache.php
@@ -29,6 +29,8 @@ class Cache
 
     public bool $use_igbinary;
 
+    protected int $error_count = 0;
+
     public function __construct(Config $config)
     {
         $this->config = $config;
@@ -63,7 +65,7 @@ class Cache
 
         // invalid cache data
         if ($inflated === false) {
-            $this->deleteItem($path);
+            $this->handleInvalidCache($path);
 
             return null;
         }
@@ -77,12 +79,28 @@ class Cache
         }
 
         if ($unserialized === false) {
-            $this->deleteItem($path);
+            $this->handleInvalidCache($path);
 
             return null;
         }
 
         return $unserialized;
+    }
+
+    protected function handleInvalidCache(string $path): void
+    {
+        $this->deleteItem($path);
+
+        $this->error_count++;
+
+        // if 10 previous items were invalid, abort since the cache is invalid and inform the user
+        // we don't report it to the user immediately, since it can happen that a few files get corrupted somehow
+        // however the impact on performance is minimal, therefore we ignore it
+        if ($this->error_count > 10) {
+            throw new RuntimeException(
+                'The cache data is corrupted. Please delete the cache directory and run Psalm again',
+            );
+        }
     }
 
     public function deleteItem(string $path): void


ro0NL:
IMHO cache should be disposable

ygottschalk:
> I think it's a bad idea that some cache errors are ignored while many (most?) are not.

`if ($this->error_count > 10)`

Why then ignore 10 more errors...?


After reading through the discussion in the original PR, I felt like there are four things important for the ppl discussing (in no particular order):
1. User should be informed about issues with cache
2. Cache should be disposable
3. Handling of cache errors should be consistent
4. Cache should be self-healing / repairing

(Not sure if I do understand 'disposable' in this context correctly, but I assume it is smth along the lines of 'if not there / corrupted do not fail execution')

I 100% agree with 1 and 4.
For 1 IMO:
- Emit an information that cache is not found / corrupted or has self-repaired itself when `--debug` or `--verbose` given, else just be silent.

ro0NL:
so it was intendted to be disposable, and this pr reverts that? why not focus on fixing the remaning cases here? instead of saying others should do it.

> (Not sure if I do understand 'disposable' in this context correctly, but I assume it is smth along the lines of 'if not there / corrupted do not fail execution')

@ygottschalk yes, and it's a core semantic of cache IMHO

@kkmuffme sorry if i annoy you :pleading_face: im just invested in psalm/GHA and cache :)

dkarlovi:
The intent of #9889 was to make the cache not absurdly large and, while doing so, a sizeable refactor was required and #8679 was fixed in the refactor. The intent of the latter was to uniformly make it forgiving / self healing and that might have been missed in some specific cases which you've noticed.

The direction here should be to fix those remaining inconsistencies, not to undo the #8679 fix. @orklah made a good call to separate these two discussions in #9914 because now the intent of this PR is more clearly visible and I can say :-1: on it without dismissing the compression configuration I had no issues with.

ygottschalk:
> This PR is not to discuss about ....

Sorry for making this a discussion....

> I wouldn't necessarily report only when --debug is set, but generally report these (in stderr?), as otherwise you might not be aware that there is an issue with cache at all.

My thoughts were focused around a clean output but having the option to investigate when someone notices a problem. My vote would go to having the issues hidden unless `--debug` / `--verbose` is set, but I am fine with the other option, too. Was just worried that every time I change some files I get like 1,000 warnings about 'Cache miss file xyz.php. Generating cache.' or even worse in case of a first run getting warnings about every file in the project. That is something we should avoid...

But while we are at adding some warnings: I would like to have some warnings to show up (not errors failing execution and in this case show them always regardless of `--debug` / `--verbose`) for the cases where config says 'use compressor lz4 / gzip' but the required extensions / functions are not present in the environment. Same for 'serilizer = igbinary' and no `ext-igbinary` present. I would expect changes to be made here: https://github.com/vimeo/psalm/pull/9924/files#diff-8d07f6a86af4900069be3c80774a29e2e7e2ffaffe434c781ba95de859cf874dR1192-R1207

dkarlovi:
I agree output shouldn't be swamped with cache-related output by default, they're transitive and should be there only if I requested them. Focus of Psalm is not to debug its cache, it's to warn the user about errors in their code.

Having Psalm spew out what's basically infrastructural notifications in the same stream actual error messages which the user is actually interested in (which is exactly the reason why they're using it to begin with) is like you're working as maintenance in a hotel and then run around yelling at guests a lightbulb in lobby #2 burnt out. They don't care, it's not why they're here, you should care because that's your job, fix it and move on.

orklah:
My two cents on that:
- If I have to choose between two solutions, I prefer having a cache that doesn't crash Psalm when unavailable
- However, it should be consistent and be applied everywhere
- I can understand why @kkmuffme would not want to work on a feature he doesn't agree with
- I can perfectly understand @dkarlovi not having time to work on it right now

So, while it's not my first preference, I'd be inclined to merge this, unless someone tells me he wants to work on the opposite direction.

It's fine, we can always change it later if it still an itch we want to scratch, but at least in the meantime, it will be consistent

Any disagreement?

orklah:
Yeah, more than current master but not more than the last release if I understood correctly. And cache corruption doesn't happen in the wild that often so it doesn't seem that big a deal for now

orklah:
yeah, that's what I was saying, your PR was not yet released so it would make no difference with the current stable release.

As I said above, I'm okay with reducing errors, but only if we're consistent

dkarlovi:
But this PR doesn't increase consistency. 🤔

It's up to you, if you merge here, make sure to reopen the issue about failing cache reads crashing Psalm because this PR re-breaks that.

orklah:
@kkmuffme just for clarity, could you put here examples of where a corrupted cache will make Psalm crash on current master?

ro0NL:
Here, take this.

```
Uncaught TypeError: Psalm\Internal\Analyzer\StatementsAnalyzer::analyzeStatement(): Argument #2 ($stmt) must be of type _HumbugBox0d94293eab9c\PhpParser\Node\Stmt, __PHP_Incomplete_Class given, called in phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php on line 167 and defined in phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php:246
Stack trace:
#0 phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Analyzer/StatementsAnalyzer.php(167): Psalm\Internal\Analyzer\StatementsAnalyzer::analyzeStatement(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(__PHP_Incomplete_Class), Object(Psalm\Context), NULL)
#1 phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Analyzer/FileAnalyzer.php(140): Psalm\Internal\Analyzer\StatementsAnalyzer->analyze(Array, Object(Psalm\Context), NULL, true)
#2 phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Codebase/Analyzer.php(1088): Psalm\Internal\Analyzer\FileAnalyzer->analyze()
#3 phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Codebase/Analyzer.php(379): Psalm\Internal\Codebase\Analyzer->analysisWorker(0, '/app/bin/consol...')
#4 phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Codebase/Analyzer.php(222): Psalm\Internal\Codebase\Analyzer->doAnalysis(Object(Psalm\Internal\Analyzer\ProjectAnalyzer), 1)
#5 phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(375): Psalm\Internal\Codebase\Analyzer->analyzeFiles(Object(Psalm\Internal\Analyzer\ProjectAnalyzer), 1, false, true)
#6 phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Cli/Psalm.php(272): Psalm\Internal\Analyzer\ProjectAnalyzer->check('/app/', false)
#7 phar:///usr/local/etc/tools/vendor/psalm/phar/psalm.phar/psalm(7): Psalm\Internal\Cli\Psalm::run(Array)
#8 /usr/local/etc/tools/vendor/psalm/phar/psalm.phar(14): require('phar:///usr/loc...')
#9 /usr/local/etc/tools/vendor/bin/psalm(120): include('/usr/local/etc/...')
#10 {main}
(Psalm 5.12.0@f90118cdeacd0088e7215e64c0c99ceca819e176 crashed due to an uncaught Throwable)
```

It was caused by swapping src to phar, and solved by wiping previous cache :roll_eyes: aka manual labor.

ygottschalk:
I do not have a lot of time but I put #9932 out there.

It is just a small starting point, but I think that is the direction to go.

Can fix (if wanted) the case given here https://github.com/vimeo/psalm/pull/9925#issuecomment-1598363798, just leave some feedback  or a👍 on the relevant comment in PR.

Would be nice if someone can confirm I did not miss anything relevant....

PR #9929, opened 2023-06-19
improve perfs for switch by not creating reverse assertions against constants

### LINKED ISSUE #9926: Hanging when using a big case statement of constants in a foreach $key => $value
When looping over an array using `foreach ($array as $key => $value)` with a large case statement of constants it seems to hang for a long time. Might be something related to #6741 or maybe not but it was the closest issue I could find to this.

It works fine for scalar values like strings, break the number of cases up into smaller blocks and also works if you're looping over values using `foreach ($array as $value)` but not in this case. 

I have a reproduction case: https://psalm.dev/r/4d3d94a149

orklah:
Fix #9926

this reformats things for symetry and clarity but the relevant part of this is
`$other_var_name[0] === '$' //don't try to assert on constant or other, it's too perf consuming`

A few years ago, I added a feature in Psalm that ensures that `$a == $b` asserts that `$a` is the same type as `$b` (as it was before) but also asserts that `$b` is the same type of `$a` (it's a bit more complicated than that because Psalm choose `$a` and `$b` not only by position)

The thing is, switch in Psalm are transformed into giants balls of OR conditions that create an exponential list of Clauses to be asserted. This PR ensure that we don't create the reverse assertion when we don't have a variable. This will probably make some pieces of code less inferable but it turns the example of #9926 from 30 seconds or so to less than 2 seconds.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -3475,56 +3475,53 @@ class AssertionFinder
             throw new UnexpectedValueException('$typed_value_position value');
         }
 
-        if ($var_name && $var_type) {
-            $identical = $conditional instanceof PhpParser\Node\Expr\BinaryOp\Identical
-                || ($other_type
-                    && (($var_type->isString(true) && $other_type->isString(true))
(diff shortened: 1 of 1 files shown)


orklah:
CI failure is a new issue found by Psalm on Psl (one of the match branch is already handled above). Possibly because match now generate less Clauses and so it gets analyzed completely and it let Psalm find the issue

PR #9931, opened 2023-06-19
Invalidate cached methods when referenced files are deleted

tscni:
There's currently no invalidation mechanism for methods referencing another file (e.g. by using a symbol from it),
Example:
1. File `A.php` with method `A::do()` has an argument of type `B`, which is a class defined in file `B.php`
2. Psalm runs and caches
3. File `B.php` is deleted
4. Psalm runs and doesn't report an issue


orklah:
Thanks!

PR #9932, opened 2023-06-20
Fix/cache fail silent

ygottschalk:
see discussion in #9925

### DIFF
--- a/src/Psalm/Internal/Cache.php
+++ b/src/Psalm/Internal/Cache.php
@@ -4,7 +4,6 @@ namespace Psalm\Internal;
 
 use Psalm\Config;
 use Psalm\Internal\Provider\Providers;
-use RuntimeException;
 
 use function file_exists;
 use function file_put_contents;
@@ -115,13 +114,10 @@ class Cache
             $compressed = $serialized;
         }
 
-        if ($compressed === false) {
-            throw new RuntimeException(
-                'Failed to compress cache data',
-            );
+        if ($compressed !== false) {
+            file_put_contents($path, $compressed, LOCK_EX);
         }
-
-        file_put_contents($path, $compressed, LOCK_EX);
+        // TODO: Error handling
     }
 
     public function getCacheDirectory(): ?string
--- a/src/Psalm/Internal/Cache.php
+++ b/src/Psalm/Internal/Cache.php
@@ -68,7 +68,7 @@ class Cache
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


ygottschalk:
https://github.com/vimeo/psalm/pull/9925#issuecomment-1598363798 will not be catched by this. For fixing that, we would need to check the unserialized array to not contain `__PHP_Incomplete_Class` for `unserialize` (and whatever the equivalent for `igbinary_unserialize` might be)

dkarlovi (approved):

ygottschalk:
> 1. Errors when failing to write to cache should be ignored too though?

> > Throw if a serious error happened (eg psalm is explicitly configured to use cache but cache dir is not writeable)

> I don't think so. Either it's disposable or not; otherwise we will end up in a discussion on what is "serious" and what is not? e.g. is a E_WARNING serious? (when we (un)serialize/compress)

...?!


> 2. use grep 'throw .*cache' on psalm's codebase to find additional places that you missed

only thing I could identify beeing relevant is `\Psalm\Internal\Provider\ClassLikeStorageProvider::exhume` which considers things to be in cache but I could not confirm that there is a check confirming that....

dkarlovi:
IMO we can add a few more cases if they're obvious, but shouldn't block this PR until we reach 100% completeness, Psalm's cache handling was a total mess, repeated in 10 different places before #9889 and nobody cared about it then, now that we've significantly moved the needle, we've all become code gourmands, nothing but the very best is even worth considering. :)

Let's be pragmatic and see this merged, the further inconsistencies (if any) can easily be fixed as discovered, and the found inconsistencies will actually be places I've missed in #9889 so they'll work exactly like the previous release, not improving, but not breaking either.

ygottschalk:
[This](https://github.com/vimeo/psalm/pull/9925#issuecomment-1598363798) should get fixed (and in here is a possible solution: https://wiki.php.net/rfc/improve_unserialize_error_handling), but that can happen in a different PR.

I would really like to have some form of output for the write error or keep throwing....

ro0NL:
> https://github.com/vimeo/psalm/pull/9925#issuecomment-1598363798

note it's no blocker for us, just the first example i encountered, and it's probably an edge case

i agree with @dkarlovi, if we cant solve this generically, then baby steps it is

orklah:
> Let's be pragmatic and see this merged

agreed :)

Thanks!

PR #9935, opened 2023-06-20
Make config warn about missing cache related functions

ygottschalk:
Added warnings since I got a 👍 [here](https://github.com/vimeo/psalm/pull/9925#issuecomment-1597079461)


orklah:
Thanks!

Issue #9918, opened 2023-06-15
Expected psalm suppression time?


ro0NL:
Hi,

We're running psalm in GitHub actions.

On our develop branch we suppress any issue :grimacing: 

The command running is: `psalm --no-progress --output-format=github --set-baseline=psalm-baseline.xml --update-baseline`

Initially ~40 MB of cache was successfully restored.

The runtime takes about ~2m, for a commit that didnt do any code changes. It's actually about the same time a commit with few code changes takes.

Is this expected, or am i missing something?

orklah:
Unfortunately, I have no answer to your main question. However, i can say that --set-baseline makes --update-baseline redundant (the first one recreate the baseline completely while the second one only adds new issues in the existing baseline). Maybe using only the one you need would change something?

ro0NL:
unfortunately it doesnt make any difference, but removed --update-baseline either way 

the bottleneck is definitely --set-baseline, since without it but with the same cache i get instant errors after modifying a file

so we need 2m for some baseline file IO :thinking:

ro0NL:
it seems the cache is just ignored/invalidated when using --set-baseline

looking at --debug output, i see a lot of `Getting+Analyzing` with --set-baseline, but mostly `Have populated` without it

ro0NL:
im gonna assume things happen for some reason

Issue #9884, opened 2023-06-08
UndefinedFunction with <extraFiles>


ro0NL:
This seems related to #5791 

I have this file structure:

```
# src/file.php
<?php foo();

# src/inc/file.php
function foo() { return 1; }
```

And this psalm config:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<psalm
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="https://getpsalm.org/schema/config"
    xsi:schemaLocation="https://getpsalm.org/schema/config"
    cacheDirectory="var"
>
    <projectFiles>
        <directory name="src"/>
    </projectFiles>
    <extraFiles>
        <directory name="src/inc"/>
        <file name="src/inc/file.php"/>
    </extraFiles>
</psalm>
```

Running psalm produces unexpected error:

```
$ psalm        
Warning: "findUnusedBaselineEntry" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.
Warning: "findUnusedCode" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.
Warning: "findUnusedBaselineEntry" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.
Warning: "findUnusedCode" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.
Target PHP version: 8.1 (inferred from current PHP version).
Scanning files...
Analyzing files...

E░

ERROR: UndefinedFunction - src/file.php:3:1 - Function foo does not exist (see https://psalm.dev/021)
foo()


------------------------------
1 errors found
------------------------------

Checks took 1.29 seconds and used 176.077MB of memory
Psalm was unable to infer types in the codebase

$ psalm -v
Psalm 5.12.0@f90118cdeacd0088e7215e64c0c99ceca819e176
```

It works with `<stubs><file name="src/inc/file.php"/></stubs>` but this does not scale. AFAIK directories are not supported. Also, the file is not a stub but actual code.

The reason there's no `require` in code is because in runtime files are requried at higher level already, and i don't want to touch it.

orklah:
Hey @ro0NL sorry for not answering before.

Did you find a solution on your own for this issue?

ro0NL:
@orklah added everything to `<stubs>` :)

Issue #7681, opened 2022-02-16
Wrong errors after second run with cache


ilyaplot:
When I run psalm with cached results after editing any file I see UndefinedClass, UndefinedDocblockClass and other errors.
This bug reproduced in all my projects on Yii2 framework.

Psalm version is 4.20.0, yii2 framework, php 7.4.

Demonstration: https://www.loom.com/share/aea22198e3374afeaf57f175a9c877dd

My psalm.xml:
```<?xml version="1.0"?>
 <psalm
        errorLevel="5"
        resolveFromConfigFile="true"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xmlns="https://getpsalm.org/schema/config"
        xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
>
    <projectFiles>
        <directory name="./"/>
        <ignoreFiles>
            <directory name="backend/runtime"/>
            <directory name="console/runtime"/>
            <directory name="console/migrations"/>
            <directory name="console/controllers"/>
            <directory name="vendor"/>
            <directory name="environments"/>
            <file name="requirements.php"/>
        </ignoreFiles>
    </projectFiles>
    <ignoreExceptions>
        <class name="yii\web\HttpException"/>
        <class name="yii\web\NotFoundHttpException"/>
        <class name="yii\web\BadRequestHttpException"/>
        <class name="yii\web\ForbiddenHttpException"/>
    </ignoreExceptions>
</psalm>
```

orklah:
could you try running with --no-diff ?

ilyaplot:
@orklah if run with --no-diff then everything is ok, no errors.

davidbyoung:
I'm also getting this issue on my project.  `--no-diff` does seem to resolve the issue.

christianlupus:
I have the same issue where `--no-diff` seems to help. I checked with 4.30 and 5.12. Both seem to be affected.

PR #9936, opened 2023-06-20
Do not invalidate cache during suppression

ro0NL:
see #9918 for background

im not aware of psalm internals, but i tried this change first and it reduces runtime practically **100%** for us when running `psalm --set-baseline=psalm-baseline.xml` _with_ cache

I also cannot spot any side-effects still this may cause, eg. expected issues are still being written to, and removed from the baseline :flushed: 

Im curious what tests do here with this change, hence a PR.

cc @orklah i must be missing something :smile:

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -421,9 +421,7 @@ final class Psalm
 
     private static function initIsDiff(array $options): bool
     {
-        return !isset($options['no-diff'])
-            && !isset($options['set-baseline'])
-            && !isset($options['update-baseline']);
+        return !isset($options['no-diff']);
     }
 
     /**


src/Psalm/Internal/Cli/Psalm.php
```diff
@@ -421,9 +421,7 @@ private static function initShowInfo(array $options): bool
 
     private static function initIsDiff(array $options): bool
     {
-        return !isset($options['no-diff'])
-            && !isset($options['set-baseline'])
-            && !isset($options['update-baseline']);
+        return !isset($options['no-diff']);
```
ro0NL, line 424: i cant really rationalize this. i feel like it hacks the caching for whatever purpose. IMHO then `--no-diff` should be passed explicitly (ref #7681)

orklah: I fear Psalm may be unable to refresh or recreate the baseline unless it has analyzed your whole codebase. It would make sense at least. So whenever your ask it to do that, it can't just analyze files that changed or were removed and launch the whole analysis.

You may have to choose between having a squeaky clean baseline and a fast analysis

ro0NL: im not sure i follow, it seems psalm is able to recreate the baseline from cache just fine :thinking: 

i can confirm after `rm psalm-baseline.xml` regenerates the same baseline file using `vendor/bin/psalm --set-baseline=psalm-baseline.xml`

ro0NL: i can also confirm after `rm psalm-baseline.xml` and intentionally forcing an error (referenced an unknown class in src/ somewhere), the baseline is updated as expected

ro0NL: i can also confirm after fixing that error i get `Baseline for issue "UndefinedClass" has 1 extra entry. (see https://psalm.dev/316)` AND it's removed from the baseline file

orklah: There's two things that can reduce the time of Psalm's analysis:
- the cache (we store parsed files and also store method/class definitions)
- the diff option (Psalm will only check changed files instead of analyzing everything again)

Both can be controlled through --no-cache and --no-diff (though I think --no-cache implies --no-diff).

I think you may have misinterpreted Psalm using the --no-diff mode when you ask it to regenerate baseline vs Psalm using no cache at all.

Could you make a few tests with --no-diff, --no-cache and --set-baseline to see if it matches your expectations?

ro0NL:
@orklah im afraid the matrix is too big to mentally coop for me :crying_cat_face: 

(with matrix i mean: local/global install vs phar/src vs v5/master vs no-diff/no-cache)

>  I think --no-cache implies --no-diff

I think so. Perhaps this PR is about: _Do not invalidate diff during suppression_ :)

I still believe the fix _looks_ legit, ppl can opt-in to the previous behaviour by passing `--no-diff` explicitly.

I've now updated our global phar installation to master. Our main project runs 2 psalm instances/configs, for non-legacy/legacy respectively.

Command running:

```
psalm --no-progress --output-format=github --set-baseline=psalm-baseline.xml && ko=0 || ko=1; \
psalm --no-progress --output-format=github -c psalm-legacy.xml --set-baseline=psalm-legacy-baseline.xml && exit $ko
```

Some numbers (all without this PR);

global phar psalm v5 no-cache: 7-8m
global phar psalm v6 cached:  6-7m

global phar psalm master no-cache: 8-9m
global phar psalm master cached:  6-7m

At this point im assuming local/global install vs phar/src does not matter (or shouldnt ;))

For comparison, we also have 2 symfony kernels (for frontend/backend respectively) that we compile in 2 envs (prod/test respectively). All wiith APP_DEBUG=1 (so the cache is invalidated based on mtime), and it takes 2m roughly.

ro0NL:
testing on my local machine with a local phar (vendor/bin/psalm.phar)

(all this is php7.4 btw)

local phar psalm master no-cache: 6-7m
local phar psalm master cached: 5-6m

at this point ive added `continue-on-error: true` for the psalm job in GHA, so eg. unit tests will run either way.

then for any commit to develop we just take >10m jobs for granted :)

ro0NL:
> You've used 100% of included services

:rofl: 

@orklah any chance to get this merged? given --no-diff remains available to opt-in?

or do you suggest to not suppress large codebase realtime, but rather eg. daily?

orklah:
> @orklah any chance to get this merged? given --no-diff remains available to opt-in?

I went back four years and found this:
https://github.com/vimeo/psalm/blame/fe294a4dc08d72c89289fb40968421a7aed4066f/src/psalm.php#L684


So yeah, this was definitely by design, there must be something deep that disallow using diff mode while also reseting the baseline

ro0NL:
@orklah, im still wondering about `--set-baseline --no-diff` v `--set-baseline`, and also why i need to repeat baseline file in --set-baseline, but that's a different story :sweat_smile: 

eventually i removed the psalm supression step from our GHA job, allow the psalm step to continue on error, disabled findUnusedBaselineEntry, and added daily psalm suppression to our backlog 

gonna leave it at this :)

orklah:
`--set-baseline --no-diff v --set-baseline`

There's no difference between those two commands. --set-baseline implies --no-diff.

We could have kept an error when using --set-baseline without --no-diff but it's noise

ro0NL:
right, i meant `--set-baseline --no-diff` v `--set-baseline --diff`

so the TLDR here is; patching the baseline based on diff (aka fast) is not (yet) supported

ro0NL:
> global phar psalm master cached: 6-7m

related to #9884, ive cleanup all the includes, so they are included once at front controller level

adding each function file as stub is still needed (it cannot be put in our composer.json due the non-legacy part of the project)

BUT this speeds up psalm suppression, im now seeing stable `real	3m39,926s`

still slow :roll_eyes:  but OK.

the side-effect is every function is loaded from stub, which ignores at least `@deprecated` at the function level. But we dont  care, since IDE still strikes them.

PR #9937, opened 2023-06-20
Do not invalidate cache during suppression [master]

ro0NL:
ref #9936 for comparison against master


ygottschalk:
FYI: circleci test-with real-project is failing for multiple PR's. Does not seem to have smth to do with this.

ro0NL:
aiming as a bugfix in #9936 :muscle:

Issue #9941, opened 2023-06-21
Possible regression due sprintf: TooManyArguments


ro0NL:
After bumping global phar installation from v5 to master, including a cache removal, we see a 130 issues about `sprintf` in our codebase:

```
Too many arguments for the number of placeholders in sprintf (see https://psalm.dev/026)
$output->writeln(sprintf('Handling product %d => %d (%d)', $productId,  $categoryId, $position));
```

ro0NL:
Sadly no

orklah:
Can you tell us in which PHP version you use to run Psalm (the runtime version, not the analysis version)

cc @kkmuffme

ro0NL:
right, both php 7.4 still

orklah:
Note: I also reproduce on my work codebase on PHP 7.4. Seems like detection based on sprintf return is broken before PHP 8

Issue #9888, opened 2023-06-09
Incorrect return types for stubbed core functions depending on PHP version


kkmuffme:
https://psalm.dev/r/7106f13383

in PHP 8 `vsprintf` will never return false, but in PHP 7 it will.
As discussed in https://github.com/vimeo/psalm/pull/9877#issuecomment-1583319344 the stubs will never fall back to the callmap - this makes declaring different return/param types between PHP versions impossible in psalm.

I suggest to change it to fall back from the stubs to the callmap if the param/return type is not declared in the stubs (but is in the callmap)

Also somehow related to https://github.com/vimeo/psalm/issues/9843 since atm no consistency checking happens in-between the different versions of the callmap and the stubs either

ygottschalk:
I thought that we have the files `stubs/Php8X.phpstub` for that...

orklah:
> I thought that we have the files stubs/Php8X.phpstub for that...

Yeah, that is a first option. We also have this:
https://github.com/vimeo/psalm/blob/bd0ba6ee957a1789cf5c5fe647893bfe3bda2425/stubs/CoreGenericFunctions.phpstub#L66

For anything that those two tools can't answer, TypeProviders are the go to.

Issue #9873, opened 2023-06-05
SprintfReturnTypeProvider not working well with array unpacking


fancyweb:
Hello, SprintfReturnTypeProvider crashes when sprintf is used with array unpacking because it considers there's one arg only, eg:
```php
$array = ['foo', 'bar', 'ccc'];
sprintf('%s %s %s', ...$array);
```

fancyweb:
Some real code that show the bug: https://github.com/symfony/symfony/pull/50565/files

fancyweb:
https://psalm.dev/r/260236cf5b

mrVrAlex:
@orklah @kkmuffme PR not full resolved this issue, look this case: https://psalm.dev/r/308b6b4ae5
Should I open new bug issue? Or just re-open it?

PR #9943, opened 2023-06-21
fix PHP 7 sprintf too many arguments false positive

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/9941

PHP 7 was an afterthough, so it was wrong originally and I couldn't test this here either, since I don't have PHP 7 available on my machine.

---

Fix https://github.com/vimeo/psalm/issues/9873
Fix https://github.com/vimeo/psalm/issues/9888


ro0NL:
uhm, it seems it found 205 more false positives with this change :sweat_smile:

orklah:
Thanks! I'll wait for feedback to prevent shipping with an obvious issue and I'll make a release

ro0NL:
@orklah @kkmuffme 

```diff
-    <TooManyArguments>
-      <code>sprintf</code>
-    </TooManyArguments>
```

:tada:

PR #9953, opened 2023-06-23
`non-empty-list` assertion for iterable

### LINKED ISSUE #9917: AssertionReconciliation: `non-empty-list` results in `list`
When I pass an iterable to an assertion which verifies that I do have a `non-empty-list`, psalm does not adapt `non-empty-list` and instead reconciles to `list`.

https://psalm.dev/r/0441efb80e

boesing:
This fixes #9917


boesing:
Hm, unsure if PSL circle CI check and the process timeout are related to this PR. 
Can some1 probably restart the failed checks?

orklah:
Tests are ok, I need to fix the real project one and the other one happens from time to time

Thanks!

PR #9955, opened 2023-06-24
Invalidate cached methods when referenced class property types change

tscni:
When the property type changes in a class referenced by a method, that method is currently not invalidated.
Example:
1. File `A.php` with method `A::foo()` has an argument of type `B`, which is a class defined in file `B.php`
   It uses `B::$value` incorrectly, raising an issue.
2. Psalm runs and caches
3. The type of `B::$value` changes to fix the issue
4. Psalm runs and still reports the same issue


orklah:
Thanks!

Issue #6664, opened 2021-10-14
Cache doesn't work


kkmuffme:
psalm takes exactly the same amount of time whether running it normally or with `--no-cache` 
While the first run should take the same amount of time for both, any repeated runs (no changes were done to the file) should be much faster with cache enabled.

The cache is generated correctly (I see how it generates it in .cache/psalm).

I run psalm for a single file only. Same issue though, when I run it for multiple files at once.

Psalm 4.10.0

kkmuffme:
No, this is not a bug with psalm emitted issues.

orklah:
can you post the output of --debug-by-line from the first and second execution when analyzing a single file?

kkmuffme:
No.

Just ran it again and confirmed that the composer.lock is identical in both cases.

orklah:
Can you try to debug on your end what makes it go into this message?

Do you have a standard install of Psalm? Do you use phar, docker or something? Do you have custom bin directory maybe?

If you don't find anything, could you try to replicate on a small repo?

Issue #4941, opened 2021-01-07
Cache doesn't work when the composer.lock is in a sub directory.


VincentLanglet:
When I run
```
symfony/vendor/bin/psalm --error-level=8
```
I get the output: 
```
Checks took 138.87 seconds and used 1,431.678MB of memory
```

If I run again the command, the output is
```
Checks took 140.04 seconds and used 1,431.678MB of memory
```

I tried `--no-cache`, I got the output
```
Checks took 89.93 seconds and used 758.369MB of memory
```

I also tried, `--no-diff`, I got the output
```
Checks took 121.00 seconds and used 1,431.675MB of memory
``` 

This is my psalm config
```
<psalm
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="https://getpsalm.org/schema/config"
    xsi:schemaLocation="https://getpsalm.org/schema/config symfony/vendor/vimeo/psalm/config.xsd"
    allowPhpStormGenerics="true"
    autoloader="symfony/bin/.phpunit/phpunit/vendor/autoload.php"
    cacheDirectory="cache/psalm"
    errorLevel="8"
    findUnusedPsalmSuppress="true"
    reportInfo="false"
>
    <projectFiles>
        <directory name="symfony/src"/>
        <directory name="symfony/tests"/>
        <ignoreFiles>
            <file name="symfony/src/Repository/.phpstorm.meta.php"/>
            <file name="symfony/tests/.phpstorm.meta.php"/>
            <directory name="symfony/src/Migrations"/>
            <directory name="symfony/vendor"/>
        </ignoreFiles>
    </projectFiles>
    <plugins>
        <pluginClass class="Psalm\PhpUnitPlugin\Plugin"/>
        <pluginClass class="Psalm\SymfonyPsalmPlugin\Plugin">
            <containerXml>symfony/var/cache/dev/srcAssuranceVie_KernelDevDebugContainer.xml</containerXml>
        </pluginClass>
    </plugins>
</psalm>
```

VincentLanglet:
I tried another cacheDirectory, and I takes 90s now. Maybe it was because the cache directory was previously in the shared folders of my vagrant.

But I don't understand why running psalm with or without the cache takes the same time. What's the interest of the cache then ?

orklah:
This is generally not the case. Maybe there's something on your codebase that makes psalm write too much in cache to a point where reading the cache is detrimental?

[...]

PR #9970, opened 2023-06-27
Missing composer lock should not invalidate cache created with missing composer lock

kkmuffme:
When the composer.lock file is missing, psalm writes the cache data, but never uses them.

Misc improvements for consistency of cache and performance when not using cache

Fix https://github.com/vimeo/psalm/issues/6664
Fix https://github.com/vimeo/psalm/issues/4941 since cache fails silently/best effort/fallback as of https://github.com/vimeo/psalm/pull/9932


orklah:
Thanks!

PR #9975, opened 2023-06-29
Fix sprintf single arg splat array

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/9873#issuecomment-1612901224


orklah:
Thanks!

Issue #5994, opened 2021-06-25
Invalid return type for template where param and return are the same array


M1ke:
https://psalm.dev/r/c75c90c4ec

The issue here is that we're wanting the docblock to declare that the function must return the input param. Psalm instead requires us to permit the function to return either the input value, or an alternative array that shares a similar type description.

This means that the common goal of "a function that takes an array, modifies it and returns it" can't be represented without also permitting the function to take an array, discard it and return something that looks like it.

M1ke:
If we unset the first key instead then Psalm has no issues with modifying it

https://psalm.dev/r/c75c90c4ec

However if we modify an internal item (even if we explicitly check it exists first) the same issue is thrown.

https://psalm.dev/r/c75c90c4ec

PR #9981, opened 2023-06-30
fix mixed replace return types for arrays

kkmuffme:
* remove useless replace return type provider code that returned incorrect type for array replacements and was worse duplicate of existing stubs
* require preg patterns to be non-empty-strings as otherwise it will throw a PHP notice
* improve return type of array replacements to be more correct (not fully correct due to https://github.com/vimeo/psalm/issues/5994#issuecomment-1614601590)

### DIFF
--- a/src/Psalm/Config/Creator.php
+++ b/src/Psalm/Config/Creator.php
@@ -23,6 +23,7 @@ use function glob;
 use function implode;
 use function is_array;
 use function is_dir;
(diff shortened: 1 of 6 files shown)


orklah:
You have a few coercion errors left

orklah:
oh nvm, github showed me an old commit

orklah:
Thanks!

PR #9982, opened 2023-06-30
Infer output format from GITHUB_ACTIONS env

ro0NL:
as per https://docs.github.com/en/actions/learn-github-actions/variables

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -416,6 +416,10 @@ final class Psalm
             return Report::TYPE_PHP_STORM;
         }
 
+        if ('true' === getenv('GITHUB_ACTIONS')) {
+            return Report::TYPE_GITHUB_ACTIONS;
+        }
+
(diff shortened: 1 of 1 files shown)


ro0NL:
we can also infer `--debug` from `GITHUB_ACTIONS=true && RUNNER_DEBUG=true`, lemme know :+1:

orklah:
I'm fine with the change but it breaks tests as we rely on output. You'll need to force the output to fix that I guess

ro0NL:
@orklah tests seems OK now

last commit infers --debug in GHA, you can try it here by re-running jobs with enabled debug logging

IMHO it's nice UX

orklah:
You have a few CS errors, otherwise seems good!

ro0NL:
@orklah 
image
how do i make it stop? i dont have a circle ci acccount :S

orklah:
Thanks!

ro0NL:
@orklah did you revert it? :joy_cat: 
image

PR #9988, opened 2023-07-03
Fixed exhuming enums as enums, not as classes

### LINKED ISSUE #9968: Type confusion regression from #9373
Introduced in ad5bf62094cf23307c34ebfdf3758addbe3d837d

Before: no error

After:

```
ERROR: MixedArgument - src/Builder/FilterBuilder.php:30:36 - Argument 1 of preg_replace cannot be mixed, expecting array<array-key, string>|string (see https://psalm.dev/030)
        $harmonized = preg_replace(self::ANY_NEWLINE, Newline::UNIX->value, $text);
```

It’s pretty hard to build a minimal reproduction case because it seems to be only happening if autoloading is involved. So I stripped down the library to a minimal reproduction case and this is the branch: https://github.com/lstrojny/uffff/tree/dev/psalm-bug-repro

It only consists of two files:

src/Builder/FilterBuilder.php
```php
<?php

namespace Uffff\Builder;

class FilterBuilder
{
    /**
     * @return callable(string):string
     */
    public function build(): callable
    {
        return static fn(string $v) => $v;
    }
}

enum Newline: string
{
    case UNIX = "\n";
    case WINDOWS = "\r\n";
    case MAC = "\r";
}

class HarmonizeNewlines
{
    // Order matters, Windows must come first
    private const ANY_NEWLINE = '/(?:' . Newline::WINDOWS->value . '|' . Newline::MAC->value . '|' . Newline::UNIX->value . ')/';

    public function __invoke(string $text): string
    {
        $harmonized = preg_replace(self::ANY_NEWLINE, Newline::UNIX->value, $text);

        assert(is_string($harmonized));

        return $harmonized;
    }
}
```

src/functions.php:

```php
<?php

declare(strict_types=1);

namespace Foo;

use Uffff\Builder\FilterBuilder;

function unicode(string $text): string
{
    $filter = (new FilterBuilder())->build();

    return $filter($text);
}
```

The two files are covered by composer autoloading like this:

```json
    "autoload": {
        "psr-4": {
            "Uffff\\": "src/"
        },
        "files": [
            "src/functions.php"
        ]
    },
```

The issue does not happen if:
 * The namespace changes
 * The FilterBuilder is removed and instead we try to use HarmonizeNewlines directly in functions.php

ygottschalk:
fixes #9968


ygottschalk:
Failing shepherd unrelated

orklah:
Thanks!

Issue #4024, opened 2020-08-18
How to tell psalm that constants are configurable?


AgentOak:
I'm using a php file with constants as configuration, like this (`config.php`):
```php
<?php
/**
 * Enables debug mode, bla bla bla
 *
 * @var bool
 */
const DEBUG = false;
```

This file is required() from a bootstrap file before classes that make use of these constants are autoloaded.

In my `psalm.xml` I list `<file name="config.php">` in `<projectFiles>`, but then psalm complains about UndefinedConstant violations in other files:

```
ERROR: UndefinedConstant - src/Templating/Template.php:29:29 - Const DEBUG is not defined (see https://psalm.dev/020)
        $options["debug"] = DEBUG;
```

I found #2644 where `<stubs>` are suggested. Adding `config.php` as a stub fixes these warnings, but now psalm also uses the predefined values of these constants for analysis and as a result finds violations like this:

```
ERROR: TypeDoesNotContainType - src/Templating/Template.php:30:29 - if (false) is impossible (see https://psalm.dev/056)
        $options["cache"] = DEBUG ? false : self::CACHE_DIR;
```

How to handle this case? I cannot define the constants without values because that is a syntax error. Thanks.

ygottschalk:
As a workaround: Try including it as an autoloader
```xml
<psalm [...] autoloader="config.php">
    <projectFiles>
        [...]
    </projectFiles>
</psalm>
```
(maybe you'll additionally need to include the file `<file name="config.php">` in `<projectFiles>`)
Hope this helps

philstutz:
We use

```php
define('DEBUG', /** @var bool $x */ $x = false);
```

in a stub - works.

https://psalm.dev/r/77d4facff4

xPaw:
Do I understand correctly that this stub workaround does not work for class constants since `define('Thing::DEBUG', ...)` wouldn't work?

PR #9992, opened 2023-07-03
make stubbed constant types configurable with comment

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/4024

With a simplified, more straightforward syntax compared to the example in the issue (which never really worked apparently and only set the type to mixed):

```
define('DEBUG', /** @var bool */ false);
```


orklah:
Thanks!

PR #9999, opened 2023-07-05
offsetExists/Get/Set/Unset types were too generic

kkmuffme:
since only array key types + null lead to expected results with "offsetSet", while they won't work with all other methods anyway.

Possibly the types can even be made more strict for some cases (e.g. only int or only string or only int|null), however I didn't have the time to test all of these, therefore I applied the narrowest possible, but broadest necessary types.


orklah:
Thanks!

PR #10010, opened 2023-07-09
Issue with template inheritance and overriding method signatures

boesing:
I have no clue how to describe the issue, maybe I am using something wrong or I don't know.
I dived through the code and found out that this code would work if the following check would return true:

__\Psalm\Internal\Codebase\Methods::getMethodReturnType__
```php
$new_contained_by_old = UnionTypeComparator::isContainedBy(
    $source_analyzer->getCodebase(),
    $overridden_storage_return_type,
    $candidate_type,
);
```

The union comparator returns `false` due to the fact that the `TTemplateParam` `TRequestedInstance` provides `mixed` but should return a super of `InstanceType` due to `TRequestedInstance extends InstanceType` and `InstanceType` is actually known as `PluginInterface`.

So I guess I do have to find out why the `TRequestedInstance` is not properly inferred as `PluginInterface` as that would probably lead to the expected result.

Fun fact, the code works once removing the overriden method `AbstractSingleInstancePluginManager#get` as it narrows the return value from `mixed` to `object` which then seems to mess up the whole type detection...

Happy if some1 has some little guidance here and can point me in the correct direction. As of now, I'd say the problem is that `TRequestedInstance` is not providing `PluginInterface` as it should, but no clue where this should happen.

### DIFF
--- a/tests/Template/ClassTemplateTest.php
+++ b/tests/Template/ClassTemplateTest.php
@@ -4207,6 +4207,68 @@ class ClassTemplateTest extends TestCase
                 'ignored_issues' => [],
                 'php_version' => '8.0',
             ],
+            'inheritedConditionalsWithMatchingTypes' => [
+                'code' => '<?php
+                    /** @template InstanceType */
+                    interface PluginManagerInterface
+                    {
+                        /**
+                         * @template TRequestedInstance extends InstanceType
+                         * @param class-string<TRequestedInstance>|string $id
+                         * @return ($id is class-string ? TRequestedInstance : InstanceType)
+                         * @throws InvalidArgumentException
+                         */
+                        public function get(string $id): mixed;
+                    }
+
+                    interface PluginInterface
+                    {}
+
+                    class ConcretePlugin implements PluginInterface
+                    {}
+
+                    /**
+                     * @template InstanceType
+                     * @template-implements PluginManagerInterface<InstanceType>
+                     */
+                    abstract class AbstractPluginManager implements PluginManagerInterface
+                    {
+                        public function get(string $id): mixed
+                        {
+                            throw new InvalidArgumentException();
+                        }
+                    }
+
+                    /**
+                     * @template InstanceType of object
+                     * @template-extends AbstractPluginManager<InstanceType>
+                     */
+                    abstract class AbstractSingleInstancePluginManager extends AbstractPluginManager
+                    {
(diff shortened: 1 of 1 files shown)


boesing:
@orklah @weirdan any ideas? :-/

orklah:
I'll try to take a look in the next few days :)

tests/Template/ClassTemplateTest.php
```diff
@@ -4207,6 +4207,68 @@ public function aggregate(...$values): null|int|float
                 'ignored_issues' => [],
                 'php_version' => '8.0',
             ],
+            'inheritedConditionalsWithMatchingTypes' => [
+                'code' => '<?php
+                    /** @template InstanceType */
+                    interface PluginManagerInterface
+                    {
+                        /**
+                         * @template TRequestedInstance extends InstanceType
```
weirdan, line 4216: I don't think we support `extends` here

boesing: it works flawless if I do not overwrite the method in the abstract class. the issue is adding the object return type which makes it less specific.

orklah: Well, I confirm we don't do anything with the extends keyword so it's just comments as far as Psalm is concerned

boesing: but it actually seems to work until I add the overridden method with the object return type 🤔

orklah: can you show an example where psalm behaves differently based on the extends?

boesing: I guess I cant. I just know that the way psalm behaves differs once I extend the `get` method and narrow the type from mixed to object

orklah: could you post two snippets so we see the difference exactly? I'm still a bit confused

boesing: Okay, I've digged a bit deeper and found out that there is a general issue with `@inheritdoc` as due to this, templates are completely dropped and therefore the whole type detection seems to be problematic.

https://psalm.dev/r/538a0d30d4

I am totally lost here, but thats maybe due to the fact that I spent too much time on other psalm related stuff in the past days.
Would really love to get some help on this but will also try to solve this on my own in the next weeks.

PR #10012, opened 2023-07-10
Always report UnusedBaseline issues

### LINKED ISSUE #10001: UnusedBaselineEntry error not showing on error free codebase
it seems that when having the `findUnusedBaselineEntry` config option set to true, the `UnusedBaselineEntry` error only shows when other issues are found and disappears when there are none.

[here](https://github.com/wickedOne/psalm-test) is a minimal standalone project to illustrate the issue (instructions for reproduction over [here](https://github.com/wickedOne/psalm-test/blob/a6bfc79414af2c217c28825f16a6ed868e4a00e1/src/Foo.php#L9-L12)).

ygottschalk:
fixes #10001
discussion in #10001

### DIFF
--- a/src/Psalm/IssueBuffer.php
+++ b/src/Psalm/IssueBuffer.php
@@ -570,9 +570,8 @@ final class IssueBuffer
             foreach (self::$issues_data as $file_path => $file_issues) {
                 usort(
                     $file_issues,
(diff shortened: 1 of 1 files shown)


orklah:
Thanks!

Sorry for the delay!

PR #10014, opened 2023-07-11
Fix inference high order function when closure param is omitted

klimick:
$result1 and $result2 must have the same type: https://psalm.dev/r/efac7ef1f4

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
@@ -268,7 +268,12 @@ class ArgumentsAnalyzer
 
             $inferred_arg_type = $statements_analyzer->node_data->getType($arg->value);
 
-            if (null !== $inferred_arg_type && null !== $template_result && null !== $param && null !== $param->type) {
+            if (null !== $inferred_arg_type
+                && null !== $template_result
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php
```diff
@@ -268,7 +268,12 @@ public static function analyze(
 
             $inferred_arg_type = $statements_analyzer->node_data->getType($arg->value);
 
-            if (null !== $inferred_arg_type && null !== $template_result && null !== $param && null !== $param->type) {
+            if (null !== $inferred_arg_type
+                && null !== $template_result
+                && null !== $param
+                && null !== $param->type
+                && !$arg->unpack
```
klimick, line 275: Was two failling tests here:
- ArrayFunctionCallTest::testValidCode(splatArrayIntersect)
- FunctionTemplateTest::testValidCode(splatTemplateParam)

And I've added `&& !$arg->unpack`.
I don't know how to fill template result when arg is being unpacking.
This branch needs for this kind inference: https://psalm.dev/r/086e0799b6
I think we can leave it as is.

klimick:
Falling of test-with-real-projects unrelated. I see same fails in the neighboring PR.

orklah:
Thanks!

PR #10024, opened 2023-07-19
Switch to amp v3

danog:


weirdan:
Is this something for Psalm 6?

danog:
Yep, it seems like it, since we're raising the minimum version

danog:
@orklah Do you think you could make another branch for v5 just like we did for v4?
I don't have any clear ETA for v6, but at least this way people that have dependency conflicts due to amp will be able to require the dev-master branch, instead (and I could also send in some PRs to remove stuff that was planned for deprecation in v6).

danog:
Addendum: I don't have any clear date for v6 in mind, but I would prefer we avoid prolonging the release like we did last time, it's no big deal to make majors once or twice a year IMO :)

danog:
Should be good to merge!

Issue #10030, opened 2023-07-21
Missing TaintedSql when using PDOStatement::bindParam or PDOStatement::bindValue


cgocast:
The following code sample does not raises a TaintedSql:
```
<?php

function deleteBindedUserId(PDO $pdo): void {
    $userId = $_POST['userid'];
    $stmt = $pdo->prepare("delete from users where user_id = :userid");
    $stmt->bindParam(':user_id', $userId);
    $stmt->execute();
}
```

whereas a very similar piece of code correctly raises a TaintedSql:
```
<?php

function deleteConcatenedUserId(PDO $pdo): void {
    $userId = $_POST['userid'];
    $stmt = $pdo->prepare("delete from users where user_id = " . $userId);
    $stmt->execute();
}
```

I'll write a PR fixing the issue.

PR #10032, opened 2023-07-21
#10030 Fix missing TaintedSql

cgocast:
We add PDOStatement::bindValue() and PDOStatement::bindParam() as sources for TaintedSql


orklah:
Thanks!

PR #10033, opened 2023-07-21
Automatic LSP container path mapping

weirdan:
This is an alternative for vimeo/psalm#8026. 
Unlike that PR, this one doesn't require any configuration (but still allows some).

Mapping is opt-in, to enable it one needs to specify `--map-folder` CLI switch to `psalm-language-server` binary. It can also be manually configured by providing an argument to `--map-folder` switch, e.g. `docker-compose exec -T php vendor/bin/psalm-language-server --map-folder=/var/www:$PWD` (variable expansion is handled by the shell).


weirdan:
@tm1000 would you mind taking a look at this? I think I addressed all concerns that were raised in the discussion we had.

orklah (approved):

tm1000 (approved):

PR #10034, opened 2023-07-23
Fix incorrect positions set in code action edits

weirdan:
Neovim requires positions to supply column as well as row.


orklah:
Thanks!

PR #10039, opened 2023-07-24
Fix combination of non-falsy-string and falsy literal

danog:


orklah:
This feels wrong. You don't have a check for non-empty-string anymore on the condition and your check for non-falsy-string checks for non-emptyness

I'd have expected a lot of tests to break :(

danog:
TNonFalsyString extends TNonEmptyString, so the two checks were actually conflicting with each other.

With this, I check either for non-falsyness or for non-emptyness, based on the exact concrete type.

orklah:
Yeah but TNonFalsyString is the most specific and is handled first so it shouldn't be an issue

Do you have a snippet of code showing the issue?

danog:
It is handled first but it doesn't matter because what follows is a check for TNonEmptyString && $other->value !== '', and since TNonFalsableString is also TNonEmptyString and '0' !== '', the new testcase I added would fail

orklah:
Thanks!

PR #10043, opened 2023-07-24
Fix scanning of intersection types

danog:
This codepath is invoked during scanning of native intersection types, and it breaks the classlike storage, leading to UndefinedClass issues during analysis.


danog:
This is just a hotfix, I really don't like the logic in https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Codebase/ClassLikes.php#L404 et al, but fixing that would require a bit more time...

orklah:
Thanks!

Issue #10035, opened 2023-07-23
[lsp] Report progress using `$/progress` notifications


weirdan:
Currently, we're using ad-hoc solution to report progress to vscode extension, using telemetry events. This requires specific client-side support. There's now a better option: reporting the progress using `$/progress` notifications, that are allowed during the initialize phase if client sends `workDoneToken` in `initialize` request. Using it would allow reporting the progress to any compliant client, not just our vscode ext.

Spec links:
* [initialize request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize)
* [workdone notifications](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workDoneProgress)

Open questions:
1. How do we prevent duplicate reports when older vscode ext connects to newer psalm server? Client sniffing?
2. Neovim doesn't seem to be sending `workDoneToken` in initialize requests (this can probably be overridden somewhere).

tm1000:
@weirdan so I believe trying to get `$` prefixes to work with https://github.com/felixfbecker/php-advanced-json-rpc is part of the issue. The way json-rpc works is it maps the RPC requests to functions and a function can't start with a $ however I haven't looked in depth at it to see

> How do we prevent duplicate reports when older vscode ext connects to newer psalm server? Client sniffing?

So the VSCode client is sent to the server but I believe that isn't in the LSP spec it's just something they do.

If I (we, whomever) can figure out a way to get the `$` to work it would be nice to also get `$/cancelRequest` to work which is useful since we run everything through a defer throttler now in amp

weirdan:
>  it maps the RPC requests to methods 

Doesn't it only happen for incoming requests though? `$/progress` is something we send, not receive.

weirdan:
As a quick test, I added (1) to `initialize` handler and got (2):

![image](https://github.com/vimeo/psalm/assets/57403/07135492-1ebc-4046-8dbb-f0ee00076e43)

weirdan:
> it maps the RPC requests to methods

The way mapping works it splits the request name using a separator (we use `/`), and then uses the last part as a method to call and all preceding parts as nested property access. So `$/cancelRequest` becomes a call to `$server->{"$"}->cancelRequest()`.

Unlike methods, property names can be arbitrary, so `$` as a property name is possible. The following worked for me (in `LanguageServer::__construct()`):

```php
        $this->{"$"} = new class($this) {
            private object $parent;
[...]

PR #10046, opened 2023-07-24
Map special RPC paths

weirdan:
Now `$/cancelRequest` will be resolved to `$server->cancelRequest()`
and `$/textDocument/whatever` to `$server->textDocument->whatever()`

/cc: @tm1000

Refs: https://github.com/vimeo/psalm/issues/10035#issuecomment-1648526809

### DIFF
--- a/src/Psalm/Internal/LanguageServer/LanguageServer.php
+++ b/src/Psalm/Internal/LanguageServer/LanguageServer.php
@@ -997,4 +997,19 @@ class LanguageServer extends Dispatcher
 
         return $filepath;
     }
+
+    // the methods below forward special paths
+    // like `$/cancelRequest` to `$this->cancelRequest()`
+    // and `$/a/b/c` to `$this->a->b->c()`
+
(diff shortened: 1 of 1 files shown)


tm1000 (approved):
Perfect. Thanks @weirdan

tm1000:
@weirdan Since you're playing in the LSP. I have one request for you if you have time

I've always wanted to know how to get psalm to reindex the project without having to restart the LSP. Right now I just restart the LSP. 

But I assume there must be another way to do it

This is useful in cases where composer has updated and now psalms cache is out of sync

weirdan:
> But I assume there must be another way to do it

No good way to do that I'm afraid. We run autoloaders (see below) and it probably won't work second time in the same process.

https://github.com/vimeo/psalm/blob/4a0a12d9e23e28ccda9b05a8461893f82c5ccd04/src/Psalm/Internal/Cli/LanguageServer.php#L259-L266

PR #10049, opened 2023-07-25
Fix #9997 dynamic properties on SimpleXmlElement

### LINKED ISSUE #9997: Dynamic properties of SimpleXMLElement should be inferred as being SimpleXMLElement
https://psalm.dev/r/100575e0f4

This reports mixed errors saying it cannot infer the type of expressions. But dynamic properties of SimpleXMLElement are always SimpleXMLElement themselves.

ygottschalk:
Refs #9997
Types dynamic property access on `SimpleXmlElement` to `SimpleXmlElement|null`.

BC impact:
`SimpleXmlElement` is not considered an universal object crate anymore. Therefore things like `$xmlElement->myExtraStringData = 'fooBar';` will not work anymore when this get merged.

I do not know why this code was in place and I am not sure if my changes are 'legit': https://github.com/vimeo/psalm/pull/10049/files#diff-a3219413916e4af22971b07d3ef169950d3b17b7cf58f48366c408b1d49ade5bR326
I looked through Atomic.php and some other file and it seemed like null checks (in most places) are done by checking `$atomic instanceof TNull` and not considering `TNamedObject('null')` as `null`. Thats why I removed the code...

Extra info copied from commit message
- make SimpleXmlElement and SimpleXmlIterator not a universal object crate
- added typed magic __get method to SimpleXmlElement
- adjusted tests to reflect new behaviour of SimpleXmlElement
- added tests
- 'fixed' code translating `null` to `TNamedObject('null')` instead of `TNull`


src/Psalm/Type/Atomic.php
```diff
@@ -323,11 +323,7 @@ private static function createInner(
                 return $analysis_php_version_id !== null ? new TNamedObject($value) : new TScalar();
 
             case 'null':
-                if ($analysis_php_version_id === null || $analysis_php_version_id >= 8_00_00) {
-                    return new TNull();
-                }
-
-                return new TNamedObject($value);
+                return new TNull();
```
orklah, line 326: This change will prevent this to work in PHP <7: https://3v4l.org/nbaYr

orklah:
aside from the change in Atomic, this seems good :)

src/Psalm/Type/Atomic.php
```diff
@@ -323,7 +323,7 @@ private static function createInner(
                 return $analysis_php_version_id !== null ? new TNamedObject($value) : new TScalar();
 
             case 'null':
-                if ($analysis_php_version_id === null || $analysis_php_version_id >= 8_00_00) {
+                if ($analysis_php_version_id === null || $analysis_php_version_id >= 7_00_00) {
```
ygottschalk, line 326: Then lets use this code in PHP<7

ygottschalk:
@orklah ready to merge. psl fail seems unrelated

orklah:
Thanks!

Issue #9974, opened 2023-06-28
Add type detection for PDOStatement::fetchAll(PDO::FETCH_CLASS, SomeClass::class)


thbley:
Example fetchAll: https://psalm.dev/r/a9b0c85b47

Add type detection for PDOStatement::fetchAll(PDO::FETCH_CLASS, SomeClass::class): SomeClass[]|false;

Currently PDOStatement::fetchAll() is currently defined as:
```
'PDOStatement::fetchAll' => ['array', 'mode='=>'int', '...args='=>'mixed'],
```
which infers to array<array-key, mixed> instead of array<array-key, SomeClass>

orklah:
Thanks for the feedback!

Turns out we already do something similar for `fetch`: https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php#L34

It should be pretty easy to adapt this to work with fetchAll too

robchett:
Looks like this issue is resolved? There was quite a lot of discussion on the merge request, but it feels beyond the scope of the tags :)

thbley:
Yes, is resolved.

PR #10055, opened 2023-07-26
Add type detection for PDOStatement::fetchAll(PDO::FETCH_CLASS, SomeClass::class)

thbley:
related to #9974

- added return type detection for PDOStatement::fetchAll, extended return type detection for PDOStatement::fetch
- extended return type detection for PDOStatement::fetchAll/fetch with FETCH_COLUMN and FETCH_KEY_PAIR
- setFetchMode() is currently not memorized
- ATTR_DEFAULT_FETCH_MODE is currently not handled

### DIFF
--- a/src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php
+++ b/src/Psalm/Internal/Provider/ReturnTypeProvider/PdoStatementReturnTypeProvider.php
@@ -116,6 +116,13 @@ class PdoStatementReturnTypeProvider implements MethodReturnTypeProviderInterfac
             case 6: // PDO::FETCH_BOUND - bool
                 return Type::getBool();
 
+            case 7: // PDO::FETCH_COLUMN - scalar|null|false
+                return new Union([
+                    new TScalar(),
+                    new TNull(),
+                    new TFalse(),
+                ]);
+
             case 8: // PDO::FETCH_CLASS - object|false
                 return new Union([
                     $fetch_class_name ? new TNamedObject($fetch_class_name) : new TObject(),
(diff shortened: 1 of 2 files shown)


weirdan:
> I'm not sure if $context->references_in_scope['fetch_mode'] = is the right way to memorize data inside a scope

It's definitely not. The docblock says it's used to track what variables reference what.

https://github.com/vimeo/psalm/blob/13d53ecbf10c31492e17bd87fda662e43ed4a920/src/Psalm/Context.php#L55-L58

Also, consider what would happen if you have 2 statements in the scope with different fetch modes.

orklah:
Thanks!

weirdan:
I wonder why don't we drop this return type provider entirely and implement it as a conditional return type for corresponding methods in stubs instead. It'll automatically work with unions as well: https://psalm.dev/r/5bcdb161c6

thbley:
> I wonder why don't we drop this return type provider entirely and implement it as a conditional return type for corresponding methods in stubs instead. It'll automatically work with unions as well: https://psalm.dev/r/5bcdb161c6

For now it would be easier, but I guess it's more difficult to include context from `$db->setAttribute(PDO::ATTR_DEFAULT_FETCH_MODE, PDO::FETCH_ASSOC);` and `$stmt->setFetchMode(PDO::FETCH_CLASS, A::class);` later.

weirdan:
To me, it looks like PDOStatement should be generic over the fetch mode: https://psalm.dev/r/55177412f8

thbley:
e.g. phpstan-dba uses generics to inspect table column properties, e.g. 

```
        $query = 'SELECT email, adaid FROM ada';
        $stmt = $pdo->query($query);
        assertType('PDOStatement<array{email: string, 0: string, adaid: int<-32768, 32767>, 1: int<-32768, 32767>}>', $stmt);

        $stmt->setFetchMode(PDO::FETCH_NUM);
        assertType('PDOStatement<array{string, int<-32768, 32767>}>', $stmt);

        $result = $stmt->fetch(PDO::FETCH_NUM);
        assertType('array{string, int<-32768, 32767>}|false', $result);
```

So either PDOStatement<PDO::FETCH_*> or PDOStatement<return-type-structure> would be an option.

weirdan:
That's fascinating; I wasn't aware of that project. It puts us in a tough spot, though, as we won't be able to add incompatible generic signature to PDOStatement now.

PR #10067, opened 2023-07-30
Remove word possibly from error

kamil-tekiela:
This sentence is very difficult to read. Alternatively, it could be:


orklah:
Yeah, I remember being confused by this one too. I would also like if we could add the actual type that was passed

PR #10068, opened 2023-07-31
Check if argument passed to isset() is not a variable

### LINKED ISSUE #9949: Difference between `isset($neverNull)` and `isset(neverNull())`
When calling `isset()` on something with a non-null type, the resulting type can only be true. Psalm infers this correctly if passed a variable, but not if it's an expression with a clear type:

https://psalm.dev/r/98dcaa0a87

Nitamet:
Added the condition to IssetAnalyzer to emit the issue when anything other than a variable is passed to isset()
Fixes #9949

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/IssetAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/IssetAnalyzer.php
@@ -3,9 +3,12 @@
 namespace Psalm\Internal\Analyzer\Statements\Expression;
 
 use PhpParser;
+use Psalm\CodeLocation;
 use Psalm\Context;
 use Psalm\Internal\Analyzer\Statements\ExpressionAnalyzer;
 use Psalm\Internal\Analyzer\StatementsAnalyzer;
+use Psalm\Issue\InvalidArgument;
+use Psalm\IssueBuffer;
 use Psalm\Type;
 
 /**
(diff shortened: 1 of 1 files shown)


weirdan:
@Nitamet thanks for your effort. Please target `5.x` (current default) branch.

Apparently, this needs some more work, so I've marked this PR as a draft. Mark it as `ready for review` when you think you're done with it.

Nitamet:
And I thought that running composer tests was enough  💀
Yeah, it seems that we should at least take checking array elements into account and maybe there are other cases.
I'll take a closer look at the Expr types and try to fix these problems.

weirdan:
You need to rebase your changes on top of `5.x` and force-push. Just changing the target branch added a bunch of master-only changes we're not going to push to `5.x` (they are for Psalm 6).

Nitamet:
I've added other valid statements and changed the branch, tests seem to be ok now, the failed one is not related to this PR.

There's a little issue related to null coalescing operator though
php.net states that foo ?? bar is just a shorthand for isset(foo) ? foo : bar, but in fact there are differences:
```php
class Foo {
    public const BAR = 'bar';
}

// This works
echo Foo::BAR ?? "doesn't exist";
// And this crashes
echo isset(Foo::BAR) ? Foo::BAR : "doesn't exist";

$var = 0;
// This works
($a =& $var) ?? "hello";
// And this crashes
isset((a =& $var)) ? (a =& $var) : "hello";
```

So if we don't add `ClassConstFetch` and `AssignRef` as valid arguments, we will get errors on valid null coalescing operators like in the example above.
And vice versa, if we add them, we won't get errors for invalid arguments `ClassConstFetch` and `AssignRef` in isset().
I'm not sure if there's an easy way to distinguish between `??` and `isset` and if that's in the scope of this PR, so for now I've just added them both as valid to avoid getting false errors.

weirdan:
> so for now I've just added them both as valid to avoid getting false errors.

Sounds good to me.

weirdan:
Thanks!

PR #10073, opened 2023-08-01
Fixed `DOMDocument::load*` signatures

weirdan:
Prior to 8.0 those methods could be called statically, returning a
`DOMDocument` instance. In 8.0 and above they no longer can be.


weirdan:
This should fix unit tets failures.

Issue #10021, opened 2023-07-15
Sprintf with only one argument without formatting directives leads to false TooFewArguments


tscni:
https://psalm.dev/r/6c9dd0f8b3

Presumably caused by
https://github.com/vimeo/psalm/blob/564c6de4ed3a4050f3e50b98b741ff1275d79a9d/src/Psalm/Internal/Provider/ReturnTypeProvider/SprintfReturnTypeProvider.php#L73-L86

It states
> it makes no sense to use sprintf when there is only 1 arg (the format) 
> as it wouldn't have any placeholders 

Which sure, is true, but that's a code style issue and not statically incorrect code.

tscni:
I guess this relates to https://github.com/vimeo/psalm/issues/9987
Seems like switching this case (and similarly for printf) over to `RedundantFunctionCall` might be the way to go?

tscni:
Sounds great!

I hadn't thought about the non-literal case at all.
Maybe the issue message could hint at the problem a bit more (something about not being able to determine placeholders)?

I wonder how worthwhile it'd be to add a new issue type for this case in general, even if there are 2 arguments or more, because it's never clear how many placeholders there really are. (e.g. a "possibly too few arguments" or "non analyzable function argument" - not sure if a similar type already exists)

ygottschalk:
>  it's just a bad practice

100% agree!

> I'll change printf with 1 arg to only report `RedundantFunctionCall`

I would either not report an issue (since `printf` does output, the call is not redundant, just bad practice) or raise some issue with a level of `Info`.

tscni:
As a user ideally I'd want to:
- Have a `sprintf()` call with a single format-less literal (or analyzable variable) argument be reported as `RedundantFunctionCall`
- Have a `printf()` call with a single format-less literal (or analyzable variable) argument be reported as "some issue type" (e.g. a `SuboptimalFunctionCall`) that doesn't necessarily have to be an error by default
- Have a `sprintf()` / `printf()` call with a single literal (or analyzable variable) argument with a formatting directive be reported as `TooFewArguments`
- Have a `sprintf()` / `printf()` call with an unanalyzable first argument, no matter the other arguments, be reported as "some issue type" (e.g. `UnanalyzableFormatArgument`) that doesn't necessarily have to be an error by default.
  Having it in a try-catch block that catches `ArgumentCountError` should suppress the issue though. (if that's the only error that could be thrown in that case, I haven't checked)

danog:
[...]

Issue #9987, opened 2023-07-03
psalm insists that `printf()` needs at least two arguments


pilif:
https://psalm.dev/r/fbda7f627d

this is totally valid code. Now granted, `echo` could have been used, but it's totally valid.

PR #10088, opened 2023-08-04
make (s)printf error reporting more correct/literal

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/10021
Fix https://github.com/vimeo/psalm/issues/9987 again now for all cases (specifically https://github.com/vimeo/psalm/issues/9987#issuecomment-1624360506)

Use RedundantFunctionCall instead of InvalidArgument, where it's technically valid. Report TooManyArguments when a format without placeholders is used Report an error for splat with vprintf/vsprintf


orklah:
Thanks!

PR #10091, opened 2023-08-08
Improve Reflection stubs

vudaltsov:
ReflectionParameter::getPosition(): non-negative-int
ReflectionProperty non-empty-string $name
ReflectionMethod non-empty-string $name

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -10631,7 +10631,7 @@ return [
 'ReflectionParameter::getDefaultValue' => ['mixed'],
 'ReflectionParameter::getDefaultValueConstantName' => ['?string'],
 'ReflectionParameter::getName' => ['string'],
-'ReflectionParameter::getPosition' => ['int'],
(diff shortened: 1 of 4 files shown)


vudaltsov:
Okay, but this is such a rear case that it's easier to not take it into account.

orklah:
This has impacts for libraries that need to inherit those class, for example BetterReflection so it would be better to let it that way

jorgsowa:
Could you limit scope to `ReflectionParameter::getPosition`? Then I think it would be useful to merge it.

vudaltsov:
Removed `ReflectionProperty` change. And added another two:
1. ReflectionAttribute::getTarget() returns Attribute::TARGET_* instead of int-mask. Even for attributes of promoted properties (https://3v4l.org/rpM1t).
2. *::getAttributes() always returns a list (https://3v4l.org/LYAMb).

weirdan (approved):

weirdan:
Thanks!

PR #10092, opened 2023-08-08
Respect stubs in all cases

mmcev106:
If stubs should NOT be respected in all cases, I can modify this to only change behavior in the specific test case. Don't hesitate to ask if you have any questions, or would prefer this be implemented in a different way.

### DIFF
--- a/src/Psalm/Internal/Codebase/Functions.php
+++ b/src/Psalm/Internal/Codebase/Functions.php
@@ -79,9 +79,8 @@ class Functions
             $function_id = substr($function_id, 1);
         }
 
-        $from_stubs = false;
         if (isset(self::$stubbed_functions[$function_id])) {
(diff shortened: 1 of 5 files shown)


mmcev106:
I'm working on fixing the code style now.

orklah:
test-with-real-projects is broken for another reason, don't worry about it ;)

mmcev106:
Thanks @orklah.  The code style has been fixed.

orklah:
I'm afraid of possible side effects from this. Could you target master instead so it will release with Psalm 6? it will be easier to manage if there are issues

mmcev106:
@orklah, I've changed the base branch to `master`, and added an empty commit to force tests to re-run.  Let me know if anything else is needed.

orklah:
Thanks!

PR #10093, opened 2023-08-08
argv and argc were inconsistent in Context with VariableFetchAnalyzer

kkmuffme:
Not sure why argv/argc are declared in context independently in the first place though


orklah:
Thanks! Not sure why it's duplicated either

PR #10098, opened 2023-08-09
Prevent autoloader or plugin exit status codes of zero from incorrectly resulting in a zero status from Psalm

mmcev106:


mmcev106:
I'm working on the failures...

mmcev106:
I think this is ready to go.

mmcev106:
Converting back to draft.  It looks like a similar problem happens with die/exit calls inside plugin hooks.  I'm going to work on a more general solution.

weirdan:
What if it worked the other way around:

1. A shutdown handler would always be registered
2. It would always report any exit as problematic
3. Unless the exit is intentional (a certain method called)
4. That certain method would unregister the crash handler and then exit

mmcev106:
@weirdan, I already has something along those lines in the works.   What do you think of the implementation?

weirdan:
The solution as it is now lacks totality and is thus hard to reason about. For me, it would be important for the crash handler to be registered **always** and report **any** exit as problematic (unless the exit is intentional).

Also, it would make sense to apply it to all cli binaries we ship (except, probably, `psalm-plugin`). There's no reason why `psalm` should have a crash handler, and `psalter` should not.

mmcev106:
@weirdan, I think you're absolutely right from an idealist perspective.   However, I looked into it a bit, and am I'm struggling to justify the time to do as you ask.  What would it take for you considering some level of incremental improvement, rather than waiting (perhaps indefinitely) for someone able to cover all cases?  In my mind this is easy to reason about, as covering some cases is better than covering none  (the old aphorism [perfect is the enemy of good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good)).

Regardless, I will implement any additional cases I encounter as I go about my primary work goals.  I should probably also share that part of my hesitation is my lack of familiarity outside of cases I've experienced directly.  I think expanding the scope to use cases I am not familiar with would significantly increase the risk of unexpected side effects.

mmcev106:
@weirdan, I guess this isn't a priority enough to think about, and/or only an "all or nothing" approach will be accepted.  Instead of continuing to work from my fork, I just modified our CI integration scripts to look for the `Psalm was able to infer` last line of output to ensure Psalm actually finished (rather than the status code).

I'm sure y'all are doing your best.  However, I should be honest and communicate that I'm growing discouraged about continuing to invest time in contributing to Psalm...

orklah:
Sorry, we clearly lack maintainers...

I understand what you feel and I lost a lot of motivation these last years too

If you would like to become a maintainer, please ping me and I'll make things move. It would be a shame to let Psalm die...

mmcev106:
@orklah, I emailed you, and am curious to hear your thoughts.

PR #10104, opened 2023-08-12
Fix parameter having `object` type in PHPDoc and only `&` in the method's definition, used with code having an object calling it's method with itself as an argument for mentioned parameter

kubawerlos:
Sorry for the title, see the explanation below:

Having such code:
```php
<?php
class Foo {
    /**
    * @param object $o
    */
    public function bar(&$o): void {}
}
$x = new Foo();
$x->bar($x);
```
will result in [internal Psalm error](https://psalm.dev/r/3374fc1feb).

1. Removing `object` from PHPDoc, or
2. removing `&` from the function's definition, or
3. not passing `$x` to method of `$x`

will result in Psalm not crashing, so this is quite an edge case.


weirdan (approved):

weirdan:
Thanks!

PR #10107, opened 2023-08-12
mysqli_field_seek returns true

kamil-tekiela:
In PHP 8.0 I made this function always return true, but the documentation got updated only now. https://github.com/php/doc-en/commit/18d96d58e38ff747678163d6a01dcb946e6810a7


weirdan:
Thanks!

PR #10110, opened 2023-08-13
Fix `hash_pbkdf2` `$options` parameter

weirdan:
`$options` parameter actually appeared in 8.1[^1] but wasn't documented until 8.1.22 [^2]

[^1]: https://github.com/php/php-src/commit/110b4e90946b6d46864c54c0e15b7a0deb6ad4b3
[^2]: https://github.com/php/php-src/commit/7cae6eb8dbee7012691f89f5358ceee44c1b32c2


weirdan:
This fixes unit test failures on Win/8.1.22 (it appears 8.1.22 is not yet available for Linux jobs).

PR #10115, opened 2023-08-15
Filter baseline when checking individual files

### LINKED ISSUE #9689: findUnusedBaselineEntry + file argument
When enabling `findUnusedBaselineEntry` and using Psalm with a file argument, I would expect only the baseline entries pertaining to one of the files passed as argument to be taken into account.

Running with a file argument is used by phpactor, for instance, like this:
`'vendor/bin/psalm' '--show-info=true' '--output-format=json' '/home/greg/dev/doctrine-orm/major/lib/Doctrine/ORM/Mapping/AssociationMapping.php'`

What happens is that a huge JSON is produced, with one UnusedBaselineEntry per entry not corresponding to the file passed as argument, then it seems phpactor gives up on displaying the errors.

I'm using Psalm 5.9.0

edsrzf:
This prevents reporting unused baseline entries for other files that aren't being checked. Fixes #9689.

Very open to suggestions on other approaches or ways to test this. It works for my use case.

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -401,7 +401,7 @@ final class Psalm
                 !$paths_to_check,
                 $start_time,
                 isset($options['stats']),
-                self::initBaseline($options, $config, $current_dir, $path_to_config),
+                self::initBaseline($options, $config, $current_dir, $path_to_config, $paths_to_check),
             );
         } else {
             self::autoGenerateConfig($project_analyzer, $current_dir, $init_source_dir, $vendor_dir);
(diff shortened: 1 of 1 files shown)


weirdan (commented):
I would probably prefer filtering to happen in `ErrorBaseline::read()`.

Also, it looks like it would be beneficial to disable all write operations for baseline when running on a subset of a codebase. I.e. allow generation and update (?) only when there are no `$paths_to_check`.

weirdan (approved):

weirdan:
Thanks!

edsrzf:
Thanks for the additions. I did intend to get back to this but didn't find the time just yet.

PR #10128, opened 2023-08-18
Add exec output param out

kkmuffme:


stubs/CoreGenericFunctions.phpstub
```diff
@@ -1224,7 +1224,7 @@ function preg_replace_callback($pattern, $callback, $subject, int $limit = -1, &
  *
  * @param non-empty-string $pattern
  * @param string $subject
- * @param mixed $matches
+ * @param array|null $matches
```
weirdan, line 1227: Looks like you accidentally committed more changes than intended.

weirdan: I'm not sure we should. It looks like non-array vars are accepted just fine: https://3v4l.org/7pgBM.

weirdan (approved):

weirdan:
Thanks!

PR #10150, opened 2023-08-25
Assertion enhancements regarding `value-of` in combination with backed enums

### LINKED ISSUE #10082: Backed enums should not behave as if they were values of their backing type
https://psalm.dev/r/a843ec35f3

But in reality it is not possible and we will get a runtime error in this case.
https://3v4l.org/ovHD0

boesing:
This patch reworks the way how psalm is interacting with enums.

In #9655, the initial `value-of` feature was introduced but also introduced a regression regarding backed enum cases were incorrectly detected as their literal representations.
This problem was reported in #10082.

While I was already working on a fix, I also extended the `value-of` assertion to reconcile `value-of<MyEnum>` to allow to assert *ANY* value of an enum case.

Fixes #10082

/cc @weirdan

### DIFF
--- a/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
+++ b/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
@@ -2991,7 +2991,10 @@ class SimpleAssertionReconciler extends Reconciler
             // For value-of<MyBackedEnum>, the assertion is meant to return *ANY* value of *ANY* enum case
             if ($enum_case_to_assert === null) {
                 foreach ($class_storage->enum_cases as $enum_case) {
-                    assert($enum_case->value !== null, 'Verified enum type above, value can not contain `null` anymore.');
+                    assert(
+                        $enum_case->value !== null,
+                        'Verified enum type above, value can not contain `null` anymore.',
+                    );
                     $reconciled_types[] = Type::getLiteral($enum_case->value);
                 }
 
--- a/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
+++ b/src/Psalm/Internal/Type/SimpleAssertionReconciler.php
@@ -83,6 +83,7 @@ use function assert;
 use function count;
 use function explode;
 use function get_class;
+use function in_array;
 use function is_int;
 use function min;
 use function strlen;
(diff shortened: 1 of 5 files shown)


src/Psalm/Internal/Type/SimpleAssertionReconciler.php
```diff
@@ -2951,6 +2955,58 @@ private static function reconcileClassConstant(
         return TypeCombiner::combine(array_values($matched_class_constant_types), $codebase);
     }
 
+    private static function reconcileValueOf(
+        Codebase $codebase,
+        TValueOf $assertion_type
+    ): ?Union {
+        $reconciled_types = [];
+
+        // For now, only enums are supported here
```
weirdan, line 2964: Our docs say that `value-of<>` also works with arrays: https://psalm.dev/docs/annotating_code/type_syntax/utility_types/#value-oft

boesing: Is there a test for this?

boesing: btw: these changes do actually target assertions, even tho `value-of` worked with `array`, that is not the case for the assertion syntax.
I do not really have a need for this and thus, we should keep this targeting enums for now.

If some1 wants to add array assertions as well, the code can be extended.

boesing:
No clue what is going on in https://github.com/vimeo/psalm/actions/runs/5981011747/job/16228173278?pr=10150#step:8:167

Does not seem to be related to this PR, can some1 confirm?

weirdan:
Definitely unrelated; it's caused by the new signature change that landed in php yesterday: https://github.com/php/php-src/pull/11430

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
weirdan: I was just making sure we don't loose any functionality.

src/Psalm/Type.php
```diff
@@ -258,6 +259,19 @@ public static function getNumericString(): Union
         return new Union([$type]);
     }
 
+    /**
+     * @param int|string $value
+     * @return TLiteralString|TLiteralInt
+     */
+    public static function getLiteral($value)
```
weirdan, line 266:
```suggestion
    public static function getLiteral($value): Atomic
```

src/Psalm/Internal/Type/SimpleAssertionReconciler.php
boesing: yah, I wasn't 100% sure as well. Since there was no test for the backed enum type as well, I just double checked that again. 😬

weirdan:
Thanks!

PR #10156, opened 2023-08-27
Prevent infinite-loop when resolving type-aliases

boesing:
The type-expander returns the same `intersection_type` in case something is not properly expandable. To avoid infinite-loop, we do explicitly verify that the expanded  alias is actually resolved

### DIFF
--- a/src/Psalm/Internal/Type/TypeParser.php
+++ b/src/Psalm/Internal/Type/TypeParser.php
@@ -1629,9 +1629,7 @@ class TypeParser
                 continue;
             }
 
-            $modified = true;
-
-            $normalized_intersection_types[] = TypeExpander::expandAtomic(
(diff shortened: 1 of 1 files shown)


weirdan (approved):

weirdan:
Would love to see a test for it though.

boesing:
I wish I would be able to provide a test. 
I stumbled upon this issue yesterday by only analyzing a single file. 
Most probably that only happened as I had some cache available based on the previous runs or sth. like that.

I also remember that I hade some weird issues (infinite loop) when switching between branches in `laminas-servicemanager` which was most likely also related to this specific bug.

I'll see if I can find some time next weekends to actually dive deeper into this so I can provide a test.

weirdan:
Thanks!

PR #10157, opened 2023-08-27
Properly inherit assertions from parents or implemented faces

boesing:
As stated in https://github.com/vimeo/psalm/pull/10010#discussion_r1306718602, psalm does not properly detect assertions on methods using `{@inheritdoc}` where the inherited method does provide assertions.

Not 100% sure if this also fixes #10010 but at least a part of it.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/ExistingAtomicMethodCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/ExistingAtomicMethodCallAnalyzer.php
@@ -778,7 +778,9 @@ class ExistingAtomicMethodCallAnalyzer extends CallAnalyzer
 
                 $replacement->rule[] = new Assertion\IsType(new TTemplateParam(
                     $assertion->type->param_name,
-                    reset($potential_assertion_providing_classlike_storage->template_types[$assertion->type->param_name]),
+                    reset(
+                        $potential_assertion_providing_classlike_storage->template_types[$assertion->type->param_name],
+                    ),
                     $potential_assertion_providing_class,
                 ));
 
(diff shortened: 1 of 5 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/ExistingAtomicMethodCallAnalyzer.php
```diff
@@ -415,11 +420,45 @@ public static function analyze(
                 }
             }
 
-            if ($method_storage->assertions) {
+            $assertions = $method_storage->assertions;
```
boesing, line 423: I am not really proud of this but it works, happy to get some ideas on how to properly refactor this to become actually more readable.

boesing: Another thought here is, why do we even need `inheritdoc`.
I am aware that it might be helpful somewhere (even tho it isn't used anywhere else than in this line), but I'd rather deprecated that property and instead recursively try to detect the inherited parent and use that docblock to merge with the methods one.

This way, every annotation/etc. would be automatically available and usable - no matter if it was inherited or not.

boesing:
No clue what is going on in https://github.com/vimeo/psalm/actions/runs/5993765961/job/16254529984?pr=10157 

Might be unrelated. I've used this to actually scan https://github.com/laminas/laminas-servicemanager/pull/204 which worked as expected.

boesing:
This also requires methods to actually provide `@inheritdoc` which is nowadays mostly omitted since the interface/extended method usually provides the details.

I am fine with actually adding `@inheritdoc` but what do you guys think about this? Should we always implicitly grab extensions from inherited classes/interfaces without requiring `@inheritdoc`? 

Just have to remove the line which explicitly checks for `$method_storage->inheritdoc`.

weirdan:
I don't think we should be using `inheritDoc` at all. Possibly there are some reasons for using it I'm unaware of though.

Perhaps we should experiment with removing those inheritdoc checks and see what tests break.

boesing:
Removed the check for `@inheritdoc`. 
Tests still passing 🎉

weirdan (approved):

PR #10161, opened 2023-08-30
`composer.json` tweaks

weirdan:
Psalm is not a library.

Also updated `composer lint` description.


drupol:
I totally agree with the change. Do you think we could now have the composer.lock file living in the repo to enable reproducible builds?

PR #10164, opened 2023-08-30
Parallel

danog:


docs/running_psalm/issues/PossiblyUnusedMethod.md
```diff
@@ -9,6 +9,7 @@ with `@psalm-api`.
 ```php
 <?php
 
+/** @psalm-api */
```
weirdan, line 12: This is supposed to be a snippet that emits `PossiblyUnusedMethod`. I think we have a test that makes sure the issue being documented is emitted by the example the doc for that issue contains.

danog: Yep noticed that, will have to either revert or patch the test

PR #10169, opened 2023-08-31
InheritorViolation was only being triggered on grand-childen classes

### LINKED ISSUE #10167: `@psalm-inheritors` does not catch error at class declaration
Took an example from the [documentation](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-inheritors): https://psalm.dev/r/0d2892f51f

No issues are reported

robchett:
Fixes #10167

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -36,6 +36,7 @@ use Psalm\Issue\DuplicateEnumCaseValue;
 use Psalm\Issue\ExtensionRequirementViolation;
 use Psalm\Issue\ImplementationRequirementViolation;
 use Psalm\Issue\InaccessibleMethod;
+use Psalm\Issue\InheritorViolation;
 use Psalm\Issue\InternalClass;
 use Psalm\Issue\InvalidEnumCaseValue;
 use Psalm\Issue\InvalidExtendClass;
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Analyzer/ClassAnalyzer.php
```diff
@@ -271,6 +272,22 @@ public function analyze(
             );
         }
 
+        $classUnion = new Union([new TNamedObject($fq_class_name)]);
```
weirdan, line 275: we use snake case for variable names

src/Psalm/Internal/Analyzer/ClassAnalyzer.php
```diff
@@ -271,6 +272,22 @@ public function analyze(
...
+        foreach ($storage->parent_classes + $storage->direct_class_interfaces as $parent_class) {
```
weirdan, line 276: This makes me wonder whether the following should be flagged or not:

```php
/** @psalm-inheritors Whatever */
interface I {}
interface IC extends I {}
```

robchett: I'm of the belief extending the interface should be explicitly stated if the annotation is used. Updated to analyse interfaces too.

weirdan (approved):

weirdan:
Thanks!

PR #10187, opened 2023-09-09
`DateTimeZone::getName()` always returns `non-empty-string`

gsteel:
Forgive me if I'm incorrect, but, DateTimeZone::__construct() does not accept an empty string, nor an invalid timezone identifier, so it's therefore infeasible for getName() to return an empty string right?


weirdan:
Thanks!

PR #10189, opened 2023-09-10
Inherited assertions v2

boesing:
With #10157, some type-juggling was introduced regarding assertions which were inherited.

Since there were still some issues (which are now reflected by the tests), I tried to fix these as well.

I found a check in https://github.com/vimeo/psalm/blob/f782767438dd1678e60f3f97cfc66e1252e79b10/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ClassTemplateParamCollector.php#L51-L53 where I do not understand why these are made. By removing them, the inheritance issues are gone (due to the fact that template parameters from class level are now properly parsed).

### DIFF
--- a/src/Psalm/Internal/Codebase/AssertionsFromInheritanceResolver.php
+++ b/src/Psalm/Internal/Codebase/AssertionsFromInheritanceResolver.php
@@ -5,17 +5,13 @@ declare(strict_types=1);
 namespace Psalm\Internal\Codebase;
 
 use Psalm\Codebase;
-use Psalm\Storage\Assertion\IsType;
 use Psalm\Storage\ClassLikeStorage;
 use Psalm\Storage\MethodStorage;
 use Psalm\Storage\Possibilities;
-use Psalm\Type\Atomic\TTemplateParam;
 
 use function array_filter;
-use function array_map;
 use function array_merge;
 use function array_values;
-use function reset;
 use function strtolower;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Psalm/Internal/Codebase/AssertionsFromInheritanceResolver.php
```diff
@@ -61,80 +57,9 @@ public function resolve(
...
-                ),
-                $potential_assertion_providing_method_storage->assertions,
-            );
+            $assertions += $potential_assertion_providing_method_storage->assertions;
         }
 
         return $assertions;
     }
-
-    /**
```
boesing, line 78: By removing https://github.com/vimeo/psalm/blob/f782767438dd1678e60f3f97cfc66e1252e79b10/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ClassTemplateParamCollector.php#L51-L53, these LoC are not necessary anymore.

boesing:
Hm, pretty sure these `ArrayAssignmentTest` issues are related to these changes but I am unsure if these are actually problematic since now it holds the `ArrayObject` types.

src/Psalm/Internal/Codebase/AssertionsFromInheritanceResolver.php
```diff
@@ -61,80 +57,9 @@ public function resolve(
              * Since the inheritance does not provide its own assertions, we have to detect those
```
orklah, line 57: Seems like this comment should change too?

boesing: well Actually we still have to detect those, but in this case we just merge. 
any suggestions? I have no clue how to modify this for being more accurate 🫤

orklah: nope! I was just making sure there was not an outdated comment left behind :)

orklah:
Thanks!

PR #10190, opened 2023-09-10
Narrow `ArrayObject` `TKey` template

boesing:
Since arrays cannot have `mixed` array keys, this patch is narrowing the `ArrowObject`.

I had a quick check in PHPStan and this change aligns with how PHPStan is handling `ArrayObject` keys as well:
https://github.com/phpstan/phpstan-src/blob/ab154e1da54d42fec751e17a1199b3e07591e85e/stubs/ArrayObject.stub#L38


weirdan (approved):

weirdan:
Thanks!

PR #10191, opened 2023-09-10
Resolve inherited conditional return types

boesing:
With #10189, we are enhancing inherited assertions. With this patch, we do also resolve recursive conditional return types.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallReturnTypeFetcher.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallReturnTypeFetcher.php
@@ -183,6 +183,7 @@ class MethodCallReturnTypeFetcher
                 $self_fq_class_name,
                 $statements_analyzer,
                 $args,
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodCallReturnTypeFetcher.php
```diff
@@ -449,7 +450,7 @@ public static function taintMethodCallResult(
                 $stmt_var_type = $context->vars_in_scope[$var_id]->setParentNodes(
                     $var_nodes,
                 );
-                
+
```
boesing, line 453: sorry, I am unable to revert this change, my IDE is auto removing these whitespaces and restoring them is more annoying than just having these deleted. I hope thats okay.

weirdan: That's totally fine.

orklah:
Thanks!

PR #10194, opened 2023-09-11
Allow class-string<T> where T is a union

### LINKED ISSUE #8582: Invalid class-string<T> where T as A|B
https://psalm.dev/r/41dd38b32b

MidnightDesign:
Fixes #8582

### DIFF
--- a/src/Psalm/Internal/Type/TypeParser.php
+++ b/src/Psalm/Internal/Type/TypeParser.php
@@ -462,12 +462,6 @@ class TypeParser
             );
         }
 
-        if (!$as->isSingle()) {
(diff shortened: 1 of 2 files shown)


MidnightDesign:
I was able to trace this condition to this commit: https://github.com/vimeo/psalm/commit/9d8c2792#diff-5d848b61764f2ca11ffb4b0f71a1a872d9d78c7ff26eba6c84a0c42f254bab37R491-R495

Since that commit is almost five years old and no tests are breaking without that condition, I don't think there's any harm in removing it.

orklah:
Thanks!

PR #10202, opened 2023-09-13
Issue handlers priority

### LINKED ISSUE #10201: Plugins can define IssueHandlers that have priority over the user configuration
When using "Config::setCustomErrorLevel" or "Config::setAdvancedErrorLevel" methods, any IssueHandler object already defined by the configuration will be overriden. In many use cases it's not the expected behavior (the user configuration should most of the time have priority over the configuration defined in a plugin by example.

gmessier:
Fix #10201

New "Config::safeSetCustomErrorLevel" and "Config::safeSetAdvancedErrorLevel" are added (to not break the current behavior and create unexpected errors). Those methods will create a IssueHandler object only if the issueHandlers array doesn't already contain this issue key.

Tests have been added to validate the behavior of the current ("Config::setCustomErrorLevel" and "Config::setAdvancedErrorLevel") and new methods.

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -1481,10 +1481,24 @@ class Config
         $this->issue_handlers[$issue_key]->setCustomLevels($config, $this->base_dir);
     }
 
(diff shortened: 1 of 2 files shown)


orklah:
Thanks!

PR #10207, opened 2023-09-15
Allow object assertions for `mixed` values

### LINKED ISSUE #10203: `psalm-assert` with object-shape does not work (anymore)
Not 100% sure where this was lost but AFAIR that worked in earlier versions.

https://psalm.dev/r/cb6b2249ae

boesing:
Fixes #10203


orklah:
Thanks!

PR #10209, opened 2023-09-16
Only the binary op 'plus' works with two arrays

### LINKED ISSUE #2123: Psalm should infer that result of multiplication is never an array.
The same goes for all numeric operations, excluding addition.

This example is used to show the types psalm infers

```php
<?php
$x = getopt('', ['value:']);
// correctly infers array<int, mixed>|string|false for $x['value']
// Observed:  PossiblyInvalidOperand - src/array.php:5:23 - Cannot add an array to a non-array int
// Expected:  PossiblyInvalidOperand - src/array.php:5:23 - Cannot use an array and a non-array int in a binary operation (or something along those lines)
$result = $x['value'] * 3;
// Observed: Argument 1 of fread expects resource, array<array-key, mixed> provided
// Expected: Should warn about $result being numeric
fread($result, 1);
```

robchett:
Treat the result of any other operation as int|float 
Fixes #2123

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
@@ -542,97 +542,97 @@ class ArithmeticOpAnalyzer
                     $has_valid_right_operand = true;
                 }
 
-                $result_type = Type::getArray();
-
                 return null;
             }
 
-            $has_valid_right_operand = true;
-            $has_valid_left_operand = true;
+            if ($parent instanceof PhpParser\Node\Expr\BinaryOp\Plus) {
+                $has_valid_right_operand = true;
+                $has_valid_left_operand = true;
 
-            if ($left_type_part instanceof TKeyedArray
-                && $right_type_part instanceof TKeyedArray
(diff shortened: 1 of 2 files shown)


orklah:
Thanks!

weirdan:
This seems to fail on master: https://github.com/vimeo/psalm/actions/runs/6316042813/job/17149847997

robchett:
It looks like this if statement is bleeding out of it's context
```php
        /** @psalm-trace $left_type_part */ 
        // -- Psalm\Type\Atomic
        if ($left_type_part instanceof TArray
            || $right_type_part instanceof TArray
            || $left_type_part instanceof TKeyedArray
            || $right_type_part instanceof TKeyedArray
        ) { ... not all branches return }
        /** @psalm-trace $left_type_part */ 
        // -- Psalm\Type\Atomic\TArray|Psalm\Type\Atomic\TKeyedArray
```

Is it possible there is a regression elsewhere in `master` that is only showing it's head here?

robchett:
Digging deeper it looks like the removal of TList was the issue, it sufficiently changed the if statement to surface what I assume is a bug - a minimal repo is this is https://psalm.dev/r/577ae8535e

Would baselining these warnings or re-annotating $left_type_part/$right_type_part to Atomic be OK until that issue can be resolved?

Issue #9764, opened 2023-05-09
Don't throw `PossiblyFalseArgument` on pass `false|list<string>` to `in_array`


vjik:
https://psalm.dev/r/e229d24bd7

Var is `false|list<string>`, but `PossiblyFalseArgument` into `in_array` do not throw.

Expected:

```
[PossiblyFalseArgument](https://psalm.dev/104) - 7:26 - Argument 2 of in_array cannot be false, possibly array<array-key, mixed> value expected
```

orklah:
It will throw if you put this config to false: https://psalm.dev/docs/running_psalm/configuration/#ignoreinternalfunctionfalsereturn

It does not by default because of the bad habit PHP had to return exotic false/null when given invalid input before PHP 8 (so Psalm would ignore them to preven bothering you to check for every possible exotic value). I'd be open to change the default value for this config in order to report them by default now that PHP 8 cleaned up everything

weirdan:
Fixed in #10211

PR #10211, opened 2023-09-17
Disable ignoreInternalFunction(False|Null)Return by default

robchett:
as per #9764

Breaking change for Psalm 6?

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -3,6 +3,8 @@
 
 - The minimum PHP version was raised to PHP 8.1.17.
 
+- [BC] The configuration settings `ignoreInternalFunctionFalseReturn` and `ignoreInternalFunctionNullReturn` are now defaulted to `false`
+
 - [BC] Switched the internal representation of `list<T>` and `non-empty-list<T>` from the TList and TNonEmptyList classes to an unsealed list shape: the TList, TNonEmptyList and TCallableList classes were removed.
   Nothing will change for users: the `list<T>` and `non-empty-list<T>` syntax will remain supported and its semantics unchanged.
   Psalm 5 already deprecates the `TList`, `TNonEmptyList` and `TCallableList` classes: use `\Psalm\Type::getListAtomic`, `\Psalm\Type::getNonEmptyListAtomic` and `\Psalm\Type::getCallableListAtomic` to instantiate list atomics, or directly instantiate TKeyedArray objects with `is_list=true` where appropriate.
--- a/bin/update-property-map.php
+++ b/bin/update-property-map.php
@@ -93,9 +93,9 @@ libxml_use_internal_errors(true);
 foreach ($files as $file) {
     $contents = file_get_contents($file);
     // FIXME: find a way to ignore custom entities, for now we strip them.
-    $contents = preg_replace('#&[a-zA-Z\d.\-_]+;#', '', $contents);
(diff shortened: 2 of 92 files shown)


orklah:
Cool!

Seems like it makes a lot of tests fail

Could you look if those are legit failures (because Psalm didn't check possible false/null values where they could happen?). If so, I'd suggest re-enabling the config for Psalm instead of trying to fix every case

robchett:
@orklah 
I've added a new config for specifying which functions to ignore null/false on, which allowed me to find out what functions were not being handled correctly. 

```xml
<ignoredInternalFunctionFalseReturn>
        <referencedFunction name="getcwd"/>
        <referencedFunction name="ob_get_clean"/>
        <referencedFunction name="glob"/>
        <referencedFunction name="filemtime"/>
        <referencedFunction name="opendir"/>
        <referencedFunction name="scandir"/>
        <referencedFunction name="readdir"/>
        <referencedFunction name="tempnam"/>
        <referencedFunction name="phpversion"/>
        <referencedFunction name="file_get_contents"/>
        <referencedFunction name="reset"/>
        <referencedFunction name="end"/>
        <referencedFunction name="fwrite"/>
        <referencedFunction name="fread"/>
        <referencedFunction name="realpath"/>
        <referencedFunction name="readlink"/>
        <referencedFunction name="json_encode"/>
        <referencedFunction name="curl_init"/>
        <referencedFunction name="base64_decode"/>
        <referencedFunction name="igbinary_serialize"/>
        <referencedFunction name="strtotime"/>
        <referencedFunction name="preg_split"/>
        <referencedFunction name="ini_get"/>
        <referencedFunction name="getopt"/>
    </ignoredInternalFunctionFalseReturn>

    <ignoredInternalFunctionNullReturn>
        <referencedFunction name="preg_replace"/>
        <referencedFunction name="array_shift"/>
        <referencedFunction name="key"/>
    </ignoredInternalFunctionNullReturn>
```

Most seem safe, but there are a few file/io ones in there that I think I should remove the suppression for and actually add checks too.

Are you happy with this approach for a fine-tuned list of suppressed functions instead of a blanket suppression on all internal methods?

orklah:
I'm not sure I understood what you meant

I was kinda expecting that the errors Psalm raised on itself were going to be places where a function can legitimately return false/null but we ignored it (probably because it can rarely happen but still). So the fix for that would have been either to fix each place in the code (pretty tedious) or just baseline those

What I understand is that you added a config so we can restore the old ignore for specific functions. I don't think this config will be used by a lot of users. They'll either don't want to bother with false/null to begin with and ignore everything or they'll want to have the cleanest code and fix what they can and baseline the rest in the meantime.

This seems like the kind of feature for power users that I would prefer to be solved with custom stubs if possible (to avoid cluttering the core too much)

jack-worman:
Casting false and null types to string is not the correct solution. This is just masking the issue. If an internal function returns null or false, it probably should throw an exception instead, since most of the time these are errors.

At my work, we use https://github.com/thecodingmachine/safe to handle these issues.

orklah:
> Casting false and null types to string is not the correct solution

Indeed but every improvement to the codebase is welcome, even when it's incremental. Before this PR, they were completely ignored and now they're explicitly handled (through casting). If this itch someone the wrong way, a future PR could improve the situation :)

orklah:
Thanks!

Issue #8047, opened 2022-06-02
Non-variable references cannot be returned by reference


jack-worman:
https://psalm.dev/r/88d9fd9adf

You cannot return a value by reference: https://3v4l.org/Ntkr0

weirdan:
Fixed in #10227

PR #10227, opened 2023-09-24
Introduce NonVariableReferenceReturn issue

tuqqu:
Emit a NonVariableReferenceReturn issue when returning by-reference functions return a non-variable reference.

Currently, PHP produces a notice when such a function is invoked: https://3v4l.org/bMabU

Close: #8047


orklah:
Thanks!

Issue #10239, opened 2023-09-30
cache issue


ging-dev:
sample code:
```php
<?php

/**
 * @template K
 */
class CacheProvider
{
    /** @var K */
    private $cache;
}
```
run psalm with the above code, then edit the docblock of `CacheProvider`:

```php
<?php

/**
 * @template T
 */
class CacheProvider
{
    /** @var K */
    private $cache;
}
```

Run `psalm` again:

- with cache: no new error messages appear. 
- with `--no-cache`: `Docblock-defined class, interface or enum named K does not exist (see https://psalm.dev/200)`

It seems `psalm` doesn't re-analyze the class in case the class's docblock changes.

PR #10240, opened 2023-09-30
fix: #10239

ging-dev:
close: #10239


ging-dev:
Ready for review, @orklah Can you take a look at it?

orklah:
Would it be possible to have a test for this?

ging-dev:
> Would it be possible to have a test for this?

Sure, added tests

orklah:
Thanks!

PR #10244, opened 2023-10-03
Add stream_socket_shutdown socket_shutdown to impure function list

### LINKED ISSUE #10235: stream_socket_shutdown & socket_shutdown call is not used
Hello!

An error is reported for the following two functions:
- stream_socket_shutdown
- socket_shutdown

The error is:
UnusedFunctionCall - The call to socket_shutdown is not used (see https://psalm.dev/206)
UnusedFunctionCall - The call to stream_socket_shutdown is not used (see https://psalm.dev/206)

This is wrong, since both functions can be used without doing anything with the returned result.

Thank you.

tuqqu:
Closes #10235 
Additionally, I have removed a duplicate entry from the list

### DIFF
--- a/dictionaries/ImpureFunctionsList.php
+++ b/dictionaries/ImpureFunctionsList.php
@@ -95,7 +95,6 @@ return [
     'mcrypt_generic_deinit' => true,
     'mcrypt_module_close' => true,
     // internal optimisation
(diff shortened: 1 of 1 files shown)


orklah:
Thanks!

PR #10248, opened 2023-10-03
Emit `MethodSignatureMismatch` when descendant does not return by reference

### LINKED ISSUE #10247: Missing `MethodSignatureMismatch` when by-ref return is absent
The following snippet is a fatal error in PHP:
```php
interface Foo {
    public function &foo(): int;
}

class Bar implements Foo {
    public function foo(): int {
        return 1;
    }
}
```
https://psalm.dev/r/141fb7ac5e

Expected behaviour:
a `MethodSignatureMismatch` should be emitted here

tuqqu:
Closes #10247

### DIFF
--- a/src/Psalm/Internal/Analyzer/MethodComparator.php
+++ b/src/Psalm/Internal/Analyzer/MethodComparator.php
@@ -313,6 +313,16 @@ class MethodComparator
             );
         }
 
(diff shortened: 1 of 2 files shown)


orklah:
Thanks!

PR #10254, opened 2023-10-04
Fix error message for implicitly returning functions with `never` return type

tuqqu:
* Before code like this

```php
function foo(): never {}
```
resulted in:
```
ERROR: InvalidReturnType

Not all code paths of foo end in a return statement, return type never expected
```

now it results in:
```
ERROR: InvalidReturnType

foo is not expected to return, but it does, either implicitly or explicitly
```

* Changed wording from `is not expected to return any values` to just `is not expected to return`, since empty returns or implicit returns do not return values

* Added tests for `never` return type


tuqqu:
@orklah the PR is ready to be merged. Please, let me know if there is anything else that needs to be done here

orklah:
Thanks!

PR #10261, opened 2023-10-07
Allow parameter names in callable docblocks with templates

### LINKED ISSUE #10217: Named type with phpstan-specific formatting cannot be parsed by psalm
Phpstan playground https://phpstan.org/r/5da87ef6-9869-40ca-b851-d4edcac82fba
Psalm playground https://psalm.dev/r/22d155399e

Source (exported types): https://github.com/webonyx/graphql-php/blob/29b5b613bd3ee34f227650f856ee8dd4ce151b88/src/Executor/ExecutionResult.php#L9-L34

----

Expected: there's no error for code that imports type `SerializableError`

Actual: 
```
ERROR: [InvalidTypeImport](https://psalm.dev/233) - 34:25 - Type alias SerializableError imported from ExecutionResult is not defined on the source class
```

```php
/**
 * @internal
 *
 * @phpstan-import-type SerializableError from ExecutionResult
 */
class Foo
{
    /**
      * @return SerializableError
     */
    public function format(\Throwable $error): array
    {
        return [
		'message' => 'message',
	];
    }
}
```

robchett:
fixes #10217. They are simply skipped, which is how `@method` is handled.

```
/**
 * @phpstan-type ErrorsHandler callable(array<Error> $errors, ErrorFormatter $formatter): SerializableErrors
 */
```

This docblock was throwing

> ERROR: [InvalidDocblock](https://psalm.dev/008) - 28:1 - Unexpected space

### DIFF
--- a/src/Psalm/Internal/Type/ParseTreeCreator.php
+++ b/src/Psalm/Internal/Type/ParseTreeCreator.php
@@ -553,24 +553,23 @@ class ParseTreeCreator
 
         $current_parent = $this->current_leaf->parent;
 
-        if ($current_parent instanceof CallableTree) {
-            return;
-        }
-
-        while ($current_parent && !$current_parent instanceof MethodTree) {
+        //while ($current_parent && !$method_or_callable_parent) {
+        while ($current_parent && !$current_parent instanceof MethodTree && !$current_parent instanceof CallableTree) {
             $this->current_leaf = $current_parent;
             $current_parent = $current_parent->parent;
         }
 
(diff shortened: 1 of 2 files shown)


robchett:
Just a not this doens't support `* @psalm-type D callable(string $c): mixed` which phpstan allows according to their docs

https://phpstan.org/writing-php-code/phpdoc-types#callables

Looks to be a fairly trivial thing to implement if it's ignored just like a non-referenced param

robchett:
Now fixed for cases with variadic, by-reference and templates are combined.

orklah:
Thanks!

PR #10262, opened 2023-10-08
Negated class_exist check on class-string converts to string instead of mixed

### LINKED ISSUE #10215: Narrowing a class-string with class_exists results in mixed instead of expected string
https://psalm.dev/r/c32c4d2cd8

We are narrowing a parameter of type `class-string` to determine if the `class_exists` or not; but if the class does NOT exist, them Psalm is telling me the type is `mixed`; however I would expect the type to be `string`. Not sure why this is happening?

robchett:
fixes #10215


orklah:
Thanks!

PR #10263, opened 2023-10-08
Maintain loop start val after increment/decrement

### LINKED ISSUE #10213: Lower boundary of `for` loop variable is not detected correctly
#### Reproducer

```php
$values = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];

for ($i = 100; $i < 110; $i++) {
    echo ((string) $values[$i - 100]) . "\n";
}
```

https://psalm.dev/r/97ca6dae65

#### Psalm's Output

ERROR: [InvalidArrayOffset](https://psalm.dev/115) - 6:20 - Cannot access value on variable $values using a int<-99, 9> offset, expecting int<0, 9>

#### Expected Output

No error

#### Notes

In the example above, Psalm correctly detects that `$i - 100` will never be greater than 9. However, it believes that the lower boundary of that operation is `-99`. Because the initial value of `$i` is 100 and because `$i` is only incremented afterwards, a value below 0 is impossible for `$i - 100`.

robchett:
Fixes #10213

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
@@ -823,12 +823,28 @@ final class ArithmeticOpAnalyzer
                                 $result_type = Type::getInt();
                             }
                         }
-                    } elseif ($parent instanceof VirtualPlus) {
-                        $start = $left_type_part instanceof TLiteralInt ? $left_type_part->value : 1;
-                        $result_type = Type::combineUnionTypes(Type::getIntRange($start, null), $result_type);
-                    } elseif ($parent instanceof VirtualMinus) {
-                        $start = $left_type_part instanceof TLiteralInt ? $left_type_part->value : 1;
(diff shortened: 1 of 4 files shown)


orklah:
(Note that I've not commented on this yet because there were errors on unit tests and I assumed you were still working on it :) )

robchett:
Yes, two steps forwards, one step back. Still working out the intricacies of the loopAnalyzer

robchett:
@orklah the lower bound is being properly detected.

I had to modify a test [tests/Loop/ForTest.php
](https://github.com/vimeo/psalm/pull/10263/files#diff-57748c2a6a2cf4964948682a48a088d50d82de6a544bd18b9abf8b2790208bf7L146) as it now shows an error which was a false-negative before. The continue statement was incorrectly suppressing an list-key offset if the form was changed very slightly. https://psalm.dev/r/303d48a67b shows it allowing list access via a negative key. It's not 100% as it hasn't reconciled the `$i > 0` and `$i--` to `int<0, 20>` before checking the array offset. I think the false positive is better than the false negative?

orklah:
Thanks! Seems like an improvement :)

PR #10266, opened 2023-10-08
Disallow `never` type for parameters

tuqqu:


orklah:
What's the purpose of this check? There's probably some fringe cases where having a never param could be useful and I don't think it would find a lot of bugs in projects

tuqqu:
@orklah it's a fatal error in PHP, https://3v4l.org/TSL24

orklah:
Oh, I didn't know, that's great then!

bdsl:
Since a value of never can never exist, a function that takes a never argument can never be called and is therefore technically useless.

There was an RFC to allow never in params but those functions would still have been unusable - the point was to have subclasses extend them into usable functions by replacing never with some other type, as a way to work around the lack of generics. The RFC was not accepted.

PR #10274, opened 2023-10-10
Fix inferring enum case value from a class constant

### LINKED ISSUE #10272: Internal Psalm error: 'Failed to infer case value' when using a constant from a class as enum value.
When using a constant from a class as the enum value, an error raises with the message `Failed to infer case value for $ENUM_CASE$`. Functionally it's accepted and works but Psalm throws an error.

Here's a code snippet to reproduce it https://psalm.dev/r/fb70dd37d2

tuqqu:
closes #10272


orklah:
Thanks!

PR #10276, opened 2023-10-11
Add intl extension stub

braindawg:
My company ran into some Psalm errors in our GitHub Actions container due to missing constants from the `intl` extension (specifically, the `NumberFormatter` class constants). This should add the constants exposed by the extension.


orklah:
Thanks!

PR #10277, opened 2023-10-11
str_replace / substr_replace signature improvements

mathroc:
* For `str_replace()` & `substr_replace()`, the return type is the same as the type of `$subject`
* For `str_replace()`, when `$replace` is an array it must be an array of strings 
* For `str_replace()`, the `$replace` argument can be an array only if `$search` is also an array, otherwise it's a `TypeError`


orklah:
Do you mean str_ireplace instead? I can find this one fine, and I don't see any stri_* one on php.net

orklah:
You have inconsistencies in the callmap. You created 4 signatures for `str_replace`. Each signature should either be present in Callmap_historical or be added in a previous PHP version (for example, if the fourth signature for str_replace was created in PHP 8.2, it won't appear in callmap_historical).

We have a documentation on psalm website about how these file works exactly, ping me if you need to find it

mathroc:
> Do you mean str_ireplace instead? I can find this one fine, and I don't see any stri_* one on php.net

:facepalm: that explains why I couldn't find it…

mathroc:
should be fine now

orklah:
Thanks!

PR #10279, opened 2023-10-13
fix final class constant type

trusting-thompson:
Fix type detection when accessing final class constants with statuc

https://psalm.dev/r/250eec0703


tuqqu:
FWIW, declaring final constants is a fatal error in PHP, so it's better to use public/protected visibility in the example

trusting-thompson:
Updated example
https://psalm.dev/r/d423062101

orklah:
Thanks!

PR #10289, opened 2023-10-17
Additional caching fixes

ging-dev:
This is an improved patch of https://github.com/vimeo/psalm/pull/10240. Add error case and fix it.


orklah:
Thanks!

PR #10290, opened 2023-10-17
Fix memory explosion with calls to method_exists

### LINKED ISSUE #9808: Several method_exists calls cause Psalm to exhaust memory
For whatever reason, six seems to be the magic number: https://psalm.dev/r/c92dc4976b

The error says:

```
PHP Fatal error:  Allowed memory size of 8589934592 bytes exhausted (tried to allocate 4294967304 bytes) in src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MissingMethodCallHandler.php on line 372
```

Looks like exponential memory growth?

robchett:
Fixes #9808. 

Growth is still 2^n, but massively better than it was ( I didn't try and work the big-O for that)

https://github.com/vimeo/psalm/blob/35183b9542b82d46f7e0d53629677220aafe27d0/src/Psalm/Internal/Analyzer/Statements/Expression/Call/MethodCallAnalyzer.php#L191

image


orklah:
Thanks!

PR #10305, opened 2023-10-21
Apply rector fixes

danog:
This PR merges 5.x into master and applies some modernization fixes using rector (+ some manual fixes too, mainly involving readonly properties).


danog:
Note that I've made the Config class final for performance purposes.

orklah:
I'm not sure about this. Every change that triggers a BC break is welcome, but I'd very much prefer avoiding any change in code style when we're between versions. Any change in V5 will be horrible to merge into master if it touches one of these changes.

Can't we wait until after V6 to do these changes?

danog:
@orklah Mmm technically this doesn't change code style in in of itself, in fact I haven't added rector to the dependencies or to the CI, this is more of a one-off pass that should not cause conflicts during merges from 5.x (except for the changes involving array_merge => array spread syntax, and those can be backported to v5, will submit a PR in a bit for that...).

orklah:
What I mean is every switch transformation to match for example will be a pain if we change something for this switch in V5. Likewise for promoted properties

danog:
>What I mean is every switch transformation to match for example will be a pain if we change something for this switch in V5. Likewise for promoted properties

Mmm yeah you're right, but on the other hand, I believe that changes to properties should be pretty rare (not sure about matches though...).

Anyway, if you want to postpone merging I can wait & create a separate PR just for the breaking changes, but personally I don't think merge conflicts from 5.x should be that big of an issue with this patchset...

orklah:
I think I'd prefer it that way, I'm not good enough at merging branches to risk it :D

danog:
OK then, could you merge https://github.com/vimeo/psalm/pull/10306 in the meantime? :)

danog:
I'll merge this now, I'll take it upon myself to do 5.x->master merges in the future (plus, I actually intend to make a first alpha/beta release for v6 soon...)

danog:
No dice, windows tests are segfaulting due to yet another PHP bug, damn it...

PR #10312, opened 2023-10-23
fix @param-out with named arguments

trusting-thompson:
Fix reference variable out type handling when using named arguments

https://psalm.dev/r/9cc467d755


orklah:
Thanks!

PR #10314, opened 2023-10-24
Narrow down `get_browser` returning array or object

simonberger:
The `$return_array` parameter decides if `get_browser` returns either `array|false` or `object|false`

Validator:
https://psalm.dev/r/9a12681e2c

### DIFF
--- a/stubs/CoreGenericFunctions.phpstub
+++ b/stubs/CoreGenericFunctions.phpstub
@@ -1797,3 +1797,9 @@ if (defined('GLOB_BRACE')) {
  * @psalm-taint-sink shell $command
  */
 function exec(string $command, &$output = null, int &$result_code = null): string|false {}
+
+/**
+* @return ($return_array is true ? array|false : object|false)
+* @psalm-ignore-falsable-return
+*/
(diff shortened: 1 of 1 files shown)


orklah:
Thanks! Can you add a @psalm-ignore-falsable-return too? It's the default behavior and it seems the false value from get_browser is not particularly meaningful

simonberger:
Ah interesting. Did not know about this annotation until now. I'll include the annotation tomorrow.

Yet I am a bit conflicted. On a correctly configured system you can trust it is indeed a wasted check, but get_browser does not work out of the box and will never work on many webhosting services as it needs an external config file loaded and configured in the php.ini.
In our case I'll log a failing return at least for now,
I still agree most users are just annoyed having to check for the false return and if not they do it also without psalm telling.

orklah:
Note that you can tell Psalm not to ignore unhandled false/null values through a config so when in doubt, I prefer allowing more leeway.

Also, those configs will have a more strict default value (so false/null values won't be ignored by default) starting on Psalm 6 so it won't make a big difference anyway

Thanks for the PR!

PR #10320, opened 2023-10-26
Fix auto completion by partial property or method

issidorov:
Supprot auto completion with usage partial string of property or method.

    namespace B;
    class A {
        public $aProp = 123;
        public $bProp = 234;
        public function bar() {
            $this->aPr
        }
    }

Expected result for auto completion `aProp`.

Full result partial code in IDE -- `$this->aProp`.


orklah:
Thanks!

Issue #10100, opened 2023-08-10
Unable to parse if there's a whitespace after trailing comma in an array shape.


hamsbrar:
https://psalm.dev/r/012810a311

weirdan:
Fixed in #10326

Issue #10095, opened 2023-08-09
class-string-map is not working as documented


stof:
The documentation shows `class-string-map<T of Foo, SomethingBasedOn<T>>` while the working syntax is `class-string-map<T as Foo, SomethingBasedOn<T>>`. It looks like `of` does not work there, only `as` (unlike in `@template`).

stof:
see https://psalm.dev/r/6f7e81fe99 for a reproducer, which is literally a copy-paste of the example in the documentation of `class-string-map`

weirdan:
Fixed in #10326

Issue #10089, opened 2023-08-06
Incorrect parse `@return` in docblock


vjik:
If return description starts from `{@see`, then psalm add this to type. It's wrong.

https://psalm.dev/r/9e5f61f5a0

weirdan:
Fixed in #10326

PR #10326, opened 2023-10-27
Consistent type docblock parsing

### LINKED ISSUE #10322: Cannot use multiline for `@template-use`
A newline in `@use` results in InvalidDocblock, while `@extends` supports multiline.

* `@use`: https://psalm.dev/r/ef05bbff9f
* `@extends`: https://psalm.dev/r/7aacc5733e

Thank you.

robchett:
fixes #10322 #10100 #10095 #10089

### DIFF
--- a/src/Psalm/Internal/Analyzer/CommentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/CommentAnalyzer.php
@@ -262,8 +262,9 @@ final class CommentAnalyzer
     {
         $docblock_type = (string) preg_replace('@^[ \t]*\*@m', '', $docblock_type);
         $docblock_type = (string) preg_replace('/,\n\s+}/', '}', $docblock_type);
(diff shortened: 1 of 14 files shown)


robchett:
Also fixes #10100

src/Psalm/Internal/Cli/Psalm.php
```diff
@@ -272,7 +272,7 @@ public static function run(array $argv): void
 
         self::emitMacPcreWarning($options, $threads);
 
-        self::restart($options, $threads, $progress);
+        //self::restart($options, $threads, $progress);
```
orklah, line 275: that should stay :)

robchett: Yes it should, forgot to exclude that (it's how I get psalm to run with Xdebug, do let me know if there is a better way). Now undone

orklah:
Thanks!

PR #10332, opened 2023-10-31
Stub constants for ZipArchive from ext-zip

braindawg:
This stubs out the class constants for `ZipArchive`, which I believe are the only constants introduced by the `zip` extension. This should allow Psalm to run over code utilizing any of these constants even if the analyzing system doesn't have ext-zip installed/enabled (e.g. a GitHub Actions container).


orklah:
Thanks!

PR #10333, opened 2023-10-31
Add constants from SOAP extension to stub

braindawg:
The SOAP extension stub was missing some constants we used (really just `SOAP_1_1` and `SOAP_1_2`), so I thought I'd add the rest of the constants declared by the extension to the stub. Values are all pulled straight from the PHP docs here:

https://www.php.net/manual/en/soap.constants.php

### DIFF
--- a/stubs/extensions/soap.phpstub
+++ b/stubs/extensions/soap.phpstub
@@ -1,5 +1,97 @@
 <?php
 
+<?php
+
+/**
+ * Constants that should be available via ext-soap in case the extension is not
+ * enabled in the Psalm-executing environment.
+ *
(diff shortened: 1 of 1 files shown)


stubs/extensions/soap.phpstub
```diff
@@ -1,5 +1,97 @@
 <?php
 
+<?php
```
orklah, line 3: There's a duplicated php tag

braindawg: Whoops! That was careless! I've gotten too used to Psalm catching all my boneheaded mistakes. Thanks!

orklah:
Thanks!

PR #10339, opened 2023-11-02
Allow enum cases to be global constants

### LINKED ISSUE #10334: Uncaught RuntimeException: Failed to infer case value for...
Using the following enum, I get the error `Uncaught RuntimeException: Failed to infer case value for OK in psalm/vendor/vimeo/psalm/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php:1457`
```php
enum UploadedFileError: int
{
     case OK = \UPLOAD_ERR_OK;
}
```

**Note**: If I comment out that line, everything works as expected. I'm not too familiar with psalm, but this might be an easy fix.

Screenshot 2023-10-31 at 2 12 54 AM

robchett:
fixes #10334


orklah:
Thanks!

weirdan:
We probably should backport this to 5.x

PR #10340, opened 2023-11-02
Fix stub for RecursiveArrayIterator::getChildren

### LINKED ISSUE #10327: Incorrect stub for RecursiveIterator, return value of getChildren() should be nullable
The [docs for RecursiveIterator](https://www.php.net/manual/en/class.recursiveiterator.php) and [stubs in PHP source](https://github.com/php/php-src/blob/b6bf7a0faf3f26711d7bd4c192b9aa0e679ad948/ext/spl/spl_iterators.stub.php#L51) specify the return value of `getChildren()` as `?RecursiveIterator`, while psalm complains in case of `null` return value:

https://psalm.dev/r/602a63575d

robchett:
fixes #10327


orklah:
Thanks!

PR #10341, opened 2023-11-02
Report unused issue handler suppressions

### LINKED ISSUE #2369: Error when error suppresions are configured which dont suppres a actual error
Similar to https://github.com/vimeo/psalm/issues/2033 but for the psalm.xml when handlers are defined like e.g. `<PossiblyInvalidOperand errorLevel="info" />` but the codebase-scan does not contain a PossiblyInvalidOperand error psalm should report that one can/should remove the suppression rule (like phpstan does)

robchett:
fixes #2369 

My first time doing something that needs to combine information from multithreaded code so there might be some mistakes there.


orklah:
Cool! Thanks!

PR #10344, opened 2023-11-03
Support @(no-)seal-(properties|methods) annotations without prefix

### LINKED ISSUE #2587: Support seal-methods/seal-properties as shorthand for psalm-seal-methods
Having a syntax that wasn't specific to Psalm would be a better indication that the user wants the comment to be interpreted that way in other IDEs or analyzers.

src/Psalm/Internal/Analyzer/CommentAnalyzer.php would need to be changed.

Phan's latest releases support `@seal-methods`/`@seal-properties`.

robchett:
Fixes #2587


orklah:
Thanks!

PR #10357, opened 2023-11-08
Inherit conditional returns

### LINKED ISSUE #3593: Conditional type 'func_num_args()' not properly inheriting
I was fiddling around with the new conditional type + func_num_args and was suprised the examples didn't work in my case.. I think it has something to do with an added layer of object inheritance in my case..

With some inheritance (broken):  https://psalm.dev/r/684a7f32b1
Flat hierarchy (working):  https://psalm.dev/r/5fd52e3490

robchett:
`PHP_MAJOR_VERSION`/`PHP_VERSION_ID`/`func_num_args()` only applied to the declaring class and not to any inheriting classes

Fixes #3593


orklah:
Thanks!

PR #10361, opened 2023-11-09
Method annotation fixes

### LINKED ISSUE #3871: Would expect `MethodSignatureMismatch`
Probably mis-understanding something, as usual. Apologies if so.
I would expect psalm to give a `MethodSignatureMismatch` error here?
https://psalm.dev/r/c4bfb387bd

### LINKED ISSUE #4546: MethodSignatureMismatch when extending class with variadic @method
If https://psalm.dev/r/935510dd68 is fine, I'd expect https://psalm.dev/r/2c63f8d3b7 to also be fine. Might be an issue with when annotations are realised?

### LINKED ISSUE #5786: @method: return type is taken into account, but parameter types are not
Hitting this on doctrine/orm: https://psalm.dev/r/0e37c0c81a

Notice how Psalm does not complain about the call to `foo()`. Type widening should be detected and OK (there should be no errors)

### LINKED ISSUE #5990: [Enhancement] Emit an error when phpdoc is impossible
https://psalm.dev/r/af5065a4c0

Given that `bar` has a native return type of `string`, shouldn't psalm be able to emit an error that the phpdoc is invalid? I'm thinking that phpdoc should only become the source of truth if there are no native typehints or the typehints are castable to each other..

robchett:
Fixes #5786 
Fixes #3871
Fixes #4546 
Fixes #5990
Also fixes this https://psalm.dev/r/edaea88e00, where an extend with an `@method` silently overwrites the parent return type

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -258,8 +258,6 @@ final class ClassAnalyzer extends ClassLikeAnalyzer
             IssueBuffer::maybeAdd($docblock_issue);
         }
 
-        $classlike_storage_provider = $codebase->classlike_storage_provider;
-
         $parent_fq_class_name = $this->parent_fq_class_name;
 
         if ($class instanceof PhpParser\Node\Stmt\Class_ && $class->extends && $parent_fq_class_name) {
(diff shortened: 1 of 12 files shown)


robchett:
Failing test arrayAssignmentTest::allowsArrayAccessNullOffset is invalid PHP - https://3v4l.org/EomWU

orklah:
Thanks!

danog:
btw @robchett this also breaks magic method calls on abstract classes: https://psalm.dev/r/085bdcc380

Trying to fix in https://github.com/vimeo/psalm/pull/10420/files, will try to make some more progress tomorrow...

Issue #9068, opened 2023-01-06
array_filter doesn't accept callback returning non-scalar


jacekkarczmarczyk:
https://psalm.dev/r/d17cbe92de

```php
array_filter(
  [(object)[], null], 
  fn ($item) => $item,
);
```

results in:

```
ERROR: [InvalidArgument](https://psalm.dev/004) - 3:42 - Argument 2 of array_filter expects callable(mixed, mixed=):scalar|null, but pure-Closure(null|object):(null|object) provided
```

You can easily workaround it, but imho you shouldn't need it
```php
array_filter(
  [(object)[], null], 
  fn ($item) => !is_null($item),
);
```

orklah:
Yeah, seems to work. Can you propose a PR? :)

Issue #8919, opened 2022-12-17
array_filter return type provider wrong when 2nd param is literal string variable


kkmuffme:
https://psalm.dev/r/5e135828ba

Both traces should be the same

Issue #6085, opened 2021-07-13
Invalid array callable wont get detected


boesing:
Hey there,

an array-callable must consist of "object"/"class name" and "method/static method" information. Any other information leads to runtime issues:

https://psalm.dev/r/b661a0cb84
https://3v4l.org/rE02P

If you triage this bug, please let me know how to help.

orklah:
Seems like a bundle of issues here
This is correctly flagged: https://psalm.dev/r/920f2bc3fa
But this is not: https://psalm.dev/r/88142fd37c
Ideally, `$a` in this example https://psalm.dev/r/69df2d109b should become `callable-array{0: Bar::class, 1: "baz"}`

I guess if I had to look at it, I'd start checking the first example and see how Psalm manage to flag the error (and where it comes from with --debug-emitted-issues) and then, I'd try to implement the same check when calling a Function or Method that has a callable parameter

Issue #3047, opened 2020-04-02
array_multisort and InvalidPassByReference


kkmuffme:
array_multisort is super bad documented in PHP itself, but it can actually be used to sort arrays by multiple columns, where the finally sorted array (by reference) is the LAST element (even though the docs say the first arg is passed by ref which isn't true e.g. in my example), causing a InvalidPassByReference in my test case.

https://psalm.dev/r/9d7e5978bf

(Note: in reality I sort multiple columns, at once, so it looks like:

```
array_multisort(
	array_column( $test, 'sort_this' ),
	SORT_DESC,
	array_column( $test, 'id' ),
	SORT_ASC,
        ...,
	$test
);
```

weirdan:
> the finally sorted array (by reference) is the LAST element (even though the docs say the first arg is passed by ref which isn't true e.g. in my example)

It's worse than that: all array arguments are (optionally!) accepted by reference and all of them are sorted, see https://3v4l.org/iZLaY

PR #10370, opened 2023-11-13
Fix misc callable bugs

kkmuffme:
Sorry, this is a quite comprehensive PR, I couldn't break it up though, as earlier commits caused me to fix issues in later commits (which sent me down a rabbit hole...)

- Fix https://github.com/vimeo/psalm/issues/9068 + array_filter callback type for keys
- Fix https://github.com/vimeo/psalm/issues/8919
- Fix https://github.com/vimeo/psalm/issues/10371
- Fix param providers for native methods not working
- Fix array_filter params callable incorrect required params depending on the 3rd argument value (at this commit psalm doesn't report an error, due to https://github.com/vimeo/psalm/issues/8438, which gets fixed by this PR later on though)
- Fix https://github.com/vimeo/psalm/issues/3047 and further improve types for array_multisort and add errors for invalid params
- Fix SimpleTypeInferer failing on bitwise operations with constants
- Fix https://psalm.dev/r/7f112fd745 - MethodComparator only reported an error for this if the parent class was user defined (= not in stubs), which is wrong, since this will cause a fatal error when running the code
- Fix ReturnTypeWillChange false positive https://psalm.dev/r/91c6992bf1 with missing return type signature
- Fix that files passed via CLI but are not in projectFiles reported only a limited, random selection of errors instead of all - treat all files passed as CLI args as if they were in projectFiles
- Fix for classes what https://github.com/vimeo/psalm/pull/8503 fixed for functions (as some issues didn't report for other reasons, which were solved since then)
- Fix https://github.com/vimeo/psalm/issues/6085
- Fix optional args not enforced in callable (fix for non-closure/arrow functions of https://github.com/vimeo/psalm/issues/8438)
- Fix array callables not treated as callable https://psalm.dev/r/23f3787207 (this is needed to fix the optional args enforcement for array callables too)
- Fix potential cache race conditions/cache not deleted with non-lowercase file paths and add missing docs

The MethodComparator change caused tons of failing tests/shepherd errors, because psalm just ignored those for itself until now, e.g. suppressedIssues should be list<string>, but some extending classes used array<int, string>,...

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -371,7 +371,7 @@ return [
 'array_diff_ukey\'1' => ['array', 'array'=>'array', 'rest'=>'array', 'arr3'=>'array', 'arg4'=>'array|callable(mixed,mixed):int', '...rest='=>'array|callable(mixed,mixed):int'],
 'array_fill' => ['array<int,mixed>', 'start_index'=>'int', 'count'=>'int', 'value'=>'mixed'],
 'array_fill_keys' => ['array', 'keys'=>'array', 'value'=>'mixed'],
-'array_filter' => ['array', 'array'=>'array', 'callback='=>'callable(mixed,mixed=):scalar|null', 'mode='=>'int'],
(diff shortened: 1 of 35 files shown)


orklah:
Thanks for the massive work!

PR #10379, opened 2023-11-16
update decimal stubs from documented URL

kkmuffme:


orklah:
It feels weird to have phpstan tags in our internal stubs. Please override the @param with the current @phpstan-param instead

orklah:
Thanks!

PR #10385, opened 2023-11-17
Finding methods and properties for auto completion feature

issidorov:
The method `Codebase::getCompletionItemsForClassishThing()` is important
for finding methods and properties in classes.

I wrote tests for it and fixed some bugs.


orklah:
Cool! Thanks!

PR #10388, opened 2023-11-18
Fix redundant condition mixed type escape hatch & loose comparison int type shouldnt be mixed

kkmuffme:
- fix loose comparisons of int with string shouldn't change change the int to empty-mixed
- revert https://github.com/vimeo/psalm/pull/7663 including previous from_docblock Mixed assignments, as the tests required 2 suppressions and created an escape hatch via mixed on higher psalm error levels, where mixed isn't reported, thus hiding potentially fatal bugs.
It's still possible to run the validation of docblock docs though: a @var declaration that contains both possible types, to ensure later code won't escape any checks (and no @psalm-suppress needed at all)
(this is also a required preparation to fix some isset issues of [#9759](https://github.com/vimeo/psalm/issues/9759)

### DIFF
--- a/src/Psalm/Internal/Type/NegatedAssertionReconciler.php
+++ b/src/Psalm/Internal/Type/NegatedAssertionReconciler.php
@@ -304,9 +304,7 @@ final class NegatedAssertionReconciler extends Reconciler
 
             $failed_reconciliation = Reconciler::RECONCILIATION_EMPTY;
 
-            return $existing_var_type->from_docblock
-                ? Type::getMixed()
-                : Type::getNever();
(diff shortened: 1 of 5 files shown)


orklah:
returning never instead of mixed is a pretty radical change from the current behaviour. I'd be fine with this for Psalm 6, what do you think about rebasing on master for this?

orklah:
nah nevermind then. Can you still rebase on 5.x? Now that the phar is fixed, I'd like to check if it works when analyzing phpunit

orklah:
Thanks!

PR #10389, opened 2023-11-18
fix CI "split" unevenly splits the number of tests

kkmuffme:
"split" unevenly splits the number of tests in a non-optimal way, which leads to:
1) some batches smaller/larger than necessary, which causes tests overall to be slower than necessary, since larger batches delay everything in the end
2) some tests run in multiple batches, which makes things slower and is confusing

This splits the batches into equal number of tests, where the first batches are by 1 element larger than later elements, if the total number of tests is not divisible by the number of total chunks.


orklah:
Thanks!

Issue #10390, opened 2023-11-19
CircleCI phar-build job is broken


edsrzf:
The CircleCI `phar-build` job is broken. ([Example failure](https://app.circleci.com/pipelines/github/vimeo/psalm/12517/workflows/52aab2ee-5aca-4965-aee3-981a65ffe1d0/jobs/41178)) According to https://github.com/vimeo/psalm/discussions/10363, this blocks releases.

I've been able to reproduce the failure locally with the following steps:

```sh
docker run -it --rm thecodingmachine/php:7.4-v4-cli /bin/bash
# Commands from this point are run inside the Docker container
$ git clone https://github.com/vimeo/psalm.git
$ cd psalm
$ composer install
$ bin/build-phar.sh
$ build/psalm.phar --version
```

This produces the error message:

```
PHP Fatal error:  Cannot declare class ComposerAutoloaderInit9e5dd54562a8c1ec1f3c2855414302bf, because the name is already in use in /usr/src/app/psalm/vendor/composer/autoload_real.php on line 5
Fatal error: Cannot declare class ComposerAutoloaderInit9e5dd54562a8c1ec1f3c2855414302bf, because the name is already in use in /usr/src/app/psalm/vendor/composer/autoload_real.php on line 5
```

From a bit of debugging, the `ComposerAutoloaderInit<hash>` class has the same class name in both the phar and the `vendor` directory. Both these autoloaders are included when `psalm.phar` runs, which is what causes the error. I'm not sure whether both autoloaders are supposed to be included but are expected to have different names, or whether only one should be included.

If I change into the `build` directory first, everything works:

```sh
$ cd build
$ ./psalm.phar --version
Psalm 5.x-dev@577808b12da647dd87abb435447dc6429361afdb
```

I don't know what changed or what the proper fix is, but I'm hoping that this information might be helpful and that this issue can serve as a place to discuss the problem.

tscni:
Looks like this was caused by Composer 2.6.4 making the autoloader generation reproducible (https://github.com/composer/composer/pull/11663)

PR #10391, opened 2023-11-21
Fix Phar build failure

tscni:
The issue was likely caused by Composer 2.6.4 making the autoloader generation (more) reproducible (composer/composer#11663)

We can either try to change the generated autoloader with an autoloader suffix, or just change the Psalm root directory for the smoke test. The latter approach seems easier. :P

Fixes https://github.com/vimeo/psalm/issues/10390


orklah:
Really cool! Thanks to you and @edsrzf for checking this oui

PR #10396, opened 2023-11-22
Fix circleci builds

danog:
Should remove the last blocker for a 5.x release @orklah


danog:
@orklah that patches the PSL repro repo since I don't have push access for it (I already fixed up the phpunit repro repo)

orklah:
Cool :)

Any idea why we don't have the jobs for circleCI appearing?

orklah:
That may mean it will fix itself once merged, let's try

PR #10402, opened 2023-11-23
Update MongoDB stubs for 1.17

alcaeus:
The parameter name for `Serializable::unserialize` has changed from `$serialized` to `$data` with the introduction of stub files in PHP 8.0. For versions before 8.0, the extension uses `$serialized` as parameter name. I'm not sure if this is worth adding to one of the call map delta files, please let me know if so.


GromNaN (approved):

orklah:
Thanks!

PR #10406, opened 2023-11-23
Fix POSIX only detection of absolute paths in config

rarila:
With #10370 a bug in ``FileFilter`` was revealed where the check for absolute paths works only on POSIX filesystems.

You'll end up with paths like ``C:\basedir\C:\basedir\path\to\source`` instead of ``C:\basedir\path\to\source`` on Windows which then triggers ``Could not resolve config path to …``.

This is just a fast emergency fix to get Psalm with passing a source dir running on Windows again.

I guess in the long run all incoming paths should be run through ``Path::normalize()`` to prevent other possible problems on Windows like

https://github.com/vimeo/psalm/blob/2897ba636551a8cb61601cc26f6ccfbba6c36591/src/Psalm/Config/FileFilter.php#L138-L143

where only the POSIX directory separator is considered. But would take quite some time and go into another PR.


orklah:
Hey! If I understood correctly, without thix fix, last release is crashing, right?

If it is, you should target 5.x branch instead so this can be released soon

rarila:
@orklah Not crashing. It just stops with that error message. And that only when you supply a source path on the commandline. So using the path from ``psalm.xml`` file like

```"C:/php.exe" ./vendor/bin/psalm```

works, but specifiying it on the commandline like

```"C:/php.exe" ./vendor/bin/psalm .\src```

fails with

> Problem parsing C:\foo\bar\psalm.xml:
>   Could not resolve config path to C:\foo\bar\\C:\foo\bar\src

rarila:
I guess that didn’t work :-)

rarila:
@orklah Created a new one (#10441) closing this one

PR #10409, opened 2023-11-26
Fix backtick analysis

danog:
The existing logic treated each component of the encapsed string as a separate argument to shell_exec, which is wrong.


danog:
Btw note that I basically transform this to a shell_exec call because that's what PHP itself does: https://github.com/php/php-src/blob/master/Zend/zend_compile.c#L9675

orklah:
Thanks!

PR #10412, opened 2023-11-26
Rename ResourceInputStream

hexmode:


danog:
ReadableResourceStream is an amp v3 class, 5.x is still using amp v2, so the existing code is correct.
Psalm v6 will use amp v3, and migration was already done there.

PR #10413, opened 2023-11-26
Switch to ReadableResourceStream

### LINKED ISSUE #10411: psalm lsp implementation uses Amp\ByteStream\ResourceInputStream
But this was renamed [two years ago](https://github.com/amphp/byte-stream/commit/b7f5920f16b4d0b69ee74a0e16f1b3a7ef0e1a90).

hexmode:
Fixes #10411


danog:
ReadableResourceStream is an amp v2 class, 5.x is still using amp v2, so the existing code is correct.
Psalm v6 will use amp v3, and migration was already done there.

PR #10422, opened 2023-11-29
Restore support for null coalesce on match expressions

tscni:
https://github.com/vimeo/psalm/pull/10068 added `isset()` restrictions that didn't consider null coalesces on match expressions.
This restores that support by converting the match expression to a virtual variable for the isset analysis, similar to other incompatible expressions.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/CoalesceAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/CoalesceAnalyzer.php
@@ -39,6 +39,7 @@ final class CoalesceAnalyzer
             || $root_expr instanceof PhpParser\Node\Expr\MethodCall
             || $root_expr instanceof PhpParser\Node\Expr\StaticCall
             || $root_expr instanceof PhpParser\Node\Expr\Cast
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/CoalesceAnalyzer.php
```diff
@@ -39,6 +39,7 @@ public static function analyze(
             || $root_expr instanceof PhpParser\Node\Expr\MethodCall
             || $root_expr instanceof PhpParser\Node\Expr\StaticCall
             || $root_expr instanceof PhpParser\Node\Expr\Cast
+            || $root_expr instanceof PhpParser\Node\Expr\Match_
```
tscni, line 42: Based on https://github.com/vimeo/psalm/pull/10068#issuecomment-1660213877, `ClassConstFetch` and `AssignRef` are currently supported in the isset analyzer for exactly the same reasons:
https://github.com/vimeo/psalm/blob/ec23f998ea38413cdea8e6c14644179b26d407b2/src/Psalm/Internal/Analyzer/Statements/Expression/IssetAnalyzer.php#L72-L73

I'm wondering whether they should be move here too.

orklah:
Thanks!

PR #10424, opened 2023-11-29
use branch name to randomize tests to ensure test dependencies will error

kkmuffme:
Tests were run randomized in the same order in all cases, which could hide dependency issues between tests.
Changed so tests will randomize using the PR branch name instead of a static string. Like that the tests will stay in the same ordering for a given PR, but each PR has them ordered in another, random way, ensuring any problematic test dependencies will at least report eventually with subsequent PRs.


orklah:
Thanks!

PR #10426, opened 2023-11-29
Fix at in array keys

danog:
@robchett Kind request, please be more careful and consider more edge cases when fixing stuff, I still haven't fully figured out a nice fix for https://github.com/vimeo/psalm/pull/10420 :)


orklah:
Thanks!

PR #10429, opened 2023-11-29
Fix static magic method pureness not being inherited from traits

tscni:
https://github.com/vimeo/psalm/pull/10385 "broke" this by propagating pseudo static methods from traits to using classes.
`AtomicStaticCallAnalyzer` was then not capable of dealing with this, because now these static pseudo methods actually exist.

As long as the methods from traits aren't actually transferred to the using class, it seems right that the logic in `AtomicStaticCallAnalyzer` uses `::getDeclaringMethodId()` instead of `::getAppearingMethodId()` for this purpose.


orklah:
Thanks!

PR #10431, opened 2023-11-30
Emit UnusedPsalmSuppress issues for suppressed issues already removed

danog:
...by plugins.


danog:
@orklah Should be good to merge now, changed the value of the findUnusedBaselineEntry config value until we drop 7.4 support in v6 due to the use of `mixed` typehints in symfony.

orklah:
Thanks!

PR #10435, opened 2023-12-01
Fix array_key_exists negation

danog:


danog:
Merged into https://github.com/vimeo/psalm/pull/10439

PR #10441, opened 2023-12-01
Fix POSIX only detection of absolute paths

rarila:
Replacing #10406, just targeting 5.x branch.

### DIFF
--- a/src/Psalm/Config/FileFilter.php
+++ b/src/Psalm/Config/FileFilter.php
@@ -7,6 +7,7 @@ use Psalm\Exception\ConfigException;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use SimpleXMLElement;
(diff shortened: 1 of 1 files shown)


rarila:
@kkmuffme I would prefer to get that in as fast as possible as emergency fix.

For the other cases of the same code, fixing that alone won’t help, as there seems to be other problems like [this](https://github.com/vimeo/psalm/blob/2897ba636551a8cb61601cc26f6ccfbba6c36591/src/Psalm/Config/FileFilter.php#L138-L143). I’m sure, there are more of this.

I guess, as nobody complained sofar, the other cases are not triggered that often. But this one probably is hit more often.

I’ll start a new branch solving all those problems as soon as possible, but I guess it takes some time (xmas is soon;-)

orklah:
Thanks!

Issue #10418, opened 2023-11-28
Could not resolve config path


M393:
On Windows 10 after upgrading from 5.15.0 to 5.16.0 I get this error:
>[INFO  - 10:52:14 AM] Starting language server
>[INFO  - 10:52:14 AM] Psalm Version was detected as 5.16.0
>[ERROR  - 10:52:15 AM] Problem parsing C:\test\psalm.xml:
>  Could not resolve config path to C:\test\\C:\test

`C:\test` is the project location, seems like the path is concatenated to itself.

Tried to remove everything from my config file, still same error:
```xml
<?xml version="1.0"?>
<psalm
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="https://getpsalm.org/schema/config"
    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
>
</psalm>
```

M393:
I'm using the vscode extension (v2.7.0). The log above is from the Output panel.

theodorejb:
I'm seeing this error notification pop up with Psalm 5.17 every time I open any PHP file in PhpStorm. For example:

> psalm: Problem parsing C:\Users\TheodoreBrown\Documents\code\main\psalm.xml:  
> Could not resolve config path to C:\Users\TheodoreBrown\Documents\code\main\\C:\Users\TheodoreBrown\AppData\Local\Temp\Psalmtemp_folder1276\php\utilities\Page.php

M393:
I updated to 5.17 just now and no longer have this issue.

Try to find out how with which parameters psalm is started. Install the last working version and get the command line arguments:
https://superuser.com/a/981599

theodorejb:
The error occurs with both 5.16 and 5.17, but not with 5.15. According to Process Monitor the following command is being run:
```
C:\Windows\system32\cmd.exe /c C:\Users\TheodoreBrown\Documents\code\main\vendor\bin\psalm.bat --output-format=checkstyle -c C:/Users/TheodoreBrown/Documents/code/main/psalm.xml --monochrome C:/Users/TheodoreBrown/AppData/Local/Temp/Psalmtemp_folder2266/php/utilities/Page.php
```

theodorejb:
@kkmuffme I get the same error for both forward slash and backslash directory separators.

theodorejb:
I found the issue and opened #10452 to fix it.

PR #10442, opened 2023-12-02
Fix CLI -r error

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/10418

This was a bug that existed for a long time but only manifested itself visibly due to changes in https://github.com/vimeo/psalm/pull/10370/commits/92bcc5cfe03396cee0a8d31e8604a93dceec1bcf


orklah:
Thanks!

PR #10443, opened 2023-12-02
Fix return type of DOMXPath::query

nielsdos:
This can also return namespace nodes, which are not a child class of DOMNode.

### DIFF
--- a/stubs/extensions/dom.phpstub
+++ b/stubs/extensions/dom.phpstub
@@ -975,7 +975,7 @@ class DOMXPath
     public function evaluate(string $expression, ?DOMNode $contextNode = null, bool $registerNodeNS = true): mixed {}
 
     /**
-     * @return DOMNodeList<DOMNode>|false
+     * @return DOMNodeList<DOMNode|DOMNameSpaceNode>|false
      */
     public function query(string $expression, ?DOMNode $contextNode = null, bool $registerNodeNS = true): mixed {}
 


orklah:
Thanks!

veewee:
@nielsdos Since you are doing a lot of work in DOM. Can you tell me if there is any specific reason why the namespace node is not a DOMNode?
I tried patching it a while back, but it seemed to scare people:

Reference : https://github.com/php/php-src/pull/6939

nielsdos:
> @nielsdos Since you are doing a lot of work in DOM. Can you tell me if there is any specific reason why the namespace node is not a DOMNode? I tried patching it a while back, but it seemed to scare people:
> 
> Reference : [php/php-src#6939](https://github.com/php/php-src/pull/6939)

I'll reply here for now but let's not disturb this place here too much. You can contact me at my nielsdos at php dot net email address if you want to discuss more in detail.

I wasn't there when the decision was made 20 years ago. This is something that I would like to do though and I think it's possible to do. It's just not entirely clear what consequences this could have for the ecosystem (extensions+existing code).
But here's what I know:
- The reason there is a separate class to begin with is because the underlying data structure in C for namespaces is completely different. This isn't much of a problem nowadays though as I rewrote a part of this code to fix crash bugs.
- There are places where we should not allow DOMNameSpaceNodes (e.g. append etc). Now those functions take a DOMNode so we have to add restrictions at the right places.
- Other extensions (both 1st and 3rd party) can rely on the fact that namespaces are not passed in.

staabm:
I tried adding this change into phpstan, which sparked some interessing questions:

https://github.com/phpstan/phpstan-src/pull/2842#issuecomment-1872509254

@nielsdos do you have concrete "real world" code examples for which this change is necessary/useful?

nielsdos:
> I tried adding this change into phpstan, which sparked some interessing questions:
> 
> [phpstan/phpstan-src#2842 (comment)](https://github.com/phpstan/phpstan-src/pull/2842#issuecomment-1872509254)
> 
> @nielsdos do you have concrete "real world" code examples for which this change is necessary/useful?

https://3v4l.org/GpOJA

PR #10452, opened 2023-12-06
Fix remaining POSIX-only absolute path detection

theodorejb:
These were missed in #10441. Fixes "Could not resolve config path" error on Windows (#10418).

Hopefully this can be released soon since Psalm 5.17 always crashes in PhpStorm on Windows.


orklah:
Thanks!

PR #10454, opened 2023-12-07
fix false positive ArgumentTypeCoercion for callback param

kkmuffme:
Fix https://psalm.dev/r/dca256db4b to not report ArgumentTypeCoercion


orklah:
Thanks!

PR #10456, opened 2023-12-07
update fidry/cpu-core-counter dependency

kkmuffme:
chore: the new version has no changes: https://github.com/theofidry/cpu-core-counter/releases/tag/1.0.0


orklah:
Thanks!

PR #10463, opened 2023-12-07
fix composer scripts running with inconsistent php versions

kkmuffme:
If composer is invoked with a specific php version, some of the scripts will run with that while others will use the default.

Fix this to ensure it will always use the php install used to invoke the script


orklah:
Thanks!

PR #10477, opened 2023-12-12
Fix type not equal when parent

kkmuffme:
parent nodes are only populated if taint/unused variable analysis is enabled

Partial, temporary fix for https://github.com/vimeo/psalm/issues/10478


orklah:
Thanks!

PR #10480, opened 2023-12-12
Use same parameter names in stubs

### LINKED ISSUE #10479: Extending `ArrayObject` and overwriting `offsetSet` leads to unsolvable ParamNameMismatch error
PHP: `8.2.13`
Psalm: `5.16.0`/`5.17.0`

We have a class which extends `ArrayObject` and overwrites the `offsetSet` method.
The method has the two arguments `$key` and `$value` as defined in `ArrayObject`. `ArrayObject` itself implements `ArrayAccess` where the method signature is different already.

We now have three errors (since Psalm 5.16.0):

- `Argument 1 of Our\Class::offsetSet has wrong name $key, expecting $offset as defined by ArrayAccess::offsetSet (see https://psalm.dev/230)`
- `Argument 1 of Our\Class::offsetSet has wrong name $key, expecting $index as defined by ArrayObject::offsetSet (see https://psalm.dev/230)`
- `Argument 2 of Our\Class::offsetSet has wrong name $value, expecting $newval as defined by ArrayObject::offsetSet (see https://psalm.dev/230)`

Error 1 and 2 contradict each other. And error 3 makes no sense (imho) because the argument is called `$value` in both `ArrayObject` and `ArrayAccess`.

The errors exist since version 5.16.0. Downgrading to 5.15.0 "solves" the problem.

danog:
Fixes #10479

### DIFF
--- a/stubs/CoreGenericClasses.phpstub
+++ b/stubs/CoreGenericClasses.phpstub
@@ -158,46 +158,46 @@ class ArrayObject implements IteratorAggregate, ArrayAccess, Serializable, Count
      * Returns whether the requested index exists
      * @link http://php.net/manual/en/arrayobject.offsetexists.php
      *
-     * @param TKey $index The index being checked.
+     * @param TKey $offset The index being checked.
      * @return bool true if the requested index exists, otherwise false
      *
      * @since 5.0.0
      */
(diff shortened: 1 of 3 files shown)


orklah:
Thanks!

Issue #10465, opened 2023-12-07
CLI: Arguments checked as file paths


gneuvill:
Hi,

New to psalm, I successfully installed it through composer, generated a config file and run it. But as I tried to enable some plugins (previously installed with composer also), I consistently got this abstruse response : `Cannot locate enable`.

I traced the error using echo statements up to the call to `CliUtils::getPathsToCheck(null)` in `Psalm/Config.php l. 1305`. Once I commented out that call, the 'enable' command ran to completion without errors. (so 92bcc5c looks like the culprit and the "enable" argument was apparently treated as a file path)

I am on linux, using php 8.2.12, composer 2.6.5, psalm 5.17.

Hope that will help.

florisluiten:
I've got the same problem.

Console output:

```bash

$ composer require --dev vimeo/psalm:5.17.0
[...snip...]

$ composer require --dev psalm/plugin-symfony
[...snip...]

$ ./vendor/bin/psalm-plugin enable
  Not enough arguments (missing: "pluginName").
enable [-c|--config CONFIG] [--] <pluginName>

$ ./vendor/bin/psalm-plugin enable psalm/plugin-symfony
Cannot locate enable
```

gneuvill:
@kkmuffme 

> Can you post the exact command you used to run psalm please?

`composer require --dev php-standard-library/psalm-plugin`
`vendor/bin/psalm-plugin enable php-standard-library/psalm-plugin`

florisluiten:
Yes, although it now complains about $paths_to_check not being defined ;)

```
$ ./vendor/bin/psalm-plugin enable psalm/plugin-symfony
PHP Warning:  Undefined variable $paths_to_check in vendor/vimeo/psalm/src/Psalm/Config.php on line 1312

[OK] Plugin enabled
```

The `psalm.xml` file is updated correctly :)

florisluiten:
> I thought it's obvious that above the "if" you added, you need to add `$paths_to_check = null;` that's why I didn't mention that

Yes it was, I was joking. Sorry :-)

commonpike:
Exactly the same error, psalm 5.18, symfony 5.4.15

Even calling `vendor/bin/psalm-plugin show` returns `Cannot locate show`

https://github.com/vimeo/psalm/issues/10465#issuecomment-1853815743 fixed it.

Never used psalm, and the first thing I have to do is hack it to get it working :-D

PR #10483, opened 2023-12-13
Fix GH-10465

florisluiten:
Fixes the "Cannot locate enable" error when using psalm-plugin enable with a Symfony project

Also see https://github.com/vimeo/psalm/issues/10465


orklah:
Thanks!

Issue #9552, opened 2023-03-22
ParadoxicalCondition for no reason?


kkmuffme:
Simplified code snippet https://psalm.dev/r/268b76bdec

No idea why/where it asserts that $unit cannot be "lb"?
=> when I comment the "continue;" the error is gone
=> this is ONLY with "lb". When I put "lbs" or anything else the error is gone too

This error is new in v5 (checked with v4 and it's not there)

weirdan:
>  => this is ONLY with "lb". When I put "lbs" or anything else the error is gone too

Looks like it happens with the value the last element of that array has. If I put `lbs` last, it starts happening with `lbs` but emits not issues with `lb`: https://psalm.dev/r/7ebfb98ceb

PR #10484, opened 2023-12-13
fix literal int/string comparisons only using one literal

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/9552
Fix https://psalm.dev/r/aada187f50 where 2 union types are not intersected and the condition contains both types

Remove incorrect/broken test, see https://github.com/vimeo/psalm/issues/10487 - this PR improves the types of this test (now 3 of 4 are correct but adds an error, compared to 1 of 4 types correct and no error before the PR), however fixing this is a separate, probably quite extensive PR.


orklah:
Thanks!

Issue #9267, opened 2023-02-11
Bug: missing InvalidArgument error when passing false to true param


kkmuffme:
https://psalm.dev/r/2c91682853

When passing `false` to the param, it should report as `InvalidArgument` just like it does for when we pass `123`

orklah:
Interesting, this is caused by this line: https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php#L864

the param "ignore_false" is set to true by default on this check. It's a little surprising because we have other mechanisms to ignore false for internal functions so I don't get why it's put to true here.

Maybe try to push a PR putting that to false and let's see where the CI goes?

orklah:
Yeah, that's the other mechanism I was talking about earlier.

Could you try to change this line:
https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php#L79
into
`            if ($input_type_part instanceof TFalse && ($ignore_false || $input_type->ignore_falsable_issues)) {`

When shepherd runs, it is supposed to ignore falsable from internal functions

orklah:
Yep, that was not the correct fix.

Can you try 
```
$input_type->ignore_nullable_issues,
$input_type->ignore_falsable_issues,
```

instead of 
```
true,
true
```
on the call here: https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php#L864
(and reverting the rest)

Apparently, it's done like that on other calls. It seems to fix your case and another test that failed on the previous fixes, let's see if the rest of CI pass

orklah:
Yeah, looks much better IMHO.

I think there's some cases where echo doesn't accept null that would need to get changed to reduce obstacles, but the rest seems legit at first glance

orklah:
Oh, there's something wrong though: https://github.com/vimeo/psalm/actions/runs/4155594465/jobs/7188799901#step:5:41
This is reported as `PossiblyInvalidArgument` but it should be a `PossiblyNullArgument` (as it seems it was before)

orklah:
No worries, seems like this bug is very old, it can wait!

PR #10488, opened 2023-12-13
add InvalidArgument error when passing false to true param

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/9267


orklah:
Thanks!

PR #10489, opened 2023-12-14
Fix Uncaught RuntimeException: PHP Error: Uninitialized string offset 0 when $pattern is empty

iMu3ic:
When $pattern in src/Psalm/Internal/Analyzer/Statements/Expression/Call/FunctionCallReturnTypeFetcher.php:638 was empty there were Uncaught RuntimeException: PHP Error: Uninitialized string offset 0. Added length check with result trimming.


orklah:
Thanks!

PR #8427, opened 2022-08-22
added SensitiveParameter, AllowDynamicProperties php 8.2 attributes

staabm:
inspired by a similar commit https://github.com/vimeo/psalm/commit/b7e70f3cd797402ce0a7c56048f80a1f8bcf140f

beware: I am not a psalm user and did not contribute a lot yet, so expect errors ;)

closes https://github.com/vimeo/psalm/issues/8262

PR #10493, opened 2023-12-15
Add support for Override attribute

### LINKED ISSUE #10404: InvalidAttribute for #[\Override] in PHP 8.3
```
ERROR: InvalidAttribute
at xxx.php:33:7
The class Override doesn't have the Attribute attribute (see https://psalm.dev/242)
    #[\Override]
```

delolmo:
Inspired by #8427.

This is my first contribution to psalm, so there may be some errors.

Closes #10404.

### DIFF
--- a/tests/AttributeTest.php
+++ b/tests/AttributeTest.php
@@ -293,6 +293,22 @@ class AttributeTest extends TestCase
                 'ignored_issues' => [],
                 'php_version' => '8.2',
             ],
+            'override' => [
+                'code' => '<?php
+
+                    namespace OverrideAttribute;
+
+                    use Override;
+
+                    class HelloWorld {
+                        #[Override]
+                        public function __invoke() {}
+                    }
+                ',
+                'assertions' => [],
+                'ignored_issues' => [],
+                'php_version' => '8.3',
+            ],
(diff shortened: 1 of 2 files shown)


vudaltsov (approved):

orklah:
Thanks!

tests/AttributeTest.php
```diff
@@ -293,6 +293,22 @@ class Foo
...
+                'code' => '<?php
+
+                    namespace OverrideAttribute;
+
+                    use Override;
+
+                    class HelloWorld {
+                        #[Override]
+                        public function __invoke() {}
+                    }
```
tscni, line 306: Fyi, this is code that should actually error, because there's no parent method to override.

delolmo: You are completely right. However, this particular piece of code would throw a PHP fatal error (i.e. `HelloWorld::__invoke() has #[\Override] attribute, but no matching parent method exists`) - not a psalm error per se. The PR, as is, creates the `#[Override]` core attribute so that psalm does not throw an `InvalidAttribute` error. That being said, I am happy to fix this if the maintainers agree on doing so - I definitely think it is a good idea.

Issue #9760, opened 2023-05-08
filter_input types too generic


kkmuffme:
https://psalm.dev/r/10b253ba58

`filter_input` types are extremely generic and should be updated so they work like the superglobals (`$_SERVER`,...) already do.
e.g. it should be `false|null|string` here

In this case https://psalm.dev/r/5b0fa23243 it should just be `string`

orklah:
that would make sense :)

PR #10498, opened 2023-12-17
filter_input & filter_var return type more specific

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/9760

I dragged this along for ages (at least 5 months) and it's time to finally PR it now.

Since there are 1000s of combinations, I only handled the most commonly used basic ones.

This is just a starting point and can be easily improved with additional types once errors/issues are reported.


orklah:
Thanks!

Issue #8075, opened 2022-06-08
`literal-string` becomes `non-empty-string` after 1000 chars


tdtm:
tl;dr: looks like after 1000 chars and over, a `literal-string`suddenly starts being considered a `non-empty-string`.

Fails with a 1000-char literal
https://psalm.dev/r/05f9492c43

Same exact thing, but with 1 fewer char passes
https://psalm.dev/r/1059fb0c2f

---

Found this adding `literal-string` to methods taking SQL strings. Some long/complex queries were triggering warnings of [ArgumentTypeCoercion](https://psalm.dev/193) despite being literals. The only apparent trigger was when the query's length.

weirdan:
There is an undocumented config attribute called `maxStringLength`: 

https://github.com/vimeo/psalm/blob/4e59398f7705b77d48aa883da1093029ef004315/src/Psalm/Config.php#L1085-L1088

https://github.com/vimeo/psalm/blob/4e59398f7705b77d48aa883da1093029ef004315/src/Psalm/Type.php#L247-L252

You may want to tweak it.

tdtm:
That does it, wonderful. Thank you!

AndrolGenhald:
> https://github.com/vimeo/psalm/blob/4e59398f7705b77d48aa883da1093029ef004315/src/Psalm/Type.php#L247-L252

I assume that's from before `TNonspecificLiteralString` was added, shouldn't it be changed to that instead of `TNonEmptyString`? That should fix the issue was well without having to change the config.

orklah:
> shouldn't it be changed to that instead of TNonEmptyString? That should fix the issue was well without having to change the config. Or maybe people are passing long literals as non-empty-string as well and we should wait until we can do literal-string&non-empty-string?

As usual when we have to juggle between types, we're confronted to this choice. You can design a psalm.dev example that will fail for any choice you'll take.
`literal-string&non-empty-string` would solve that, but Psalm's type system is not designed for that.

Maybe a refactoring of TString could allow this and keep flags for literalness, non-emptyness, lowercaseness, non-falsyness etc... but it's a monstrous task

AndrolGenhald:
> Maybe a refactoring of TString could allow this and keep flags for literalness, non-emptyness, lowercaseness, non-falsyness etc... but it's a monstrous task

I'd much rather just get around to finishing my type system rewrite so that we _can_ do `literal-string&non-empty-string`.

boesing:
[...]

PR #10499, opened 2023-12-17
Fix empty literal string becomes non-empty-string

kkmuffme:
- when max literal string length is 0, which means literal strings are disabled, an empty literal string suddenly became a non-empty-string
- change default fallback type to non-falsy-string, unless it's 0; see https://github.com/vimeo/psalm/issues/8075
-- this also fixes concat of literal non-falsy-string with generic string to stay non-falsy-string instead of becoming a non-empty-string https://psalm.dev/r/323e33ae8a


orklah:
Thanks!

Issue #10501, opened 2023-12-18
Internal: non-strict checking leads to inconsistent behavior


kkmuffme:
https://psalm.dev/r/30076adf91

There are phpcs rules that disallow this already, however it might be too intrusive, bc it will disallow this in all cases, instead of only when necessary (e.g. `Squiz.Operators.ComparisonOperatorUsage.ImplicitFalse` and `Squiz.Operators.ComparisonOperatorUsage.ImplicitTrue` or `Squiz.Operators.ComparisonOperatorUsage`)

There should be a psalm check for these, where a union of falsy types is flagged.

This is an internal issue (e.g. https://github.com/vimeo/psalm/pull/10500 was only necessary bc of the non-strict checking) but I probably also affects psalm users's codebases.

PR #10502, opened 2023-12-18
report error for non-strict or empty comparison on truthy+falsy union

kkmuffme:
- Fix https://github.com/vimeo/psalm/issues/10501 => when a type can contain types that can be truthy AND falsy, it will require strict comparison
- fix it for empty()
- fix empty always returned bool even in cases where it can only be true/false


orklah:
I have no issue with baselining real bugs in Psalm itself.

Though it seems tests are failing a lot too and this is more an issue...

src/Psalm/Issue/RiskyTruthyFalsyComparison.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Psalm\Issue;
+
+use Psalm\CodeLocation;
+
+final class RiskyTruthyFalsyComparison extends CodeIssue
+{
+    public const ERROR_LEVEL = 4;
```
orklah, line 9: I think the level should be adjusted to something like 2. It may find bug but it's pretty opiniated IMHO

orklah:
You can update the baseline on your side by just running psalm with --update-baseline. Other than that and my comment above, seems good to merge :)

danog:
@kkmuffme Don't worry about that, I just did that, thanks for the PR!

veewee:
Just wondering : not all of these checks should be considered Risky. For example:

```php
/**
 * Returns canonicalized absolute pathname.
 * The resulting path will have no symbolic links, '/./' or '/../' components.
 *
 * @return non-empty-string|null
 */
function canonicalize(string $path): ?string
{
    return realpath($path) ?: null;
}
```

This is a nice shortcut which is not risky at all.
If you want to avoid this, it should be rewritten to something like:

```php
function canonicalize(string $path): ?string
{
    $path = realpath($path);

    return false === $path || '' === $path  ? null : $path;
}
```

which is rather verbose.

Does it make sense to be very strict about Risky checks by default or are there acceptable scenarios ?

danog:
Hmm, @kkmuffme might be actually worth to add another separate issue for RiskyTruthyFalsyElvisOperator...

veewee:
Psalm is smart enough to narrow down types in situations like these, so I'm using it as a shortcut to do so in a lot of places. It's giving me quite some errors which I wouldn't consider harmfull but rather expected behaviour.
I'm not sure if dealing with elvis is enough, cause it's also valid when doing ternary checks or even in if-based validations.

But mostly I'm not sure if doing strict comparison over taking the educated shortcut is adding a lot of value.
So for now I'm gonna opt-out on this feature.

theodorejb:
I'm surprised to see a change like this introduced in a minor release. It is a very common pattern to use checks like:

```php
if (!$possiblyEmptyStringOrNull) {
    throw new Exception("Missing required value");
}

// or
if ($intFlagOrNull) {
    doThingSinceOptionIsEnabled();
}
```

I also don't see these as risky - it's using the language as designed to easily handle `null` values the same way as an empty string or `0` integer.

theodorejb:
> could you clarify why?

Because these code patterns are very common and typically used intentionally to simplify code.

Now instead of `if (!$possiblyEmptyStringOrNull)` I have to write `if ($possiblyEmptyStringOrNull !== null && $possiblyEmptyStringOrNull !== '' && $possiblyEmptyStringOrNull !== '0')` for the equivalent check. It makes the code far more verbose.

danog:
All in all, this new check is quite strict and is bound to affect a lot of codebases, but IMO, it's something that can catch potential bugs, which is always a good thing, and the reason why I merged this.

I had some doubts on whether it should be on level 1 instead of level 2, maybe it's better to lower its level...

theodorejb:
> Does your codebase allow `if ($possiblyEmptyStringOrNull == false)` ?

At least in my mind that doesn't express the same intent (even if it does the same thing). It's explicitly comparing to a type not contained in the expression, vs. using the `!` operator to check for falsey values.

theodorejb:
> The impetus is a bug in psalm itself that was caused by one of these unstrict comparisons: #10500

It looks like in this case the bug was with `!$arrayOrNull` checks, which incorrectly handled an empty array the same as `null`.

Perhaps the new check could be limited to array comparisons like this? That does seem far more risky than when checking for falsy strings/integers.

bwoebi:
I would love to see separate checks for if (something-which-can-be-string) and if(something-never-string).
From my experience I've never had issues with array|null or bool|null or such, but only with strings, where "0" was also included, and `if ("" != $str)` instead `if ($str)` would be the right choice. For explicit checking for "0", "" and false/null there's then `if ($str == true)`.

aguerre:
I'm confused with this new report on empty() function, as it is usually used to check if it's null or "empty scalar value" (like 0 or ''). There is no risky comparison in it. 
It's too sad that i had to deactivate this rule because of empty() checks warnings even though it is useful for other cases.
Can you imagine to split this rule in two ?

aguerre:
Yes, check this : https://psalm.dev/r/f95fd49c39

aguerre:
This is a french example where all fields are required, sorry for the lack of precision.
Here we WANT to check if null or empty string. The user shall not pass if there is no address set (untouched field or empty string post). I don't want to differentiate at this point.

stronk7:
I'm afraid, but this is the first time that I'm going to need to disable something because the solution doesn't make any sense to me.

Sure that it has been introduced for a good reason, but it destroys any attempt to use elvis, empty() and simple conditions when you're sure that the behaviour is the same for null/false/empty string (aka any falsy).

I honestly cannot see any advantage having to switch to the long alternative in a lot of cases. Said with all respects!

https://psalm.dev/r/78ca579efd

Ciao :-)

PR #10504, opened 2023-12-19
Merge 5.x

danog:


danog:
@robchett Heads up, I've reverted a part of https://github.com/vimeo/psalm/pull/10361 (https://github.com/vimeo/psalm/pull/10504/commits/4dd06f72964eb9098df92356895673d9f836a7d7) because it breaks purity analysis.

Your implementation isn't exactly ideal, there's a lot of custom logic based on magic methods that breaks if they're suddenly converted to concrete methods.

Issue #4899, opened 2020-12-28
Uncaught Exception: $storage should not be null


danog:
Getting the following exception while scanning a huge project with multiple threads enabled.

```
Uncaught Exception: $storage should not be null for Module\Comment\Orm\MarkFakeMapper::getkey
Stack trace in the forked worker:
#0 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ClassAnalyzer.php(2038): Psalm\Internal\Codebase\Methods->getStorage()
#1 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ClassAnalyzer.php(2009): Psalm\Internal\Analyzer\ClassAnalyzer::analyzeClassMethodReturnType()
#2 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ClassAnalyzer.php(758): Psalm\Internal\Analyzer\ClassAnalyzer->analyzeClassMethod()
#3 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/FileAnalyzer.php(210): Psalm\Internal\Analyzer\ClassAnalyzer->analyze()
#4 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(340): Psalm\Internal\Analyzer\FileAnalyzer->analyze()
#5 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Fork/Pool.php(193): Psalm\Internal\Codebase\Analyzer->Psalm\Internal\Codebase\{closure}()
#6 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(406): Psalm\Internal\Fork\Pool->__construct()
#7 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(269): Psalm\Internal\Codebase\Analyzer->doAnalysis()
#8 /root/.composer/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(639): Psalm\Internal\Codebase\Analyzer->analyzeFiles()
#9 /root/.composer/vendor/vimeo/psalm/src/psalm.php(676): Psalm\Internal\Analyzer\ProjectAnalyzer->check()
#10 /root/.composer/vendor/vimeo/psalm/psalm(2): require_once('/root/.composer...')
[...]

Issue #9672, opened 2023-04-18
ParseError: Problem loading method: $storage should not be null for Index::action                          


olleharstedt:
limesurvey/application/controllers/survey/index.php:32:5:error - ParseError: Problem loading method: $storage should not be null for Index::action                          

You have any idea what this error means? There's no syntax error in the file. Can't find anything about any $storage variable.

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/survey/index.php#L32

olleharstedt:
> Hey @olleharstedt, can you reproduce the issue on https://psalm.dev ?

Nop, too much code, little bot.

olleharstedt:
Hm seems related to the fact that the class is called `index`? And that we have _two_ such classes in the code-base. Heh.

orklah:
The error means that Psalm has tried to retrieve an object through an index that should be populated with every method but it's empty for Index::action.

Yeah, having two different classes in the same namespace *could* lead to such issues depending on which class is returned by your autoloading I suppose

Issue #9357, opened 2023-02-21
Uncaught Exception: UnexpectedValueException $storage should not be null for XXX


lens0021:
I have a private project and have set up psalm.xml file.
I am sorry the project is private so I can't upload the full source code of the project.

## ❌ Case 1

Executing `psalm` throws an uncaught exception. 
```console
% psalm               
Warning: "findUnusedBaselineEntry" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.
Warning: "findUnusedCode" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.

Install the opcache extension to make use of JIT on PHP 8.0+ for a 20%+ performance boost!

Target PHP version: 5.4 (inferred from composer.json) (unsupported extensions: mcrypt, grpc).
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░   60 / 1074 (5%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  120 / 1074 (11%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  180 / 1074 (16%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  240 / 1074 (22%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  300 / 1074 (27%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  360 / 1074 (33%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  420 / 1074 (39%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  480 / 1074 (44%)
[...]

PR #10515, opened 2023-12-22
Fix issue #4899

iMu3ic:
This issue is caused by the fact that class/interface could be present in two (or more) different files.
That's why the ClassStorage could be created twice, in two different threads.
In the first one, the class storage could be associated with one list of statements.
In another thread, the class with another list of statements is scanned instead, causing an error while fetching the first one list of statements.
So, it is multithreading issue, and I check the statement with the same name in foreach loop. And it works for 2, 4 threads.
I suppose that there is a more elegant fix for this issue.


iMu3ic:
@orklah could you please help me to fix tests?

iMu3ic:
Try to fix:
https://github.com/vimeo/psalm/issues/9672
https://github.com/vimeo/psalm/issues/9357
https://github.com/vimeo/psalm/issues/4899

orklah:
Hey! Sorry for not answering earlier, I was pretty busy!

Don't worry about tests about InternalCallmapHandler (mb_get_inf) this is unrelated to your PR, something changed in PHP itself.

However, errors from Run Shepherd should be adressed. Do you need help with those?

danog:
Found & merged a better fix.

Issue #10522, opened 2024-01-02
Psalm validation crashes laptop due to memory issues


mkroeders:
When running psalm on `microsoft/microsoft-graph` it crashes due to spawnining multiple, ~10, PHP processes. Each process continuously increases it's memory footprint, beginning from 0,5GB to over 5GB each. Which effectively makes makes it run out of memory;

Composer.json;
```
{
  "require": {
    "microsoft/microsoft-graph": "^2.1"
  },
  "require-dev": {
    "vimeo/psalm": "^5.0.0"
  },
  "config": {
    "allow-plugins": {
      "php-http/discovery": false
    }
  }
}

```
psalm.xml

```
<?xml version="1.0"?>
<psalm
        xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xmlns="https://getpsalm.org/schema/config"
        errorLevel="1"
    errorBaseline="psalm-baseline.xml"
>
    <projectFiles>
        <ignoreFiles>
            <directory name="vendor"/>
        </ignoreFiles>
    </projectFiles>
    <issueHandlers>
        <MixedAssignment errorLevel="suppress"/>
    </issueHandlers>
</psalm>
```

test.php
```
<?php
declare(strict_types=1);

require __DIR__ . '/vendor/autoload.php';

use Microsoft\Graph\GraphServiceClient;

$graphServiceClient = new GraphServiceClient($tokenContext, $scopes);
```

First thought the issue was with PHPstorm so reported it there, [bug report](https://youtrack.jetbrains.com/issue/WI-75665).

orklah:
Well, a known issue is that Psalm threads will follow the php.ini directive (or use the memory-limit option if provided) for *each* thread.

While each thread individually will probably consume less memory than analyzing everything in the same thread, it will inevitably have an overlap.

Psalm can also be pretty resource consuming on big codebases. Both of these combined can eat up a laptop memory pretty fast. Have you tried running in a single thread?

mkroeders:
If I use the option `--threads=1` I get a memory error;

```
PHP Fatal error:  Allowed memory size of 8589934592 bytes exhausted (tried to allocate 20480 bytes) in /vendor/vimeo/psalm/src/Psalm/Internal/Cache.php on line 76
[...]

PR #10532, opened 2024-01-08
Reduce memory consumption of caching and parallel processing without igbinary

sj-i:
This fix reduces the memory consumption of caching and parallel processing with default serialization.

See https://github.com/vimeo/psalm/issues/10522#issuecomment-1881729504 for more detail.

With this fix, the test script in the above issue can be analyzed with 3GB memory_limit, doesn't hit 5GB or 8GB memory consumption per process.

```
./vendor/bin/psalm --memory-limit=3G test.php
```

### DIFF
--- a/src/Psalm/Aliases.php
+++ b/src/Psalm/Aliases.php
@@ -2,8 +2,12 @@
 
 namespace Psalm;
 
+use Psalm\Storage\UnserializeMemoryUsageSuppressionTrait;
+
 final class Aliases
 {
+    use UnserializeMemoryUsageSuppressionTrait;
+
     /**
      * @var array<lowercase-string, string>
      */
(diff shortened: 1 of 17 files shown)


orklah:
Wow! That's pretty impressive!

Thanks for the massive work in this analysis and the PR associated!

staabm:
@sj-i I would like to test the same case in PHPStan to get an idea whether we have the same problem. 

Could you give me an idea on how to reproduce the psalm problem.

sj-i:
@staabm 
Hi!

Does the test script in #10522 consumes large amount of memory with PHPStan too?

At a quick glance, PHPStan uses `\var_export()` and `__set_state()` for caching things, and uses JSON in combination with a kind of hydration for IPC. So, PHPStan probably doesn't have the same problem. I'm currently considering the possibility of using `\var_export()` and `__set_state()` for Psalm also.

Anyway, this issue isn't the right place for talking about how to analyze the memory usage of PHPStan.

You can create an issue or a discussion thread in the PHPStan project, with a title like "Evaluate the memory usage", and then call me there if you want to see an usage example of Reli for PHPStan.

Issue #10535, opened 2024-01-09
Add SimpleXMLElement::__set() method handling


kthaler:
## Description
As of https://github.com/vimeo/psalm/pull/10049, we are unable to dynamically set properties in a `SimpleXMLElement`. So code like `$xmlElement->newProp = $newVal` causes psalm to flag the line with a `InvalidPropertyAssignment` error. 

## Potential Solution
We suggest updating [the Psalm stubs](https://github.com/vimeo/psalm/blob/5.x/stubs/extensions/simplexml.phpstub#L68) for SimpleXMLElement to include handling of the __set() method.

orklah:
Seems fine, could you push a PR?

PR #10536, opened 2024-01-09
Implement __set method in SimpleXMLElement stub

kthaler:
Adds the magic `__set` method to the SimpleXMLElement class in the simplexml stub to handle property assignments.

Closes https://github.com/vimeo/psalm/issues/10535


orklah:
We seem to have two tests named setPropertiesOfSimpleXMLElement1 and setPropertiesOfSimpleXMLElement2 that are failing.

They seem to do exactly what you want but they were expecting an error. I don't see why we would want to emit an error for those, do you see something wrong with those snippets?

If not, I think we should just remove them

orklah:
Thanks!

PR #10537, opened 2024-01-10
update master

### LINKED ISSUE #10460: Array keys not retained when using splat operator to define class constant
When using the splat operator to define a class constant, Psalm seems to resolve the resulting array as having integer keys instead of retaining the key type of the source array.

https://psalm.dev/r/32157859fc

weirdan:


danog:
Just did this yesterday :)

PR #10542, opened 2024-01-12
remove redundat directory separator which caused "//" in path not found errors

kkmuffme:


orklah:
Thanks!

gready-hub:
Hi there. I've found this to break how the error baseline is used. Due to the removal of the trailing `/` from the config path Psalm does not locate the files from the error baseline due too this one missing character.

Either the generate baseline needs to add the prefix or we need to remove the prefix `/` when checking if the path is in the `$issues_baseline` map.

From `src/Psalm/Internal/Cli/Psalm.php`
```php
        if ($paths_to_check !== null) {
            $filtered_issue_baseline = [];
            foreach ($paths_to_check as $path_to_check) {
                // These two lines below are the problem... 
                $path_to_check = substr($path_to_check, strlen($config->base_dir));
                if (isset($issue_baseline[$path_to_check])) {
                    $filtered_issue_baseline[$path_to_check] = $issue_baseline[$path_to_check];
                }
            }
            $issue_baseline = $filtered_issue_baseline;
        }
```

This resolves the issue I have been experiencing... Either increment by one or remove prefix `/` prior.
```
// ...
$path_to_check = substr($path_to_check, strlen($config->base_dir) + 1);
// ...
```

PR #10543, opened 2024-01-12
Fix mb_get_info can return null - CI failing bc of reflection

kkmuffme:
See https://github.com/php/php-src/issues/12753


orklah:
Thanks!

Issue #7361, opened 2022-01-09
`extract($arr, EXTR_SKIP)` should not invalidate previous variables


gharlan:
https://psalm.dev/r/35e85a72e3

When using the `EXTR_SKIP` flag for `extract()`, the existing variable types should not be invalidated by the `extract` call.

orklah:
It's handled here: https://github.com/vimeo/psalm/blob/7c24d33bb20e127676f8e99e4f974f83eabbb2e0/src/Psalm/Internal/Analyzer/Statements/Expression/Call/NamedFunctionCallHandler.php#L200

Currently, every variable in the scope except a few exception is turned into mixed

PR #10544, opened 2024-01-14
add support for extract to set variables for keyed arrays and respect EXTR_SKIP

kkmuffme:
Add support for extract set variables in keyed arrays and also:
Fix https://github.com/vimeo/psalm/issues/7361


orklah:
Thanks!

PR #10566, opened 2024-01-17
Reproducible builds (fixes #10446)

### LINKED ISSUE #10446: Adding `composer.lock` in VCS ?
Hi there,

I'm writing to suggest an enhancement for the `vimeo/psalm` project to improve its integration and ensure long-term reliability, particularly for package managers like apt, rpm, yum, nix, etc etc.

The primary goal here is to enable reproducibility of the exact build of the PHAR released, down to the last bit. This is crucial, especially when distributing it as a binary for reasons of reliability and security.

To achieve this, I propose adding a `composer.lock` file to the `vimeo/psalm` repository. This file will ensure that `psalm` is reproducible on any system, regardless of environmental differences, by locking down specific versions of dependencies. Its absence currently hinders the ability to precisely replicate the builds you provide in your GitHub releases, introducing potential variability.

For example, in Nix, we have to provide our own `composer.lock` for each update, as seen here: https://github.com/NixOS/nixpkgs/pull/271920

For reference, here are some projects that have recognized the value of versioning the `composer.lock` file:

- https://github.com/sebastianbergmann/phpunit/pull/5576
- https://github.com/bobthecow/psysh/issues/767
- https://github.com/phpro/grumphp/issues/1095
- https://github.com/NixOS/nixpkgs/pull/255253
- https://github.com/phpstan/phpstan-src/blob/1.11.x/composer.lock

If there are concerns about including the `composer.lock` file in the main repository, could it be an option to add it to the release artifacts on GitHub instead? This compromise would still greatly benefit reproducibility and consistency in dependency management.

Thank you for considering this enhancement. I look forward to your response.

danog:


vendor-bin/box/composer.json
```diff
@@ -2,11 +2,14 @@
 	"minimum-stability": "dev",
 	"prefer-stable": true,
 	"require": {
-		"humbug/box": "^3.16.0"
+		"humbug/box": "^4"
```
drupol, line 5: I would require `^4.6` here.

composer.json
```diff
@@ -74,7 +74,8 @@
         },
         "optimize-autoloader": true,
         "sort-packages": true,
-        "platform-check": true
+        "platform-check": true,
+        "autoloader-suffix": "PsalmAutoloader"
```
drupol, line 78: This is not required if you use Composer >= 2.6.4, see relevant PR @ https://github.com/composer/composer/pull/11663

danog: This is not relevant for us, as the autoloader suffix is reused only if composer.lock is commited, which I've decided not do for psalm itself (only box) for a number of reasons, like https://github.com/vimeo/psalm/issues/10446#issuecomment-1896650439

drupol:
Hi,

What's missing in this PR? I really wish we could move forward with it.

drupol:
Hi,

I am sorry to dig up old topics, but we (the Nix and reproducible build communities) would really benefit to be able to use the `composer.lock` file that has been used to generate the PHAR file.

Would it be possible to publish it along with the PHAR release?

drupol:
Thank you ! I am glad of this outcome, I think this should be the default behaviour when publishing a PHAR. 

I pushed an update in Nix and remove the custom `composer.lock` that I was using @ https://github.com/NixOS/nixpkgs/pull/384574

PR #10567, opened 2024-01-18
Update nikic/php-parser to 5.0

edsrzf:
I saw #10562, and I've also been playing around with this.

The main issues I encountered in this update were:
* Removal of `PhpParser\Node\Stmt\Throw`. Throws are now always treated as expressions.
* Doc comments are only associated with one node. This affects expression statements and expressions passed as arguments. I'm not super happy with my fixes for this inside `ReflectorVisitor`, so very open to suggestions.
* Some node inheritance hierarchies have changed.


edsrzf:
I've marked this as ready for review. I still have some doubts about the way I've solved a few of these issues, but am hopeful that feedback will help ensure that the best solution is the one that gets merged.

I've incorporated several changes upstream in the `5.x` branch and rebased this PR, so the diff is a little smaller than it originally was. I'm not sure there's anything else I could reasonably pull out, but let me know if you see anything!

composer.json
```diff
@@ -33,15 +33,12 @@
         "felixfbecker/language-server-protocol": "^1.5.2",
         "fidry/cpu-core-counter": "^0.4.1 || ^0.5.1 || ^1.0.0",
         "netresearch/jsonmapper": "^1.0 || ^2.0 || ^3.0 || ^4.0",
-        "nikic/php-parser": "^4.16",
+        "nikic/php-parser": "^5.0.0",
```
derrabus, line 36: This is a dependency that Psalm shares with other tools like PHPUnit. In order to allow downstream projects to upgrade their dependencies gracefully, Psalm should produce a release first that is compatible with both major versions.

```suggestion
        "nikic/php-parser": "^4.18 || ^5.0",
```

edsrzf: I'm open to trying to create a version of this PR that would land on the `5.x` branch that would work with both parser v4 and v5. I think the main thing would be going back to the deprecated node names (eg `Encapsed` vs `Interpolated`), which are still supported in v5 via class aliases.

The thing I'd worry about is additional complexity, maintenance, and testing involved in supporting two versions. The differences are non-trivial, as evidenced by this PR. PHPStan also looks like it's planning to jump to parser v5 on its 2.0 release, with its 1.x releases only supporting parser v4.

I'd like feedback from the maintainers before making any major changes to this PR or creating a new one.

danog:
Thank you for this PR.
I also initially intended to support both v4 and v5 in my php-parser upgrade PR, but then also noticed the non-triviality of the changes and the extra maintenance burden of supporting both version, and pivoted towards v5-only support in my PR.

If you would want to provide a second PR to 5.x to support both v4 and v5, that would of course be appreciated, but personally I don't think dropping v4 support for Psalm v6 is that big of a deal (regardless, I would absolutely merge a PR adding support for it).

weirdan:
@danog it appears it broke psalm.dev sandbox. All snippets return
``` 
Internal Psalm error on line 25 - /src/OnlineChecker.php: Call to undefined method PhpParser\ParserFactory::create()
```

weirdan:
> it broke psalm.dev sandbox

Fixed now.

BenMorel:
Will this be merged into `5.x`?

edsrzf:
I don't have any plans to backport this to 5.x, personally.

Slamdunk:
@edsrzf with no rush and no pressure, may I ask you if there's an expected release date for `v6`?

edsrzf:
I'm not a maintainer, so it's not up to me, but I started a discussion here: https://github.com/vimeo/psalm/discussions/10701

VincentLanglet:
I tried to work on the Back port of this PR on 5.x.
https://github.com/vimeo/psalm/pull/11035

I just have one error to fix, any help is welcomed.
https://github.com/vimeo/psalm/actions/runs/10094010337/job/27910940740?pr=11035

VincentLanglet:
PR on 5.x is ready https://github.com/vimeo/psalm/pull/11035 open to review or tests on real project :)

PR #10575, opened 2024-01-19
Assert on var of nullsafe expr

### LINKED ISSUE #10554: If a null-safe method call does not return null, it implies that the object is definitely not null
Trivially `if ($foo->check() === false) { /* $foo is not null! */ }`.

In fact psalm seems to this $foo is null, any not just possibly null here.

https://psalm.dev/r/ccf3505ede

danog:
Fixes #10554

### DIFF
--- a/tests/MethodCallTest.php
+++ b/tests/MethodCallTest.php
@@ -196,6 +196,60 @@ class MethodCallTest extends TestCase
     public function providerValidCodeParse(): iterable
     {
         return [
+            'nullSafeCallNotNullMakesVarNotNull' => [
+                'code' => '<?php
+                    class Foo {
+                        public function check(): bool {
+                            return false;
+                        }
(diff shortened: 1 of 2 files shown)


tests/MethodCallTest.php
```diff
@@ -196,6 +196,60 @@ function printInt(int $int): void {
...
+                            return false;
+                        }
+                        public function do(): void {}
+                    }
+
+                    /** @var ?Foo */
+                    $foo = null;
+
+                    if ($foo?->check() === null) {
+                        /** @psalm-check-type-exact $foo = null */
```
orklah, line 212: I would be curious to see the same test but with check that returns "?bool". Because then you can't infer anything from the nullsafe call

danog: Yep my exact thoughts, I still need to finish this PR

Issue #9838, opened 2023-05-30
Checked alternative for `@var` annotation / Override Psalm's template type inference


janopae:
Sometimes, you need to declare a type manually, as the inferred type is too specific:

```php
/**
* @template Tkey of array-key
* @template T
*/
class ArrayCollection {
    /**
    * @param array<Tkey, T> $array
    */
    public function __construct(private array $array){}
}



$collection = new ArrayCollection(['a', 'b']);

/** @palm-trace $collection: ArrayCollection<int<1,2>, 'a'|'b'> */

```

This type would probably be pretty useless in most cases, because which function would take an `ArrayCollection<int<1,2>, 'a'|'b'>` as a parameter? The problem is that [you can't always pass a subtype into a supertype declaration](https://psalm.dev/docs/annotating_code/templated_annotations/#template-covariance) (so you couldn't pass this variable into a function that takes an `ArrayCollection<int, string>`, for example).

You could do it like this:

```php

/** @var ArrayCollection<array-key, string> $collection */
$collection = new ArrayCollection(['a', 'b']);
```

However, inline `@var` annoations are equivalents to `@psalm-ignore` and destroy the soundness of the type system, because they are completely unchecked. So in the future, the signature of the constructor (or better imagine a factory method instead) might change and return something else that is incompatible with the declared type, and nobody would notice.

I'm not aware of a way to provide a less specific type in such a case while still having psalm keeping track of it.

I'd propose either
1. changing the way inline `@var` annotations work to being type-checked (huge breaking change)
2. add a new annotation like `@templated`, that syntactically works like a type cast (so you'd write `$collection = /** @templated ArrayCollection<string, int> */ new ArrayCollection();` and don't need a variable name) and overrides Psalms automatic template type inference and  ("templated" because AFAIK templates are the only kind of type where [you can't always pass a subtype into a supertype declaration](https://psalm.dev/docs/annotating_code/templated_annotations/#template-covariance) and therefore need this. Alternative name: `@cast`)

### A super-dooper advanced solution

[...]

PR #10577, opened 2024-01-19
Allow specifying a custom default template type

danog:
This is the first step in my plan to fix https://github.com/vimeo/psalm/issues/9838.

First of all, the default template types all SPL datastructures will be changed to `never`; then, a new `@satisfies` assertion will be introduced to avoid the (ab)use of `@var`.

The semantics of `@satisfies` will be entirely similar to typescript's [satisfies operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html), and will allow, among other things, typechecked call-site specification of template parameters.  

Here's a few examples of how I generally plan `@satifies` to work (and how the SPL stubs will be tweaked), I'm open to suggestions:

- https://psalm.dev/r/fcc7e48e65
- https://psalm.dev/r/59402fff1a

### DIFF
--- a/docs/annotating_code/templated_annotations.md
+++ b/docs/annotating_code/templated_annotations.md
@@ -113,6 +113,54 @@ function array_combine(array $arr, array $arr2) {}
 - `@template` tag order matters for class docblocks, as they dictate the order in which those generic parameters are referenced in docblocks.
 - The names of your templated types (e.g. `TKey`, `TValue`) don't matter outside the scope of the class or function in which they're declared.
 
+## Default template values
+
+By default, if a class template parameter is not initialized in the constructor or by a `@var` annotation, Psalm will initialize it to `mixed`.  
+
+However, a custom default initial type can also by provided using the following syntax: 
+
+```php
+<?php
+
+/**
+ * Template T can have any type (`mixed`), but if it's not specified, it will default to `never`.
+ * 
+ * @template T as mixed = never
+ */
+class MyContainer {
+  /** @var list<T> */
+  private array $values = [];
+
+  /** @param T $value */
+  public function addValue($value): void {
+    $this->values []= $value;
+  }
+
+  /** @return list<T> */
+  public function getValue(): array {
+    return $this->values;
+  }
+}
+
+$t1 = new MyContainer;
+
+/** @psalm-trace $t1 */; // MyContainer<never>
+
+// InvalidArgument: Argument 1 of MyContainer::addValue expects never, but 123 provided
+$t1->addValue(123); 
+
+
+/** @var MyContainer<int> Always specify template parameters! */
+$t2 = new MyContainer;
+
+/** @psalm-trace $t2 */; // MyContainer<int>
+
+// OK!
+$t2->addValue(123); 
+```
+
+This can be often useful, like in the above example, to always force specification of a template type when constructing generic objects by specifying `never` as default type: `never` is the [bottom type](https://psalm.dev/docs/annotating_code/type_syntax/top_bottom_types/#never), and thus all usages of methods relying on an *uninitialized* template type will use `never` and will emit Psalm issues, essentially warning the user to explicitly specify a template parameter when constructing the class.  
+
 ## @param class-string&lt;T&gt;
 
 Psalm also allows you to parameterize class types
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/MethodCallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/MethodCallAnalyzer.php
@@ -30,11 +30,13 @@ use Psalm\Issue\UndefinedMagicMethod;
 use Psalm\Issue\UndefinedMethod;
 use Psalm\IssueBuffer;
 use Psalm\Type;
+use Psalm\Type\Atomic\TConditional;
 use Psalm\Type\Atomic\TNamedObject;
 use Psalm\Type\Atomic\TObject;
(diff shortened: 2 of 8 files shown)


janopae:
I think this is a great effort that would handle the edge cases presented in #9838, along with generally proviting greater accuracy when dealing with values inside containers.

I think it is important that `@satisfies` only works on creation, and can't be added later-on. We then still have the open question whether it should work for factory methods, and under which conditions a function would qualify as "factory function" – see https://github.com/vimeo/psalm/issues/9838#issuecomment-1578787773.

klimick:
> I think it is important that @satisfies only works on creation

Why it should be limited to the `new` expression?
[Typescript allows this](https://www.typescriptlang.org/play?#code/FAYw9gdgzgLgBANwIxwLxwOYFMYGUwC2OAFgJYQYAUAlHFAIYylQBmpWUcASluAE4ATADyw+5DABo6MMRTgAfOBACuBAEZY+APgDcoSLEQAmNJhz4iMMhRp1GzNh268wgkTPFTR4hUtUa+XzUwMAAbLHoIXWBgFmUIECZIMzxCEnFbAG9gOFy+HGU+CDhs3LK4FlgALjgkCRzyuggBGoAiAGIABk6WFm7W+sa4Kz4amWUsQdyAXzLgaeAgA)

klimick:
I can't undestand how it solves this issue: https://psalm.dev/r/7da42be7d7

```php
/** @satisfies MyContainer<int> */
$t = new MyContainer;

/** @psalm-trace $t */; // MyContainer<never: int>

// OK!
$t->addValue(123); 

/** @psalm-trace $t */; // MyContainer<123: int>

/**
 * @param MyContainer<int> $container
 */
function test(MyContainer $container): void {}
test($t);

/** @psalm-trace $t */; // What type will be here?
```

janopae:
> Why it should be limited to the new expression?
[Typescript allows this](https://www.typescriptlang.org/play?#code/FAYw9gdgzgLgBANwIxwLxwOYFMYGUwC2OAFgJYQYAUAlHFAIYylQBmpWUcASluAE4ATADyw+5DABo6MMRTgAfOBACuBAEZY+APgDcoSLEQAmNJhz4iMMhRp1GzNh268wgkTPFTR4hUtUa+XzUwMAAbLHoIXWBgFmUIECZIMzxCEnFbAG9gOFy+HGU+CDhs3LK4FlgALjgkCRzyuggBGoAiAGIABk6WFm7W+sa4Kz4amWUsQdyAXzLgaeAgA)

As I wrote in my comment, I don't think it should be limited to the `new` expression, but also allow for factory functions (like the one in your TypeScript example).

But you shouldn't be able to do [this](https://www.typescriptlang.org/play?#code/FAYwNghgzlAEDKB7AtgUwC4AsCWA7A5gGIQjqIBOAnrAN7CwP4ZJpZ74AUAlLfQ-+QwBXcrlhsoAOigQAbqkIUAMhHSpyAVSioOdfvoYAzKOgBcsAIwAaPgYZRcAE3MAiAMQAGD4cNeXNu34scnN0ciFUAIMAXy4AblsDW2jbGXlHAH5zGmMzSzjYB2dYdy8fPytxTBDxcNRohP4AB3JsWVVUQrkFZQ7NbQ5EACMAK1RSbNzzCwKi109vXw9-KpqwiNjeOwlpbsdYAF5YYbHSRoNBdBExE-H0c9gUlOBDIVxSbEQxbXRCEwAVRD-OocWTmABK4wojgAPCZWgRKvD2LAAD6wXBCZBDdRo2BDRCIMCoCC4AB8PD0DFkklyh1qEQSzxAXxMsEMJDIVHpuFQAHcECgMDgCMRSBRKNwEizcGzZPSOeKqJImOgWML2NwuuhsFBDNhUHBISzyLDkYjCmEUejMdjceiCUSSeSEsAfn90IDgRFQfFgMBsIZYBw3o5UPrefsAIQHI6KrmUXbpSkMWwyqBOyRgRCceMSpOoRy0tkAaks8VgAHpK8dEE04IBkAnZnPzaULxfQsF1sAAButUD2MYg+cBnsAgA).

When using `satisfies`, the instance must not have been saved somewhere else before, where the `statisfied` requirement can't be ensured.

danog:
@kkmuffme 

1. Sure
2. Sure, that can be done.
3. This is an error like in 2, `@satisfies` is not `@var`.

weirdan:
> `@template T as mixed = whatever`

The choice of the syntax is backward incompatible as previously `= whatever` would be treated as a part of the comment. To be on the safe side we must choose a syntax that is currently rejected by Psalm, e.g. `@template T as mixed : whatever`.

edsrzf:
If I understand correctly, there are two or three things happening in this change:
* A way to specify a default template parameter -- this is the backward incompatible change.
* A safe way to specify template parameters.
* A way to make a variable's type "sticky", so it can't change, more like other languages. (This is the part I'm not sure I'm correct about. I'm taking this from the second snippet link in the PR description.)

Given that only the first thing is backward incompatible, maybe we could focus efforts on just this part, in order to make progress toward Psalm 6.0? This by itself still seems valuable, and the rest can be done later without breaking compatibility.

A couple other thoughts on the overall idea:

This seems a little different from TypeScript's `satisfies`. In TypeScript, `satisfies` is purely an assertion and never changes the type of a variable. But here it sounds like we're wanting to change the variable's type?

I wonder if it would be worth further separating the second and third concepts above? It's not obvious to me that safely specifying template parameters should also cause the variable's type to be unchangeable.

PR #10586, opened 2024-01-22
Fix template replacement edge case

klimick:
Try to fix this issue: https://psalm.dev/r/775f5a3985

Interestingly, the problem only appears when using the function/method.
There is no issue with `new` expression: https://psalm.dev/r/4e1c42b658


klimick:
To be honest, I haven't found a better way

vudaltsov (approved):

orklah:
Thanks!

PR #10588, opened 2024-01-23
Partial revert "Fix auto completion by partial property or method"

issidorov:
Filtering is not necessary. Clients using LSP should filter the results themselves. That's what it says in the documentation.

Can see on [page](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList):

```typescript
export interface CompletionList {
    /**
    * This list is not complete. Further typing should result in recomputing
    * this list.
    *
    * Recomputed lists have all their items replaced (not appended) in the
    * incomplete completion sessions.
    */
    isIncomplete: boolean;

    ...
}
```

In our code, this property `isIncomplete` is passed as `false`.
Therefore, we must transfer everything we have, and the client's IDE must filter.

This reverts commit d6faff2844dbcd40df108052fbdd929bdeafc7a9.


src/Psalm/Codebase.php
```diff
@@ -2036,26 +2036,6 @@ public function getCompletionItemsForClassishThing(
         return $completion_items;
     }
 
-    /**
-     * @param list<CompletionItem> $items
-     * @return list<CompletionItem>
-     */
-    public function filterCompletionItemsByBeginLiteralPart(array $items, string $literal_part): array
```
weirdan, line 2043: You'll need to keep the method to satisfy backward compatibility checks, but mark it `@deprecated to be removed in Psalm 6`.

issidorov: Thanks, I'll fix it. I didn't notice that a release had already been made. I thought this method was not yet in release.

issidorov: > PossiblyUnusedMethod: Cannot find any calls to method Psalm\Codebase::filterCompletionItemsByBeginLiteralPart

another problem )))

issidorov: problem is fixed by annotation `@api`...

weirdan (approved):

weirdan:
Thanks!

PR #10594, opened 2024-01-25
Switch condition order

edsrzf:
This change is for forward-compatibility with `nikic/php-parser` 5.0, where `InterpolatedStringPart` (née `EncapsedStringPart`) is no longer considered an expression. Thus we can't pass it to `NodeTypeProvider::getType()` anymore. Since that call returns `null` anyway, we can swap the condition order. Everything still works and Psalm type-checking is happy.

This also might be a tiny performance improvement since it lets the common, cheap `instanceof` check come before a method call, but I haven't actually benchmarked it.

I've extracted this change from #10567 to reduce the size of the diff, reduce merge conflicts, and make review easier. I will rebase that PR once this is merged and the 5.x branch is merged to master.


orklah:
Thanks!

Issue #10414, opened 2023-11-27
Intersection between UnitEnum / BackedEnum and Interface


hockdudu:
When a function has a parameter of `\BackedEnum&Interface` (intersection between Enum and Interface), Psalm throws an exception:

```
Psalm runner output: 

Internal Psalm error on line 47 - /vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php: Could not get class storage for backedenum
```

Here's the code: https://psalm.dev/r/4d1f328b6f

My use case is more or less the following: https://psalm.dev/r/7fb43699a6

In my case, I can fix it by just setting the parameter to `Interface` (remove the intersection with Enum), but still, here's the bug report. There could be valid cases where one needs to have an Enum (e.g. to get all cases) and still call the interface method at the same time.

(This can be reproduced with `\UnitEnum` as well)

weirdan:
Duplicate of #7520

PR #10602, opened 2024-01-27
Fix intersection with internal interfaces

weirdan:
Fixes vimeo/psalm#10414


weirdan:
Nope, this didn't work. It appears the problem is caused by TypehintResolver trying too hard to simplify intersection during the scanning phase, before the codebase has been fully populated.

PR #10610, opened 2024-01-28
Remove unnecessary null type from `initialized_methods` Context property

theodorejb:
Also optimize ConfigFileTest slightly.


weirdan:
Looks like yu need to target `master` instead.

theodorejb:
@weirdan Thanks. The type declarations were already added to master, so this PR is much smaller now.

weirdan:
Thanks!

PR #10612, opened 2024-01-29
Re-work CheckTrivialExprVisitor

edsrzf:
In php-parser 5.0, `ClosureUse` is no longer considered an expression. This requires changes to Psalm's `CheckTrivialExprVisitor`, which stores an array of "non-trivial" `Expr` nodes.

However the only use of this array is to count whether or not it's empty. Instead of keeping the array, we can keep a boolean, and avoid needing to change the types in this class when we upgrade to php-parser 5.0.

See #10567, which this was extracted from.


weirdan:
Thanks!

Issue #3081, opened 2020-04-06
ownerDocument can be null


snapshotpl:
Following docs https://www.php.net/manual/en/class.domnode.php#domnode.props.ownerdocument

https://psalm.dev/r/7eac58778c

PR #10619, opened 2024-01-31
Fix ownerDocument type in dom-ext classes

fluffycondor:
Introduced in #3081 https://github.com/vimeo/psalm/commit/26fcde10fa937bb84896bc37ccf653b2f4c162db

PHP docs says:
> _ownerDocument_
> The DOMDocument object associated with this node, or null if this node is a DOMDocument

https://www.php.net/manual/en/class.domnode.php#domnode.props.ownerdocument

And actually PHP enforces this contract:
https://3v4l.org/tWoCc

PHP 8.1+ throws DOMException about object broken state if DOMNode doesn't have an onwerDocument, so it can't be null.
PHP 8.1- allows get null from this field, but it throws a warning about object broken state. While it's possible to add this behavior change in CallMap_81_delta, I believe that manipulating an object with broken state should be discouraged, and Psalm shouldn't cover this edge case.

Probably it should be forbidden to call DOMNode and its children constructor directly—except for DOMDocument—to prevent creating dom object with broken state. Marking those with `@internal` would be the easiest solution, but it could be better to show custom issue message like "using DOMNode and its children constructor directly results in broken object's internal state, use corresponding DOMDocument::create* method instead". 
I'm not sure how it could be right to reinforce this check, so if somebody gives me a hint, I'm ready to make a PR :)


weirdan (approved):

weirdan:
Thanks!

KevinVanSonsbeek:
Not gonna lie, i find this change in behavior a bit odd. Escalating from a warning to a fatal error. In a minor update.

As the [manual](https://www.php.net/manual/en/class.domnode.php#domnode.props.ownerdocument) does not mention this change in behavior. But i also don't see any mentions of the DOM logic in the php 8.1 migration guide. (Which i find kinda odd tbh)

Could this possibly be a regression from other changes in php 8.1? (Might open an issue on the source repo to check. It would be preferable for the change to be noted on the wiki atleast somewhere)

theseer:
Imho, this change is wrong and not following the actual implementation in PHP.

The example linked above (https://3v4l.org/tWoCc) about PHP enforcing the non-nullable type is also incorrect, as a modified example demonstrates: https://3v4l.org/fGc7t

In other words: The `ownerDocument`-Property can very much (still) be null. Whether or not this is a desirable behavior is a different story.

PR #10620, opened 2024-01-31
Encourage using DOMDocument factory methods to prevent objects broken state

fluffycondor:
Manually created DOMNode objects have broken internal state because they don't have `ownerDocument` until added to one.
It leads to DOMException/warning `Invalid State Error` when calling `ownerDocument` field.
This PR encourages using factory methods that immediately bind freshly created object to `ownerDocument` and thus their internal state is valid.

### DIFF
--- a/stubs/extensions/dom.phpstub
+++ b/stubs/extensions/dom.phpstub
@@ -256,6 +256,9 @@ class DOMDocumentFragment extends DOMNode implements DOMParentNode
     /** @readonly */
     public int $childElementCount;
 
+    /**
+     * @psalm-internal DOMDocument::createDocumentFragment
+     */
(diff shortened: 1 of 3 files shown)


fluffycondor:
Looks like the failed check is unrelated

weirdan:
I'm not so sure about this. How people use it is usually `...->appendChild(new DOM...)`, and that's perfectly valid.

Source: https://github.com/search?q=%22new+DOMText%22+language%3APHP&type=code

weirdan:
Perhaps we could do something like this instead: https://psalm.dev/r/26c873ea00

weirdan:
Or this: https://psalm.dev/r/4bc562b64e

Issue #10561, opened 2024-01-17
Assumed filter_var return type can be incorrect


pontus-mp:
Psalm gets a bit confused about the return value from `filter_var` when called with `FILTER_VALIDATE_INT`, a numeric value and some options.

https://psalm.dev/r/c1579d2079

NicolasCARPi:
This bug is introduced in https://github.com/vimeo/psalm/pull/10498 and the PR author indicates:

>  I only handled the most commonly used basic ones. This is just a starting point and can be easily improved with additional types once errors/issues are reported.

So here we are, reporting issues and gently pinging @kkmuffme to ask if they would have the time to look into it :)

Downgrading to 5.18.0 is the current workaround if one doesn't want to add ignore comments.

PR #10621, opened 2024-01-31
Fix numeric scalar validate filter var input return type wrong

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/10561


weirdan (approved):

weirdan:
Thanks!

PR #10629, opened 2024-02-01
Analyze dynamic names for static property and const fetches

edsrzf:
While trying to use PHP 8.3's new dynamic class constant syntax, I found that the dynamic constant name wasn't being analyzed. Then I tried to see what dynamic static properties were doing and noticed that they had the same problem!

This fixes both issues.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php
@@ -206,7 +206,15 @@ final class ClassConstAnalyzer
             }
 
             if (!$stmt->name instanceof PhpParser\Node\Identifier) {
-                return true;
+                $was_inside_general_use = $context->inside_general_use;
+
+                $context->inside_general_use = true;
+
+                $ret = ExpressionAnalyzer::analyze($statements_analyzer, $stmt->name, $context);
+
+                $context->inside_general_use = $was_inside_general_use;
+
+                return $ret;
             }
 
             $const_id = $fq_class_name . '::' . $stmt->name;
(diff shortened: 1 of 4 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php
```diff
@@ -206,7 +206,15 @@ public static function analyzeFetch(
             }
 
             if (!$stmt->name instanceof PhpParser\Node\Identifier) {
-                return true;
+                $was_inside_general_use = $context->inside_general_use;
+
+                $context->inside_general_use = true;
+
+                $ret = ExpressionAnalyzer::analyze($statements_analyzer, $stmt->name, $context);
```
weirdan, line 213: I suppose it should be version guarded, to make sure we don't allow this syntax when target version is below 8.3

weirdan: Also it's possible to access enum cases in the same way in 8.3, but it doesn't seem to be handled.

weirdan (changes requested):

src/Psalm/Internal/Analyzer/Statements/Expression/ClassConstAnalyzer.php
edsrzf: Ah, I thought the version check would be handled by the parser, but apparently not. What behavior would you expect on a lower version than 8.3? `ParseError`?

weirdan: `ParseError` sounds good, as that's what PHP emits: https://3v4l.org/DfCB2

weirdan (approved):

weirdan:
Thanks!

PR #10630, opened 2024-02-01
Fix additional places where base_dir was broken due to missing separator

kkmuffme:
* Improves upon https://github.com/vimeo/psalm/pull/10542 and https://github.com/vimeo/psalm/pull/10628
* Fix (some) psalm output incorrectly sent to STDOUT instead of STDERR and exit status 0 on error
* improve return type of getcwd, file_put_contents, readlink and realpath
* fix psalm E2E tests potentially re-using/sharing data between independent tests


src/Psalm/Config.php
```diff
@@ -1817,7 +1817,7 @@ private function getPluginClassForPath(Codebase $codebase, string $path, string
     public function shortenFileName(string $to): string
     {
         if (!is_file($to)) {
-            return preg_replace('/^' . preg_quote($this->base_dir, '/') . '/', '', $to, 1);
+            return preg_replace('/^' . preg_quote($this->base_dir . DIRECTORY_SEPARATOR, '/') . '?/', '', $to, 1);
```
weirdan, line 1820: I think directory separator shouldn't be optional in the regexp:
image

src/Psalm/Internal/Cli/Psalter.php
```diff
@@ -354,7 +364,11 @@ public static function run(array $argv): void
             );
 
             if ($doc_block_add_new_line_before_return === null) {
-                die('--add-newline-between-docblock-annotations expects a boolean value [true|false|1|0]' . PHP_EOL);
```
weirdan, line 357: We should also add `die` to `psalm.xml/forbiddenFunctions`, so that future code consistently uses `exit` instead.

weirdan (changes requested):

src/Psalm/Internal/Cli/Psalter.php
weirdan: Just `psalm.xml.dist`. Even though `die()` and `exit()` are synonymous, I have the impression that people generally use `die()` with a string argument and `exit()` with an int, and we want Psalm sources to use the latter.

weirdan (approved):

weirdan:
Thanks!

PR #10637, opened 2024-02-02
CallMap: Adjust return type for `inotify_add_watch()` to `int|false`

### LINKED ISSUE #10636: Wrong return type for `inotify_add_watch()`, should be `int|false`
Heya!

This is about https://www.php.net/manual/en/function.inotify-add-watch, which clearly returns an `int` as per documentation. However, it doesn't: If the path doesn't exist, it raises a PHP warning and returns `false`.

Code for reproduction:
```php
<?php

declare(strict_types=1);

$inotify = inotify_init();

$watch = inotify_add_watch($inotify, '/nonexistant/path', IN_ALL_EVENTS);
if ($watch === false) {
    echo "inotify_add_watch() failed\n";
}
```

Output:
```raw
PHP Warning:  inotify_add_watch(): No such file or directory in /home/uli/simplequeue/inotify-bug.php on line 7
PHP Stack trace:
PHP   1. {main}() /home/uli/simplequeue/inotify-bug.php:0
PHP   2. inotify_add_watch($inotify_instance = resource(4) of type (stream), $pathname = '/nonexistant/path', $mask = 4095) /home/uli/simplequeue/inotify-bug.php:7
inotify_add_watch() failed
```

Up to here it's a PHP bug, not a Psalm one. However, check out [this example](https://psalm.dev/r/4e5bb84c24). Psalm there reports a not helpful "TypeDoesNotContainType - int does not contain false" for the valid check on the returned value.

In addition, but that's just the online playground, it does know `inotify_add_watch()`, but it doesn't know any of the constants like `IN_ALL_EVENTS`. That's a minor issue though.

Greetings!

Uli

UlrichEckhardt:
This should fix #10636

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -5700,7 +5700,7 @@ return [
 'ini_restore' => ['void', 'option'=>'string'],
 'ini_parse_quantity' => ['int', 'shorthand'=>'non-empty-string'],
 'ini_set' => ['string|false', 'option'=>'string', 'value'=>'string|int|float|bool|null'],
-'inotify_add_watch' => ['int', 'inotify_instance'=>'resource', 'pathname'=>'string', 'mask'=>'int'],
+'inotify_add_watch' => ['int|false', 'inotify_instance'=>'resource', 'pathname'=>'string', 'mask'=>'int'],
 'inotify_init' => ['resource|false'],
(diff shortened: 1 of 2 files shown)


weirdan (approved):

weirdan:
Thanks!

PR #10639, opened 2024-02-02
Allow sebastian/diff v6

simPod:
This allows installing phpunit v11 in consumer apps since it requires sebastian/diff v6.


weirdan:
Thanks!

Issue #10560, opened 2024-01-16
"Failed to infer case value" when enum case references another enum


aurimasrim:
Hi,
I've encountered an issue where I get "Failed to infer case value" when a case of an enum references the value of another enum's case.

```php
enum SomeEnum: string
{
    case SOME_VALUE = 'SOME_VALUE';
}

enum AnotherEnum: string
{
    case ANOTHER_VALUE = SomeEnum::SOME_VALUE->value;
}
```

https://psalm.dev/r/d4e839fc63

PR #10643, opened 2024-02-02
Fix enum crash

AegirLeet:
Fixes https://github.com/vimeo/psalm/issues/10560.

First commit fixes the crash.

Second commit fixes the `InvalidEnumCaseValue` revealed once the crashing is fixed.

Third commit adds two test cases for the crash fix.

I'd like to add some more test cases to validate the value is resolved correctly, but not sure where.

### DIFF
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php
@@ -1437,10 +1437,16 @@ final class ClassLikeNodeScanner
         $case_location = new CodeLocation($this->file_scanner, $stmt);
 
         if ($stmt->expr !== null) {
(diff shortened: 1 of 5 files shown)


weirdan:
Does https://psalm.dev/r/38dc5a68ea (referencing the enum below rather than above) work with your PR?

AegirLeet:
No, that still fails. Looks like the `ClassLikeStorageProvider` doesn't know about `SomeEnum` at that point. This seems to be a general issue with enums referencing things declared later. Seems to work fine with two classes and `const`s though (https://psalm.dev/r/5d409f074e).

I can look into it, but probably not before monday.

AegirLeet:
@weirdan Is this still needed with https://github.com/vimeo/psalm/pull/10655?

weirdan:
No, that PR addresses the issue you were trying to fix in a more general way.

Issue #10626, opened 2024-01-31
Support PHP 8.3 Override attribute


edsrzf:
Psalm should support PHP 8.3's new `Override` attribute in two ways:

* [x] When a method has an `Override` attribute but does not override a parent method, raise an issue. (Any method with an `Override` attribute that would cause a fatal runtime error should also be detected statically by Psalm.) Call it `InvalidOverride`. (Implemented in #10644.)
* [ ] When a method of a child class overrides that of a parent class and there is no `Override` attribute, raise an issue. Call it `MissingOverrideAttribute`.

Both of these would apply only when the PHP version is >= 8.3, and the second one should probably have its own configuration option.

I'm happy to work on the implementation and use this issue to track progress, discuss the ideas, and hash out any details, like the error names or configuration.

weirdan:
Don't we have that already? See #10493

weirdan:
> Maybe call it `InvalidOverrideAttribute`?

Just `InvalidOverride` would seem more fitting - as there's nothing wrong with the attribute itself.

> Maybe call it `MissingOverrideAttribute`?

Sounds good to me.

weirdan:
> Don't we have that already? See https://github.com/vimeo/psalm/pull/10493

That PR just adds the attribute stub.

PR #10644, opened 2024-02-03
New InvalidOverride issue for Override attribute

edsrzf:
`InvalidOverride` is a new issue emitted when the PHP 8.3 `Override` attribute is added to a method, but there is no matching method in a parent class/interface. This results in a fatal runtime error and only affects code that has opted in (by using the attribute in the first place), so I've added it to error level 7.

There are other cases where an `Override` attribute could cause a fatal runtime error due to mismatching signatures, but I believe all these cases are already covered by `MethodSignatureMismatch`. This happens regardless of whether or not the `Override` attribute is present and is already enabled at error level 7, so I don't think there are any further changes necessary.

This implements half of #10626.

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -283,6 +283,7 @@
             <xs:element name="InvalidNamedArgument" type="ArgumentIssueHandlerType" minOccurs="0" />
             <xs:element name="InvalidNullableReturnType" type="IssueHandlerType" minOccurs="0" />
             <xs:element name="InvalidOperand" type="IssueHandlerType" minOccurs="0" />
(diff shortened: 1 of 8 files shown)


weirdan:
See docs on adding new issue types: https://psalm.dev/docs/contributing/adding_issues/ - it's a somewhat involved process.

edsrzf:
Thanks, I'd missed a couple of those steps. I think I've got them all now.

weirdan (approved):

weirdan:
Thanks!

PR #10645, opened 2024-02-03
Update `.gitattributes`

drupol:
Hello there,

When downloading a snapshot archive from Github (example: https://github.com/vimeo/psalm/archive/refs/heads/5.x.zip) based on a specific commit ID, branch or tag for example, it is not possible to build PSalm because the `/examples` directory is missing, and it is required in `composer.json` (https://github.com/vimeo/psalm/blob/5.x/composer.json#L100).

This PR update the `.gitattributes` file so it doesn't get stripped out by Github.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,7 +10,6 @@
 /psalm-baseline.xml export-ignore
 /dictionaries/scripts export-ignore
 /.editorconfig export-ignore
-/examples export-ignore
 /.gitattributes export-ignore
 /.github export-ignore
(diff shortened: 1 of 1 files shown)


weirdan:
Do you mean the phar build?

drupol:
Yes,

The PHAR build, or just a regular build. Since it is mentioned in the `composer.json`, it shouldn't be stripped out.

drupol:
Closed in favor of #10647

Issue #10573, opened 2024-01-18
@psalm-check-type prepends current namespace to built-in types


vudaltsov:
https://psalm.dev/r/f45e4c4a68

PR #10648, opened 2024-02-03
Fix check-type when using builtin types from within a namespace

robchett:
Fix for #10573 

When looking for a better alternative to `Type::getFQCLNFromString` I found `TypeTokenizer::getFullyQualifiedTokens` also supports types defined with `psalm-type` so I've supported those too.


weirdan (approved):

weirdan:
Note that github does not recognize 'fix for #issuenumber' as a closing keyword, but it recognizes a number of other: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #10651, opened 2024-02-04
New MissingOverrideAttribute issue

### LINKED ISSUE #10626: Support PHP 8.3 Override attribute

edsrzf:
`MissingOverrideAttribute` is a new issue emitted when a method overrides a parent method, but does not have the PHP 8.3 `Override` attribute. This issue is only emitted under PHP 8.3 when the Psalm config flag `ensureOverrideAttribute` is turned on.

Fixes #10626.

### DIFF
--- a/tests/OverrideTest.php
+++ b/tests/OverrideTest.php
@@ -6,8 +6,6 @@ use Psalm\Config;
 use Psalm\Tests\Traits\InvalidCodeAnalysisTestTrait;
 use Psalm\Tests\Traits\ValidCodeAnalysisTestTrait;
 
-use const DIRECTORY_SEPARATOR;
-
 class OverrideTest extends TestCase
 {
     use ValidCodeAnalysisTestTrait;
@@ -83,7 +81,7 @@ class OverrideTest extends TestCase
                         public function f(): void {}
                     }
                 ',
-                'error_message' => 'InvalidOverride - src' . DIRECTORY_SEPARATOR . 'somefile.php:3:25',
+                'error_message' => 'InvalidOverride',
                 'error_levels' => [],
                 'php_version' => '8.3',
             ],
@@ -97,7 +95,7 @@ class OverrideTest extends TestCase
                         public function f(): void {}
                     }
                 ',
-                'error_message' => 'MissingOverrideAttribute - src' . DIRECTORY_SEPARATOR . 'somefile.php:7:25',
+                'error_message' => 'MissingOverrideAttribute',
                 'error_levels' => [],
                 'php_version' => '8.3',
             ],
@@ -113,7 +111,7 @@ class OverrideTest extends TestCase
                         public function f(): void {}
                     }
                 ',
-                'error_message' => 'MissingOverrideAttribute - src' . DIRECTORY_SEPARATOR . 'somefile.php:9:25',
+                'error_message' => 'MissingOverrideAttribute',
                 'error_levels' => [],
                 'php_version' => '8.3',
             ],
@@ -131,7 +129,7 @@ class OverrideTest extends TestCase
                         public function __construct() {}
                     }
                 ',
-                'error_message' => 'InvalidOverride - src' . DIRECTORY_SEPARATOR . 'somefile.php:10:25',
+                'error_message' => 'InvalidOverride',
                 'error_levels' => [],
                 'php_version' => '8.3',
             ],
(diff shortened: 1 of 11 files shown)


tests/OverrideTest.php
```diff
@@ -0,0 +1,180 @@
...
+    {
+        return [
+            'noParent' => [
+                'code' => '<?php
+                    class C {
+                        #[Override]
+                        public function f(): void {}
+                    }
+                ',
+                'error_message' => 'InvalidOverride - src' . DIRECTORY_SEPARATOR . 'somefile.php:3:25',
```
weirdan, line 86: The `error_message` is used for substring match, so you don't have to spell out the entire message here. 

```suggestion
                'error_message' => 'InvalidOverride',
```

tests/OverrideTest.php
```diff
@@ -0,0 +1,180 @@
...
+                'code' => '<?php
+                    class C {
+                        public function f(): void {}
+                    }
+
+                    class C2 extends C {
+                        public function f(): void {}
+                    }
+                ',
+                'error_message' => 'MissingOverrideAttribute - src' . DIRECTORY_SEPARATOR . 'somefile.php:7:25',
```
weirdan, line 100:
```suggestion
                'error_message' => 'MissingOverrideAttribute',
```

tests/OverrideTest.php
```diff
@@ -0,0 +1,180 @@
...
+                        abstract public function f(): void;
+                    }
+
+                    class C {
+                        use T;
+
+                        public function f(): void {}
+                    }
+                ',
+                'error_message' => 'MissingOverrideAttribute - src' . DIRECTORY_SEPARATOR . 'somefile.php:9:25',
```
weirdan, line 116:
```suggestion
                'error_message' => 'MissingOverrideAttribute',
```

tests/OverrideTest.php
```diff
@@ -0,0 +1,180 @@
...
+                    class C {
+                        public function __construct() {}
+                    }
+
+                    class C2 extends C {
+                        #[Override]
+                        public function __construct() {}
+                    }
+                ',
+                'error_message' => 'InvalidOverride - src' . DIRECTORY_SEPARATOR . 'somefile.php:10:25',
```
weirdan, line 134:
```suggestion
                'error_message' => 'InvalidOverride',
```

tests/OverrideTest.php
```diff
@@ -0,0 +1,180 @@
...
+                'code' => '<?php
+                    interface I {
+                        public function f(): void;
+                    }
+
+                    interface I2 extends I {
+                        public function f(): void;
+                    }
+                ',
+                'error_message' => 'MissingOverrideAttribute - src' . DIRECTORY_SEPARATOR . 'somefile.php:7:25',
```
weirdan, line 148:
```suggestion
                'error_message' => 'MissingOverrideAttribute',
```

tests/OverrideTest.php
```diff
@@ -0,0 +1,180 @@
...
+                    class C {
+                        private function f(): void {}
+                    }
+
+                    class C2 extends C {
+                        #[Override]
+                        private function f(): void {}
+                    }
+                ',
+                'error_message' => 'InvalidOverride - src' . DIRECTORY_SEPARATOR . 'somefile.php:7:25',
```
weirdan, line 163:
```suggestion
                'error_message' => 'InvalidOverride',
```

tests/OverrideTest.php
```diff
@@ -0,0 +1,180 @@
...
+                'php_version' => '8.3',
+            ],
+            'interfaceWithNoParent' => [
+                'code' => '<?php
+                    interface I {
+                        #[Override]
+                        public function f(): void;
+                    }
+                ',
+                'error_message' => 'InvalidOverride - src' . DIRECTORY_SEPARATOR . 'somefile.php:3:25',
```
weirdan, line 174:
```suggestion
                'error_message' => 'InvalidOverride',
```

weirdan (changes requested):

weirdan (approved):

tests/OverrideTest.php
edsrzf: I saw some other tests doing this. I thought maybe it would be useful to keep the line and column number to ensure that the code location is correct. I don't feel strongly about it, though, and am fine with the change you've made.

PR #10653, opened 2024-02-04
Allow nikic/php-parser v5

simPod:
`phpunit/php-code-coverage 11.0.0 requires nikic/php-parser ^5.0`


simPod:
This is too hard to support both v4 and v5

BacLuc:
A really nice workaround is using the package [psalm/phar](https://packagist.org/packages/psalm/phar) instead of vimeo/psalm.

The change is also small, see here: https://github.com/ecamp/ecamp3/pull/6256

PR #10665, opened 2024-02-06
Fix parsing magic method annotations

issidorov:
Fix parsing for code:

```
/**
 * @method static (string|int)[] getArray()
 * @method static (callable() : string) getCallable()
 */
class MyClass {}
```


weirdan (approved):

weirdan:
Thanks!

PR #10675, opened 2024-02-07
Prevent mixed|null

### LINKED ISSUE #5594: False-positive mixed nullable properties set in constructor promotion
https://psalm.dev/r/e60098079c

Psalm is requiring the impossible type `mixed|null` if an overloaded `mixed` property is set to `null` in constructor promotion.
The same does not occur if the property is defined as a traditional class property (B2)

Attempting to used mixed|null results in Fatal error: Type mixed can only be used as a standalone type

robchett:
fixes #5594


robchett:
Has changed this code from a `MixedArgument` error to a `PossiblyNullArgument`

https://psalm.dev/r/345e90d09b

weirdan (approved):

weirdan:
Thanks!

Issue #10669, opened 2024-02-06
Misparse of enums in list{}


bwoebi:
https://psalm.dev/r/04fcfb0955

While `@return A::X|A::Y` standalone works, putting it in a `list{}` will misinterpret the first part of the expression up to the first colon as array key of the list instead of recognizing the enum value.

danog:
Just wrap the parts: https://psalm.dev/r/61d2d0ec1c

bwoebi:
@danog That seems rather like a workaround than a fix? I see no particular parse conflict here, why it should not be possible to parse that?

danog:
Yeah that can be fixed!

weirdan:
The reason for not allowing constants in list keys seems to be somewhat technical. Our type AST for keyed arrays uses native php types (int/string) to represent shape keys (they become actual array kes of `TKeyedArray::$properties`) rather than Psalm types. Thus, we cannot differentiate between `A::B` and `"A::B"` - they both have the same string representation, and we cannot reliably resolve constant values during the scanning phase as referenced classes may not have been scanned yet. And we won't be ever able to use standalone constants (e.g. PHP_VERSION) in shape _keys_ using current syntax as it's ambiguous.  

But we should be able to support constants in list _value_ types by backtracking and reinterpreting the part being parsed, as a value type, the same way we can do it currently with `list{1}`.

PR #10678, opened 2024-02-07
Class consts in array shapes

weirdan:


src/Psalm/Internal/Type/TypeParser.php
```diff
@@ -1468,8 +1468,12 @@ private static function getTypeFromKeyedArrayTree(
                     if ($const_name === 'class') {
                         $property_key = $fq_classlike_name;
                         $class_string = true;
-                    } else {
+                    } elseif ($property_branch->value[0] === '"' || $property_branch->value[0] === "'") {
                         $property_key = $property_branch->value;
+                    } else {
+                        throw new TypeParseTreeException(
+                            ':: in array key is only allowed for ::class',
```
orklah, line 1475: that message should be updated now that we support multiple cases. Should we list what is not yet supported instead?

weirdan: No, we still do not support '::' in shape _keys_, even though we do support it now in shape _value_ types (see https://github.com/vimeo/psalm/issues/10669#issuecomment-1933287962), so the message is still correct.

weirdan: Also, see these passing tests below:
https://github.com/vimeo/psalm/pull/10678/files#diff-5e43e612fcbdb0af385d5f84675e4f7d04415e1c607c8d81873fb4914f26309dR364-R373

Issue #9649, opened 2023-04-13
`@psalm-this-out` doesn't work on `__construct()`


fluffycondor:
https://psalm.dev/r/2c2d54f466

**Expected**: no errors.
**Got**: `IfThisIsMismatch`.

As a result, now I have to typehint the object manually:
https://psalm.dev/r/6d5e4476ea

PR #10690, opened 2024-02-11
Process `@psalm-this-out` on `__construct()` as well

weirdan:
Fixes vimeo/psalm#9649


danog:
I personally think that a cleaner solution is to add support for default template values, I already implemented that in https://github.com/vimeo/psalm/pull/10577, you can copy-paste my implementation

weirdan:
image

PR #10695, opened 2024-02-11
Fix composer warnings

weirdan:
These two classes weren't autoloadable because their FQCNs didn't match
the file paths.


edsrzf:
Oops, sorry about that. Seems like something that CI could have caught somewhere?

orklah (approved):

weirdan:
`composer dump-autoload --strict-psr` should do the trick

weirdan:
> Oops, sorry about that. Seems like something that CI could have caught somewhere?

It kinda did (BCC was failing) but nobody bothered to check its output because the PR was targeting the master branch where BC breaks are allowed.

Issue #6985, opened 2021-11-24
Generator `send` type is not checked when traversing it with foreach


weirdan:
https://psalm.dev/r/11773a3883

I'd expect something along the lines of `Cannot iterate over a generator with non-nullable send() parameter type`.

PR #10697, opened 2024-02-12
Forbid iterating over generators with non-nullable `send()`

weirdan:
Fixes vimeo/psalm#6985

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLike/ReturnTypeAnalyzer.php
@@ -307,15 +307,6 @@ final class ReturnTypeAnalyzer
             $source->getParentFQCLN(),
         );
 
-        // hack until we have proper yield type collection
-        if ($function_like_storage
-            && $function_like_storage->has_yield
-            && !$inferred_yield_type
-            && !$inferred_return_type->isVoid()
-        ) {
-            $inferred_return_type = new Union([new TNamedObject('Generator')]);
-        }
-
         if ($is_to_string) {
             $union_comparison_results = new TypeComparisonResult();
             if (!$inferred_return_type->hasMixed() &&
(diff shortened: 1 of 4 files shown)


weirdan:
Some failures look valid, including the test-with-real-projects/PSL, as declaring `TSend` as `void` seems wrong. `TSend` is used as both the `send()` parameter type _and_ the `yield` return type, and `void` as a parameter type is rather misleading. Also, as described in the linked issue, `foreach()` invokes `send(null)` during iteration, so a generator with `TSend=void` is rather useless - it cannot be iterated over. 

/cc: @azjezz

weirdan:
The only remaining failures are for `Generator<..., ..., void, ...>` - we have to decide whether we want to allow that (and treat it as `null`) or reject it (see above).

orklah:
I may be wrong on that, but I believe Psalter itself used to (or still) generate signature with void generators like that

Anyway, maybe we could mimic the behavior of phpstan for this?

I don't see the need of creating one more difference for such a niche case.

weirdan:
> Anyway, maybe we could mimic the behavior of phpstan for this?

PHPStan doesn't flag iteration over `Generator<..., ..., void, ...>`, but not because it special cases `void`, but because it's not aware that `foreach` invokes `send(null)`: https://phpstan.org/r/86c02395-e906-4493-a354-99a39d135323, https://3v4l.org/ONh6K

Just like Psalm before this pull request.

@ondrejmirtes WDYT?

My preferred approach, I think, would be to allow people to ~shoot themselves in the foot~ specify `Generator<..., ..., void, ...>` for the time being but then flag all iterations over this generator. Then in a couple of years start flagging `Generator<..., ..., void, ...>` just like we flag `@param void` now.

weirdan:
I'm generally satisfied with how it works now.

@veewee heads up, this will break PSL because `Awaitable::iterate()` produces a generator that nominally cannot be iterated over:
https://github.com/azjezz/psl/blob/b85e5e1c8e2385bf108bb3a09fb36b459dbc1a07/src/Psl/Async/Awaitable.php#L52-L54

veewee:
Thanks for the heads-up. It should be fixed in https://github.com/azjezz/psl/pull/444

PR #10704, opened 2024-02-13
Fix detecting magic static methods

issidorov:
Key fixes for static magic methods:
- behavior with `@psalm-seal-methods` and related...
- detection the issue "InvalidStaticInvocation"
- detection the issue "NonStaticSelfCall"

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodVisibilityAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodVisibilityAnalyzer.php
@@ -40,6 +40,8 @@ final class MethodVisibilityAnalyzer
         $fq_classlike_name = $method_id->fq_class_name;
         $method_name = $method_id->method_name;
 
+        $with_pseudo = true;
+
         if ($codebase_methods->visibility_provider->has($fq_classlike_name)) {
             $method_visible = $codebase_methods->visibility_provider->isMethodVisible(
                 $source,
@@ -65,7 +67,7 @@ final class MethodVisibilityAnalyzer
             }
         }
 
-        $declaring_method_id = $codebase_methods->getDeclaringMethodId($method_id);
+        $declaring_method_id = $codebase_methods->getDeclaringMethodId($method_id, $with_pseudo);
 
         if (!$declaring_method_id) {
             if ($method_name === '__construct'
@@ -109,7 +111,7 @@ final class MethodVisibilityAnalyzer
             return null;
         }
 
-        $storage = $codebase->methods->getStorage($declaring_method_id);
+        $storage = $codebase->methods->getStorage($declaring_method_id, $with_pseudo);
         $visibility = $storage->visibility;
 
         if ($appearing_method_name
--- a/tests/MagicMethodAnnotationTest.php
+++ b/tests/MagicMethodAnnotationTest.php
@@ -46,6 +46,35 @@ class MagicMethodAnnotationTest extends TestCase
         $this->analyzeFile('somefile.php', new Context());
     }
 
+    public function testPhpDocMethodWhenUndefinedWithStatic(): void
+    {
+        Config::getInstance()->use_phpdoc_method_without_magic_or_parent = true;
+
+        $this->addFile(
+            'somefile.php',
+            '<?php
+                /**
+                 * @method static string getString()
+                 * @method static void setInteger(int $integer)
+                 * @method static mixed setString(int $integer)
+                 * @method static bool getBool(string $foo)
+                 * @method static (string|int)[] getArray()
+                 * @method static (callable() : string) getCallable()
+                 */
+                class Child {}
+
+                $a = Child::getString();
+                Child::setInteger(4);
+                /** @psalm-suppress MixedAssignment */
+                $b = Child::setString(5);
+                $c = Child::getBool("hello");
+                $d = Child::getArray();
+                $e = Child::getCallable();',
+        );
+
+        $this->analyzeFile('somefile.php', new Context());
+    }
+
     public function testPhpDocMethodWhenTemplated(): void
     {
         Config::getInstance()->use_phpdoc_method_without_magic_or_parent = true;
@@ -92,6 +121,28 @@ class MagicMethodAnnotationTest extends TestCase
                 $child = new Child();
 
                 $child->getString();',
+        );
+
+        $context = new Context();
+
+        $this->analyzeFile('somefile.php', $context);
+    }
+
+    public function testAnnotationWithoutCallConfigWithStatic(): void
+    {
+        $this->expectExceptionMessage('UndefinedMethod');
+        $this->expectException(CodeException::class);
+        Config::getInstance()->use_phpdoc_method_without_magic_or_parent = false;
+
+        $this->addFile(
+            'somefile.php',
+            '<?php
+                /**
+                 * @method static string getString()
+                 */
+                class Child {}
+
+                Child::getString();',
         );
 
         $context = new Context();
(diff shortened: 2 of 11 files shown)


tests/MagicMethodAnnotationTest.php
```diff
@@ -817,6 +960,7 @@ class A {}
                     class B extends A {}
                     function consumeInt(int $i): void {}
 
+                    /** @psalm-suppress UndefinedMethod */
                     consumeInt(B::bar());',
```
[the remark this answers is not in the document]
weirdan: I'd suggest adding another test:

```php
[
   'code' => <<<'PHP'
   <?php
   /** @method static int bar() */
   class A {}
   class B extends A {}

   /** @psalm-suppress UndefinedMethod */
   $a = B::bar(123, "whatever");
   PHP,
   'assertions' => [
       '$a===' => 'mixed',
   ],
],
```
and one more:
```php
[
   'code' => <<<'PHP'
   <?php
   /** @method static int bar() */
   class A {}
   class B extends A {}

   /** @psalm-suppress UndefinedMethod */
   $a = B::bar(function_does_not_exist(123));
   PHP,
   'error_message' => 'UndefinedFunction',
],
```

This will make sure that ignoring the method also ignores its parameters / return value, but the statements inside the argument list are still validated (thus `UndefinedFunction` in the second case)

tests/Traits/InvalidCodeAnalysisWithIssuesTestTrait.php
```diff
@@ -0,0 +1,112 @@
...
+ * "CodeException":
+ *
+ * ```
+ * $codebase->config->throw_exception = true; // or false
+ * ```
+ *
+ * If "throw_exception" is set to `true`, code execution stops.
+ *
+ * If "throw_exception" is set to `false`, the code may continue
+ * to execute, and an error may potentially occur.
```
weirdan, line 33: You mean a situation like:
1. we emitted an issue
2. then continued executing
3. and crashed with, e.g. `InvalidArgumentException` down the line?

tests/TestCase.php
```diff
@@ -155,6 +158,31 @@ protected function getTestName(bool $withDataSet = true): string
...
+            foreach ($file_issues as $issue) {
+                $fullIssueMessage = $issue->type . ' - ' . $issue->file_name . ':' . $issue->line_from . ':' . $issue->column_from . ' - ' . $issue->message;
+                $issueMessages[] = $fullIssueMessage;
+                if (preg_match('/\b' . preg_quote($expected, '/') . '\b/', $fullIssueMessage)) {
+                    $res = true;
+                }
+            }
+        }
+        if (!$message) {
+            $message = "Failed asserting that issue with \"$expected\".";
```
weirdan, line 176:
```suggestion
            $message = "Failed asserting that issue with \"$expected\" was emitted.";
```

tests/TestCase.php
```diff
@@ -155,6 +158,31 @@ protected function getTestName(bool $withDataSet = true): string
...
+            if (count($issueMessages)) {
+                $message .= "\n" . 'Other exists issues:' . "\n  - " . implode("\n  - ", $issueMessages);
+            } else {
+                $message .= ' Issues is not exists.';
```
weirdan, line 180:
```suggestion
                $message .= "\n" . 'Other issues reported:' . "\n  - " . implode("\n  - ", $issueMessages);
            } else {
                $message .= ' No issues reported.';
```

tests/TestCase.php
```diff
@@ -155,6 +158,31 @@ protected function getTestName(bool $withDataSet = true): string
         return $this->getName($withDataSet);
     }
 
+    public static function assertHasIssueType(string $expected, string $message = ''): void
+    {
+        $issueMessages = [];
```
weirdan, line 163: Most of the codebase uses snake_case for variable/property names

weirdan (changes requested):

tests/Traits/InvalidCodeAnalysisWithIssuesTestTrait.php
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * This commit allows testing for the second case, when the value of
```
issidorov, line 35: > This commit allows

This is not commit. This is description.

tests/TestCase.php
issidorov: Thanks. I forgot about it. I'll fix it.

src/Psalm/Internal/Analyzer/Statements/Expression/Call/Method/MethodVisibilityAnalyzer.php
```diff
@@ -40,6 +40,8 @@ public static function analyze(
         $fq_classlike_name = $method_id->fq_class_name;
         $method_name = $method_id->method_name;
 
+        $with_pseudo = true;
```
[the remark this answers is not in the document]
issidorov: I've been thinking about it too.
This is a declaration of a local variable.
And it seemed to me more beautiful to do this at the beginning of the method, next to the declaration of other variables.

Does it need to be fixed? Please confirm.

tests/TestCase.php
```diff
@@ -155,6 +158,31 @@ protected function getTestName(bool $withDataSet = true): string
         return $this->getName($withDataSet);
     }
 
+    public static function assertHasIssueType(string $expected, string $message = ''): void
```
issidorov, line 161: Need rename from `assertHasIssueType` to `assertHasIssue`.

tests/MagicMethodAnnotationTest.php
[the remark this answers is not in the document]
issidorov: I agree, the test has become confusing.

It's probably better to fix this test. I think your version with `assertions` is better.
But I would remove the confusing code with `@psalm-suppress UndefinedMethod`, and replace it with `error_levels`.

```
[
    'code' => <<<'PHP'
        <?php
        /**
         * @method static int bar()
         */
        class A {}
        class B extends A {}

        $a = B::bar();
        PHP,
    'assertions' => [
        '$a===' => 'int',
    ],
    'error_levels' => [
        'UndefinedMethod',
    ],
],
```

@weirdan , what do you think about this version of the code?

tests/MagicMethodAnnotationTest.php
```diff
@@ -809,15 +952,41 @@ function consumeInt(int $i): void {}
...
+                    class B extends A {}
+
+                    $a = B::bar(123, "whatever");
+                    PHP,
+                'assertions' => [
+                    '$a===' => 'int',
+                ],
+                'ignored_issues' => [
+                    'UndefinedMethod',
+                    'TooManyArguments',
```
weirdan, line 988: I believe this is incorrect. If we claim the method does not exist, how do we know it has no parameters and returns int? Since `UndefinedMethod` is suppressed, it should accept any number of arguments and return mixed.

weirdan: In other words the test needs to be reverted to what I suggested and the code fixed to match the suggested test.

tests/Traits/InvalidCodeAnalysisWithIssuesTestTrait.php
weirdan, line 33:
```suggestion
 * When `throw_exception` is set to `true`, code execution stops once 
 * the first issue is emitted, thus it may mask any problems after 
 * that point.  
 *
 * When `throw_exception` is set to `false`, the code will continue 
 * to be executed and we can uncover some additional bugs. 
```

weirdan (changes requested):

tests/Traits/InvalidCodeAnalysisWithIssuesTestTrait.php
issidorov: Thanks!

weirdan (approved):

weirdan:
Thanks!

Issue #10707, opened 2024-02-15
5.22 crash on scan: Uncaught RuntimeException: PHP Error: Uninitialized string offset 0 


aszenz:
```
Uncaught RuntimeException: PHP Error: Uninitialized string offset 0 in /home/runner/work/app/vendor/vimeo/psalm/src/Psalm/Internal/Type/TypeTokenizer.php:380 for command with CLI args "/home/runner/work/app/vendor/bin/psalm --output-format=github --no-cache" in /home/runner/work/app/vendor/vimeo/psalm/src/Psalm/Internal/ErrorHandler.php:75
Stack trace:
#0 /home/runner/work/app/vendor/vimeo/psalm/src/Psalm/Internal/Type/TypeTokenizer.php(380): Psalm\Internal\ErrorHandler::Psalm\Internal\{closure}()
#1 /home/runner/work/app/vendor/vimeo/psalm/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php(1964): Psalm\Internal\Type\TypeTokenizer::getFullyQualifiedTokens()
#2 /home/runner/work/app/vendor/vimeo/psalm/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php(1894): Psalm\Internal\PhpVisitor\Reflector\ClassLikeNodeScanner::getTypeAliasesFromCommentLines()
#3 /home/runner/work/app/vendor/vimeo/psalm/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeNodeScanner.php(386): Psalm\Internal\PhpVisitor\Reflector\ClassLikeNodeScanner::getTypeAliasesFromComment()
#4 /home/runner/work/app/vendor/vimeo/psalm/src/Psalm/Internal/PhpVisitor/ReflectorVisitor.php(155): Psalm\Internal\PhpVisitor\Reflector\ClassLikeNodeScanner->start()
#5 /home/runner/work/app/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(200): Psalm\Internal\PhpVisitor\ReflectorVisitor->enterNode()
#6 /home/runner/work/app/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(114): PhpParser\NodeTraverser->traverseArray()
#7 /home/runner/work/app/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(223): PhpParser\NodeTraverser->traverseNode()
[...]

PR #10711, opened 2024-02-15
Improve parsing of psalm-type

danog:


weirdan:
@danog can you also take a look at #10707 - I have a gut feeling the fix should be implemented in the same place.

weirdan (approved):
See #10713 which includes all commits from this PR and fixes the linked issue.

weirdan:
See #

Issue #9782, opened 2023-05-13
Loop: uppercase literal strings inferred as non-empty-lowercase-string when combined with numeric-strings


kkmuffme:
https://psalm.dev/r/5dda068487

When a string is concat with int/numeric-string, psalm discards any literal non-lowercase-strings and just assumes they're all lowercase.

When using "normal" (non-numeric) strings it works correctly though (as the lower example shows)

PR #10722, opened 2024-02-19
Fix non-empty-lowercase-string handling with literal non-lowercase strings

kkmuffme:
* Fix https://github.com/vimeo/psalm/issues/9782 and related issues
* add explicit handling for non-falsy-string to not fallback non-falsy-string and 0 to string
* revert https://github.com/vimeo/psalm/pull/10039 to make it more exact and fix the test to be exact @danog @orklah


danog (approved):

weirdan:
Thanks!

Issue #10350, opened 2023-11-06
Internal Psalm Error: `Could not get class storage for <interface>`


ohader:
### Error Message using latest release [`vimeo/psalm:5.15.0`](https://github.com/vimeo/psalm/releases/tag/5.15.0)

> Uncaught InvalidArgumentException: Could not get class storage for sessionhandlerinterface in .../vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:45

This scenario occurred while analyzing [`\Symfony\Component\HttpFoundation\Session\Storage\Handler\MigratingSessionHandler`](https://github.com/symfony/http-foundation/blob/6.3/Session/Storage/Handler/MigratingSessionHandler.php#L25-L26). 

I've created a simplified version at https://psalm.dev/r/4c29e2bae8

The problem seems to be related to using intersection types in class property declarations. Resolving that results in the mentioned `Could not get class storage for <interface>` error.

```php
class Subject implements \JsonSerializable
{
    // @hint Demo: Remove the intersection type here to see how the error disappears
    // `Internal Psalm error on line 47 - /vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php: Could not get class storage for jsonserializable`
    private \JsonSerializable&\Stringable $property;
...
```

### Relevant stack trace

→ I'm guessing, that the [`TypeHintResolver` is causing this when resolving intersection types](https://github.com/vimeo/psalm/blob/5.15.0/src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php#L112)

```
#0 .../vimeo/psalm/src/Psalm/Internal/Codebase/ClassLikes.php(723): Psalm\Internal\Provider\ClassLikeStorageProvider->get('sessionhandleri...')
#1 .../vimeo/psalm/src/Psalm/Internal/Codebase/ClassLikes.php(713): Psalm\Internal\Codebase\ClassLikes->getParentInterfaces('sessionhandleri...')
#2 .../vimeo/psalm/src/Psalm/Codebase.php(761): Psalm\Internal\Codebase\ClassLikes->interfaceExtends('sessionhandleri...', 'sessionupdateti...')
#3 .../vimeo/psalm/src/Psalm/Internal/Type/Comparator/ObjectComparator.php(358): Psalm\Codebase->interfaceExtends('sessionhandleri...', 'sessionupdateti...')
[...]

Issue #10152, opened 2023-08-26
Could not get class storage for redis


thbley:
Psalm 5.15 crashes with:

```
public Redis&App $redis;
```

but succeeds with:
```
private Redis $redis;
```
https://psalm.dev/r/8eca4012df
https://psalm.dev/r/8c20742f57

```
Target PHP version: 8.2 (inferred from current PHP version) Enabled extensions: pdo, redis (unsupported extensions: pdo_pgsql, spl, pcre, zend-opcache).
Scanning files...
Uncaught InvalidArgumentException: Could not get class storage for redis in /var/www/tests/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:45
Stack trace:
#0 /var/www/tests/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/ClassLikes.php(613): Psalm\Internal\Provider\ClassLikeStorageProvider->get('redis')
#1 /var/www/tests/vendor/vimeo/psalm/src/Psalm/Codebase.php(705): Psalm\Internal\Codebase\ClassLikes->classExtends('redis', 'pdo')
#2 /var/www/tests/vendor/vimeo/psalm/src/Psalm/Internal/Type/Comparator/ObjectComparator.php(340): Psalm\Codebase->classExtendsOrImplements('redis', 'pdo')
#3 /var/www/tests/vendor/vimeo/psalm/src/Psalm/Internal/Type/Comparator/ObjectComparator.php(117): Psalm\Internal\Type\Comparator\ObjectComparator::isIntersectionShallowlyContainedBy(Object(Psalm\Codebase), Object(Psalm\Type\Atomic\TNamedObject), Object(Psalm\Type\Atomic\TNamedObject), 'pdo', false, false, NULL)
#4 /var/www/tests/vendor/vimeo/psalm/src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php(354): Psalm\Internal\Type\Comparator\ObjectComparator::isShallowlyContainedBy(Object(Psalm\Codebase), Object(Psalm\Type\Atomic\TNamedObject), Object(Psalm\Type\Atomic\TNamedObject), false, NULL)
[...]

Issue #7520, opened 2022-01-29
Intersection types with built-in classes does not work


davidbyoung:
https://psalm.dev/r/07aeb2b8c9

AndrolGenhald:
It shouldn't crash like that, but intersecting two classes doesn't make much sense anyway. Either we already know one class extends the other, or we know they can't possibly intersect. It seems to work correctly for interfaces: https://psalm.dev/r/1b136de642.

davidbyoung:
My real use-case is intersecting an interface and `MockObject` in a unit test.  In this case, my interface was `SessionHandlerInterface`.  Here's a repro of that issue:  https://psalm.dev/r/f544acf7cc

davidbyoung:
(it still breaks if you make `Foo` an interface in my above example)

davidbyoung:
The order of the type seems to matter.  For example, [this works](https://psalm.dev/r/09f17deb82), but [this does not](https://psalm.dev/r/e9af3116a7).

orklah:
ouch, this breaks badly. I'll look into it.

I'm not sure what will happen by intersecting a mock object and an interface though... Either the mock object already implement the interface and it should already work or it doesn't and it's weird

AndrolGenhald:
> Either the mock object already implement the interface and it should already work or it doesn't and it's weird

Intersecting objects and interfaces in makes sense when the class is non-final, since children can implement interfaces.

davidbyoung:
Fwiw [PHPUnit returns an intersection type](https://github.com/sebastianbergmann/phpunit/blob/b32796b9c4c19bdd7b8080bc36bbf3182c7cb3f6/src/Framework/TestCase.php#L1322) when creating a mock

orklah:
Thanks for feedback :)

So far, I found out that the issue happens because of Psalm didn't manage to scan SessionHandlerInterface before it tries to create the intersectionType (for that, Psalm needs to know the class storage). 

I'm still digging

AndrolGenhald:
@orklah Given that [docblock types](https://psalm.dev/r/39c247533a) work I'd take a look at the PR that added support for intersection signature types, maybe there's something happening during scanning that needs to happen during analysis?

orklah:
That's highly possible, but I'm confused by the fact it works when you flip the intersection around. If I solve the mystery here and the answer is not right, I'll try to fix it further

AndrolGenhald:
[...]

PR #10720, opened 2024-02-19
Catch missing classlike exceptions during scanning

weirdan:
This may resolve #10706, although I'm not exactly convinced it's the
best way.

Fixes: #7520 
Fixes: #10350
Fixes: #10152

PR #10724, opened 2024-02-19
Reprocess missing class-like storage events

ohader:
The scenario described in #10350 fails, since not all other references used in intersection types were complete resolved when scanning a particular file. Instead of catching exceptions during the type resolution process - which would just "mute" the problem and drop important information - those failures in resolving class-like storages are tracked for a potential rescanning.

This PR does the following:

* introduces a new `ClassStorageNotFoundException` (instead of  global `InvalidArgumentException`) to handle missing class-like storage references specifically
* tracks those storage failures during the scanning process in `$files_to_rescan` for later processing
* adds bool return type to `\Psalm\Internal\Codebase\Scanner::scanAPath()` - whether scan was successful
* modifies `Codebase::scanFiles()` to actually reprocess those files to be rescanned - after changes of the previous run were populated to the codebase
* adds test cases for the known scenario of issue #10350 

---

Fixes: https://github.com/vimeo/psalm/issues/7520
Fixes: https://github.com/vimeo/psalm/issues/10350
Fixes: https://github.com/vimeo/psalm/issues/10152

---

### TODO

+ [ ] tackle `\Psalm\Internal\Codebase\Scanner::scanAPath()` return type in forked processes - currently only tested with `$pool_size = 1` in test cases

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -433,11 +433,16 @@ final class Codebase
      */
     public function scanFiles(int $threads = 1): void
     {
-        $has_changes = $this->scanner->scanFiles($this->classlikes, $threads);
-
-        if ($has_changes) {
-            $this->populator->populateCodebase();
-        }
+        $max = 10;
+        do {
+            $has_changes = $this->scanner->scanFiles($this->classlikes, $threads);
+            if ($has_changes) {
+                $this->populator->populateCodebase();
+            }
+            // in case previous scans failed, let's try again after previously found
+            // references (class-like storage items and dependencies) were populated
+            $shall_rescan = $this->scanner->prepareRescanFiles();
+        } while ($max-- > 0 && $shall_rescan);
     }
 
     public function getFileContents(string $file_path): string
--- a/src/Psalm/Exception/ClassStorageNotFoundException.php
+++ b/src/Psalm/Exception/ClassStorageNotFoundException.php
@@ -5,22 +5,19 @@ declare(strict_types=1);
 namespace Psalm\Exception;
 
 use InvalidArgumentException;
+use Throwable;
 
 /**
  * To be thrown when a `\Psalm\Storage\ClassLikeStorage` for a given name could not be resolved.
  */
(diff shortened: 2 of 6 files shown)


ohader:
Alternative approach to PR #10720

weirdan:
> introduces a new ClassStorageNotFoundException (instead of global InvalidArgumentException) to handle missing class-like storage references specifically

It was literally my next step :grin: :

image

ohader:
> I wonder if rescanning whole files is really necessary. My idea was to walk classlikes / functionlikes after the codebase population and tighten the intersections. E.g. if the return method was left as `MySessionHandler&SessionHandlerInterface` where `MySessionHandler` implements `SessionHandlerInterface` (because either `MySessionHandler` or `SessionHandlerInterface` were unavailable during the initial scan) it would be tightened to `MySessionHandler`.

I had something similar in my mind as well, but did not find a good my to "communicate" from the specific inner parts to the outer parts that would trigger those optimizations after codebase population. Basically like:

* hold unresolved/failed statements/names in memory (probably required for anonymous classes)
* process different strategies to resolve the types (e.g. simplifying the types like `ArrayObject&Countable` → `ArrayObject`; if that fails, reprocess the parser statements; if that fails reprocess the whole file) - this might be an additional setting in `psalm.xml` config file
* in case not all previously unresolved statements/names could be resolved, the process could either fail with an exception or continue with a generic type, e.g. `TMixed` - this might be another config option in `psalm.xml`

simonberger:
I would have tested if this attempt makes a difference for the crash I still have, but unfortunately you branched it from psalm 6 instead of 5 which I not yet are able to install.

ohader:
> I would have tested if this attempt makes a difference for the crash I still have, but unfortunately you branched it from psalm 6 instead of 5 which I not yet are able to install.

Uh, sorry... I usually start working in `main` (`master`) and the back-port to stable branches. I forgot it's different here...
Let's close this PR (I can take care of the forward port later, once we found a good foundation code-wise).

→ `5.x` PR at https://github.com/vimeo/psalm/pull/10730

Issue #10725, opened 2024-02-19
Taint Graph not working anymore using generic class @template


ohader:
https://psalm.dev/r/4d62e72f27

* worked in 5.12.0
* fails since 5.13.0 (incl. current 6.x master branch)

https://github.com/vimeo/psalm/compare/5.12.0...5.13.0

update: required an additional custom plugin, that was modifying parser statements

ohader:
→ bisected down to da8c1da8be2db41cd10054062676719e275c5238

ohader:
While trying to isolate the problem further, it turned out that the (good) taint analysis stems from a custom plugin - not from the core functionality of PsalmPHP.

The plugin used the `BeforeFileAnalysisEvent` to modify parser statements before actually being analyzed.
By dropping the in-memory cache in commit da8c1da8be2db41cd10054062676719e275c5238 those statement modifications are lost...

I still need to look, how this could be done differently...

ohader:
I just want to share what was going on in my custom scenario, having a custom plugin.

https://github.com/vimeo/psalm/blob/5.22.1/src/Psalm/Internal/Analyzer/FileAnalyzer.php#L147-L155

* `BeforeFileAnalysisEvent` was used to modify the statements by fetching and modifying them in a consecutive call to `$codebase->getStatementsForFile($this->file_path)`
* this worked prior to commit da8c1da8be2db41cd10054062676719e275c5238, since the array of `list<PhpParser\Node\Stmt>` was served from the in-memory cache `StatementsVolatileCache`
* my approach was basically a work-around which worked well (back then)

In order to make this an official feature, the event would have to pass along those `Stmt` nodes explicitly and fetch them again in the mentioned location in `FileAnalyzer`, e.g.

```php
class BeforeFileAnalysisEvent
{
    // ...
    /**
     * @return list<PhpParser\Node\Stmt>
     */
    public function getStmts(): array {
        return $this->stmts;
    }
    // ...
}

PR #10728, opened 2024-02-19
Add list of statements to BeforeFileAnalysisEvent

ohader:
Fixes: #10725

### DIFF
--- a/src/Psalm/Internal/Analyzer/FileAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FileAnalyzer.php
@@ -150,9 +150,9 @@ class FileAnalyzer extends SourceAnalyzer
             return;
         }
 
-        $event = new BeforeFileAnalysisEvent($this, $this->context, $file_storage, $codebase);
-
+        $event = new BeforeFileAnalysisEvent($this, $this->context, $file_storage, $codebase, $stmts);
         $codebase->config->eventDispatcher->dispatchBeforeFileAnalysis($event);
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/FileAnalyzer.php
```diff
@@ -150,9 +150,9 @@ public function analyze(
             return;
         }
 
-        $event = new BeforeFileAnalysisEvent($this, $this->context, $file_storage, $codebase);
-
+        $event = new BeforeFileAnalysisEvent($this, $this->context, $file_storage, $codebase, $stmts);
         $codebase->config->eventDispatcher->dispatchBeforeFileAnalysis($event);
+        $stmts = $event->getStmts();
```
weirdan, line 155: This implies the `$stmts` array may be changed by the handler, but there's no method on BeforeAnalysisEvent to set the updated array of statements.

Note that plugins may still change the statement objects in this array - there's no need to fetch statements from the event to support that.

ohader: I'd be fine adding a `setStmts(array $stmts)` method to the event as well - or remove the line above from `FileAnalyzer` (since the references of inside that array are kept).

My feeling is, that dropping the line from above and not introducing `setStmts(array $stmts)` has less side effects.

weirdan (approved):

weirdan:
Thanks!

Issue #10521, opened 2024-01-01
Taint should be ignored on int cast


mmcev106:
@cgocast, I believe [commit #94a98c](https://github.com/vimeo/psalm/commit/94a98ccddd68445665813c4989b5d75d555695ed) may need some follow-up work.  It's causing dozens of false positives for us, which I have simplified to the following example:

https://psalm.dev/r/c775928c77

I don't believe appending a hard coded string to an untainted integer value should cause it to become tainted.  I think the same incorrect logic might also apply to `taintedInputAfterIntCast`, as prepending `"aaaa"` is something the developer has explicitly chosen to do (not a taint).  Am I missing something?

orklah:
Well, I do agree that both of your examples should behave the same way.

However, I'm not certain which way it should go.

The relevant issue/discussion we had is here: https://github.com/vimeo/psalm/issues/10238

The example snippet is indeed bad. you could also alter it to show that a LIMIT $i could be dangerous in some cases and there are probably many cases I don't think of.

TaintedHTML (and more widely, any echo/print in the code) is more complex though, it doesn't deal with performance critical systems and if it's designed to be returned to the client, it's mostly safe.
However, you could be using html to generate a quotation PDF and the client managed to push 0 for $i in `"Total: $".$i` so this is a legit taint I believe
You could also be echoing something that you retrieve through ob_ functions to do something critical, it's hard to tell

Contrary to the "normal" analysis of Psalm, which aims to reduce false positive to an acceptable compromise, we should probably aim for security based analysis to reduce false negative to a minimum because people may rely on it too much and create a false sense of security

mmcev106:
@orklah, I'm not sure I understand how an integer value could be considered tainted, even for SQL.  Tainted SQL becomes problematic when there is risk of SQL injection, which is not an issue for integers.  It is a valid use case for the application to allow any integer to be passed from user input to a SQL query, and up to the application to determine if permissions allow that action on that integer ID.  You can pass any integer with prepared statements as well (which of course does not and should not report a taint).  I think the logic in #10238 that lead up to this change might be flawed...  Am I missing something?

orklah:
> Tainted SQL becomes problematic when there is risk of SQL injection

[...]

PR #10242, opened 2023-10-02
Allow tainted numerics except for 'html' and 'has_quotes'

cgocast:
Fix https://github.com/vimeo/psalm/issues/10238

PR #10729, opened 2024-02-20
Revert "Allow tainted numerics except for 'html' and 'has_quotes'"

ohader:
see comments and details at https://github.com/vimeo/psalm/issues/10521#issuecomment-1954151988

Fixes: vimeo/psalm#10521
Reverts: vimeo/psalm#10242


orklah (approved):
While I do consider the change legit, it's true that it have inconsistencies and it may have surprised people. It could come back  as an opt in or with another issue type once the issue is fixed

weirdan:
Thanks!

Issue #10706, opened 2024-02-14
5.22.0 regression: Could not get class storage for psl\range\upperboundrangeinterface


zerkms:
It was fine in 5.21.1

```
Target PHP version: 8.3 (inferred from current PHP version).
Scanning files...
Uncaught InvalidArgumentException: Could not get class storage for psl\range\upperboundrangeinterface in /app-appname/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:45
Stack trace:
#0 /app-appname/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/ClassLikes.php(726): Psalm\Internal\Provider\ClassLikeStorageProvider->get()
#1 /app-appname/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/ClassLikes.php(716): Psalm\Internal\Codebase\ClassLikes->getParentInterfaces()
#2 /app-appname/vendor/vimeo/psalm/src/Psalm/Codebase.php(763): Psalm\Internal\Codebase\ClassLikes->interfaceExtends()
#3 /app-appname/vendor/vimeo/psalm/src/Psalm/Internal/Type/Comparator/ObjectComparator.php(357): Psalm\Codebase->interfaceExtends()
#4 /app-appname/vendor/vimeo/psalm/src/Psalm/Internal/Type/Comparator/ObjectComparator.php(117): Psalm\Internal\Type\Comparator\ObjectComparator::isIntersectionShallowlyContainedBy()
#5 /app-appname/vendor/vimeo/psalm/src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php(350): Psalm\Internal\Type\Comparator\ObjectComparator::isShallowlyContainedBy()
#6 /app-appname/vendor/vimeo/psalm/src/Psalm/Type.php(891): Psalm\Internal\Type\Comparator\AtomicTypeComparator::isContainedBy()
#7 /app-appname/vendor/vimeo/psalm/src/Psalm/Type.php(768): Psalm\Type::intersectAtomicTypes()
#8 /app-appname/vendor/vimeo/psalm/src/Psalm/Internal/PhpVisitor/Reflector/TypeHintResolver.php(112): Psalm\Type::intersectUnionTypes()
[...]

PR #10730, opened 2024-02-20
Avoid exceptions on fetching uninitialized names in ClassLikes

ohader:
Similar to the behavior of `ClassLikes::classImplements()` (#5984),
the following methods will return a false/empty value in case a
specific class name has not been initialized yet:

+ `ClassLikes::classExtends()` returns `false`
+ `ClassLikes::getParentInterfaces()` returns `[]`

---

Fixes: https://github.com/vimeo/psalm/issues/7520
Fixes: https://github.com/vimeo/psalm/issues/10350
Fixes: https://github.com/vimeo/psalm/issues/10152

### DIFF
--- a/src/Psalm/Internal/Codebase/ClassLikes.php
+++ b/src/Psalm/Internal/Codebase/ClassLikes.php
@@ -613,6 +613,9 @@ final class ClassLikes
             return false;
         }
 
+        if (!$this->classlike_storage_provider->has($unaliased_fq_class_name)) {
+            return false;
+        }
         $class_storage = $this->classlike_storage_provider->get($unaliased_fq_class_name);
 
         if ($from_api && !$class_storage->populated) {
@@ -722,7 +725,9 @@ final class ClassLikes
     public function getParentInterfaces(string $fq_interface_name): array
     {
         $fq_interface_name = strtolower($fq_interface_name);
-
+        if (!$this->classlike_storage_provider->has($fq_interface_name)) {
+            return [];
+        }
         return $this->classlike_storage_provider->get($fq_interface_name)->parent_interfaces;
     }
 
--- a/src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/UnionTypeComparator.php
@@ -496,7 +496,7 @@ final class UnionTypeComparator
     /**
      * @return list<Atomic>
      */
-    private static function getTypeParts(
+    public static function getTypeParts(
         Codebase $codebase,
         Union $union_type
     ): array {
--- a/tests/Internal/Codebase/InternalCallMapHandlerTest.php
+++ b/tests/Internal/Codebase/InternalCallMapHandlerTest.php
@@ -16,6 +16,7 @@ use Psalm\Tests\Internal\Provider\FakeParserCacheProvider;
 use Psalm\Tests\TestCase;
 use Psalm\Tests\TestConfig;
 use Psalm\Type;
+use Psalm\Type\Atomic\TNamedObject;
 use ReflectionException;
 use ReflectionFunction;
 use ReflectionFunctionAbstract;
@@ -23,6 +24,8 @@ use ReflectionMethod;
 use ReflectionParameter;
 use ReflectionType;
 
+use function array_filter;
+use function array_map;
 use function array_shift;
 use function class_exists;
 use function count;
@@ -39,6 +42,7 @@ use function print_r;
 use function strcmp;
 use function strncmp;
 use function strpos;
+use function strtolower;
 use function substr;
 use function version_compare;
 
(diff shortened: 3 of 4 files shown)


ohader:
_Previous comment from `master` PR are still valid (https://github.com/vimeo/psalm/pull/10724#issuecomment-1953042355)_

> I wonder if rescanning whole files is really necessary. My idea was to walk classlikes / functionlikes after the codebase population and tighten the intersections. E.g. if the return method was left as `MySessionHandler&SessionHandlerInterface` where `MySessionHandler` implements `SessionHandlerInterface` (because either `MySessionHandler` or `SessionHandlerInterface` were unavailable during the initial scan) it would be tightened to `MySessionHandler`.

I had something similar in my mind as well, but did not find a good way to "communicate" from the specific inner parts to the outer parts that would trigger those optimizations after codebase population. Basically like:

* hold unresolved/failed statements/names in memory (probably required for anonymous classes)
* process different strategies to resolve the types (e.g. simplifying the types like `ArrayObject&Countable` → `ArrayObject`; if that fails, reprocess the parser statements; if that fails reprocess the whole file) - this might be an additional setting in `psalm.xml` config file
* in case not all previously unresolved statements/names could be resolved, the process could either fail with an exception or continue with a generic type, e.g. `TMixed` - this might be another config option in `psalm.xml`

simonberger:
I did test it. [While my crash pre-scanning the project files](https://github.com/vimeo/psalm/issues/10706#issuecomment-1952070106) does not happen anymore, I got a crash while scanning the project files.

```
Uncaught Exception: Psalm\Exception\UnpreparedAnalysisException File /var/www/project/src/SomeMockFile.php has not been properly scanned

Emitted in /var/www/project/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/FileAnalyzer.php:124
Stack trace in the forked worker:
#0 /var/www/project/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(1591): Psalm\Internal\Analyzer\FileAnalyzer->analyze()
#1 /var/www/project/vendor/vimeo/psalm/src/Psalm/Internal/Fork/Pool.php(191): Psalm\Internal\Codebase\Analyzer->analysisWorker()
#2 /var/www/project/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(372): Psalm\Internal\Fork\Pool->__construct()
#3 /var/www/project/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(272): Psalm\Internal\Codebase\Analyzer->doAnalysis()
#4 /var/www/project/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(552): Psalm\Internal\Codebase\Analyzer->analyzeFiles()

```

ohader:
> I did test it. [While my crash pre-scanning the project files](https://github.com/vimeo/psalm/issues/10706#issuecomment-1952070106) does not happen anymore, I got a crash while scanning the project files.
> 
> ```
> Uncaught Exception: Psalm\Exception\UnpreparedAnalysisException File /var/www/project/src/SomeMockFile.php has not been properly scanned
> 
> Emitted in /var/www/project/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/FileAnalyzer.php:124
> ```

It seems this is a new/different issue now. Can you try running Psalm single-threaded (`./psalm --threads=1 ...`)?
Just to make sure it's not the issue I referenced in the `TODO` section of this PR description. Thx!

ohader:
While working on additional tests for unknown/undefined interfaces used in intersections, I realized that the behavior is different. Thus the `\Stringable&\JsonSerializable` example failed since both the classes actually existed - when those classes are not existing, the process continues and silently ignores that lack of information.

Thus, it seems there might be a much simpler solution than reprocessing, which is more in line with the rest of the assumptions in PsalmPHP:
* add some more `$classlike_storage_provider->has($interface)` checks and we're basically good to go
* there are already plenty of these `try .. catch` blocks and `->has(..)` checks in `\Psalm\Internal\Codebase\ClassLikes`(https://github.com/vimeo/psalm/blob/5.x/src/Psalm/Internal/Codebase/ClassLikes.php) already
* I dumped the previous draft-code to a separate branch at https://github.com/ohader/psalm/tree/10350-intersections-outdated-5.x in case it gets relevant again

I've updated the PR accordingly...

ohader:
Test cases explicitly expect an exception to infer further type checks... which feels super weird:

https://github.com/vimeo/psalm/blob/1c36da6ddacddd1719b3e453d0e152230276065b/tests/Internal/Codebase/InternalCallMapHandlerTest.php#L649-L659

`ClassLikes::classExtends()` now returns `false` instead of throwing an exception by implicitly invoking `$this->classlike_storage_provider->get` and waiting for that `InvalidArgumentException`.

ohader:
I'll surrender now... still `static` keyword failures in the CallMap tests...

ohader:
→ Phew... I'm dropping this change in favor of the original PR at https://github.com/vimeo/psalm/pull/10720

PR #10733, opened 2024-02-20
Fix RiskyTruthyFalsyComparison irrelevant errors when there is no explicit truthy/falsy type

kkmuffme:
and update the error description.

e.g. `int|float` isn't an issue, only `int|float|false` would be risky since 0 and false may not implicate the same thing/require different handling.


weirdan (approved):

weirdan:
Thanks!

PR #10742, opened 2024-02-23
Add @psalm-no-discard annotation and report UnusedFunctionCall for @psalm-taint-escape calls

### LINKED ISSUE #11381: Add support for #[\NoDiscard]
Will ship with PHP 8.5 https://wiki.php.net/rfc/marking_return_value_as_important

There is a finished but unmerged PR for `@psalm-no-discard` https://github.com/vimeo/psalm/pull/10742 which is essentially the same thing for phpdoc, which could be extended (or merged and a separate PR) for this functionality to be implemented.

### LINKED ISSUE #11638: Add support for (void) typecast PHP 8.5
https://psalm.dev/r/5ab15137a3

### LINKED ISSUE #11639: Bad error message for deprecated (unset) cast
https://psalm.dev/r/cf6d00f0cd

kkmuffme:
By default psalm considers all functions as impure and UnusedFunctionCall errors are only emitted for pure functions, since pure functions have no side-effects and therefore their return value must be used (otherwise the function call is useless)

There are some cases, where you want to force people to use the return value of a function call even if it's not pure - e.g. for escaping functions or when upgrading legacy code bases where a function returned void/never previously, but now returns a value that must get validated.
Currently, there is no valid way to do that (you could add @psalm-pure and suppress that on the function, but that will lead to other issues)

I propose a new @psalm-no-discard to get this behavior.
~~(I'd prefer @psalm-require-use however I think that might lead to confusion with @use/@template-use/@psalm-use)~~

Additionally, anything annotated with @psalm-taint-escape will also report unused function calls by default, since the escape function call only makes sense if the returned value is used.

Since phpstan doesn't offer this functionality in the first place (since by default it treats everything as pure https://phpstan.org/writing-php-code/phpdocs-basics#impure-functions), there are no PHP eco-system big picture concerns either.

Another benefit of the annotation is that atm the attribute will result in a fatal https://github.com/php/php-src/issues/20879 if a class is autoloaded, but the class it extends is stubbed

---

Summary:
- fix @psalm-pure + @psalm-assert-if-true not reporting UnusedFunctionCalls on methods https://psalm.dev/r/7093ba0a35 (already worked for functions https://psalm.dev/r/afb33d970e)
- fix methods that return void should not be reporting UnusedMethodCall since the return value can't be used
- report unused function/method calls if the function is annotated with `@psalm-no-discard` or has a `@psalm-taint-escape` annotation
- add support for PHP 8.5 `#[NoDiscard]` to report unused return value. Fix https://github.com/vimeo/psalm/issues/11381
- add support for PHP 8.5 `(void)` typecase, as this is required to handle NoDiscard attribute. Fix https://github.com/vimeo/psalm/issues/11638
- fix bad error message and wrong handling for deprecated and removed `(unset)` typecast. Fix https://github.com/vimeo/psalm/issues/11639

### DIFF
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -535,7 +535,7 @@ $counter->count = 5; // This will fail, as it's mutating a property directly
 
 ### `@psalm-no-discard`
 
-Like PHP 8.5 `#[NoDiscard]` attribute
+Like PHP 8.5 `#[NoDiscard]` attribute, except that it can also be resolved with @psalm-suppress instead of using a `(void)` cast, since that is only available in PHP 8.5+. 
 
 ```php
 <?php
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -533,6 +533,25 @@ echo $counter->count; // outputs 1
 $counter->count = 5; // This will fail, as it's mutating a property directly
 ```
 
+### `@psalm-require-usage`
+
+Used to annotate an impure function or method the unused result of which Psalm should report.
+
+```php
+<?php
+/**
+ * @psalm-require-usage
+ *
+ * @return int
+ */
+function foo() {
+	return rand();
+}
+
+foo(); // reports UnusedFunctionCall
+echo foo(); // no error
+```
+
 ### `@psalm-trace`
 
 You can use this annotation to trace inferred type (applied to the *next* statement).
--- a/docs/annotating_code/supported_annotations.md
+++ b/docs/annotating_code/supported_annotations.md
@@ -533,14 +533,14 @@ echo $counter->count; // outputs 1
 $counter->count = 5; // This will fail, as it's mutating a property directly
 ```
 
-### `@psalm-require-usage`
+### `@psalm-no-discard`
 
-Used to annotate an impure function or method the unused result of which Psalm should report.
+Like PHP 8.5 `#[NoDiscard]` attribute
 
 ```php
 <?php
 /**
- * @psalm-require-usage
+ * @psalm-no-discard
  *
  * @return int
  */
--- a/docs/running_psalm/issues/ImpureFunctionCall.md
+++ b/docs/running_psalm/issues/ImpureFunctionCall.md
@@ -17,7 +17,7 @@ function impure(array $a) : array {
 }
 
 /** @psalm-pure */
-function filterOdd(array $a) : void {
-    impure($a);
+function filterOdd(array $a) : array {
+    return impure($a);
 }
 ```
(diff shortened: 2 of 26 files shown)


ohader:
This seems to be very useful! 👍 
We had to deal with similar patterns in 3rd party code analyzing security vulnerabilities - like visualized in this example:

https://psalm.dev/r/4daa849fa9

```php
/**
 * @psalm-require-usage
 */
class AccessDeniedResponse implements ResponseInterface {}
```

### Questions concerning this PR

* Would it quality to introduce a new `CodeIssue` class for that, e.g. `MissingRequiredUsage` instead of `UnusedFunctionCall`
* Would it be possible to extend `@psalm-require-usage` to class instances that need to be used? Thus, basically making it available for other concepts (classes, variables, ...) as well...

weirdan:
@ohader the thing you're describing looks much more complicated then what's proposed here. While a return value usage is fairly easy to define (any read operation), usage of an atomic in a union is much less clear. E.g. would all of the following constitute usage of `AccessDenied`?

https://psalm.dev/r/fcb87e545b

Also note that in all of those cases we don't really see `AccessDenied` in any return type, only its parent.

docs/annotating_code/supported_annotations.md
```diff
@@ -503,6 +503,25 @@ echo $counter->count; // outputs 1
 $counter->count = 5; // This will fail, as it's mutating a property directly
 ```
 
+### `@psalm-require-usage`
+
+Used to annotate a impure functions, which Psalm should require the call to be used.
```
weirdan, line 508:
```suggestion
Used to annotate an impure function or method the unused result of which Psalm should report.
```

weirdan (changes requested):
We should probably add one more thing here. Since `void|never + require-usage` is pointless it would make sense to flag those combinations. E.g.
```php
/**
 * @psalm-require-usage
 */
function f(): never { exit; } 
// InvalidDocblock(?): @require-usage is incompatible with `never` return type
```

ohader:
@weirdan Thanks for providing the examples! I agree that my idea would make it much more complicated. Besides that, I did not want to block this change nor request any changes in particular. Sorry for hijacking this PR with my own agenda...

kkmuffme:
@danog could you please merge this, then I could PR https://github.com/vimeo/psalm/issues/11381 since this is now available with PHP 8.5

Or should I change `@psalm-require-usage` to `@no-discard` (and `@psalm-no-discard`) and implement support for NoDiscard in this PR right away?

kkmuffme:
Changed to @psalm-no-discard now to match the PHP attribute naming

Additionally implemented support for the native PHP attribute now:
- add support for PHP 8.5 #[NoDiscard] to report unused return value. Fix https://github.com/vimeo/psalm/issues/11381
- add support for PHP 8.5 (void) typecase, as this is required to handle NoDiscard attribute. Fix https://github.com/vimeo/psalm/issues/11638
- fix bad error message and wrong handling for deprecated and removed (unset) typecast. Fix https://github.com/vimeo/psalm/issues/11639

Failing tests unrelated from https://github.com/vimeo/psalm/pull/11569 and bc of old base branch (already fixed, but I cannot retrigger the CI myself)

kkmuffme:
@danog correct me if I'm wrong, but in case weirdan isn't available to review anymore, would you be so kind please

PR #10745, opened 2024-02-24
Fix getFunctionIdsFromCallableArg to return the function id for array($this, 'foo') in params providers for example

kkmuffme:
Required to get consistent types in e.g. params providers used in plugins.


weirdan (changes requested):
Can you provide a test demonstrating what this change affects?

kkmuffme (commented):
Added test

Issue #10747, opened 2024-02-24
UnimplementedAbstractMethod on all enums when running PHAR version on Windows


Shira-3749:
I've encountered an obscure bug when the .phar version of Psalm is run on Windows.

(This only happens with the .phar version. It works in both cases when installed via Composer.)

Psalm 5.22.2@d768d914152dbbf3486c36398802f74e80cfde48
PHP 8.3.3 (cli)

Given this _psalm.xml_:

```xml
<?xml version="1.0"?>
<psalm
    errorLevel="1"
    resolveFromConfigFile="true"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="https://getpsalm.org/schema/config"
    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
    findUnusedBaselineEntry="true"
    findUnusedCode="false"
>
    <projectFiles>
        <directory name="src" />
    </projectFiles>
</psalm>

```

And this simple enum definition in src/Example.php:

```php
<?php

enum Foo {
    case A;
    case B;
}
```

When I run psalm.phar on the Windows host, I get the following output:

```
$ ./psalm.phar 
Target PHP version: 8.3 (inferred from current PHP version).
Scanning files...
Analyzing files...

ERROR: UnimplementedAbstractMethod - src/Example.php:3:6 - Method cases is not defined on class Foo, defined abstract in UnitEnum (see https://psalm.dev/101)
enum Foo {

------------------------------
1 errors found
------------------------------

Checks took 0.01 seconds and used 15.275MB of memory
No files analyzed
Psalm was unable to infer types in the codebase
```

But this passes when run inside a Docker container (based on php:8.3-apache).

```
root@d519a37baed3:/var/www/html/test/# ./psalm.phar 
Target PHP version: 8.3 (inferred from current PHP version).
Scanning files...
Analyzing files...

░
------------------------------
                              
       No errors found!       
                              
------------------------------

Checks took 1.17 seconds and used 30.963MB of memory
Psalm was unable to infer types in the codebase
```

weirdan:
Reproduced: https://github.com/weirdan/10747

weirdan:
The fun part is that the issue disappear when you extract the phar. Something tells me it's going to be fun to debug.

PR #10748, opened 2024-02-25
Fix loading stubs from phar file on Windows

### LINKED ISSUE #2858: array_keys() issue only with phar version
For this code (a bit too much going on to reproduce in sandbox):

```php
	/**
	 * @return Stopwatch_Record[]
	 */
	public function get_records(): array {

		/** @psalm-var array<string, StopwatchEvent> $events */
		$events = $this->stopwatch->getSectionEvents( '__root__' );

		/** @var string $name */
		foreach ( array_keys( $events ) as $name ) {
			if ( $this->stopwatch->isStarted( $name ) ) {
				unset( $events[ $name ] );
			}
		}

		return array_map( [ $this, 'transform' ], array_keys( $events ), $events );
	}

	/**
	 * @param string         $name  Event name.
	 * @param StopwatchEvent $event Stopwatch event instance.
	 *
	 * @return Stopwatch_Record
	 */
	protected function transform( string $name, StopwatchEvent $event ): Stopwatch_Record {

		return new Stopwatch_Record( $name, $event );
	}
```
https://github.com/Rarst/laps/blob/168a226859dade7b431d2d69ba751fb3667ee146/src/Record/Collector/Stopwatch_Collector.php#L49-L76

Getting this error:

> ERROR: InvalidScalarArgument - src\Record\Collector\Stopwatch_Collector.php:64:21 - First parameter of closure passed to function array_map expects string, int|string provided
                return array_map( [ $this, 'transform' ], array_keys( $events ), $events );

However `array_keys( $events )` should only contain/pass strings since `$events` is documented to only have string keys above `array<string, StopwatchEvent>`.

### LINKED ISSUE #9394: psalm/phar and github phar works differently from vimeo/psalm
I have a simple class src\Foo.php:

```php
<?php

/**
 * @template T
 * @implements IteratorAggregate<string,T>
 */
class Foo implements IteratorAggregate
{
	/**
	 * @var array<string, T>
	 */
	private $data;
	
	/**
	 * @param array<string, T> $data
	 */
	public function __construct(array $data = [])
	{
		$this->data = $data;
	}
	
	/**
	 * @return ArrayIterator
	 */
	#[\ReturnTypeWillChange]
	public function getIterator()
	{
		return new ArrayIterator($this->data);
	}
}
```
And psalm.xml:
```xml
<?xml version="1.0"?>
<psalm
	errorLevel="1"
	ignoreInternalFunctionFalseReturn="false"
	ignoreInternalFunctionNullReturn="false"
	cacheDirectory=".psalm-cache"
	phpVersion="5.6"
	findUnusedCode="false"
>
	<projectFiles>
		<directory name="src"/>
	</projectFiles>
</psalm>
```

When I install psalm using composer:
```
> composer require --dev vimeo/psalm
...
Using version ^5.7 for vimeo/psalm
> .\vendor\bin\psalm.bat --no-cache
Warning: "findUnusedBaselineEntry" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.

Install the opcache extension to make use of JIT on PHP 8.0+ for a 20%+ performance boost!

Target PHP version: 5.6 (set by config file).
Scanning files...
Analyzing files...

░
                              
       No errors found!       
                              
Checks took 0.49 seconds and used 73.594MB of memory
Psalm was able to infer types for 100% of the codebase
```

If I install psalm/phar instead:
```
> composer remove --dev vimeo/psalm
> composer require --dev psalm/phar
...
Using version ^5.7 for psalm/phar
> .\vendor\bin\psalm.phar.bat --no-cache
Warning: "findUnusedBaselineEntry" will be defaulted to "true" in Psalm 6. You should explicitly enable or disable this setting.

Install the opcache extension to make use of JIT on PHP 8.0+ for a 20%+ performance boost!

Target PHP version: 5.6 (set by config file).
Scanning files...
Analyzing files...

E

ERROR: TooManyTemplateParams - src/Foo.php:7:22 - Foo has too many template params when extending IteratorAggregate, expecting 0 (see https://psalm.dev/184)
class Foo implements IteratorAggregate


ERROR: InvalidAttribute - src/Foo.php:25:4 - The class ReturnTypeWillChange doesn't have the Attribute attribute (see https://psalm.dev/242)
        #[\ReturnTypeWillChange]


2 errors found

Checks took 0.12 seconds and used 26.961MB of memory
Psalm was able to infer types for 100% of the codebase
```
The same (2 errors) result is given when I download the PHAR from github.

Why would this happen, and how to work around it using the PHAR?

weirdan:
Fixes vimeo/psalm#10747
Fixes #9394 
Fixes #2858

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -2326,6 +2326,10 @@ class Config
 
         foreach ($stub_files as $file_path) {
             $file_path = str_replace(['/', '\\'], DIRECTORY_SEPARATOR, $file_path);
+            // fix mangled phar paths on Windows
+            if (strpos($file_path, 'phar:\\\\') === 0) {
+                $file_path = 'phar://'. substr($file_path, 7);
+            }
             $codebase->scanner->addFileToDeepScan($file_path);
         }
 
@@ -2423,6 +2427,10 @@ class Config
 
         foreach ($stub_files as $file_path) {
             $file_path = str_replace(['/', '\\'], DIRECTORY_SEPARATOR, $file_path);
+            // fix mangled phar paths on Windows
+            if (strpos($file_path, 'phar:\\\\') === 0) {
+                $file_path = 'phar://' . substr($file_path, 7);
+            }
             $codebase->scanner->addFileToDeepScan($file_path);
         }
 


weirdan:
That `str_replace()` was converting `phar://C:/path/to/psalm.phar\stubs\Php81.phpstub` into `phar:\\C:\path\to\psalm.phar\stubs\Php81.phpstub`. But `phar:\\` is an invalid scheme.

Issue #10578, opened 2024-01-19
Operand of type false is always falsy


tdclaritum:
The following code 

```php
<?php

/** @var list<array{id: int, template_id: int}> $records */
$records = json_decode(file_get_contents('test.json'), true);
$processed = [];
foreach($records as $record) {
    if (empty($processed[$record['template_id']])) {
        $processed[$record['template_id']] = [];
    }
    $processed[$record['template_id']][] = $record['id'];
}
```

produces the following error 

```
ERROR: TypeDoesNotContainType - a.php:7:9 - Operand of type false is always falsy (see https://psalm.dev/056)
    if (empty($processed[$record['template_id']])) {
```

but it shouldn't. It seems the issue is that `psalm` doesn't know the data inside `$records`. If `$records` is defined like
```php
$records = [
    ['id' => 1, 'template_id' => 1],
    ['id' => 2, 'template_id' => 1],
    ['id' => 3, 'template_id' => 2],
];
```
there is no error reported.

Reproduce on psalm.dev: https://psalm.dev/r/5921391c4a

**Workaround**: replace `empty()` with `! isset()` - `if (empty($processed[$record['template_id']])) {`  => `if (! isset($processed[$record['template_id']])) {`

Edit: syntax highlighting

edsrzf:
It looks like this behavior was introduced in 5.20 with #10502. I ran into it when trying to update from 5.18 to 5.22.

edsrzf:
A smaller repro that's essentially the same thing, though it results in a different error: https://psalm.dev/r/2e5b073a15

It seems like this is may be an intentional change from #10502?

In my repro:

* Psalm types `$a['a']` as `true`, even though the key `'a'` might not exist in the array.
* Due to the changes in the PR, `empty(true)` now type checks as the type `false`, where previously it had the less specific type `bool`.
* Psalm reports this as a redundant condition.

This all makes sense from Psalm's point of view. I can't see an easy way to fix it except to revert that part of #10502 and go back to typing the expression as `bool`.

The workaround using `isset` works because `isset` doesn't have this behavior. It always type checks to `bool`, no matter what.

You could maybe make the case that `isset` should be preferred over `empty` for these types of checks, but it feels wrong to me to report this as a redundant or invalid condition when it's not.

weirdan:
[...]

PR #10752, opened 2024-02-26
Set inside_isset false when analyzing ArrayDimFetch index

edsrzf:
For an expression like `isset($a[$b])`, the `inside_isset` context flag wasn't being reset to `false` while analyzing the index (`$b` in this case). This meant that certain issues would not be detected for the index expression, such as undefined variables or possibly undefined offset. I found this while working on a fix for #10578 and decided to address it separately.

This resulted in new errors in two places in the Psalm code base, which I've also fixed.


weirdan (approved):

orklah (approved):

weirdan:
Thanks!

PR #10753, opened 2024-02-27
Set inside_isset = false when analyzing arguments

edsrzf:
This is similar to #10752, but for call arguments. This means that certain types of errors could go undetected when inside call arguments, eg: https://psalm.dev/r/1291966e74 `$a` should be flagged as undefined.

This luckily doesn't result in any new errors in the Psalm code base.


weirdan (approved):

weirdan:
Thanks!

PR #10756, opened 2024-02-28
When inside isset, make array fetch result nullable

### LINKED ISSUE #10578: Operand of type false is always falsy

### LINKED ISSUE #10715: Regression with RedundantConditionGivenDocblockType and DocblockTypeContradiction
Psalm incorrectly reports truthy and falsy values, and marks a redundant check or a doc type contradiction:

This broke in [5.19.1](https://github.com/vimeo/psalm/releases/tag/5.19.1) I think

Consider:
```php
/**
 * @param array<string, true> $array
 */
function func(array $array): int
{
    if (!empty($array['stuff'])) return 1;
    
    if (empty($array['more stuff'])) return 2;

    return 3;
}
```

Psalm returns:
```
Psalm output (using commit [c488d40](https://github.com/vimeo/psalm/commit/c488d40)): 

ERROR: [RedundantConditionGivenDocblockType](https://psalm.dev/156) - 8:6 - Operand of type true is always truthy

ERROR: [DocblockTypeContradiction](https://psalm.dev/155) - 10:9 - Operand of type false is always falsy
```

https://psalm.dev/r/892d59e94e

### LINKED ISSUE #10716: Regression: incorrectly reported TypeDoesNotContainType in a loop
Psalm started incorrectly reporting TypeDoesNotContainType some time after [5.19.1](https://github.com/vimeo/psalm/releases/tag/5.19.1)

Consider a two dimensional array, the code returns a count for each unique value in "key" column
```php
/**
 * @param array{key: string}[] $array
 */
function func(array $array): array
{
    $r = [];
    foreach ($array as $row) {
        $r[$row['key']] = empty($r[$row['key']])
            ? 1
            : $r[$row['key']] + 1;
    }
    return $r;
}
```

psalm reports
```
Psalm output (using commit [c488d40](https://github.com/vimeo/psalm/commit/c488d40)): 

ERROR: [TypeDoesNotContainType](https://psalm.dev/056) - 10:27 - Operand of type false is always falsy
```

https://psalm.dev/r/707051b9f6

edsrzf:
This prevents false positive for various types of issues inside empty, such as `RedundantConditionGivenDocblockType` and `TypeDoesNotContainType`.

This fix tends to produce more `RiskyTruthyFalsyComparison` issues than previous for cases like `empty($a[$i])`, because it's not clear whether it's checking for the array not containing the index, for a falsy value in the array, or both. I've updated the baseline here rather than fix the errors because I'm not actually sure of the best way to resolve the new issues.

Fixes #10578, fixes #10715, fixes #10716 (I've tested against the reproduction cases for all these issues and it's the same underlying cause)

### DIFF
--- a/tests/TypeReconciliation/EmptyTest.php
+++ b/tests/TypeReconciliation/EmptyTest.php
@@ -771,6 +771,13 @@ class EmptyTest extends TestCase
                     }',
                 'error_message' => 'RedundantConditionGivenDocblockType',
             ],
+            'redundantEmptyArrayFetch' => [
+                'code' => '<?php
+                    /** @var array<true> $a */;
+                    assert(isset($a["a"]));
+                    if (empty($a["a"])) {}',
+                'error_message' => 'DocblockTypeContradiction',
+            ],
         ];
     }
 }
--- a/tests/TypeReconciliation/EmptyTest.php
+++ b/tests/TypeReconciliation/EmptyTest.php
@@ -231,6 +231,8 @@ class EmptyTest extends TestCase
                     function foo(array $o) : void {
                         if (empty($o[0]) && empty($o[1])) {}
                     }',
+                'assertions' => [],
+                'ignored_issues' => ['RiskyTruthyFalsyComparison'],
             ],
             'multipleEmptiesInConditionWithMixedOffset' => [
                 'code' => '<?php
@@ -315,6 +317,8 @@ class EmptyTest extends TestCase
                             if (empty($data[Foo::ONE])) {}
                         }
                     }',
+                'assertions' => [],
+                'ignored_issues' => ['RiskyTruthyFalsyComparison'],
             ],
             'doubleEmptyCheckTwoArrays' => [
                 'code' => '<?php
@@ -330,6 +334,8 @@ class EmptyTest extends TestCase
                     function foo(array $arr) : void {
                         if (empty($arr["a"]) && empty($arr["b"])) {}
                     }',
+                'assertions' => [],
+                'ignored_issues' => ['RiskyTruthyFalsyComparison'],
             ],
             'doubleEmptyCheckOnTKeyedArrayVariableOffsets' => [
                 'code' => '<?php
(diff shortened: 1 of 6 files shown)


tests/TypeReconciliation/EmptyTest.php
```diff
@@ -626,6 +634,11 @@ function test(string $s): void {
                     '$x===' => 'true',
                 ],
             ],
+            'emptyArrayFetch' => [
+                'code' => '<?php
+                    /** @var array<true> $a */
+                    if (empty($a["a"])) {}',
+            ],
```
weirdan, line 641: I'd suggest adding another test like this:
```php
/** @var array<true> $a*/;
assert(isset($a["a"]));
if (empty($a["a"])) {} // it should report `TypeDoesNotContainType` (or paradox) here
```

edsrzf: The error I get is `DocblockTypeContradiction`, which...doesn't seem 100% correct, since the contradiction doesn't come from the docblock type, but rather the assertion. But I suspect this is a separate issue?

edsrzf:
@weirdan any more thoughts on this change? I'm pretty keen to get the linked issues fixed.

weirdan (approved):

weirdan:
Thanks!

PR #10759, opened 2024-02-29
void|never shouldn't be combined to null but keep void

kkmuffme:
Since psalm internally often checks for isVoid (or isNever) which should include cases of void|never from what I saw, but since the type is changed to null some errors aren't reported.
Initiated by https://github.com/vimeo/psalm/pull/10742#pullrequestreview-1899640189

### DIFF
--- a/tests/FileManipulation/MissingReturnTypeTest.php
+++ b/tests/FileManipulation/MissingReturnTypeTest.php
@@ -538,9 +538,9 @@ final class MissingReturnTypeTest extends FileManipulationTestCase
                     }',
                 'output' => '<?php
                     /**
-                     * @return null|string
+                     * @return string|void
                      *
-                     * @psalm-return \'hello\'|null
+                     * @psalm-return \'hello\'|void
                      */
                     function foo() {
                       if (rand(0, 1)) {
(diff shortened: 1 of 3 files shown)


tests/FileManipulation/MissingReturnTypeTest.php
```diff
@@ -533,9 +533,9 @@ function bar() {
                     }',
                 'output' => '<?php
                     /**
-                     * @return null|string
+                     * @return string|void
```
weirdan, line 536: `void` in a union looks wrong. Looking at the code here I'd expect `null|string` actually.

tests/FunctionCallTest.php
```diff
@@ -3181,6 +3181,18 @@ function foo(never $_): void
                 'ignored_issues' => [],
                 'php_version' => '8.1',
             ],
+            'disallowVoidNeverTypeForParam' => [
+                'code' => '<?php
+                    /**
+                     * this is just one example where psalm uses isVoid and isNever, which is why void|never should not be collapsed to null
+                     * @param void|never $i
```
weirdan, line 3188: This should result in `void` (as `never` is the bottom type) and then trigger `ReservedWord`

PR #10769, opened 2024-03-02
Since annotations outside phpstub should not infer php version

### LINKED ISSUE #10761: Psalm ignores functions with certain `@since` annotations, and then reports a ParseError
It seems that Psalm is interpreting a `@since` docblock annotation, when the value matches `/^[4578]\.\d(\.\d+)?$/`, as indicating a _PHP_ version ([ref](https://github.com/vimeo/psalm/blob/9aa450bbe6aa67fe4789457e6382d1598c575644/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php#L402-L410)) and ignores the function definition if the value is newer than (a mangled version of[^1]) the value of `--php-version` or `<psalm phpVersion>` ([ref1](https://github.com/vimeo/psalm/blob/9aa450bbe6aa67fe4789457e6382d1598c575644/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php#L464-L476), [ref2](https://github.com/vimeo/psalm/blob/9aa450bbe6aa67fe4789457e6382d1598c575644/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php#L1048-L1060)).

Then, since the function was ignored, Psalm later reports a ParseError about "$storage should not be null" because it doesn't know anything about the function (cf. https://github.com/vimeo/psalm/issues/9672#issuecomment-1513456138).

This interferes with the common (and [documented](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/since.html)) use of `@since` to indicate the version of _the project_ in which a function was added.

See https://psalm.dev/r/b16dc9cdc5 for an example.

[^1]: Due to another bug in the code ([here](https://github.com/vimeo/psalm/blob/9aa450bbe6aa67fe4789457e6382d1598c575644/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php#L466) and [here](https://github.com/vimeo/psalm/blob/9aa450bbe6aa67fe4789457e6382d1598c575644/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php#L1050)), it interprets any phpVersion "8.x" as if it were "8.8", any "7.x" as if it were "7.7", and so on.

kkmuffme:
- `@since` annotations should only infer PHP version in .phpstub files or for `@since 8.0.0 PHP` refs https://github.com/vimeo/psalm/issues/10761
- fix composer PHP versions 8.2 and 8.3 ignored

Fixes #10761

### DIFF
--- a/src/Psalm/Config.php
+++ b/src/Psalm/Config.php
@@ -2781,8 +2781,22 @@ class Config
                 $version_parser = new VersionParser();
 
                 $constraint = $version_parser->parseConstraints($php_version);
+                $php_versions = [
+                    '5.4',
+                    '5.5',
+                    '5.6',
+                    '7.0',
+                    '7.1',
+                    '7.2',
+                    '7.3',
+                    '7.4',
+                    '8.0',
+                    '8.1',
+                    '8.2',
+                    '8.3',
+                ];
 
-                foreach (['5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1'] as $candidate) {
+                foreach ($php_versions as $candidate) {
                     if ($constraint->matches(new Constraint('<=', "$candidate.0.0-dev"))
                         || $constraint->matches(new Constraint('<=', "$candidate.999"))
                     ) {
--- a/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php
@@ -23,6 +23,7 @@ use function count;
 use function explode;
 use function implode;
 use function in_array;
+use function pathinfo;
 use function preg_last_error_msg;
 use function preg_match;
 use function preg_replace;
@@ -36,6 +37,8 @@ use function strtolower;
 use function substr;
 use function substr_count;
 use function trim;
+
+use const PATHINFO_EXTENSION;
 
 /**
  * @internal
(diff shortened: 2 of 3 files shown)


src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php
```diff
@@ -417,11 +420,15 @@ public static function parse(
...
-
-                $info->since_php_major_version = (int)$since_parts[0];
-                $info->since_php_minor_version = (int)$since_parts[1];
+            // only for phpstub files or @since 8.0.0 PHP
+            // since @since is commonly used with the project version, not the PHP version
+            // https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/since.html
+            // https://github.com/vimeo/psalm/issues/10761
+            if (preg_match('/^([4578])\.(\d)(\.\d+)?(\s+PHP)?$/i', $since, $since_match)
+                && isset($since_match[1])&& isset($since_match[2])
+                && (!empty($since_match[4]) || pathinfo($code_location->file_name, PATHINFO_EXTENSION) === 'phpstub')) {
```
weirdan, line 429: It'll do for now, but we should probably limit this to only the stubs Psalm itself ships in future. When I renamed the stubs I didn't want `.phpstub` to be anything specific to Psalm, but hoped to establish a naming convention for any stubs across the ecosystem.

src/Psalm/Config.php
```diff
@@ -2781,8 +2781,22 @@ public function getPHPVersionFromComposerJson(): ?string
...
+                    '7.0',
+                    '7.1',
+                    '7.2',
+                    '7.3',
+                    '7.4',
+                    '8.0',
+                    '8.1',
+                    '8.2',
+                    '8.3',
+                ];
```
weirdan, line 2797: Note to self: we keep forgetting to update PHP version numbers in different places. This means we need to centralize PHP version handling in a single place.

weirdan (approved):

src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockParser.php
weirdan: Many plugins use `.phpstub` extension, and not for built-in classes / extensions: https://github.com/search?q=org%3Apsalm%20phpstub&type=code

Issue #6935, opened 2021-11-16
Deprecated/not available functions not reported


kkmuffme:
Assuming psalm.dev runs on PHP 7.4.x (phpVersion=7.4 in config, since I cannot see the actual config)

https://psalm.dev/r/5201b68036

Should give an error:
https://www.php.net/manual/en/migration72.deprecated.php

===

Also vice versa, if phpVersion=7.0, I expect to get an error for this:
https://psalm.dev/r/208b2cb781

as the array_key_first function was added in 7.3.0 https://www.php.net/manual/de/function.array-key-first.php

weirdan:
It appears Psalm reflects the runtime function definition when it can't find the requested function in callmap:
![image](https://user-images.githubusercontent.com/57403/142190010-950ff997-10a3-422b-981a-f97b3b53cced.png)

kkmuffme:
@weirdan I guess there is no way around hardcoding all deprecated/newly introduced functions and just checking against that in psalm?

There is a (non-maintained?) tool, that already has all that data collected up to PHP 8.0 (I think PHP 8.0 isn't fully complete, but before that I'm sure it is): https://github.com/wapmorgan/PhpDeprecationDetector/tree/master/data

weirdan:
We have most of that data in callmaps (added and removed functions). If we're going to add deprecations for internal functions, it would make sense to have them there as well.

PR #10771, opened 2024-03-02
Report UndefinedFunction/Method correctly if analyzed PHP version is older than currently used PHP version

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/6935

- this also fixes a bug where the callmaps cache wasn't cleared in a case where it should be cleared
- this also fixes a bug where a function would not report as undefined if reflection encountered an error
- this also fixes an issue where reflection would incorrectly modify function/method info if the PHP version running Psalm is older than the declared --php-version (or via config)


kkmuffme:
Replaced by https://github.com/vimeo/psalm/pull/11708 for 6.x branch

PR #10777, opened 2024-03-03
Variadic params should always be optional and fix redundant optional variadic annotation

### LINKED ISSUE #9188: Inconsistent handling of "empty"-status of arrays as return values
**Description**
The assumption about whether arrays are non-empty seems to be inconsistent between the types `list` and `array` as output typing from a function.  

**To Reproduce**
https://psalm.dev/r/7320dbad2e
In this first snippet, the return type of the function `map` is typed with "list<T>". When attempting to use the spread operator on this return value as the parameter of a function which requires at least 1 member to be present, it triggers a warning "possibly undefined". 

https://psalm.dev/r/5f017e6180
The second snippet is the same setup -- but the return value is array<T> not list<T> -- but the above error does not occur. 

**Expected behavior**
I don't have intimate knowledge of Psalm internals, so I can't say whether the one behaviour or the other is correct -- but it does seem that it is inconsistent between the two.

I would expect that Psalm would consider both array<T> and list<T> to be potentially undefined in this spread operator -- or that they would both not cause a problem. 

**Environment (please complete the following information):**
- OS: Linux Fedora
- PHP version 8.1
- Psalm version 5.5.0

Thanks a lot for a great library !

### LINKED ISSUE #10787: PHP variadic internal functions should not allow named calls
https://psalm.dev/r/14572dea00

This is a fatal error.

The reason it's wrong in psalm is that PHP's docs itself have been wrong all along (since the beginning), which is why this is wrong in psalm too I guess - see https://github.com/php/doc-en/issues/3233

kkmuffme:
- fix variadic params have wrong type in PHP 8/8.1
- Fix https://github.com/vimeo/psalm/issues/10787
- add error when using named arguments in PHP <8
- fix report an error in all cases when unpacking string keyed array in PHP < 8.1 in all cases (reported for < 8.0 in some incorrectly)
- fix no errors reported for unpacked union (e.g. array<string>|int)
- fix keyed array with fallback didn't report the same errors as unkeyed array for toomany/toofewarguments
- add error for possibly too many/few arguments with array unpacking
- fix named arguments after positional arguments not reported in unpacking
- fix https://github.com/vimeo/psalm/issues/9188

### DIFF
--- a/examples/TemplateChecker.php
+++ b/examples/TemplateChecker.php
@@ -62,7 +62,7 @@ final class TemplateAnalyzer extends Psalm\Internal\Analyzer\FileAnalyzer
                     throw new InvalidArgumentException('Could not interpret doc comment correctly');
                 }
 
-                /** @psalm-suppress ArgumentTypeCoercion */
+                /** @psalm-suppress ArgumentTypeCoercion, TooFewArguments */
                 $method_id = new MethodIdentifier(...explode('::', $matches[1]));
 
                 $this_params = $this->checkMethod($method_id, $first_stmt, $codebase);
--- a/examples/plugins/FunctionCasingChecker.php
+++ b/examples/plugins/FunctionCasingChecker.php
@@ -35,7 +35,7 @@ final class FunctionCasingChecker implements AfterFunctionCallAnalysisInterface,
         }
 
         try {
-            /** @psalm-suppress ArgumentTypeCoercion */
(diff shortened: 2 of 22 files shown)


weirdan:
189 new errors reported in PHPUnit look quite worrying: https://app.circleci.com/pipelines/github/vimeo/psalm/13214/workflows/7c12912e-4b6a-4e48-9ba6-06a9d0015c09/jobs/43049

PR #10781, opened 2024-03-03
Update CallMap for sqlsrv_connect and sqlsrv_errors to match reflection

theodorejb:
Otherwise tests fail when the sqlsrv extension is enabled.


weirdan (approved):

weirdan:
Thanks!

Issue #10461, opened 2023-12-07
False positives in the function forward_static_call_array


jorgsowa:
Hello,
The following code produces two false positives
```
<?php

/** @return class-string */
function getClassName(object $class): string {
    return $class::class;
};

function fromArrays(array $data, object $object)
{
    $className = getClassName($object);

    return forward_static_call_array([$className, 'from'], $data);
}
```
https://psalm.dev/r/ecc7244555

We know that the variable passed to `forward_static_call_array` is a class string, so it should be treated as a callable.

```
ERROR: [InvalidArgument](https://psalm.dev/004) - 12:38 - Argument 1 of forward_static_call_array expects callable, but list{class-string, 'from'} provided
```

```
ERROR: [ArgumentTypeCoercion](https://psalm.dev/193) - 12:60 - Argument 2 of forward_static_call_array expects list<mixed>, but parent type array<array-key, mixed> provided
```

weirdan:
> We know that the variable passed to forward_static_call_array is a class string, so it should be treated as a callable.

we don't know if there's a method named `from()` on that class (because we don't know the class name in the first place), so we can't do that: https://3v4l.org/pb3Rr

The latter issue is a useful safeguard as lists work for every PHP version, but string-keyed arrays need to match the signature (and Psalm doesn't seem to be able to infer it): https://3v4l.org/LvpJt

jorgsowa:
That's true, but if we pass type `list{class-string, string}` to function `forward_static_call_array` we shouldn't have any warning, because this construct creates a callable which is expected. The code is valid and in this case, the only workaround is to add the warning to baseline.

weirdan:
> because this construct creates a callable which is expected

That's the thing with callables - they are only callable if they can be called. Otherwise, they are just arrays / strings.

With that said, this should work (but doesn't): https://psalm.dev/r/a1bdbc0e6b

weirdan:
> This also produces a fatal error when run in PHP, doesn't it?
No, it works just fine: https://3v4l.org/lTQWQ

weirdan:
> But apparently forward_static_call_array also works outside of a class context as your example above shows?

:man_shrugging: I was confused by that passage in the docs too.

PR #10805, opened 2024-03-10
Allow more callable types as subtypes of `callable`

weirdan:
Fixes vimeo/psalm#10461

### DIFF
--- a/src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php
+++ b/src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php
@@ -8,6 +8,7 @@ use Psalm\Type\Atomic;
 use Psalm\Type\Atomic\Scalar;
 use Psalm\Type\Atomic\TArray;
 use Psalm\Type\Atomic\TCallable;
+use Psalm\Type\Atomic\TCallableArray;
 use Psalm\Type\Atomic\TCallableKeyedArray;
 use Psalm\Type\Atomic\TCallableObject;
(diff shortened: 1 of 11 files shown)


src/Psalm/Internal/Type/Comparator/AtomicTypeComparator.php
```diff
@@ -191,7 +192,12 @@ public static function isContainedBy(
         }
 
         if (($container_type_part instanceof TCallable
-                && $input_type_part instanceof TCallable)
+            && ($input_type_part instanceof TCallable
+                || $input_type_part instanceof TCallableArray
+                || $input_type_part instanceof TCallableObject
+                || $input_type_part instanceof TCallableString
+                || $input_type_part instanceof TCallableKeyedArray
```
SCIF, line 199: Looks like `TCallableInterface` would help to prevent such cases. And would help to make clearer distinction between `callable` and it's implementations.

danog: +1

zerkms:
Can it be relevant that now the following code:

```php
->controller([MetricsController::class, 'prometheus'])
```
for `controller` method (symfony): 
```php
final public function controller(callable|string|array $controller): static
```
triggers
```
Argument 1 of Symfony\Component\Routing\Loader\Configurator\RouteConfigurator::controller expects a public static callable, but a non-static callable provided 
```
?

Not reproducible on psalm.dev though

Issue #10807, opened 2024-03-11
Method `__callStatic()` breaks asserts on 5.23


SCIF:
https://psalm.dev/r/2a4ebd4d08

SCIF:
I assume not only asserts affected, but much more, however, I haven't checked the damaged area.

SCIF:
Repro https://github.com/vimeo/psalm/pull/10808

antonioeatgoat:
Here is another example 👇 

https://psalm.dev/r/1c3825394e

weirdan:
@antonioeatgoat this error is valid, see https://docs.phpdoc.org/2.9/references/phpdoc/tags/method.html:

> If the intended method is static, the `static` keyword can be placed before the return type to communicate that. **There must be a return type, `static` on its own would mean that the method returns an instance of the child class which the method is called on.**

antonioeatgoat:
Thank you so much @weirdan, I missed it. Then I guess it was just not recognised by previous versions.

weirdan:
Fixed in #10812

PR #10808, opened 2024-03-11
Test for #10807

SCIF:


weirdan:
Merged as part of #10812

PR #10481, opened 2023-12-12
report error for invalid array key type

kkmuffme:
- Add error for invalid array key types in docblock
- Fix [#8680](https://github.com/vimeo/psalm/issues/8680)
See also [#9295](https://github.com/vimeo/psalm/issues/9295)

PR #10814, opened 2024-03-12
Improve string-int juggle consistency in array keys and display for int-like strings in type

kkmuffme:
- display `array{'08': string}` as `array{'08': string}` instead of `array{08: string}` since it's not juggled to int, but can be confused as if it were (since `array{'8': string}` which is juggled displays as `array{8: string}` since it's an int, but in the first case it's not. Just like how it works with floats, e.g. `array{'15.3': string}`)
- further improve string-int juggling handling which was previously already improved by me in https://github.com/vimeo/psalm/pull/10481 (which is where tests have been added already for various cases)
- fix https://psalm.dev/r/3b401c6f88

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
@@ -49,7 +49,6 @@ use function in_array;
 use function is_int;
 use function is_numeric;
 use function is_string;
-use function preg_match;
 use function trim;
 
 use const FILTER_VALIDATE_INT;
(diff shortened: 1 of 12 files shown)


jack-worman:
You could add the follow test:

```php
    {
        return [
            'php is amazing' => [
                'code' => <<<'PHP'
                    <?php
                    $a = ['15' => 'a'];
                    PHP,
                'assertions' => ['$a===' => "array{15: 'a'}"],
            ],
        ];
    }
```

danog:
The correct behavior would be to treat numeric-string in array keys as int, because that's (unfortunately) what PHP does.
The logic here isn't too clear, what's the behavior you're trying to reach?

@jack-worman's testcase is correct.

weirdan:
I think in this specific case that `preg_match()` is a dead code because the `$key` comes from an actual PHP array key. So even if we assigned `"15" => "whatever"`, it would get internally converted into `15 => "whatever"` by PHP itself.

src/Psalm/Internal/Analyzer/Statements/Expression/ArrayAnalyzer.php
```diff
@@ -237,6 +241,38 @@ public static function analyze(
         return true;
     }
 
+    /**
+     * @param string|int $literal_array_key
+     * @return false|int
+     * @psalm-assert-if-false !numeric $literal_array_key
+     */
+    public static function getLiteralArrayKeyInt(
```
weirdan, line 249: I first thought we could replace it with something like:

```php
function convertIntishArrayKey(int|string $k): false|int {
    return is_int(key([$k => null])) ? $k : false;
}
```
but then realized that then target PHP version may have different rules.

tests/ArrayKeysTest.php
```diff
@@ -183,6 +183,26 @@ public function test(bool $v): array {
...
+            'intStringKeyAsInt' => [
+                'code' => '<?php
+                    $a = ["15" => "a"];
+                    $b = ["15.7" => "a"];
+                    // since PHP 8 this is_numeric but will not be int key
+                    $c = ["15 " => "a"];
+                    $d = ["-15" => "a"];
+                    // see https://github.com/php/php-src/issues/9029#issuecomment-1186226676
+                    $e = ["+15" => "a"];
+                    $f = ["015" => "a"];
```
weirdan, line 195: I'd also add another example with the scientific format - e.g. `1e2`. It's also numeric, but is not converted to int.

weirdan (approved):

weirdan (approved):

weirdan:
Thanks!

PR #10815, opened 2024-03-13
add support for named arguments for filter_var and filter_input

### LINKED ISSUE #10809: filter_input return type confusion when using named parameters and leaving out the `filter` argument
https://psalm.dev/r/23882663dc

all calls are equivalent, yet the last one does not yield the expected result.

pilif:
this fixes #10809


weirdan (approved):

weirdan:
@pilif Thanks!

weirdan:
@kkmuffme please refrain from such comments

1. They are annoying. I receive all updates on all pull requests and issues, there's no need to add additional noise
2. You're not a boss of mine, don't tell me what to do
3. I'm not the only maintainer here
4. The others don't need to be bugged either
5. We're volunteers here and do the work in our spare time when we have the time and energy

Now there could be times when an old PR or an issue may need to be brought up, but that certainly wasn't the case here or on other PRs you pinged me recently.

PR #10834, opened 2024-03-15
Greater > and Less Than < comparisons with non-numeric, non-string, non-resource types should report error

### LINKED ISSUE #10833: Greater > and Less Than < comparisons with non-numeric, non-string, non-resource types should report error
e.g. https://psalm.dev/r/e63cef6890 this is a redundant condition but isn't reported, or https://psalm.dev/r/29eb392672 which causes a PHP notice.

However there are tons of things one can do wrong with those operators:
https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.types
and it would take an immense amount of time to figure out all possible cases and all report them as separate errors.

As a first, I suggest to report all > and < operations that contain an array, object, probably even null and bool as possible error - and refine this if e.g. there are too many false positives for some cases (e.g. `int|false > 100`)

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/10833

For int|null cases the errors could be slightly annoying, however null can refer to the minimum value (PHP_INT_MIN) but is compared as 0, or it refers to a non-numeric state and shouldn't be used in comparison in the first place (similar to what RiskyTruthyFalsyComparison checks, which I guess would apply here too, but I want to keep it simple for now)

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
@@ -23,15 +23,20 @@ use Psalm\Internal\MethodIdentifier;
 use Psalm\Issue\DocblockTypeContradiction;
 use Psalm\Issue\ImpureMethodCall;
 use Psalm\Issue\InvalidOperand;
+use Psalm\Issue\PossiblyInvalidOperand;
 use Psalm\Issue\RedundantCondition;
 use Psalm\Issue\RedundantConditionGivenDocblockType;
 use Psalm\Issue\TypeDoesNotContainType;
 use Psalm\IssueBuffer;
 use Psalm\Plugin\EventHandler\Event\AddRemoveTaintsEvent;
 use Psalm\Type;
+use Psalm\Type\Atomic\TFloat;
+use Psalm\Type\Atomic\TInt;
 use Psalm\Type\Atomic\TLiteralInt;
 use Psalm\Type\Atomic\TLiteralString;
 use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Atomic\TResource;
+use Psalm\Type\Atomic\TString;
 use Psalm\Type\Union;
 use UnexpectedValueException;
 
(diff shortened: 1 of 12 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
```diff
@@ -240,19 +245,65 @@ public static function analyze(
...
+                        $statements_analyzer->getSuppressedIssues(),
+                    );
+                } else {
+                    foreach ($stmt_left_type->getAtomicTypes() as $atomic_type) {
+                        if ($atomic_type instanceof TString
+                            || $atomic_type instanceof TInt
+                            || $atomic_type instanceof TFloat
+                            || $atomic_type instanceof TResource) {
+                            continue;
+                        }
```
jack-worman, line 268: also allow `DateTimeInterface`?

dictionaries/CallMap.php
```diff
@@ -2793,11 +2793,11 @@
...
 'file_get_contents' => ['string|false', 'filename'=>'string', 'use_include_path='=>'bool', 'context='=>'?resource', 'offset='=>'int', 'length='=>'?int'],
 'file_put_contents' => ['int<0, max>|false', 'filename'=>'string', 'data'=>'string|resource|array<string>', 'flags='=>'int', 'context='=>'resource'],
-'fileatime' => ['int|false', 'filename'=>'string'],
-'filectime' => ['int|false', 'filename'=>'string'],
+'fileatime' => ['int<0, max>|false', 'filename'=>'string'],
+'filectime' => ['int<0, max>|false', 'filename'=>'string'],
 'filegroup' => ['int|false', 'filename'=>'string'],
 'fileinode' => ['int|false', 'filename'=>'string'],
-'filemtime' => ['int|false', 'filename'=>'string'],
+'filemtime' => ['int<0, max>|false', 'filename'=>'string'],
```
jack-worman, line 2800: Counter example:
```php
$filePath = tempnam(sys_get_temp_dir(), '_');
touch($filePath, -42, -100);
var_dump(filemtime($filePath));
var_dump(fileatime($filePath));
```

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
```diff
@@ -240,19 +251,117 @@ public static function analyze(
...
+                                 && min($right_literal_int_floats) >= 0)
+                                || ($stmt instanceof PhpParser\Node\Expr\BinaryOp\GreaterOrEqual
+                                    && min($right_literal_int_floats) > 0)
+                                || ($stmt instanceof PhpParser\Node\Expr\BinaryOp\Smaller
+                                    && max($right_literal_int_floats) <= 0)
+                                || ($stmt instanceof PhpParser\Node\Expr\BinaryOp\SmallerOrEqual
+                                    && max($right_literal_int_floats) < 0))
+                        ) {
+                            continue;
+                        }
```
jack-worman, line 301: This is repeated (almost identically) below. We should abstract to a private function

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
```diff
@@ -240,19 +251,117 @@ public static function analyze(
...
+                            || $atomic_type instanceof TArrayKey
+                            || $atomic_type instanceof TFloat
+                            || $atomic_type instanceof TResource) {
+                            continue;
+                        }
+
+                        if (AtomicTypeComparator::isContainedBy(
+                            $statements_analyzer->getCodebase(),
+                            $atomic_type,
+                            new TNamedObject('DateTimeInterface'),
```
jack-worman, line 283: [GMP](https://www.php.net/manual/en/class.gmp.php) objects also overload operators

tests/BinaryOperationTest.php
```diff
@@ -1218,6 +1262,104 @@ function foo(string $s1, string $s2): string {
...
+            ],
+            'greaterEqualObjectNotice' => [
+                'code' => '<?php
+                    $a = rand(0, 1) > 0 ? rand() : new stdClass();
+                    if ($a >= 0) {
+                        echo "yes";
+                    }',
+                'error_message' => 'PossiblyInvalidOperand',
+            ],
+            'lessDateTimeInterfaceInvalid' => [
```
jack-worman, line 1321: Add a testcase for `DateTimeInterface > 0` giving an error.

src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOpAnalyzer.php
weirdan: It looks like the following classes from php-src have custom comparison behaviour:

```console
$ rg '\.compare = ' | rg -v 'zend_objects_not_comparable'

ext/date/php_date.c:	date_object_handlers_date.compare = date_object_compare_date;
ext/date/php_date.c:	date_object_handlers_immutable.compare = date_object_compare_date;
ext/date/php_date.c:	date_object_handlers_timezone.compare = date_object_compare_timezone;
ext/date/php_date.c:	date_object_handlers_interval.compare = date_interval_compare_objects;
Zend/zend_closures.c:	closure_handlers.compare = zend_closure_compare;
ext/gmp/gmp.c:	gmp_object_handlers.compare = gmp_compare;
ext/intl/breakiterator/breakiterator_class.cpp:	BreakIterator_handlers.compare = BreakIterator_compare_objects;
ext/intl/timezone/timezone_class.cpp:	TimeZone_handlers.compare = TimeZone_compare_objects;
ext/simplexml/simplexml.c:	sxe_object_handlers.compare = sxe_objects_compare;
ext/spl/spl_array.c:	spl_handler_ArrayObject.compare = spl_array_compare_objects;
```

weirdan: > According to the docs [php.net/manual/en/class.dateinterval.php](https://www.php.net/manual/en/class.dateinterval.php) this is incomparable - if I understand this correctly we cannot use <>

Correct, it just emits a warning and fails:
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/date/php_date.c#L4273-L4281

> I guess this applies also to
>> ext/date/php_date.c: date_object_handlers_timezone.compare = date_object_compare_timezone;

No, there are actually several defined ways (depending on the TZ type) to compare `DateTimeZone`:
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/date/php_date.c#L2035-L2048

But not `IntlTimezone` - this one is basically unorderable:
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/intl/timezone/timezone_class.cpp#L281-L287 

Closures are unorderable too:
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/Zend/zend_closures.c#L429-L457

Same for BreakIterator (but looks buggy):
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/intl/breakiterator/breakiterator_class.cpp#L88-L92

Ditto for SimpleXMLObject: 
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/simplexml/simplexml.c#L1233-L1250

SplArray compares as array:
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/spl/spl_array.c#L896-L899

For FFI (not in the list above) PHP compares pointers:
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/ffi/ffi.c#L1771

SplObjectStorage (also not in the list) are sorted as arrays:
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/spl/spl_observer.c#L397

GMP is properly comparable:
https://github.com/php/php-src/blob/f2ec6e480615f4a5633eb07ad8bfb7b2cadcb921/ext/gmp/gmp.c#L434

> Should I disable the error for these like I did for DateTimeInterface?

Since we probably don't want to allow comparing arrays (with `> / <`) this leaves:
* DateTime/DateTimeImmutable
* GMP
* DateTimeZone

Yes, please disable the error for them.

weirdan:
Why allow comparing resources though?

Issue #10837, opened 2024-03-16
self, parent and static in callable are deprecated since PHP 8.2 and throw a notice


kkmuffme:
https://psalm.dev/r/24b9ea4a40

Issue #10823, opened 2024-03-13
Callable should only allow public methods if used as arg in a non-$this/self context


kkmuffme:
https://psalm.dev/r/7a6d5d9849

Issue #8509, opened 2022-09-24
Non-static methods called statically should report error


kkmuffme:
https://psalm.dev/r/8d8e02aefb

Issue #8508, opened 2022-09-24
Callable visibility: calling private methods as callable should raise error


kkmuffme:
https://psalm.dev/r/6c28e36db4

AndrolGenhald:
At the same time, it should be allowed if `do_action` were a member of `foo`.

PR #10839, opened 2024-03-16
Add misc missing errors for invalid callable methods

kkmuffme:
- Fix https://github.com/vimeo/psalm/issues/10837
- Fix https://github.com/vimeo/psalm/issues/10823
- Fix https://github.com/vimeo/psalm/issues/8509
- Fix https://github.com/vimeo/psalm/issues/8508

### DIFF
--- a/tests/CallableTest.php
+++ b/tests/CallableTest.php
@@ -1786,6 +1786,421 @@ class CallableTest extends TestCase
 
                     takesCallable(function() { return; });',
             ],
+            'callableMethodOutOfClassContextStaticPublic' => [
+                'code' => '<?php
+                /**
+                 * @param callable $callable
+                 * @return void
+                 */
+                function run($callable) {
+                    call_user_func($callable);
+                }
+
+                class Foo {
+                    public static function hello(): void {
+                        echo "hello";
+                    }
+                }
+
+                $foo = new Foo();
+                run(array($foo, "hello"));',
+            ],
+            'callableMethodOutOfClassContextNonStatic' => [
+                'code' => '<?php
+                /**
+                 * @param callable $callable
+                 * @return void
+                 */
+                function run($callable) {
+                    call_user_func($callable);
+                }
+
+                class Foo {
+                    public function hello(): void {
+                        echo "hello";
+                    }
+                }
+
+                $foo = new Foo();
+                run(array($foo, "hello"));',
+            ],
+            'callableClassStringArrayMethodOutOfClassContextStaticPublic' => [
+                'code' => '<?php
+                /**
+                 * @param callable $callable
+                 * @return void
+                 */
+                function run($callable) {
+                    call_user_func($callable);
+                }
+
+                class Foo {
+                    public static function hello(): void {
+                        echo "hello";
+                    }
+                }
+
+                run(array(Foo::class, "hello"));',
+            ],
+            'callableClassStringMethodOutOfClassContextStaticPublic' => [
+                'code' => '<?php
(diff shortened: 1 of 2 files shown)


tests/CallableTest.php
```diff
@@ -2222,7 +2637,7 @@ public function bar(): bool {
                     function foo($arg) {}
 
                     class A {
-                        public function bar(): bool {
+                        public static function bar(): bool {
```
weirdan, line 2640:
```suggestion
                        public function bar(): bool {
```

weirdan (approved):

weirdan:
Thanks! Especially for the ton of new tests.

VincentLanglet:
Hi @kkmuffme @weirdan, I think this PR introduce a false-positive.

I get errors on SonataMediaBundle as shown by 
https://github.com/sonata-project/SonataMediaBundle/actions/runs/8921432277/job/24501522954?pr=2452
on line like https://github.com/sonata-project/SonataMediaBundle/blob/1f8bee3c639c70eb431396716894fce20da640b9/src/Twig/Extension/FormatterMediaExtension.php#L48

But the TwigFunction doesn't require a static callable:
https://github.com/twigphp/Twig/blob/68e4e5545f85e0dc9f09dff7f9db473577234fb9/src/TwigFunction.php#L31-L34

VincentLanglet:
> Can you create a minimal reproduction on psalm.dev please that can be used as a unit test when fixing this issue?

I tried, but unfortunately I still didn't understood why I cannot reproduce it on psalm.dev with something like
https://psalm.dev/r/2d2959e71b

But it's easy to get the error on this Sonata branch https://github.com/sonata-project/SonataMediaBundle/pull/2452
by running `composer install` and `make  psalm`. I run psalm on every commit between 5.23.0 and 5.24.0 and the error comes with your commit.

> Can you perhaps link me to one example of how that callable is actually used then [twigphp/Twig@`68e4e55`/src/TwigFunction.php#L61C21-L61C32](https://github.com/twigphp/Twig/blob/68e4e5545f85e0dc9f09dff7f9db473577234fb9/src/TwigFunction.php#L61C21-L61C32) ?

Twig has a special behavior 
```
$compiler->raw(sprintf('$this->env->getRuntime(\'%s\')->%s', $callable[0], $callable[1]));
```
that's why the phpdoc is `callable|array{class-string, string}` in my case it should match with `array{class-string, string}`.

VincentLanglet:
> I don't have time to dive into this atm, but I think the underlying issue is that: PHP (& psalm) treat `array{class-string, string}` as callable (even before this PR, as it's used in various places) but the only way that this is callable is if the method ("string") is static. Otherwise this wouldn't be callable and a fatal error.
> 
> [3v4l.org/2ct4q](https://3v4l.org/2ct4q)

But as soon as you change to `callable|array`, it works for PHP https://3v4l.org/JZgNh

> There is no way to fix this for psalm really, bc it would create other issues if `array{class-string, string}` wouldn't be treated as callable anymore. It would be possible to make it work for specific cases where we have literal strings and can validate that the method exists, however there is no 100% way

I found a way https://github.com/vimeo/psalm/pull/10935

michnovka:
This PR causes this issue - https://github.com/vimeo/psalm/issues/10971 and reports incorrectly InvalidArgument

VincentLanglet:
> This PR causes this issue - #10971 and reports incorrectly InvalidArgument

Already fixed in https://github.com/vimeo/psalm/pull/10935

PR #10842, opened 2024-03-19
fix tests running with other than called PHP binary if called with a non-default PHP binary

kkmuffme:
e.g. you run the tests with a php83 executable but your default php is PHP 7.4, it will suddenly change while running the tests leading to false positive errors


jack-worman:
> @jack-worman please merge

I am not a maintainer :)

PR #10858, opened 2024-03-25
Don't emit MissingOverrideAttribute for implicit Stringable implementations

### LINKED ISSUE #10846: False positive #[Override] required for __toString()
It's not reproducible at psalm-dev, but here is a repro repository: https://github.com/zerkms/psalm-10846-repro

```php
<?php
declare(strict_types=1);

namespace Zerkms;

final readonly class Test
{
    public function __toString(): string
    {
        return '';
    }
}
```
```
ERROR: MissingOverrideAttribute - src/Test.php:8:5 - Method __toString should have the "Override" attribute
```

edsrzf:
Fixes #10846.


weirdan (approved):

weirdan:
Thanks!

PR #10859, opened 2024-03-25
Specify array return type of session_get_cookie_params

jorgsowa:
Specifies the return type of function `session_get_cookie_params`. The options have been changed in PHP 7.3.


weirdan:
Thanks!

PR #10861, opened 2024-03-25
Add more specific return type of session_status

jorgsowa:
Return value of this function may be only one of 3 session status constants.

### DIFF
--- a/dictionaries/CallMap.php
+++ b/dictionaries/CallMap.php
@@ -11121,7 +11121,7 @@ return [
 'session_set_save_handler' => ['bool', 'open'=>'callable(string,string):bool', 'close'=>'callable():bool', 'read'=>'callable(string):string', 'write'=>'callable(string,string):bool', 'destroy'=>'callable(string):bool', 'gc'=>'callable(string):bool', 'create_sid='=>'callable():string', 'validate_sid='=>'callable(string):bool', 'update_timestamp='=>'callable(string):bool'],
 'session_set_save_handler\'1' => ['bool', 'open'=>'SessionHandlerInterface', 'close='=>'bool'],
 'session_start' => ['bool', 'options='=>'array'],
-'session_status' => ['int'],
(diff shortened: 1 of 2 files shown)


weirdan:
I don't think constants are supported in callmaps.

jorgsowa:
@weirdan could they be supported? I can look into this if this would be helpful

weirdan:
You can put constant values there.

dictionaries/CallMap.php
```diff
@@ -11121,7 +11121,7 @@
 'session_set_save_handler' => ['bool', 'open'=>'callable(string,string):bool', 'close'=>'callable():bool', 'read'=>'callable(string):string', 'write'=>'callable(string,string):bool', 'destroy'=>'callable(string):bool', 'gc'=>'callable(string):bool', 'create_sid='=>'callable():string', 'validate_sid='=>'callable(string):bool', 'update_timestamp='=>'callable(string):bool'],
 'session_set_save_handler\'1' => ['bool', 'open'=>'SessionHandlerInterface', 'close='=>'bool'],
 'session_start' => ['bool', 'options='=>'array'],
-'session_status' => ['int'],
+'session_status' => ['PHP_SESSION_ACTIVE|PHP_SESSION_NONE|PHP_SESSION_ACTIVE'],
```
weirdan, line 11124:
```suggestion
'session_status' => ['1|2'],
```

weirdan: Also you had a duplicate there

weirdan (changes requested):

dictionaries/CallMap.php
jorgsowa: I took a different approach. Hope it works.

weirdan: That's clever, but I would prefer values though. Many of the constants in PHP are not really constants - e.g. `PHP_BINARY`, `PHP_INT_SIZE`, `PASSWORD_DEFAULT`, etc. Using constants in the callmap the way you did establishes an unsafe example for future contributors / maintainers who may not be necessarily aware of those details.

Issue #9584, opened 2023-03-30
@psalm-internal must not propagate to private functions


scapital:
https://psalm.dev/r/0657193833

As W\Worker::internalRun is private, so it can not be used not in W\Worker.

scapital:
@kkmuffme 

> It's possible to extend the class and then it's called from a different namespace, e.g. https://psalm.dev/r/4e660ece0a

If class is internal to some namespace/class, there are not ability to extend it in another namespace, your example has error
```
W\Worker is internal to C\Consumer but called from D\NewWorker
```
and that's ok.

> therefore I think it does make sense and this behavior should not be changed

Do you mean, it is ok, if some class is marked as internal, it does not has ability to call his own private methods?

PR #10866, opened 2024-03-27
document that @psalm-internal works for namespace + class too

kkmuffme:
as used in https://github.com/vimeo/psalm/issues/9584


weirdan:
It can even go as granular as methods (when used on methods): https://psalm.dev/r/83f0c9c098

weirdan:
I thought you may want to :grin:

weirdan (approved):

weirdan:
Thanks!

PR #10883, opened 2024-04-05
Update PHP 8.2 Call map delta with refined types for string comparison functions

gsteel:
Since PHP 8.2 the following string comparison functions now return `1|0|-1` rather than `int`:

- `strcasecmp`
- `strcmp`
- `strnatcasecmp`
- `strnatcmp`
- `strncasecmp`
- `strncmp`


gsteel:
Is there anything missing here to get this patch over the line?

orklah:
Thanks! Sorry, I missed your PR earlier

Issue #10890, opened 2024-04-09
jsonSerialize triggers PossiblyUnusedMethod even when json_encode is called


josephwynn-sc:
https://psalm.dev/r/9a3c68649b

```php
<?php

class JsonSerializableClass implements \JsonSerializable
{
    public function jsonSerialize(): array
    {
        return ["hello"];
    }
}

$x = new JsonSerializableClass();

echo json_encode($x);
echo json_encode([
    "nested" => "structure",
    "x" => $x,
]);
```

The psalm.dev link shows this as INFO but locally it's an ERROR (see below) 

```
Psalm 5.23.1@8471a896ccea3526b26d082f4461eeea467f10a4

ERROR: PossiblyUnusedMethod - test.php:5:21 - Cannot find any calls to method JsonSerializableClass::jsonSerialize (see https://psalm.dev/087)
    public function jsonSerialize(): array
```

Using the following config:

```xml
<?xml version="1.0"?>
<psalm errorLevel="3" findUnusedBaselineEntry="true" findUnusedCode="true">
    <projectFiles>
        <file name="test.php" />
    </projectFiles>
</psalm>
```

PR #10891, opened 2024-04-09
Ignore jsonSerialize for implementors of JsonSerializable

josephwynn-sc:
This is my proposed fix for #10890. It follows the same pattern as the special case for serialize/unserialize introduced in 02fd1659ef74a41e6c4a7dcbf52df60fe43862ea.


staabm:
Does the same problem exist for e.g. `Countable`-interface and `count()` method, or `Throwable` or other builtin interfaces?

if so, maybe just ignore all methods which implement a php-src builtin interface?

orklah:
Thanks!

PR #10902, opened 2024-04-11
Throw exception instead of silently logging issues occurred during scan

danog:
This is the first step to debug a nasty issue occurring while scanning MadelineProto.


orklah:
Thanks!

PR #10907, opened 2024-04-14
Add backed enum type annotation

### LINKED ISSUE #8268: Add way to annotate BackedEnum backing type
- Add template to `BackedEnum` interface
 - Add annotation to specify backing type when declaring enum (`@enum<int>` maybe?)
 - Check annotation against PHP backing type (can only be `string` or `int`, but you can't use `@enum<positive-int>` with `enum Foo: string`)
 - Check annotation against cases (`case Bar = 0` should show an issue for `@enum<positive-int>`)
 - The `BackedEnum` template needs to be filled in from the `@enum<[type]>` annotation so that `BackedEnum::from`, `::tryFrom`, and `::cases` use the correct type

michaelcozzolino:
This fixes #8268.

Basically I added the `/** @template T of int|string */` in the BackedEnum stub and then i managed to add issues when
the implemented template type mismatches with the backing type. 
// will give an error
```
                    // will give an error
                    /**
                     * @implements BackedEnum<string>
                     */
                    enum NumberEnum: int {
                        case Zero = 0;
                        case One = 1;
                    }
```
```
                    // won't give an error
                    /**
                     * @implements BackedEnum<non-negative-int>
                     */
                    enum NumberEnum: int {
                        case Zero = 0;
                        case One = 1;
                    }
```
Furhtermore, it's not mandatory to add the@implements annotation if the backing type is just int or string

```
                    // won't give an error
                    /**
                     * @implements BackedEnum<int>
                     */
                    enum NumberEnum: int {
                        case Zero = 0;
                        case One = 1;
                    }
```
```
                    // won't give an error
                    enum NumberEnum: int {
                        case Zero = 0;
                        case One = 1;
                    }
```
Further cases can be seen in the tests.
I am not sure this is the right implementation. Let me know if, in case, we can make some adjustments.

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -2592,6 +2592,8 @@ final class ClassAnalyzer extends ClassLikeAnalyzer
                 }
 
                 if ($is_string_case_value && $storage->enum_type === 'string') {
+                    $case_value = (string) $case_value;
+
                     if ($enum_implemented_type instanceof Type\Atomic\TNonEmptyString) {
                         if (trim($case_value) === '') {
                             IssueBuffer::maybeAdd(
@@ -2610,6 +2612,8 @@ final class ClassAnalyzer extends ClassLikeAnalyzer
                 }
 
                 if ($is_int_case_value && $storage->enum_type === 'int') {
+                    $case_value = (int) $case_value;
+
                     if ($enum_implemented_type instanceof Type\Atomic\TIntRange) {
                         if ($enum_implemented_type->isPositive() && $case_value < 1) {
                             IssueBuffer::maybeAdd(
--- a/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassAnalyzer.php
@@ -74,6 +74,7 @@ use Psalm\Storage\ClassLikeStorage;
 use Psalm\Storage\FunctionLikeParameter;
 use Psalm\Storage\MethodStorage;
 use Psalm\Type;
+use Psalm\Type\Atomic;
 use Psalm\Type\Atomic\TGenericObject;
 use Psalm\Type\Atomic\TMixed;
 use Psalm\Type\Atomic\TNamedObject;
@@ -84,6 +85,7 @@ use Psalm\Type\Union;
 use UnexpectedValueException;
 
 use function array_filter;
+use function array_key_first;
 use function array_keys;
 use function array_map;
 use function array_merge;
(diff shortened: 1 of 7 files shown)


michaelcozzolino:
@orklah @danog @weirdan please review!

staabm:
I wonder why this phpdoc typing is necessary. 

All enum values are known at static analysis time, so why doesn't psalm just turn the `->value`s type into a union of all case-types?

michaelcozzolino:
> I wonder why this phpdoc typing is necessary.
> 
> All enum values are known at static analysis time, so why doesn't psalm just turn the `->value`s type into a union of all case-types?

and how will you know the type of the enum you want to infer? let's say you want an enum having only non-empty-string. how would you know this?

staabm:
so you want a type which accepts any enum which is backed by a string? I can't think of a use-case for such a vague type. couldn't you accept any `literal-string` then? whats special about a string when it comes from a enum?

michaelcozzolino:
> so you want a type which accepts any enum which is backed by a string?

well, when you do enum MyEnum: string{...} aren't your already doing it? i just want to be strict about the types to pass in an enum, i would like to have enum having non empty strings as i would have classes with properties that are non empty string

staabm:
you can't pass a type into a enum. every enum has a native type per default and cannot be defined in a wrong way without a fatal runtime error, e.g. https://3v4l.org/VlZAK

I don't see which information psalm could take from the phpdoc which is not already there by the raw definition.

don't get me wrong: I am not a psalm developer, so I just share my thoughts.
lets see what psalm guys say about said.

michaelcozzolino:
> from the phpdoc which is not already there by th

Checking that the backed type is the same as the case value is not the main goal of this pr even if it's a subcase of it. Actually the main goal of the pr is to say that you want to infer a subtype of the backed type in order to have an enum whose case values respect both types.
So in case of 
```/**
                     * @implements BackedEnum<non-negative-int>
                     */
                    enum NumberEnum: int {
                        case Zero = 0;
                        case One = 1;
                       case MinusOne = -1
                    }
```

you would get an error.
your example, as i already said is a subcase of this pr, but it does not mean that it should not be considered. At the end psalm will still give you errors for lines of code that would probably generate a fatal error, such as usages of non existing methods.

orklah:
My main issue here is that it will make any BackedEnum mandatory to be templated. If PHPStan doesn't, it will lead to complains about not being able to use both. It could be great to discuss it with them to know if it can be implemented on their side (and if they agree with the feature)

michaelcozzolino:
> My main issue here is that it will make any BackedEnum mandatory to be templated. If PHPStan doesn't, it will lead to complains about not being able to use both. It could be great to discuss it with them to know if it can be implemented on their side (and if they agree with the feature)

they suggested to use custom stubs to achieve that 
https://github.com/phpstan/phpstan/issues/10981
at this point, i suppose we should add a new annotation such as @psalm-enum, what do you think?

PR #10912, opened 2024-04-18
Fix conditional on non empty literal string

### LINKED ISSUE #10911: `literal-string is non-empty-literal-string` seems to be considered as always true
Hi, @orklah @weirdan 

I was trying to fix https://github.com/vimeo/psalm/issues/10909
and discovered an issue with conditional type which does block me.

https://psalm.dev/r/2d6c0230ed

```
 * @return (
 *    $s is non-empty-literal-string ? float : int
 * )
```
seems to give the same result than
```
 * @return (
 *    $s is literal-string ? float : int
 * )
```
like `literal-string is non-empty-literal-string` was always true...

This is the expected behavior https://phpstan.org/r/0ebe5412-7b05-4dd4-9ca3-2e5724c18c4f

Any idea where the fix need to be done ?

VincentLanglet:
Closes #10911


orklah:
Thanks!

PR #10923, opened 2024-04-27
Add `mail` to impure functions list

### LINKED ISSUE #10918: UnusedFunctionCall false positive on mail standard extension
When calling `mail` as [documented in the PHP manual](https://www.php.net/manual/en/function.mail), Psalm throws an UnusedFunctionCall because the boolean return value is not captured and utilized.

https://psalm.dev/r/79996b7e5c

smaddock:
Fixes #10918.


orklah:
Thanks!

Issue #10926, opened 2024-04-28
literal-string is non-empty-string seems to be considered as always true


VincentLanglet:
Similar to https://github.com/vimeo/psalm/issues/10911

The reproducer can be https://psalm.dev/r/e9071d2408

PR #10927, opened 2024-04-28
Fix condition literal-string is non-empty-string

VincentLanglet:
Hi @orklah, this PR is similar to https://github.com/vimeo/psalm/pull/10912

Closes https://github.com/vimeo/psalm/issues/10926

### DIFF
--- a/src/Psalm/Internal/LanguageServer/LanguageServer.php
+++ b/src/Psalm/Internal/LanguageServer/LanguageServer.php
@@ -400,7 +400,8 @@ final class LanguageServer extends Dispatcher
             $this->path_mapper->configureClientRoot($this->getPathPart($rootUri));
         }
 
-        return call(
+        /** @var Promise<InitializeResult> $promise */
+        $promise = call(
             /** @return Generator<int, true, mixed, InitializeResult> */
             function () use ($workDoneToken) {
                 $progress = $this->client->makeProgress($workDoneToken ?? uniqid('tkn', true));
@@ -578,6 +579,8 @@ final class LanguageServer extends Dispatcher
                 return new InitializeResult($serverCapabilities, $initializeResultServerInfo);
             },
         );
+
+        return $promise;
     }
 
     /**
--- a/src/Psalm/Internal/Type/ArrayType.php
+++ b/src/Psalm/Internal/Type/ArrayType.php
@@ -35,8 +35,10 @@ final class ArrayType
 
     /**
      * @return (
-     *     $type is TArrayKey ? self : (
-     *         $type is TArray ? self : null
+     *     $type is TKeyedArray ? self : (
+     *         $type is TNonEmptyArray ? self : (
+     *             $type is TArray ? self : null
+     *         )
      *     )
      * )
      */
--- a/src/Psalm/Internal/Type/TemplateInferredTypeReplacer.php
+++ b/src/Psalm/Internal/Type/TemplateInferredTypeReplacer.php
@@ -442,19 +442,11 @@ final class TemplateInferredTypeReplacer
                         || $conditional_type->getId() !== 'float')
                 ) {
                     $matching_if_types[] = $candidate_atomic_type;
-                } elseif (!UnionTypeComparator::isContainedBy(
-                    $codebase,
-                    $conditional_type,
+                } elseif (null === Type::intersectUnionTypes(
                     new Union([$candidate_atomic_type]),
-                    false,
-                    false,
-                    null,
-                    false,
-                    false,
-                )
-                    && ($candidate_atomic_type->getKey() !== 'string'
-                        || false === strpos($conditional_type->getKey(), 'string'))
-                ) {
+                    $conditional_type,
+                    $codebase
+                )) {
                     $matching_else_types[] = $candidate_atomic_type;
                 }
             }
--- a/src/Psalm/Internal/Type/TemplateInferredTypeReplacer.php
+++ b/src/Psalm/Internal/Type/TemplateInferredTypeReplacer.php
@@ -453,7 +453,7 @@ final class TemplateInferredTypeReplacer
                     false,
                 )
                     && ($candidate_atomic_type->getKey() !== 'string'
-                        || $conditional_type->getKey() !== 'string')
+                        || false === strpos($conditional_type->getKey(), 'string'))
                 ) {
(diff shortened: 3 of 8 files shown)


src/Psalm/Internal/Type/TemplateInferredTypeReplacer.php
```diff
@@ -451,7 +451,10 @@ private static function replaceConditional(
                     null,
                     false,
                     false,
-                )) {
+                )
```
VincentLanglet, line 454: The logic is

When doing `A is B ? C : D`:
- If A is contained by B => It's C
- If B is contained by A => It's D
- Else, it's C|D

I think this logic is wrong and does not work well with string type like `literal-string` and `non-empty-string`.
it should be
- If A is contained by B => It's C
- If A&B = NEVER => It's D
- Else it's C|D

But I dunno if a check `A&B=Never` already exist for psalm, so I added a special check for string instead.

src/Psalm/Internal/Type/TemplateInferredTypeReplacer.php
```diff
@@ -442,15 +442,10 @@ private static function replaceConditional(
                         || $conditional_type->getId() !== 'float')
                 ) {
                     $matching_if_types[] = $candidate_atomic_type;
-                } elseif (!UnionTypeComparator::isContainedBy(
-                    $codebase,
-                    $conditional_type,
+                } elseif (null === Type::intersectUnionTypes(
```
VincentLanglet, line 445: The logic is

When doing `A is B ? C : D`:
- If A is contained by B => It's C
- If B is contained by A => It's D
- Else, it's C|D

I think this logic is wrong and does not work well with string type like `literal-string` and `non-empty-string`.
it should be
- If A is contained by B => It's C
- If A&B = NEVER => It's D
- Else it's C|D

src/Psalm/Internal/Type/ArrayType.php
```diff
@@ -35,8 +35,10 @@ public function __construct(Union $key, Union $value, bool $is_list, ?int $count
 
     /**
      * @return (
-     *     $type is TArrayKey ? self : (
```
VincentLanglet, line 38: `TArrayKey` doesn't exist.

As soon as I fixed `TemplateInferredTypeReplacer.php`, the error was detected.

src/Psalm/Internal/Type/TemplateInferredTypeReplacer.php
```diff
@@ -451,6 +451,12 @@ private static function replaceConditional(
                     null,
                     false,
                     false,
+                ) && null === Type::intersectUnionTypes(
```
VincentLanglet, line 454: The logic was

When doing `A is B ? C : D`:
- If A is contained by B => It's C
- If B is not contained by A => It's D
- Else, it's C|D

I think this logic is wrong and does not work well with string type like `literal-string` and `non-empty-string`.
`literal-string is non-empty-string ? int : float`.
Since `non-empty-string` is not contained by `literal-string` it was considering the result as a float.
But since `literal-string&non-empty-string` is non empty, the result should have been `int|float`.

src/Psalm/Type.php
```diff
@@ -787,8 +789,26 @@ public static function intersectUnionTypes(
 
             //if a type is contained by the other, the intersection is the narrowest type
             if (!$intersection_performed) {
-                $type_1_in_2 = UnionTypeComparator::isContainedBy($codebase, $type_1, $type_2);
-                $type_2_in_1 = UnionTypeComparator::isContainedBy($codebase, $type_2, $type_1);
+                $type_1_in_2 = UnionTypeComparator::isContainedBy(
```
VincentLanglet, line 792: When I tried to use `intersectUnionTypes` with `$allow_float_int_equality = false` I got a bug.

Seems like those parameters were not passed to every calls.

src/Psalm/Internal/LanguageServer/LanguageServer.php
```diff
@@ -400,7 +400,8 @@ public function initialize(
             $this->path_mapper->configureClientRoot($this->getPathPart($rootUri));
         }
 
-        return call(
+        /** @var Promise<InitializeResult> $promise */
+        $promise = call(
```
VincentLanglet, line 404: This avoid
```
The type 'Amp\Promise<LanguageServerProtocol\InitializeResult>|Amp\Promise<mixed>' is more general than the declared return type 'Amp\Promise<LanguageServerProtocol\InitializeResult>' for Psalm\Internal\LanguageServer\LanguageServer::initialize
```

This is technically true because the conditional return type is
```
(T is Promise ? false : (T is \Generator ? (TGenerator is Promise ? Promise<TGeneratorPromise> : Promise<TGeneratorReturn>) : int))
```
and the check `InitializeResult is Promise` cannot be resolved since `InitializeResult&Promise` would be a possible type.

VincentLanglet:
The only error I get now is coming from the test https://github.com/vimeo/psalm/pull/10927/files/91ea61e4170656cb7a54347271d054e9037106d6#diff-19d1ea03f9fa43528c7190e1c91a6a4f216c48b742e3e2dcdd4e75e750c6a7a5

But this failure also happen in 5.x as soon as I fix the phpdoc of the tests
https://github.com/vimeo/psalm/pull/10929
I search a long time and dunno why psalm is failing to get the right type.

Should I comment the test in a separate PR @orklah since the test is wrong as shown by https://github.com/vimeo/psalm/pull/10929 ?

VincentLanglet:
> The only error I get now is coming from the test [vimeo/psalm/pull/10927/files/91ea61e4170656cb7a54347271d054e9037106d6#diff-19d1ea03f9fa43528c7190e1c91a6a4f216c48b742e3e2dcdd4e75e750c6a7a5](https://github.com/vimeo/psalm/pull/10927/files/91ea61e4170656cb7a54347271d054e9037106d6#diff-19d1ea03f9fa43528c7190e1c91a6a4f216c48b742e3e2dcdd4e75e750c6a7a5)
> 
> But this failure also happen in 5.x as soon as I fix the phpdoc of the tests #10929 I search a long time and dunno why psalm is failing to get the right type.
> 
> Should I comment the test in a separate PR @orklah since the test is wrong as shown by #10929 ?

Any recommendation about such situation @weirdan ?

PR #10934, opened 2024-05-02
redis: add possible types for `Redis#auth` method

boesing:
With v6.0.0, `Redis#auth` also accepts an array with 2 values where the first value represents the `user` and the second `password`.

I am not sure if they accept `array` as of 6.0 but as of the "official" stub, `auth` does also accept an array with just the password:
https://github.com/phpredis/phpredis/blob/f350dc342cb3520b2ee663664a22c25f29bc8aaf/redis.stub.php#L581-L594

### DIFF
--- a/stubs/extensions/redis.phpstub
+++ b/stubs/extensions/redis.phpstub
@@ -36,7 +36,8 @@ class Redis {
 	/** @return false|int|Redis */
     public function append(string $key, string $value) {}
 
-    public function auth(mixed $credentials): bool {}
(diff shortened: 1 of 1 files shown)


orklah:
I'm not sure Psalm can correctly process the attribute here, can you remove it just in case?

boesing:
ah I can, yes. Was thinking it might have benefits for tentative thingy.
will drop that somewhen the next days

weirdan:
Thanks!

PR #10935, opened 2024-05-03
Avoid false positive about array which are non-callable

VincentLanglet:
From discussion https://github.com/vimeo/psalm/pull/10839#issuecomment-2090041036

This should be valid, cf https://3v4l.org/JZgNh

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
@@ -63,6 +63,7 @@ use Psalm\Type\Atomic\TNamedObject;
 use Psalm\Type\Union;
 use UnexpectedValueException;
 
+use function array_filter;
 use function count;
 use function explode;
 use function implode;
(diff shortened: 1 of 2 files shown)


VincentLanglet:
> I get the point of this PR and it's good, the question is since the `array{class-string, string}` type generally indicates a callable.

`generally` but not always. If someone wants to indicate a callable, he HAVE TO use the right type `callable`.

> Are we ok with it not catching obvious errors anymore in that case?

The opposite question would be "Are we ok to return false positive for people using correctly the phpdoc in order to keep catching errors for people which used the wrong phpdoc". I would say NEVER.

And if we're doing nothing we'll end with a lot of issue like https://github.com/vimeo/psalm/issues/10936 which makes psalm unusable with Symfony/Twig.

I agree catching a wrong method could be great, but first the false-positive need to be fixed.

VincentLanglet:
> Agree, 2 things that come to my mind at this point:
> 
> * if we do that for arrays, this should also be changed for strings (e.g. callable|string) since it's the same issue (in the PR/same file just a bit further down in the elseif). Can you update this in the PR?
> * can you confirm that this will still report an error with this PR:
> 
> ```
> /**
>  * @param int|callable $arg
>  */
> function foo($arg): void {}
> foo([\DateTime::class, "errorHere"]);',
> ```

Both are added in the last commit.

weirdan (approved):

weirdan:
Thanks!

michnovka:
This PR did not fix the issue with setting protected function as callable. 

https://psalm.dev/r/3c405968f1

This code reports on error on psalm.dev, as it uses outdated version there. But with 5.24 and 5.25 it complains that 

> setCode expects a public callable, but a non-public callable provided (see https://psalm.dev/004)

However setting a protected function this way is totally fine. The issue would be setting a private function

Issue #10940, opened 2024-05-03
Casting int range should keep literals


kkmuffme:
https://psalm.dev/r/b6f6d6b075

For a reasonable int range size, casting an int range should behave like casting int's that aren't in a range.

PR #10941, opened 2024-05-03
Casting int-range should keep literals

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/10940
500 limit taken from TypeCombiner


weirdan (approved):

weirdan:
Thanks!

PR #10955, opened 2024-05-05
PHAR: add `config.xsd` to generated PHAR package

boesing:
In most projects I am working on, the `config.xsd` is referenced from within the projects `psalm.xml` (via `xsi:noNamespaceSchemaLocation`).
That provides decent auto-completion, etc. and thus is usually recommended to represent the latest configuration schema.
That was done by having `vendor/vimeo/psalm/config.xsd` as that schema is located for `vimeo/psalm` usage.

When using `psalm/phar`, that does not work as the file is not part of the pacakge. This PR is changing that and will add the `config.xsd` with whatever next version is generated after merging these changes.


weirdan:
Duplicate of #10938

Issue #10957, opened 2024-05-05
CI: phpunit should fail on deprecated notices


kkmuffme:
If psalm itself triggers PHP notices in CI these aren't reported and the tests don't fail
e.g. https://github.com/vimeo/psalm/issues/10950 that issue would have been caught 3 years ago and I wonder how nobody has encountered that notice when running CI locally with a stricter error handler :)

Issue #10950, opened 2024-05-04
Psalm PHP notice when using modulo, bitwise or,... on non-int


kkmuffme:
https://psalm.dev/r/e2fc45cb56

When analyzing this code with psalm triggers a PHP notice
>Deprecated: Implicit conversion from float ... to int loses precision

Issue #10949, opened 2024-05-04
Report error for float used with modulo


kkmuffme:
No error reported https://psalm.dev/r/7f4bfa83e6
PHP triggers notice https://3v4l.org/Cph3O
>Deprecated: Implicit conversion from float 25.4 to int loses precision

PR #10958, opened 2024-05-05
CI should fail on PHP deprecated errors and report errors for deprecated implicit float to int precision loss

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/10957
Fix https://github.com/vimeo/psalm/issues/10950
Fix https://github.com/vimeo/psalm/issues/10949

Update parts of CI to use PHP 8.3 instead of older PHP versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,7 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
+          - "7.4"
           - "8.0"
           - "8.1"
           - "8.2"
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.0'
+          php-version: '7.4'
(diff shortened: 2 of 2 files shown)


staabm:
In my experience you also need to set a proper php.ini in the github action. 

See e.g. https://github.com/nikic/PHP-Parser/commit/f2e037f8ea4c4bac1a5a36b4d8334f2aa581491a

.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.4'
+          php-version: '8.3'
```
weirdan, line 60: Please revert this. For 5.x branch our lowest supported version is 7.4, and PHPCS (Slevomat coding standard) uses the version it runs on to customize some sniffs. Thus it should match the lowest version we support.

.github/workflows/windows-ci.yml
```diff
@@ -54,8 +54,8 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.0'
-          ini-values: zend.assertions=1, assert.exception=1, opcache.enable_cli=1, opcache.jit=function, opcache.jit_buffer_size=512M
+          php-version: '8.3'
```
weirdan, line 57: would actually prefer this lowered to 7.4

PR #10959, opened 2024-05-05
Fix int-range concat to behave like union of int types

### LINKED ISSUE #10947: int-range behaves different than literal int even when identical
https://psalm.dev/r/639ce43827

### LINKED ISSUE #10965: int range returns incorrect result for pow
https://psalm.dev/r/05888237f2

Should be identical, but it's not.
There's a test for it (IntRangeTest pow) but it's wrong there too, as it asserts -1 incorrectly.

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/10947
Fix https://github.com/vimeo/psalm/issues/10965
500 limit taken from TypeCombiner (to ensure int-range will behave exactly like union of ints)


orklah:
Can you rebase on 6.x?

PR #10960, opened 2024-05-05
Fix wrong and inconsistent PHP native types in atomic

kkmuffme:
See https://www.php.net/manual/en/language.types.declarations.php


src/Psalm/Type/Atomic/TIterable.php
```diff
@@ -101,7 +101,9 @@ public function toPhpString(
 
     public function canBeFullyExpressedInPhp(int $analysis_php_version_id): bool
     {
-        return $this->type_params[0]->isMixed() && $this->type_params[1]->isMixed();
+        return $analysis_php_version_id >= 7_01_00
```
SCIF, line 104: Does `7.1` actually should be supported?

PR #10969, opened 2024-05-08
Precise preg_match_all return type

VincentLanglet:


VincentLanglet:
> afaik positive-int is being deprecated, should be `int<0, max>` where you use that in a union with 0

I updated the notation then

VincentLanglet:
Friendly ping @weirdan @orklah if you have time to review :)

stubs/CoreGenericFunctions.phpstub
```diff
@@ -1249,7 +1249,7 @@ function preg_replace_callback($pattern, $callback, $subject, int $limit = -1, &
  *              )
  *          )
  *        ) $matches
- * @return int|false
+ * @return 0|positive-int|false
```
weirdan, line 1252: `@return int<0,max>|false` perhaps?

VincentLanglet: Indeed, changed

weirdan:
Thanks!

PR #10989, opened 2024-05-24
Fix false `MissingOverrideAttribute` issue

### LINKED ISSUE #10982: MissingOverrideAttribute false positive with traits used by child classes
https://psalm.dev/r/70ba038824

If you do what psalm says you're supposed to do and add the `Override` attribute to the trait method, you will create invalid code that PHP refuses to run (and psalm also flags as an error).

I think `B` should be totally free to import its own copy of the `Foo` trait.

ghostwriter:
This patch resolves #10982, via the following changes:
 
 - Add test case `tests/OverrideTest.php`
 -  Update `src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php`

### DIFF
--- a/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
@@ -1995,6 +1995,7 @@ abstract class FunctionLikeAnalyzer extends SourceAnalyzer
                 if (!$has_override_attribute
                     && $codebase->config->ensure_override_attribute
                     && $overridden_method_ids
+                    && $storage->visibility !== ClassLikeAnalyzer::VISIBILITY_PRIVATE
                     && $storage->cased_name !== '__construct'
                     && ($storage->cased_name !== '__toString'
                        || isset($appearing_class_storage->direct_class_interfaces['stringable']))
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/FunctionLikeAnalyzer.php
```diff
@@ -1995,6 +1995,7 @@ private function getFunctionInformation(
                 if (!$has_override_attribute
                     && $codebase->config->ensure_override_attribute
                     && $overridden_method_ids
+                    && $storage->visibility !== ClassLikeAnalyzer::VISIBILITY_PRIVATE
```
edsrzf, line 1998: Unfortunately this doesn't fully fix the issue.

Even if the trait method is `protected` or `public`, there's a false positive: https://psalm.dev/r/4dce1a9157 And in that code, it would still be incorrect to add an `Override` attribute, since the method in `A` doesn't override anything. The attribute causes an error at runtime.

I think ideally we'd not consider methods that come from traits, but I can't see an easy way to check for this.

ghostwriter: @edsrzf Thank you for reviewing the Pull request,

You're right!

> Methods from a used trait behave as if the method definition was copied and pasted into the target class. Specifically the `#[\Override]` attribute on a trait method requires the existence of a matching method in a parent class or implemented interface.

I'll take a closer look.

edsrzf (approved):
LGTM, though I'm not a maintainer, so we need to wait for one before merging.

ghostwriter:
Resolved in https://github.com/vimeo/psalm/pull/11284

PR #10999, opened 2024-06-02
JetBrains IDE SARIF support improvement

llaville:
Hello,

As PHPStorm user, I've noticed that auto-detection of SARIF reports are not well handled when file report extension is only `.sarif`

> [!TIP] 
> See appendix M for files extension supported : 
> https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141791204

With this PR, the `.sarif.json` extension is also supported and PHPStorm integration is better.

See screenshots 

With only `.sarif` extension 

psalm sarif

With `.sarif.json` extension

psalm sarif json


llaville:
Consider to use PR #11000 that include this fix !

PR #11009, opened 2024-06-07
Improve mysql fetch_field* return type

MoonE:
While a table column name cannot be null, it is still possible to use an empty string as an alias. In this case name is an empty string.
```sql
SELECT'';
```
`def` and `catalog` are constants, and `max_length` is always `0` since php 8.1
https://www.php.net/manual/en/mysqli-result.fetch-field.php#refsect1-mysqli-result.fetch-field-returnvalues


weirdan (approved):

weirdan:
Thanks!

PR #11010, opened 2024-06-08
Fix signature of Locale::canonicalize.

ADmad:
The method can return null, https://www.php.net/manual/en/locale.canonicalize.php.


weirdan:
Indeed: https://3v4l.org/Bco3Y

weirdan:
This needs more work. See https://psalm.dev/docs/contributing/editing_callmaps/#typical-changes

weirdan:
Since the return type now matches the reflection, please remove the signature from `InternalCallMapHandlerTest::$ignoredReturnTypeOnlyFunctions`

weirdan:
Thanks!

PR #11025, opened 2024-06-24
Add source as parent node in CodeBase::addTaintSource()

### LINKED ISSUE #10057: Adding custom taint sources via plugin ignored

cgocast:
Fixes #10057

### DIFF
--- a/src/Psalm/Codebase.php
+++ b/src/Psalm/Codebase.php
@@ -2453,6 +2453,9 @@ final class Codebase
 
         $this->taint_flow_graph->addSource($source);
 
+        /** @psalm-suppress InaccessibleProperty */
+        $expr_type->parent_nodes[$source->id] = $source;
+
         return $expr_type->addParentNodes([$source->id => $source]);
     }
 


cgocast:
This bugfix is not perfect but it is much simpler than #10206.

@orklah, please can you have a look at this PR ?

Patrick-Remy:
Maybe both changes are required, in my PR I am also adding massive Taint-related tests and fixing examples. 
Anyway, in my PR I still had issues that some taints weren't found where I hadn't found a solution for.
I will check if merging/applying your change into mine will fix the outstanding failing tests!

cgocast:
Indeed my PR lacks taint-related tests and I agree we should merge your tests.

I am not convinced by the changes you made in the Analyzers. For instance, I was able to raise a `TaintedHtml` on 
```
<?php // --taint-analysis

$foo = $bad_data;
echo $foo; 
```
without changing the Analyzers

Patrick-Remy:
Even rebased and with your commit applied, the test you mentioned is failing in the CI: https://github.com/vimeo/psalm/actions/runs/9747343000/job/26899836333?pr=10206

Therefor I will revert your change from my branch again, as mentioned in the PR description `addTaintSource()` was rewritten to immutability / purity. Adding again a side-effect there is a bit anti-design in my opinion.

Patrick-Remy:
@cgocast I think you can close this PR, as #10206 is merged and released

PR #11042, opened 2024-07-11
Use more precise return types for getimagesize* functions

MidnightDesign:
- Key 0 (the image width) is always positive
- Key 1 (the image height) is always positive
- Key 2 is always one of the `IMAGETYPE_*` constants


MidnightDesign:
Is there an easy way to add tests for these signature changes?

weirdan:
You also need to update either the delta files or the historic callmap. See https://psalm.dev/docs/contributing/editing_callmaps/

PR #11081, opened 2024-08-21
Consistently emit issues for properties on classes with unknown mixins

issidorov:
The tests being added are similar to those just above in the code.


weirdan (approved):

weirdan:
Thanks!

PR #11100, opened 2024-09-19
Fix return type of IteratorIterator::getInnerIterator

alcaeus:
When creating an `IteratorIterator` with an iterator of `TIterator`, we can assume that `getInnerIterator` will return that same `TIterator` type. Technically, this change could be applied to `OuterIterator`, although that would require us to introduce a `TIterator` template to `OuterIterator`, which may break existing usages beyond the default iterators for which stubs are provided. Please advise if it's ok to change `OuterIterator` to get an additional template for the iterator it's wrapping.

### DIFF
--- a/stubs/CoreGenericIterators.phpstub
+++ b/stubs/CoreGenericIterators.phpstub
@@ -127,6 +127,11 @@ class IteratorIterator implements OuterIterator {
      * @psalm-ignore-nullable-return
      */
     public function key() {}
+
+    /**
+     * @return TIterator
+     */
+    public function getInnerIterator();
 }
 
 /**


GromNaN:
Thanks for the PR! The fix is correct.

A test can be added:

```diff
--- a/tests/Template/ClassTemplateExtendsTest.php
+++ b/tests/Template/ClassTemplateExtendsTest.php
@@ -3766,6 +3766,26 @@ final class ClassTemplateExtendsTest extends TestCase
                          public abstract function accept () {}
                     }',
             ],
+            'iteratorIteratorGetInnerIteratorReturnsTIterator' => [
+                'code' => '<?php
+                    /**
+                     * @implements Iterator<int, string>
+                     */
+                    class MyIterator implements Iterator {
+                        public function current(): string { return ""; }
+                        public function key(): int { return 0; }
+                        public function next(): void {}
+                        public function rewind(): void {}
+                        public function valid(): bool { return false; }
+                    }
+
+                    $inner = new MyIterator();
+                    $iter = new IteratorIterator($inner);
+                    $result = $iter->getInnerIterator();',
+                'assertions' => [
+                    '$result' => 'MyIterator',
+                ],
+            ],
```

PR #11108, opened 2024-10-01
chore: amp package version upgrade to 3.x

sankarsuda:
I am facing challenge when integration psalm with grumphp which require amp version of 3.x


mougrim:
Hi all!

@weirdan, is there way to merge it?

Do you need any help here?

orklah:
Migrating to Psalm 6 should solve the issue now :)

PR #11109, opened 2024-10-01
Update method signatures for ext-mongodb 1.20

alcaeus:
This includes a tiny fix for the template annotation in `MongoDB\BSON\Iterator`, as well as updates to the stubs for `MongoDB\Driver\CursorInterface` and `MongoDB\Driver\Cursor` to prepare for the upcoming 2.0 release.

### DIFF
--- a/stubs/extensions/mongodb.phpstub
+++ b/stubs/extensions/mongodb.phpstub
@@ -264,6 +264,7 @@ namespace MongoDB\Driver
         {
         }
 
+        /** @return ($asInt64 is true ? Int64 : CursorId)
         public function getId(bool $asInt64 = false): Int64|CursorId
         {
         }


stubs/extensions/mongodb.phpstub
```diff
@@ -207,44 +235,48 @@ namespace MongoDB\Driver
...
-        /**
-         * @return array<TValue>
-         */
-        public function toArray()
+        /** @return array<TValue> */
+        public function toArray(): array
+        {
+        }
+
+        public function getId(bool $asInt64 = false): Int64|CursorId
```
GromNaN, line 267: You can improve the return type. The default return type change in 2.0. If we could specify the return type depending on the ext-mongodb version, that would be nice.
```suggestion
        /** @return ($asInt64 is true ? Int64 : CursorId) */
        public function getId(bool $asInt64 = false): Int64|CursorId
```

alcaeus: I thought I already had this in, but I didn't add it in our own stubs. It's added now 👍

orklah:
could you rebase on 6.x?

alcaeus:
@orklah sorry for the late reply. Do we still need to maintain CallMap entries, or do we only need the stubs for 6.x? I noticed that the file is no longer there.

danog:
The file was moved, see https://psalm.dev/docs/contributing/editing_callmaps/

PR #11134, opened 2024-10-20
Fix parameter name in SplObjectStorage

### LINKED ISSUE #11133: ::attach has wrong name $info, expecting $data as defined by SplObjectStorage::attach
https://psalm.dev/r/a10fdd663a

I was expecting no error as the parameter name is `$info` which is exactly what it should be. https://www.php.net/manual/en/splobjectstorage.attach.php Where is `$data` coming from?

kamil-tekiela:
Fixes #11133


orklah:
Can you rebase on 6.x?

PR #11136, opened 2024-10-21
Die when PHP doc not found in update-property-map.php

M393:
Otherwise an exception is thrown right below, where `$docDir` is used.


orklah:
Can you rebase on 6.x?

orklah:
Thanks!

PR #11138, opened 2024-10-21
Fix return type of ReflectionEnum::getBackingType for PHP 8.2

M393:
https://www.php.net/manual/en/reflectionenum.getbackingtype.php#refsect1-reflectionenum.getbackingtype-changelog


orklah:
Can you rebase on 6.x?

PR #11140, opened 2024-10-23
start making curl types more specific

kkmuffme:
Start to make curl param and return types more specific.

Things that still need improving:
- curl_getinfo
- curl_setopt
- param types where many are listed can potentially be more restricted when one goes through the actual constant values and their meaning. I just used the widest set possible according to the docs now.

Note: for the return type https://curl.se/libcurl/c/libcurl-errors.html needs to be used, bc the PHP docs https://www.php.net/manual/en/curl.constants.php are missing various values (the non-listed values can be returned, but not used as param, that's maybe why some of them are missing in the docs, but not sure)


orklah:
Can you rebase on 6.x?

PR #11150, opened 2024-11-06
Fix array_reverse type for `list{0: int, 1?: int}` etc.

### LINKED ISSUE #11148: array_reverse on variable length list gives strange results
https://psalm.dev/r/090d9ad9ef

Wrong: `array{0?: 'b', 1: 'a'}`
Correct: `array{0: 'a'|'b', 1?: 'a'}`

MoonE:
Fix #11148

I'm not entirely sure if `$first_arg_array->properties` can have string keys? The type suggests it but not sure if it is possible here, because currently those would be lost with my implementation.


orklah:
Please rebase on 6.x so we can merge it :)

orklah:
Thanks!

PR #11152, opened 2024-11-09
Fix type of property-get with coalesce

### LINKED ISSUE #11151: `@property-read` dynamic property type not working sometimes
https://psalm.dev/r/f818e2d9d9

Type of `$_c` should be `string` instead of `''|mixed`.

MoonE:
Fixes #11151

If the property was not already known inside the IssetAnalyzer::analyze function, it was added as mixed to the context.

This adds a check for dynamic properties.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/IssetAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/IssetAnalyzer.php
@@ -7,6 +7,7 @@ namespace Psalm\Internal\Analyzer\Statements\Expression;
 use PhpParser;
 use Psalm\CodeLocation;
 use Psalm\Context;
+use Psalm\Internal\Analyzer\Statements\Expression\Fetch\InstancePropertyFetchAnalyzer;
 use Psalm\Internal\Analyzer\Statements\ExpressionAnalyzer;
 use Psalm\Internal\Analyzer\StatementsAnalyzer;
 use Psalm\Issue\InvalidArgument;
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/IssetAnalyzer.php
```diff
@@ -30,7 +31,16 @@ public static function analyze(
...
+                    if (isset($context->vars_in_scope['$this'])) {
+                        InstancePropertyFetchAnalyzer::analyze(
+                            $statements_analyzer,
+                            $isset_var,
+                            $context,
+                        );
+                    }
+                    if (!isset($context->vars_in_scope[$var_id])) {
+                        $context->vars_in_scope[$var_id] = Type::getMixed();
+                    }
```
MoonE, line 43: The return value of InstancePropertyFetchAnalyzer::analyze is discarded, not sure if/how it needs to be handled.

src/Psalm/Internal/Analyzer/Statements/Expression/IssetAnalyzer.php
```diff
@@ -30,7 +31,16 @@ public static function analyze(
                 $var_id = '$this->' . $isset_var->name->name;
 
                 if (!isset($context->vars_in_scope[$var_id])) {
-                    $context->vars_in_scope[$var_id] = Type::getMixed();
+                    if (isset($context->vars_in_scope['$this'])) {
```
MoonE, line 34: This check is necessary to keep this test passing:
https://github.com/vimeo/psalm/blob/3f52ded9c37a9ab227b8052e132223efd97acc44/tests/TypeReconciliation/ScopeTest.php#L165-L174

This checks if this is inside an object context and dynamic property access is possible.

orklah:
Thanks!

PR #11153, opened 2024-11-10
Fix some more tests when running phpunit with args

MoonE:
Related to  #11122

Running this failed because of the argv values
```sh
./vendor/bin/phpunit --filter '/^Psalm\\Tests\\VariadicTest/'
```


orklah:
Thanks!

PR #11155, opened 2024-11-11
Allow running generate_testsuites.php from a different dir

MoonE:
Now the script uses its own location as a basis instead of the current working dir.


orklah:
Thanks!

PR #11156, opened 2024-11-12
Fix object properties lose type

### LINKED ISSUE #9767: Object property type changed when used in if condition
non-empty-mixed is added to the type of `g`, but it should be unchanged `object|null`
https://psalm.dev/r/286d439d27

### LINKED ISSUE #10380: Accessing different nested object properties trigger MixedPropertyFetch
Hi,

I've noticed an issue when accessing multiple nested object properties.

Hopefully this snippet explains the issue:

```
<?php
/**
 * @var object{
 *     Nested: object{
 *         foo: string,
 *         bar: string,
 *    }
 * } $object
 */
if ($object->Nested->foo !== 'Foo') {
    return null;
}
if ($object->Nested->bar !== 'Bar') { // the MixedPropertyFetch is unexpected here
    return null;
}
```
Link to the snippet https://psalm.dev/r/4ec39c5d44

FYI this only happens with nested objects, accessing multiple properties directly on `$response` works as expected.

### LINKED ISSUE #11041: Psalm seems to forget (nested) type information after evaluation
Consider the following [snippet](https://psalm.dev/r/e128b3793a):

```php
/**
 * @psalm-type Foo = object{
 *     bar: object{
 *         baz: string|null,
 *     },
 * }
 */

/** @var Foo */
$foo = (object) [
	'bar' => (object) [
		'baz' => null,
	],
];

if ($foo->bar->baz === null) {
	$foo->bar->baz = 'foobarbaz';
	return;
}

$foo->bar->baz = 'foobarbaz';
```

As you can see, in lines 19 and 23, the exact same thing happens: We're setting a nested object property.
Up until line 19, everything is fine, and Psalm happily fetches the nested property value in line 18 and also allows to set the value in line 19. However, in line 23, it no longer seems to know the type of the `bar` property; it inferred `mixed`.

Now, there are a lot of different things you can do to break it even more, or partially unbreak it.

For example, moving line 23 before the `if` will not trigger anything.

```php
$foo->bar->baz = 'foobarbaz';

if ($foo->bar->baz === null) {
	$foo->bar->baz = 'foobarbaz';
	return;
}
```

Or storing the conditional in a variable will make the issue go away too:

```php
$isNull = $foo->bar->baz === null;
if ($isNull) {
	$foo->bar->baz = 'foobarbaz';
	return;
}

$foo->bar->baz = 'foobarbaz';
```

Funnily, using `is_null` instead of manually comparing with `null` will work, too, even when inline:

```php
if (is_null($foo->bar->baz)) {
	$foo->bar->baz = 'foobarbaz';
	return;
}

$foo->bar->baz = 'foobarbaz';
```

Or `empty()` or `!isset()` etc.

---

Is there anything I am doing wrong? Is Psalm?

Does that look like a bug to you too?

Happy to provide additional information, but it looks like Psalm discards (partial) information after a manual comparison inside a control structure (or something like that), and thus falls back to `mixed`, which then leads to various issues like `MixedPropertyAssignment`, `MixedPropertyFetch`, `MixedArrayAssignment`, `MixedOperand` etc.

MoonE:
Fixes #9767
Fixes #10380
Fixes #11041

### DIFF
--- a/src/Psalm/Type/Reconciler.php
+++ b/src/Psalm/Type/Reconciler.php
@@ -54,6 +54,7 @@ use Psalm\Type\Atomic\TNamedObject;
 use Psalm\Type\Atomic\TNever;
 use Psalm\Type\Atomic\TNull;
 use Psalm\Type\Atomic\TObject;
+use Psalm\Type\Atomic\TObjectWithProperties;
 use Psalm\Type\Atomic\TString;
 use Psalm\Type\Atomic\TTemplateParam;
 use ReflectionProperty;
@@ -845,6 +846,9 @@ class Reconciler
 
                         if ($existing_key_type_part instanceof TNull) {
                             $class_property_type = Type::getNull();
+                        } elseif ($existing_key_type_part instanceof TObjectWithProperties) {
+                            $class_property_type =
+                                $existing_key_type_part->properties[$property_name] ?? Type::getMixed();
                         } elseif ($existing_key_type_part instanceof TMixed
                             || $existing_key_type_part instanceof TObject
                             || ($existing_key_type_part instanceof TNamedObject
--- a/tests/ReturnTypeProvider/GetObjectVarsTest.php
+++ b/tests/ReturnTypeProvider/GetObjectVarsTest.php
@@ -24,6 +24,19 @@ class GetObjectVarsTest extends TestCase
             'assertions' => ['$ret' => 'array{prop: string}'],
         ];
 
+        yield 'retrurnsNotMixed' => [
+            'code' => '<?php
+                /**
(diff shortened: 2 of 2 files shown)


MoonE:
Wondering if this should default to `never` instead of `mixed`...

Is there a difference between `object` and `object{}`? As in one is an object which could have any properties the other is an object without any properties.
What about `object{p: int, ...}` and `object{p: int}`? Seems like the `...` part is ignored for objects.

orklah:
Thanks!

orklah:
I actually merged that if 5.x that won't have more releases. Could you propose this fix for 6.x?

PR #11157, opened 2024-11-12
Add missing reset of context inside_general_use

MoonE:
Everywhere else this property is reset before returning, just here it seems to be missing.
Not sure what the ramifications are as the tests seem to pass either way.

See  #7195


orklah:
Thanks!

PR #11168, opened 2024-11-25
Decrease opcache.jit_buffer_size to 128M

edsrzf:
As of PHP 8.4, this is the maximum value available for AArch64. Without this change, running Psalm with PHP 8.4 on that architecture results in a warning:

```
Warning: JIT on AArch64 doesn't support opcache.jit_buffer_size above 128M.
```

It's still possible to configure up to 2G for x86_64, so we could keep the 512M value on that architecture at the cost of a bit more complexity to detect the PHP version and current CPU architecture. (Looks like `php_uname('m')` would be the way to check CPU arch.)


edsrzf:
Based on the CI failures, maybe we need 512M in the Windows configuration? I'm not sure what to make of the failure. I'll await general feedback before I change anything.

danog:
Not merging this, already merged in 6.x.

PR #11172, opened 2024-11-27
Add phpstub for amqp extension (php-amqp)

mkilmanas:
Disclaimer: first time contributor here

This adds phpstub file for [AMQP extension](https://github.com/php-amqp/php-amqp) by copying contents of their [originally provided stubs](https://github.com/php-amqp/php-amqp/tree/v2.1.2/stubs) and concatenating into a single file.

What I am uncertain of and would like your feedback on:
- Should stubs in psalm contain original docblocks with explanations? Or only those with annotations that help with types, deprecations, etc?
- I have added a link to their LICENCE file (copying the whole file over seems a bit of an overkill). Is that sufficient?
- I have found two places to register/allow the new extension - have I missed any others? It did not seem like any tests would be specifically checking for each extension, so no new tests added either.

Looking forward to your comments (and hopefully a merge at the end)

### DIFF
--- a/config.xsd
+++ b/config.xsd
@@ -729,6 +729,7 @@
 
     <xs:simpleType name="ExtensionType">
         <xs:restriction base="xs:string">
+            <xs:enumeration value="amqp"/>
             <xs:enumeration value="apcu"/>
             <xs:enumeration value="decimal"/>
             <xs:enumeration value="dom"/>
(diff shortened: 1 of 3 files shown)


mkilmanas:
Hey, what's the status and overall position with this PR - is it at all worth pursuing further?

2 checks failed for debatable reasons:
a) The PR does not have any labels assigned, which I am not in position to add myself, so it's up for maintainers, I suppose
b) Backwards compatibility check is failing, but I find the actual reason to be extremely formal and cannot think of any real-life situation where that could actually break anything or cause any other kind of trouble.

Please let me know what you think and anything that I could improve here.

danog:
Thank you!

Issue #11107, opened 2024-09-30
PHP 8.4 support


kkmuffme:
- [ ] https://wiki.php.net/rfc/deprecations_php_8_4

- [ ] https://github.com/php/php-src/blob/php-8.4.0RC1/UPGRADING#L102 (e.g. `The type of PHP_DEBUG and PHP_ZTS constants changed to bool.`)

- [ ] update ProjectAnalyzer `private const PHP_SUPPORTED_VERSIONS_REGEX = '^(5\.[456]|7\.[01234]|8\.[0123])(\..*)?$';`

- [ ] update composer.json

- [ ] all changes done for 8.2/8.3 in PR https://github.com/vimeo/psalm/pull/10028 need to be done for 8.4 too

edsrzf:
#11166 and #11168 fix a couple warnings I got when running Psalm with PHP 8.4.

llaville:
Psalm error handler (https://github.com/vimeo/psalm/blob/5.26.1/src/Psalm/Internal/ErrorHandler.php#L62) do not allow to disable deprecations at runtime ( `php -d error_reporting=24575 vendor/bin/psalm` )

Bolkarr:
"This version of Nextcloud is not compatible with PHP>=8.4.
You are currently running 8.4.1."

Is there a distinction between 8.4 and 8.4.1 ?

juliangums:
Is there any ETA on 8.4 support?

alexander-schranz:
v6 with PHP 8.4 is released: https://github.com/vimeo/psalm/releases/tag/6.0.0. Can be closed /cc @danog

theodorejb:
Does Psalm 6 support static analysis of all the new PHP 8.4 features like property hooks and asymmetric visibility?

danog:
Psalm v6 now properly runs on PHP 8.4, and all signatures are up-to-date for PHP 8.4.

Property hook and asymmetric visibilty support is planned, but currently there is no funding for it.
I would gladly accept a [support contract](https://psalm.dev/support), funding the implementation of these two features.

PR #11190, opened 2024-12-23
Fix wrong PHP_INT_MAX/MIN and PHP_DEBUG PHP 8.4 type

### LINKED ISSUE #11189: PHP_INT_MAX/MIN types make no sense
https://psalm.dev/r/56c6ecbbab

kkmuffme:
* Fix https://github.com/vimeo/psalm/issues/11189
=> technically on 32-bit systems the values are different, however 32-bit systems don't have any practical relevance anymore (e.g. PhpStorm IDE is only available as 64-bit since years, therefore it's safe to assume 32-bit PHP can be ignored)
* Fix PHP_DEBUG constant for PHP 8.4 part of https://github.com/vimeo/psalm/issues/11107 => https://php.watch/versions/8.4/PHP_ZTS-PHP_DEBUG-const-type-change

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ConstFetchAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ConstFetchAnalyzer.php
@@ -25,6 +25,9 @@ use function explode;
 use function implode;
 use function strtolower;
 
+use const PHP_INT_MAX;
+use const PHP_INT_MIN;
+
 /**
  * @internal
  */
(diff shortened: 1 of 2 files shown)


src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/ConstFetchAnalyzer.php
```diff
@@ -164,13 +167,22 @@ public static function getGlobalConstType(
...
+                case 'PHP_DEBUG':
                 case 'PHP_ZTS':
-                    return Type::getInt();
+                    if ($codebase->analysis_php_version_id >= 8_04_00) {
+                        return Type::getBool();
+                    }
+                    return Type::getIntRange(0, 1);
 
                 case 'PHP_INT_MAX':
+                    return Type::getInt(false, PHP_INT_MAX);
```
danog, line 184: This and PHP_INT_MIN should not be a literal, instead it should be a generic positive integer, because i.e. running psalm on 32-bit and 64-bit systems will return different results.

danog: PHP_INT_MAX is already correct at least on master, PHP_INT_MIN needs Type::getIntRange(null, -1)

staabm: Raspberry PI is the most used 32 bit PHP AFAIK

Issue #9807, opened 2023-05-23
The issue message for the $<tmp coalesce var>101 is not understandable


jorgsowa:
https://psalm.dev/r/ec8201d718

The description of the issues is not understandable. The term `$<tmp coalesce var>101` doesn't explain what's wrong exactly. The phrase `$<tmp coalesce var>101` should be probably annotated differently.

```
ERROR: [RedundantCondition](https://psalm.dev/122) - 7:10 - Type array<array-key, mixed> for $<tmp coalesce var>101 is never null

ERROR: [TypeDoesNotContainNull](https://psalm.dev/090) - 7:33 - Cannot resolve types for $<tmp coalesce var>101 - array<array-key, mixed> does not contain null
```

orklah:
It comes from here: https://github.com/vimeo/psalm/blob/1c19260cddea29803cba2ef31c3a49189bcdeb0c/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/CoalesceAnalyzer.php#L46

When processing a coalesce operator, Psalm will create a fake expression that looks like this: `isset($<tmp coalesce var>XXX) ? $<tmp coalesce var>XXX : [];`

the `$<tmp coalesce var>XXX` must be absolutely unique in the scope because otherwise it may clash with an existing variable already existing. That's why the XXX is actually the position of the first char of the expression in the file, that way there won't be two clashing variables.

I agree the current syntax is confusing, feel free to make a PR and propose something better (keep in mind that it should be a unique name and must not be something that could already be a variable in the scope)

AkshatHotCode:
@orklah I've noticed an issue related to Psalm errors and code readability in our project. I'm planning to work on this by introducing a more descriptive variable name in a coalesce operation.

Would appreciate any thoughts or collaboration on this. Thanks!

orklah:
CoalesceAnalyzer is in need of a lot of love!

We have multiple bugs related to the logic in here and the fact that we transform the coalesce into a fake ternary is causing some of those.

But if you want to focus on the readability issue, my previous message summarized what is needed, a unique string that can't conflict with any variable in the scope nor any other similar ternary (for example, imagine your have `$a ?? []`; in two consecutive lines, the string you generate must be different between the two

We could maybe go with `Type array<array-key, mixed> for **$a at line 108 char 5486** is never null` that way it is both unique and descriptive

AkshatHotCode:
[...]

PR #11202, opened 2025-01-18
The issue for  message #9807

wetas2020:
This change makes the error messages more understandable. Instead of: Type array<array-key, mixed> for $<tmp coalesce var>101 is never null You'll now see something like:
Type array<array-key, mixed> for $coalesce_temp_line7_pos101 is never null


orklah:
Can you rebase on 6.x?

PR #11204, opened 2025-01-22
fix DOCUMENT_ROOT is always defined in $_SERVER as well as some others

kkmuffme:
I learned some additional things since I created that code in psalm initially and made it more correct now


orklah:
Can you target 6.x?

kkmuffme:
Done and rebased

PR #11214, opened 2025-01-28
Multithreaded refactoring

danog:
Improvements for multithreaded mode


danog:
Pushed directly to master.

PR #11225, opened 2025-01-30
Log files being processed for too long

danog:


orklah:
Hey! If you didn't know already, there is a similar feature for functions : https://github.com/vimeo/psalm/blob/487e9ce313f37121fb6835fbc710c4716e3afa7f/src/Psalm/IssueBuffer.php#L811 
It could be great to have a similar way to enable or log for both

danog:
Will consider that, for now the main usecase of this specific feature is to provide logs with really slow scans *at runtime*, i.e. I currently have a scan that takes around 2 days to complete (getting stuck on 98% after like 10 minutes :)

Issue #11227, opened 2025-01-30
Support uppercase-string


VincentLanglet:
In the same way psalm supports `lowercase-string` PHPStan does.
But PHPStan also supports `uppercase-string`. (https://github.com/phpstan/phpstan-src/pull/3613)
https://phpstan.org/r/1bb5f514-6781-48a8-a775-70c8fbe8bb3a

Psalm currently does not understand the type https://psalm.dev/r/f5835a8213

When using both PHPStan and Psalm it's kinda annoying since it forces to use double annotation.

It would be great to supports uppercase-string for psalm too, but I feel like it may be more complicated in PHPStan because
- PHPStan use an array of accessory (non-empty, non-falsy, lowercase, numeric, uppercase, literal, ...) so can easily represente any intersection of those types.
- Psalm use a class for every existing intersection. So psalm has TNonEmptyString, TNonFalsyString, TNumericString, TLowercaseString, TNonEmptyLowercaseString, but is already missing classes like TNonFalsyLowercaseString, TNonFalsyNumericString (to exclude `0`), TNumericLowercaseString (if you need to exclude `10E2`) and introducing TUppercaseString would require multiple new classes (knowning that even `TLowercaseUppercaseString` is technically possible for string without letter like numbers or special characters).

So at least would it be easily possible for Psalm to consider `uppercase-string` as an alias for `string`, in order to avoid using double annotation when using both PHPStan and psalm ?

orklah:
Yeah, we should at least have an alias. That would be pretty easy to do by adding to this switch: https://github.com/vimeo/psalm/blob/6.x/src/Psalm/Type/Atomic.php#L279

PR #10975, opened 2024-05-15
Add support for phpstan-pure

VincentLanglet:

PR #8654, opened 2022-11-03
Add support for phpstan assertions

VincentLanglet:

PR #5648, opened 2021-04-19
Add support for phpstan import-type and type annotations

VincentLanglet:
Trying to solve https://github.com/vimeo/psalm/issues/5647

PR #11228, opened 2025-01-31
Parse uppercase-string

VincentLanglet:
Hi @danog 

As a quick workaround for https://github.com/vimeo/psalm/issues/11227, this avoid to duplicate annotation when using the new syntax `uppercase-string` from PHPStan.

Suggested by @orklah https://github.com/vimeo/psalm/issues/11227#issuecomment-2626442084

### DIFF
--- a/src/Psalm/Type/Atomic.php
+++ b/src/Psalm/Type/Atomic.php
@@ -184,6 +184,7 @@ abstract class Atomic implements TypeNode, Stringable
                 return new TFloat();
 
             case 'string':
+            case 'uppercase-string': // TODO implement TUppercaseString
                 return new TString();
 
             case 'bool':
@@ -270,6 +271,7 @@ abstract class Atomic implements TypeNode, Stringable
                 return Type::getNonEmptyListAtomic(Type::getMixed(false, $from_docblock));
 
             case 'non-empty-string':
+            case 'non-empty-uppercase-string': // TODO implement TNonEmptyUppercaseString
                 return new TNonEmptyString();
 
             case 'truthy-string':


orklah:
Mind adding some tests? :)

danog:
Personally, I'd rather see a full implementation

orklah:
Well, yeah, me too, but should we wait and make people juggle between phpstan and psalm until someone create the full implementation?

VincentLanglet:
> Personally, I'd rather see a full implementation

Sure, but who'll do the full implementation ?
- There is 50+ occurences of `LowercaseString` in psalm codebase to copy with UppercaseString
- There is possible extra-work/reflection to handle string which is both lowercase and uppercase

Personally I almost know nothing about the Psalm codebase, and the lack of activity in 2024 pushed me to drop psalm in favor of using only PHPStan because psalm started to be to much behind. The last time I tried to take time for Psalm was unfortunately for nothing with https://github.com/vimeo/psalm/pull/11035. 


I'm really happy for Psalm that the tool got a new motivated maintainer. But I don't think implementing uppercaseString will be in your priority when we see all the existing issues.

Currently I'm trying to try again Psalm with the v6 and it's a little bit discouraging to get error because psalm and phpstan doesn't speak the same language. I already did in the past some (minor) work to improve this compatibility like
- https://github.com/vimeo/psalm/pull/5648
- https://github.com/vimeo/psalm/pull/8654
- https://github.com/vimeo/psalm/pull/8678
- https://github.com/vimeo/psalm/pull/10975

But a full implementation will be too much for me. I'm sorry. So until we'll see a full implementation, it would give a better dev experience to consider it as a string alias. PHPStan used the same strategy for `lowercase-string` when Psalm introduced it, and it took years before getting a full implementation of lowercaseString in PHPStan ; it was a pretty satisfying solution as a PHPStan user.

Feel free to close this if not wanted, and/or wait for someone planning to do the full implementation.

Otherwise I can add test if wanted.

VincentLanglet:
> Mind adding some tests? :)

Any recommendation in which file it should be @orklah ?

orklah:
> > Mind adding some tests? :)
> 
> Any recommendation in which file it should be @orklah ?

Nevermind, I thought we had a test that compared the phpdoc type with a class in Psalm but I must have imagined it :D

PR #11237, opened 2025-02-03
Add new psalm-review tool

danog:


BafS:
It seems to not be available when using psalm/phar, would it be possible to add a flag there like we have with `--alter`?

danog:
Sure!

PR #11243, opened 2025-02-04
Fix issue with insufficient shared memory for interned strings buffer

yethee:
This PR fixes an issue of inconsistency between options `opcache.interned_strings_buffer` and `opcache.memory_consumption`.

I encountered a problem when psalm crashed with an error:

```
Fatal Error Insufficient shared memory for interned strings buffer! (tried to allocate 67109072 bytes)
```

This error happened when option `opcache.memory_consumption` has value `64` or less. *In my case the value was `64`*.

### DIFF
--- a/src/Psalm/Internal/Fork/PsalmRestarter.php
+++ b/src/Psalm/Internal/Fork/PsalmRestarter.php
@@ -103,6 +103,12 @@ final class PsalmRestarter extends XdebugHandler
                     return true;
                 }
             }
+
+            $requiredMemoryConsumption = $this->getRequiredMemoryConsumption();
+
+            if ((int)ini_get('opcache.memory_consumption') < $requiredMemoryConsumption) {
+                return true;
+            }
         }
 
         return $default || $this->required;
@@ -169,6 +175,12 @@ final class PsalmRestarter extends XdebugHandler
             foreach (self::REQUIRED_OPCACHE_SETTINGS as $key => $value) {
                 $additional_options []= "-dopcache.{$key}={$value}";
             }
+
+            $requiredMemoryConsumption = $this->getRequiredMemoryConsumption();
+
+            if ((int)ini_get('opcache.memory_consumption') < $requiredMemoryConsumption) {
+                $additional_options []= "-dopcache.memory_consumption={$requiredMemoryConsumption}";
+            }
         }
 
         array_splice(
@@ -180,5 +192,24 @@ final class PsalmRestarter extends XdebugHandler
         assert(count($command) > 1);
 
         parent::restart($command);
+    }
+
+    /**
+     * @return positive-int
+     */
+    private function getRequiredMemoryConsumption(): int
+    {
+        // Reserve for byte-codes
+        $result = 256;
+
+        if (isset(self::REQUIRED_OPCACHE_SETTINGS['jit_buffer_size'])) {
+            $result += self::REQUIRED_OPCACHE_SETTINGS['jit_buffer_size'] / 1024 / 1024;
+        }
+
+        if (isset(self::REQUIRED_OPCACHE_SETTINGS['interned_strings_buffer'])) {
+            $result += self::REQUIRED_OPCACHE_SETTINGS['interned_strings_buffer'];
+        }
+
+        return $result;
     }
 }


danog:
I'd raise this a bit more to be frank, or even better add some logic to raise it if it's less than 64mb.

yethee:
`128` is the default value defined in php.
I wasn't looking for a threshold at which the problem disappears, but immediately tried the default value for `opcache.memory_consumption`.

I've tried other options now:

```
$ php -dopcache.memory_consumption=96 vendor/bin/psalm
Fatal Error Insufficient shared memory!

$ php -dopcache.memory_consumption=112 vendor/bin/psalm
Fatal Error Insufficient shared memory!

$ php -dopcache.memory_consumption=120 vendor/bin/psalm
Fatal Error Could not allocate JIT root traces buffer!
```

I think we can add a condition that the value of `memory_consumption` should be at least twice as large as `interned_strings_buffer`.

danog:
Well, it should be bigger than jit_buffer_size + interned_strings_buffer + some amount to account for the uncompiled bytecode of the classes

yethee:
> + some amount to account for the source bytecode

I think 256 MB should be enough for most projects.

src/Psalm/Internal/Fork/PsalmRestarter.php
```diff
@@ -181,4 +193,23 @@ protected function restart($command): void
...
+
+        if (isset(self::REQUIRED_OPCACHE_SETTINGS['jit_buffer_size'])) {
+            $result += self::REQUIRED_OPCACHE_SETTINGS['jit_buffer_size'] / 1024 / 1024;
+        }
+
+        if (isset(self::REQUIRED_OPCACHE_SETTINGS['interned_strings_buffer'])) {
+            $result += self::REQUIRED_OPCACHE_SETTINGS['interned_strings_buffer'];
+        }
+
+        return $result;
```
danog, line 213: return this in bytes

src/Psalm/Internal/Fork/PsalmRestarter.php
```diff
@@ -181,4 +193,23 @@ protected function restart($command): void
...
+    /**
+     * @return positive-int
+     */
+    private function getRequiredMemoryConsumption(): int
+    {
+        // Reserve for byte-codes
+        $result = 256;
+
+        if (isset(self::REQUIRED_OPCACHE_SETTINGS['jit_buffer_size'])) {
+            $result += self::REQUIRED_OPCACHE_SETTINGS['jit_buffer_size'] / 1024 / 1024;
```
danog, line 206: bytes not megabytes

yethee: We need convert to MB, because value for `memory_consumption` should be set in megabytes and `jit_buffer_size` is defined in bytes.

> [`opcache.memory_consumption int`](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.memory-consumption)
> The amount of memory used to store interned strings, in megabytes. The maximum value is 32767 on 64bit architectures, and 4095 on 32bit architectures.

> [`opcache.jit_buffer_size int`](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.jit-buffer-size)
> The amount of shared memory to reserve for compiled JIT code. A zero value disables the JIT.

>
> When an [int](https://www.php.net/manual/en/language.types.integer.php) is used, the value is measured in bytes. Shorthand notation, as described in [this FAQ](https://www.php.net/manual/en/faq.using.php#faq.using.shorthandbytes), may also be used.

src/Psalm/Internal/Fork/PsalmRestarter.php
yethee: According to the documentation, the value for `memory_consumption` is set in megabytes.

src/Psalm/Internal/Fork/PsalmRestarter.php
```diff
@@ -169,6 +175,12 @@ protected function restart($command): void
             foreach (self::REQUIRED_OPCACHE_SETTINGS as $key => $value) {
                 $additional_options []= "-dopcache.{$key}={$value}";
             }
+
+            $requiredMemoryConsumption = $this->getRequiredMemoryConsumption();
+
+            if ((int)ini_get('opcache.memory_consumption') < $requiredMemoryConsumption) {
```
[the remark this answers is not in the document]
yethee: Shorthand notation is not supported for this option.

danog:
Thanks!

PR #11255, opened 2025-02-06
Fix config error in LSP

tm1000:
This fixes an error in the language server when trying to recall the config from the client

Also hello @danog! I mantain https://github.com/psalm/psalm-vscode-plugin and occasionally work on the LSP here


tm1000:
@danog done!

PR #11257, opened 2025-02-06
Improve performance and accuracy of types comparison

danog:


src/Psalm/Internal/Analyzer/Statements/Expression/SimpleTypeInferer.php
```diff
@@ -92,11 +92,11 @@ public static function infer(
                         return Type::getString($result);
                     }
 
-                    if ($left->isSingle() && $left->getSingleAtomic() instanceof TNonEmptyString) {
+                    if ($left->isSingle() && $left->isNonEmptyString()) {
```
orklah, line 95: Hey @danog! The isSingle check feels redundant. I think it was there for simplicity in order to be able to call getSingleAtomic just after but isNonEmptyString should be able to look through every Atomic

danog: Right, hang on let me push a fix

PR #11258, opened 2025-02-06
Fix TemplateInferredTypeReplacer

VincentLanglet:
This reverts commit 7b4ab31.


VincentLanglet:
You can see that the test failing is wrong just by looking at
```
$container = new SpecificObjectContainer(new stdClass());
$object = $container->build(SpecificObject::class);
$nonSpecificObject = $container->build("whatever");
```

Since the implementation is
```
public function __construct(private readonly mixed $instance){}

public function build(string $name): mixed
{
     return $this->instance;
}
```

build is just returning the instance passed to the constructor.

So both `$object` and `$nonSpecificObject` are supposed to be `stdClass`,
but the test expect a wrong assertion
```
'$object===' => 'SpecificObject',
```

danog:
actually you know what, that entire test is clearly too messy, let's get rid of it

VincentLanglet:
> actually you know what, that entire test is clearly too messy, let's get rid of it

Done

VincentLanglet:
Thanks to you I was able to get https://github.com/vimeo/psalm/pull/11252 working :)

tests/Template/ConditionalReturnTypeTest.php
```diff
@@ -1043,7 +977,7 @@ function getSomething(string $string)
                 'code' => '<?php
                     /**
                      * @param string|int $stringOrInt
-                     * @psalm-return ($stringOrInt is (non-empty-string|int) ? string : int)
+                     * @psalm-return ($stringOrInt is (non-empty-string)|int ? string : int)
```
SCIF, line 980: Why is that? Are you it's not a bug? This change looks very suspicious imho

PR #11269, opened 2025-02-07
Fix type assertion for Arithemtic addition and subtraction

### LINKED ISSUE #11241: Type inference issue when using arithmetic assignments
When upgrading to Psalm 6 we've noticed some new errors in the codebase.

Which seem to be caused by psalm incorrectly tracking the type of a variable, if it is created as an integer, and then updated through arithmetic addition or subtraction assignments.

See
https://psalm.dev/r/6385ed8d8d 
and
https://psalm.dev/r/c9b1c78210

This problem only seems to happen during addition and subtraction though. The other types of arithmetic assignments change the type of the variable back to either `int` or `float|int`.

KevinVanSonsbeek:
Closes #11241

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
@@ -831,7 +831,7 @@ final class ArithmeticOpAnalyzer
                                 $new_type = new TIntRange(null, $left_type_part->value + $sum);
                             }
                         } elseif ($left_type_part instanceof TLiteralInt) {
-                            if ($context->inside_assignment) {
+                            if ($context && $context->inside_assignment) {
                                 $new_type = new TInt();
                             } else {
                                 $new_type = new TLiteralInt($left_type_part->value + $sum);
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/ArithmeticOpAnalyzer.php
@@ -823,6 +823,8 @@ final class ArithmeticOpAnalyzer
                             }
                         }
                     } elseif ($parent instanceof VirtualPlus || $parent instanceof VirtualMinus) {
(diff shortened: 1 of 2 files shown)


orklah:
Cool! Seems like an improvement but I'm not sure it fixes everything

Can you check this for example? https://psalm.dev/r/6f97b61d3d

What you did is replace `$sum` usage by a generic `TInt` in one of the branches. It's better because it stops assuming we have a literal 1, but the issue is with $sum everywhere I think

We should fix it here if we can
https://github.com/vimeo/psalm/pull/11269/files#diff-38fd2d45971b85013dfc252b81a4d63816a2f0b9e6d15b5168c5878ad2dd65c4R826

I think we need to check what's in `$right_type_part` instead of assuming we have 1 or -1.
If it's a literal int, the operation is easy to do. If we have something else, we can just assume a generic TInt will be returned for the whole https://github.com/vimeo/psalm/pull/11269/files#diff-38fd2d45971b85013dfc252b81a4d63816a2f0b9e6d15b5168c5878ad2dd65c4R825 branch

KevinVanSonsbeek:
I can't really seem to reproduce the sum issues in other situations.

The example you sent with an int-range is handled in an earlier branch. Same with additions/subtractions that have an TLiteralInt on the right side, and are not inside of a loop.

If i check the right type part, and grab that value if it is a `TLiteralInt` it actually ends up breaking int-ranges for loops.
ex:
```php
// Replacing the sum assignment with this:
  if ($right_type_part instanceof TLiteralInt) {
      $sum = $right_type_part->value;
  } else {
      $sum = $parent instanceof VirtualPlus ? 1 : -1;
  }
```

Causess the `incrementInLoop` test to fail:
```
-    '$i' => 'int<0, 10>'
-    '$j' => 'int<100, 110>'
+    '$i' => 'int<0, 11>'
+    '$j' => 'int<100, 111>'
```
Increments in loops also seem to be the only situation where the righthand side is a TLiteralInt from what i can see in the tests present.

orklah:
Fair enough! Can you just leave a comment above the $sum initialisation with something like "This seems arbitrary defined as 1/-1, this may need a better handling if there's issues"

Just to facilitate a future debugging :p

Thanks!

orklah:
Thanks!

PR #11270, opened 2025-02-07
Force usage of latest patch version to avoid JIT bugs

danog:
Fixes https://github.com/vimeo/psalm/issues/11268

### DIFF
--- a/src/Psalm/Internal/CliUtils.php
+++ b/src/Psalm/Internal/CliUtils.php
@@ -543,7 +543,12 @@ final class CliUtils
         $issues = [];
 
         $major_minor = PHP_VERSION_ID - (PHP_VERSION_ID % 100);
-        foreach ([8_01_31 => '8.1.31', 8_02_27 => '8.2.27', 8_03_16 => '8.3.16', 8_04_03 => '8.4.3'] as $version => $txt) {
+        foreach ([
+            8_01_31 => '8.1.31',
+            8_02_27 => '8.2.27',
+            8_03_16 => '8.3.16',
+            8_04_03 => '8.4.3',
+        ] as $version => $txt) {
             $version_m = $version - ($version % 100);
             if ($version_m === $major_minor) {
                 if (PHP_VERSION_ID < $version) {
--- a/src/Psalm/Internal/CliUtils.php
+++ b/src/Psalm/Internal/CliUtils.php
@@ -524,9 +524,6 @@ final class CliUtils
 
     public static function checkRuntimeRequirements(): void
     {
-        $required_php_version = 8_01_17;
-        $required_php_version_text = '8.1.17';
-
         // the following list was taken from vendor/composer/platform_check.php
         // It includes both Psalm's requirements (from composer.json) and the
         // requirements of our dependencies `netresearch/jsonmapper` and
@@ -545,9 +542,16 @@ final class CliUtils
         ];
         $issues = [];
 
-        if (PHP_VERSION_ID < $required_php_version) {
-            $issues[] = 'Psalm requires a PHP version ">= ' . $required_php_version_text . '".'
+        $major_minor = PHP_VERSION_ID - (PHP_VERSION_ID % 100);
+        foreach ([8_01_31 => '8.1.31', 8_02_27 => '8.2.27', 8_03_16 => '8.3.16', 8_04_03 => '8.4.3'] as $version => $txt) {
+            $version_m = $version - ($version % 100);
+            if ($version_m === $major_minor) {
+                if (PHP_VERSION_ID < $version) {
+                    $issues[] = 'Psalm requires a PHP version ">= ' . $txt . '".'
                         . ' You are running ' . PHP_VERSION . '.';
+                }
+                break;
+            }
         }
 
         $missing_extensions = array_filter(
--- a/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php
@@ -77,7 +77,6 @@ use function mkdir;
 use function number_format;
 use function preg_match;
 use function rename;
-use function sprintf;
 use function str_ends_with;
 use function str_starts_with;
 use function strlen;
@@ -87,6 +86,7 @@ use function substr;
 use function usort;
 
 use const PHP_EOL;
+use const PHP_VERSION;
 use const PSALM_VERSION;
 use const STDERR;
 
@@ -280,7 +280,7 @@ final class ProjectAnalyzer
             clearstatcache(true, $cache_directory);
             if (is_dir($cache_directory)) {
                 $this->progress->debug(
-                    'Composer lockfile change detected, clearing cache directory ' . $cache_directory . "\n",
+                    'Composer lockfile change detected, clearing cache directory ' . $cache_directory . PHP_EOL,
                 );
 
                 Config::removeCacheDirectory($cache_directory);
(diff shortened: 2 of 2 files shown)


MoonE:
Isn't that a bit harsh? 

If JIT is the problem, then it could be  check for that in combination with the version.
I'm currently on Ubuntu 24.10 and the latest available version there is php 8.3.11.

Also if psalm is hard-coded to require a specific php version it should say so in the composer.json. Otherwise composer will happily download a version that can't be used.

danog:
That is kind of the point, hardcoding a composer.json version will prevent users from getting newer versions of psalm, but a runtime check will load it just fine.

The ondrej ppa repos for Ubuntu have the latest 8.3.16 with all the fixes; I do not feel comfortable with slowing down Psalm due to some distros lagging behind in patch releases.

DanielBadura:
But letting people install psalm despite knowing it will not run is just bad dx. I just ran into it and was confused why i even could install the version if it's not runnning with it...

IMHO bumping the min version in composer.json is here the way to enforce newer php versions.

theodorejb:
> hardcoding a composer.json version will prevent users from getting newer versions of psalm

It shouldn't be an exact version, just a minimum version. For example: 

```json
"php": "~8.1.31 || ~8.2.27 || ~8.3.16 || ~8.4.3"
```

danog:
Either way, specifying a more strict min version will silently prevent composer update from loading the latest version.

For a development tool like Psalm, I consider a runtime check to be more than acceptable.

src/Psalm/Internal/CliUtils.php
```diff
@@ -545,9 +542,21 @@ public static function checkRuntimeRequirements(): void
...
+            8_04_03 => '8.4.3',
+        ] as $version => $txt) {
+            $version_m = $version - ($version % 100);
+            if ($version_m === $major_minor) {
+                if (PHP_VERSION_ID < $version) {
+                    $issues[] = 'Psalm requires a PHP version ">= ' . $txt . '".'
                         . ' You are running ' . PHP_VERSION . '.';
+                }
+                break;
+            }
```
Ocramius, line 559: Please just declare `composer.json` `"conflict"` ranges next time: this has caused me quite a few headaches, yesterday.

I'd rather get an older psalm version, than one that crashes at startup.

danog: Thing is, the version bump was made for a specific reason, to avoid crashes and other subtle bugs that would have caused more lost time in pointless debugging than a simple PHP version upgrade...

Ocramius: From a consumer perspective, a conflict range would've made this clear, and could've been relaxed until a bugfix is applied (if that would ever allow relaxing the dependency range).

A runtime crash is the worst of both worlds:

* it doesn't work
* I don't know it until I run it

Early signaling here is `composer update` failing to get the latest release.

The current effective result I have right now is me  + other people I collaborate with having to manually go in `composer.json` and pinning `vimeo/psalm` to an older release.

In fact, I currently use `vimeo/psalm:6.4.0` (rollback) on all my systems, and added `vimeo/psalm:6.5.0` as an explicit conflict for my colleagues.

danog: The problem with conflict ranges is that composer does not make them at all clear to users of a package; the dependency resolver simply works around them to fit the current version constraints, regardless of the side effects (bugs in older PHP versions).

This is not the only runtime check, and some requirements (like the VM overcommit requirements) cannot even be expressed using composer.

I will not be adding a dependency constraint to composer.json, because as I said, Psalm is a development tool, and the development environment is not as rigid as a production environment (to a certain extent, i.e. minor bumps will still be expressed using composer, as they are somewhat related to the production environment).

Ocramius: > the dependency resolver simply works around them to fit the current version constraints, regardless of the side effects (bugs in older PHP versions).

We've been working this way for well over a decade now: not getting "latest" is fine, and `composer why-not` tells you so.

I've even built an entire package around this: https://github.com/Roave/SecurityAdvisories/ - it's designed to exclude dependencies that are a hazard, and people do use it, consciously.

Runtime checks are 100% not the way, and amplify noise around this.

I (relatively experienced PHP dev) for sure wasted a lot of time yesterday dealing with this while trying to work on https://github.com/psalm/psalm-plugin-phpunit/pull/149 - had to roll back `vimeo/psalm` tons of times while trying out dependency ranges.

What moving this to a runtime check achieved here is:

1. confusion (people are putting feedback here for a reason: nobody wants to piss you off, rest assured of that :-P )
2. almost all `composer.lock` upgrade pipelines ( @renovate / @dependabot ) will fail due to an invalid upgrade
3. time wasted during development

## What should be done next?

The correct way forward here is:

1. revert patch
2. tag new patch release (so that upgrades allow jumping to a working release again, including automated upgrades performed so far)
3. re-apply patch
4. add dependency range in `"conflict"`
5. tag new patch release

Note that deleting a release is non-viable, unless it's a security risk.

As for the "it's a development tool": it does not matter if it is - it's part of the dependency graph, and the dependency graph is designed to say "these work together" or "these don't work together".

In addition to that, even if the dependency graph is split between `--dev` and `--no-dev` dependencies, the "these work together" should still apply for the `--dev` part.

danog: Marco, I value your input, and look forward to collaborating with you (will be taking a look at your PRs next week).

I am not *too* willing to die on this hill: in general, it's not about this specific change, it's about the possiblity for me as the maintainer of Psalm to direct users to make changes to their environment, to avoid them footguns when working with Psalm.

In my long time as a PHP engineer, I have always tried to proactively prevent bugs from even reaching users, guaranteeing the best possible experience when actually using my tools, be it using warnings to update to the last version on startup (like in MadelineProto, will probably implement this in Psalm as well, given that phpstan has them too), or using runtime exceptions for particularly bad platform bugs (i.e. did something similar for MadelineProto for a recent fiber-related bug, and @trowski did the same inside of amphp itself as well), or even automatic static analysis on code written by the users, to proactively detect possible issues.

I do not want users to waste time debugging weird exceptions and crashes caused by PHP bugs (like recently, I already got two new issues caused by a PHP bug on older versions of PHP, combined with the new amphp-based concurrency model). 
 
A (patch release, not even a major/minor!) PHP version upgrade is a minor inconvenience, bypassable using any modern PHP distribution method (ondrej ppas on debian/ubuntu), and does not require any additional refactoring work on the codebase on behalf of users.

This is also the case for other runtime checks as well (such as https://github.com/vimeo/psalm/pull/11218).

I am a user and great fan of your security-advisories project, because it allows library developers to bypass composer's long-time limitations, to prevent users from getting vulnerable/broken versions of libraries.

I am trying to apply the same rationale behind security-advisories to Psalm as well, using the only way I have available to me, given that we're talking about PHP platform bugs, and not issues with userland libraries.

Ocramius: I feel like the message didn't come through, but I'm also just making a point, and not really willing to explain further than I already did :D 

We should probably talk about it over :beers: at PHPDay or such :)

danog: Ayy, now that's a nice idea!

Kravets1996:
I'm absolutely okay with updating PHP on my local machine, but there is a problem when you have complex infrastructure with different environments. For example, our company have about 50 PHP projects on different platforms (frameworks, CMS etc.). 

All of this projects have different environments: shared hosting, VPN, dedicated servers, clusters of servers, docker/kubernetes deployments. All of this projects have some kind of CI/CD pipeline which use a plenty of base Docker images for CI/CD and test/production.

For us it is not possible to always have latest PHP patch, partially because some of the base Docker images for projects not ours, but rather community driven, open-source images. And it's totally okay to be a little outdated for this images, because they are used in CI/CD.

So for us it became impossible to use latest psalm on most of our project, where it was successfuly used for years. 

I totally agree with all comments above and strongly recommend take a look at this decision again.

danog:
This was created for a client that specifically had issues with Psalm on an old PHP version in CI due to JIT bugs.

I am preparing an official Psalm docker image that can and should be used to avoid environment issues, in the meantime just using the official php:8.4 image should solve all issues (+, psalm already bundles stubs for most extensions, they just need to be enabled in the config).

PR #11273, opened 2025-02-09
Bump `symfony/filesystem`

weirdan:
The versions listed are the earliest in their respective branches that
do not crash Psalm on PHP 8.4.

The crash was due to implicit nullable parameter in
`Symfony\Component\Filesystem\Path::getFilenameWithoutExtension()`


weirdan:
The crash can be seen here:
https://github.com/psalm/codeception-psalm-module/actions/runs/13230311585/job/36926488221?pr=57#step:10:68

PR #11277, opened 2025-02-12
Fix key type in PossiblyUndefinedStringArrayOffset.md

emmanuelGuiton:
Fixed array key type mistake : string instead of int.


orklah:
Thanks!

PR #11283, opened 2025-02-14
Enable more features by default

danog:


Ocramius:
As mentioned in #11288, I believe this should target `7.x`.

Discussion over there: this comment is mostly cross-linking.

danog:
For future reference: https://github.com/vimeo/psalm/issues/11288#issuecomment-2661377990

M393:
@jorgsowa See https://github.com/vimeo/psalm/issues/11301#issuecomment-2662892263
In lower versions it will be seen as a normal comment by php.

danog:
& on php 8.0-8.2, simply require the symfony polyfill to polyfill the Override class

Issue #11184, opened 2024-12-13
urlencode(non-empty-string) returns non-empty-string


mathroc:
Hi! 

is it ok to update the callmap to make Psalm understand that `urlencode(non-empty-string)` will always return a `non-empty-string` ?

I guess there are many over functions like that and I'm not planning to do all of them though :sweat_smile:

mathroc:
https://psalm.dev/r/b603b51a2c

orklah:
Yeah, please do :)

PR #11292, opened 2025-02-16
[GH-11184] urlencode return an empty string when input is empty

mathroc:


dictionaries/override/CallMap.php
```diff
@@ -50216,13 +50216,23 @@
   ),
   'rawurldecode' => 
   array (
-    0 => 'string',
-    'string' => 'string',
+    0 => '\'\'',
```
orklah, line 50219: I'm not used to the new callmap format yet but the way I understand it, this is supposed to be the return type. Shouldn't that be string?

mathroc: I'm not sure of the format either, what I'm trying to do is tell that if the input is an empty string (`''`) the return type is an empty string too, and if the input is a non empty string, so is the output

I need to test it

orklah: I guess that could work, even if we rarely go this precise in callmaps, however in this case you'll be missing the main case (you handle empty-string, non-empty-string but you should also handle any-string :p

mathroc: right, a string can only be empty or non-empty, but that doesn't mean that we know that at the type level ! 🤦
I added back the any string variants. still need to test it. I wonder which variants Psalm will take if multiple could match. If they are tried in order I guess I'll have to put the any-string variant last. time to write some tests 👌

PR #11293, opened 2025-02-16
Fix condition on lowercase-string

VincentLanglet:
Similar to https://github.com/vimeo/psalm/pull/11252
cc @orklah


orklah:
Thanks!

PR #11307, opened 2025-02-17
Add docker image

danog:


bin/docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN apt-get update && apt-get -y --no-install-recommends install git unzip && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+ADD bin/docker/php.ini /usr/local/lib/php.ini
+
+COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
+
+ARG PSALM_REV=dev-master
+RUN COMPOSER_ALLOW_SUPERUSER=1 \
+    COMPOSER_HOME="/composer" \
+    composer global require vimeo/psalm:${PSALM_REV} --prefer-dist --no-progress --dev && \
```
weirdan, line 28: Won't this suffer from a race condition?

E.g. when you're building a container for a new, freshly tagged release you'd want packagist.org to pick it up first, otherwise this would fail, wouldn't it?

danog: That is indeed correct, and I thought of this already, will add a waiting loop later

Issue #11322, opened 2025-02-20
`Uncaught Error: Call to undefined function _HumbugBox44cdd6d36ff3\array_any() ` for 6.8.1 release


chenrui333:
👋 while testing psalm.phar 6.8.1 release, seeing some crash issue as below:

```
==> /home/linuxbrew/.linuxbrew/Cellar/psalm/6.8.1/bin/psalm --init
  Uncaught Error: Call to undefined function _HumbugBox44cdd6d36ff3\array_any() in phar:///home/linuxbrew/.linuxbrew/Cellar/psalm/6.8.1/libexec/psalm/src/Psalm/Internal/Fork/PsalmRestarter.php:52
  Stack trace:
  #0 phar:///home/linuxbrew/.linuxbrew/Cellar/psalm/6.8.1/libexec/psalm/vendor/composer/xdebug-handler/src/XdebugHandler.php(116): Psalm\Internal\Fork\PsalmRestarter->requiresRestart(false)
  #1 phar:///home/linuxbrew/.linuxbrew/Cellar/psalm/6.8.1/libexec/psalm/src/Psalm/Internal/Cli/Psalm.php(645): _HumbugBox44cdd6d36ff3\Composer\XdebugHandler\XdebugHandler->check()
  #2 phar:///home/linuxbrew/.linuxbrew/Cellar/psalm/6.8.1/libexec/psalm/src/Psalm/Internal/Cli/Psalm.php(234): Psalm\Internal\Cli\Psalm::restart(Array, 1, 1, Object(Psalm\Progress\LongProgress))
  #3 phar:///home/linuxbrew/.linuxbrew/Cellar/psalm/6.8.1/libexec/psalm/psalm(7): Psalm\Internal\Cli\Psalm::run(Array)
  #4 /home/linuxbrew/.linuxbrew/Cellar/psalm/6.8.1/libexec/psalm(14): require('phar:///home/li...')
  #5 /home/linuxbrew/.linuxbrew/Cellar/psalm/6.8.1/bin/psalm(2): require('/home/linuxbrew...')
  #6 {main}
  (Psalm 6.8.1@f12f96cde537a424fe2a9366105b5e[71](https://github.com/Homebrew/homebrew-core/actions/runs/13414729294/job/37472910776?pr=208273#step:4:72)46d4fc23 crashed due to an uncaught Throwable)
  Calculating best config level based on project files
  Error: psalm: failed
  An exception occurred within a child process:
[...]

PR #11323, opened 2025-02-20
Fix phar generation

danog:


danog:
Fixes https://github.com/vimeo/psalm/issues/11322

PR #11326, opened 2025-02-20
Regenerate callmap, add \Roundable parameter possibility to round PHP8.4

goossda:
This PR is for adding Roundable to the round- function as a possible parameter.

I also had to change the gen_callmap.sh to generate a new callmap.
After generation, more files did change, the changes to the round-function did pick up.

### DIFF
--- a/dictionaries/override/CallMap_84_delta.php
+++ b/dictionaries/override/CallMap_84_delta.php
@@ -1,87 +1,87 @@
 <?php // phpcs:ignoreFile
 
 return array (
-  'added' => 
+  'added' =>
   array (
-    'array_find_key' => 
+    'array_find_key' =>
     array (
       0 => 'array-key|null',
       'array' => 'array<array-key, mixed>',
       'callback' => 'callable',
     ),
-    'sodium_crypto_aead_aes256gcm_keygen' => 
+    'sodium_crypto_aead_aes256gcm_keygen' =>
     array (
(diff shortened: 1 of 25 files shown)


dictionaries/override/CallMap_84_delta.php
```diff
@@ -991,201 +1008,201 @@
...
       array (
         0 => 'true',
       ),
     ),
   ),
-  'removed' => 
+  'removed' =>
   array (
   ),
-);
```
danog, line 1191: You should have added the new signature in here as well as in CallMap.php.

goossda: I don't really understand, because the override delta has this signature?

```
    'round' =>
    array (
      'old' =>
      array (
        0 => 'float',
        'num' => 'float|int',
        'precision=' => 'int',
        'mode=' => 'int<0, max>',
      ),
      'new' =>
      array (
        0 => 'float',
        'num' => 'float|int',
        'precision=' => 'int',
        'mode=' => 'RoundingMode|int<0, max>',
      ),
    ),
```

danog: Yep, this seems correct

goossda: I don't think i need to commit something else right? :) 
It should be all there, however Github doesn't really allow us to check it online.

PR #11328, opened 2025-02-21
Fixes PHP Error: Undefined array key 1

tm1000:
This fixes a fatal crash with improperly defined template_covariants


danog:
Thanks!

PR #11364, opened 2025-03-20
Allow concatenation between (string|int) and similar types

### LINKED ISSUE #11337: Allow int|string and string concatenation
#11329 allowed concatenations between string and int variables in addition to concatenations between strings, but the type `int|string` should probably also be an acceptable type in such concatenations.

Example with false positive warning: https://psalm.dev/r/1ff6a14ddb

haas-dtv:
Fixes #11337


danog:
Looks okay, can you also add a few unit tests

haas-dtv:
I've added a few tests

danog:
Btw, could you also submit this against 6.x

Issue #11368, opened 2025-03-20
False positive on a method call inside a wrapper class when the called method's param is template T of static


Guuzen:
We have an abstract Uuid class in our project, which we use to differentiate between different types of Ids when we compare them, so it has an equals method. 

When we compare Ids inside a wrapper class using the method, psalm complains about a wrong type of the param of the equals method of the Uuid class and somehow infers, that the type should have a connection to the wrapper class. 

Expected Behavior:
no errors

Actual Behavior:
`Argument 1 of UserId::equals expects User&static, but parent type UserId provided`

The problem appears starting with psalm 6.5 so I guess this PR might be related https://github.com/vimeo/psalm/pull/11267

https://psalm.dev/r/6f7419118e

PR #11370, opened 2025-03-20
Fix binding of T of static when called from a class scope

robchett:
Partial fix for #11368

Issue appears to cement itself here - 
https://github.com/vimeo/psalm/blob/e052de4275fb6cdbfedfef1d883a7e90732ea609/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php#L418-L421
as `context->self = "User"` but the method was called on an instance of `UserId`

After updating to 

```php
 !$statements_analyzer->isStatic()
    && (!$method_id || $method_id->method_name !== '__construct')
    ? ($method_id ? $method_id->fq_class_name : $context->self)
    : null,
```

https://psalm.dev/r/a32e05de27
> ERROR: [ArgumentTypeCoercion](https://psalm.dev/193) - 19:33 - Argument 1 of UserId1::equals expects User&static, but parent type UserId1 provided

becomes 

> ERROR: [ArgumentTypeCoercion](https://psalm.dev/193) - 19:33 - Argument 1 of UserId1::equals expects UserId1&static, but parent type UserId1 provided

Now, using self instead of static works
```
    /**
     * @template T of self
     * @param T $uuid
     */
    final public function equals($uuid): void {}
```

I've seen the "A&static, but parent type A" style messaging elsewhere and unsure if this is now a legitimate issue or not.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentAnalyzer.php
@@ -417,7 +417,7 @@ final class ArgumentAnalyzer
                 $argument_offset,
                 !$statements_analyzer->isStatic()
                     && (!$method_id || $method_id->method_name !== '__construct')
-                    ? $context->self
+                    ? ($method_id ? $method_id->fq_class_name : $context->self)
                     : null,
                 $context->calling_method_id ?: $context->calling_function_id,
             );
--- a/tests/Template/FunctionTemplateTest.php
+++ b/tests/Template/FunctionTemplateTest.php
@@ -1850,12 +1850,35 @@ final class FunctionTemplateTest extends TestCase
                     );
 
                     $test = $list->sortByArr(
-                        /** 
+                        /**
                          * @return list{0: mixed, 1: mixed}
                          */
                         static fn ($item): array => [$item, $item],
                         1,
                     );',
+            ],
+            'staticScope#11368' => [
+                'code' => '<?php
+                    abstract class Uuid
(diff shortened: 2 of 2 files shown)


robchett:
Ignore the above snippet - I pasted the wrong link. https://psalm.dev/r/a32e05de27

Guuzen:
> I've seen the "A&static, but parent type A" style messaging elsewhere and unsure if this is now a legitimate issue or not.

Could you please elaborate?

robchett:
Not sure how I managed to add the same link twice https://psalm.dev/r/1028cf9de2

The major issue is resolved by this PR - I'm unure if this secondary warning is a bug or expected as I've seen it in other places but have always resolved by changing the type definitions (I can't find any other examples to hand).

With this PR the error is:

> ERROR: [ArgumentTypeCoercion] Argument 1 of UserId2::equals expects UserId2&static, but parent type UserId2 provided

In my mind `UserId2&static == UserId2`, but I may be mistaken

Guuzen:
Well, I am not sure either. But to me it feels that you are right, they are equal.

I can also add, that it feels strange to use self instead of static, because in PHP self is about the class where code is defined and static is about the class of the object from which code is called. Self would mean that it is possible to pass any subtypes of Uuid which is undesirable.

Actually, I am not sure that our use case is valid (although it had been working till the recent change).
We can relatively easily change it to class level generic instead of method level generic and it will work just fine (except that we need to specify the type for each class). It is also possible to specify the bound for only Uuid types, so it feels like there is no problem.
Something like this https://psalm.dev/r/54e8166be5

PR #11390, opened 2025-04-10
fix: alter of ClassMustBeFinal inserts final before attributes

### LINKED ISSUE #11289: `ClassMustBeFinal` alter leading to syntax error with attributes
The new `vendor/bin/psalm --alter --issues=ClassMustBeFinal` (running v6.6) does not take into account attributes and introduces syntax errors when run.

The following:

```php
#[Entity('addresses')]
readonly class Address
{
}
```

Gets "fixed" into:

```php
final #[Entity('addresses')]
readonly class Address
{
}
```

Which then is a syntax error in PHP (and in Psalm too). As class attributes are often used in PHP, this is a huge pain, especially as Psalm now advocates for running this alteration.

Tofandel:
Fixes #11289


danog:
Thank you for the PR, could you please add a testcase?

PR #11416, opened 2025-05-02
Disable diff mode

danog:
Forcefully disables diff mode to avoid a whole set of issues.
Proper fixes will be coming up, in the meantime this means caching can be safely re-enabled.


dan-rf:
Was there ever any followup to this? This PR makes the cache effectively useless as all files in the project will always be analysed regardless of changes.

PR #11443, opened 2025-05-19
fix(analyzer): fix crash when analzing first class callable

azjezz:
This PR fixes a runtime crash with first-class callables and significantly refactors conditional logic analysis for clarity and maintainability.

Example:

```php
<?php

while (count(...) > 1) { /* panics */ }
```

---

Key Changes:

- Fixes Crash: Resolves an assertion exception when `->getArgs()` is called on a `FuncCall` node representing a first-class callable. The new `isFunctionCallToOneOf()` helper now safely handles this.
-  Reduces Duplication & Improves Clarity:
    - Consolidated duplicated logic in various conditional checks (e.g., numeric comparisons like hasInferior/SuperiorNumberCheck are now handled by getComparisonLiteralOperand() and getLiteralIntegerFromExpression()).
    - Similar refactoring applied to gettype, get_debug_type, get_class, and count/sizeof checks.
    - Introduced multiple helper methods to centralize checks and make intent clearer.
- Optimizes: Implements short-circuiting for some boolean expressions.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/AssertionFinder.php
@@ -1404,19 +1404,13 @@ final class AssertionFinder
     private static function hasGetTypeCheck(
         PhpParser\Node\Expr\BinaryOp $conditional,
     ): bool|int {
-        if ($conditional->right instanceof PhpParser\Node\Expr\FuncCall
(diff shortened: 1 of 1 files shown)


azjezz:
@danog please let me know if i should split this PR into smaller ones. I initially was just reading through `AssertionFinder` and came across the bug mentioned above, and got carried away with refactoring and optimizing things :)

Issue #11442, opened 2025-05-18
Uncaught TypeError: str_starts_with(): Argument #1 ($haystack) must be of type string, int given in  vendor/vimeo/psalm/src/Psalm/DocComment.php:57


vrkansagara:
> ./vendor/bin/psalm -- --no-cache --find-dead-code --find-unused-psalm-suppress --long-progress

Running on PHP 8.4.7, Psalm dev-master@8506dfcb5657c5d649d6a4b165cc0ba8cde8a597.

JIT acceleration: ON

JIT compilation in progress...

JIT compilation took 5 seconds.                                                

Target PHP version: 8.4 (inferred from current PHP version) Enabled extensions: dom, mysqli.

Scanning files (8 threads)...

2130 / 2150...Uncaught TypeError: str_starts_with(): Argument #1 ($haystack) must be of type string, int given in /vendor/vimeo/psalm/src/Psalm/DocComment.php:57
Stack trace:
#0 /vendor/vimeo/psalm/src/Psalm/DocComment.php(57): str_starts_with()

VincentLanglet:
You should isolate the code which crash with psalm.
It seems like you could have something like a weird annotation with an integer `@42`.

Issue #11217, opened 2025-01-29
Psalm crash when parsing a PHPDoc tag that looks like a number


jlewisizone:
The following code will cause Psalm to crash:
```php
<?php
/** @1 */
```

```php
<?php
/**
 * @1
 */
```

```php
<?php
/** @1.3 */
```

The following will not crash Psalm

```php
<?php
/** @1x */
```

This can be replicated on the Psalm website: https://psalm.dev/r/d5db1ec4ea

Full stacktrace below
```

Uncaught TypeError: str_starts_with(): Argument #1 ($haystack) must be of type string, int given in C:\redacted\vendor\vimeo\psalm\src\Psalm\DocComment.php:57
Stack trace:
#0 C:\redacted\vendor\vimeo\psalm\src\Psalm\DocComment.php(57): str_starts_with(1, 'psalm-')
#1 C:\redacted\vendor\vimeo\psalm\src\Psalm\Internal\PhpVisitor\Reflector\ClassLikeNodeScanner.php(1897): Psalm\DocComment::parsePreservingLength(Object(PhpParser\Comment\Doc))
#2 C:\redacted\vendor\vimeo\psalm\src\Psalm\Internal\PhpVisitor\ReflectorVisitor.php(110): Psalm\Internal\PhpVisitor\Reflector\ClassLikeNodeScanner::getTypeAliasesFromComment(Object(PhpParser\Comment\Doc), Object(Psalm\Aliases), Array, NULL)
#3 C:\redacted\vendor\nikic\php-parser\lib\PhpParser\NodeTraverser.php(196): Psalm\Internal\PhpVisitor\ReflectorVisitor->enterNode(Object(PhpParser\Node\Stmt\Nop))
#4 C:\redacted\vendor\nikic\php-parser\lib\PhpParser\NodeTraverser.php(76): PhpParser\NodeTraverser->traverseArray(Array)
#5 C:\redacted\vendor\vimeo\psalm\src\Psalm\Internal\Scanner\FileScanner.php(72): PhpParser\NodeTraverser->traverse(Array)
#6 C:\redacted\vendor\vimeo\psalm\src\Psalm\Internal\Codebase\Scanner.php(535): Psalm\Internal\Scanner\FileScanner->scan(Object(Psalm\Codebase), Object(Psalm\Storage\FileStorage), false, Object(Psalm\Progress\DefaultProgress))
[...]

PR #11445, opened 2025-05-20
Fix crash on numeric annotation

VincentLanglet:
Closes https://github.com/vimeo/psalm/issues/11217
Closes https://github.com/vimeo/psalm/issues/11442

If I keep `@param array<string, array<int, string>> $specials`, the `is_string` check is reported as useless, so I'm not sure if you prefer @danog 
- To change it to `@param array<int|string, array<int, string>> $specials` like I did
- Or to suppress the psalm error on the useless `is_string` check

### DIFF
--- a/src/Psalm/DocComment.php
+++ b/src/Psalm/DocComment.php
@@ -55,7 +55,8 @@ final class DocComment
         }
 
         foreach ($parsed_docblock->tags as $special_key => $_) {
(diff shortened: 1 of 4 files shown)


danog:
Let's turn this into a cast + suppress the issue on the is_string check

danog:
Thank you!

PR #11446, opened 2025-05-21
Avoid Uncaught TypeError when using str_starts_with()

10n:
Cast $special_key to string to avoid Uncaught TypeError: str_starts_with(): Argument #1 ($haystack) must be of type string, int given


danog:
Fixed by https://github.com/vimeo/psalm/pull/11445

PR #11453, opened 2025-05-28
Update stubs for ext-mongodb 2.0

alcaeus:
This PR updates the function map for version 2.0 of the MongoDB extension.

As a semi-related question, we've recently released version 2.1, but the docker image used to regenerate the dictionaries doesn't yet have that version installed. How often are these images updated, and is this something we can trigger on our own after releasing a new version?

### DIFF
--- a/dictionaries/CallMap_81.php
+++ b/dictionaries/CallMap_81.php
@@ -46838,15 +46838,6 @@ return array (
   array (
     0 => 'mixed',
   ),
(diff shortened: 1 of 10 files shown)


danog:
The docker image is automatically rebuilt on every commit by https://github.com/vimeo/psalm/blob/6.x/bin/stubs/build_docker.sh, it was last built a few days ago, and it will get rebuilt now after this PR is merged

danog:
Thank you!

PR #11459, opened 2025-05-30
feat(analyzer): add `HiddenGeneratorReturn` issue

### LINKED ISSUE #11458: incorrect `InvalidReturnStatement` when returning from an iterable generator.
Psalm incorrectly triggers an `InvalidReturnStatement` when a generator function, with a declared return type of `iterable`, uses a return statement with a value. It appears Psalm attempts to validate this return value against the `iterable<K, V>`, rather than treating it as the generator's `R` type.

If such a function has a declared return type like `iterable<K, V>` or `iterable<V>`, its actual signature is closer to `Generator<K, V, S = mixed, R = mixed>`.

The `return $value;` statement within a generator sets the `R` type parameter. This value is distinct from the `iterable`.

Currently, Psalm seems to incorrectly assume that `$value` in `return $value;` must conform to `iterable<K, V>`.

The following code demonstrates the issue:

```php
<?php

/**
 * @param array<string, string> $array
 * @return iterable<string>
 */
function generator(array $array): iterable
{
    yield $array['key'] ?? 'default';

    return 1; // this is an error, because `1` does not match `iterable<string>`
}

/**
 * @param array<string, string> $array
 * @return iterable<string>
 */
function generator2(array $array): iterable
{
    yield $array['key'] ?? 'default';

    return []; // this is fine because `[]` matches `iterable<string>`
}
```

Suggested Resolution:

When analyzing a generator function with an `iterable` return type:

- Recognize that `return $value;` defines the `R` type of the underlying `Generator`.
- do not validate `$value` against the `iterable<K, V>`
- If the `R` type needs to be statically checked, Psalm could perhaps assume `R` is mixed when only `iterable` is specified, or require the more explicit `Generator` type hint if the `R` type is important to the user.

The current behavior is misleading because the value from return in a generator is not meant to be an instance of the `iterable` type itself, but rather the result obtained from `Generator::getReturn()`.

azjezz:
closes #11458

### DIFF
--- a/src/Psalm/Issue/HiddenGeneratorReturn.php
+++ b/src/Psalm/Issue/HiddenGeneratorReturn.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Issue;
+
+final class HiddenGeneratorReturn extends CodeIssue
+{
+    public const ERROR_LEVEL = 4;
+    public const SHORTCODE = 284;
+}
--- a/tests/Loop/ForeachTest.php
+++ b/tests/Loop/ForeachTest.php
@@ -1176,7 +1176,10 @@ final class ForeachTest extends TestCase
             'generatorWithUnspecifiedSend' => [
                 'code' => <<<'PHP'
                     <?php
-                    /** @return Generator<int,int> */
+                    /**
+                     * @return Generator<int,int>
+                     * @psalm-suppress MixedReturnStatement
+                     */
                     function gen() : Generator {
                         return yield 1;
                     }
@@ -1187,7 +1190,10 @@ final class ForeachTest extends TestCase
             'generatorWithMixedSend' => [
                 'code' => <<<'PHP'
                     <?php
-                    /** @return Generator<int,int, mixed, mixed> */
+                    /**
+                     * @return Generator<int,int, mixed, mixed>
+                     * @psalm-suppress MixedReturnStatement
+                     */
                     function gen() : Generator {
                         return yield 1;
                     }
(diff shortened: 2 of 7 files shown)


src/Psalm/Issue/HiddenGeneratorReturn.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\Issue;
+
+final class HiddenGeneratorReturn extends CodeIssue
+{
+    public const ERROR_LEVEL = 4;
+    public const SHORTCODE = 284;
```
azjezz, line 10: not sure if those values are correct.

azjezz:
> Error: src/Psalm/Type/UnionTrait.php:950:21: PossiblyUnusedMethod: Cannot find any calls to method Psalm\Type\Union::isGenerator (see https://psalm.dev/087)

Is it safe to remove `isGenerator` from `UnionTrait`? a plugin could be depending on it..

tests/Loop/ForeachTest.php
```diff
@@ -1187,7 +1190,10 @@ function gen() : Generator {
             'generatorWithMixedSend' => [
                 'code' => <<<'PHP'
                     <?php
-                    /** @return Generator<int,int, mixed, mixed> */
+                    /**
+                     * @return Generator<int,int, mixed, mixed>
+                     * @psalm-suppress MixedReturnStatement
+                     */
```
azjezz, line 1196: same here

tests/Loop/ForeachTest.php
```diff
@@ -1176,7 +1176,10 @@ function f(array $a): array {
             'generatorWithUnspecifiedSend' => [
                 'code' => <<<'PHP'
                     <?php
-                    /** @return Generator<int,int> */
+                    /**
+                     * @return Generator<int,int>
+                     * @psalm-suppress MixedReturnStatement
+                     */
```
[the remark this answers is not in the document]
azjezz: It could make sense to not report `MixedReturnStatement` here actually, we do that in mago, the reason is the return type of the function is not `mixed` but rather `TReturn` on the `Generator` type is, so it might be a good idea to either not report it, or have a different error code that tells user to have a better `TReturn` for the generator other than `mixed` 🤔

PR #11467, opened 2025-06-03
fix(analyzer): correct composite string handling, and improve string casting

### LINKED ISSUE #11464: invalid `UnusedVariable` when the variable is used in a composite string.
Ref: https://psalm.dev/r/041f63fa9f

In `EncapsulatedStringAnalyzer`, before analzing an expression part, `in_general_use` prop of `context` should have been set to true, but that does not seem to be the case now, which results in unused variable error, even when the variable is indeed used.

See: https://github.com/vimeo/psalm/blob/cf420941d061a57050b6c468ef2c778faf40aee2/src/Psalm/Internal/Analyzer/Statements/Expression/EncapsulatedStringAnalyzer.php#L51-L55

### LINKED ISSUE #11465: non-empty composite string made up of only expressions is incorrectly inferred as `string` instead of `non-empty-string`
Ref: https://psalm.dev/r/fc8c7e3ae2

Expected: No issues

---

Psalm seems to fail here: https://github.com/vimeo/psalm/blob/cf420941d061a57050b6c468ef2c778faf40aee2/src/Psalm/Internal/Analyzer/Statements/Expression/EncapsulatedStringAnalyzer.php#L70-L86

As it assumes that the string would be non-empty, only if it was marked as non-empty *before* the expression part, but that does not have to be the case, in the example above, the only part in the string is `$x` which is never checked for "non-empty" as `$non_empty` is set to `false` by default.

---

<details>

<summary>For reference, this is the logic used in mago, which correctly identifies the result as non-empty-string:</summary>

```rust

impl Analyzable for CompositeString {
    fn analyze<'a>(
        &self,
        context: &mut Context<'a>,
        block_context: &mut BlockContext<'a>,
        analysis_data: &mut FunctionAnalysisData,
    ) -> Result<(), AnalysisError> {
        let mut non_empty = false;
        let mut all_literals = true;
        let mut resulting_string = Some(String::new());

        for part in self.parts().as_slice() {
            let part_type = match part {
                StringPart::Literal(literal_string_part) => {
                    let literal_string_value = context.interner.lookup(&literal_string_part.value);

                    non_empty = non_empty || !literal_string_value.is_empty();

                    if let Some(resulting_string) = resulting_string.as_mut() {
                        resulting_string.push_str(literal_string_value);
                    }

                    continue;
                }
                StringPart::Expression(expression) => {
                    let was_inside_general_use = block_context.inside_general_use;
                    block_context.inside_general_use = true;
                    expression.analyze(context, block_context, analysis_data)?;
                    block_context.inside_general_use = was_inside_general_use;

                    analysis_data.get_expression_type(expression)
                }
                StringPart::BracedExpression(braced_expression) => {
                    let was_inside_general_use = block_context.inside_general_use;
                    block_context.inside_general_use = true;
                    braced_expression.expression.analyze(context, block_context, analysis_data)?;
                    block_context.inside_general_use = was_inside_general_use;

                    analysis_data.get_expression_type(&braced_expression.expression)
                }
            };

            let Some(part_type) = part_type else {
                all_literals = false;
                resulting_string = None;

                // TODO: maybe it is worth reporting an issue here?
                continue;
            };

            let casted_part_type = cast_type_to_string(part_type, context, part.span());
            let mut is_non_empty_part = true;
            let mut is_part_literal = true;
            let mut part_literal_string = None;
            let mut could_specify_literals = true;

            for atomic_literal in &casted_part_type.types {
                is_non_empty_part = is_non_empty_part && atomic_literal.is_non_empty_string();

                // Get the literal string specification, if any.
                let TAtomic::Scalar(TScalar::String(TString { literal: Some(literal), .. })) = atomic_literal else {
                    is_part_literal = false;

                    continue;
                };

                if !is_part_literal {
                    continue;
                }

                match literal {
                    TStringLiteral::Unspecified => {
                        resulting_string = None;
                        part_literal_string = None;
                        could_specify_literals = false;
                    }
                    TStringLiteral::Value(literal_string) => {
                        if !could_specify_literals {
                            // We can't specify literals for this part, as
                            // one or more parts contain specified literal strings
                            // that are not identical.
                            continue;
                        }

                        match part_literal_string {
                            None => {
                                part_literal_string = Some(literal_string);
                            }
                            Some(previous_part_string) if !previous_part_string.eq(literal_string) => {
                                // This part of the string type contain a specified literal string
                                // that is different from a previous part.
                                part_literal_string = None;
                                could_specify_literals = false;
                            }
                            _ => {
                                // This part literal is equal to the previous one
                            }
                        }
                    }
                }
            }

            non_empty = non_empty || is_non_empty_part;
            all_literals = all_literals && is_part_literal;

            if !is_part_literal {
                resulting_string = None;
            } else if could_specify_literals {
                if let Some(part_literal_string) = part_literal_string {
                    if let Some(resulting_string) = resulting_string.as_mut() {
                        resulting_string.push_str(&part_literal_string);
                    }
                }
            }
        }

        let resulting_type = if non_empty {
            if let Some(literal_string) = resulting_string {
                get_literal_string(literal_string)
            } else if all_literals {
                get_non_empty_unspecified_literal_string()
            } else {
                get_non_empty_string()
            }
        } else if all_literals {
            get_unspecified_literal_string()
        } else {
            get_string()
        };

        analysis_data.set_expression_type(self, resulting_type);

        Ok(())
    }
}
```

</details>

### LINKED ISSUE #11466: casting `array`/`list` to a `string` should result in literal `"Array"` instead of generic `string`
Ref: https://psalm.dev/r/00980af153

I would expect psalm to still emit an error or a warning here, but the result should be `"Array"`, not `string`.

3v4l: https://3v4l.org/i3Z9h

azjezz:
closes #11465
closes #11466
closes #11464

### DIFF
--- a/tests/ToStringTest.php
+++ b/tests/ToStringTest.php
@@ -208,6 +208,16 @@ final class ToStringTest extends TestCase
 
                     echo implode(":", [new Bar()]);',
             ],
+            'resourceOrFalseToString' => [
+                'code' => '<?php
+                    $a = fopen("php://memory", "r");
+                    if (rand(0, 1)) {
+                        $a = [];
+                    }
+
+                    // b = "Array"|""|"Resource id #1"
+                    $b = (string) $a;',
+            ],
         ];
     }
 
@@ -313,15 +323,6 @@ final class ToStringTest extends TestCase
                     $a = fopen("php://memory", "r");
                     takesString($a);',
                 'error_message' => 'InvalidArgument',
-            ],
-            'resourceOrFalseToString' => [
-                'code' => '<?php
-                    $a = fopen("php://memory", "r");
-                    if (rand(0, 1)) {
-                        $a = [];
-                    }
-                    $b = (string) $a;',
-                'error_message' => 'PossiblyInvalidCast',
             ],
             'cannotCastInsideString' => [
                 'code' => '<?php
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1175,11 +1175,6 @@
       <code><![CDATA[$context->self]]></code>
     </RiskyTruthyFalsyComparison>
   </file>
-  <file src="src/Psalm/Internal/Analyzer/Statements/Expression/EncapsulatedStringAnalyzer.php">
-    <PossiblyInvalidOperand>
-      <code><![CDATA[$casted_part_type->getSingleLiteral()->value]]></code>
-    </PossiblyInvalidOperand>
-  </file>
   <file src="src/Psalm/Internal/Analyzer/Statements/Expression/ExpressionIdentifier.php">
     <ImplicitToStringCast>
       <code><![CDATA[$stmt->dim->name]]></code>
@@ -2450,9 +2445,6 @@
     <PossiblyInvalidArgument>
       <code><![CDATA[$this->link_format]]></code>
     </PossiblyInvalidArgument>
-    <PossiblyInvalidCast>
-      <code><![CDATA[$this->link_format]]></code>
-    </PossiblyInvalidCast>
     <PossiblyInvalidPropertyAssignmentValue>
       <code><![CDATA[(
                 ini_get('xdebug.file_link_format') ?: get_cfg_var('xdebug.file_link_format')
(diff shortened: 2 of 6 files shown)


tests/ToStringTest.php
```diff
@@ -208,6 +208,16 @@ public function __toString() {
...
+                'code' => '<?php
+                    $a = fopen("php://memory", "r");
+                    if (rand(0, 1)) {
+                        $a = [];
+                    }
+
+                    // b = "Array"|""|"Resource id #1"
+                    $b = (string) $a;',
+            ],
         ];
```
azjezz, line 221: previously considered failing due to the assumpton plsam made that `[]` can't be cast to string.

azjezz:
Unrelated note, there seem to be too many classes to represent string currently, I think it would make sense to reduce all `T*String` classes into a single, `TString` class, with properties that act as flags + `literal` string that can be either `null` ( not a literal ), or an instance of `LiteralString` which can have a value or not ( specified literal vs unspecified ).

Issue #11491, opened 2025-06-25
"Could not get class storage for"


HenkPoley:
From Issue #10895

Repo with crashing case, and some non crashing cases: https://github.com/HenkPoley/PsalmCouldNotGetClassStorageFor

Looks like there are ways to trigger a load of the throwable type class, and then psalm is fine.

`allowStringToStandInForClass="true"` (or false) is irrelevant to this crash (`php ./composer.phar crash-irrelevant-allowStringToStandInForClass`).

```
php ./composer.phar crash
[..]
Analyzing files...
Analyzing /Users/henk/PsalmCouldNotGetClassStorageFor/src/CrashingClass.php
/Users/henk/PsalmCouldNotGetClassStorageFor/src/CrashingClass.php:13
/Users/henk/PsalmCouldNotGetClassStorageFor/src/CrashingClass.php:17
Uncaught InvalidArgumentException: Could not get class storage for assertionerror in /Users/henk/PsalmCouldNotGetClassStorageFor/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:45
Stack trace:
#0 /Users/henk/PsalmCouldNotGetClassStorageFor/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php(329): Psalm\Internal\Provider\ClassLikeStorageProvider->get('AssertionError')
#1 /Users/henk/PsalmCouldNotGetClassStorageFor/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/NewAnalyzer.php(212): Psalm\Internal\Analyzer\ClassLikeAnalyzer::checkFullyQualifiedClassLikeName(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), 'AssertionError', Object(Psalm\CodeLocation), 'Henk\\PsalmCould...', 'henk\\psalmcould...', Array)
#2 /Users/henk/PsalmCouldNotGetClassStorageFor/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/Statements/ExpressionAnalyzer.php(306): Psalm\Internal\Analyzer\Statements\Expression\Call\NewAnalyzer::analyze(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(PhpParser\Node\Expr\New_), Object(Psalm\Context), NULL)
[...]

PR #11501, opened 2025-07-07
InvalidArgumentException due to dynamic class instantiation using a variable (Issue #11491)

HenkPoley:
Add missing (built-in?) classes in Codebase->classlike_storage_provider using ReflectionClass.

Similar to \Psalm\Internal\Codebase\Scanner::convertClassesToFilePaths()

Should fix issue #11491.


HenkPoley:
Hmm, that looks like a bunch of unrelated stuff (?)

PR #11517, opened 2025-07-17
Use more relax php version range

samsonasik:
on some system, eg: macos using macports, the php version latest may not up to date, I think it is better to have more relax version to avoid error:

```
vimeo/psalm[6.12.0, ..., 6.13.0] require php ~8.1.31 || ~8.2.27 || ~8.3.16 || ~8.4.3 -> your php version (8.2.24) does not satisfy that requirement.
```


samsonasik:
Ready to merge 👍

gonzalo1987:
Please yes 🙏

danog:
The version range was chosen due to critical bugs in older PHP versions that directly affect psalm.

samsonasik:
psalm is mostly used as dev dependency, imo, it doesn't need to be in latest php, as CI is also green

gonzalo1987:
> psalm is mostly used as dev dependency, imo, it doesn't need to be in latest php, as CI is also green

I think so, It would be nice if it were considered

danog:
You don't understand, the php bugs affect directly Psalm, causing crashes and false positives or negatives during scans.

Issue #11555, opened 2025-09-12
Psalm fails to run if its path is not in the `opcache.restrict_api` setting.


dmvrtx:
Opcache extension has an `opcache.restrict_api` settings to allow access to the API only for the scripts within the specified path. Unfortunately, API still remains declared, i.e. `function_exists('opcache_get_status')` will return true, while calling `opcache_get_status` will result in a `RuntimeException`.

### To reproduce

 - Edit the appropriate `php.ini` file on your platform and set `opcache.restrict_api="/tmp"` or any similar path.
 - Try running `psalm`
 - The error occurs:

```
Running on PHP 8.4.11, Psalm 6.13.1@1e3b7f0a8ab32b23197b91107adc0a7ed8a05b51.
Uncaught RuntimeException: PHP Error: Zend OPcache API is restricted by "restrict_api" configuration directive in /<redacted>/composer/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php:999 for command with CLI args "/<redacted>/vendor/bin/psalm" in /<redacted>/composer/vendor/vimeo/psalm/src/Psalm/Internal/ErrorHandler.php:76
Stack trace:
#0 [internal function]: Psalm\Internal\ErrorHandler::{closure:Psalm\Internal\ErrorHandler::installErrorHandler():69}()
#1 /<redacted>/composer/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(999): opcache_get_status()
#2 /<redacted>/composer/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(290): Psalm\Internal\Cli\Psalm::restart()
#3 /<redacted>/composer/vendor/vimeo/psalm/psalm(9): Psalm\Internal\Cli\Psalm::run()
#4 /<redacted>/composer/vendor/bin/psalm(119): include('...')
#5 {main}
(Psalm 6.13.1@1e3b7f0a8ab32b23197b91107adc0a7ed8a05b51 crashed due to an uncaught Throwable)
```

PR #11556, opened 2025-09-12
Fix the opcache status retrieval if restricted paths are set.

dmvrtx:
Fixes vimeo/psalm#11555

To avoid a `RuntimeException` if opcache API cannot be used, wrap a call to `opcache_get_status` in a `try ... catch` and use the result in the later comparisons.

### Testing instructions

 - Edit the appropriate `php.ini` file on your platform and set `opcache.restrict_api="/tmp"` or any similar path.
 - Try running `psalm` from the branch
 - It should proceed with the following message:
 
```
Running on PHP 8.4.11, Psalm 6.13.1@1e3b7f0a8ab32b23197b91107adc0a7ed8a05b51.

JIT acceleration: OFF (an error occurred while enabling JIT)
Please report this to https://github.com/vimeo/psalm with your OS and PHP configuration!
```

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -996,12 +996,7 @@ final class Psalm
 
         $hasJit = false;
         if (function_exists('opcache_get_status')) {
-            try {
-                $opcache_status = opcache_get_status();
-            } catch (RuntimeException) {
-                $opcache_status = [];
-            }
-            if (true === ($opcache_status['jit']['on'] ?? false)) {
+            if (true === (opcache_get_status()['jit']['on'] ?? false)) {
                 $hasJit = true;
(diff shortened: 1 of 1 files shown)


danog:
This should be fixed by tweaking the opcache config in PsalmRestarter

dmvrtx:
> This should be fixed by tweaking the opcache config in PsalmRestarter

I'm trying to understand how it should be done, but I don't think I'm getting the idea. PsalmRestarter tries to update certain settings, but `restrict_api` can only be set in the `php.ini`. Just adding it in the PsalmRestarter produces no effect for me.

dmvrtx:
@danog , I changed the approach to override `restrict_api` through PsalmRestarter. I tested and it worked well. Thanks for the initial pointer.

danog:
Perfect, thank you!

PR #11598, opened 2025-12-02
Fix `Imagick::getImageBlob()` return type

HypeMC:
The [latest version](https://github.com/Imagick/imagick/releases/tag/3.8.1) of the Imagick extension [fixes the return type](https://github.com/Imagick/imagick/pull/735) of `Imagick::getImageBlob()`, which was previously incorrectly marked as nullable. This PR updates the callmap to align with this change.


danog:
Already fixed upstream, but thanks anyway!

PR #11596, opened 2025-12-01
ci: add nikic/php-parser deprecations to Psalm baseline

jorgsowa:
There are new deprecations in nikic
PHP-Parser v5.6.2

 https://github.com/nikic/PHP-Parser/commit/3374502720536d6a2d8c24de12be34ab2820cded

PR #11599, opened 2025-12-02
fix: deprecated null offset

jorgsowa:
Fixes null offsets in codebase for PHP 8.5

> PHP Error: Using null as an array offset is deprecated, use an empty string instead

Psalm check fails because another PR must be merged first: https://github.com/vimeo/psalm/pull/11596


danog:
Thank you

PR #11613, opened 2025-12-10
Only restart with custom JIT settings when `forceJit` is enabled

### LINKED ISSUE #11589: Option to disable JIT for better performance
I'm on Windows 11, with Psalm 6.13.1 and PHP 8.4.14 (x64 NTS build).

Every time I run Psalm, it displays "JIT compilation in progress..." for about 5 seconds before starting to scan files. There doesn't seem to be a configuration option to turn this off, but for testing I manually edited `vendor/vimeo/psalm/src/Psalm/Internal/Fork/PsalmRestarter.php`, setting `'enable_cli' => 0` in the `REQUIRED_OPCACHE_SETTINGS` array.

After doing this, I see a consistent 5 second reduction in the time it takes to run Psalm on my project. I benchmarked 10 runs before and after the above change (using `time psalm` in Bash). While memory usage increased by about 10%, the total time to run Psalm is consistently 33% faster:

Image

Perhaps there is a benefit to always enabling JIT compilation on other platforms (the docs mention an [up to +20% performance](https://psalm.dev/docs/running_psalm/configuration/#forcejit) increase), but at least on Windows it would be nice to have an option to disable it for better performance.

theodorejb:
This reduces Psalm runtime on one of my projects by 33% (from 15 seconds down to 10 seconds).

When running on the Psalm codebase itself, this has very little impact (49-50 seconds before, 46-47 seconds after this change).

Closes #11589

### DIFF
--- a/src/Psalm/Internal/Cli/Psalm.php
+++ b/src/Psalm/Internal/Cli/Psalm.php
@@ -953,6 +953,7 @@ final class Psalm
         Progress $progress,
     ): void {
         $ini_handler = new PsalmRestarter('PSALM');
+        $ini_handler->forceJit = $force_jit;
 
         if (isset($options['disable-extension'])) {
             if (is_array($options['disable-extension'])) {
@@ -1001,10 +1002,14 @@ final class Psalm
                 $progress->write(PHP_EOL
                     . 'JIT acceleration: ON'
                     . PHP_EOL . PHP_EOL);
-            } else {
+            } elseif ($force_jit) {
                 $progress->write(PHP_EOL
                     . 'JIT acceleration: OFF (an error occurred while enabling JIT)' . PHP_EOL
                     . 'Please report this to https://github.com/vimeo/psalm with your OS and PHP configuration!'
+                    . PHP_EOL . PHP_EOL);
+            } else {
+                $progress->write(PHP_EOL
+                    . 'JIT acceleration: OFF'
                     . PHP_EOL . PHP_EOL);
             }
         } else {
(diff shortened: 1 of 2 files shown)


danog:
Don't really want to merge this, as the main thing that can have some impact is preloading in combination with JIT.

On smaller codebases preloading has a negative performance impact, on an medimum-size codebases it has a positive impact.

JIT by itself without preloading has pretty much the same behavior, only the positive impact starts earlier (medium-small codebases).

I am currently working on a PHP AOT compiler (mainly for Psalm), which should solve the issue altogether.

danog:
If anything, there should be a flag to control preloading, however preloading fixes a negative performance impact caused by repeated JIT compilation in multithreaded mode.

A possible optimization path can be the reduction of the number of preloaded classes (arguably there are some classes which probably shouldn't be in the preloaded list as they aren't even used in multithreaded mode, like the LangaugeServerProtocol classes, AdvancedJsonRpc, LibDNS).

theodorejb:
There should be some way to turn off Psalm's overriding of Opcache settings, since for many projects it negatively affects performance. Even on very large projects, some may want to use different Opcache settings than the ones currently forced by Psalm.

Prior to this PR, the `forceJit` option is pointless IMO, since it merely exits the process if the JIT isn't enabled, rather than allowing users to choose whether Psalm forces its own Opcache settings.

danog:
The `forceJit` option behaves as designed, to make sure CI tasks don't suddenly stop using JIT due to a configuration change, slowing down scans.

As I said, the only thing that can really affect performance is preloading (in some conditions, on smaller projects); in most other cases, it is a performance improvement, and JIT is a performance improvement almost across the board (in fact, even scanning Psalm itself without JIT is slower than scanning it with JIT).

danog:
If you'd like to submit a PR to add an option to disable preloading (including maybe even by default when running in single-threaded mode), this will improve performance on smaller scans without hurting (or requiring manual config changes) on larger codebases.

PR #11620, opened 2025-12-13
Returning null from params provider should not overwrite already infe…

### LINKED ISSUE #11619: FunctionParamsProvider getFunctionParams() nullable but leads to type overwritten
if a plugin implements getFunctionParams but returns null, it still overwrites any existing params already inferred

This is inconsistent compared to the ReturnTypeProvider and does not allow fall back to the default value

kkmuffme:
…rred param details

Fix https://github.com/vimeo/psalm/issues/11619


kkmuffme:
Tests are failing bc the types for array_multisort in CallMap_xx.php (and _delta) are technically wrong (however no way to change this) and in ArrayMultisortParamsProvider.php it's missing handling for those cases and returns null (instead of just setting the param to functionlikeparameter)
If I remember correctly I was the one that actually created it, but in wise foreshadowing put:
> // too complex for now

Well, this needs to get fixed/improved at some point :-)

@danog is it ok to just @psalm-suppress the issue in the shepherd issue for now? As I don't really have time to improve the array_multisort atm

PR #11625, opened 2025-12-21
Remove redefinition of Attribute constants.

### LINKED ISSUE #11610: Attribute is not repeatable error in PHP 8.5
For all repeatable attributes I get the following error in Psalm 6.14.1 and PHP 8.5 (when they are repeated):

```
ERROR: InvalidAttribute - src/App/Controller/CustomerController.php:41:7 - Attribute Route is not repeatable (see https://psalm.dev/242)
    #[Route('/customer/create', methods: ['GET', 'POST'], name: 'customer_create')]
```

In PHP 8.4 this error does not occur, so something about PHP 8.5 seems to throw Psalm off.

hschletz:
The Attribute::* class constants were redefined "since that class might not exist at runtime". That applies only to PHP < 8.0. Now that Psalm requires PHP 8.1 or later, the redefinition is no longer necessary. Remove the redefinition and use Attribute::* instead.

This fixes #11610, caused by PHP 8.5 having changed the internal value for Attribute::IS_REPEATABLE from 64 to 128.

### DIFF
--- a/src/Psalm/Internal/Analyzer/AttributesAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/AttributesAnalyzer.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Psalm\Internal\Analyzer;
 
+use Attribute as GlobalAttribute;
 use Generator;
 use PhpParser\Node\Arg;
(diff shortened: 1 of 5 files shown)


danog:
Thank you!

PR #11641, opened 2026-01-08
Fix and improve return and ref/param-out type of array_pop for keyed list and array

### LINKED ISSUE #11590: list is erroneously defined as non-empty-list
I'm not a PHP developer and I might be doing something wrong, but in this case it seems like the condition and adding the second element to the array are not being calculated correctly. This code violates the non-empty list invariant.

```
/** @param non-empty-list<int> $a */ 
function foo(array $a): void { echo $a[0]; }

$d = [];
$d[] = 0;
if (time() & 1) {
    $d[] = 2;
}
array_pop($d);
foo($d); // No error!
```

https://psalm.dev/r/60f16c5ce5

https://phpstan.org/r/4710c606-9675-49a3-9b2a-bd11a4e56fc3

kkmuffme:
- Fix https://github.com/vimeo/psalm/issues/11590
- Don't collapse keyed array with all optional keys to generic array on function call (array_pop)


kkmuffme:
Failing tests unrelated

PR #11569, opened 2025-10-20
Support PHP 8.4 interface property hooks

adamaveray:
Adds support for PHP 8.4 property hooks, specifically addressing the parser issue being triggered for property hooks on interfaces.

This is the syntax that this PR adds support for: https://psalm.dev/r/6e926c6c01

Closes #11452.

PR #11649, opened 2026-01-20
Array map impure unlink unused

### LINKED ISSUE #6687: Unused function call for array_map unlink
https://psalm.dev/r/33a00f8479

While technically unused, PHP throws notices if any of the files in the array don't exist (and therefore cannot be unlinked).
Most people would not use the return value here.

Since psalm does not complain about an unused return value of "unlink" it should not complain about unused return value when "unlink" is used in array_map*

*I assume there are other similar functions, for which this should apply too - if psalm does not complain about unused return value if the function is called standalone, it should not complain about unused return value if it's used in array_map

kkmuffme:
- Fix https://github.com/vimeo/psalm/issues/6687
- consistency/stability improvements for must-use and never return callable
- fix impure PHP native functions legacy reported as pure when using Reflection because it was forgotten to add the newly added impure handling there

### DIFF
--- a/src/Psalm/Internal/Codebase/Functions.php
+++ b/src/Psalm/Internal/Codebase/Functions.php
@@ -466,7 +466,15 @@ final class Functions
 
         if (!isset($function_callable->params)
             || ($args !== null && count($args) === 0)
-            || ($function_callable->return_type && $function_callable->return_type->isVoid())
+        ) {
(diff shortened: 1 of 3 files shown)


kkmuffme:
Failing tests unrelated from https://github.com/vimeo/psalm/pull/11569

PR #11691, opened 2026-02-28
Fix missing and inconsistent reporting of type hint issues and messages

### LINKED ISSUE #4047: False MethodSignatureMismatch when a trait implements an interface method.
I'm using traits to satisfy an Interface in multiple objects, I get MethodSignatureMismatch I think it's a false positive, since the code works fine on php7.4 and php8.0.

Example: https://psalm.dev/r/722d154afb

### LINKED ISSUE #11686: MethodSignatureMismatch suggests docblock-like type instead of native type hint type
https://psalm.dev/r/36d9734244

Got
>Argument 1 of Y::run has wrong type 'array<array-key, mixed>', expecting 'array<array-key, mixed>|string' as defined by X::run

Expected 
>Argument 1 of Y::run has wrong type 'array', expecting 'array|string' as defined by X::run

Also the error message is a bit too restrictive - since it expects that type or a covariant/contravariant type of it

### LINKED ISSUE #11687: False positive MethodSignatureMismatch errors if parent class is stubbed
https://psalm.dev/r/7955401cd3

If the parent class A is stubbed, it reports an error that Argument 1 of B::run has wrong type, expecting array{hello: int|non-empty-string, world: array}|false but array<array-key, mixed>|false provided

### LINKED ISSUE #11688: #[ReturnTypeWillChange] does not report an error when used on non-built ins and also causes MethodSignatureMismatch to not report on built-ins
psalm.dev/r/281be02fc0

#[ReturnTypeWillChange] only works on built-ins, should report an error when used on non-built ins

Also psalm does not report an error when #[ReturnTypeWillChange] is used on non-built ins and no return type is specified - this must report an error, since it will be a fatal error when the code is actually run

### LINKED ISSUE #11689: If param has type hint but parent param does not, it results in a fatal error but psalm does not report an error for it
https://psalm.dev/r/0b3fcac776

### LINKED ISSUE #11690: Error message for no return type hint is misleading, since it shows empty string instead of mixed
https://psalm.dev/r/fce8359883

>Method B::run with return type '' is different to return type 'array<array-key, mixed>|false' of inherited method A::run

kkmuffme:
* Fix https://github.com/vimeo/psalm/issues/11686
* Fix https://github.com/vimeo/psalm/issues/11687
* Reduce duplicate code related to param type hint validation of stubbed/non-stubbed code (related to https://github.com/vimeo/psalm/issues/11687)
* Fix https://github.com/vimeo/psalm/issues/11688
* Fix https://github.com/vimeo/psalm/issues/11689
* Fix https://github.com/vimeo/psalm/issues/11690
* Fix https://github.com/vimeo/psalm/issues/4047

### DIFF
--- a/src/Psalm/Internal/Analyzer/MethodComparator.php
+++ b/src/Psalm/Internal/Analyzer/MethodComparator.php
@@ -21,6 +21,7 @@ use Psalm\Internal\Type\TypeExpander;
 use Psalm\Issue\ConstructorSignatureMismatch;
 use Psalm\Issue\ImplementedParamTypeMismatch;
 use Psalm\Issue\ImplementedReturnTypeMismatch;
+use Psalm\Issue\InvalidAttribute;
 use Psalm\Issue\LessSpecificImplementedReturnType;
 use Psalm\Issue\MethodSignatureMismatch;
 use Psalm\Issue\MethodSignatureMustProvideReturnType;
@@ -37,7 +38,6 @@ use Psalm\Storage\ClassLikeStorage;
 use Psalm\Storage\FunctionLikeParameter;
 use Psalm\Storage\MethodStorage;
 use Psalm\Type;
-use Psalm\Type\Atomic\TNull;
 use Psalm\Type\Atomic\TTemplateParam;
 use Psalm\Type\Union;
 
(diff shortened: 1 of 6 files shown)


kkmuffme:
Didn't update the baseline to avoid conflicts when merging, if it gets updated by another PR in the meantime (and also bc of the issue I fixed in https://github.com/vimeo/psalm/pull/11698 I cannot update the baseline either until that PR is merged)

PR #11700, opened 2026-03-01
Fix InvalidDocblock for assertions could not be suppressed

### LINKED ISSUE #11662: Impossible to suppress InvalidDocblock from library code
I'm not quite sure, maybe this is related to #11385.

I'm using the `sensiolabs/gotenberg-bundle`, which itself seems to only use @phpstan tags.
One of their classes defines a function, which uses @phpstan-assert tags.
https://github.com/sensiolabs/GotenbergBundle/blob/1.x/src/Builder/Result/GotenbergFileResult.php#L34-L42

When calling that function from my code, psalm will emit an InvalidDocblock error on my call of that function.
It seems to be impossible to suppress that error on the calling side. No matter where I put the @psalm-suppress, it keeps complaining.

I've created a reproducer here: https://psalm.dev/r/dbf0e9e8e4

Keep in mind, that I can't modify the doc block on the method itself, as it actually comes from a library.

kkmuffme:
Fix https://github.com/vimeo/psalm/issues/11662

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/CallAnalyzer.php
@@ -682,6 +682,7 @@ abstract class CallAnalyzer
                             'Assert notation is malformed',
                             new CodeLocation($statements_analyzer, $expr),
                         ),
+                        $statements_analyzer->getSuppressedIssues(),
                     );
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


danog:
I'd rather fix the parsing of phpstan blocks instead...

kkmuffme:
The primary issue here isn't with phpstan doc, but with any invalid docblock (including psalm) from an external lib that cannot be suppressed - that's what this PR fixes.

PR #11720, opened 2026-03-10
Regenerate callmaps to include new ext-mongodb 2.2 API

alcaeus:
I apologise for the size of the diff. I regenerated the callmaps using `./bin/stubs/gen_callmap.sh`, expecting only changes related to the new version of the extension, but it included a host of changes in dictionaries/Callmap_73.php that I couldn't explain.


danog:
Thank you!

PR #11728, opened 2026-03-13
Add TDate template parameter to DatePeriod stubs

### LINKED ISSUE #11727: DatePeriod stub missing TDate template: iteration type not preserved
## Description

The `DatePeriod` stub in `stubs/Php82.phpstub` has a `Start` template for constructor overload resolution but lacks a template parameter for the **iteration value type**. This means iterating over a `DatePeriod` constructed with `DateTime` objects yields `DateTimeInterface` instead of `DateTime`.

## Current stub

```php
/**
 * @template-covariant Start of string|DateTimeInterface
 * @implements IteratorAggregate<int, DateTimeInterface>  // <-- hardcoded
 */
class DatePeriod implements IteratorAggregate
{
    /** @psalm-return (Start is string ? Iterator<int, DateTime> : Iterator<int, Start>) */
    public function getIterator(): Iterator {}
}
```

The `getIterator()` method correctly resolves `Start` to the concrete type, but the `@implements` line hardcodes `DateTimeInterface`. This causes inconsistency — tools that check the `@implements` declaration (e.g., for `TooManyTemplateParams` validation against other stubs) see only 1 template param, while PHPStorm stubs declare 2.

## PHPStorm stubs (used by PHPStan) for comparison

```php
/**
 * @template TDate of DateTimeInterface
 * @template TEnd of ?DateTimeInterface
 * @implements IteratorAggregate<int, TDate>
 */
class DatePeriod implements IteratorAggregate
```

Key differences:
- PHPStorm has `TDate` (what gets iterated) + `TEnd` (end date type) — 2 template params
- Psalm has only `Start` (1 template param) and hardcodes the `@implements` to `DateTimeInterface`

## Problem in practice

When codebases target both PHPStan and Psalm, writing `DatePeriod<DateTime>` or `DatePeriod<DateTime, DateTime>` in PHPDoc causes `TooManyTemplateParams` in Psalm (since it only has 1 template), while bare `DatePeriod` loses type precision in PHPStan.

Additionally, the `getIterator()` conditional return type handles the typing correctly at runtime resolution, but the `@implements IteratorAggregate<int, DateTimeInterface>` declaration doesn't reflect this — any code that relies on the implemented interface type (rather than calling `getIterator()` directly) gets the less specific `DateTimeInterface`.

## Proposed fix

Add a `TDate` template parameter and use it in `@implements`:

```php
/**
 * @psalm-immutable
 *
 * @template-covariant Start of string|DateTimeInterface
 * @template-covariant TDate of DateTimeInterface
 * @implements IteratorAggregate<int, TDate>
 */
class DatePeriod implements IteratorAggregate
{
    /**
     * @param Start $start
     * @param (Start is string ? int-mask<self::EXCLUDE_START_DATE, self::INCLUDE_END_DATE> : DateInterval) $interval
     * @param (Start is string ? never : (DateTimeInterface|positive-int)) $end
     * @param (Start is string ? never : int-mask<self::EXCLUDE_START_DATE, self::INCLUDE_END_DATE>) $options
     */
    public function __construct($start, $interval = 0, $end = 1, $options = 0) {}

    /** @psalm-return Iterator<int, TDate> */
    public function getIterator(): Iterator {}
}
```

This aligns with PHPStorm stubs and preserves the concrete `DateTimeInterface` subtype through iteration. The `Start` template can remain for constructor overload resolution.

## Environment

- Psalm version: 6.x (stubs/Php82.phpstub)
- PHP version: 8.2+

alies-dev:
Fixes #11727
- Add `TDate of DateTimeInterface` template parameter to all three `DatePeriod` stubs (PHP <8.0, 8.0-8.1, 8.2+)
- Use `@psalm-this-out` on the constructor to bind `TDate` based on the `Start` parameter (`Start is string ? DateTime : Start`)
- Update `@implements` to use `TDate` instead of hardcoded `DateTimeInterface`
- Simplify `getIterator()` return type from conditional to `Iterator<int, TDate>` (PHP 8.0+)

This aligns with [PHPStorm stubs](https://github.com/JetBrains/phpstorm-stubs/blob/5f24d7aef6ada9de9a894ce899345b3054d6c639/date/date_c.php#L1144-L1151) (used by PHPStan) and resolves `TooManyTemplateParams` when codebases use `DatePeriod<DateTime, DateTime>` in PHPDocs for cross-tool compatibility.


danog:
Thanks!

Issue #11721, opened 2026-03-12
Sprintf precision parameter type misidentification


tommander:
Hello everyone,
  this simple code returns an error on the 3rd line (sprintf) and I believe it should not.

> ERROR: InvalidArgument - Argument 1 of sprintf is invalid - Precision must be an integer (see https://psalm.dev/004)

```php
$int = 1;
$flt = 1.234;
echo sprintf('%.*f', $int, $flt);
```

Feel free to play around with it. What I have tried:

1. `declare(strict_types=1);` --- doesn't work
2. `/** @var int */` --- doesn't work
3. Removing `$int` entirely and using an integer liteal instead --- doesn't work
4. Typecast `(int)` --- Psalm sees it as redundant (_yup, can't agree more..._).

If this is a bug, I can dive into it and find out why, but at this moment I'm not sure if I'm not just missing something (simple, I expect).

Debian Trixie amd64
PHP: 8.2.30 (cli)
Psalm: 6.15.1

https://psalm.dev/r/d42e5842ff

kkmuffme:
Yeah it's a bug that should be easy to fix
Fyi: be aware that in PHP <8 this behaves differently though https://3v4l.org/6tsnp#veol

Some time ago I worked on it, check https://github.com/vimeo/psalm/pull/10088 and https://github.com/vimeo/psalm/pull/9877 (there are probably newer ones, they're just the first of mine I found, since I did most of the implementation for it afair) 

These should give you some quick insight and help you get started on a PR (check the actual/latest code obviously, since I guess the sprintf provider has changed slightly since those PRs)

PR #11729, opened 2026-03-15
Fix false InvalidArgument for sprintf precision placeholders

eyupcanakman:
Fix https://github.com/vimeo/psalm/issues/11721

Psalm validates some sprintf/printf formats by calling them with dummy values. For `.*` precision placeholders that used empty strings, PHP 8 raises a false `ValueError` and Psalm reported `InvalidArgument`.

Validate complex `*` width and precision placeholders with integer dummies before falling back to the generic return type. Add regressions for `%.*f`, positional variants, and related arity errors.


danog:
Thanks!

PR #11733, opened 2026-03-17
[7.x] Support parenthesized union types in @method parameter annotations

### LINKED ISSUE #11730: Parser fails on parenthesized union types in @method parameter types
## Description

Psalm's type parser fails when an `@method` annotation uses parenthesized union types for parameters, such as:

```php
@method $this lock(('none'|'shared'|'default'|'exclusive') $value)
```

This causes `TypeParseTreeException('Unrecognised token (')` during docblock parsing, which silently discards **all** `@method` annotations from the class. As a result, every magic method call on that class triggers `UndefinedMagicMethod`.

## Impact

This directly affects **Laravel 12+**, where `ColumnDefinition` and `ForeignKeyDefinition` use this syntax:

```php
/**
 * @method $this nullable(bool $value = true)
 * @method $this default(mixed $value)
 * @method $this lock(('none'|'shared'|'default'|'exclusive') $value)
 * ...
 */
class ColumnDefinition extends Fluent {}
```

Because parsing fails on the `lock()` annotation, Psalm loses all `@method` tags, causing hundreds of `UndefinedMagicMethod` false positives in any project with migrations:

```
ERROR: UndefinedMagicMethod
Magic method Illuminate\Database\Schema\ColumnDefinition::nullable does not exist

ERROR: UndefinedMagicMethod
Magic method Illuminate\Database\Schema\ForeignKeyDefinition::references does not exist
```

The syntax `('none'|'shared'|'default'|'exclusive')` is valid in PHPStan's phpdoc-parser and is increasingly used by libraries.

## Reproduction

```php
<?php

/**
 * @method $this foo('a'|'b' $value)
 * @method $this bar(('a'|'b') $value)
 */
class Test {
    public function __call(string $name, array $arguments): static { return $this; }
}

(new Test())->foo('a'); // ✅ OK
(new Test())->bar('a'); // ❌ UndefinedMagicMethod
```

https://psalm.dev/r/5d8bbcb3fa

## Root cause

In `ParseTreeCreator::handleOpenRoundBracket()`, when `current_leaf` is a `Value` node, the parser throws unconditionally:

```php
if ($this->current_leaf instanceof Value) {
    throw new TypeParseTreeException('Unrecognised token (');
}
```

This prevents parenthesized expressions from being used as type grouping inside `@method` parameter types. Additionally, `createMethodParam()` uses a simple `)` check without tracking nesting depth, so even if the first issue were fixed, the method boundary detection would break.

## Expected behavior

Psalm should parse `('none'|'shared'|'default'|'exclusive')` as a grouped union type (same as `'none'|'shared'|'default'|'exclusive'`) when used as a parameter type in `@method` annotations.

## Versions

- Psalm: 7.0.0-beta16
- PHP: 8.3+
- Laravel: 12.54.1

alies-dev:
Fixes #11730

The regex-based end-of-method detection in `ClassLikeDocblockParser` matched the **first** `)` not preceded by `array(`, without tracking nesting depth. When `@method` parameter types used parenthesized unions like `('none'|'shared'|'default'|'exclusive')`, the regex matched the `)` closing the union type instead of the `)` closing the method parameters. This truncated the method entry prematurely, causing **all** `@method` annotations on the class to be silently discarded — resulting in hundreds of `UndefinedMagicMethod` false positives.

This directly affects **Laravel 12+**, where `ColumnDefinition` and `ForeignKeyDefinition` use this syntax.

### Changes

- Replace the regex with a nesting-depth-aware parenthesis scanner that correctly finds the matching `)` for the method's opening `(`
- Remove the now-unused `PREG_OFFSET_CAPTURE` import
- Add unit tests in `ClassLikeDocblockParserTest` covering parenthesized params with multiple params, static modifier, and return type suffix
- Add end-to-end test in `MagicMethodAnnotationTest` matching the exact Laravel `ColumnDefinition` example from the issue

### Before

```
ERROR: UndefinedMagicMethod - Magic method ColumnDefinition::nullable does not exist
ERROR: UndefinedMagicMethod - Magic method ForeignKeyDefinition::references does not exist
```

### After

All `@method` annotations are correctly parsed, including those with parenthesized union parameter types.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1933,7 +1933,6 @@
       <code><![CDATA[$e]]></code>
     </ImplicitToStringCast>
     <PossiblyUndefinedIntArrayOffset>
-      <code><![CDATA[$matches[0]]]></code>
       <code><![CDATA[$method_tree->children[0]]]></code>
       <code><![CDATA[$method_tree->children[1]]]></code>
     </PossiblyUndefinedIntArrayOffset>
--- a/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
+++ b/src/Psalm/Internal/PhpVisitor/Reflector/ClassLikeDocblockParser.php
@@ -389,7 +389,9 @@ final class ClassLikeDocblockParser
                     if ($method_close_paren !== null) {
                         $after_paren = substr($method_entry, $method_close_paren + 1);
                         // Optionally consume return type annotation after the closing paren
-                        if (preg_match('/^ ?(\: ?(\??[\\\\a-zA-Z0-9_]+))/', $after_paren, $return_matches)) {
+                        if (preg_match('/^ ?(\: ?(\??[\\\\a-zA-Z0-9_]+))/', $after_paren, $return_matches)
+                            && isset($return_matches[0])
+                        ) {
                             $end = $method_close_paren + 1 + strlen($return_matches[0]);
                             $method_entry = substr($method_entry, 0, $end);
                         } else {


danog:
Thank you!

alies-dev:
@danog 
You are too fast, I even didn't have a chance to remove WIP from the PR title 😁

PR #11735, opened 2026-03-17
[7.x] Allow plugin issues to specify custom documentation URLs

### LINKED ISSUE #11732: Allow plugin issues to specify custom documentation URLs
## Summary

Plugin-defined issues (subclasses of `PluginIssue`) have no mechanism to provide custom documentation URLs. The current `link` field in `IssueData` is derived exclusively from `SHORTCODE` and hardcoded to the `psalm.dev` domain, which is not meaningful for third-party plugin issues.

## Current behavior

In `IssueData::__construct()` ([IssueData.php, line ~49](https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Analyzer/IssueData.php#L49)):

```php
$this->link = $shortcode ? 'https://psalm.dev/' . str_pad((string) $shortcode, 3, "0", STR_PAD_LEFT) : '';
```

- Built-in Psalm issues use `SHORTCODE` to generate links like `https://psalm.dev/042`, which resolve to documentation pages.
- Plugin issues inherit `SHORTCODE = 0` from `CodeIssue`, resulting in an empty link.
- Even if a plugin overrides `SHORTCODE`, the generated URL would point to `psalm.dev`, which has no documentation for third-party issues.

## Expected behavior

Plugin authors should be able to associate documentation URLs with their custom issues. When a plugin issue is reported (in console output, SARIF, or other report formats), the link should point to the plugin's own documentation explaining the issue and how to resolve it.

## Suggested approach

Introduce an overridable constant or method on `CodeIssue` that allows specifying a custom documentation URL:

```php
abstract class CodeIssue
{
    public const DOCUMENTATION_URL = '';
    // ...
}
```

Then in `IssueData`:

```php
$this->link = match (true) {
    $documentation_url !== '' => $documentation_url,
    $shortcode > 0 => 'https://psalm.dev/' . str_pad((string) $shortcode, 3, "0", STR_PAD_LEFT),
    default => '',
};
```

This would allow plugin authors to write:

```php
final class NoEnvOutsideConfig extends PluginIssue
{
    public const DOCUMENTATION_URL = 'https://github.com/psalm/psalm-plugin-laravel/blob/master/docs/issues/NoEnvOutsideConfig.md';
}
```

Alternatively, a method-based approach (`getDocumentationUrl(): string`) would also work and provides more flexibility for dynamic URL generation.

## Context

Discovered while developing a custom rule for [psalm-plugin-laravel](https://github.com/psalm/psalm-plugin-laravel). The plugin defines issue classes that users need guidance on, but there is no way to link them to the plugin's documentation.

alies-dev:
## Summary

Plugin-defined issues (subclasses of `PluginIssue`) had no mechanism to provide custom documentation URLs. The `link` field in `IssueData` was derived exclusively from `SHORTCODE` and hardcoded to the `psalm.dev` domain, which is not meaningful for third-party plugin issues.

This adds a `DOCUMENTATION_URL` constant to `CodeIssue` that plugin authors can override to link to their own documentation:

```php
final class NoEnvOutsideConfig extends PluginIssue
{
    public const DOCUMENTATION_URL = 'https://github.com/psalm/psalm-plugin-laravel/blob/master/docs/issues/NoEnvOutsideConfig.md';
}
```

The custom URL takes priority over the shortcode-based psalm.dev link. Fully backward-compatible — existing behavior is unchanged.

Fixes #11732

### DIFF
--- a/src/Psalm/Internal/Analyzer/IssueData.php
+++ b/src/Psalm/Internal/Analyzer/IssueData.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Psalm\Internal\Analyzer;
 
+use function is_string;
 use function str_pad;
 
 use const STR_PAD_LEFT;
@@ -45,7 +46,12 @@ final class IssueData
         public ?array $taint_trace = null,
         public ?array $other_references = null,
         public readonly ?string $dupe_key = null,
+        ?string $documentation_url = null,
     ) {
-        $this->link = $shortcode ? 'https://psalm.dev/' . str_pad((string) $shortcode, 3, "0", STR_PAD_LEFT) : '';
+        $this->link = match (true) {
+            is_string($documentation_url) => $documentation_url,
+            $shortcode > 0 => 'https://psalm.dev/' . str_pad((string) $shortcode, 3, "0", STR_PAD_LEFT),
+            default => '',
+        };
     }
 }
--- a/src/Psalm/Issue/CodeIssue.php
+++ b/src/Psalm/Issue/CodeIssue.php
@@ -16,6 +16,8 @@ abstract class CodeIssue
     public const ERROR_LEVEL = -1;
     /** @var int<0, max> */
     public const SHORTCODE = 0;
+    /** @var non-empty-string|null */
+    public const DOCUMENTATION_URL = null;
 
     public ?string $dupe_key = null;
 
@@ -107,6 +109,7 @@ abstract class CodeIssue
                 ]
                 : null,
             $this->dupe_key,
+            static::DOCUMENTATION_URL,
         );
     }
 }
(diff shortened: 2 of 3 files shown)


src/Psalm/Internal/Analyzer/IssueData.php
```diff
@@ -45,7 +45,12 @@ public function __construct(
         public ?array $taint_trace = null,
         public ?array $other_references = null,
         public readonly ?string $dupe_key = null,
+        string $documentation_url = '',
     ) {
-        $this->link = $shortcode ? 'https://psalm.dev/' . str_pad((string) $shortcode, 3, "0", STR_PAD_LEFT) : '';
+        $this->link = match (true) {
+            $documentation_url !== '' => $documentation_url,
```
andrew-demb: Should we validate this as a proper URL and ensure proper sanitization depending on where it is used (e.g., UI rendering vs CLI output)?

danog:
Thanks!

PR #11736, opened 2026-03-17
[7.x] Respect `ERROR_LEVEL` for plugin-defined issues

### LINKED ISSUE #11731: PluginIssue subclasses cannot utilize ERROR_LEVEL for error-level based reporting
## Summary

`CodeIssue::ERROR_LEVEL` is not respected for plugin-defined issues that extend `PluginIssue`. Plugin authors can override `ERROR_LEVEL` in their issue classes, but the value is never evaluated by Psalm's reporting logic.

## Current behavior

In `Config::getReportingLevelForFile()` ([Config.php, line ~2112](https://github.com/vimeo/psalm/blob/master/src/Psalm/Config.php#L2112)), the method attempts to resolve the issue class under the `Psalm\Issue\` namespace:

```php
$issue_class = 'Psalm\\Issue\\' . $issue_type;

if (!class_exists($issue_class) || !is_a($issue_class, CodeIssue::class, true)) {
    return self::REPORT_ERROR;
}

$issue_level = $issue_class::ERROR_LEVEL;

if ($issue_level > 0 && $issue_level < $this->level) {
    return self::REPORT_INFO;
}
```

Since plugin issues live in their own namespace (e.g., `Vendor\Plugin\Issues\MyIssue`), the `class_exists('Psalm\Issue\MyIssue')` check fails, and the method returns `REPORT_ERROR` immediately — the `ERROR_LEVEL` check on the actual class is never reached.

As a result, all `PluginIssue` subclasses are unconditionally treated as errors regardless of their `ERROR_LEVEL` value and the project's configured `errorLevel`.

## Expected behavior

Plugin-defined issues should be able to participate in Psalm's error-level system. If a plugin issue declares `ERROR_LEVEL = 2`, it should be reported as `info` (rather than `error`) when the project's `errorLevel` is set to 3 or higher, consistent with how built-in Psalm issues behave.

## Suggested approach

Resolve the issue class from `IssueData` or look up the actual FQCN rather than assuming the `Psalm\Issue\` namespace. For example:

```php
$issue_class = 'Psalm\\Issue\\' . $issue_type;

if (!class_exists($issue_class) || !is_a($issue_class, CodeIssue::class, true)) {
    // Fallback: check if the issue type was registered by a plugin
    // and resolve its actual class to read ERROR_LEVEL
    return self::REPORT_ERROR;
}
```

## Context

Discovered while developing a custom rule for [psalm-plugin-laravel](https://github.com/psalm/psalm-plugin-laravel). The rule defines a `PluginIssue` subclass with `ERROR_LEVEL = 1`, but the constant has no effect on reporting behavior.

alies-dev:
`PluginIssue` subclasses that override `ERROR_LEVEL` had no effect on reporting behavior. All plugin issues were unconditionally treated as errors regardless of their `ERROR_LEVEL` value.

The root cause: `Config::getReportingLevelForFile()` resolves issue classes by prepending `Psalm\Issue\` to the short issue type name. Plugin issues live in their own namespaces (e.g. `Vendor\Plugin\Issues\MyIssue`), so `class_exists()` fails and the method returns `REPORT_ERROR` immediately — the `ERROR_LEVEL` check is never reached.

This adds an `ERROR_LEVEL` check in `getReportingLevelForIssue()`, which has access to the actual `CodeIssue` instance. The check only applies when no explicit issue handler is configured, so user config (e.g. `<MyIssue errorLevel="error"/>`) takes priority.

Fixes #11731

### DIFF
--- a/src/Psalm/Internal/Analyzer/IssueData.php
+++ b/src/Psalm/Internal/Analyzer/IssueData.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Psalm\Internal\Analyzer;
 
-use function is_string;
 use function str_pad;
 
 use const STR_PAD_LEFT;


danog:
Thanks!

danog:
Could you fix the shepherd issue

alies-dev:
@danog 
>Could you fix the shepherd issue

done

PR #11756, opened 2026-03-24
Add "memcached" to list of supported extensions.

ADmad:
Currently adding "memcached" to the "enableExtensions" list in `psalm.xml` generates an assertion error.


ADmad:
Does `memcached` also need to be added to the `$php_extensions_supported_by_psalm_callmaps` list?

ADmad:
Ping

Issue #11280, opened 2025-02-13
Uncaught UnexpectedValueException: $storage should not be null


gijs-blanken:
Hi everyone,

I'm running Psalm 6.5.1 and I'm running into an issue that was not present on the 5.20.0 branch when we were on PHP 8.3 (we are now on 8.4). I noticed a commit which supposedly fixed this error (https://github.com/vimeo/psalm/pull/11232). But unfortunately it persists for me. I also tested the analyses with threads=1 and scan-threads=1 but to no avail. I'm not quite sure what extra context to add, please let me know!

Thanks in advance 😅 

```shell
Uncaught UnexpectedValueException: $storage should not be null for App\Facades\Defuse::__callstatic in /www/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Methods.php:1154
Stack trace:
#0 /www/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php(528): Psalm\Internal\Codebase\Methods->getStorage(Object(Psalm\Internal\MethodIdentifier))
#1 /www/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticMethod/AtomicStaticCallAnalyzer.php(204): Psalm\Internal\Analyzer\Statements\Expression\Call\StaticMethod\AtomicStaticCallAnalyzer::handleNamedCall(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(PhpParser\Node\Expr\StaticCall), Object(PhpParser\Node\Identifier), Object(Psalm\Context), Object(Psalm\Type\Atomic\TNamedObject), Array, 'App\\Facades\\Def...', false, false, NULL)
#2 /www/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/Statements/Expression/Call/StaticCallAnalyzer.php(230): Psalm\Internal\Analyzer\Statements\Expression\Call\StaticMethod\AtomicStaticCallAnalyzer::analyze(Object(Psalm\Internal\Analyzer\StatementsAnalyzer), Object(PhpParser\Node\Expr\StaticCall), Object(Psalm\Context), Object(Psalm\Type\Atomic\TNamedObject), false, false, false, false, NULL)
[...]

PR #11760, opened 2026-03-24
[6.x] Fix InterfaceAnalyzer crash when storage is overwritten by reflection

alies-dev:
Related to [#11280](https://github.com/vimeo/psalm/issues/11280#issuecomment-4121612344)

- When a project declares an interface with the same FQCN as a vendor class loaded via reflection (e.g. in `<ignoreFiles>`), the `ClassLikeStorage` gets overwritten during population. The `InterfaceAnalyzer` then crashes trying to analyze AST methods that don't exist in the overwritten storage.
- Adds a guard in `InterfaceAnalyzer::analyze()` to skip methods missing from storage. `DuplicateClass` is already emitted during scanning, so analysis of the duplicate is pointless anyway.



## Before the fix

Here's the test output from Psalm without this fix:
```text
  E                                                                   1 / 1 (100%)                                                                                                                                        
   
  There was 1 error:                                                                                                                                                                                                      
                                                                                                
  1) Psalm\Tests\IncludeTest::testInterfaceAnalysisDoesNotCrashWhenStorageOverwritten                                                                                                                                     
  UnexpectedValueException: $storage should not be null for Foo\Bar::somemethod
                                                                                                                                                                                                                          
  /…/src/Psalm/Internal/Codebase/Methods.php:1154                                                                                                                                                                         
  /…/src/Psalm/Internal/Analyzer/MethodAnalyzer.php:73     
  /…/src/Psalm/Internal/Analyzer/InterfaceAnalyzer.php:161                                                                                                                                                                
  /…/src/Psalm/Internal/Analyzer/FileAnalyzer.php:192                                                                                                                                                                     
  /…/tests/IncludeTest.php:988                             
                                                                                                                                                                                                                          
  ERRORS!                                                                                       
  Tests: 1, Assertions: 0, Errors: 1.                      
```                                                                                                                                           

The stack trace matches the crash path exactly: `Methods::getStorage` → `MethodAnalyzer::__construct` → `InterfaceAnalyzer::analyze` → `FileAnalyzer::analyze.`


danog:
Thanks!

PR #11782, opened 2026-04-01
Fix repo link

maximal:
https://github.com/danog/psalm-plugin-skeleton does not exist, so using https://github.com/psalm/psalm-plugin-skeleton


danog:
Thanks, my mistake

PR #11789, opened 2026-04-06
[6.x] Skip trait in `getMethodMutations` when storage is missing (fixes "Could not get class storage")

### LINKED ISSUE #8953: Uncaught InvalidArgumentException: Could not get class storage for lexik\bundle\jwtauthenticationbundle\security\authenticator\forwardcompatauthenticatortrait
when running psalm --init

Uncaught InvalidArgumentException: Could not get class storage for lexik\bundle\jwtauthenticationbundle\security\authenticator\forwardcompatauthenticatortrait in phar:///var/www/blackbird/psalm.phar/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:39

psalm I got in dev-master version
"lexik/jwt-authentication-bundle": "2.16.0"

the problematic file I try to ignore. I tried to run no diff etc

alies-dev:
Fixes #8953

Psalm 6.x crashes with `InvalidArgumentException: Could not get class storage for illuminate\console\concerns\promptsformissinginput` when analyzing Laravel projects containing classes that extend `Illuminate\Console\Command`.

**Root cause**: `ClassLikeAnalyzer::getMethodMutations()` iterates AST `TraitUse` nodes unconditionally. A trait may be registered via reflection (present in `existing_traits`) but absent from `ClassLikeStorageProvider::$storage`. The call to `getTraitNode()` → `classlike_storage_provider->get()` throws `InvalidArgumentException`.
**Fix**: Add a 4-line `has()` guard before accessing trait storage, same pattern as the interface fix in 03037f74.
**Real-world trigger**: Any Laravel class extending `Illuminate\Console\Command` (which uses `PromptsForMissingInput` trait) with an uninitialized typed property that triggers `collect_initializations` → `getMethodMutations` → crash.

## Stack Trace

```
InvalidArgumentException: Could not get class storage for illuminate\console\concerns\promptsformissinginput
  at src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:45
  ClassLikes.php:771       → getTraitNode("Illuminate\Console\Concerns\PromptsForMissingInput")
  ClassLikeAnalyzer.php:141 → getTraitNode(...)        ← CRASH SITE
  FileAnalyzer.php:409     → getMethodMutations(...)
  ProjectAnalyzer.php:1205 → getMethodMutations(...)
  FileAnalyzer.php:372     → analyzing ExportMakeCommand.php
```

## Test

```
php vendor/bin/phpunit tests/IncludeTest.php --filter=testGetMethodMutationsDoesNotCrashWhenTraitStorageMissing
```

It failed with:

```
InvalidArgumentException: Could not get class storage for foo\mytrait
  at src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php:137
```

The stack trace matched the expected crash path through `ClassLikeAnalyzer::getMethodMutations()` → `ClassLikeStorageProvider::get()`.

## Notes

Follows the same guard pattern as the interface fix in 03037f74 (#11760)

---

> [!NOTE]
> **Medium Risk**
> Touches core mutation-tracking logic for traits; while the change is a small guard, it can alter analysis behavior by silently skipping unscanned traits and should be validated against real-world projects.
> 
> **Overview**
> Prevents `ClassLikeAnalyzer::getMethodMutations()` from throwing when encountering a `TraitUse` whose `ClassLikeStorage` is missing by adding a `classlike_storage_provider->has()` guard and skipping that trait.
> 
> Adds a regression test in `IncludeTest` that simulates a trait registered via reflection but absent from storage (by removing it from `ClassLikeStorageProvider::$storage`) and asserts analyzing a child class no longer crashes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit deff79deaa0b1d6bc1fd06a43177c40f5b905e75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/ClassLikeAnalyzer.php
@@ -133,6 +133,10 @@ abstract class ClassLikeAnalyzer extends SourceAnalyzer
                         $this->source->getAliases(),
                     );
 
+                    if (!$codebase->classlike_storage_provider->has($fq_trait_name)) {
+                        continue;
+                    }
+
                     $trait_file_analyzer = $project_analyzer->getFileAnalyzerForClassLike($fq_trait_name);
                     $trait_node = $codebase->classlikes->getTraitNode($fq_trait_name);
                     $trait_storage = $codebase->classlike_storage_provider->get($fq_trait_name);
--- a/tests/IncludeTest.php
+++ b/tests/IncludeTest.php
@@ -988,4 +988,89 @@ final class IncludeTest extends TestCase
         // from MethodAnalyzer when the method was missing from storage.
         $file_analyzer->analyze();
     }
+
+    /**
+     * Regression test: ClassLikeAnalyzer::getMethodMutations() must not crash
+     * when a used trait's storage is missing (e.g. a vendor trait that was
+     * registered via reflection but never scanned). The trait should be
+     * skipped gracefully instead of throwing InvalidArgumentException.
+     *
+     * Real-world scenario: Illuminate\Console\Command uses trait
+     * PromptsForMissingInput. The trait exists (registered via reflection)
+     * but was never scanned, so ClassLikeStorageProvider has no entry for it.
+     * When Psalm analyzes a user class extending Command and traces
+     * constructor mutations, ClassLikeAnalyzer::getMethodMutations() iterates
+     * the AST's TraitUse nodes and must handle missing storage gracefully.
+     */
+    public function testGetMethodMutationsDoesNotCrashWhenTraitStorageMissing(): void
+    {
+        $codebase = $this->project_analyzer->getCodebase();
+        $config = $codebase->config;
+
+        $config->throw_exception = false;
+
+        // Two files: parent class uses a trait, child class triggers
+        // getMethodMutations by calling parent::__construct().
+        $parent_path = (string) getcwd() . DIRECTORY_SEPARATOR . 'parent.php';
+        $child_path = (string) getcwd() . DIRECTORY_SEPARATOR . 'child.php';
+
+        // The parent class uses a trait whose storage will be removed.
+        // The trait is intentionally empty — it has no methods, so its
+        // removal won't leave stale declaring_method_ids in class storage.
+        $this->addFile($parent_path, '<?php
+            namespace Foo;
+
+            trait MyTrait {}
+
+            class Base {
+                use MyTrait;
+
+                public function __construct() {}
+            }
+        ');
+
+        // The child class has an uninitialized typed property and calls
+        // parent::__construct(). This triggers collect_initializations,
+        // which calls getMethodMutations on Base to trace mutations.
+        // Base's AST has "use MyTrait", and getMethodMutations will
+        // try to resolve that trait.
+        $this->addFile($child_path, '<?php
+            namespace Foo;
+
+            class Child extends Base {
+                public string $name;
+
+                public function __construct() {
+                    parent::__construct();
+                    $this->name = "test";
+                }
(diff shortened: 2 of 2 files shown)


alies-dev:
UPD: it seems like this PR fixes symptoms of the issue, it's better to find and fix the root cause instead

alies-dev:
## Root cause investigation: this PR treats a symptom

I dug into the original repro path on `SpartnerNL/Laravel-Excel` against `master` and traced exactly why trait storage goes missing. Posting findings before closing this PR.

### What actually happens

1. The trait file `vendor/laravel/framework/src/Illuminate/Console/Concerns/PromptsForMissingInput.php` is **queued, located via composer, and "Deep scanning" is logged** for it. So `FileScanner::scan` does run.
2. Inside `FileScanner::scan`, `getStatementsForFile` returns `[]` with `has_errors=true`. The file is 3,989 bytes of valid PHP (verified with `php -l`), but the parser produces zero top-level statements.
3. Because there are no statements, `ReflectorVisitor::enterNode` is never called for the `Trait_` node, which means `ClassLikeNodeScanner::start` is never reached, which means `ClassLikeStorageProvider::create` is never called for the trait.
4. Storage for `Illuminate\Console\Concerns\PromptsForMissingInput` therefore does not exist after scan or after populate. (Storage probe at end of populate confirms: only the sibling interface `Illuminate\Contracts\Console\PromptsForMissingInput` is present.)
5. Later, analysis of any user class extending `Illuminate\Console\Command` walks `Command`'s `TraitUse` AST nodes in `ClassLikeAnalyzer::getMethodMutations` and crashes on `getTraitNode` / `storage->get()`. That is the crash this PR guards.

### Why parsing returns zero statements

* `laravel-excel/composer.json` declares `"php": "^7.0 || ^8.0"`, so Psalm logs `Target PHP version: 7.0 (inferred from composer.json)`.
* `vendor/laravel/framework/composer.json` requires `"php": "^8.3"`.
* Psalm parses **every** file (including vendored Laravel) against the inferred 7.0 target.
* `PromptsForMissingInput.php` line 44 uses `match (true) { ... }`, which is PHP 8.0 only. PHP-Parser configured for 7.0 rejects it, the error has no column info so it is also not surfaced as a `ParseError` issue (`StatementsProvider.php:424`), and parsing returns `[]`.

### Verification

Setting `phpVersion="8.3"` in laravel-excel's `psalm.xml`, **without applying this PR**, makes the run complete cleanly with normal issue counts. The "missing trait storage" condition simply does not arise once the file actually parses.

### Implication

The crash is the loud symptom of a quieter, broader problem: any vendor file that uses post-target syntax silently disappears from analysis. Classes, interfaces, enums, traits all behave the same way. The trait crash is the most visible failure; the rest produce wrong types, ghost `MissingDependency` chains, and false negatives with no diagnostic the user can act on.

Issue #8953 (lexik/jwt `ForwardCompatAuthenticatorTrait`) is almost certainly the same shape.

### Better fixes than this PR

1. When `parseStatements` returns `has_errors=true`, attach the failure to the file storage. At populate or analyze time, when storage is missing for a class-like that was queued, emit a clear diagnostic that names the file, the parse error, and the inferred PHP target. Today that signal is silently swallowed.
2. When inferring `phpVersion`, walk all of root + dependencies' constraints and take the highest floor, so the analyzer does not run below what the vendored code requires to load.

### Closing

This PR mirrors the precedent in 03037f74 (interfaces) and would prevent a real crash, but it does so by hiding a class of silent failures rather than fixing them. Closing in favour of pursuing the upstream fix. Happy to open a separate issue for "silent vendor parse failures from PHP version target mismatch" if that helps triage.

PR #11849, opened 2026-05-17
Fix DateInterval::createFromDateString return type for PHP 8.3+

xelaris:
On PHP 8.3 and newer `DateInterval::createFromDateString()` throws `DateMalformedIntervalStringException` on failure instead of returning `false` (see php/php-src#10366). Psalm still reports `PossiblyFalseReference` when calling a method on the return value of `createFromDateString`, even when targeting PHP 8.3 and newer.

See also https://www.php.net/manual/en/dateinterval.createfromdatestring.php#refsect1-dateinterval.createfromdatestring-changelog

The return type change was considered in `CallMap_84_delta.php`, but I think it should be in `CallMap_83_delta.php`.

The CallMap lookup seems to be skipped for Psalm's own stubs. I added a `Php83.phpstub` that overrides the `createFromDateString` definition in `stubs/CoreGenericClasses.phpstub` with return type `DateInterval`.

https://github.com/vimeo/psalm/blob/ca151242c84d8962b921bf59c509b49362ce0eec/stubs/CoreGenericClasses.phpstub#L624-L631

There are two new test cases in `CoreStubsTest`. One is confirming the return type is still `DateInterval|false` for PHP 8.2 and the other one is confirming it's `DateInterval` on PHP 8.3.

### DIFF
--- a/dictionaries/CallMap_83.php
+++ b/dictionaries/CallMap_83.php
@@ -13170,7 +13170,7 @@ return array (
   ),
   'dateinterval::createfromdatestring' => 
   array (
-    0 => 'DateInterval|false',
+    0 => 'DateInterval',
     'datetime' => 'string',
   ),
   'dateinterval::format' => 
--- a/dictionaries/override/CallMap_83_delta.php
+++ b/dictionaries/override/CallMap_83_delta.php
@@ -28,6 +28,19 @@ return array (
         'timezone' => 'DateTimeZone|IntlTimeZone|null|string',
       ),
     ),
+    'dateinterval::createfromdatestring' =>
+    array (
+      'old' =>
+      array (
+        0 => 'DateInterval|false',
+        'datetime' => 'string',
+      ),
+      'new' =>
+      array (
+        0 => 'DateInterval',
+        'datetime' => 'string',
+      ),
+    ),
     'gc_status' => 
     array (
       'old' => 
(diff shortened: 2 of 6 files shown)


zoltan-schwarz:
I'm on php8.4, but it is still inferring `false` as possible return type. Tried direct cli argument `--php-version=` also defining in the config file, it's there in the `composer.json`. I even defined a local stub. Still I resolved `false`. Am I missing something.

xelaris:
> I'm on php8.4, but it is still inferring `false` as possible return type. Tried direct cli argument `--php-version=` also defining in the config file, it's there in the `composer.json`. I even defined a local stub. Still I resolved `false`. Am I missing something.

it has been a while since I looked into this, but if I remember correctly, this is because `stubs/CoreGenericClasses.phpstub` defines the return type as `DateInterval|false` and the CallMap lookup is skipped for Psalm's internal stubs.

So, without this PR being merged, even for PHP 8.4 and despite the CallMap, the stub takes effect and Psalm assumes `DateInterval|false`, regardless of the PHP version.

I'm not sure why your local stub doesn't work though.

We are using `psalm-suppress` to suppress the error:

```php
$dateTime = DateTimeImmutable::createFromInterface(/* ... */);
$interval = DateInterval::createFromDateString(/* ... */);

/**
 * @psalm-suppress PossiblyFalseArgument DateInterval::createFromDateString
 * php >= 8.3 returns a DateInterval object or throws but never returns false
 */
return $dateTime->add($interval);
```

PR #11852, opened 2026-05-18
Fix: Stub generation for union types that result to scalar could not be converted to identifier

### LINKED ISSUE #11805: StubsGenerator crashes with UnexpectedValueException when signature return type contains string|int|float|bool (collapsed to TScalar)
### Description

Running Psalm with `--generate-stubs` crashes with an uncaught `UnexpectedValueException` when a method's signature return type is `string|int|float|bool|null`. Psalm's `TypeCombiner` collapses `string|int|float|bool` to the internal `TScalar` pseudo-type, but the stub generator cannot convert `TScalar` back to a valid PHP type node because `scalar` is not a valid PHP type keyword.

### Error

```
Uncaught UnexpectedValueException: scalar could not be converted to an identifier
in .../vendor/vimeo/psalm/src/Psalm/Internal/Stubs/Generator/StubsGenerator.php:329

Stack trace:
#0 .../ClassLikeStubGenerator.php(283): StubsGenerator::getParserTypeFromPsalmType(Object(Psalm\Type\Union))
#1 .../ClassLikeStubGenerator.php(42): ClassLikeStubGenerator::getMethodNodes(Object(Psalm\Storage\ClassLikeStorage))
#2 .../StubsGenerator.php(89): ClassLikeStubGenerator::getClassLikeNode(..., 'InputBag')
#3 .../Psalm.php(1373): StubsGenerator::getAll(...)
#4 .../Psalm.php(421): Psalm::generateStubs(...)
```

### Root Cause

The bug is a round-trip failure in the stub generator:

1. `InputBag::get()` (Symfony `symfony/http-foundation`) has a PHP return type of `string|int|float|bool|null` and a docblock `@return TDefault|TInput` where both templates are bounded by `string|int|float|bool|null`.

2. When Psalm resolves the type union, `TypeCombiner` collapses `string|int|float|bool` into the internal `TScalar` pseudo-type ([TypeCombiner.php:327–338](https://github.com/vimeo/psalm/blob/master/src/Psalm/Internal/Type/TypeCombiner.php)):

```php
if (isset($combination->value_types['string'])
    && isset($combination->value_types['int'])
    && isset($combination->value_types['bool'])
    && isset($combination->value_types['float'])
) {
    unset(...);
    $combination->value_types['scalar'] = new TScalar;
}
```

3. The stub generator stores this collapsed `TScalar|null` as the method's `signature_return_type`.

4. `StubsGenerator::getParserTypeFromPsalmType()` hits the `instanceof Scalar` branch and calls `$atomic_type->toPhpString()` — which returns `null` for `TScalar` because `scalar` is not valid PHP syntax.

5. Line 329 throws the exception:

```php
if ($identifier_string === null) {
    throw new UnexpectedValueException(
        $atomic_type->getId() . ' could not be converted to an identifier'
    );
}
```

### Expected Behavior

The stub generator should handle `TScalar` gracefully — either by expanding it back to `string|int|float|bool` or by skipping the return type declaration (falling back to no return type in the stub).

### Suggested Fix

In `StubsGenerator::getParserTypeFromPsalmType()`, add a special case before the `instanceof Scalar` branch to expand `TScalar` into its constituent types:

```php
if ($atomic_type instanceof TScalar && !$atomic_type instanceof TString && ...) {
    // TScalar has no PHP equivalent; expand to string|int|float|bool
    // or return null/mixed
}
```

Alternatively, `TypeCombiner` could avoid collapsing `string|int|float|bool` to `TScalar` when the types originate from a PHP signature (as opposed to a docblock-inferred type).

### Psalm Version

```
Psalm dev-master@d9c74776b9c98d459dd7f98b12e0ab0daf4af721
```

### Reproducer

Any class that overrides a method and has a PHP return type of `string|int|float|bool|null`, combined with a docblock template return type whose bounds resolve to the same union. Symfony's `InputBag::get()` triggers this reliably:

```php
// symfony/http-foundation InputBag::get()
/** @template TDefault of string|int|float|bool|null */
/** @return TDefault|TInput */
public function get(string $key, mixed $default = null): string|int|float|bool|null
```

Run: `vendor/bin/psalm --generate-stubs=stubs.php`

Mirco91:
Encountered the same issue described in detail [here](https://github.com/vimeo/psalm/issues/11805) by @alies-dev and wanted to fix this issue.

If atomic_type is an object from class TScalar, treat it like TNull

TScalar is an instance of Scalar, though `toPhpString` would always return `null`, thus we would always encounter an UnexpectedValueException.

Fixes: #11805

### DIFF
--- a/src/Psalm/Internal/Stubs/Generator/StubsGenerator.php
+++ b/src/Psalm/Internal/Stubs/Generator/StubsGenerator.php
@@ -18,6 +18,7 @@ use Psalm\Type\Atomic\TLiteralString;
 use Psalm\Type\Atomic\TNamedObject;
 use Psalm\Type\Atomic\TNull;
 use Psalm\Type\Atomic\TObject;
+use Psalm\Type\Atomic\TScalar;
 use Psalm\Type\Atomic\TTrue;
 use Psalm\Type\Atomic\Scalar;
 use PhpParser;
@@ -314,7 +315,7 @@ final class StubsGenerator
         $nullable = $type->isNullable();
 
         foreach ($type->getAtomicTypes() as $atomic_type) {
-            if ($atomic_type instanceof TNull) {
+            if ($atomic_type instanceof TNull || $atomic_type instanceof TScalar) {
                 continue;
             }
 
@@ -345,7 +346,6 @@ final class StubsGenerator
                 if ($nullable) {
                     return new VirtualNullableType($name_node);
                 }
-
                 return $name_node;
             }
         }
--- a/src/Psalm/Type/Atomic/TLiteralFloat.php
+++ b/src/Psalm/Type/Atomic/TLiteralFloat.php
@@ -6,6 +6,8 @@ namespace Psalm\Type\Atomic;
 
 use Override;
 
+use function is_nan;
+
 /**
  * Denotes a floating point value where the exact numeric value is known.
  *


Mirco91:
I'm unsure how to fix the test-with-real-projects tests. I would be grateful, if someone could help me / give me a hint

Issue #11460, opened 2025-05-31
ClassMustBeFinal tries to insert final into strange places.


HenkPoley:
Psalm 6.12.0@cf420941d061a57050b6c468ef2c778faf40aee2

Odd places such as inserting `final` before the initial `<?php`

```diff
diff --git a/src/Application.php b/src/Application.php
--- a/src/Application.php	(revision 1c2b1fbab260d4d4b4c17f9068947f8e9b418d70)
+++ b/src/Application.php	(date 1748664994185)
@@ -1,4 +1,4 @@
-<?php
+final <?php
 
 namespace HenkPoley\DocBlockDoctor;
 
```

Or in the middle of the `use` list.

```diff
diff --git a/src/AstUtils.php b/src/AstUtils.php
--- a/src/AstUtils.php	(revision 1c2b1fbab260d4d4b4c17f9068947f8e9b418d70)
+++ b/src/AstUtils.php	(date 1748664994185)
@@ -3,7 +3,7 @@
 namespace HenkPoley\DocBlockDoctor;
 
 use PhpParser\Node;
-use PhpParser\Node\Expr\Assign;
+final use PhpParser\Node\Expr\Assign;
 use PhpParser\NodeFinder;
 
 /**
```

From this code at revision `1c2b1fb` https://github.com/HenkPoley/PhpDocBlockDoctor/commit/1c2b1fbab260d4d4b4c17f9068947f8e9b418d70

HenkPoley:
This does not appear to emit any 'ERROR:'. So it is not possible to reproduce on the demo page of https://psalm.dev/

HenkPoley:
`--no-cache` **nor** `--clear-cache` / `--clear-global-cache` has any influence.

So it is not #11299.

pl38:
I am seeing similar

```@@ -1,4 +1,4 @@
-<?php
+final <?php
 
 namespace Database\Seeders;
 
@@ -14,69 +14,4 @@
 class UserSeeder extends Seeder
 {
```

I am also using ```"psalm/plugin-laravel": "^3.0",``` if that has any bearing on it.

emmanuelGuiton:
Same here (Psalm 6.13.0) :
```
-use My\App\Config\Config;
+ufinal se My\App\Config\Config;
```
```
-<?php
+final <?php
```

Xerkus:
This issue makes fixer unusable. Patch from linked PR works

tomudding:
[...]

PR #11542, opened 2025-08-20
fix: apply final before the class keyword, obey the suppressions

sawilde:
A couple of issues relating to ClassMustBeFinal
1. Address the incorrect placement of the final keyword - https://github.com/vimeo/psalm/issues/11538, https://github.com/vimeo/psalm/issues/11460
2. Respect suppression when using the --alter switch - https://github.com/vimeo/psalm/issues/11412

I couldn't find any tests that would break when I altered the code, so any advice on where to put tests to help evaluate this fix would be appreciated

PR #11855, opened 2026-05-19
Fix ClassMustBeFinal final insertion for #11460

HenkPoley:
ClassMustBeFinal used a text search over the class statement snippet and treated the relative match offset as a file offset. During --alter this could insert "final" near the start of the file, such as before <?php, before a namespace/use statement, or inside leading text.

The fixes implemented:

- uses an absolute file offset instead of a relative snippet offset
- inserts before readonly class, producing final readonly class
- avoids adding manipulations for suppressed ClassMustBeFinal issues

It also fixes remaining cases not covered by PR #11542:

- preserves attribute placement by inserting after attributes and before class
- avoids duplicate indentation by not preserving indentation for "final "
- avoids matching "class" inside docblocks or other non-declaration text
- there are PHPUnit tests

The insertion point is now derived from the actual class declaration token in the snippet, then translated back to the corresponding file offset before the file manipulation is added. This should fix Issue #11460.

---

This is LLM-assisted code, GPT-5.5 Low in Codex. I did review it, and looks like it does what it should do. I see [`token-get-all()`](https://www.php.net/manual/en/function.token-get-all.php) is also used to (re-)parse PHP in some other gnarly spots. Further fixes of the regex code from the original PR did not seem to be a viable approach.

### DIFF
--- a/src/Psalm/Internal/Codebase/ClassLikes.php
+++ b/src/Psalm/Internal/Codebase/ClassLikes.php
@@ -58,16 +58,25 @@ use function explode;
 use function get_declared_classes;
 use function get_declared_interfaces;
 use function implode;
+use function is_array;
+use function ltrim;
 use function preg_match;
 use function preg_quote;
 use function preg_replace;
+use function str_starts_with;
 use function strlen;
-use function strpos;
 use function strrpos;
 use function strtolower;
 use function substr;
+use function token_get_all;
 
 use const PHP_EOL;
+use const T_CLASS;
+use const T_COMMENT;
+use const T_DOC_COMMENT;
+use const T_READONLY;
+use const T_STRING;
+use const T_WHITESPACE;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


HenkPoley:
Hmm, the CircleCI failure is inside src/Psalm/Type/Atomic/TLiteralFloat.php; which I did not patch.

Am I supposed to base my patch on top of the last non-CircleCI-failing push?

HenkPoley:
Just to show how annoying this bug is with a bog standard Laravel install:

```shell
composer global require laravel/installer
# Just accept the defaults to create a new Laravel project folder
laravel new barebones-laravel
cd barebones-laravel
# Add laravel-psalm to make working with Laravel's magic less annoying
composer config minimum-stability dev && composer config prefer-stable true
composer require --dev psalm/plugin-laravel:^4.*
./vendor/bin/psalm-laravel init
# So we can track what Psalm butchers
git init
git add .
git commit -m 'Initial commit'
# optional
./vendor/bin/psalm-laravel analyze
# Now muck up the PHP code with `psalter` 👇
./vendor/bin/psalter --issues=all
# Search for 'final' by typing /final
git diff
```

Excerpts where `final` is inserted incorrectly, without any patching:

```diff
diff --git a/app/Models/User.php b/app/Models/User.php
index 68f3a66..cfc4480 100644
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1,4 +1,4 @@
-<?php
+final <?php
 
 namespace App\Models;
 
diff --git a/app/Providers/AppServiceProvider.php b/app/Providers/AppServiceProvider.php
index 452e6b6..925a894 100644
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -1,4 +1,4 @@
-<?php
+final <?php
 
 namespace App\Providers;
 
diff --git a/database/factories/UserFactory.php b/database/factories/UserFactory.php
index c4ceb07..a89da1b 100644
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Database\Factories;
+namespace Database\Factoriefinal s;
 
 use App\Models\User;
 use Illuminate\Database\Eloquent\Factories\Factory;
diff --git a/database/seeders/DatabaseSeeder.php b/database/seeders/DatabaseSeeder.php
index 6b901f8..3af438d 100644
--- a/database/seeders/DatabaseSeeder.php
+++ b/database/seeders/DatabaseSeeder.php
@@ -1,4 +1,4 @@
-<?php
+final <?php
 
 namespace Database\Seeders;

```

HenkPoley:
I verified that my PR still places the `final` keyword in the correct place when using Psalm 7.0.0-beta19@7e751c06a756fa64dc4c759c09fe4a173afcb43; also known as Psalm 7.0.0-beta19.

PR #11856, opened 2026-05-19
Fix for coding standard issue in commit ca15124

HenkPoley:
Current PHPCS/Slevomat requires is_nan() to be imported instead of referenced through the fallback global namespace.


HenkPoley:
Looks like CircleCI says https://github.com/psalm/endtoend-test-phpunit needs its baselines updated.

PR #11859, opened 2026-05-22
Fix coalesce with function call array key

### LINKED ISSUE #11857: InvalidReturnStatement / InvalidReturnType false positive when array key is a function call
In the code below, `foo` cannot return `null`, but Psalm still raises [`NullableReturnStatement`](https://psalm.dev/139) and [`InvalidNullableReturnType`](https://psalm.dev/144)

```php
/** @param array<string, ?string> $foo */
function foo(array $foo, string $key, string $default): string {
    return $foo[bar($key)] ?? $default;
}

function bar(string $key): string {
    return "aa$key";
}
```

https://psalm.dev/r/f8585fb593

Possibly a common root cause with #8213 and/or #10233

HenkPoley:
Fixes #11857

ExpressionIdentifier::getExtendedVarId() is used to find out whether Psalm can give the entire left-hand array lookup a stable internal id. If it can, normal isset() reconciliation is enough. If it cannot, as with bar($key), then the id === null, and we create a temporary variable so Psalm can narrow that temporary instead.

It does not try to prove that `bar($key)`  is stable, pure, or repeatable.

### DIFF
--- a/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/CoalesceAnalyzer.php
+++ b/src/Psalm/Internal/Analyzer/Statements/Expression/BinaryOp/CoalesceAnalyzer.php
@@ -6,6 +6,7 @@ namespace Psalm\Internal\Analyzer\Statements\Expression\BinaryOp;
 
 use PhpParser;
 use Psalm\Context;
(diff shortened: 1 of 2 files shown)


HenkPoley:
Hmm, looks like this is not a complete fix.

PR #11861, opened 2026-05-24
Allow sebastian/diff 9

liviuconcioiu:


niconoe-:
Thanks for that PR.

Would some maintainers consider releasing a new patch version of Psalm with this? It prenvents to use both Psalm and PHPUnit 13.2 (which requires sebastian/diff ^9.0) at the same time :cry: 

Thanks a lot!

LordSimal:
will this be upwards merged to the 7.x beta as well or does this require its own PR?

djairhogeuens:
@danog Any chance this could be included in a new 6.x release please?

PR #11863, opened 2026-05-28
fix: pin GitHub Actions to SHA for supply chain security

riccardosarro:
## Summary

Pin all GitHub Actions to full commit SHAs for supply chain security.

Actions referenced by tag or branch have been resolved to their commit SHA, with the original ref preserved as an inline comment. Where a sub-action had unpinned transitive dependencies, the action was upgraded to the closest newer version where all sub-actions are fully pinned.

### DIFF
--- a/.github/workflows/bcc.yml
+++ b/.github/workflows/bcc.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Set up PHP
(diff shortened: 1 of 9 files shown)


security-claw:
:white_check_mark: pure-pin, merged — all 7 actions pinned to SHA at same version (no bumps). `Code Style Analysis` failure is pre-existing on `6.x` base branch (run #25955569666).

PR #11891, opened 2026-06-23
[6.x] Allow `#[\Override]` on properties (PHP 8.5)

### LINKED ISSUE #11890: False positive InvalidAttribute: `#[\Override]` on properties is valid in PHP 8.5
### Summary

On PHP 8.5, `#[\Override]` is allowed on **properties** ([RFC: Add `#[\Override]` to properties](https://wiki.php.net/rfc/marking_overriden_properties)). Psalm reports a false positive `InvalidAttribute` for that valid usage.

### Reproduction

```php
<?php

class A
{
    protected int $x = 1;
}

class B extends A
{
    #[\Override]
    protected int $x = 2;
}
```

### Expected

No error. PHP 8.5 accepts this:

```
$ php -r 'class A { protected int $x = 1; } class B extends A { #[\Override] protected int $x = 2; } echo "accepted\n";'
accepted
```

### Actual

```
ERROR: InvalidAttribute - Attribute \Override cannot be used on a property (see https://psalm.dev/242)
```

### Root cause

The bundled stub `stubs/CoreGenericAttributes.phpstub` declares `Override` with `TARGET_METHOD` only:

```php
/** @psalm-immutable */
#[Attribute(Attribute::TARGET_METHOD)]
final class Override
{
    public function __construct() {}
}
```

As of PHP 8.5 the real target set is `method | property`. Confirmed against the engine (class constants are **not** a valid target, so they must be excluded):

```
$ php -r 'class A { const X = 1; } class B extends A { #[\Override] const X = 2; }'
PHP Fatal error: Attribute "Override" cannot target class constant (allowed targets: method, property)
```

### Suggested fix

```diff
-#[Attribute(Attribute::TARGET_METHOD)]
+#[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_PROPERTY)]
 final class Override
```

Ideally guarded so `TARGET_PROPERTY` is only treated as valid under PHP 8.5+ target analysis.

### Environment

- Psalm 7.0.0-beta19
- PHP 8.5.7
- Stub unchanged on `master` / `5.x` at time of filing.

alies-dev:
PHP 8.5 lets you put `#[\Override]` on a property, but Psalm reported `InvalidAttribute` for it. The bundled stub declared the attribute as `TARGET_METHOD` only, so `AttributesAnalyzer` rejected it anywhere other than a method. Promoted constructor properties hit the same false positive.

Adding `TARGET_PROPERTY` to the stub fixes it. Class constants stay excluded, since the engine fatals there ("allowed targets: method, property").

One caveat: the stub loads for every `phpVersion`, not just 8.5 and above, so the property target now applies no matter what you target. A per-version attribute target set can't be expressed through stubs, because version stubs are cumulative and redeclaring the class only appends a second `#[Attribute]` that `getAttributeClassFlags` never reads. Psalm is already this loose about attribute availability elsewhere (`Deprecated`, for one).

Tests cover a plain property, a promoted property, and the method case that has to keep working. A negative case keeps `#[\Override]` rejected on class constants.

Fixes #11890

### DIFF
--- a/stubs/CoreGenericAttributes.phpstub
+++ b/stubs/CoreGenericAttributes.phpstub
@@ -8,7 +8,7 @@ final class AllowDynamicProperties
 }
 
 /** @psalm-immutable */
-#[Attribute(Attribute::TARGET_METHOD)]
+#[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_PROPERTY)]
 final class Override
 {
     public function __construct() {}
--- a/tests/AttributeTest.php
+++ b/tests/AttributeTest.php
@@ -43,6 +43,32 @@ final class AttributeTest extends TestCase
     public function providerValidCodeParse(): iterable
     {
         return [
+            'overrideAttributeOnOverriddenMethod' => [
+                'code' => '<?php
+                    class A
+                    {
+                        public function f(): void {}
+                    }
+
+                    class B extends A
+                    {
+                        #[\Override]
+                        public function f(): void {}
+                    }',
+            ],
+            'overrideAttributeOnOverriddenProperty' => [
+                'code' => '<?php
+                    class A
+                    {
+                        protected int $x = 1;
+                    }
+
+                    class B extends A
+                    {
+                        #[\Override]
+                        protected int $x = 2;
+                    }',
+            ],
             'classAndPropertyAttributesExists' => [
                 'code' => '<?php
                     namespace Foo;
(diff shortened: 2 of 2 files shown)


alies-dev:
UPD: I'll create another PR, stacked on that, that will have a proper reporting and auto-fix features. The currect one is a hotfix no-brainer one aims to quick-patch that noisy issue

Issue #8251, opened 2022-07-13
Badly placed parenthesis in conditional return causes crash


gsteel:
I hope this issue isn't just noise 🤞

In short, if you mess up where you close parenthesis for a conditional return, psalm crashes rather than yield an `InvalidDockblock` issue:

https://psalm.dev/r/cea33e07ae

AndrolGenhald:
Turns out I ran into this myself! Duplicate of #7575, I'll try to look at that in the next week.

PR #11914, opened 2026-08-06
Throw TypeParseTreeException instead of crashing on a bare TemplateIsTree

### LINKED ISSUE #7575: Crash on incorrectly formatted conditional return
I couldn't remember where the parentheses go, and it crashed! https://psalm.dev/r/3933e35e36

### LINKED ISSUE #10827: Psalm crash with invalid annotation
reproducer: https://psalm.dev/r/9889237f4f

the error:

> Uncaught InvalidArgumentException: Unrecognised parse tree type Psalm\Internal\Type\ParseTree\TemplateIsTree in phar:///var/www/vendor/psalm/phar/psalm.phar/src/Psalm/Internal/Type/TypeParser.php:228

Eljees:
Fixes #10827.
Fixes #7575.
(#8251 was closed as a duplicate of #7575 and is fixed by the same change.)

### Problem

A docblock with misplaced parentheses in a conditional return type makes Psalm crash with an internal error instead of reporting `InvalidDocblock`:

```php
<?php
/** @return ($n is int) ? int : float */
function double(int|float $n): int|float { return $n * 2; }
```

```
Uncaught InvalidArgumentException: Unrecognised parse tree type
Psalm\Internal\Type\ParseTree\TemplateIsTree in src/Psalm/Internal/Type/TypeParser.php:402
```

The annotation is indeed wrong — the correct form is `($n is int ? int : float)` — but a typo in a docblock should be a diagnostic, not a crash.

### Why it happens

`ParseTreeCreator::handleQuestionMark()` only folds a `TemplateIsTree` into a `ConditionalTree` when the `?` arrives while the current leaf *is* that `TemplateIsTree`. With `($n is int) ? …` the closing parenthesis terminates the `EncapsulationTree` first, so the `?` takes the `NullableTree` branch and the `TemplateIsTree` is left bare.

The docblock scanner then truncates the return type at the balanced expression — the string that actually reaches the parser is `($n is int)`, without any branches — and `TypeParser::getTypeFromTree()` has no case for a bare `TemplateIsTree`, so it falls through to the catch-all `throw new InvalidArgumentException(...)`.

That matters because the caller only expects `TypeParseTreeException`: `FunctionLikeDocblockScanner::handleReturn()` wraps the parse in `try` and catches `TypeParseTreeException` to emit `InvalidDocblock`. An `InvalidArgumentException` escapes that catch and takes the whole run down.

### Fix

Throw `TypeParseTreeException` for a bare `TemplateIsTree`, the same way the neighbouring malformed-conditional cases in this file already do (`'Unrecognized template'`, `'Invalid conditional'`, `'Misplaced brackets'`). The existing `catch` then turns it into an ordinary `InvalidDocblock`:

```
ERROR: InvalidDocblock - src/a.php:3:1 - Invalid conditional, expecting the form (T is X ? Y : Z)
       in docblock for double
```

### Tests

* `tests/TypeParseTest.php` — two cases covering the shapes the docblock scanner actually produces: `(T is string)` and `(T is string) ? string`. Both throw `InvalidArgumentException` before the change and pass after it.
* `tests/AnnotationTest.php::providerInvalidCodeParse` — the reporters' snippets from #10827 and #7575, expecting `InvalidDocblock`.

Note that `(T is string) ? string : int` was *not* affected: that string never reaches the parser intact, and in isolation it already raises `TypeParseTreeException: Saw : outside of object-like array`.

### Verified locally (PHP 8.5.8, on `6.x`)

| gate | result |
|---|---|
| new unit tests without the fix | 2 failures, both `InvalidArgumentException: Unrecognised parse tree type …TemplateIsTree` |
| new unit tests with the fix | OK (2 tests, 2 assertions) |
| new annotation cases without / with the fix | 2 failures → OK (2 tests, 4 assertions) |
| `tests/TypeParseTest.php` in full | OK (176 tests, 186 assertions) |
| `tests/AnnotationTest.php` in full | OK (155 tests, 222 assertions, 2 skipped — also skipped on a clean tree) |
| `tests/TypeCombinationTest.php` | OK (122 tests, 234 assertions) |
| `composer lint`, `composer cs`, `composer psalm` on the changed files | clean |
| CLI reproducers for #10827 and #7575 | `Uncaught InvalidArgumentException` → `InvalidDocblock` |

The full `composer phpunit` (paratest, entire suite) was not run — only the files above.

*AI-assisted: I used Claude to help trace the parse-tree path and to draft the guard and the tests. I reviewed every line, ran the new tests against an unpatched `6.x` first to confirm they fail without the change, and ran the gates in the table locally; the analysis and the runs are mine.*

### DIFF
--- a/src/Psalm/Internal/Type/TypeParser.php
+++ b/src/Psalm/Internal/Type/TypeParser.php
@@ -399,7 +399,7 @@ final class TypeParser
         }
 
         if (!$parse_tree instanceof Value) {
-            throw new InvalidArgumentException('Unrecognised parse tree type ' . $parse_tree::class);
+            throw new TypeParseTreeException('Unrecognised parse tree type ' . $parse_tree::class);
         }
 
         if ($parse_tree->value[0] === '"' || $parse_tree->value[0] === '\'') {
--- a/tests/AnnotationTest.php
+++ b/tests/AnnotationTest.php
@@ -2047,6 +2047,17 @@ final class AnnotationTest extends TestCase
                 ",
                 'error_message' => 'InvalidDocblock',
             ],
+            'parenthesizedConditionalCondition' => [
+                'code' => '<?php
+                    /** @return ($n is int) ? int : float */
+                    function double(int|float $n): int|float {
+                        return $n * 2;
+                    }
+                ',
+                'error_message' => 'InvalidDocblock',
+                'error_levels' => [],
+                'php_version' => '8.0',
+            ],
             'promotedPropertiesDocumentationFailsWhenSendingBadTypeAgainstParam' => [
                 'code' => '<?php
                     final class UserRole


Eljees:
A note on the red `ci/circleci: test-with-real-projects`, since it is not caused by this PR.

That job is failing on `6.x` itself. The four most recent commits on the branch that ran it are all red, going back to #11874 on 10 June:

| commit | date | test-with-real-projects |
|---|---|---|
| `7385c40` | 23 Jun | failure |
| `0090471` | 23 Jun | failure |
| `be7afcf` | 11 Jun | failure |
| `778320a` | 10 Jun | failure |

Every other check on this PR is green, and the change itself is two lines in `TypeParser.php` plus tests. Let me know if you would like me to rebase once the branch job is healthy again.

Eljees:
Rebased onto current `6.x` (`82c6021`). All three CircleCI jobs are green on the new head `af1626f`, including `test-with-real-projects`, which was the red one before.

That earlier failure was not about this change: it came from the run of 6 August, where the job stopped after 20 seconds inside `bin/ci/test-with-real-projects.sh phpunit`, before Psalm was ever invoked. Nothing in the diff moved in the meantime — it is the same one-line change plus its test, byte for byte — so the difference is the base it was tested against.

The change itself is unchanged: `getTypeFromTree()` throws `TypeParseTreeException` instead of `InvalidArgumentException` when the parse tree is not a `Value`, so a docblock like `@return ($n is int) ? int : float` written with parentheses around the condition is reported as an `InvalidDocblock` rather than crashing the analysis. The test in `tests/AnnotationTest.php` covers that form.

PR #11942, opened 2026-09-14
Resolve taint graph to a fixed point with sink-reachability pruning

danog:
## Summary

Taint resolution previously ran a fixed **40-round** BFS and then stopped at a hard-coded nesting limit (`$depth = 40`). This had two problems:

- Taint flows longer than 40 hops were **silently missed** (false negatives).
- The 40-round budget was unrelated to whether resolution had actually converged.

This PR reworks resolution into a monotone worklist that runs to a **true fixed point**, so the depth limit is no longer needed, and adds pruning so that running to convergence stays cheap.

## Changes

- **Run to convergence** instead of a fixed 40 rounds. Deep taint flows are no longer missed. Termination is guaranteed by the existing `(id, taints)` visited guard, whose state space is finite.
- **Monotone taint-accumulation guard.** The taint transfer functions `(taints | added) & ~removed` are monotone, so once an unspecialized node has propagated a set of taints, re-processing it with a subset reaches no new sink. Skipping those bounds per-node work by the number of taint kinds rather than `2^kinds`. It is applied only to unspecialized nodes, since specialized routing depends on the specialization context, not just the node id.
- **Sink-reachability pruning.** A backward reachability pass computes which nodes can reach a sink; sources that cannot are dropped, and propagation into nodes that cannot is skipped. On real codebases the full graph is large but this relevant sub-graph is small. Specialized and unspecialized nodes are linked both by their authoritative `unspecialized_id` field and by the id-string ` specialized in ` separator, so the prune never drops a live node.

## Correctness

- All existing taint tests pass, including the exact issue-trace assertions.
- Adds a regression test (`testTaintFlowDeeperThanLegacyResolutionDepth`) exercising a taint flow deeper than the old 40-hop cap; it fails if a 40-round cap is reintroduced and passes now.
- Pruning is sound in both directions: over-linking the reverse graph only prunes *less*, and the two linking signals (field + id-string) are complementary, each independently sufficient on the current test suite.

Note: this is a behaviour change — taint flows longer than 40 hops that were previously dropped will now be reported.


danog:
@cursor review

PR #11945, opened 2026-09-14
Taint php://input & STDIN sources, and propagate taint through closures

danog:
## Summary

Two related taint-analysis improvements (master / bitmask design).

### 1. Reading from `php://input` / `php://stdin` is a taint source

- `fopen()`, `file_get_contents()` and `file()` called with a literal `php://input` or `php://stdin` path now return a taint source.
- Stream reading functions (`stream_get_contents`, `fgets`, `fread`, `fgetcsv`, `fscanf`, ...) propagate the taint of their stream-resource argument, so data read from a tainted `fopen()` handle stays tainted.
- The predefined `STDIN` constant (the `php://stdin` stream) is itself a taint source, so e.g. `fgets(STDIN)` / `stream_get_contents(STDIN)` are tainted.

### 2. Taint sources propagate through closures / first-class callables

Invoking a callable value dropped the underlying function's taint-source behavior, because the call target is an expression rather than a `Node\Name`. `TClosure`/`TCallable` now remember the underlying function id (metadata only — it does not affect the structural type, comparison, or `getId()`), so on invocation the source handling is re-dispatched. Argument sinks already propagated via the callable's params.

This makes the following correctly report a tainted read:

```php
$f = file_get_contents(...);
echo $f('php://input'); // TaintedHtml
```

## Test plan

Added `TaintTest` cases covering each source and the first-class-callable flow. CI runs the full suite.


danog:
bugbot run

danog:
bugbot run

danog:
bugbot run

danog:
bugbot review

danog:
bugbot review

PR #11947, opened 2026-09-14
Use upstreamed amphp/parallel ForkContext

danog:
## Summary

`amphp/parallel` 2.4.0 ships the fork-based `Context` implementation that previously lived in-tree as `Psalm\Internal\Fork\ForkContext`. This bumps the constraint to `^2.4.0`, removes the local copy, and wires Psalm up to the upstream `Amp\Parallel\Context\ForkContext`.

Two Psalm-specific behaviours the in-tree context provided are preserved so this is behaviour-neutral:

- **igbinary serialization** — upstream `ForkContext::start()` now defaults to `NativeSerializer` but accepts a `Serializer` argument, so the pool passes `IgbinarySerializer` when the extension is loaded (matching the previous auto-selection).
- **Worker-death diagnostics** — the old context turned fatal signals into Psalm-specific guidance (SIGSEGV → "THIS IS A PHP BUG, report to vimeo/psalm & php-src", SIGKILL → "likely OOM killer, increase swap or set `arrayCache=false`"). Upstream just reports `Worker exited due to signal N` with the signal in the exception code. A thin `SignalDiagnosticContext` decorator re-applies the original messages by inspecting that code, so users see the same hints as before.

## Changes

- `composer.json`: `amphp/parallel` `^2.3` → `^2.4.0`
- Removed `src/Psalm/Internal/Fork/ForkContext.php`
- Added `src/Psalm/Internal/Fork/SignalDiagnosticContext.php` (diagnostics decorator)
- `Pool.php`: start the upstream `ForkContext` (swapped arg order, explicit serializer) and wrap the result in the decorator
- `PreloaderList.php`: preload the upstream `ForkContext` and the new decorator instead of the removed class

## Testing

- `php -l` on all changed files
- Psalm self-analysis on the changed Fork files: no errors
- Ran Psalm with `--threads=2` to exercise the fork pool end-to-end (fork + igbinary serialization + cross-process result gathering all work)


danog:
bugbot review

danog:
bugbot review

PR #11948, opened 2026-09-14
Fix multiprocess non-determinism in taint/data-flow node locations

danog:
When analysis runs across forked workers, each worker builds a partial DataFlowGraph that the parent merges with `+=` (first value seen per key wins). The same DataFlowNode id could be created at several sites with different `code_location`s (and different sink taints), so which value survived the merge depended on which work-stolen file a worker happened to process first. Because a taint issue is reported at its node's `code_location`, this shifted the reported location -- and the per-file / snippet occurrence counts -- of taint findings between otherwise identical runs, so baselined issues silently reappeared and `--set-baseline` never converged.

Make a node's location and sink taints a pure function of its identity:

* getForCallableArg()/getForCallableReturn() no longer accept an independent CodeLocation (breaking change). A callable node stands for a builtin/magic/callable-object/dynamic call and has no storage to derive a canonical location from, so its location is now always its specialization (the callsite) -- which is already part of the node id. It is therefore no longer possible to give the same (unspecialized) shared id a different callsite location in different processes.

* getForMethodArgument() derives its sink taints from the parameter's storage rather than from a caller-supplied argument (breaking change), so the same argument node always carries the same taints no matter where it is created.

* Sites that built nodes for inherited methods without resolving storage (MethodCallReturnTypeFetcher's inherited-method returns, and ArgumentAnalyzer's argument sinks and dependent-classlikes loop) now resolve the declaring method's storage, so the location is the canonical parameter/return location wherever the id is produced.

Verified on a ~9k-file project: `--set-baseline` followed by repeated analysis now reports zero new issues across many consecutive runs (previously every rerun leaked a different set of taint findings). psalm's own TaintTest suite passes unchanged.

Claude-Session: https://claude.ai/code/session_01YYjYCCjPFamLerruby1Vko


danog:
bugbot run

danog:
bugbot run

danog:
bugbot review

danog:
Superseded by a 4-PR atomic stack that splits this change into reviewable, independently-mergeable pieces:

- #11951 — Resolve the taint graph to a fixed point with sink-reachability pruning
- #11952 — Derive data-flow node locations & sink taints from storage (deterministic findings)
- #11953 — Make DataFlowNode locations deterministic by construction (private constructor)
- #11954 — Key taint argument nodes by declared parameter index (fix named-argument sinks)

Please review the stack in order (11951 → 11954).

PR #11955, opened 2026-09-15
Detect unused free functions; simplify dead-code roots

danog:
## Summary

Adds dead-code detection for **free (non-method) functions** and simplifies how the code-use graph decides reachability roots.

### New `UnusedFunction` issue
- Reported under `--find-dead-code` when a free function is never referenced.
- Function references are now recorded in the code-use graph at call sites (`FunctionCallAnalyzer`), so calls — including first-class callables `foo(...)` and callable-strings like `array_map('foo', …)` — keep a function alive.
- Free functions are no longer unconditional roots, so genuinely dead functions are reported.
- `@api` free functions are treated as public-API entry points (kept alive, and they keep what they reference alive).
- Closures/arrow functions are intentionally **not** reported (they have no representation in the reachability graph; their body references are attributed to the enclosing scope).

### Simplified reachability roots
- **Stop specially treating external `@api` interfaces/classes as roots.** `@api` is handled uniformly by the public-API marking that already follows from scanning `@api` docblocks, so a class that is only reachable by extending a vendored type must itself be marked `@api` to count as an entry point. This reverts the external-interface override special-case; the escape hatch for a plugin/framework entry point is to mark **your own** class/function `@api`.
- `CodeUseGraph` now receives the classlike storage provider via its constructor, and the root predicate lives in an overridable `isRoot()` method.

### Message tweaks
- `UnusedClass`, `PossiblyUnusedMethod`, `PossiblyUnusedProperty` and `UnusedFunction` messages now note that adding `@api` is the fix **only** for code genuinely used from outside this codebase (controller entry points, library APIs covered by a BC promise), otherwise it is dead code and should be removed.

## Tests
- New `UnusedCodeTest` cases: dead free function reported; called function / first-class-callable / callable-string kept alive; `@api` function not reported; `@api` class implementing an external interface kept alive while a non-`@api` one is not.
- New `CodeUseGraphTest` case asserting a free-function node is not a root by itself.
- Docs (`UnusedFunction.md`, `issues.md`) and `config.xsd` handler added.


danog:
Addressed the Bugbot findings:

- **First-class callable skips reference** — real bug, fixed in 112a1fd7b. `FunctionCallAnalyzer` returned early for `foo(...)` before recording the reference, so a function used only via a first-class callable was wrongly reported. The `addReferenceToFunctionLike` call now runs right after the function id is resolved, before the first-class-callable early return. Added a regression test (`testFunctionReferencedByFirstClassCallableIsUsed`).

- **String callables not marked used** — false positive. String/array callables (`array_map('fn', …)`) are already recorded through the callable-argument analysis, so such a function is not reported as unused. This is covered by `testFunctionReferencedByCallableStringIsUsed`, which passes.

danog:
bugbot run

danog:
bugbot run

PR #11960, opened 2026-09-15
[6.x] Report InvalidDocblock instead of crashing on unparseable `@self-out` / `@if-this-is` types

### LINKED ISSUE #11958: Uncaught TypeParseTreeException: an unparseable `@self-out` / `@this-out` type crashes the scan phase
`@self-out` / `@this-out` is the only docblock tag in Psalm whose type is parsed without a `try { } catch (TypeParseTreeException $e)`. Any type Psalm cannot parse there escapes as an uncaught throwable during the scan phase, killing the run before a single issue is reported. It cannot be suppressed, and `<ignoreFiles>` does not help because that skips analysis, not scanning.

`laravel/framework` v13.32.0 shipped the first widely installed docblock that hits this, so every Psalm 6 and Psalm 7 run against a project resolving laravel/framework >= 13.32 now dies outright. This holds even for projects whose own files never reference Laravel: composer's autoload `files` entries (Laravel's helpers) pull `Collection.php` into the scan phase.

### Reproducer

`repro.php`:

```php

final class A
{
    // 1. parameter-conditional type
    /** @phpstan-this-out ($key is null ? static : $this) */
    public function s(?int $key): void {}

    // 2. plain unparseable type — so this is not only about conditionals
    /** @psalm-self-out garbage<<< */
    public function t(): void {}
}
```

`psalm.xml`:

```xml
<?xml version="1.0"?>
<psalm errorLevel="1" xmlns="https://getpsalm.org/schema/config">
    <projectFiles>
        <directory name="."/>
    </projectFiles>
</psalm>
```

https://psalm.dev/r/80ce3d4b4a

### Actual

```
Uncaught Psalm\Exception\TypeParseTreeException: Unrecognized template '$key' in .../src/Psalm/Internal/Type/TypeParser.php:342
Stack trace:
#0 .../src/Psalm/Internal/Type/TypeParser.php(281): Psalm\Internal\Type\TypeParser::getTypeFromTree()
#1 .../src/Psalm/Internal/Type/TypeParser.php(158): Psalm\Internal\Type\TypeParser::getTypeFromTree()
#2 .../src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockScanner.php(319): Psalm\Internal\Type\TypeParser::parseTokens()
#3 .../src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeNodeScanner.php(495): Psalm\Internal\PhpVisitor\Reflector\FunctionLikeDocblockScanner::addDocblockInfo()
...
(Psalm 7.0.0-beta21@5df32a9 crashed due to an uncaught Throwable)
```

Case 2 crashes the same way, with `Unexpected token <` from `ParseTreeCreator`.

### Expected

An `InvalidDocblock` issue, which is what every other tag in the same file already produces:

```php
/** @return ($nope is null ? int : string) */  // InvalidDocblock: Unrecognized template '$nope' in docblock for A::r
/** @return garbage<<< */                       // InvalidDocblock: Unexpected token < in docblock for A::u
/** @param-out garbage<<< $x */                 // InvalidDocblock: Badly-formatted @param in docblock for A::v
```

Note that `@return` accepts the parameter-conditional form, because it routes the type through `getConditionalSanitizedTypeTokens()`, which rewrites `$param` tokens into generated `TGeneratedFromParamN` template params. The `self_out` path parses the raw tokens, so `ConditionalTree` looks `$key` up in the template map, does not find it, and throws.

### Where

`src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockScanner::addDocblockInfo()`. Of the eight `TypeParser::parseTokens()` call sites in that file, the `self_out` and `if_this_is` blocks are the only two with no surrounding `catch (TypeParseTreeException $e)`:

```
$ grep -n "TypeParser::parseTokens\|catch (TypeParseTreeException" src/Psalm/Internal/PhpVisitor/Reflector/FunctionLikeDocblockScanner.php
257:                $storage->global_types[$global['name']] = TypeParser::parseTokens(
266:            } catch (TypeParseTreeException $e) {
319:            $out_type = TypeParser::parseTokens(     <-- self_out, unguarded
336:            $out_type = TypeParser::parseTokens(     <-- if_this_is, unguarded
614:            $namespaced_type = TypeParser::parseTokens(
628:        } catch (TypeParseTreeException $e) {
793:                $new_param_type = TypeParser::parseTokens(
806:            } catch (TypeParseTreeException $e) {
983:            $storage->return_type = TypeParser::parseTokens(
1039:        } catch (TypeParseTreeException $e) {
1180:            $removed_taint = TypeParser::parseTokens(
1197:        } catch (TypeParseTreeException $e) {
1387:            $out_type = TypeParser::parseTokens(
1398:        } catch (TypeParseTreeException $e) {
1451:                        $template_type = TypeParser::parseTokens(
1462:                    } catch (TypeParseTreeException $e) {
```

(Line numbers from 7.0.0-beta21; the structure is the same on 6.x.)

Making the parameter-conditional form actually work looks like a separate, larger change. Degrading to `InvalidDocblock` would already turn an unrecoverable crash into a suppressible issue.

### Versions

- Reproduced on 6.17.2, 7.0.0-beta20 and 7.0.0-beta21. Not a regression in any of them.
- All three tag spellings crash: `@self-out`, `@psalm-self-out`, `@phpstan-this-out`.
- PHP 8.4 and 8.5.

### Real-world trigger

[laravel/framework v13.32.0, `Collection.php:1061`](https://github.com/laravel/framework/blob/v13.32.0/src/Illuminate/Collections/Collection.php#L1061), added in laravel/framework#61571:

```php
 * @phpstan-this-out ($key is null ? static<TKey|int, TValue> : $this)
```

`$key` is a declared parameter of `prepend()`, so the docblock is well formed and PHPStan accepts it. Pinning `laravel/framework` below v13.32.0 is the only workaround I found.

Possibly related: #10126 (conditional `@param-out` silently ignored).

alies-dev:
Fixes #11958

### Problem

In `FunctionLikeDocblockScanner::addDocblockInfo()`, the `self_out` and `if_this_is` blocks were the only two of the eight `TypeParser::parseTokens()` call sites in that file with no surrounding `catch (TypeParseTreeException $e)`. Any type Psalm cannot parse in `@self-out` / `@psalm-self-out` / `@psalm-this-out` / `@phpstan-this-out` / `@psalm-if-this-is` therefore escaped as an uncaught throwable during the **scan** phase, killing the run before a single issue was reported. It could not be suppressed, and `<ignoreFiles>` did not help because that skips analysis, not scanning.

`laravel/framework` v13.32.0 shipped `@phpstan-this-out ($key is null ? static<TKey|int, TValue> : $this)` in `Collection.php` https://github.com/laravel/framework/blob/cdd8b33c246719acdd118c705ce8c7ab5ef48a96/src/Illuminate/Collections/Collection.php#L1061, so every Psalm 6/7 run against a project resolving that version died outright — even for projects whose own files never reference Laravel, since composer's autoload `files` entries pull `Collection.php` into the scan phase.

### Change

Both blocks now wrap the parse in `catch (TypeParseTreeException $e)` and append an `InvalidDocblock` to `$storage->docblock_issues[]`, using the exact same form as the six already-guarded sites in the file. On failure `self_out_type` / `if_this_is_type` stay `null`, so downstream analysis behaves as if the tag were absent. The parse result is assigned straight to the storage property, dropping the now-pointless intermediate local.

Actually *supporting* the parameter-conditional form (routing through `getConditionalSanitizedTypeTokens()` and resolving the generated `TGeneratedFromParamN` templates at every callsite applying `self_out_type`) is a separate, larger change and is deliberately not attempted here. This turns an unrecoverable crash into a suppressible issue, matching every other tag in the same file.

### Before / after

```php
final class A
{
    /** @phpstan-this-out ($key is null ? static : $this) */
    public function s(?int $key): void {}

    /** @psalm-self-out garbage<<< */
    public function t(): void {}
}
```

Before: `Uncaught Psalm\Exception\TypeParseTreeException: Unrecognized template '$key'` at `FunctionLikeDocblockScanner.php:321`.

After:

```
ERROR: InvalidDocblock - repro.php:7:5 - Unrecognized template '$key' in docblock for A::s
ERROR: InvalidDocblock - repro.php:11:5 - Unexpected token < in docblock for A::t
```

`<InvalidDocblock errorLevel="suppress"/>` silences both.

### Tests

- `tests/ThisOutTest.php`: gains `InvalidCodeAnalysisTestTrait` + `providerInvalidCodeParse` with an unparseable `@psalm-self-out` case and the laravel-shaped `@phpstan-this-out ($key is null ? static : $this)` case.
- `tests/IfThisIsTest.php`: gains an unparseable `@psalm-if-this-is` case.

Both new files' tests fail (as fatals) without the src change and pass with it. `ThisOutTest` 5/5, `IfThisIsTest` 11/11, `Template/FunctionTemplateTest` 121/121. Full suite locally shows only pre-existing `InternalCallMapHandlerTest` reflection failures on PHP 8.5, reproducible on an unmodified tree. `phpcs` and Psalm self-analysis clean.


alies-dev:
A note on a **remaining, pre-existing** gap that this PR does not close, so #11958 should not be treated as fully resolved by it:

`TypeParser::parseTokens()` can also throw `InvalidArgumentException`, not just `TypeParseTreeException`. A degenerate parenthesised `is` expression that is not a complete ternary still aborts the run past the new catch:

```php
/** @phpstan-this-out ($key is null) */
public function s(?int $key): void {}
```

```
Uncaught InvalidArgumentException: Unrecognised parse tree type Psalm\Internal\Type\ParseTree\TemplateIsTree
  in src/Psalm/Internal/Type/TypeParser.php:402
```

This is **not a regression from this PR** and not specific to `@self-out`: the identical input kills the run through the already-guarded `@return` path too (verified locally on an unmodified tree), because `src/Psalm/Internal/Type/TypeParser.php:402` raises `InvalidArgumentException` for any unhandled parse tree, which none of the eight docblock catch sites expect.

Fixing it belongs in `TypeParser` rather than here, so that all eight call sites benefit at once — e.g. raising `TypeParseTreeException` instead of `InvalidArgumentException` for an unrecognised tree. Happy to follow up with that as a separate PR if maintainers agree on the approach.
___
UPD: I created a new PR to ~fix~ bypass that issue too: https://github.com/vimeo/psalm/pull/11962. Preper fix requires a bit more work (parameter-conditional type support for @self-out / @this-out)

PR #11961, opened 2026-09-15
[7.x] Report every taint flow into a sink, not just the first to arrive

### LINKED ISSUE #11959: [7.x] Taint: one of two flows into the same sink is dropped when one path is longer (visited guard also suppresses reporting)
When two taint flows reach the same sink node with the same taint mask, only the flow that arrives first is reported. The `visited_source_ids` guard in `TaintFlowGraph::getChildNodes()` sits above the sink-reporting block, so it suppresses *reporting* as well as *propagation*. It only bites when the two flows arrive in different resolution rounds, i.e. when one path is longer than the other.

This is a latent bug that pre-dates `7.0.0-beta20`. What changed in `beta21` is round parity, not the resolver: `6ff035824` ("Strip nosql taint on string casts") inserts an extra `string-cast` pass-through node on one of the two paths, pushing that flow into a later round.

I originally filed this as a plugin-dependent beta20 → beta21 regression. Both halves of that framing were wrong: the plugin is not involved at all, and `beta20` is affected too. Title and description updated accordingly.

### Reproducer, no plugin, no Laravel

`somefile.php`:

```php
<?php

/**
 * @psalm-flow ($value) -> return
 */
function relay(string $value): string { return $value; }

/**
 * @psalm-taint-sink shell $cmd
 */
function runCmd(string $cmd): void {}

runCmd((string)($_GET["direct"] ?? ""));
runCmd(relay((string)($_GET["relayed"] ?? "")));
```

```bash
vendor/bin/psalm --taint-analysis --no-cache --threads=1 --no-progress --no-diff
```

Actual: one `TaintedShell`.
Expected: two `TaintedShell`, one per flow.

Both calls pass tainted input into `runCmd()`'s shell sink. The second goes through `relay()`, so it is one hop longer and reaches the shared sink node one round later with an identical taint mask. Fails on `7.0.0-beta20` and on `master` alike.

### Evidence from the resolver

Logging every sink arrival in `getChildNodes()` on the original Laravel reproducer (kept below):

```
beta20:  SINK to=…pendingrequest::get#1 taints=124927 match=512 from=…b.php:703-739
         SINK to=…pendingrequest::get#1 taints=124927 match=512 from=…a.php:363-389
master:  SINK to=…pendingrequest::get#1 taints=124927 match=512 from=…b.php:703-739
         SKIP-VISITED to=…pendingrequest::get#1 taints=124927 from=…a.php:363-389
```

Same sink node, same mask `124927`, same matching bit `512` (`ssrf`). On `beta20` both arrive in one round and both report. On `master` the a.php flow arrives a round later and is dropped by

```php
if (isset($visited_source_ids[$to_id][$new_taints])) {
    continue;
}
```

`git bisect` between the two tags points at `6ff035824`, which only changes `CastAnalyzer`: the implicit `mixed` → `string` cast for `PendingRequest::get(string $url)` now routes the value through a `string-cast` node, adding the hop that changes the parity.

### Related: the same guard duplicates findings

`IssueBuffer::add()` skips deduplication for taint issues (`if ($is_tainted || !self::alreadyEmitted($emitted_key))`), so the same flow expanded under two different `specialized_calls` sets reports twice at the same file, line and column. That explains the duplicate `TaintedShell` I saw in the same batch.

### Fix

Separate "do not re-propagate" from "do not report" in `TaintFlowGraph::getChildNodes()`:

- Keep an early `continue` when the destination is already visited **and is not a sink**, so the hot path stays as cheap as it is now.
- A visited **sink** falls through to the reporting block, because a later-arriving flow into it is a finding of its own.
- Deduplicate reporting on the flow's own endpoints — `(source id, sink id, taints)` — which yields one finding per origin/sink/mask pair and also removes the duplicate reports described above.
- Guard only the frontier enqueue with `$already_visited`, so the fixed point still terminates.

Verified: both findings restored on the plugin-free reproducer and on the Laravel one, on top of `master`. `tests/TaintTest.php` 259/259 and `tests/EndToEnd` 16/16 pass; the remaining full-suite failures (`InternalCallMapHandlerTest` on PHP 8.5, `ReportOutputTest::testSarifReport`) are pre-existing and reproduce unpatched.

<details>
<summary>Original Laravel-based report (superseded by the reproducer above)</summary>

Between `7.0.0-beta20` and `7.0.0-beta21`, two taint flows that enter the same sink method from two different files collapse into one report: the second finding is dropped. Analyzed separately, each file reports its own finding on both versions.

I could not reduce this below a `psalm-plugin-laravel` dependency, so the reproducer below needs the plugin.

#### Setup

```bash
composer require --dev vimeo/psalm:7.0.0-beta21 psalm/plugin-laravel:dev-master#ece3ad4cdf0b9e781676c0ef4ebcdb2bd33cf5e1
```

The plugin must be `dev-master`, not a release: beta21 requires `psalm/psalm-plugin-api` 0.2.0, and the last plugin release still requires ^0.1.0, so it is not installable alongside beta21.

`laravel/framework` has to be pinned below v13.32.0, because v13.32.0 introduces a docblock that crashes Psalm's scan phase outright (#11958). I used v13.31.0.

`a.php`:

```php

use Illuminate\Http\RedirectResponse;
use Illuminate\Http\Request;
use Illuminate\Validation\Rule;

/** @psalm-suppress MixedArgument */
function store(Request $request): RedirectResponse {
    $request->validate([
        'reply_to' => ['required', Rule::email()],
    ]);

    (new \Illuminate\Http\Client\PendingRequest())->get($request->input('reply_to'));
    return redirect()->to($request->input('reply_to'));
}
```

`b.php`:

```php

namespace App\Http\Requests;

use Illuminate\Contracts\Validation\ValidationRule;
use Illuminate\Foundation\Http\FormRequest;

/**
 * @psalm-taint-escape header
 * @psalm-taint-escape cookie
 */
final class DnsEmailRule implements ValidationRule
{
    #[\Override]
    public function validate(string $attribute, mixed $value, \Closure $fail): void {}
}

final class ContactRequestNew extends FormRequest
{
    public function rules(): array
    {
        return ['team_email' => ['required', 'string', new DnsEmailRule()]];
    }
}

function direct(ContactRequestNew $request): \Illuminate\Http\RedirectResponse {
    (new \Illuminate\Http\Client\PendingRequest())->get($request->safe()->input('team_email'));
    return redirect()->to($request->safe()->input('team_email'));
}
```

`psalm.xml`:

```xml
<?xml version="1.0"?>
<psalm errorLevel="1" runTaintAnalysis="false" xmlns="https://getpsalm.org/schema/config">
    <projectFiles>
        <file name="a.php"/>
        <file name="b.php"/>
    </projectFiles>
    <plugins>
        <pluginClass class="Psalm\LaravelPlugin\Plugin"/>
    </plugins>
</psalm>
```

#### Run

```bash
vendor/bin/psalm --taint-analysis --no-progress --no-diff --output-format=compact
```

Actual, `7.0.0-beta21`:

```
ERROR b.php:27:57 TaintedSSRF: Detected tainted network request
```

Expected, and what `7.0.0-beta20` reports:

```
ERROR a.php:13:57 TaintedSSRF: Detected tainted network request
ERROR b.php:27:57 TaintedSSRF: Detected tainted network request
```

#### Scope

The plugin's taint fixture suite co-analyzes 348 files. Re-running it single-threaded gives 775 issues on beta20 and 764 on beta21. Ten `TaintedSSRF` findings are lost cross-file in the way shown above, and one `TaintedShell` is reported twice at the same file, line and column.

</details>

alies-dev:
Fixes #11959.

### Problem

When two taint flows reach the same sink node with the same taint mask, only the one that arrives first is reported. The `visited_source_ids` guard in `TaintFlowGraph::getChildNodes()` sat above the sink-reporting block, so it suppressed *reporting* as well as *propagation*:

```php
if (isset($visited_source_ids[$to_id][$new_taints])) {
    continue;
}
```

`visited_source_ids` is marked at expansion time, so a sink that entered the frontier on its first arrival is expanded — and marked visited — one round later. Any second flow arriving in a later round with the same mask is dropped before it can report.

It only bites when the two paths have different lengths. That is a latent bug, not a beta21 regression: it reproduces on `7.0.0-beta20` too. What `beta21` changed is round parity — `6ff035824` ("Strip nosql taint on string casts") inserts a `string-cast` pass-through node on one of the two paths.

Plugin-free reproducer (fails on `beta20` and on `master` alike, one `TaintedShell` instead of two):

```php
<?php

/** @psalm-flow ($value) -> return */
function relay(string $value): string { return $value; }

/** @psalm-taint-sink shell $cmd */
function runCmd(string $cmd): void {}

runCmd((string)($_GET["direct"] ?? ""));
runCmd(relay((string)($_GET["relayed"] ?? "")));
```

### Fix

Separate "do not re-propagate" from "do not report":

- A visited destination that is **not** a sink still hits an early `continue`, so the hot loop stays as cheap as before; the extra `isset($sinks[$to_id])` is only evaluated for already-visited nodes.
- A visited **sink** falls through to the reporting block, because a later-arriving flow into it is a finding of its own.
- Reporting is deduplicated on `(flow origin, sink id, taints)`, where the origin is the root of the node's `taintSource` chain (`getFlowOrigin()`), computed only for an edge that actually matches sink taints.
- Only the frontier enqueue is guarded by `$already_visited`, so the fixed point still terminates; `$reported_flows` is bounded by the same finite triple space.

Nothing between the guard and the enqueue has side effects beyond the fetch-ignore checks and reporting itself, so re-entering that stretch for a visited sink is safe.

Keying on the origin rather than on the sink's immediate predecessor matters because the predecessor is shared by every call site that funnels through the same (possibly de-specialized) node. It also still collapses the same flow re-expanded under two different `specialized_calls` sets into one finding, which is what removes the duplicate `TaintedShell` reported at identical file/line/column (`IssueBuffer::add()` skips dedup for taint issues).

### Behaviour note

Flows from distinct origins that converge at an intermediate, non-sink node still collapse to one finding: the propagation guard merges them upstream of the sink, so the second flow never reaches it. That is pre-existing fixed-point behaviour, unchanged here — verified by a shared-wrapper case that reports one issue both with and without this patch.

### Verification

- New regression test `twoFlowsOfDifferentLengthIntoTheSameSink` in `tests/TaintTest.php`: fails unpatched (1 `TaintedShell`), passes patched (2). Checked in both directions.
- `tests/TaintTest.php`: 259/259. `tests/EndToEnd`: 16/16.
- The original Laravel reproducer from #11959 reports both `a.php:13:57` and `b.php:27:57` again on top of `master`.
- Remaining full-suite failures (`InternalCallMapHandlerTest` on PHP 8.5, `ReportOutputTest::testSarifReport`) are pre-existing and reproduce without this patch.
- `psalm` self-analysis and `phpcs` clean on the changed files.


danog:
bugbot review

PR #11962, opened 2026-09-15
[6.x] Throw `TypeParseTreeException` for an incomplete conditional type

alies-dev:
> [!NOTE]
> **Stacked on #11960.** The base here is `6.x` because a fork branch cannot be used as a PR base, so the diff currently shows both commits. Only the second one — `Throw TypeParseTreeException for an incomplete conditional type` — belongs to this PR. I'll rebase once #11960 lands. Happy to fold it into #11960 instead if you prefer a single PR.

### Problem

Follow-up to the gap I noted in https://github.com/vimeo/psalm/pull/11960#issuecomment-5684881368.

`TypeParser::getTypeFromTree()` handles `ConditionalTree` but has no branch for `TemplateIsTree`, the node `ParseTreeCreator` builds for an `is` token before a `?` upgrades it into a `ConditionalTree`. An `is` expression that never gets its ternary therefore falls through to the invariant guard at the bottom:

```php
if (!$parse_tree instanceof Value) {
    throw new InvalidArgumentException('Unrecognised parse tree type ' . $parse_tree::class);
}
```

None of the docblock parse sites catch `InvalidArgumentException`, so this killed the entire run:

```php
/** @return ($key is null) */
public function c(?int $key) { return null; }
```

```
Uncaught InvalidArgumentException: Unrecognised parse tree type Psalm\Internal\Type\ParseTree\TemplateIsTree
  in src/Psalm/Internal/Type/TypeParser.php:402
```

This was not specific to `@self-out` — `@return`, `@param-out` and `@psalm-if-this-is` all died on the same input, even though all three already have a `catch (TypeParseTreeException)`.

### Change

Reject `TemplateIsTree` in `getTypeFromTree()` with a `TypeParseTreeException`, so every one of the eight docblock call sites degrades to `InvalidDocblock` as intended:

```
ERROR: InvalidDocblock - dangling.php:11:5 - Invalid conditional, expected ? after is in docblock for F::c
```

Handling it in `getTypeFromTree()` rather than in `ParseTreeCreator::create()` is deliberate: `getTypeFromTree()` already recurses, so nested occurrences such as `list<($key is null)>` are covered for free, whereas a tail check in `create()` would only see the root node and would need a full tree walk.

The `InvalidArgumentException` at the bottom of `getTypeFromTree()` is left alone. I probed the other `ParseTree` subclasses that reach it (`Root`, `FieldEllipsis`, `CallableParamTree`, `MethodParamTree`, `KeyedArrayPropertyTree`) with malformed docblocks — `array{a: int, ...}`, `list<...>`, `callable(int...)`, `callable(int):`, `array{a:}`, `Foo::bar()`, `int...`, `...` — and none of them are reachable from user input; they are either consumed by their parent node or rejected earlier with a `TypeParseTreeException`. `TemplateIsTree` was the only real hole, so the guard stays as a genuine internal invariant.


alies-dev:
Pushed a second commit, 94eca6323, that closes a sibling hole Copilot spotted on #11964 (https://github.com/vimeo/psalm/pull/11964#pullrequestreview-5216373689). It is pre-existing on `6.x`, not introduced by any PR in this stack:

```php
/** @return ($key is ? int : string) */
```

crashes the scan with `Undefined array key 0` in `TypeParser::getTypeFromTree()`. The `?` promotes the childless `TemplateIsTree` into a `ConditionalTree` before the top-level guard from the first commit gets a chance to see it, and the `ConditionalTree` branch then reads `condition->children[0]` unchecked. The new commit adds the missing count check next to the existing `children !== 2` one, so it degrades to `InvalidDocblock` like the other incomplete forms. Covered by a `TypeParseTest` case and an `InvalidDocblock` case in `ConditionalReturnTypeTest`, both confirmed to crash the run without the fix. The guard also makes one `psalm-baseline.xml` entry redundant, so that line is dropped.

Issue #10126, opened 2023-08-18
Conditional @param-out broken - most cases of psalm's phpstubs affected


kkmuffme:
https://psalm.dev/r/14a1e38511
(copy of `preg_match` from psalm's stubs)

Line 22 is wrong
But the real problem is line 24 `$value: mixed` which is not correct - it completely ignores the @param-out it seems?

I assume this worked at some point previously, since it's used extensively in CoreGenericFunctions.phpstub
Atm all these stubs by-reference types are broken.

How are no tests failing here, since all the conditional @param-out (or @psalm-param-out) do not work correctly?

weirdan:
Apparently built-in functions are handled differently: https://psalm.dev/r/dde5e62304

weirdan:
But if you reverse the arguments, the conditional type is resolved: https://psalm.dev/r/f20095f04a

Coincidentally this what Psalm does internally for global functions called `preg_match()` and `preg_match_all()` (and only those):

https://github.com/vimeo/psalm/blob/bf8e150f8ff4199171c1ea6ca88bbf1f27b14c66/src/Psalm/Internal/Analyzer/Statements/Expression/Call/ArgumentsAnalyzer.php#L693-L695

PR #11964, opened 2026-09-15
[6.x] Support parameter-conditional types in `@self-out` / `@this-out`

alies-dev:
> [!NOTE]
> **Stacked on #11960 and #11962.** The base is `6.x` because a fork branch cannot be a PR base, so the diff currently shows all three commits. Only the last one, `Support parameter-conditional types in @self-out/@this-out`, belongs to this PR. I'll rebase once the other two land.

## Context

The crash in #11958 came from Laravel's `Collection::prepend()` docblock:

```php
/** @phpstan-this-out ($key is null ? static<TKey|int, TValue> : $this) */
public function prepend(mixed $value, ?int $key = null): static
```

#11960 turned that crash into an `InvalidDocblock`, but the tag is still rejected with `Unrecognized template '$key'`. The reason is a gap in the scanner only: `@return` sends its type through `getConditionalSanitizedTypeTokens()`, which rewrites `$param` references into generated `TGeneratedFromParamN` template params and registers them on the method, while the `self_out` block parsed the raw tokens, so `ConditionalTree` looked up the literal `$key` and failed. The same conditional on `@return` was already accepted, and so was `@psalm-self-out (T is null ? ...)` over an explicit `@template T`, which resolved correctly after the call. Only the parameter form was missing.

## Solution

The `self_out` block in `FunctionLikeDocblockScanner::addDocblockInfo()` now runs the same sanitizer `handleReturn()` uses and reassigns the returned `$function_template_types` so the later `if_this_is`, `handleRemovedTaint` and `handleReturn` steps see the generated params. That is the whole source change (eight lines). Nothing at the call sites needed touching: `ExistingAtomicMethodCallAnalyzer` already runs `checkMethodArgs()` before applying `self_out_type` through `MethodCallReturnTypeFetcher::replaceTemplateTypes()`, so the `TGeneratedFromParamN` lower bounds are filled from the actual arguments and `TemplateInferredTypeReplacer` collapses the conditional to the right branch. `NewAnalyzer` does the same for constructors. When the argument cannot decide the branch (a `?int` variable, say) the result is the union of both branches, never a leaked `TConditional`. Collisions between a conditional `@return` and a conditional `@self-out` on one method are not possible: the generated name is keyed on the parameter index and the sanitizer reuses an existing entry rather than creating a second one.

`@psalm-if-this-is` is deliberately left out. It is checked against `$this` before the call's arguments are analyzed (and again in `FunctionLikeAnalyzer` with no arguments at all), so a conditional on those arguments can never be decided there. A parameter conditional on it still degrades to `InvalidDocblock`, which is now pinned by a test.

Reusing the sanitizer has a few side effects worth knowing about. Relative names in `@self-out` are now resolved the way `@return` resolves them, so `@psalm-self-out self` inside a trait yields the using class instead of the trait (pinned by a test). `func_num_args()` and the PHP version constants now work in `@self-out` conditionals too. A parameter referenced by the conditional gets its type rewritten to a template param, so inside the method body an `int $key` reads as `int|float`; that is pre-existing behaviour of conditional `@return` and not new here.

Still unsupported: conditional `@param-out` has the identical gap (#10126) and remains `InvalidDocblock`. That is a separate PR.

Tests cover the Laravel shape with the null and non-null branches asserted separately (`Coll<int|string, bool>&static` vs `Coll<string, bool>&static`), a method with both a conditional `@return` and a conditional `@self-out`, the constructor path via `new`, the trait `self` case, and the three failure modes (`garbage<<<`, an undeclared parameter, an incomplete `($key is null)`). The full suite passes apart from the pre-existing PHP 8.5 `InternalCallMapHandlerTest` reflection mismatches. A CLI run against a scratch project with the real Laravel docblock traces both branches correctly.

Refs #11958.


alies-dev:
Re the suppressed Copilot note on `TypeParser.php` (https://github.com/vimeo/psalm/pull/11964#pullrequestreview-5216373689): confirmed, and it is not new to this stack. `@return ($key is ? int : string)` crashes the scan on current `6.x` with `Undefined array key 0`, because `?` promotes the childless `TemplateIsTree` into a `ConditionalTree` before the top-level guard runs. This PR only made the same path reachable through `@self-out`.

Fixed in #11962 (commit 94eca6323, see https://github.com/vimeo/psalm/pull/11962#issuecomment-5689108583), which is where the other incomplete-conditional guard lives. Merged that branch forward here and added `conditionalWithoutIsTypeIsReportedInsteadOfCrashing` to `ThisOutTest`, which crashes without the guard and reports `InvalidDocblock` with it.

